PR #62, opened 2018-05-02
DeveloperGuide: add setup instructions

eugenepeh:
To get the ball rolling -> borrowed most of it from addressbook 😛

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -1,11 +1,46 @@
 # Developer Guide
 Thank you for contributing to RepoSense!
+- [Setting Up](#setting-up)
+- [Architecture](#architecture)
+- [Notable POJOs](#notable-pojos)
+- [Frontend Components](#frontend-components)
+- [Dashboard Generator Components](#dashboard-generator-components)
+- [Git Utilities](#git-utilities)
+- [Analyzer Components](#analyzer-components)
+- [HTML Dashboard](#html-dashboard)
+
+## Setting up
+
+### Prerequisites
+
+1. **JDK `1.8.0_60`**  or later <br>
+> This app only works with Java 8 version. <br>
+2. **Git** on the command line <br>
+> Ensure that you're able to use it on the OS terminal. <br>
+3. **IntelliJ** IDE [Optional]
+
+### Setting up the project in your computer using IntelliJ
+1. Fork this repo, and clone the fork to your computer.
+2. Open IntelliJ (if you are not in the welcome screen, click `File` > `Close Project` to close the existing project dialog first).
+3. Set up the correct JDK version for Gradle.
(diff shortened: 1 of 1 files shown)


eugenepeh:
@harryggg could you help me review see if I missed out anything in the setup instructions?

docs/developer_guide.md
```diff
@@ -1,56 +1,92 @@
...
+### 1.1. Prerequisites
+
+1. **JDK `1.8.0_60`**  or later<br>
+> This app only works with Java 8 version. <br>
+
+2. **IntelliJ** IDE
+3. **Git** on the command line
+> Ensure that you're able to use it on the OS terminal. <br>
+
+### 1.2. Setting up the project in your computer
```
damithc: Quick comment: may be not use section numbers? Markdown cannot generate section numbers automatically. Hard-coding them means it is hard to insert/delete sections later.

eugenepeh: hmm, true. In that case, shall we port over to asciidoc while its still in the early stage?

docs/developer_guide.md
```diff
@@ -1,11 +1,47 @@
...
+- [HTML Dashboard](#html-dashboard)
+
+## Setting up
+
+### Prerequisites
+
+1. **JDK `1.8.0_60`**  or later<br>
+> This app only works with Java 8 version. <br>
+
+2. **IntelliJ** IDE
```
harryggg: Why are we restricting IDE to IntelliJ

harryggg:
Sorry for the late reply... will try to reply within 12 hours

docs/developer_guide.md
eugenepeh: hmm, restricting is not my intention but a common IDE will probably be easier for multiple developers to work together. IntelliJ is the dedicated IDE for java in the jetbrains suite of tools. 
I'm open to suggestions though.

harryggg: In the context of this project, I cannot think of any specific advantages for unifying IDE... But I think it will be good for beginners! 
The other two requirements are JDK and Git, which are both compulsory, so it might imply that IntelliJ is compilsory too. It might cause some confusion.
Maybe mark it as optional to avoid confusion? Thx

damithc: Yes, we can make Intellij an example only.

eugenepeh: > In the context of this project, I cannot think of any specific advantages for unifying IDE... But I think it will be good for beginners!
The other two requirements are JDK and Git, which are both compulsory, so it might imply that IntelliJ is compilsory too. It might cause some confusion.
Maybe mark it as optional to avoid confusion? Thx

Alright

eugenepeh:
Amended IntelliJ to optional - is there anything you'd like to add to verifying setup?

harryggg:
LGTM. Thx!

PR #64, opened 2018-05-05
Setup Travis CI

eugenepeh:
Reference: https://github.com/eugenepeh/RepoSense/pull/2


eugenepeh:
@harryggg Just like check

<pre>
analyzer.AnnotatorAnalyzerTest STANDARD_OUT
    Directory does not exist.
    Directory does not exist.
    cloning harryggg/GitTest...
    cloning done!
</pre>

Got all these messages from running tests on gradle. What directory is missing?

damithc:
> Got all these messages from running tests on gradle. What directory is missing?

Most likely just an info message.  If that is the case, we should rephrase it to avoid appearing like an error message.

harryggg:
The message is shown when there the repository is not found, which means RepoSense needs to clone it first. I will rephrase it now.

harryggg:
LGTM.

PR #65, opened 2018-05-05
Travis: add checkstyle to CI test

eugenepeh:
Proposed commit message:

When writing a computer program it is important to keep in mind that the program should not only run correctly, but also be written in a clear way so that other people can understand it.

This can be enforced by incorporating Checkstyle, which is a development tool that help programmers write Java code that adheres to a coding standard, into Travis CI.

Let's add Checkstyle to the Travis CI configuration to automate the process.

### DIFF
--- a/src/main/java/analyzer/AnnotatorAnalyzer.java
+++ b/src/main/java/analyzer/AnnotatorAnalyzer.java
@@ -1,17 +1,14 @@
 package analyzer;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import dataObject.Author;
 import dataObject.FileInfo;
 import dataObject.LineInfo;
 import dataObject.RepoConfiguration;
 
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
-
-/**
- * Created by matanghao1 on 5/8/17.
- */
 public class AnnotatorAnalyzer {
 
     private static final String AUTHOR_TAG = "@@author";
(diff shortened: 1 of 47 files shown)


src/main/java/analyzer/AnnotatorAnalyzer.java
```diff
@@ -1,14 +1,13 @@
...
 import dataObject.FileInfo;
 import dataObject.LineInfo;
 import dataObject.RepoConfiguration;
 
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
-
 /**
  * Created by matanghao1 on 5/8/17.
```
damithc: Remove these too? We don't usually mention author or date in header comments

src/main/java/analyzer/AnnotatorAnalyzer.java
```diff
@@ -19,30 +18,35 @@
     private static final Pattern PATTERN_AUTHOR_NAME_FORMAT = Pattern.compile(REGEX_AUTHOR_NAME_FORMAT);
     private static final int MATCHER_GROUP_AUTHOR_NAME = 1;
 
-
+    /**
+     *
+     */
```
damithc: I'd rather not have empty header comments. We can hold back the rule to require header comments for all public members if it is too much work to enforce retroactively.

src/main/java/analyzer/AnnotatorAnalyzer.java
eugenepeh: Okay

PR #68, opened 2018-05-07
Gradle: replace build with shadowJar plugin

eugenepeh:
Proposed commit message:

Gradle build compiles our source files and libraries used in the program into separate jar files, and delivers the entire package as a zip file.

It is rather inconvenient as user would have to extract from the zip file to use it and also have more unnecessary files to deal with. Additionally, if we package only our own class files into the JAR file, it will not work properly unless the user has all the other JAR files (i.e. third party libraries) our classes depend on.

Let's replace the gradle build with shadowJar plugin to package all dependencies into a single JAR file which enhance usability and portability.

### DIFF
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -12,6 +12,7 @@
 
 ## How to Generate Dashboard
 1. Download the latest executable Jar on our [release](https://github.com/reposense/RepoSense/releases/latest).
+   * Alternatively, you can compile the executable Jar yourself by following our [build from source guide](Build.md).
 2. Execute it on the OS terminal. <br>
 Usage: `java -jar RepoSense.jar -config CSV_CONFIG_FILE_PATH [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY]`
 3. The dashboard can be in the folder designated in OUTPUT_DIRECTORY, or current working directory otherwise, as index.html.
--- a/docs/user_guide.md
+++ b/docs/user_guide.md
@@ -5,16 +5,20 @@
 4. Profit!
 
 ## Dependencies
-1. Git
-2. Gradle
+1. **JDK `1.8.0_60`**  or later
+   * This app only works with Java 8 version.
+2. **Git** on the command line
+   * Ensure that you're able to use it on the OS terminal.
+
 ## How to Generate Dashboard
-```sh
-$ gradle clean build
+1. Download the latest executable Jar on our [release](https://github.com/reposense/RepoSense/releases/latest).
+2. Execute it on the OS terminal. <br>
+Usage: `java -jar RepoSense.jar -config CSV_CONFIG_FILE_PATH [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY]`
+3. The dashboard can be in the folder designated in OUTPUT_DIRECTORY, or current working directory otherwise, as index.html.
+
+Sample usage:
 ```
-Unzip the file in build/distributions/
-```
-$ cd build/distributions/RepoSense/bin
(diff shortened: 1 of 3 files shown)


damithc:
Give a detailed commit message per PR?

eugenepeh:
On hold until #77 is merge so that we can make a v1.0 release.

docs/user_guide.md
```diff
@@ -5,16 +5,20 @@
...
+1. Download the latest executable Jar on our [release](https://github.com/reposense/RepoSense/releases/latest).
+2. Execute it on the OS terminal. <br>
+Usage: `java -jar RepoSense.jar -config CSV_CONFIG_FILE_PATH [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY]`
+3. The dashboard can be in the folder designated in OUTPUT_DIRECTORY, or current working directory otherwise, as index.html.
+
+Sample usage:
 ```
-$ cd build/distributions/RepoSense/bin
-$ ./RepoSense -config CSV_path.csv -output output_path/ -since 01/10/2017 -until 01/11/2017
+$ java -jar RepoSense.jar -config CSV_path.csv -output output_path/ -since 01/10/2017 -until 01/11/2017
```
damithc: shall we document both 'run from executable' and 'build from source' documentation?

docs/user_guide.md
```diff
@@ -5,16 +5,20 @@
 4. Profit!
 
 ## Dependencies
-1. Git
-2. Gradle
+1. **JDK `1.8.0_60`**  or later <br>
+> This app only works with Java 8 version. <br>
```
damithc: Let's not use 'quote' formatting for things that are not really quotes. 
No need for `<br>` if you use correct indentation. e.g.,

1. **JDK `1.8.0_60`**  or later
   This app only works with Java 8 version.

```
1. **JDK `1.8.0_60`**  or later
   This app only works with Java 8 version.
```

docs/user_guide.md
eugenepeh: hmm ok

docs/user_guide.md
eugenepeh: Hmm, I think it only works in comment. It works even without indentation.

damithc: Try this (extra line break)
```
1. **JDK `1.8.0_60`**  or later

   * This app only works with Java 8 version.
```

BTW, the app doesn't work for Java 9, 10?

eugenepeh: I mean the code quote version works but not this one without `<br>` 

1. **JDK `1.8.0_60`**  or later
   This app only works with Java 8 version.

> BTW, the app doesn't work for Java 9, 10?

Yup, seems like the code is build in Java 8 so we will need to upgrade to Java 9 / 10 for it to work with latest java jdk. Meanwhile the workaround is by overwriting the system variable using
`set JAVA_HOME=JAVA_8_JDK_DIRECTORY` in windows.

eugenepeh: >A java 8 app should work within Java 9 or 10 unless we use some backward incompatible library or API. Did you try with later versions?

Yup, my default jdk is java 9. I'm unable to build it with java 9 but I have not investigate on the cause, so currently I'm still using the workaround.

eugenepeh: image

Seems like its due to gradle out-of-date

damithc: Got it. In this PR we can require users to use Java 8. Hopefully, we can make it compatible with Java 9 soon.

eugenepeh:
>Got it. In this PR we can require users to use Java 8. Hopefully, we can make it compatible with Java 9 soon.

On it.

For your final review. Should I make a v1.0 release directly after this PR or would you like me to upgrade to java 9 first?

damithc:
> For your final review. Should I make a v1.0 release directly after this PR or would you like me to upgrade to java 9 first?

Can do the release first.

PR #77, opened 2018-05-09
Remove GUI #70

### LINKED ISSUE #70: Remove GUI
Remove GUI code for the time being. Can bring back later if needed.

eugenepeh:
Fix #70 

Proposed message:

Gui mode is inactivate-able without using gradle. This requires user
to download and execute the production code, which makes it not user-
friendly.

Let's focus on console mode and put gui production on hold.


harryggg:
Should we just leave the class there, just in case in a distant future we want GUI back?

eugenepeh:
We can always look up the git history when we're ready to bring it back. 
Leaving it there might confuse other contributors.

damithc:
> We can always look up the git history when we're ready to bring it back.
> Leaving it there might confuse other contributors.

Yes, let's aggressively remove unused code but make each removal a separate PR so that it is easy to reverse them later.

eugenepeh:
>Yes, let's aggressively remove unused code but make each removal a separate PR so that it is easy to reverse them later.

Done!

eugenepeh:
@damithc for your final review

harryggg:
That help message is super helpful.. can u show it in CLI too, when users input is not valid? Thank you so much

eugenepeh:
>User guide needs updating right?
I don't see any GUI code being deleted. Delete all GUI-related code that is not used.

Oh, I thought your earlier comment `make each removal a separate PR so that it is easy to reverse them later.` meant this PR. Will revert it.

> That help message is super helpful.. can u show it in CLI too, when users input is not valid? Thank you so much

Sure.

eugenepeh:
>User guide needs updating right?

The user guide update is in shadowJar PR. I see no need to update for this PR because there's error in current one anyway. So it would be better to package it together with our first v1.0 release.

damithc:
> The user guide update is in shadowJar PR. I see no need to update for this PR because there's error in current one anyway. So it would be better to package it together with our first jar release.

I mean the online user guide. If you don't update it, the user guide becomes inconsistent with the product. Note that some people who want to live on the bleeding edge will build the product from code rather than use the jar file.

eugenepeh:
>I mean the online user guide. If you don't update it, the user guide becomes inconsistent with the product. Note that some people who want to live on the bleeding edge will build the product from code rather than use the jar file.

Updated the user guide as requested.

docs/user_guide.md
```diff
@@ -8,19 +8,14 @@
 1. Git
 2. Gradle
 ## How to Generate Dashboard
-1. GUI
-```sh
-$ gradle clean build
-$ gradle clean run
-```
-2. CLI
+Using CLI
```
damithc: This line is redundant, as this is the _only_ way to use the app.

eugenepeh:
>Ideally, the help message should have been a separate PR -- because we squash at merge, each PR should contain only one change.

Agree, added into this PR to replace it the default functionality of no argument though.

PR #80, opened 2018-05-10
 Prefix all packages with organisation name

eugenepeh:
Proposed message:

Prefix our packages with organisation name ensures uniqueness. It also ensures that our packages will be grouped together, following our import order coding standard.

Let's prefix all our packages name with our organisation name.


damithc:
> Prefix our packages with organisation name ensures uniqueness. It also ensures that our packages will be grouped together, following our import order coding standard.

Suggested:
~~Prefix~~ Prefixing our packages with ~~organisation~~ organization name prevents name clashes. It also ensures that our packages will be grouped together, following our import order coding standard.

PR #83, opened 2018-05-10
Gradle: upgrade to version 4.7

### LINKED ISSUE #81: Update gradle to latest version
image

Gradle does not recognise Java 9 version. 

This is likely due to gradle version being out-dated for too long.

Let's update gradle to latest version so that it is compatible with the newer version of Java.

eugenepeh:
Fix #81. Seems easier than I thought... have nothing else to fix other than url 🤣 

Proposed message:
```
Gradle does not recognise Java 9 version.

This is likely due to gradle version being out-dated for too long.

Let's update gradle to latest version so that it is compatible with the newer version of Java.
```

Onhold till #68 is merged so that I can update the docs too.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,11 +19,12 @@ test {
 }
 
 dependencies {
-    compile group: 'commons-collections', name: 'commons-collections', version: '3.2'
-    compile 'com.github.javaparser:javaparser-core:3.0.1'
-    compile 'com.google.code.gson:gson:2.6.2'
-    testCompile group: 'junit', name: 'junit', version: '4.12'
-    compile group: 'org.json', name: 'json', version: '20180130'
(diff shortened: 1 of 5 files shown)


eugenepeh:
jdk10 is not available in travis yet as mentioned in https://github.com/travis-ci/travis-ci/issues/9368

build.gradle
```diff
@@ -41,7 +42,7 @@ tasks.shadowJar.dependsOn("myZip");
 tasks.run.dependsOn("myZip");
 
 configurations {
-  functionalCompile.extendsFrom testCompile
+  functionalImplementation.extendsFrom testImplementation
```
damithc: A side comment: We should use a terms such as 'System Tests' instead of 'functional'.

eugenepeh: Opened issue #85

PR #97, opened 2018-05-15
Refactor Client class into multiple parser classes

yong24s:
Not ready for review.

For discussion, please.


damithc:
Refactor :-)

src/main/java/reposense/ConfigParser/Argument.java
```diff
@@ -0,0 +1,52 @@
+package reposense.ConfigParser;
+
+import java.io.File;
+import java.util.Date;
+import java.util.Optional;
+
+public final class Argument {
```
eugenepeh: if the arguments are literate-able, I think it should be `Configurations` rather than `Argument`

src/main/java/reposense/ConfigParser/CliArgumentsParser.java
```diff
@@ -0,0 +1,82 @@
...
+            if (!args[i].startsWith(ARGUMENT_PREFIX_DASH)) {
+                throw new ParseException(PARSE_EXCEPTION_MESSAGE_INVALID_INPUTS);
+            }
+
+            String key = args[i].substring(1);
+            argumentMap.put(key, args[i + 1]);
+        }
+
+        if (isAllMandatoryArgumentsPresent()) {
+            throw new ParseException(PARSE_EXCEPTION_MESSAGE_NO_CSV_FILE);
```
eugenepeh: `PARSE_EXCEPTION_MESSAGE_NO_CSV_FILE` should be only a subset of `isAllMandatoryArgumentsPresent`

src/main/java/reposense/ConfigParser/CliArgumentsParser.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Parses user-supplied arguments into an Argument object.
+     *
+     * @return Argument object, which contains the user supplied arguments.
+     * @throws ParseException If the given inputs or the csv file fail to parse or mandatory fields are missing.
+     */
+    @Override
+    public Argument parse(String[] args) throws ParseException{
+        for (int i = 0; i < args.length; i = i + 2) {
```
eugenepeh: for future expand-ability, I think we shouldn't limit to 1 value per parameter

yong24s:
As suggested, I will close this PR and open a new one.

PR #102, opened 2018-05-16
splitting the results files & converting the results into JSON format

ongspxm:
proposed message

```
We wanna generate the report for each repo iteratively to improve
performance. Here is a proposed structure that we can adopt for future
frontend work.

Files are converted into JSON formatted, rather than the initial JS
script which loads all the information into a pre-defined variable
name.

Folders are reorganized to support the multiple reports format. This
potentially frees up the reporting capabilities of reposense, making it
possible for other developers to extend upon the generated results.

Template is updated to display the separated results through the means
of artifically merging them through dummy files like `summary.js` and
`report.js`.

Let's write some scripts to split the reports from the demo-dashboard. But 
for the work on the template, it is temporary and work will commence on a 
seperated single file template to load the reports.
```


ongspxm:
generated data also need to checkstyle ah?

ongspxm:
template updated to "fake merge" the data files together.

it is pretty hacky, but i am planning to redo the whole thing when i change the frontend to a single template which loads the data dynamically.

ongspxm:
ready for review

damithc:
Are we going to have all these json files in the repo?

ongspxm:
this is too big a task, we are gonna split in into different pr. Work resumes on #111

ongspxm:
this is too big a task, we are gonna split in into different pr. Work resumes on #111

PR #91, opened 2018-05-14
Fix untested javascript files in functional test

ongspxm:

PR #106, opened 2018-05-16
Restore functional test

eugenepeh:
Continuing work of #91

Proposed message:

```
The assertion of javascript files has been omitted due to incorrect
keyword specified in file extension check.

This causes functional test passes irregardless of result which
doesn't provide a clear indication on the state of our project.

Let's fix the incorrect extension, and also all the OS and machine
ambiguities to restore our functional test.
```


src/functional/java/reposense/Entry.java
```diff
@@ -44,17 +44,18 @@ private void verifyAllJson(File expectedDirectory, String actualRelative) {
...
+                String relativeDirectory = file.getAbsolutePath()
+                        .replace("\\", "/").split("expected/")[1];
                 assertJson(file, relativeDirectory, actualRelative);
             }
         }
     }
 
     private void assertJson(File expectedJson, String expectedPosition, String actualRelative) {
-        File actual = new File(actualRelative + expectedPosition);
+        File actual = new File(actualRelative + "/" + expectedPosition);
```
damithc: Should we use an OS-independent method such as `File.separator` instead of `/`?

eugenepeh: I'm thinking of switching to [Paths](https://docs.oracle.com/javase/10/docs/api/java/nio/file/Paths.html) once I cleared this up. For those that are going to be printed to json files, we still need to use a fixed standard e.g. `/` though.

eugenepeh: On second thought, too many File occurrence to fix at the moment. Will change according to suggestion.

eugenepeh:
@damithc Prof, for your final review.

damithc:
Suggested commit message:
```
1. Javascript files are being omitted from functional tests 
due to incorrect keyword specified in file extension check. 
Let's correct the extension.

2. The output of the analysis vary (cosmetically) based on
the OS and the computer. This means it is impossible to
specify the 'expected' output for functional tests. 
Lets tweak the code to produce the exact output same  
output irrespective of OS and the computer.
```

Issue #69, opened 2018-05-07
Tweak code structure to match proposed architecture


damithc:
Proposed architecture is in this file
[diagrams.pptx](https://github.com/reposense/RepoSense/files/1978676/diagrams.pptx)

![image](https://user-images.githubusercontent.com/1673303/39687802-8d71c278-5202-11e8-9e51-90a42c3c0f0d.png)

PR #111, opened 2018-05-17
Split contribution summary by repository #69

eugenepeh:


eugenepeh:
@ongspxm I think some of the functions are still not using json format, can you rectify it?

ongspxm:
are you referring to the the template?

eugenepeh:
@ongspxm can you fix something tomorrow? 

Currently because the result for each repo is generated iteratively, the min and max date (of all the repo) is not known till the end. Hence, the start date and end date of all the repo to be the same unless I modifiy their configuration which is technically not right to do so. As a result, now if you run the program without setting the `-since` and `-until` parameter, the ramp chart will be jumbled (some user will start from 04 oct and some will start from 30 sep).

I need you to rectify the min and max date thing. I've tried fixing the getMin and getMax function in utils.js but that doesn't affect it, so its probably the Vue app or something.

Thanks.

On a side note, we should probably change weekly contribution to be calculated base on daily contribution as well.

eugenepeh:
More info: selecting the start date on the html date picker solves this problem, so perhaps can look into that direction?

ongspxm:
roger that, working on it nao

ongspxm:
i ran this pr on `summary_full`, gives the same starting date leh.. `9/30/17`, but the maxDate is not calculated properly though.. i will fix that

ongspxm:
eh.. i pr into architecture right?

yong24s:
@eugenepeh Due to security reasons, Chrome does not allow a file:// resource to do XHR.
image

But if I spawn a python server to serve the files. It performs as expected.
image

While Firefox does not have this issue at all.

eugenepeh:
>Looks good, but need to consider updating docs to mention the chrome issue, probably a guide to start a local server if using chrome.

Hmm, thats a legacy issue in this project though. Would be better to fix it in another PR.

eugenepeh:
Good catch, rectified.

eugenepeh:
Just realized that I've forgotten to remove the `RepoInfo` class and update the developer guide too.

@ongspxm @yamidark @AdityaA1998 @yong24s  for your review again.

src/main/java/reposense/report/ContributionSummaryGenerator.java
```diff
@@ -14,40 +14,30 @@
...
+     */
+    public static RepoContributionSummary analyzeContribution(
+            RepoConfiguration config,
+            List<CommitInfo> commitInfos,
+            HashMap<Author, Integer> authorContributionMap,
+            HashSet<Author> suspiciousAuthors) {
+        Date startDate = config.getFromDate() == null ? getStartDate(commitInfos) : config.getFromDate();
+        RepoContributionSummary summary = new RepoContributionSummary();
+        summary.setAuthorWeeklyIntervalContributions(
+                getAuthorIntervalContributions(config, commitInfos, startDate, 7, suspiciousAuthors));
```
ongspxm: previously mentioned to remove the weekly and do it in the frontend instead? in this PR or open new?

eugenepeh: In another PR instead

eugenepeh:
@damithc for your review. This PR a bit big due to lots of coupling.

PR #122, opened 2018-05-22
TestUtil: add file content comparison method

eugenepeh:
Proposed message:

```
Unlike diff tools, Assert.assertEquals does not provide useful 
information when the content in 2 File objects are not the same. 

This makes troubleshooting difficult. As without a clear indication, 
it is hard to trace the errors.

Let's implement the functionality of printing line differences by 
adding the method TestUtil#compareFileContents. 
```

### DIFF
--- a/src/functional/java/reposense/Entry.java
+++ b/src/functional/java/reposense/Entry.java
@@ -2,7 +2,6 @@ package reposense;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.List;
@@ -14,6 +13,7 @@ import reposense.dataobject.RepoConfiguration;
 import reposense.report.RepoInfoFileGenerator;
 import reposense.system.CsvConfigurationBuilder;
 import reposense.util.FileUtil;
+import reposense.util.TestUtil;
 
 
 public class Entry {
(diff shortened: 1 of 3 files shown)


yong24s:
LGTM. 

Adding on to @ongspxm inputs on why Travis may randomly fail, our investigations have shown that the "actual" file may be missing due to a race condition.

    private void assertJson(File expectedJson, String expectedPosition, String actualRelative) {
        File actual = new File(actualRelative + expectedPosition);
        >> Assert.assertTrue(actual.exists());
     }

src/functional/java/reposense/Entry.java
```diff
@@ -57,19 +57,11 @@ private void verifyAllJson(File expectedDirectory, String actualRelative) {
     private void assertJson(File expectedJson, String expectedPosition, String actualRelative) {
         File actual = new File(actualRelative + expectedPosition);
```
yamidark: `actualJson`?

src/test/java/reposense/util/TestUtil.java
```diff
@@ -0,0 +1,58 @@
...
+     * Also prints out error message if the lines count are different,
+     * else prints out maximum {@code maxTraceCounts} lines of content difference (if any).
+     */
+    public static boolean compareFileContents(File expected, File actual, int maxTraceCounts) throws IOException {
+        int traceCounts = 0;
+
+        System.out.println(String.format(MESSAGE_COMPARING_FILES, expected, actual));
+
+        String[] expectedContent = new String(Files.readAllBytes(expected.toPath()))
+                .replace("\r", "").split("\n");
```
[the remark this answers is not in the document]
eugenepeh: This is unlikely to happen elsewhere because it can only occur in file comparison, where Git likes to auto crlf files in windows.

eugenepeh:
My original intention was to do something like diff but I realized that such implementation would be overkill for this purpose.

Several reasons: 
1. Not practical for command line; Not possible for gradle to print out entire chunk of difference due to the large text sizes.
2. We don't need so much information. Since this method is implemented for testing, the input files were meant to be perfect match. Hence we only need to know the starting line of the difference and its content to rectify it. 

Hence, I implemented `compareFileContents` which prints only the first line of difference and the corresponding line number for troubleshooting purposes. Developing it on our own would also give us more control over its functionality and behavior.

src/test/java/reposense/util/TestUtil.java
```diff
@@ -0,0 +1,58 @@
...
+    private static final String MESSAGE_COMPARING_FILES = "Comparing files %s & %s\n";
+
+    private static final String MESSAGE_LINE_CONTENT_DIFFERENT = "Content different at line number %d:\n"
+            + "<< %s\n"
+            + ">> %s\n";
+
+    private static final String MESSAGE_LINES_LENGTH_DIFFERENT = "The files' lines count do not match.";
+
+    /**
+     * Returns true if the files' contents are the same, and false otherwise.
```
damithc: `and false otherwise` can be omitted.

PR #123, opened 2018-05-23
Implement a logger 

### LINKED ISSUE #98: Replace non-operation system output with logging
Output that users don't need to be concern with, shouldn't be displayed on CLI output stream.

Let's replace them with logging.

AdityaA1998:
Fixes #98 

Proposed message: 

```
Currently, all the unnecessary information - not concerning the user is being printed 
on console. 
Let's implement a logger to display only relevant information on the console and save the rest of 
it on an external log file. 
```


AdityaA1998:
@eugenepeh this PR is ready to review

eugenepeh:
Can you look into how to integrate it with gradle, and console arguments as well? 

Because ultimately we want to be able to configure what will be shown.

src/main/java/reposense/git/GitChecker.java
```diff
@@ -15,7 +20,7 @@ public static void checkoutBranch(String root, String branch) {
     }
 
     public static void checkOutToCommit(String root, CommitInfo commit) {
-        System.out.println("Checking out " + commit.getHash() + "time:" + commit.getTime());
+        logger.info("Checking out " + commit.getHash() + "time:" + commit.getTime());
```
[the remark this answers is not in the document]
AdityaA1998: Using String format is actually slower than using String concatenate/builder.
Results: https://stackoverflow.com/questions/925423/is-it-better-practice-to-use-string-format-over-string-concatenation-in-java

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,99 @@
...
+        addFileHandler(logger);
+
+        return logger;
+    }
+
+    /**
+     * Creates a Logger for the given class name.
+     */
+    public static <T> Logger getLogger(Class<T> clazz) {
+        if (clazz == null) {
```
ongspxm: lolz, clazz

src/main/java/reposense/git/GitCloner.java
```diff
@@ -24,7 +30,7 @@ public static void downloadRepo(String organization, String repoName, String bra
         try {
             GitChecker.checkout(FileUtil.getRepoDirectory(organization, repoName), branchName);
         } catch (RuntimeException e) {
-            System.out.println("Error: Branch does not exist! Analyze terminated.");
+            logger.info("Error: Branch does not exist! Analyze terminated.");
```
eugenepeh: shouldn't this be logger.error? Also, the `Error:` prefix can be removed now.

src/main/java/reposense/report/RepoInfoFileGenerator.java
```diff
@@ -33,7 +37,7 @@ public static void generateReposReport(List<RepoConfiguration> repoConfigs, Stri
             try {
                 GitCloner.downloadRepo(config.getOrganization(), config.getRepoName(), config.getBranch());
             } catch (GitClonerException e) {
-                System.out.println("Exception met when cloning the repo, will skip this one");
+                logger.info("Exception met when cloning the repo, will skip this one");
```
eugenepeh: logger.warning

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Configures and manages the loggers and handlers, including their levels
+ */
+public class LogsManager {
+    private static final int FILE_COUNT = 2;
+    private static final int MEGABYTE = (int) (Math.pow(2.0, 20));
+    private static final int MAX_FILE_SIZE_IN_BYTES = 5 * MEGABYTE; // 5MB
+    private static final String LOG_FILE = "reposense.log";
+    private static Level currentConsoleLogLevel = Level.INFO;
+    private static Level currentFileLogLevel = Level.INFO;
```
eugenepeh: Perhaps create setters for the 2 log level to make it configurable?

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Configures and manages the loggers and handlers, including their levels
+ */
+public class LogsManager {
+    private static final int FILE_COUNT = 2;
+    private static final int MEGABYTE = (int) (Math.pow(2.0, 20));
+    private static final int MAX_FILE_SIZE_IN_BYTES = 5 * MEGABYTE; // 5MB
+    private static final String LOG_FILE = "reposense.log";
+    private static Level currentConsoleLogLevel = Level.INFO;
```
yamidark: should have an empty line to separate the `static final` from `static` variables (not idea how this passed checkstyle)

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,99 @@
...
+
+        removeHandlers(logger);
+        addConsoleHandler(logger);
+        addFileHandler(logger);
+
+        return logger;
+    }
+
+    /**
+     * Creates a Logger for the given class name.
```
yamidark: `{@code Logger}`

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,99 @@
...
+    private static void removeHandlers(Logger logger) {
+        Handler[] handlers = logger.getHandlers();
+        for (Handler handler : handlers) {
+            logger.removeHandler(handler);
+        }
+    }
+
+    /**
+     * Adds the {@code fileHandler} to the {@code logger}.
+     * Creates {@code fileHandler} if it is null.
```
yamidark: Same as above, `fileHandler` is `private` and not visible outside of this class

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.logging.Logger;
+import java.util.logging.SimpleFormatter;
+
+/**
+ * Configures and manages the loggers and handlers, including their levels
+ */
+public class LogsManager {
+    private static final int FILE_COUNT = 2;
+    private static final int MEGABYTE = (int) (Math.pow(2.0, 20));
+    private static final int MAX_FILE_SIZE_IN_BYTES = 5 * MEGABYTE; // 5MB
```
yamidark: Unnecessary comment

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import java.util.logging.SimpleFormatter;
+
+/**
+ * Configures and manages the loggers and handlers, including their levels
+ */
+public class LogsManager {
+    private static final int FILE_COUNT = 2;
+    private static final int MEGABYTE = (int) (Math.pow(2.0, 20));
```
yamidark: Don't think we  should be using `Math.pow` which returns a `double` and can have floating point errors, also when we cast this back to `int`. Can try using the absolute value, or use `1 << 20`

src/main/java/reposense/system/LogsManager.java
AdityaA1998: For _all_ the info to be stored in the log file, we need to set it to the lowest level; i.e. info in our project.

AdityaA1998:
@eugenepeh @yamidark ready to review

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final int MAX_FILE_SIZE_IN_BYTES = 5 * MEGABYTE; // 5MB
+    private static final String LOG_FILE = "reposense.log";
+
+    private static Level currentConsoleLogLevel = Level.INFO;
+    private static Level currentFileLogLevel = Level.INFO;
+    private static FileHandler fileHandler;
+    private static ConsoleHandler consoleHandler;
+
+    public static Logger getLogger(String name) {
+        Logger logger = Logger.getLogger(name);
```
yamidark: I think we should keep a list of all loggers created in `LogsManager`, so that later when we need to change the `currentConsoleLogLevel` based on user input, we can update each classes' logger.

src/main/java/reposense/system/CustomLogFormatter.java
```diff
@@ -0,0 +1,31 @@
+package reposense.system;
+
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.logging.Handler;
+import java.util.logging.LogRecord;
+import java.util.logging.SimpleFormatter;
+
+public class CustomLogFormatter extends SimpleFormatter {
```
yamidark: Javadoc comments for class

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    private static FileHandler createFileHandler() throws IOException {
+        FileHandler fileHandler = new FileHandler(LOG_FILE, MAX_FILE_SIZE_IN_BYTES, FILE_COUNT, true);
+        fileHandler.setFormatter(new SimpleFormatter());
+        fileHandler.setLevel(currentFileLogLevel);
+        return fileHandler;
+    }
+
+    private static ConsoleHandler createConsoleHandler() {
+        CustomLogFormatter consoleHandlerFormat = new CustomLogFormatter();
```
yamidark: No need to create, just directly pass into `setFormatter`

src/main/java/reposense/system/CustomLogFormatter.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private static final DateFormat dateFormat = new SimpleDateFormat("dd/MM/yyyy hh:mm:ss");
+
+    @Override
+    public synchronized String format(LogRecord record) {
+        StringBuilder builder = new StringBuilder(1000);
```
eugenepeh: is it necessary to initialise the initial capacity?

src/main/java/reposense/analyzer/CheckStyleParser.java
```diff
@@ -27,9 +29,11 @@
     private static final String SEVERITY = "severity";
     private static final String LINE = "line";
 
+    private static final Logger logger = LogsManager.getLogger(CheckStyleParser.class);
+
     public static void aggregateStyleIssue(FileInfo fileInfo, String rootRepo) {
         String raw = CommandRunner.checkStyleRaw(rootRepo + '/' + fileInfo.getPath());
-        System.out.println("checking style of:" + fileInfo.getPath());
+        logger.info("checking style of:" + fileInfo.getPath());
```
eugenepeh: `Checking`

src/main/java/reposense/report/ContributionSummaryGenerator.java
```diff
@@ -15,7 +15,6 @@
 import reposense.dataobject.RepoConfiguration;
 import reposense.dataobject.RepoContributionSummary;
 
-
```
eugenepeh: this change is not necessary

src/main/java/reposense/system/CustomLogFormatter.java
```diff
@@ -0,0 +1,31 @@
...
+        builder.append(formatMessage(record));
+        builder.append("\n");
+        return builder.toString();
+    }
+
+    public String getHead(Handler h) {
+        return super.getHead(h);
+    }
+
+    public String getTail(Handler h) {
```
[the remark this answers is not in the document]
AdityaA1998: These methods are necessary for the formatter

src/main/java/reposense/system/LogsManager.java
AdityaA1998: I think so, instead of maintaining a list, whenever a user initialises the logger, he can pass the level in the getLogger function!

src/main/java/reposense/system/CustomLogFormatter.java
AdityaA1998: Noop

src/main/java/reposense/system/CustomLogFormatter.java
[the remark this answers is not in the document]
yamidark: Just tested, these methods aren't necessary, can remove them

AdityaA1998:
@damithc Prof for your review

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,109 @@
...
+ * Configures and manages the loggers and handlers, including their levels
+ */
+public class LogsManager {
+    private static final int FILE_COUNT = 2;
+    private static final int MEGABYTE = (1 << 20);
+    private static final int MAX_FILE_SIZE_IN_BYTES = 5 * MEGABYTE; // 5MB
+    private static final String LOG_FILE = "reposense.log";
+
+    private static Level currentConsoleLogLevel = Level.INFO;
+    private static Level currentFileLogLevel = Level.INFO;
```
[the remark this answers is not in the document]
AdityaA1998: You need to set the level to the lowest in order to display all the information. `Level.SEVERE` is the highest and will only display all the severities

.gitignore
```diff
@@ -9,3 +9,4 @@ build/
 *.zip
 
 repos/
+reposense.log.0
```
damithc: We allow 2 log files right? This line seems to ignore just one file.
May be create all log files in one `/logs` folder and ignore the entire folder?

AdityaA1998:
@damithc resolved the log folder issue. Ready for review

damithc:
Looks like there are unrelated files in the PR?
It's best to wait till CI passes before requesting a review.

AdityaA1998:
@damithc prof for your review

src/main/java/reposense/system/CustomLogFormatter.java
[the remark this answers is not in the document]
damithc: What's the final decision on this? Remove or keep? @yamidark @AdityaA1998

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Configures and manages the loggers and handlers, including their levels
+ */
+public class LogsManager {
+    private static final int FILE_COUNT = 2;
+    private static final int MEGABYTE = (1 << 20);
+    private static final int MAX_FILE_SIZE_IN_BYTES = 5 * MEGABYTE; // 5MB
+    private static final String LOG_FILE_LOCATION = "./logs/";
+    private static final String LOG_FILE =  LOG_FILE_LOCATION + "reposense.log";
```
damithc: Add a comment about this block of constants to explain the overall outcome i.e., where files are saved, how and when log files are rolled over, etc.

src/main/java/reposense/system/CustomLogFormatter.java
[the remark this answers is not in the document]
AdityaA1998: We can remove these methods

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Adds the File Handler to the {@code logger}.
+     * Creates File Handler if it is null.
+     */
+    private static void addFileHandler(Logger logger) {
+        File file = new File(LOG_FILE_LOCATION);
+        if (!file.exists() && file.mkdir()) {
+            logger.info("Log folder has been successfully created");
```
yamidark: Can we use `nio.Path` instead of `io.File` like this:
```Java
Path filePath = Paths.get(LOG_FILE_LOCATION);
            if (!Files.exists(filePath)) {
                Files.createDirectory(filePath);
                logger.info("Log folder has been successfully created");
            }
```

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,109 @@
...
+     */
+    private static void addFileHandler(Logger logger) {
+        try {
+            if (fileHandler == null) {
+                fileHandler = createFileHandler();
+            }
+            logger.addHandler(fileHandler);
+        } catch (IOException e) {
+            logger.warning("Error adding file handler for logger.");
+        }
```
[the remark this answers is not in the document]
yamidark: ping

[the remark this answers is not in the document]
yamidark: Whoops, forgot that the `logger` used here already belongs to that `class`, so no need for this :sweat_smile:

AdityaA1998:
@damithc prof for your review

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,128 @@
...
+     */
+    private static void addConsoleHandler(Logger logger) {
+        if (consoleHandler == null) {
+            consoleHandler = createConsoleHandler();
+        }
+        logger.addHandler(consoleHandler);
+    }
+
+    /**
+     * Remove all the handlers from {@code logger}.
```
damithc: Removes

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,128 @@
...
+            }
+        }
+        try {
+            if (fileHandler == null) {
+                fileHandler = createFileHandler();
+            }
+            logger.addHandler(fileHandler);
+        } catch (IOException e) {
+            logger.warning("Error adding file handler for logger.");
+        }
```
damithc: If we fail to create the directory, is there any point in adding the file handler?

AdityaA1998: No point in creating a handler then. Noted will `return` as soon as the exception is caught.

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,128 @@
...
+     */
+    private static void addFileHandler(Logger logger) {
+        Path path = Paths.get(LOG_FILE_LOCATION);
+
+        if (!Files.exists(path)) {
+            try {
+                Files.createDirectory(path);
+                logger.info("Log folder has been successfully created");
+            } catch (IOException io) {
+                logger.warning("There was problem creating a log folder");
```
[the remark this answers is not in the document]
AdityaA1998: Noted.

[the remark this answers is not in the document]
AdityaA1998: The logger has two outputs, one on the console and one in the file. If there is a problem with creating a directory, the warning message can still be displayed on the console.

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Adds the File Handler to the {@code logger}.
+     * Creates File Handler if it is null.
+     */
+    private static void addFileHandler(Logger logger) {
+        Path path = Paths.get(LOG_FILE_LOCATION);
+
+        if (!Files.exists(path)) {
+            try {
+                Files.createDirectory(path);
```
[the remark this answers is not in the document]
AdityaA1998: `createDirectories` results in an issue that it only throws an error when there is a security problem, otherwise it doesn't throw any error (eg. if the the directory exists) which results in extra unnecessary log messages.

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * Adds the File Handler to the {@code logger}.
+     * Creates File Handler if it is null.
+     */
+    private static void addFileHandler(Logger logger) {
+        Path path = Paths.get(LOG_FILE_LOCATION);
+
+        if (!Files.exists(path)) {
```
yong24s: Maybe the block of code can be extracted to a method? Since it does not really do```addFileHandler```but more of a satisfying conditions/dependencies.
```
if (!Files.exists(path)) {
        try {
            Files.createDirectories(path);
            logger.info("Log folder has been successfully created");
        } catch (IOException io) {
            logger.warning("There was problem creating a log folder");
            return;
        }
    }
```

src/main/java/reposense/system/LogsManager.java
AdityaA1998: Noted

AdityaA1998:
@eugenepeh ready for review

AdityaA1998:
@damithc prof ready for review

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,137 @@
...
+                logger.info("Log folder has been successfully created");
+
+                if (fileHandler == null) {
+                    fileHandler = createFileHandler();
+                }
+                logger.addHandler(fileHandler);
+            } catch (IOException e) {
+                logger.severe(LogsManager.getErrorDetails(e));
+            }
+        }
```
damithc: The whole code block is executed only if the file does not exist?

If that is the case, a guard clause will reduce the nesting depth of the method.

```java
if (Files.exists(path)){
    return;
}
...
```

AdityaA1998: There was an error. `fileHandler` will be added even if the directory exists. The scope of `if` statement was wrong. It has been rectified.

damithc: Need to be more careful in future.

src/main/java/reposense/system/LogsManager.java
```diff
@@ -0,0 +1,137 @@
...
+    public static void setConsoleHandlerLevel(Level level) {
+        currentConsoleLogLevel = level;
+    }
+
+    public static void setFileConsoleHandlerLevel(Level level) {
+        currentFileLogLevel = level;
+    }
+
+    /**
+     * Returns a detailed message of the t, including the stack trace.
```
damithc: `t` should be in `{@code ...}` as it refers to the parameter name?

src/main/java/reposense/analyzer/FileInfoGenerator.java
```diff
@@ -22,7 +26,7 @@ public static FileInfo generateFileInfo(String repoRoot, String relativePath) {
                 lineNum += 1;
             }
         } catch (IOException ioe) {
-            ioe.printStackTrace();
+            logger.severe(LogsManager.getErrorDetails(ioe));
```
eugenepeh: Instead of doing this, have you tried this overload method?
https://docs.oracle.com/javase/7/docs/api/java/util/logging/Logger.html#log(java.util.logging.Level,%20java.lang.String,%20java.lang.Throwable)

eugenepeh:
image

Still has 2 lines not ported over to logger.

PR #124, opened 2018-05-23
changing frontend to a single file format

ongspxm:
changes to shift to a single template system where content is dynamically loaded. wip.
here just for the travis test.

mostly a transfer of features, with a more vue-powered structure. major architectural changes, but most doesn't affect the output, here are some that do.

## things to consider
- [ ] scale of ramp slices & contribution bar

## changes to data results
- weekly commits calculated from daily commits

### result not displayed immediately, loaded from hash or user input
image
this single file template is capable of being used with different set of results data.

### changing display style to make it obvious when grouped by repos
image
currently when grouped by repo, it is hard to tell, but this manner of grouping them with a bolded title allows users to identify different repos.


ongspxm:
closed. put to #129

PR #125, opened 2018-05-23
Fix datetime inconsistency

eugenepeh:
```
Several date time inconsistency exists in the project.
1. Instead of designated constants, raw values are being used in month
   field of Calender setter.
2. Time zone and hours were not specified in the git log command
   execution.

This leads to disagreeing outputs when running the program at
different timing or on different machine.

Let's fix the date time inconsistency by
1. using the designated constants,
2. and fixed formatted values.
```

### DIFF
--- a/src/main/java/reposense/util/Constants.java
+++ b/src/main/java/reposense/util/Constants.java
@@ -16,7 +16,8 @@ public class Constants {
     public static final String LOG_SPLITTER = "\\|";
     public static final DateFormat GIT_ISO_FORMAT = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
     public static final DateFormat REPORT_NAME_FORMAT = new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss");
-    public static final DateFormat GIT_LOG_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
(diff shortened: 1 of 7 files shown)


yamidark:
Should we update the User Guide and documentation that the `since` and `until` dates input are based on `+08:00` timezone?

src/main/java/reposense/util/Constants.java
```diff
@@ -16,7 +16,8 @@
     public static final String LOG_SPLITTER = "\\|";
     public static final DateFormat GIT_ISO_FORMAT = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
     public static final DateFormat REPORT_NAME_FORMAT = new SimpleDateFormat("yyyy-MM-dd-HH-mm-ss");
-    public static final DateFormat GIT_LOG_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
+    public static final DateFormat GIT_LOG_SINCE_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'00:00:00+08:00");
```
yong24s: Not sure if we should standardize the timezone to GMT +8 for all date operations.
It is reasonable to use it for system testing but more research has to been done for users of other timezone.

I'll try to test it to see if it cause any problems.

eugenepeh:
I guess I will just add a notice in user guide for now. Would be nice if we make add a cli parameter in the future.
e.g. `-timezone local`, `-timezone +08:00`??

Or perhaps we can go with a config file too

@yamidark @yong24s for your review again.

yong24s:
LGTM :1st_place_medal:

Issue #84, opened 2018-05-11
Functional test is too computational intensive


eugenepeh:
Functional test is too computational intensive.

It takes up over 2 GB and full CPU usage on Windows laptop with i7 duo core processor. It also seems that it has memory leak issues.

Let's look for means to reduce the computational requirement of functional test.

PR #126, opened 2018-05-24
Improve performance

### LINKED ISSUE #94: Replace string concatenation with StringBuilder or similar function
While compiler _may_ optimize the code to use the StringBuilder alternative, there's the possibility that it might fail to do so.

More discussion can be found here:
https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

yamidark:
Fixes #94.
Part of #84.
```
The analyzing of contribution for each file in the program is very slow 
due to several factors:
1. The `git blame` process produces a lot of unnecessary line information, 
causing a large amount of unnecessary time spent reading I/O. 
2. StreamGobbler uses String concatenation when reading from process InputStream,
which is polynomial in time based on the number of lines.

Let's improve performance of the program and functional test by:
1. Filter the `git blame` process output using `findstr` in Windows OS or `grep`
for other OS, thus having a smaller I/O to read.
2. Change StreamGobbler to read directly from InputStream instead of using
a BufferedReader, and use StringBuilder instead of String concatenation.
```

### DIFF
--- a/src/main/java/reposense/git/GitBlamer.java
+++ b/src/main/java/reposense/git/GitBlamer.java
@@ -1,8 +1,5 @@
 package reposense.git;
 
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import reposense.dataobject.Author;
 import reposense.dataobject.FileInfo;
 import reposense.dataobject.RepoConfiguration;
@@ -11,24 +8,19 @@ import reposense.system.CommandRunner;
 
 public class GitBlamer {
 
-    private static final Pattern INSERTION_PATTERN = Pattern.compile("^author (.+)");
+    private static final int AUTHOR_NAME_OFFSET = "author ".length();
 
     public static void aggregateBlameInfo(FileInfo fileInfo, RepoConfiguration config) {
-        //System.out.println("blaming " + fileInfo.getPath());
         String raw = CommandRunner.blameRaw(config.getRepoRoot(), fileInfo.getPath());
         String[] rawLines = raw.split("\n");
         int lineCount = 0;
-        for (int i = 0; i < rawLines.length; i++) {
-            Matcher m = INSERTION_PATTERN.matcher(rawLines[i]);
-            if (m.find()) {
-                String authorRawName = m.group(1);
-                Author author = config.getAuthorAliasMap().get(authorRawName);
-                if (author == null) {
-                    author = new Author("-");
-                }
-                fileInfo.getLines().get(lineCount++).setAuthor(author);
+        for (String line: rawLines) {
+            String authorRawName = line.substring(AUTHOR_NAME_OFFSET);
+            Author author = config.getAuthorAliasMap().get(authorRawName);
+            if (author == null) {
+                author = new Author("-");
             }
+            fileInfo.getLines().get(lineCount++).setAuthor(author);
         }
-
     }
 }
--- a/src/main/java/reposense/system/CommandRunner.java
+++ b/src/main/java/reposense/system/CommandRunner.java
@@ -8,7 +8,6 @@ import java.util.Date;
 
 import reposense.util.Constants;
 
-
 public class CommandRunner {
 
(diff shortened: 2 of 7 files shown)


yamidark:
@eugenepeh @yong24s For your review.

src/main/java/reposense/system/StreamGobbler.java
```diff
@@ -1,15 +1,14 @@
...
 
-import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.InputStreamReader;
-
 
 public class StreamGobbler extends Thread {
 
+    private static final byte[] BUFFER = new byte[1 << 13];
```
eugenepeh: rationale for using 2^13?

src/main/java/reposense/git/GitBlamer.java
```diff
@@ -11,24 +8,19 @@
 
 public class GitBlamer {
 
-    private static final Pattern INSERTION_PATTERN = Pattern.compile("^author (.+)");
+    private static final int AUTHOR_NAME_OFFSET = 7;
```
eugenepeh: Do you think if we put "author ".length() would be better?

Because I think its more explanatory that way

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -31,7 +30,13 @@ public static void checkOut(String root, String hash) {
 
     public static String blameRaw(String root, String fileDirectory) {
         File rootFile = new File(root);
-        return runCommand(rootFile, "git blame " + addQuote(fileDirectory) + " -w -C -C -M --line-porcelain");
+        String blameCommand = "git blame " + addQuote(fileDirectory) + " -w -C -C -M --line-porcelain | ";
+        if (isWindows) {
+            blameCommand += "findstr /B /C:" + addQuote("author ");
+        } else {
+            blameCommand += "grep " + addQuote("^author .*");
```
eugenepeh: grep isn't installed by default for some linux system though. So if we use grep, we need to include it in the user guide & developer guide as well.

src/main/java/reposense/system/StreamGobbler.java
yamidark: Based on my own testing, around 2^11 - 2^13 gives the best speed.

eugenepeh: Hmm, can you explore using [ByteBuffer](https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html)?

yamidark:
@eugenepeh @yong24s For your review again.

src/main/java/reposense/system/StreamGobbler.java
```diff
@@ -1,15 +1,17 @@
...
 import java.io.InputStream;
-import java.io.InputStreamReader;
-
+import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
+import java.nio.channels.ReadableByteChannel;
 
 public class StreamGobbler extends Thread {
 
+    private final ByteBuffer buffer = ByteBuffer.allocate(1 << 13);
```
yong24s: Good to mention that this is 8KB.

src/main/java/reposense/system/CommandRunner.java
ongspxm: lolz.... `addQuote`, what about using slash escape?

yamidark: `addQuote` is using slash escape 😅

yamidark:
Based on my own testing and Travis, the runtime for functional tests improves by about 1min, from around 6-7min to 5-6mins on average.

eugenepeh:
@AdityaA1998 can you test if it works properly on mac?

eugenepeh:
For future error, please post the error message on the thread so that we can take note of it.

AdityaA1998:
Yep! It works

PR #128, opened 2018-05-25
Increase functionality of arguments parser

yong24s:


src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Verifies that {@code sinceDate} is earlier than {@code untilDate}
+     *
+     * @throws IllegalArgumentException If user-supplied values are invalid
+     */
+    private static void verifyDatesAreInCorrectRange(Optional<Date> sinceDate, Optional<Date> untilDate) {
+        if (sinceDate.isPresent() && untilDate.isPresent() && sinceDate.get().getTime() > untilDate.get().getTime()) {
+            throw new IllegalArgumentException(MESSAGE_SINCE_DATE_LATER_THAN_UNTIL_DATE);
+        }
```
ongspxm: should split out the error, just do a `sinceDate.get()`, will throw its own error.
`MESSAGE_SINICE_DATE_LATER_THAN_UNTIL_DATE` not really describe when sinceDate and untilDate is empty.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,91 @@
...
+public class ArgsParser {
+    private static final String PROGRAM_USAGE = "java -jar RepoSense.jar";
+    private static final String PROGRAM_DESCRIPTION =
+            "RepoSense is a contribution analysis tool for Git repositories.\n";
+    private static final String MESSAGE_SINCE_DATE_LATER_THAN_UNTIL_DATE =
+            "\"Since Date\" cannot be later than \"Until Date\"";
+
+    private static ArgumentParser getArgumentParser() {
+        ArgumentParser parser = ArgumentParsers
+                .newFor(PROGRAM_USAGE)
```
yamidark: I think this should be the name of our program `RepoSense`?

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,91 @@
...
+
+        return cliArguments;
+    }
+
+    /**
+     * Verifies that {@code sinceDate} is earlier than {@code untilDate}
+     *
+     * @throws IllegalArgumentException If user-supplied values are invalid
+     */
+    private static void verifyDatesAreInCorrectRange(Optional<Date> sinceDate, Optional<Date> untilDate) {
```
yamidark: i think we should do a check that the `Dates` are present inside the `Optional` 1st, before the `Date` here directly for the check.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,91 @@
...
+                .newFor(PROGRAM_USAGE)
+                .build()
+                .defaultHelp(true)
+                .description(PROGRAM_DESCRIPTION);
+
+        parser.addArgument("-config")
+                .required(true)
+                .type(Arguments.fileType().verifyExists().verifyIsFile().verifyCanRead())
+                .metavar("PATH")
+                .help("The CSV File  to the CSV config file");
```
yamidark: `The path to the CSV config file to read`

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,91 @@
...
+import net.sourceforge.argparse4j.inf.ArgumentParser;
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+import net.sourceforge.argparse4j.inf.MutuallyExclusiveGroup;
+import net.sourceforge.argparse4j.inf.Namespace;
+import reposense.frontend.CliArguments;
+
+public class ArgsParser {
+    private static final String PROGRAM_USAGE = "java -jar RepoSense.jar";
+    private static final String PROGRAM_DESCRIPTION =
+            "RepoSense is a contribution analysis tool for Git repositories.\n";
```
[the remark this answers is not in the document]
yong24s: Yes. Line is longer than 120 chars.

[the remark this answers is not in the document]
yamidark: As in the `'\n' at the end

src/main/java/reposense/parser/DateArgumentType.java
```diff
@@ -0,0 +1,27 @@
...
+ * Verifies and parses a string-formatted date to a Date object.
+ */
+public class DateArgumentType implements ArgumentType<Optional<Date>> {
+    private static final String PARSE_EXCEPTION_MESSAGE_INVALID_DATE_STRING_FORMAT = "Invalid Date: %s";
+
+    @Override
+    public Optional<Date> convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
+        try {
+            return Optional.of(Constants.CLI_ARGS_DATE_FORMAT.parse(value));
+        } catch (java.text.ParseException pex) {
```
eugenepeh: why not just import this exception? Also `pe`

yong24s:
image

src/main/java/reposense/frontend/RepoSense.java
```diff
@@ -27,19 +24,6 @@ public static void main(String[] args) {
...
-    private static void showHelpMessage() {
-        System.out.println("usage: java -jar RepoSense.jar -config CSV_CONFIG_FILE_PATH\n"
-                + "   [-output OUTPUT_DIRECTORY]\n"
-                + "   [-since DD/MM/YYYY]\n"
-                + "   [-until DD/MM/YYYY]\n");
-        System.out.println("-config: Mandatory. The path to the CSV config file.");
-        System.out.println("-output: Optional. The path to the dashboard generated.\n"
-                + "   If not provided, it will be generated in the current directory.");
-        System.out.println("-since : Optional. start date of analysis. Format: dd/MM/yyyy");
-        System.out.println("-until : Optional. end date of analysis. Format: dd/MM/yyyy");
```
eugenepeh: How about using the old messages instead?

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,92 @@
...
+
+            Path configFilePath = Paths.get(results.get("config"));
+            Path outputFilePath = Paths.get(results.get("output"));
+            Optional<Date> sinceDate = results.get("since");
+            Optional<Date> untilDate = results.get("until");
+
+            verifyDatesRangeIsCorrect(sinceDate, untilDate);
+
+            cliArguments = new CliArguments(configFilePath, outputFilePath, sinceDate, untilDate);
+        } catch (ArgumentParserException apex) {
```
yamidark: `ape`
no need to add the `x` for each Exception

src/main/java/reposense/parser/DateArgumentType.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Optional;
+
+import net.sourceforge.argparse4j.inf.Argument;
+import net.sourceforge.argparse4j.inf.ArgumentParser;
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+import net.sourceforge.argparse4j.inf.ArgumentType;
+import reposense.util.Constants;
+
+/**
+ * Verifies and parses a string-formatted date to a Date object.
```
yamidark: `{@code Date}`

src/main/java/reposense/frontend/RepoSense.java
yong24s: Yes. Done.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+    private Date getDate(int year, int month, int date) {
+        return new Calendar.Builder()
+                            .setDate(year, month, date)
+                            .setTimeOfDay(0, 0, 0)
+                            .build()
+                            .getTime();
+    }
+
+    @Test
+    public void allCorrectInputs_success() throws ArgumentParserException {
```
yamidark: `parse_allCorrectInputs_success()`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+        Assert.assertEquals(outputDirectoryPath.toAbsolutePath(), cliArguments.getTargetFilePath().toAbsolutePath());
+
+        Date expectedSinceDate = getDate(2017, Calendar.JULY, 1);
+        Date expectedUntilDate = getDate(2017, Calendar.NOVEMBER, 30);
+        Assert.assertEquals(expectedSinceDate, cliArguments.getSinceDate().get());
+        Assert.assertEquals(expectedUntilDate, cliArguments.getUntilDate().get());
+    }
+
+    @Test
+    public void absolutePath_toConfigFile_success() throws ArgumentParserException {
```
yamidark: `parse_configFileOnly_success()`
Think we should put the `absolutePath` and `relativePath` test here together

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+    @Test
+    public void relativePath_toConfigFile_success() throws ArgumentParserException {
+        String input = String.format("-config %s", "sample_small.csv");
+        String[] args = input.split(" ");
+        ArgsParser.parse(args);
+    }
+
+    @Test
+    public void absolutePath_toOutputDirectory_success() throws ArgumentParserException {
```
yamidark: `parse_configFileAndOutputDirectory_success()`
same as above, we can mix the `absolutePath` and `relativePath` test here together

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+
+import reposense.frontend.CliArguments;
+
+public class ArgsParserTest {
+
+    private String projectDirectory;
+
+    public ArgsParserTest() {
+        projectDirectory = System.getProperty("user.dir");
```
yamidark: set this as `private static final` directly above, no need for this constructor

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+        return new Calendar.Builder()
+                            .setDate(year, month, date)
+                            .setTimeOfDay(0, 0, 0)
+                            .build()
+                            .getTime();
+    }
+
+    @Test
+    public void allCorrectInputs_success() throws ArgumentParserException {
+        Path configFilePath = Paths.get(projectDirectory + "/sample_small.csv");
```
yamidark: for `Paths`, I think we can use `Paths.get(projectDirectory, "sample_small.csv")`, instead of manually adding the file separator ourselves

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+        Assert.assertEquals(expectedSinceDate, cliArguments.getSinceDate().get());
+        Assert.assertEquals(expectedUntilDate, cliArguments.getUntilDate().get());
+    }
+
+    @Test
+    public void absolutePath_toConfigFile_success() throws ArgumentParserException {
+        String directory = Paths.get(projectDirectory + "/sample_small.csv").toAbsolutePath().toString();
+        String input = String.format("-config %s", directory);
+        String[] args = input.split(" ");
+        ArgsParser.parse(args);
```
yamidark: Add the `Assert.assertEquals` tests here

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+        Date expectedSinceDate = getDate(2017, Calendar.JULY, 1);
+        Date expectedUntilDate = getDate(2017, Calendar.NOVEMBER, 30);
+        Assert.assertEquals(expectedSinceDate, cliArguments.getSinceDate().get());
+        Assert.assertEquals(expectedUntilDate, cliArguments.getUntilDate().get());
+    }
+
+    @Test
+    public void absolutePath_toConfigFile_success() throws ArgumentParserException {
+        String directory = Paths.get(projectDirectory + "/sample_small.csv").toAbsolutePath().toString();
```
yamidark: This is the config file, which is not a directory right?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+    @Test(expected = ArgumentParserException.class)
+    public void directory_forConfigFile_notAllowed() throws ArgumentParserException {
+        String directory = Paths.get(projectDirectory).getParent().toAbsolutePath().toString();
+        String input = String.format("-config %s", directory);
+        String[] args = input.split(" ");
+        ArgsParser.parse(args);
+    }
+
+    @Test(expected = ArgumentParserException.class)
+    public void nonExistentFile_forConfigFile_notAllowed() throws ArgumentParserException {
```
yamidark: `parse_conFigFileNotExists_throwsArgumentParserException`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,177 @@
...
+    public void untilDate_correctFormat_success() throws ArgumentParserException {
+        String configFile = Paths.get(projectDirectory + "/sample_small.csv").toString();
+        String untilDate = "30/11/2017";
+        String input = String.format("-config %s -until %s", configFile, untilDate);
+        String[] args = input.split(" ");
+        ArgsParser.parse(args);
+    }
+
+    @Test
+    public void optionalArguments_haveDefaultValues() throws ArgumentParserException {
```
yamidark: assertions can be done in the `configFileOnly` test above

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,179 @@
...
+                            .setDate(year, month, date)
+                            .setTimeOfDay(0, 0, 0)
+                            .build()
+                            .getTime();
+    }
+
+    @Test
+    public void parse_allCorrectInputs_success() throws ArgumentParserException {
+        Path configFilePath = Paths.get(projectDirectory, "sample_small.csv");
+        Path outputDirectoryPath = Paths.get(projectDirectory);
```
eugenepeh: most of these can be place in `@setup`?

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,101 @@
...
+            "\"Since Date\" cannot be later than \"Until Date\"";
+
+    private static ArgumentParser getArgumentParser() {
+        ArgumentParser parser = ArgumentParsers
+                .newFor(PROGRAM_USAGE)
+                .addHelp(false)
+                .build()
+                .description(PROGRAM_DESCRIPTION);
+
+        parser.formatHelp();
```
eugenepeh: this line seems redundant since its a getter.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,98 @@
...
+        parser.addArgument("-until")
+                .metavar("dd/MM/yyyy")
+                .type(new DateArgumentType())
+                .setDefault(Optional.empty())
+                .help("The date to stop filtering.");
+
+        return parser;
+    }
+
+    public static CliArguments parse(String[] args) throws ParseException {
```
eugenepeh: missing header comments

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,98 @@
...
+
+            verifyDatesRangeIsCorrect(sinceDate, untilDate);
+            return new CliArguments(configFilePath, outputFilePath, sinceDate, untilDate);
+        } catch (ArgumentParserException ape) {
+            throw new ParseException(String.format(getArgumentParser().formatUsage() + ape.getMessage()));
+        }
+    }
+
+    /**
+     * Verifies that {@code sinceDate} is earlier than {@code untilDate}
```
eugenepeh: missing `.` at the end.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,178 @@
...
+import reposense.exception.ParseException;
+import reposense.frontend.CliArguments;
+
+public class ArgsParserTest {
+
+    private static final String projectDirectory = System.getProperty("user.dir");
+
+    private Date getDate(int year, int month, int date) {
+        return new Calendar.Builder()
+                            .setDate(year, month, date)
```
eugenepeh: indentation of 8 spaces from prev line for linewrap

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,178 @@
...
+        String input = String.format("-config %s", configFile);
+        String[] args = input.split(" ");
+        CliArguments cliArgumentsOfAbsolutePath =  ArgsParser.parse(args);
+
+        input = String.format("-config %s", "sample_small.csv");
+        args = input.split(" ");
+        CliArguments cliArgumentsOfRelativePath = ArgsParser.parse(args);
+
+        Assert.assertEquals(cliArgumentsOfAbsolutePath.getConfigFilePath(),
+                cliArgumentsOfRelativePath.getConfigFilePath());
```
eugenepeh: You shouldn't parse both to compare. Once should be thrown into parse to obtain the actual, the other should be the expected value.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        Date expectedSinceDate = getDate(2017, Calendar.JULY, 1);
+        Date expectedUntilDate = getDate(2017, Calendar.NOVEMBER, 30);
+        Assert.assertEquals(expectedSinceDate, cliArguments.getSinceDate().get());
+        Assert.assertEquals(expectedUntilDate, cliArguments.getUntilDate().get());
+    }
+
+    @Test
+    public void parse_configFileOnly_success() throws ParseException {
+        String configFile = Paths.get(projectDirectory, "sample_small.csv").toAbsolutePath().toString();
```
eugenepeh: same implementation as above, but why coded differently?
Also, if you going to code this for every method, use `@setup` or `@beforetest` instead.

yong24s:
@eugenepeh @yamidark @AdityaA1998 
Ready for review.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,101 @@
...
+                .type(new DateArgumentType())
+                .setDefault(Optional.empty())
+                .help("The date to stop filtering.");
+
+        return parser;
+    }
+
+    /**
+     * Parses the given string arguments to a {@code CliArguments} object.
+     * @throws ParseException if the given string fails to parse to a CliArguments object.
```
yamidark: `@throws ParseException if the given string arguments fails to parse to a {@code CliArguments} object`

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,101 @@
...
+            Optional<Date> sinceDate = results.get("since");
+            Optional<Date> untilDate = results.get("until");
+
+            Path configFilePath = Paths.get(configFile.toURI());
+            Path outputFilePath = Paths.get(outputFile.toURI());
+
+            verifyDatesRangeIsCorrect(sinceDate, untilDate);
+            return new CliArguments(configFilePath, outputFilePath, sinceDate, untilDate);
+        } catch (ArgumentParserException ape) {
+            throw new ParseException(String.format(getArgumentParser().formatUsage() + ape.getMessage()));
```
yamidark: is `String.format` even needed?

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    /**
+     * Verifies that {@code sinceDate} is earlier than {@code untilDate}.
+     * @throws IllegalArgumentException if user-supplied values are invalid.
```
yamidark: `@throws IllegalArgumentException if {@code sinceDate} supplied is later than {@code untilDate}`
if the supplied values are invalid, we would get a `ParseException` doing `parse` instead right?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+    private static final String OUTPUT_DIRECTORY_ABSOLUTE = Paths.get(PROJECT_DIRECTORY, "build").toString();
+
+    private static final String CONFIG_FILE_RELATIVE = Paths.get("sample_small.csv").toString();
+    private static final String OUTPUT_DIRECTORY_RELATIVE = Paths.get("build").toString();
+
+    private Date getDate(int year, int month, int date) {
+        return new Calendar
+                .Builder()
+                .setDate(year, month, date)
+                .setTimeOfDay(0, 0, 0)
```
yamidark: is this necessary?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Test(expected = ParseException.class)
+    public void absoluteDirectory_forConfigFile_notAllowed() throws ParseException {
+        String absDirectory = Paths.get(PROJECT_DIRECTORY).getParent().toString();
+        String input = String.format("-config %s", absDirectory);
+        ArgsParser.parse(input.split(INPUT_DELIMITER));
+    }
+
+    @Test(expected = ParseException.class)
+    public void relativeDirectory_forConfigFile_notAllowed() throws ParseException {
```
yamidark: `_throwsParseException()`
same for below

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+        CliArguments cliArgumentsOfRelativePath = ArgsParser.parse(input.split(INPUT_DELIMITER));
+
+        Path expected = Paths.get(CONFIG_FILE_ABSOLUTE);
+        Assert.assertEquals(expected, cliArgumentsOfAbsolutePath.getConfigFilePath());
+        Assert.assertEquals(expected, cliArgumentsOfRelativePath.getConfigFilePath());
+    }
+
+    @Test
+    public void parse_configFileAndOutputDirectory_success() throws ParseException {
+        String input = String.format("-config %s -output %s", CONFIG_FILE_ABSOLUTE, OUTPUT_DIRECTORY_RELATIVE);
```
yamidark: for all these tests and below, can we have a variable:
`private static final String DEFAULT_MANDATORY_ARGS = "-config " + CONFIG_FILE_ABSOLUTE;"`
then each of the other tests will use this `DEFAULT_MANDATORY_ARGS + String.format(...))`, make it easier to see which specific argument we testing

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    @Test
+    public void optionalArguments_haveDefaultValues() throws ParseException {
+        String input = String.format("-config %s", CONFIG_FILE_ABSOLUTE);
+        CliArguments cliArguments = ArgsParser.parse(input.split(INPUT_DELIMITER));
+
+        Assert.assertEquals(Optional.empty(), cliArguments.getSinceDate());
+        Assert.assertEquals(Optional.empty(), cliArguments.getUntilDate());
+        Assert.assertEquals(new File(".").getAbsolutePath(), cliArguments.getTargetFilePath().toString());
```
yamidark: Can put this check inside the `configFileOnly` test

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Test(expected = ParseException.class)
+    public void nonExistentDirectory_forOutputDirectory_notAllowed() throws ParseException {
+        String nonExistentDirectory = Paths.get(PROJECT_DIRECTORY, "some_non_existent_dir/").toString();
+        String input = String.format("-config %s -output %s", CONFIG_FILE_ABSOLUTE, nonExistentDirectory);
+        ArgsParser.parse(input.split(INPUT_DELIMITER));
+    }
+
+    @Test(expected = ParseException.class)
+    public void sinceDate_nonSupportedFormats_notAllowed() throws ParseException {
```
yamidark: test for if `sinceDate` later than `untilDate` throws `iae` as well

src/main/java/reposense/frontend/RepoSense.java
```diff
@@ -5,18 +5,16 @@
 import java.util.List;
 
 import reposense.dataobject.RepoConfiguration;
+import reposense.exception.ParseException;
+import reposense.parser.ArgsParser;
 import reposense.report.RepoInfoFileGenerator;
 import reposense.system.CsvConfigurationBuilder;
```
AdityaA1998: Need any header comment?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,174 @@
...
+    private static final String OUTPUT_DIRECTORY_RELATIVE = Paths.get("build").toString();
+    private static final String DEFAULT_MANDATORY_ARGS = "-config " + CONFIG_FILE_ABSOLUTE + " ";
+
+    @Test
+    public void parse_allCorrectInputs_success() throws ParseException {
+        String input = DEFAULT_MANDATORY_ARGS + String.format("-output %s -since 01/07/2017 -until 30/11/2017",
+                OUTPUT_DIRECTORY_ABSOLUTE);
+        CliArguments cliArguments = ArgsParser.parse(input.split(INPUT_DELIMITER));
+        Assert.assertEquals(CONFIG_FILE_ABSOLUTE, cliArguments.getConfigFilePath().toString());
+        Assert.assertEquals(OUTPUT_DIRECTORY_ABSOLUTE, cliArguments.getTargetFilePath().toString());
```
yamidark: This 2 should be testing `Path` equals as well (below)?

src/test/java/reposense/parser/ArgsParserTest.java
yong24s: Done. :+1:

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    private static final String CONFIG_FILE_ABSOLUTE = Paths.get(PROJECT_DIRECTORY, "sample_small.csv").toString();
+    private static final String OUTPUT_DIRECTORY_ABSOLUTE = Paths.get(PROJECT_DIRECTORY, "build").toString();
+    private static final String CONFIG_FILE_RELATIVE = Paths.get("sample_small.csv").toString();
+    private static final String OUTPUT_DIRECTORY_RELATIVE = Paths.get("build").toString();
+    private static final String DEFAULT_MANDATORY_ARGS = "-config " + CONFIG_FILE_ABSOLUTE + " ";
+
+    @Test
+    public void parse_allCorrectInputs_success() throws ParseException {
+        String input = DEFAULT_MANDATORY_ARGS + String.format("-output %s -since 01/07/2017 -until 30/11/2017",
```
yamidark: Since this portion is also testing for the `config` settings, should add it the `String.format` instead of using `DEFAULT_MANDATORY_ARGS`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,174 @@
...
+        Date expectedSinceDate = TestUtil.getDate(2017, Calendar.JULY, 1);
+        Date expectedUntilDate = TestUtil.getDate(2017, Calendar.NOVEMBER, 30);
+        Assert.assertEquals(expectedSinceDate, cliArguments.getSinceDate().get());
+        Assert.assertEquals(expectedUntilDate, cliArguments.getUntilDate().get());
+    }
+
+    @Test
+    public void parse_configFileOnly_success() throws ParseException {
+        String input = String.format("-config %s", CONFIG_FILE_ABSOLUTE);
+        CliArguments cliArgumentsOfAbsolutePath =  ArgsParser.parse(input.split(INPUT_DELIMITER));
```
yamidark: extra whitespace after `=`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,174 @@
...
+        CliArguments cliArgumentsOfRelativePath = ArgsParser.parse(input.split(INPUT_DELIMITER));
+
+        Path expected = Paths.get(CONFIG_FILE_ABSOLUTE);
+        Assert.assertEquals(expected, cliArgumentsOfAbsolutePath.getConfigFilePath());
+        Assert.assertEquals(expected, cliArgumentsOfRelativePath.getConfigFilePath());
+
+        //Optional arguments have default values
+        Assert.assertEquals(Optional.empty(), cliArgumentsOfAbsolutePath.getSinceDate());
+        Assert.assertEquals(Optional.empty(), cliArgumentsOfAbsolutePath.getUntilDate());
+        Assert.assertEquals(new File(".").getAbsolutePath(), cliArgumentsOfAbsolutePath.getTargetFilePath().toString());
```
yamidark: Use `new File(".").toPath()` to test using `Path`

src/test/java/reposense/util/TestUtil.java
```diff
@@ -55,4 +57,16 @@ public static boolean compareFileContents(Path expected, Path actual, int maxTra
         }
         return true;
     }
+
+    /**
+    * Creates and returns a {@code Date} object with the specified {@code year}, {@code month}, {@code day} specified
```
yamidark: no need mention `specified` twice

src/test/java/reposense/parser/ArgsParserTest.java
yamidark: Actually, from this [SO](https://stackoverflow.com/questions/29368308/java-nio-how-is-files-issamefile-different-from-path-equals), `Paths.equals(Object)` may not return `true` even if both `Path` points to the same file/directory, instead we should use `Files.isSameFile(Path, Path)` here.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,176 @@
...
+    public void parse_configFileOnly_success() throws ParseException, IOException {
+        String input = String.format("-config %s", CONFIG_FILE_ABSOLUTE);
+        CliArguments cliArgumentsOfAbsolutePath = ArgsParser.parse(input.split(INPUT_DELIMITER));
+
+        input = String.format("-config %s", CONFIG_FILE_RELATIVE);
+        CliArguments cliArgumentsOfRelativePath = ArgsParser.parse(input.split(INPUT_DELIMITER));
+
+        Path expected = Paths.get(CONFIG_FILE_ABSOLUTE);
+        Assert.assertEquals(expected, cliArgumentsOfAbsolutePath.getConfigFilePath());
+        Assert.assertEquals(expected, cliArgumentsOfRelativePath.getConfigFilePath());
```
yamidark: This part too `Files.isSameFile(Path, Path)`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,176 @@
...
+    private static final String OUTPUT_DIRECTORY_RELATIVE = Paths.get("build").toString();
+    private static final String DEFAULT_MANDATORY_ARGS = "-config " + CONFIG_FILE_ABSOLUTE + " ";
+
+    @Test
+    public void parse_allCorrectInputs_success() throws ParseException {
+        String input = String.format("-config %s -output %s -since 01/07/2017 -until 30/11/2017",
+                CONFIG_FILE_ABSOLUTE, OUTPUT_DIRECTORY_ABSOLUTE);
+        CliArguments cliArguments = ArgsParser.parse(input.split(INPUT_DELIMITER));
+        Assert.assertEquals(CONFIG_FILE_ABSOLUTE, cliArguments.getConfigFilePath().toString());
+        Assert.assertEquals(OUTPUT_DIRECTORY_ABSOLUTE, cliArguments.getOutputFilePath().toString());
```
yamidark: This part too `Files.isSameFile(Path, Path)`

src/main/java/reposense/frontend/RepoSense.java
yong24s: As discussed, probably not.

yong24s:
@eugenepeh For your review, please.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -0,0 +1,102 @@
...
+            verifyDatesRangeIsCorrect(sinceDate, untilDate);
+            return new CliArguments(configFilePath, outputFilePath, sinceDate, untilDate);
+        } catch (ArgumentParserException ape) {
+            throw new ParseException(getArgumentParser().formatUsage() + ape.getMessage() + "\n");
+        }
+    }
+
+    /**
+     * Verifies that {@code sinceDate} is earlier than {@code untilDate}.
+     * @throws ParseException if {@code sinceDate} supplied is later than {@code untilDate}
```
eugenepeh: missing eol full-stop.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+import org.junit.Test;
+
+import reposense.exception.ParseException;
+import reposense.frontend.CliArguments;
+import reposense.util.TestUtil;
+
+public class ArgsParserTest {
+
+    private static final String INPUT_DELIMITER = " ";
+    private static final String PROJECT_DIRECTORY = System.getProperty("user.dir");
```
eugenepeh: Not very good, what if I try to run the test from some other directory?

Would it be safer to do it somewhat similarly like functional test?https://github.com/reposense/RepoSense/blob/b7bab4cea243aca7ea857d49d7c3935ee1e276c0/src/functional/java/reposense/Entry.java#L43

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    private static final String CONFIG_FILE_ABSOLUTE = Paths.get(PROJECT_DIRECTORY, "sample_small.csv").toString();
+    private static final String OUTPUT_DIRECTORY_ABSOLUTE = Paths.get(PROJECT_DIRECTORY, "build").toString();
+    private static final String CONFIG_FILE_RELATIVE = Paths.get("sample_small.csv").toString();
+    private static final String OUTPUT_DIRECTORY_RELATIVE = Paths.get("build").toString();
```
eugenepeh: Try not to use build - that is occupied by gradle.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+        Assert.assertEquals(Optional.empty(), cliArgumentsOfRelativePath.getSinceDate());
+        Assert.assertEquals(Optional.empty(), cliArgumentsOfRelativePath.getUntilDate());
+        Assert.assertTrue(Files.isSameFile(Paths.get("."), cliArgumentsOfRelativePath.getOutputFilePath()));
+    }
+
+    @Test
+    public void parse_configFileAndOutputDirectory_success() throws ParseException, IOException {
+        String input = String.format("-config %s -output %s", CONFIG_FILE_ABSOLUTE, OUTPUT_DIRECTORY_RELATIVE);
+        CliArguments cliArgAbsConfigRelOutput = ArgsParser.parse(input.split(INPUT_DELIMITER));
```
eugenepeh: Have you explore such tools as mentioned in this SO? https://stackoverflow.com/questions/3259143/split-a-string-containing-command-line-parameters-into-a-string-in-java

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+        String absConfigFile = Paths.get(PROJECT_DIRECTORY, "somefile_not_existent.csv").toString();
+        String input = String.format("-config %s", absConfigFile);
+        ArgsParser.parse(input.split(INPUT_DELIMITER));
+    }
+
+    @Test(expected = ParseException.class)
+    public void parse_relativeConfigFileNotExists_throwsArgumentParserException() throws ParseException {
+        String relConfigFile = Paths.get("somefile_not_existent.csv").toString();
+        String input = String.format("-config %s", relConfigFile);
+        ArgsParser.parse(input.split(INPUT_DELIMITER));
```
eugenepeh: I think absolute is sufficient for this case. How about testing empty config param?
`-config `

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+    @Test(expected = ParseException.class)
+    public void nonExistentDirectory_forOutputDirectory_throwsParseException() throws ParseException {
+        String nonExistentDirectory = Paths.get(PROJECT_DIRECTORY, "some_non_existent_dir/").toString();
+        String input = DEFAULT_MANDATORY_ARGS + String.format("-output %s", nonExistentDirectory);
+        ArgsParser.parse(input.split(INPUT_DELIMITER));
+    }
+
+    @Test(expected = ParseException.class)
+    public void sinceDate_nonSupportedFormats_throwsParseException() throws ParseException {
```
eugenepeh: `unsupportedFormats`

src/test/java/reposense/parser/ArgsParserTest.java
yong24s: I have tried implementing with ```ClassLoader``` to get **project base directory** but JUnit tests only passed in IntelliJ and failed in gradlew and travis. 

On further investigations, it seems like ```gradlew``` can only be run in the **project base directory** so using ```System.getProperty("user.dir")``` makes sense.

src/test/java/reposense/parser/ArgsParserTest.java
eugenepeh: Yes... but this prevent the mistake where developers might accidentally type an additional space in the input line and try to find the error all over the place.

eugenepeh: Also, it will be nice to keep the test arguments as close as possible to what user might type.

src/main/java/reposense/parser/DateArgumentType.java
```diff
@@ -0,0 +1,28 @@
...
+ * Verifies and parses a string-formatted date to a {@code Date} object.
+ */
+public class DateArgumentType implements ArgumentType<Optional<Date>> {
+    private static final String PARSE_EXCEPTION_MESSAGE_INVALID_DATE_STRING_FORMAT = "Invalid Date: %s";
+
+    @Override
+    public Optional<Date> convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
+        try {
+            return Optional.of(Constants.CLI_ARGS_DATE_FORMAT.parse(value));
+        } catch (ParseException pe) {
```
eugenepeh: so where does the parse exception come from?

eugenepeh: Also consider shifting Constants.CLI_ARGS_DATE_FORMAT to here instead.

src/test/java/reposense/parser/ArgsParserTest.java
eugenepeh: Hmm okay, but place test files within the test folder (maybe like functional, create its own resource folder) to prevent any modification not relevant to the test.

src/test/java/reposense/parser/ArgsParserTest.java
eugenepeh: Also, `cliArgAbsConfigRelOutput` is too complicated. Your first line is explanatory enough. 
Just name it as usual, otherwise you'll have to make all the others occurrence follow this standard too.

src/main/java/reposense/parser/DateArgumentType.java
yong24s: It is actually ```java.text.ParseException```. I will use its full package name to avoid confusion.

eugenepeh:
https://stackoverflow.com/questions/21829295/remove-filename-from-a-url-path-in-java
for relative path

eugenepeh:
Seems like travis for ur pr has hanged. Please rebase onto latest master too.

yong24s:
@damithc Prof., for your review, please.

src/main/java/reposense/frontend/RepoSense.java
```diff
@@ -5,41 +5,26 @@
...
             Date fromDate = cliArguments.getSinceDate().orElse(null);
             Date toDate = cliArguments.getUntilDate().orElse(null);
 
             List<RepoConfiguration> configs = CsvConfigurationBuilder.buildConfigs(configFile, fromDate, toDate);
             RepoInfoFileGenerator.generateReposReport(configs, targetFile.toAbsolutePath().toString());
-        } catch (IllegalArgumentException ex) {
-            System.out.print(ex.getMessage());
-            showHelpMessage();
+        } catch (ParseException pe) {
+            System.out.print(pe.getMessage());
```
eugenepeh: one more minor fix - due to the recent logger addition, please change this to use logger

yamidark:
Not sure why, but on my Windows, I'm getting an error:
```
java.lang.ExceptionInInitializerError
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:217)
	at org.junit.runners.BlockJUnit4ClassRunner$1.runReflectiveCall(BlockJUnit4ClassRunner.java:266)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:263)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:137)
	at com.intellij.junit4.JUnit4IdeaTestRunner.startRunnerWithArgs(JUnit4IdeaTestRunner.java:68)
	at com.intellij.rt.execution.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:47)
	at com.intellij.rt.execution.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:242)
	at com.intellij.rt.execution.junit.JUnitStarter.main(JUnitStarter.java:70)
Caused by: java.nio.file.InvalidPathException: Illegal char <:> at index 2: /C:/Stuff/SE-EDU/RepoSense/out/test
```
due to this line:
` private static final Path CONFIG_FILE_ABSOLUTE = Paths.get(ArgsParserTest.class.getClassLoader()
            .getResource("output/parser_test.csv").getPath());`

eugenepeh:
Issue verified, seems like it's because URL will return an additional `/` at the front.
 @yong24s please rectify the issue.

yamidark:
>seems like it's because URL will return an additional / at the front.

Seems like using `toPath()` on the `URI` will result in the additional ` /` at the front.
We can resolve this by using `toURI()` instead, but this will require throwing/catching of an `URISyntaxException`.
One way would be to static constructors with `try/catch` blocks (not tested),
or we can use JUnit `BeforeClass` to initialize each variable, but they can no longer be final (tested).

eugenepeh:
>Seems like using toPath() on the URI will result in the additional / at the front.
We can resolve this by using toURI() instead, but this will require throwing/catching of an URISyntaxException.
One way would be to static constructors with try/catch blocks (not tested),
or we can use JUnit BeforeClass to initialize each variable, but they can no longer be final (tested).

Keep it simpler, replacefirst will do. Most likely it doesn't happen in linux/mac because a slash infront wouldn't make a different. Which is also why this problem did not occur to me when I was using a network drive.

Issue #92, opened 2018-05-14
FileInfo: date range does not affect code contribution


eugenepeh:
The date range entered in the CLI does not affect the total code contribution collated in the FileInfo.

harryggg:
go thru commit list -> find the latest commit thats in the date range -> checkout that commit -> blame
However, in the current architecture, the date range in dashboard cant affect the total code contribution.

PR #137, opened 2018-06-04
FileInfo: date range does not affect code contribution #92

yamidark:
The date range entered in the CLI does not affect the total code contribution collated in the FileInfo.

This is because the git blame command used is analyzing the entire commit history, instead of only
the commits within the date range specified.

Let's update the CommandRunner#blameRaw(String, String) git blame command to specify the
date range of commits to analyze, if specified.


yamidark:
@yong24s @AdityaA1998 @ongspxm for your review

yong24s:
LGTM.

yamidark:
@eugenepeh for your review too

src/main/java/reposense/dataobject/RepoConfiguration.java
```diff
@@ -9,7 +9,6 @@
 
 import reposense.util.FileUtil;
 
-
```
eugenepeh: avoid unnecessary change if you didn't modify the file.

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -25,14 +25,37 @@ public static String gitLog(String root, Date fromDate, Date toDate) {
...
         Path rootPath = Paths.get(root);
         runCommand(rootPath, "git checkout " + hash);
     }
 
-    public static String blameRaw(String root, String fileDirectory) {
+    /**
+     * Checkout to the latest commit before {@code untilDate} in {@code branchName} branch
+     * if {@code untilDate} is not null
+     */
+    public static void checkoutToDate(String root, String branchName, Date untilDate) {
```
eugenepeh: As quoted from git documentation
`When you are not interested in changes older than version v2.6.18, or changes older than 3 weeks, you can use revision range specifiers similar to git rev-list:`

if its similar to git rev list, is there any point in doing this?

src/test/java/reposense/system/CommandRunnerTest.java
```diff
@@ -58,7 +58,8 @@ public void logWithoutContentTest() {
     @Test
     public void blameRawTest() {
         try {
-            String content = CommandRunner.blameRaw(TestConstants.LOCAL_TEST_REPO_ADDRESS, "blameTest.java");
+            String content = CommandRunner.blameRaw(
+                    TestConstants.LOCAL_TEST_REPO_ADDRESS, "blameTest.java", null, null);
```
eugenepeh: instead of doing this, perhaps an overload method will make it looks cleaner?

src/main/java/reposense/system/CommandRunner.java
yamidark: This is to ensure the `FileInfo` generated is based on the latest commit of the date range.
Should there be any changes to the files after we analyze from the latest commit of the date range, the blame results generated using this date range may not correspond to the correct lines of the 'updated' files, and we may even get `IndexOutOfBoundException` if the file has deleted lines.

src/main/java/reposense/dataobject/RepoConfiguration.java
yamidark: Whoops, modified this file earlier but removed the changes xD

src/test/java/reposense/system/CommandRunnerTest.java
yamidark: Hmmm, but this overloaded method would also be used for tests though?

src/main/java/reposense/system/CommandRunner.java
eugenepeh: i see. can add it to the header comments too?

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -129,5 +123,11 @@ private static boolean isWindows() {
         return (System.getProperty("os.name").toLowerCase().indexOf("win") >= 0);
     }
 
-
+    private static String getFilterBlameResultsCommand() {
```
eugenepeh: Perhaps `getAuthorFilterCommand` would be better?

src/test/java/reposense/util/TestUtil.java
```diff
@@ -55,4 +57,14 @@ public static boolean compareFileContents(Path expected, Path actual, int maxTra
         }
         return true;
     }
+
+    /**
+     * Creates and returns a {@code Date} object with the specified {@code year}, {@code month}, {@code day} specified
```
eugenepeh: full stop behind

yamidark:
@ongspxm @yong24s @AdityaA1998 for your review

yong24s:
LGTM :medal_sports:

src/main/java/reposense/git/GitChecker.java
```diff
@@ -20,8 +22,15 @@ public static void checkOutToCommit(String root, CommitInfo commit) {
     }
 
     public static void checkout(String root, String commitHash) {
-        CommandRunner.checkOut(root, commitHash);
+        CommandRunner.checkout(root, commitHash);
     }
 
+    /**
+     * Checkout to the latest commit before {@code untilDate} in {@code branchName} branch
```
damithc: Checks out?

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -108,5 +115,31 @@ private static boolean isWindows() {
...
+    /**
+     * Returns the {@code String} command which filters the git blame output to produce only the necessary author
+     * name for each line.
+     */
+    private static String getAuthorFilterCommand() {
+        if (isWindows) {
+            return "| findstr /B /C:" + addQuote("author ");
+        } else {
+            return "| grep " + addQuote("^author .*");
+        }
```
damithc: A good place to use the tertiary operator? `return isWindows ? ... : ...`

src/test/java/reposense/system/CommandRunnerTest.java
```diff
@@ -22,48 +26,63 @@ public void cloneTest() {
...
+            Assert.assertFalse(content.isEmpty());
+
+            // no sinceDate or untilDate supplied -> success
+            content = CommandRunner.blameRaw(
+                    TestConstants.LOCAL_TEST_REPO_ADDRESS, "blameTest.java", null, null);
+            Assert.assertFalse(content.isEmpty());
         } catch (Exception e) {
-            Assert.assertTrue(false);
+            fail();
         }
```
damithc: Can avoid the try-catch if you add `throws Exception` to the method signature. An exception fails the test anyway.

yamidark:
@damithc @eugenepeh Made the changes and rebased to `master`

yamidark:
>hmm ran thrice, keep having problem with daterange but no problem with the other.
Could you please look into it?

Just found out that Windows command prompt does not allow command substitution, thus failing for the 
```
"git checkout `git rev-list 1 --before="
                + Constants.GIT_LOG_UNTIL_DATE_FORMAT.format(untilDate) + " " + branchName + "`";
```
command.
Resolved this by using [Loop command](https://ss64.com/nt/for_f.html) for Windows

eugenepeh:
do you think we can do something like that to enhance readability?

```
substituteCommand = git rev-list -n 1 --before=" + Constants.GIT_LOG_UNTIL_DATE_FORMAT.format(untilDate) + " " + branchName;

checkoutCommand = "FOR /f %g IN (" + substitute_command + ") DO git checkout %g";
```

yamidark:
@eugenepeh 
>do you think we can do something like that to enhance readability?

Hmm, something like this?
```Java
        String checkoutCommand;
        String substituteCommand = "`git rev-list -1 --before="
                + Constants.GIT_LOG_UNTIL_DATE_FORMAT.format(untilDate) + " " + branchName + "`";

        if (isWindows) {
            checkoutCommand = "for /f %g in (" + substituteCommand + ") do git checkout %g";
        } else {
            checkoutCommand = "git checkout " + substituteCommand;
        }
```

eugenepeh:
image

Still seems to have some error. Are you able to replicate it?

yamidark:
@eugenepeh 
>Still seems to have some error. Are you able to replicate it?

Resolved this. Forgot Windows need to use `'` single quotes while bash uses `, and mixed them up earlier
:sweat:

eugenepeh:
@AdityaA1998 please review this PR as well

AdityaA1998:
LGTM. It works perfectly fine on macOS as well

PR #141, opened 2018-06-11
Fix failing ArgParser unit tests on Windows systems

yong24s:
Proposal Message:
```
ArgsParser unit tests use ClassLoader#getResource(String) to get the path to
a resource file/directory.

According to this SO post[1], ClassLoader#getResource(String) adds an extra
leading slash before a path on Windows systems, which may cause the path
we get to be invalid, resulting in all ArgsParser's unit tests to fail.

Let's update ArgsParserTest to use File(String).toPath() to prevent
the extra leading slash in Windows system.

[1]: Leading slash before disk name in Windows:
https://stackoverflow.com/questions/15662820/getresource-puts-a-leading-before-the-disk-name-using-java-1-7-windows-7
```

### DIFF
--- a/src/test/java/reposense/parser/ArgsParserTest.java
+++ b/src/test/java/reposense/parser/ArgsParserTest.java
@@ -21,10 +21,10 @@ import reposense.util.TestUtil;
 public class ArgsParserTest {
 
     private static final Path PROJECT_DIRECTORY = Paths.get(System.getProperty("user.dir"));
-    private static final Path CONFIG_FILE_ABSOLUTE = Paths.get(new File(ArgsParserTest.class.getClassLoader()
-            .getResource("output/parser_test.csv").getPath()).getPath());
-    private static final Path OUTPUT_DIRECTORY_ABSOLUTE = Paths.get(new File(ArgsParserTest.class.getClassLoader()
-            .getResource("output").getPath()).getPath());
+    private static final Path CONFIG_FILE_ABSOLUTE = new File(ArgsParserTest.class.getClassLoader()
+            .getResource("output/parser_test.csv").getFile()).toPath();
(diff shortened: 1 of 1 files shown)


src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -20,10 +21,10 @@
...
 
     private static final Path PROJECT_DIRECTORY = Paths.get(System.getProperty("user.dir"));
-    private static final Path CONFIG_FILE_ABSOLUTE = Paths.get(ArgsParserTest.class.getClassLoader()
-            .getResource("output/parser_test.csv").getPath());
-    private static final Path OUTPUT_DIRECTORY_ABSOLUTE = Paths.get(ArgsParserTest.class.getClassLoader()
-            .getResource("output").getPath());
+    private static final Path CONFIG_FILE_ABSOLUTE = Paths.get(new File(ArgsParserTest.class.getClassLoader()
+            .getResource("output/parser_test.csv").getPath()).getPath());
+    private static final Path OUTPUT_DIRECTORY_ABSOLUTE = Paths.get(new File(ArgsParserTest.class.getClassLoader()
+            .getResource("output").getPath()).getPath());
```
yamidark: Looks abit weird to have 2 `.getPath()` though 😅
Maybe this would look nicer and simpler?
```Java
new File(ArgsParserTest.class.getClassLoader()
            .getResource("output/parser_test.csv").getFile()).toPath();
```

yamidark:
`getResource(String)` -> `ClassLoader#getResource(String)`

AdityaA1998:
Yep! All the tests pass on macOS.

yong24s:
@eugenepeh, for review, please.

PR #152, opened 2018-06-26
FileUtil: add zip function for offline preview

AdityaA1998:
Proposed message:
```
In order to view the dashboard, we need to run a local server as the security 
in web browsers does not allow it to traverse the directories on local machine.

Let's zip all the necessary json files together to make our application 
compatible with all the browsers and bypass the need to run the application
on local server. 
```


eugenepeh:
@ongspxm @AdityaA1998 just a quick question, is it necessary to zip and unzip?

AdityaA1998:
@eugenepeh we need to zip as the browser cannot read files due to the security features, and we need to run the local server to see the dash board. After zipping, all those issues are solved. Also, it solves the issue of browser compatibility. 

Unzip is being used for the template.

eugenepeh:
Is there any benefit compare to using a json object stored in a js file?

ongspxm:
if we dump the whole json object into one js file, then the browser will die trying to load just the summary. by loading it from a zip, the files are kept somewhere in the the browser "memory" until we need the file then load it into the js object.

AdityaA1998:
@yamidark @yong24s @ongspxm review plz

eugenepeh:
> if we dump the whole json object into one js file, then the browser will die trying to load just the summary. by loading it from a zip, the files are kept somewhere in the the browser "memory" until we need the file then load it into the js object.

What if you break it into several objects stored in the same js file as mentioned in this [SO](https://stackoverflow.com/questions/4833480/have-i-reached-the-limits-of-the-size-of-objects-javascript-in-my-browser-can-ha/4833756)? How's the performance in comparison?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +126,67 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+     */
+    public static void zip(String directoryPath) throws IOException {
+
+        String zipFileName = "archiveJSON.zip";
+
+        List<Path> allJsonFiles = listJsonFiles(directoryPath);
+
+        try {
+            //byte buffer for I/O
+            byte[] buffer = new byte[1024];
```
yong24s: Try using ```ByteBuffer``` if possible

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +126,67 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+                }
+
+                while ((length = fis.read(buffer)) > 0) {
+                    zos.write(buffer, 0, length);
+                }
+                zos.closeEntry();
+                fis.close();
+            }
+
+            zos.close();
```
yong24s: Its unsafe to close streams here, do it in ```finally``` block or use a ```try-with-resource``` block.

Read: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

ongspxm:
> What if you break it into several objects stored in the same js file as mentioned in this SO? How's the performance in comparison?

the zip feature is a workaround, the aim is to split the reports into multiple files so that when served up on a server, we are able to cut the number of request for a more responsive dashboard.

to run locally (without running server), the problem is with traversing directory. to work around that, we have to get all the content into one file either ways.

two ways to do this with the current state of the backend. 
1. generate everything into one giant js file
2. fake the directory traversal, by traversing a zip file rather than local path

assuming we are generating the one big file (OBF) post report generation (because we wouldn't want to hog the memory, and would like to churn out the reports as soon as possible).

**option 1**
backend wise, 1 will require the parsing of the JSON files then dumping them into the OBF after parsing it into the individual objects then printing it back out. 

"why not just copy the contents and prepend in a stadard object name, like we did last time?" well, it will still work, but for every new component we add, we will have to alter this part to add in the new object.

frontend wise, this will require some major changes to how the objects is read into the application. Have to maintain both local-side and server-side data representation, which will be double work.

**option 2**
using zip as our OBF means that we can skip the parsing and just wholesale dump all the json file in, we don't even need to maintain this much. assuming we are gonna stick to json files in the near future.

frontend side, we dun really need to do much, just change the XHR to traverse the zip file rather than the actual URI.

at the end of the day, i just wanna write less code... please dun do this to me... :confounded:

yamidark:
As mentioned earlier, can try integrating this into `ReportGenerator` and `FileUtil.writeJsonFile(Object, Path)` instead of traversing the entire directory and reread each JSON file.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+    }
+
+    /**
+     * Zips all the JSON files contained in the {@generatedFolder} directory.
+     * Creates the zip folder in the {@code directoryPath}
+     */
+    public static void zip(String directoryPath, String generatedFolder) {
+
+        String zipFileName = "archiveJSON.zip";
+
```
yong24s: Unnecessary line

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+        }
+
+    }
+
     private static String attachJsPrefix(String original, String prefix) {
         return "var " + prefix + " = " + original;
     }
 
+    /**
+     * Checks if the given {@path} is of summary.json
```
yong24s: ```{@code path}```

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+                    .collect(Collectors.toList());
+        } catch (IOException ioe) {
+            logger.log(Level.SEVERE, ioe.getMessage(), ioe);
+        }
+
+        return allJsonFiles;
+    }
+
+    /**
+     * Zips all the JSON files contained in the {@generatedFolder} directory.
```
yong24s: ```{@code generatedFolder}```

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+
+                    while ((length = fis.read(buffer.array())) > 0) {
+                        zos.write(buffer.array(), 0, length);
+                    }
+                }
+            }
+        } catch (IOException ioe) {
+            logger.log(Level.SEVERE, ioe.getMessage(), ioe);
+        }
+
```
yong24s: Unnecessary line

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+        return allJsonFiles;
+    }
+
+    /**
+     * Zips all the JSON files contained in the {@generatedFolder} directory.
+     * Creates the zip folder in the {@code directoryPath}
+     */
+    public static void zip(String directoryPath, String generatedFolder) {
+
```
yong24s: Unnecessary line

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+        String zipFileName = "archiveJSON.zip";
+
+        List<Path> allJsonFiles = listJsonFiles(directoryPath + File.separator + generatedFolder);
+
+        //byte buffer for I/O
```
yong24s: Redundant comment

eugenepeh:
a quick comment 
- put the unzip method together.
- clean up the unzip method according to the comments given to the zip as well.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+        }
+
+        return allJsonFiles;
+    }
+
+    /**
+     * Zips all the JSON files contained in the {@generatedFolder} directory.
+     * Creates the zip folder in the {@code directoryPath}
+     */
+    public static void zip(String directoryPath, String generatedFolder) {
```
yong24s: Should name the method ```zip``` to be more specific to fit the method header. 
For example, ```zipJson```.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+
+        String zipFileName = "archiveJSON.zip";
```
yong24s: As this method is in ```FileUtil```, this variable should not be hard coded. Pass it as a parameters, to make this method reusable.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+        //byte buffer for I/O
+        ByteBuffer buffer = ByteBuffer.allocate(1 << 10);
+        int length;
+        try (FileOutputStream fos = new FileOutputStream(directoryPath + File.separator + zipFileName);
+             ZipOutputStream zos = new ZipOutputStream(fos)) {
+            for (int i = 0; i < allJsonFiles.size(); i++) {
+                try (FileInputStream fis = new FileInputStream(allJsonFiles.get(i).toFile())) {
+
+                    // begin writing a new ZIP entry, positions the stream to the start of the entry data
+                    if (isSummaryJson(allJsonFiles.get(i))) {
```
yong24s: Same problem with the hard-coded ```archiveJSON.zip```. This method is very specific to be in ```FileUtil```.

Perhaps, you can generalize main zipping logic of this method, then use another wrapper to do a specific function.

Can discuss with @yamidark. If you need a code example, just @mention me.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,70 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
     private static String attachJsPrefix(String original, String prefix) {
         return "var " + prefix + " = " + original;
     }
 
+    /**
+     * Checks if the given {@path} is of summary.json
+     */
+    private static boolean isSummaryJson(Path path) {
+        return path.getFileName().toString().equals("summary.json");
+    }
```
ongspxm: maybe can do a isFileName or something, move the filename out.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,67 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+             ZipOutputStream zos = new ZipOutputStream(fos)) {
+            for (int i = 0; i < allJsonFiles.size(); i++) {
+                try (FileInputStream fis = new FileInputStream(allJsonFiles.get(i).toFile())) {
+
+                    // begin writing a new ZIP entry, positions the stream to the start of the entry data
+                    if (isFilePathValid(allJsonFiles.get(i), summaryJson)) {
+                        zos.putNextEntry(new ZipEntry(allJsonFiles.get(i).getFileName().toString()));
+                    } else {
+                        zos.putNextEntry(new ZipEntry(allJsonFiles.get(i).getParent().getFileName().toString()
+                                + File.separator + allJsonFiles.get(i).getFileName().toString()));
```
yamidark: Its not clear what this is doing, perhaps abstract this out to a method `addParentDirectory(Path)`?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,67 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+    }
+
     private static String attachJsPrefix(String original, String prefix) {
         return "var " + prefix + " = " + original;
     }
 
+    /**
+     * Checks if the given {@code path} is of {@code file}
+     */
+    private static boolean isFilePathValid(Path path, String file) {
```
yamidark: The method name does not reflect what it is doing, should be `checkFileName`? Update the header comments accordingly.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,67 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
         }
     }
 
+    /**
+     * Generates a list of paths of {@code fileType} contained in the given {@code directoryName} directory.
+     */
+    public static List<Path> listFilePath(String directoryName, String fileType) {
```
yamidark: `listFilePath` kinda sounds like we are printing out the file paths, which is not what we are doing here. Perhaps rename this as `getFilePaths`? 
Also, since this seems to just a helper method within `FileUtil`, can make it private.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +127,67 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
         }
     }
 
+    /**
+     * Generates a list of paths of {@code fileType} contained in the given {@code directoryName} directory.
```
yamidark: `Generates a list of paths...` -> `Returns a list of {@code Path}...`

src/main/java/reposense/RepoSense.java
```diff
@@ -23,8 +24,11 @@ public static void main(String[] args) {
             List<RepoConfiguration> configs = CsvParser.parse(cliArguments.getConfigFilePath());
             RepoConfiguration.setDatesToRepoConfigs(configs, cliArguments.getSinceDate(), cliArguments.getUntilDate());
 
-            ReportGenerator.generateReposReport(
+            String generatedFolderName = ReportGenerator.generateReposReport(
                     configs, cliArguments.getOutputFilePath().toAbsolutePath().toString());
+
+            FileUtil.zipJson(cliArguments.getOutputFilePath().toAbsolutePath().toString(), generatedFolderName);
```
yamidark: Should we put this into `ReportGenerator#generateReposReport(...)` instead, that we can do a test for the `.zip` file as well in functional test?

yamidark:
Remember to rebase to master too

src/main/java/reposense/RepoSense.java
AdityaA1998: But is zipping the jsons a part of generating the report?

yamidark: >But is zipping the jsons a part of generating the report?

Hmmm, not too sure on that part, let's ignore this for now.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +140,31 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+    /**
+     * Returns a list of paths of {@code fileType} contained in the given {@code directoryName} directory.
+     */
+    private static List<Path> getFilePath(String directoryName, String fileType) {
+        Path directory = Paths.get(directoryName);
+        List<Path> filePaths = null;
+        try {
+            filePaths = Files.walk(directory)
+                    .filter(p -> p.toString().endsWith(fileType))
+                    .distinct()
```
yamidark: Is there a need for `distinct()`? Each file should be already be distinct.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +62,57 @@ public static void deleteDirectory(String root) throws IOException {
...
-            while (entry != null) {
-                String entryName = entry.getName();
-                Path path = Paths.get(destinationFolder, entryName);
+    /**
+     * Unzips the contents of the {@code zipInput} and stores in the {@code destinationFolder}
+     */
+    public static void unzip(ZipInputStream zipInput, String destinationFolder) {
+        ByteBuffer buffer = ByteBuffer.allocate(1 << 11);
+        ZipEntry entry;
+        try (ZipInputStream zi = zipInput) {
```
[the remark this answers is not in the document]
AdityaA1998: I implemented it to maintain consistency, and to make sure that zipInputStream is auto-closed in case any exception occurs

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +62,57 @@ public static void deleteDirectory(String root) throws IOException {
...
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the JSON files contained in the {@code sourcePath} directory.
+     * Creates the zip folder in the {@code outputPath}
+     */
+    public static void zipJson(String outputPath, String sourcePath) {
+        List<Path> allJsonFiles = getFilePath(outputPath + File.separator + sourcePath, ".json");
+
+        ByteBuffer buffer = ByteBuffer.allocate(1 << 10);
+        int length;
```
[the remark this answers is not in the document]
AdityaA1998: It is redundant to initialise it as is only used after being initialised in the while loop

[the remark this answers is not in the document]
yamidark: Sorry, I meant to declare it then, as in put them together.

src/main/java/reposense/util/FileUtil.java
yamidark: `Files.walk(..)` shouldn't visit the same file more than once right?

src/main/java/reposense/util/FileUtil.java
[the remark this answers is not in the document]
yamidark: Hmm, still looks very weird though, maybe we can make the caller `FileUtil#copyTemplate(...)` to pass an `InputStream` instead, and initialize the `ZipInputStream` here instead?

AdityaA1998:
@eugenepeh @yamidark @yong24s @ongspxm for review

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +139,30 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
         }
     }
 
+    /**
+     * Returns a list of {@code Path} of {@code fileType} contained in the given {@code directoryName} directory.
+     */
+    private static List<Path> getFilePath(String directoryName, String fileType) {
```
yamidark: getFilePath**s**

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +63,55 @@ public static void deleteDirectory(String root) throws IOException {
...
-
-        try {
-            Files.createDirectories(directory);
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the JSON files contained in the {@code sourcePath} directory.
+     * Creates the zip folder in the {@code outputPath}
+     */
+    public static void zipJson(String outputPath, String sourcePath) {
+        List<Path> allJsonFiles = getFilePath(outputPath + File.separator + sourcePath, ".json");
```
yamidark: Put this inside the `try` block as well, then `getFilePaths(..)` can just throw the `IOException` up here.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +63,55 @@ public static void deleteDirectory(String root) throws IOException {
...
 
-            while (entry != null) {
-                String entryName = entry.getName();
-                Path path = Paths.get(destinationFolder, entryName);
+    /**
+     * Unzips the contents of the {@code zis} and stores in the {@code destinationFolder}
+     */
+    public static void unzip(ZipInputStream zis, String destinationFolder) {
+        ZipEntry entry;
+        try (ZipInputStream zi = zis) {
```
yamidark: As mentioned earlier, can we make `unzip` take in an `InputStream` instead, and create the new `ZipInputStream` here instead?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +63,55 @@ public static void deleteDirectory(String root) throws IOException {
...
-
-        // buffer for read and write data to file
-        byte[] buffer = new byte[2048];
-
-        try {
-            Files.createDirectories(directory);
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the JSON files contained in the {@code sourcePath} directory.
+     * Creates the zip folder in the {@code outputPath}
```
yamidark: `.` at the end.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +63,55 @@ public static void deleteDirectory(String root) throws IOException {
...
+     */
+    public static void zipJson(String outputPath, String sourcePath) {
+
```
yamidark: unnecessary line break.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +63,55 @@ public static void deleteDirectory(String root) throws IOException {
...
 
-            while (entry != null) {
-                String entryName = entry.getName();
-                Path path = Paths.get(destinationFolder, entryName);
+    /**
+     * Unzips the contents of the {@code is} and stores in the {@code destinationFolder}
+     */
+    public static void unzip(InputStream is, String destinationFolder) {
+        ZipEntry entry;
+        try (ZipInputStream zi = new ZipInputStream(is)) {
```
yamidark: `zi` -> `zis`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +63,54 @@ public static void deleteDirectory(String root) throws IOException {
...
+        }
+    }
 
-            while (entry != null) {
-                String entryName = entry.getName();
-                Path path = Paths.get(destinationFolder, entryName);
+    /**
+     * Unzips the contents of the {@code is} and stores in the {@code destinationFolder}.
+     */
+    public static void unzip(InputStream is, String destinationFolder) {
```
yong24s: Change all occurrence of destinationFolder -> outputPath

src/main/java/reposense/util/FileUtil.java
```diff
@@ -106,7 +122,7 @@ public static void unzip(ZipInputStream zipInput, String destinationFolder) {
     public static void copyTemplate(String outputPath, String reportName) {
         String templateLocation = outputPath + File.separator + reportName;
         InputStream is = RepoSense.class.getResourceAsStream(Constants.TEMPLATE_ZIP_ADDRESS);
-        FileUtil.unzip(new ZipInputStream(is), templateLocation);
+        FileUtil.unzip(is, templateLocation);
```
yong24s: Why not pass the path of template zip as a parameter? i.e. ```String sourcePath```
So that the ```unzip``` has similar contract as ```zipJson```

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +63,54 @@ public static void deleteDirectory(String root) throws IOException {
...
-        byte[] buffer = new byte[2048];
-
-        try {
-            Files.createDirectories(directory);
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the JSON files contained in the {@code sourcePath} directory.
+     * Creates the zip folder in the {@code outputPath}.
+     */
+    public static void zipJson(String outputPath, String sourcePath) {
```
yong24s: I think we should use ```src``` then ```dest```, which is the generally followed approach.

```(String outputPath, String sourcePath)``` -> ```(String sourcePath, String outputPath)```

**Comment by yong24s** [comment]
Minors nits, else LGTM.

File: src/main/java/reposense/util/FileUtil.java [hunk 30]
```diff
@@ -57,44 +63,55 @@ public static void deleteDirectory(String root) throws IOException {
...
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the JSON files contained in the {@code sourcePath} directory.
+     * Creates the zip folder in the {@code outputPath}.
+     */
+    public static void zipJson(String sourcePath, String outputPath) {
+        int length;
+        try (FileOutputStream fos = new FileOutputStream(outputPath + File.separator + Constants.JSON_ZIP_FILE);
+             ZipOutputStream zos = new ZipOutputStream(fos)) {
+            List<Path> allJsonFiles = getFilePaths(outputPath + File.separator + sourcePath, ".json");
```
eugenepeh: why is sourcepath dependent on output path?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -28,6 +32,8 @@
     private static Logger logger = LogsManager.getLogger(FileUtil.class);
 
     private static final String GITHUB_API_DATE_FORMAT = "yyyy-MM-dd";
+    private static final ByteBuffer buffer = ByteBuffer.allocate(1 << 11);
```
eugenepeh: same as the other buffer implementation, please add a similar comment

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +63,55 @@ public static void deleteDirectory(String root) throws IOException {
...
-
-        // buffer for read and write data to file
-        byte[] buffer = new byte[2048];
-
-        try {
-            Files.createDirectories(directory);
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the JSON files contained in the {@code sourcePath} directory.
+     * Creates the zip folder in the {@code outputPath}.
```
eugenepeh: ~zip~ `zipped`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -122,8 +138,26 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+                .filter(p -> p.toString().endsWith(fileType))
+                .collect(Collectors.toList());
+        return filePaths;
+    }
+
+    /**
+     * Returns the child contained in the {@code sourcePath} of the {@code fullPath}.
+     */
+    private static String getChild(String fullPath, String sourcePath) {
+        return fullPath.substring(fullPath.indexOf(sourcePath) + sourcePath.length());
```
eugenepeh: how is this even called child?

AdityaA1998:
@eugenepeh for review

AdityaA1998:
@eugenepeh for review

src/main/java/reposense/util/FileUtil.java
```diff
@@ -65,16 +65,19 @@ public static void deleteDirectory(String root) throws IOException {
...
+    public static void zipJson(Path sourcePath, Path outputPath) {
         int length;
-        try (FileOutputStream fos = new FileOutputStream(outputPath + File.separator + Constants.JSON_ZIP_FILE);
-             ZipOutputStream zos = new ZipOutputStream(fos)) {
-            List<Path> allJsonFiles = getFilePaths(outputPath + File.separator + sourcePath, ".json");
+        try (FileOutputStream fos =
+                     new FileOutputStream(outputPath + File.separator + JSON_ZIP_FILE);
+             ZipOutputStream zos =
+                     new ZipOutputStream(fos)
+        ) {
```
eugenepeh: Perhaps something like this, would be more readable.
```
try (
    FileOutputStream fos = new FileOutputStream(outputPath + File.separator + JSON_ZIP_FILE);
    ZipOutputStream zos = new ZipOutputStream(fos)
) {
```

yamidark:
@AdityaA1998  On Windows, seems like if we supply an `out` folder, the `.json` files won't be inside the zip file. Can you investigate this?

AdityaA1998:
@yamidark on it.

yong24s:
@eugenepeh 
Zip works on Ubuntu 18.04. Tested with and without ```-output``` argument.

yamidark:
@AdityaA1998 @eugenepeh Works on Windows as well 👍

src/main/java/reposense/util/FileUtil.java
```diff
@@ -121,8 +139,18 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
 
+    /**
+     * Returns a list of {@code Path} of {@code fileType} contained in the given {@code directoryPath} directory.
+     */
+    private static List<Path> getFilePaths(Path directoryPath, String fileType) throws IOException {
+        List<Path> filePaths;
+        filePaths = Files.walk(directoryPath)
+                .filter(p -> p.toString().endsWith(fileType))
+                .collect(Collectors.toList());
+        return filePaths;
```
yamidark: We can just return this straight?

src/main/java/reposense/RepoSense.java
```diff
@@ -25,6 +26,9 @@ public static void main(String[] args) {
 
             ReportGenerator.generateReposReport(
                     configs, cliArguments.getOutputFilePath().toAbsolutePath().toString());
+
+            FileUtil.zipJson(cliArguments.getOutputFilePath().toAbsolutePath());
+
```
yamidark: extra line break.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -28,6 +32,8 @@
     private static Logger logger = LogsManager.getLogger(FileUtil.class);
 
     private static final String GITHUB_API_DATE_FORMAT = "yyyy-MM-dd";
+    private static final String JSON_ZIP_FILE = "archiveJSON.zip";
+    private static final ByteBuffer buffer = ByteBuffer.allocate(1 << 11); //2KB
```
yamidark: whitespace after `//`

AdityaA1998:
@eugenepeh plz review.

eugenepeh:
@AdityaA1998 apparently you did not follow the previous comment, please check your own work again.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +64,59 @@ public static void deleteDirectory(String root) throws IOException {
...
+        }
+    }
 
-            while (entry != null) {
-                String entryName = entry.getName();
-                Path path = Paths.get(destinationFolder, entryName);
+    /**
+     * Unzips the contents of the {@code zipSourcePath} and stores in the {@code outputPath}.
+     */
+    public static void unzip(String zipSourcePath, String outputPath) {
```
eugenepeh: why does `zip` uses `Path and `unzip` uses `String`?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,44 +64,59 @@ public static void deleteDirectory(String root) throws IOException {
...
-
-        // buffer for read and write data to file
-        byte[] buffer = new byte[2048];
-
-        try {
-            Files.createDirectories(directory);
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the JSON files contained in the {@code sourcePath} and its subdirectories.
+     * Creates the zipped folder in the {@code sourcePath}.
```
eugenepeh: please add a description for the name of the zipped file.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -28,6 +32,9 @@
     private static Logger logger = LogsManager.getLogger(FileUtil.class);
 
     private static final String GITHUB_API_DATE_FORMAT = "yyyy-MM-dd";
+    private static final String TEMPLATE_ZIP_ADDRESS = "/templateZip.zip";
+    private static final String JSON_ZIP_FILE = "archiveJSON.zip";
```
eugenepeh: why is the zip and unzip folder name different?

AdityaA1998: Both the methods `zipJson(..)` and `unzip(..)` are used on different files. 

`zipJson(...)` is used to zip all the generated json files and put them in a `archiveJSON.zip`; however, `unzip(...)` is used to unzip the `TEMPLATE_ZIP_ADDRESS` which contains all the html, css, js files.

eugenepeh: In that case, please add the comments. 
Especially those with specific purpose, used outside java code base.

damithc:
How is this PR going? Is this needed for v1.1?

eugenepeh:
@ongspxm , it seems that #163 does not require this PR. Can you please confirm whether if this is needed?

ongspxm:
Netlify doesnt require this to work, netlify is our server in this case.
But if you clone the  repo, build on your own, and open on your own comp
without starting a server, you will still need the zip file.

My suggestion is we go ahead with the merge, since we have added the steps
to get the dashboard in the user docs already.

eugenepeh:
>Netlify doesnt require this to work, netlify is our server in this case.
But if you clone the  repo, build on your own, and open on your own comp
without starting a server, you will still need the zip file.

I see. Will user need to select the zip folder on the index page or is it automated?

eugenepeh:
> User will need to select manually.

This is not very ideal as it is not standardized with the server version which makes
- netlify preview alone insufficient
- 2 different structure to maintain in the future

@ongspxm if this is the way to go, can you please stick to only one of them? E.g. if its on server, use the zip file directory, else ask the user to choose it manually.

eugenepeh:
```
Means on java side we only generate the zip file, then skip the json?

Means we dun get benefit of partial loading.

Cos on local side, when choosing the zip file, its almost instantaneous.

But on server side, you will have to retrieve the whole zip file in one go.
```

Personally, I think having 2 different standard to maintain is horrendous. It usually introduces more blind spots and helps your product break down more often than expectation.

@damithc what do you think?

damithc:
> @damithc what do you think?

I agree; having two ways to do the same thing is not ideal. I guess we'll have to do for embedded-server approach in future. For the moment, I'm OK with the zip file approach as a temporary measure just so that we can move forward. I'm worried that we cannot get reposense ready for the coming semester. I need to try v1.1 ASAP to decide what more we need to do before start of semester.

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -0,0 +1,34 @@
...
+public class FileUtilTest {
+
+    private static final Path OUTPUT_DIRECTORY_ABSOLUTE = new File(FileUtilTest.class.getClassLoader()
+            .getResource("reposense-report-test").getFile()).toPath().toAbsolutePath();
+    private static final Path ARCHIVE_JSON_PATH = Paths.get(OUTPUT_DIRECTORY_ABSOLUTE.toString(), "archiveJSON.zip");
+
+    @Test
+    public void zipTest_success() {
+        FileUtil.zipJson(OUTPUT_DIRECTORY_ABSOLUTE);
+        Assert.assertTrue(Files.exists(ARCHIVE_JSON_PATH));
```
eugenepeh: Quick remark - 
Is this sufficient to know that your function is working well? What if the result does not meet your expectation?

AdityaA1998: I thought about it too, but to test the content of the zip file we need to unzip it which makes it dependent on unzip(...) so I'm looking for an alternate method.

eugenepeh: alright, @AdityaA1998 please highlight when its ready for next iteration of review.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -121,8 +151,27 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
         }
     }
 
+    /**
+     * Zips all the Json files present in the {@code sourcePath}.
```
yamidark: `inside {@code sourcePath} directory`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,55 +65,77 @@ public static void deleteDirectory(String root) throws IOException {
...
+        } catch (IOException ioe) {
+            logger.log(Level.SEVERE, ioe.getMessage(), ioe);
+        }
+    }
 
-            while (entry != null) {
-                String entryName = entry.getName();
-                Path path = Paths.get(destinationFolder, entryName);
+    /**
+     * Unzips the contents of the {@code zipSourcePath} and stores in the {@code outputPath}.
```
yamidark: `Unzips the contents of the {@code zipSourcePath} into {@code outputPath}.`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,55 +65,77 @@ public static void deleteDirectory(String root) throws IOException {
...
+    public static void zip(Path sourceAndOutputPath, String... fileType) {
+        FileUtil.zip(sourceAndOutputPath, sourceAndOutputPath, fileType);
+    }
 
-        try {
-            Files.createDirectories(directory);
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the {@code fileType} files contained in the {@code sourcePath} and its subdirectories.
+     * Creates the zipped {@code ZIP_FILE} file in the {@code outputPath}.
```
yamidark: remove `{code ZIP_FILE}`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,55 +65,77 @@ public static void deleteDirectory(String root) throws IOException {
...
-    public static void unzip(ZipInputStream zipInput, String destinationFolder) {
-        Path directory = Paths.get(destinationFolder);
-
-        // buffer for read and write data to file
-        byte[] buffer = new byte[2048];
+    /**
+     * Zips all the files of {@code fileType} contained in the {@code sourceAndOutputPath}.
+     * Zipped file is contained in the same location {@code sourceAndOutputPath} as {@code fileTypes}'s location.
+     */
+    public static void zip(Path sourceAndOutputPath, String... fileType) {
```
yamidark: `Zips all the files of {@code fileType} contained inside {@code sourceAndOutputPath} directory into the same folder.`?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,55 +65,77 @@ public static void deleteDirectory(String root) throws IOException {
...
+    }
 
-        try {
-            Files.createDirectories(directory);
-            ZipEntry entry = zipInput.getNextEntry();
+    /**
+     * Zips all the {@code fileType} files contained in the {@code sourcePath} and its subdirectories.
+     * Creates the zipped {@code ZIP_FILE} file in the {@code outputPath}.
+     */
+    public static void zip(Path sourcePath, Path outputPath, String... fileType) {
```
yamidark: `fileTypes` plural

src/main/java/reposense/util/FileUtil.java
```diff
@@ -57,55 +65,77 @@ public static void deleteDirectory(String root) throws IOException {
...
 
     /**
-     * Copies the template files to the {@code outputPath}.
+     * Copies the template files from {@code sourcePath} to the {@code outputPath}.
      */
-    public static void copyTemplate(String outputPath) {
-        InputStream is = RepoSense.class.getResourceAsStream(Constants.TEMPLATE_ZIP_ADDRESS);
-        FileUtil.unzip(new ZipInputStream(is), outputPath);
+    public static void copyTemplate(String sourcePath, String outputPath) {
+        FileUtil.unzip(Paths.get(sourcePath).toAbsolutePath(), Paths.get(outputPath));
```
yamidark: there shouldn't be a need for `toAbsolutePath()`?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -121,8 +151,27 @@ public static void copyDirectoryFiles(Path src, Path dest) throws IOException {
...
+    }
+
+    /**
+     * Returns a list of {@code Path} of {@code fileType} contained in the given {@code directoryPath} directory.
+     */
+    private static List<Path> getFilePaths(Path directoryPath, String... fileType) throws IOException {
+        List<Path> filteredFilePaths = new ArrayList<>();
+        for (String extension : fileType) {
+            filteredFilePaths.addAll(Files.walk(directoryPath)
+                    .filter(p -> p.toString().endsWith(extension))
```
yamidark: Write a helper method for the filtering, so that we can collect all inside the list at once, instead of having to go through each `extension`.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -2,33 +2,41 @@
...
 import reposense.system.LogsManager;
 
 
 public class FileUtil {
     private static Logger logger = LogsManager.getLogger(FileUtil.class);
-
     private static final String GITHUB_API_DATE_FORMAT = "yyyy-MM-dd";
 
+    // zip file which contains all the specified file types
+    private static final String ZIP_FILE = "archive.zip";
```
yamidark: make this `public` and import it for use in your unit test instead of redefining

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+    @Test
+    public void zip_validFileType_success() throws IOException {
+        FileUtil.zip(OUTPUT_DIRECTORY_ABSOLUTE, ".csv", ".zip");
+        Assert.assertTrue(Files.exists(ARCHIVE_ZIP_PATH));
+        Assert.assertTrue(Files.size(ARCHIVE_ZIP_PATH) > 0);
+    }
+
+    @Test
+    public void zip_invalidFileType_fail() {
+        FileUtil.zip(REPORT_DIRECTORY_ABSOLUTE, ".jsonnn");
```
yamidark: this isn't an invalid file type? should instead be a file type that does not exist in directory, thus the zip file should exist but is empty?

AdityaA1998: For an empty zip, `Files.exist(Path ...)` returns false; however, the empty zip does exist.

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -38,14 +38,14 @@ public void zip_invalidLocation_fail() {
...
+        FileUtil.zip(OUTPUT_DIRECTORY_ABSOLUTE, ".csv");
         Assert.assertTrue(Files.exists(ARCHIVE_ZIP_PATH));
         Assert.assertTrue(Files.size(ARCHIVE_ZIP_PATH) > 0);
     }
 
     @Test
     public void zip_invalidFileType_fail() {
-        FileUtil.zip(REPORT_DIRECTORY_ABSOLUTE, ".jsonnn");
+        FileUtil.zip(REPORT_DIRECTORY_ABSOLUTE, ".pdf");
         Assert.assertFalse(Files.exists(ARCHIVE_ZIP_PATH));
```
yamidark: Add a comment that the 'empty' zip will return 'false' for the exist test, even though it gets created.

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+    public void zip_validFileType_success() throws IOException {
+        FileUtil.zip(OUTPUT_DIRECTORY_ABSOLUTE, ".csv");
+        Assert.assertTrue(Files.exists(ARCHIVE_ZIP_PATH));
+        Assert.assertTrue(Files.size(ARCHIVE_ZIP_PATH) > 0);
+    }
+
+    @Test
+    public void zip_invalidFileType_fail() {
+        FileUtil.zip(REPORT_DIRECTORY_ABSOLUTE, ".pdf");
+        Assert.assertFalse(Files.exists(ARCHIVE_ZIP_PATH));
```
yamidark: This test doesn't actually fail, but instead gives an empty `zip` file right? The file type also isn't invalid, just that its non-existent in the folder to zip. Also, add a comment that the 'empty' `zip` file would return `false` for the `Files.exist(Path)` here, since its not intuitive.

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Test
+    public void unzip_validZipFile_success() {
+        FileUtil.unzip(TEMPLATE_ZIP_PATH, TEMPLATE_DIRECTORY_ABSOLUTE);
+        Assert.assertTrue(Files.exists(TEMPLATE_DIRECTORY_ABSOLUTE));
```
yamidark: Can we do something like `verifyAllJson(Path, String)` in `Entry` here, to check that all files have been unzipped?

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Test
+    public void unzip_invalidZipFile_fail() {
+        Path invalidZipFile = Paths.get(OUTPUT_DIRECTORY_ABSOLUTE.toString(), "parser_test.csv");
+        FileUtil.unzip(invalidZipFile, OUTPUT_DIRECTORY_ABSOLUTE);
+        Assert.assertFalse(Files.exists(Paths.get(OUTPUT_DIRECTORY_ABSOLUTE.toString(), "parser_test")));
```
yamidark: We should just assert that the `OUTPUT_DIRECTORY_ABSOLUTE` doesn't exist, since no files are getting unzipped out.

AdityaA1998:
@yamidark @yong24s @ongspxm for review

eugenepeh:
Actually to simplify the test, you can just use the function to zip it once, visually inspect that it is correct, then use it as expected result for comparison.

yong24s:
LTGM. Not sure, but I think you can reduce the size of the test files.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -2,30 +2,39 @@
...
 import reposense.system.LogsManager;
 
 
 public class FileUtil {
-    private static Logger logger = LogsManager.getLogger(FileUtil.class);
 
+    // zip file which contains all the specified file types
+    public static final String ZIP_FILE = "archive.zip";
+
+    private static Logger logger = LogsManager.getLogger(FileUtil.class);
```
yamidark: should be `private static final Logger logger...`

src/main/java/reposense/util/Constants.java
```diff
@@ -2,6 +2,6 @@
 
 public class Constants {
     public static final String REPOS_ADDRESS = "repos";
-    public static final String TEMPLATE_ZIP_ADDRESS = "/templateZip.zip";
+    public static final String STATIC_INDIVIDUAL_REPORT_TEMPLATE_ADDRESS = "repo_report";
```
yamidark: this is unused? remove it

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.nio.file.Paths;
+
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class FileUtilTest {
+
+    private static final Path REPORT_DIRECTORY_ABSOLUTE = new File(FileUtilTest.class.getClassLoader()
+            .getResource("reposense-report-test").getFile()).toPath().toAbsolutePath();
```
eugenepeh: would prefer that if all these files are only used in this test case, put them under a same folder.

eugenepeh:
Also please take note of @yong24s 's suggestion

>LGTM. Not sure, but I think you can reduce the size of the test files.

There isn't a need to use full fledge json files if its just a proof of concept. Especially when the content isn't even tested.

yong24s:
Tested on Linux. It works.

yamidark:
The zip generated doesn't work on Windows, as the zip file created is missing some directory entries.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -55,55 +64,81 @@ public static void deleteDirectory(String root) throws IOException {
...
+                    int length;
+                    while ((length = is.read(buffer.array())) > 0) {
+                        zos.write(buffer.array(), 0, length);
+                    }
+                }
+            }
+        } catch (IOException ioe) {
+            logger.log(Level.SEVERE, ioe.getMessage(), ioe);
+        }
+    }
```
yong24s: Suggested fix to Windows as follows:

```
    public static void zip(Path sourcePath, Path outputPath, String... fileTypes) {
        try (
                FileOutputStream fos = new FileOutputStream(outputPath + File.separator + ZIP_FILE);
                ZipOutputStream zos = new ZipOutputStream(fos)
        ) {
            Set<Path> allFilePaths = getFilePaths(sourcePath, fileTypes);
            for (Path filePath : allFilePaths) {
                String path = sourcePath.relativize(filePath.toAbsolutePath()).toString();
                String zipEntryPath = Files.isDirectory(filePath) ? path + "/" : path;
                zos.putNextEntry(new ZipEntry(zipEntryPath.replace("\\", "/")));

                if (Files.isRegularFile(filePath)) {
                    byte[] data = Files.readAllBytes(filePath);
                    zos.write(data, 0, data.length);
                }

                zos.closeEntry();
            }
        } catch (IOException ioe) {
            logger.log(Level.SEVERE, ioe.getMessage(), ioe);
        }
    }
```

yamidark: Verified on my Windows 10 side as well 👍

yong24s:
For documentation:

For Windows and Mac OSX, the container parent directory has be to created using ```ZipOutputStream#putNextEntry``` for interoperability. 

Additionally, on Windows, all `\` has to be replaced with `/`.

src/main/java/reposense/util/FileUtil.java
yong24s: I don't think we need `InputStream` to read a file in byte as we are dealing with very small files.

yamidark: >I don't think we need InputStream to read the bytes of a file as we are dealing with very small files

Erm, some of the `.json` files can go up to several 100k LOC (our `authorship.json` for `testrepo-charlie` is at 3739KB at the moment, will increase in other larger projects), so we should probably shift back to `InputStream` and read using a buffer.

yong24s: Ah... let's keep the `InputStream` 👍

src/main/java/reposense/util/FileUtil.java
```diff
@@ -82,15 +82,12 @@ public static void zip(Path sourcePath, Path outputPath, String... fileTypes) {
         ) {
             Set<Path> allFiles = getFilePaths(sourcePath, fileTypes);
             for (Path path : allFiles) {
+                String filePath = sourcePath.relativize(path.toAbsolutePath()).toString();
+                String zipEntry = Files.isDirectory(path) ? filePath + File.separator : filePath;
                 try (InputStream is = Files.newInputStream(path)) {
```
yamidark: This will cause `IOException` since we are trying to open an `InputStream` for a directory, thus resulting in an empty zip being generated. You should do the directory check outside and skip this portion.

AdityaA1998: `IOException` will be caused if you try and write something to `ZipOutStream` after putting the entry. I'm checking if its a regular file before to the `ZipOutputStream`.

yong24s: Your implementation will fail on Windows. 

Try:
```
Set<Path> allFilePaths = getFilePaths(sourcePath, fileTypes);
    for (Path filePath : allFilePaths) {
        String path = sourcePath.relativize(filePath.toAbsolutePath()).toString();
        String zipEntryPath = Files.isDirectory(filePath) ? path + "/" : path;
        zos.putNextEntry(new ZipEntry(zipEntryPath.replace("\\", "/")));

        if (Files.isRegularFile(filePath)) {
            try (InputStream is = Files.newInputStream(filePath)) {
                int length;
                while ((length = is.read(buffer.array())) > 0) {
                    zos.write(buffer.array(), 0, length);
                }
           }
       }

     zos.closeEntry();
}
```

src/main/java/reposense/util/FileUtil.java
```diff
@@ -55,55 +64,78 @@ public static void deleteDirectory(String root) throws IOException {
...
+                ZipOutputStream zos = new ZipOutputStream(fos)
+        ) {
+            Set<Path> allFiles = getFilePaths(sourcePath, fileTypes);
+            for (Path path : allFiles) {
+                String filePath = sourcePath.relativize(path.toAbsolutePath()).toString();
+                String zipEntry = Files.isDirectory(path) ? filePath + File.separator : filePath;
+                try (InputStream is = Files.newInputStream(path)) {
+                    zos.putNextEntry(new ZipEntry(zipEntry.replace("\\", "/")));
+                    int length;
+                    while (Files.isRegularFile(path) && (length = is.read(buffer.array())) > 0) {
```
yong24s: This implementation is smelly.

Why open InputStream for both `folder` and `file`?

You can restructure the block by moving `Files.isRegularFile(path)` upwards.

If the reason for `InputStream` is to write to file, then the question is - Do I need to open `InputStream` for folder?

yong24s:
When dealing with `java.io` or `java.nio`, be sure the code is clean to ensure interoperability between OSes.
Otherwise, the code is bound to fail.

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class FileUtilTest {
+
+    private static final Path FILE_UTIL_TEST_DIRECTORY = new File(FileUtilTest.class.getClassLoader()
+            .getResource("FileUtilTest").getFile()).toPath().toAbsolutePath();
+    private static final Path ARCHIVE_ZIP_PATH = Paths.get(FILE_UTIL_TEST_DIRECTORY.toString(),
+            FileUtil.ZIP_FILE);
```
damithc: In cases like this, try to insert the line break at a higher node of the syntax tree, to improve readability. e.g.,
```java
private static final Path ARCHIVE_ZIP_PATH = 
        Paths.get(FILE_UTIL_TEST_DIRECTORY.toString(), FileUtil.ZIP_FILE);
```

PR #158, opened 2018-06-29
Remove navbar

ongspxm:
as recommended. dun see first

```
After the removal of the feature to support multiple reports, the navigation
bar contains only the reposense logo.

The bar now plays only an aesthetic role and has no functional purpose, which
is a waste of vertical space.

Similarly the filters on the summary tab could be more concisely displayed to
fully ultilize vertical space.

Let's set the default path for retreieving report data to `/`, and make the
necessary UI changes to improve vertical space usage.
```


eugenepeh:
@ongspxm you need to make your PR smaller, 2000 LoC is too much for a single PR

ongspxm:
haha, dun view this first... view this after the frontend merged

PR #160, opened 2018-07-02
Dashboard: restore view state using query parameters #224

### LINKED ISSUE #224: Add back the bookmark feature
In v1.0 we had a feature to generate a permalink that users can bookmark to visit a particular view of the dashboard. This feature is very important because tutors should be able to go to a specific filtered view in one step. 

v1.1.1a doesn't seem to have this feature anymore.

Suggested: add the feature back.

ongspxm:
fixes #224 

```
The summary dashboard plays an important role in highlighting and filtering the
report data to provide the most relevant view for the users.

Permalinks was a feature implemented in the previous dashboard that enables the
users to quickly save the state of the dashboard to quickly return to it when
needed.

While useful, this feature was not core to the usage of the dashboard, hence
left out during the feature for feature reimplementation during the dashboard
structural changes.

Lets implement this permalink feature in the main vue module. New fields will
be emitted from their individual modules into the global  object.

The summary module will read the information from the global object then load
the filtered view accordingly.

Implementing it in the main modules also provides us the option of opening up
other modules for the permalink features if needed in the future.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -62,6 +62,11 @@ html
             .mui-textfield.summary-picker__date
               input(type="date", v-model="filterUntilDate")
               label until
+          .mui-select.summary-picker__input
+            select(v-model="filterTimeFrame")
+              option(value="day") Day 
+              option(value="week") Week 
+            label details 
           .mui-checkbox.summary-picker__checkboxes.summary-picker__input
             label
               input(type="checkbox", v-model="filterSortReverse").summary-picker__checkbox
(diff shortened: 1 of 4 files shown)


damithc:
> @damithc i have updated the timeframe thingy to use the dropdown so that we can implement the per commit view in the future, please check netlify to see if its alright.

Looks good. Can you rename it as `granularity`?

eugenepeh:
@ongspxm doesn't seems to work when I change the sort, copy the url and paste again

PR #163, opened 2018-07-02
Dashboard revamp

eugenepeh:
This PR contains the work done by @ongspxm 

Proposed merge message:
```
There are 2 main templates loaded separately to process the information
from the generated report. This requires the same template files to be
regenerated multiple times, which greatly increases the size of the
generated report unnecessarily. Ideally, it should only contains the
data for the report.

Also, as the dashboard uses vue for its templating engine mostly, let's
better utilize the framework to restructure the dashboard into a single
page application which loads content from the report dynamically.
```


ongspxm:
@eugenepeh ready for merge

eugenepeh:
@ongspxm please proposed a merge message for this PR

ongspxm:
@damithc ready for review

**proposed commit message**

```
There are 2 main templates loaded seperately to process the information from the generated report. This requires the same template files to be regenerated multiple times, which greatly increases the size of the generated report unnecessarily, which ideally, should only contains the data for the report.

The current dashboard uses vue for its templating engine mostly. Let's more fully ultilize the vue framework to restructure the dashboard into a SPA which loads the content from the report dynamically.
```

damithc:
`more fully ultilize` -> `better utilize`?
What's `SPA`? Use the full version of the term?

template/static/js/api.js
```diff
@@ -0,0 +1,96 @@
+/* util funcs */
```
damithc: Use full words e.g., functions

ongspxm: so i make a new pr to frontend for the nits?

ongspxm:
**edited proposed message**

```
There are 2 main templates loaded seperately to process the information
from the generated report. This requires the same template files to be
regenerated multiple times, which greatly increases the size of the
generated report unnecessarily, which ideally, should only contains the
data for the report.

The current dashboard uses vue for its templating engine mostly. Let's
more better ultilize the vue framework to restructure the dashboard into
a single page application which loads the content from the report dynamically.
```

template/static/js/api.js
damithc: Nits can be fixed in this PR. No need to get my review again though.

ongspxm:
nits and conflict resolved in #173

eugenepeh:
@ongspxm seems like the additional instructions on `UserGuide` were not removed, please do a proper merge to the master to resolve the conflict.

ongspxm:
roger that, opened a new PR in #176

eugenepeh:
@ongspxm seems like need to merge one more time due to the #175 quick fix

eugenepeh:
Also for the PR title, since there were many changes, I will just simplify it to `Dashboard revamp`.
Do you think that's appropriate? @ongspxm @damithc

damithc:
> I will just simplify it to Dashboard revamp.
> Do you think that's appropriate?

Sure

PR #166, opened 2018-07-04
fixing week view. after #160

ongspxm:
after #160

**proposed commit message**
```
In the summary view, the week view is generated from the information in
the daily view. However when the daily commits do not match up with the
week, (ie, date range doesn't end on a Sunday), the week view
generation messes up.

This happens when the week view generation process tries to access a
commit in an unpadded list of day commits.

Let's add in a check to confirm the availability of the day commit to
ensure smooth running of the function.
```


ongspxm:
tired of waiting, gonna put #166 here as well

ongspxm:
together with #198

Issue #56, opened 2018-04-21
Allow Configurable File types


harryggg:
Currently, allowed file types (.java, .adoc)are hardcoded, we should allow users to configure what file types he wants to allow

Issue #51, opened 2018-04-06
Ramp chart: selection doesn't work after search feature is used


damithc:
Steps:
1. Search for a keyword
2. Clear keyword and press enter -> full list is shown
3. Click on a name

Expected: load the code of that person
Actual: nothing happens

harryggg:
very weird bug

eugenepeh:
Issue doesn't seems to exist anymore

PR #167, opened 2018-07-04
allow configurable file types #56

LiXiaoooowei:
Also fix bug for issue `selection does not work after search feature is used #51`.

Changes:
 - specify the types of document to retrieve from git by adding an extra option `-doctype` with different document types seperated by `/`. For example, `java -jar RepoSense.jar -config [path to config file] -output [path to output directory] -doctype java/adoc` or `gradlew run -Dargs="-config [path to config file] -output [path to output directory] -doctype java/adoc"`. `-doctype` flag is optional, and default to `java/adoc`.
 - add support to filter commits and code details based on the type of document. User can choose to filter based on any combination of document types.


eugenepeh:
Just a heads up - there seems to be many unnecessary changes, so take note of that.
Also, try to keep your PR small.
Reason being that
- it is costly to revert a big PR in case of bug
- it is also costly that in case your work get deprecated by latest api and etc
- lastly, it is hard for reviewers (and yourself) to detect bugs in a large PR
- and many other reasons.

LiXiaoooowei:
Thank you for the feedback!
I have attached a screenshot below of how the filter works.
demo

LiXiaoooowei:
The last 6 commits are failed attempts to fix coding style issues reported by Codacy. Please ignore those commits if you are reviewing this PR.

damithc:
PR description should contain: Fixes #56

damithc:
BTW, the intention of #56 is to be able to instruct RepoSense which file types analyze. Currently it is hard-coded to java and adoc. As such, the file types should be given to RepoSense _before_ it runs, via command line or via the config file. I don't think it is suitable to be configured via the dashboard UI (which can only be done _after_ RepoSense has ran. Perhaps use a command line parameter like this? `-types js,java,adoc,md`? What do others think? @yong24s

damithc:
> BTW, the intention of #56 is to be able to instruct RepoSense which file types analyze. Currently it is hard-coded to java and adoc. As such, the file types should be given to RepoSense before it runs, via command line or via the config file. I don't think it is suitable to be configured via the dashboard UI (which can only be done after RepoSense has ran. Perhaps use a command line parameter like this? -types js,java,adoc,md? What do others think? @yong24s

Let's continue this discussion at #56.

PR #168, opened 2018-07-05
Fix code panel not showing the correct files #214

### LINKED ISSUE #214: Code panel: not showing the correct files
When switching from one to the other code panel's result, the corresponding files was not updated.
E.g. below should show KeyCommand.java but its still the README.adoc from the other author `Ahm (lithiumlkid)`

image

ongspxm:
fixes #214

```
As the tab content is changed, some of the data for the vue module will
be overwritten and it messes up the display of the authorship tab,
causing the case where multiple files gets jumbled up together and the
display goes wild.

To prevent that from happening, let's tag the `v_authorship` module
with a key which forces vue to re-render `v_summary` as a new instance
with a fresh set of data.
```

### DIFF
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -11,11 +11,7 @@ window.app = new window.Vue({
 
     isTabActive: false,
     isTabAuthorship: false,
-    isTabIssues: false,
-
-    tabRepo: '',
-    tabAuthor: '',
-    tabAuthorName: '',
+    tabAuthorship: {},
   },
   methods: {
     // model functions //
@@ -62,14 +58,11 @@ window.app = new window.Vue({
 
     deactivateTabs() {
       this.isTabAuthorship = false;
-      this.isTabIssues = false;
     },
 
     updateTabAuthorship(obj) {
       this.deactivateTabs();
-      this.tabRepo = obj.repo;
-      this.tabAuthor = obj.author;
-      this.tabAuthorName = obj.name;
+      this.tabAuthorship = { ...obj };
 
       this.isTabActive = true;
       this.isTabAuthorship = true;
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -11,6 +11,7 @@ window.app = new window.Vue({
 
     isTabActive: false,
     isTabAuthorship: false,
+    tabInfo: {},
     tabAuthorship: {},
   },
   methods: {
@@ -62,10 +63,14 @@ window.app = new window.Vue({
 
     updateTabAuthorship(obj) {
       this.deactivateTabs();
-      this.tabAuthorship = { ...obj };
+      this.tabInfo.tabAuthorship = { ...obj };
 
       this.isTabActive = true;
       this.isTabAuthorship = true;
+    },
+
+    generateKey(dataObj) {
+      return JSON.stringify(dataObj);
     },
   },
   components: {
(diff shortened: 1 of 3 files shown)


ongspxm:
No problem, will get it done. Your wish is my command.

frontend/src/static/js/main.js
```diff
@@ -62,14 +58,11 @@ window.app = new window.Vue({
...
       this.isTabAuthorship = false;
-      this.isTabIssues = false;
     },
 
     updateTabAuthorship(obj) {
       this.deactivateTabs();
-      this.tabRepo = obj.repo;
-      this.tabAuthor = obj.author;
-      this.tabAuthorName = obj.name;
+      this.tabAuthorship = { ...obj };
```
eugenepeh: if you want to support extensions, perhaps this should be given a generic name instead of `tabAuthorship`?

ongspxm: oh, the idea is to have one object for each module, so instead of `tabAuthor`, `tabAuthorName` and `tabRepo` for the authorship tab, we have just one `tabAuthorship` object that is used to passed all the details into the authorship module.

eugenepeh: but in future, you're going to use that tab to display commits etc as well isn't it? so it wouldn't be right to call it `tabAuthorship` anymore.

ongspxm:
I am planning to create individual objects for different "modules". So for
the tab which shows the list of commits, there be one `tabCommits`.

The reason for this design is to allow for the content of multiple tabs to
be loaded at the same time, eg (commit view and diff view), when there are
modules where it makes sense to include multiple tabs.

ongspxm:
Or i could create a tabInfo, which is a mapping that includes various
objects like so:

```
tabInfo : {
    tabAuthorship: {…},
    tabCommits: {…}
}
```

Would you prefer this design?

eugenepeh:
>Or i could create a tabInfo, which is a mapping that includes various
objects like so:

Hmm, the problem is, your PR title is not matching what you doing.
It seems like what you're trying to do is to loading of several tab at once while your title puts support extension, usually means extend the functionality, which I don't see it in this case. Perhaps you need to update the PR title.

Also by tab, do you mean table? if so, its better to choose another short form because its very misleading especially when your code are meant to work with browser.

ongspxm:
> Also by tab, do you mean table? if so, its better to choose another short
form because its very misleading especially when your code are meant to
work with browser
haha, like literal tabs, like window tabs tabs

> Perhaps you need to update the PR title.
okay, "architecture changes to consolidate tab information"?

eugenepeh:
> okay, "architecture changes to consolidate tab information"?

I wouldn't really call a single variable shrink an architecture change, perhaps you would like to include a bit more work in this PR to produce a bigger impact.

eugenepeh:
> okay, "architecture changes to consolidate tab information"?

On the other hand, perhaps you can just name this title accordingly to #214 
e.g. `Fix code panel not showing the correct files` 
That would make more sense than your proposed title

eugenepeh:
> haha, like literal tabs, like window tabs tabs
```
+                  v-bind:key="tabAuthorship.repo+'_'+tabAuthorship.author"
+                  v-bind:repo="tabAuthorship.repo",
+                  v-bind:author="tabAuthorship.author",
+                  v-bind:name="tabAuthorship.name")
```

It looks plain weird that you're populating the tab with the an object named `tabAuthorship`

ongspxm:
@eugenepeh updated, take a look?

eugenepeh:
As the tab content is changed, some of the data for the vue module will
be overwritten and it messes up the display of the authorship tab,
causing the case where multiple files gets jumbled up together and the
display goes wild.

-> `When switching to display another author's code contribution, some of the code panel's ui elements continue to display old data.`

PR #169, opened 2018-07-05
Fix #51: ramp chart selection does not work after search feature is used

### LINKED ISSUE #51: Ramp chart: selection doesn't work after search feature is used

LiXiaoooowei:
Upon click on ramp chart, the click event fires an ajax call to `result.js` load relevant authorship details. The event binding is lost when DOM is reloaded. Therefore, using jQuery's `on` method when document ready will bind event to `element` other than `event target`. Details see here https://stackoverflow.com/questions/13767919/jquery-event-wont-fire-after-ajax-call

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ repos/
 logs/
 out/
 ft_temp/
+outputs/
--- a/template/index.html
+++ b/template/index.html
@@ -61,38 +61,38 @@
               },
               methods: vueMethods
           });
-          $(".author_name_tag").click(function() {
-              displayName = $(this).attr("displayName");
-              author = $(this).attr("author");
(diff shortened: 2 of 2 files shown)


damithc:
@LiXiaoooowei let others know when this PR is ready for review

LiXiaoooowei:
@damithc @eugenepeh Ready for review

damithc:
Updated PR to match issue title, as per our convention.

LiXiaoooowei:
@damithc Noted. Thank you!

ongspxm:
Sflr. i have tried out the new frontend, doesnt seem to have this issue.

PR #157, opened 2018-06-29
CommitsReporter: add support for per-commit-view

yamidark:
Part of #134, #135 
```
The dashboard only supports daily and weekly interval views for the
commits.

This is due to commits.json file only holding the daily and weekly
interval contributions summary data.

As part of implementing a per-commit-view on the dahsboard, let's
update commits.json to store the list of each CommitResult inside
the daily interval contribution data. This list of CommitResult with the
full commit hash will also allow for tracking of previously analyzed
commits for incremental build.

As the weekly interval view is now generated on the frontend based
on the daily view by #129, there is also no longer a need to store the
weekly interval contribution data in commits.json.

Let's remove the weekly interval contribution data in commits.json
to reduce the size of the json files.
```
Note: since the frontend v2 is still not merged in and this PR is branched of master, 
the dashboard currently cannot be seen.

Issue #134, opened 2018-05-31
Generate weekly contribution dynamically from daily 


eugenepeh:
The start day of the weekly contribution calculation is dependent on the date of first commit in the repository, within the specified period.

This can creates the problem of unequal comparison as the start date of all repositories may differ.

For simplicity, there are 2 solutions to this issue:
1. Fix the start day of calculation
2. Generate it dynamically from daily contributions

The second option would be preferable because:
1. Low cost / Cheap operation
2. Expandable to more options such as `select-able start day`, or `monthly contribution`.

damithc:
In case this is relevant: we also want to do a per-commit view in addition to per-day and per-week views we currently have.

ongspxm:
this is kind of done in #129, as of now, its sitting in the frontend branch

yamidark:
Resolved in #163

PR #174, opened 2018-07-09
[On hold] generating daily view from raw commits. relies on #168, #157

ongspxm:
part of #134, depends on #157, #168

```
Currently, all the commit information is formatted and processed during
the report generation and the dashboard only used for the displaying of
the information.

This means that to support both daily and weekly view, the details for
the different views will all have to be generated and stored as part of
the report. This duplicated information artifically increases the size
of the report as well, without providing much additional information.

Let's shift to generating and storing only the raw commits of the
project, then generate the day view for the summary module from the raw
commit information.
```


damithc:
Does this result in shifting some logic from java to js, or the shift is within js?

ongspxm:
yep, this shifts the consolidating of the daily insertion and deletion from java to js. this is an extension of a comment in #134, where it is suggested to generate the weekly view from the daily view. 

considering we are gonna have per-commit view as our lowest level (smallest) view, it would be natural to base our daily and weekly view off the raw data for consistent results.

damithc:
hmm... I'm not that keen to move critical logic to js, or to move too much work to the browser. Java is a better place to keep logic for ease of testing. Generating files is a one-time cost and the report viewer is not affected. The more we try to generate things on the fly on the browser side, the more it will affect the end-user experience.

yamidark:
@ongspxm Do update your PR to the `per-commit-view` branch.

ongspxm:
closing this due to requirement changes.

PR #159, opened 2018-06-29
Add support for user-supplied file formats #56, #177

yong24s:
Fixes #56, #177

```
File formats .java and .adoc are hardcoded in the git log command of
RepoSense.

As the file formats are not changable, this means RepoSense may only
be useful for users who work on repository that consisting mainly of
java or adoc.

Let's allow an optional file formats to be passed in as a parameter so
that user could use RepoSense on repositories with different codebases.
```

PR #178, opened 2018-07-10
Add more default file types to analyze #177

### LINKED ISSUE #177: Add more default file types to analyze
Current: .java and .adoc files are analyzed by default
Suggested: until we allow files types to be configurable, add these file types as defaults -- js, md, css, html, cs, json, xml, py, fxml, tag, jsp, gradle

yamidark:
Fixes #177
```
RepoSense only tracks '.java' and '.adoc' file types for analyse.

This may not be sufficient for visualization as not many other projects that
are not mainly coded in Java or use asciidoctor for their documents.

Let's include more file types by default in the analysis.
```

### DIFF
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -1,7 +1,6 @@
 package reposense.model;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
@@ -21,9 +20,8 @@ public class RepoConfiguration {
     private Date untilDate;
 
     private transient boolean needCheckStyle = false;
+    private transient List<String> fileFormats;
     private transient int commitNum = 1;
-    private transient List<String> whiteListedFileTypes = Arrays.asList(".java", ".adoc", ".js", ".md", ".css",
-            ".html", ".cs", ".json", ".xml", ".py", ".fxml", ".tag", ".jsp", ".gradle");
     private transient List<String> ignoreDirectoryList = new ArrayList<>();
     private transient List<Author> authorList = new ArrayList<>();
     private transient TreeMap<String, Author> authorAliasMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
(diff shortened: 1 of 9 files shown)


src/main/java/reposense/system/CommandRunner.java
```diff
@@ -151,4 +152,13 @@ private static String getGitDateRangeArgs(Date sinceDate, Date untilDate) {
 
         return gitDateRangeArgs;
     }
+
+    private static String getGitFileTypesArgs(List<String> fileTypeList) {
```
eugenepeh: `convertToGitFileTypesArgs` is probably more appropriate

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -21,6 +22,8 @@
 
     private transient boolean needCheckStyle = false;
     private transient int commitNum = 1;
+    private transient List<String> whiteListedFileTypes = Arrays.asList(".java", ".adoc", ".js", ".md", ".css",
+            ".html", ".cs", ".json", ".xml", ".py", ".fxml", ".tag", ".jsp", ".gradle");
```
eugenepeh: I'm thinking whether it would be better to configure this at the main because once you integrate it with the parser/config, it would be clearer to state it like `load param if exist else load default` rather than overwrite it silently at the backend.

yamidark: @yong24s is already integrating this into parser in #159, will put this in `main` temporary in this PR for v1.1

yong24s:
LGTM

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -7,22 +7,23 @@
...
 
-    public static String gitLog(String root, Date sinceDate, Date untilDate) {
+    public static String gitLog(String root, Date sinceDate, Date untilDate, List<String> fileFormats) {
         Path rootPath = Paths.get(root);
 
         String command = "git log --no-merges ";
         command += getGitDateRangeArgs(sinceDate, untilDate);
-        command += " --pretty=format:\"%h|%aN|%ad|%s\" --date=iso --shortstat -- \"*.java\" -- \"*.adoc\"";
+        command += " --pretty=format:\"%h|%aN|%ad|%s\" --date=iso --shortstat";
+        command += convertToGitFileFormatssArgs(fileFormats);
```
eugenepeh: `convertToGitFileFormatssArgs` -> typo here. I think we can just shorten it to `convertToGitArgs` since fileFormats is already an input param.

eugenepeh:
RepoSense only tracks '.java' and '.adoc' file types for analyse.

This may not be sufficient for ~analysising~ `visualization as not` many other projects that are
~not~ mainly coded in Java or use asciidoctor for their documents.

Let's ~add more default file types to analyse.~ `include more file types by default in the analysis`

yamidark:
@damithc Prof, for your review.

yamidark:
@eugenepeh Ready for merge.

eugenepeh:
to be merged after #163

PR #150, opened 2018-06-22
Modification to accommodate netlify

yong24s:
Propose Message:
```
To validate a PR, a contributor would have to generate the dashboard
on his/her computer.

This results in additional lead-time as the contributor would have to
re-run the entire functional test.

Let's deploy the generated dashboard to netlify to streamline the
process of validation.
```

PR #187, opened 2018-07-13
Fix dashboard padding function

### LINKED ISSUE #186: Frontend should read assets relative to the given url path
For example, when hosting on github pages ```https://yong24s.github.io/dashboard-demo/```, the ```summary.json``` is located at  ```https://yong24s.github.io/dashboard-demo/summary.json```and not ``` https://yong24s.github.io/summary.json```

ongspxm:
also fixes #186 

```
When displaying contributors with no commit history, the dashboard will
crash and stop rendering ramp charts for subsequent repositories.

This is caused by the padding of the ramp charts. The padding is added
to the ramp chart to align the charts across different users and
projects.

The amount of padding required is calculated from the `sinceDate` of
the first and last date entry of the user. This becomes problematic
when the user have no date entries at all, causing the system to run
into problem, and hence, skipping the ramp chart generation for the
particular project.

To fix this, let's alter the padding function to take into the edge
case of a empty date range, and artifically fill up the whole timeline
with generated empty dates.
```

### DIFF
--- a/frontend/src/static/js/api.js
+++ b/frontend/src/static/js/api.js
@@ -5,15 +5,15 @@ window.enquery = (key, val) => `${key}=${encodeURIComponent(val)}`;
 const REPORT_DIR = '.';
 
 // data retrieval functions //
-function loadJSON(file, fn) {
+function loadJSON(fname, fn) {
   const err = () => window.alert('unable to get file');
 
   if (window.REPORT_ZIP) {
-    window.REPORT_ZIP.file(file.slice(2)).async('text')
+    window.REPORT_ZIP.file(fname.slice(2)).async('text')
       .then(txt => fn(JSON.parse(txt)));
   } else {
     const xhr = new XMLHttpRequest();
-    xhr.open('GET', file);
+    xhr.open('GET', fname);
     xhr.onload = function xhrOnload() {
       if (xhr.status === 200) {
         fn(JSON.parse(xhr.responseText));
--- a/frontend/src/static/js/api.js
+++ b/frontend/src/static/js/api.js
@@ -2,14 +2,14 @@
 window.$ = id => document.getElementById(id);
 window.enquery = (key, val) => `${key}=${encodeURIComponent(val)}`;
 
-const REPORT_DIR = '';
+const REPORT_DIR = '.';
 
 // data retrieval functions //
 function loadJSON(file, fn) {
   const err = () => window.alert('unable to get file');
 
   if (window.REPORT_ZIP) {
-    window.REPORT_ZIP.file(file.slice(1)).async('text')
+    window.REPORT_ZIP.file(file.slice(2)).async('text')
       .then(txt => fn(JSON.parse(txt)));
   } else {
     const xhr = new XMLHttpRequest();
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -258,6 +258,10 @@ window.vSummary = {
       const userFirst = user.dailyCommits[0];
       const userLast = user.dailyCommits[user.dailyCommits.length - 1];
 
+      if(!userFirst){
+        return null;
+      }
+
       let sinceDate = this.filterSinceDate;
       if (!sinceDate) {
         ({ sinceDate } = userFirst);
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -258,7 +258,7 @@ window.vSummary = {
       const userFirst = user.dailyCommits[0];
       const userLast = user.dailyCommits[user.dailyCommits.length - 1];
 
-      if(!userFirst){
+      if (!userFirst) {
         return null;
       }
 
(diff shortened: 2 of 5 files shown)


ongspxm:
@AdityaA1998 @yong24s pls check, tanks

yong24s:
Update the propose message and tag the issues.

When displaying contributors with no commit history, the dashboard will
**crash and stop rendering ramp charts for subsequent repositories.**
~~face error in generating some of the ramp charts.~~

ongspxm:
@eugenepeh check pls, thank you, we're out for lunch

frontend/src/static/js/v_summary.js
```diff
@@ -271,7 +271,8 @@ window.vSummary = {
       if (this.filterGroupWeek) {
         sinceDate = dateRounding(sinceDate, 1);
       }
-      let diff = getIntervalDay(userFirst.sinceDate, sinceDate);
+      let diff = userFirst ?
```
eugenepeh: > const userFirst = user.dailyCommits[0];

In the event that user.dailyCommits is an empty array, wouldn't it cause an invalid index error?

ongspxm: javascript behavior is to return `undefined`

ongspxm: oh, actually on second thoughts, i might rewrite the fix to play around this idea of an invalid index.

yong24s:
LGTM.

eugenepeh:
@yong24s is this a current problem in the https://github.com/reposense/dashboard-demo/pull/3?

yong24s:
I have applied the fixes directly. Dashboard is running fine for us.

yong24s:
@eugenepeh Sorry about that. I forgot to apply the newer fixes.

@ongspxm Any repos that have an empty author contribution should be fine right? Try **CS2103JAN2018-F14-B1/main** ?

ongspxm:
affirmative
image

yong24s:
```
CS2103JAN2018-F14-B1,main,master,lithiumlkid,AHMAD...HAZAT,
CS2103JAN2018-F14-B1,main,master,codeeong,CODEE...EN CI,
CS2103JAN2018-F14-B1,main,master,jordancjq,JORDA...IA QI,
CS2103JAN2018-F14-B1,main,master,lohtianwei,LOH T...N WEI,
```

ongspxm:
to clarify, netlify runs sample.csv right?

yong24s:
Yes.

From #150:
```mkdir output && ./gradlew clean build && ./gradlew run -Dargs="-config sample.csv -output output"```

**Comment by eugenepeh** [comment]
hmm, from the repo, it seems that [codeeong](https://github.com/CS2103JAN2018-F14-B1/main/commits?author=codeeong) do have some work though. I think we need to further investigate this.

Meanwhile,

File: template/static/js/v_summary.js [hunk 2]
```diff
@@ -258,6 +258,10 @@ window.vSummary = {
       const userFirst = user.dailyCommits[0];
       const userLast = user.dailyCommits[user.dailyCommits.length - 1];
 
+      if(!userFirst){
```
eugenepeh: cosmetic tweak, space before `{`

sample.csv
```diff
@@ -7,3 +7,7 @@ CS2103AUG2017-W09-B2,main,master,charlesgoh,Cha,Charles Goh
 CS2103AUG2017-W09-B2,main,master,Esilocke,Esi,
 CS2103AUG2017-W09-B2,main,master,jeffreygohkw,Jef,Jeffrey Goh
 CS2103AUG2017-W09-B2,main,master,wangyiming1019,Wan,ACER\kyle;Wang Yiming
+CS2103JAN2018-F14-B1,main,master,lithiumlkid,AHMAD...HAZAT,
+CS2103JAN2018-F14-B1,main,master,codeeong,CODEE...EN CI,
+CS2103JAN2018-F14-B1,main,master,jordancjq,JORDA...IA QI,
+CS2103JAN2018-F14-B1,main,master,lohtianwei,LOH T...N WEI,
```
eugenepeh: similar to the other repos, please shorten their display name

frontend/src/static/js/api.js
```diff
@@ -2,14 +2,14 @@
...
-const REPORT_DIR = '';
+const REPORT_DIR = '.';
 
 // data retrieval functions //
 function loadJSON(file, fn) {
   const err = () => window.alert('unable to get file');
 
   if (window.REPORT_ZIP) {
-    window.REPORT_ZIP.file(file.slice(1)).async('text')
+    window.REPORT_ZIP.file(file.slice(2)).async('text')
```
eugenepeh: is it safe to do slicing on zip file?

eugenepeh:
Also, just to highlight something, 
Note that the newly added repo belongs to a later semester.

It seems that the date range is not updated

image

and the latest commit for the repo ends on 05 apr 2018

image

frontend/src/static/js/api.js
ongspxm: file is a string, i will rename it to `fname` for clarity

ongspxm:
> It seems that the date range is not updated

the default date may be an issue for too big a range, but this will be settled together with some ui bugs and fixes in later pr

Issue #135, opened 2018-06-01
Enable incremental report build


eugenepeh:
Generate the entire report takes time, hence we should reduce the time of sub-sequence build by doing it incrementally. 

 Current idea to implement this:
- Do pre-processing to detect changes from the HEAD of last analysis (analysis->HEAD).
- Get list of file changes since the last analysis
  This can be done using `git diff --name-only <hash1> <hash2>`
- CommitsReporter:
  Count daily contribution and sum to existing
- AuthorshipReporter:
  Git blame only on the affected files

Some scenarios that we may have to skip pre-processing and regenerate the entire report
- If `analysis->HEAD` is not found
- If date param is set
- If previous report doesn't exist

fzdy1914:
The issue mentioned is not a problem now, as we auto set the date range now.

eugenepeh:
@fzdy1914 actually the idea of this is to avoid regenerating whole report from scratch in some sort of "append" or modify affected style in a continuous/incremental style.

E.g. 
RepoSense is scheduled to perform analysis daily, however if a target repo is still the same as the day before, we will skip performing analysis on it. 

This is meant to reduce unnecessary analysis.

fzdy1914:
OK, got the idea.

damithc:
Let's put this on hold as keeping data from the previous runs is not feasible when the report generation is done in a CI environment (e.g., on GitHub Actions), as is the case for all our RepoSense reports that are online at the moment.

PR #191, opened 2018-07-13
AuthorshipReporter: add support for incremental authorship build #135

yamidark:
Part of #135


src/test/java/reposense/util/TestConstants.java
```diff
@@ -5,12 +5,7 @@
 public class TestConstants {
     public static final String TEST_ORG = "reposense";
     public static final String TEST_REPO = "testrepo-Alpha";
+    public static final String TEST_MASTER_BRANCH = "master";
```
eugenepeh: can we shift this to git checkout? also, make a default method that checkouts to master?

yamidark:
>seems like there are some clean up involved. Can we split the PR up to make it clearer to objective?

Noted, will split into cleanup and the actual adding of incremental.

src/test/java/reposense/util/TestConstants.java
yamidark: Actually, I think we can should use `config.getBranch()` instead, so no need for this 'constant'. Will remove it.

yamidark:
@eugenepeh Split the cleanup to a different PR (#193) to `incremental branch` as well. 
This PR is dependent on that PR.

yamidark:
@ongspxm @yong24s @AdityaA1998 for your reviews

PR #195, opened 2018-07-16
remove template from source code

ongspxm:
as refered to in #171 

```
Since the dashboard revamp in v1.1, `template` no longer contains the source of the dashboard. Instead, the source files are contained in `frontend/src`, andcompiled into the corresponding `html` and `css` files.

As good practice, these generated template files shouldn't be track. Let's include the generation of all these template files in the gradle build process instead.
```


eugenepeh:
https://plugins.gradle.org/search?term=sass
https://plugins.gradle.org/search?term=npm

ongspxm:
@eugenepeh @yong24s @yamidark @AdityaA1998 please review, push this in first before the other ui changes

damithc:
Shouldn't this PR be in v1.11 as well (because a PR depending on this one is already in v1.11)?

ongspxm:
@damithc ready for review

PR #196, opened 2018-07-16
Fix invalid path when accessing jar resources  #194

### LINKED ISSUE #194: Invalid path when accessing jar resources
Following #152, executing RepoSense.jar causes unzip function to create an invalid path when accessing resources. Let's revert unzip function to access template resource directly using InputStream instead.

```
java -jar RepoSense.jar -config sample.csv
11:28:03 - Path = file:\C:\Users\Mickey\Desktop\GitHub\RepoSense\RepoSense.jar!\templateZip.zip C:\Users\Mickey\Desktop\GitHub\RepoSense\.\reposense-report
Exception in thread "main" java.nio.file.InvalidPathException: Illegal char <:> at index 4: file:\C:\Users\Mickey\Desktop\GitHub\RepoSense\RepoSense.jar!\templateZip.zip
        at java.base/sun.nio.fs.WindowsPathParser.normalize(Unknown Source)
        at java.base/sun.nio.fs.WindowsPathParser.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsPathParser.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsPath.parse(Unknown Source)
        at java.base/sun.nio.fs.WindowsFileSystem.getPath(Unknown Source)
        at java.base/java.nio.file.Paths.get(Unknown Source)
        at reposense.util.FileUtil.copyTemplate(FileUtil.java:142)
        at reposense.report.ReportGenerator.generateReposReport(ReportGenerator.java:35)
        at reposense.RepoSense.main(RepoSense.java:32)

```

yong24s:
Fixes #194 

```
RepoSense 1.1RC2 executed as a JAR throws an InvalidPathException when
accessing templateZip.zip.

This is because the location of the templateZip.zip cannot be located
by its path on the disk but by its relative path in the JAR file. 

So let's change the way to retrieve templateZip.zip from 
RepoSense.class.getClassLoader().getResource("templateZip.zip") to 
RepoSense.class.getResourceAsStream("/templateZip.zip").
```

### DIFF
--- a/src/main/java/reposense/util/FileUtil.java
+++ b/src/main/java/reposense/util/FileUtil.java
@@ -102,22 +102,21 @@ public class FileUtil {
 
     /**
      * Unzips the contents of the {@code zipSourcePath} into {@code outputPath}.
+     * @throws IOException if {@code zipSourcePath} is an invalid path.
      */
-    public static void unzip(Path zipSourcePath, Path outputPath) {
-        try (
-                InputStream is = Files.newInputStream(zipSourcePath);
-        ) {
+    public static void unzip(Path zipSourcePath, Path outputPath) throws IOException {
+        try (InputStream is = Files.newInputStream(zipSourcePath)) {
             unzip(is, outputPath);
-        } catch (IOException e) {
-            logger.log(Level.SEVERE, e.getMessage(), e);
         }
     }
 
-    public static void unzip(InputStream is, Path outputPath) {
-        ZipEntry entry;
-        try (
-                ZipInputStream zis = new ZipInputStream(is)
-        ) {
+    /**
+     * Unzips the contents of the {@code is} into {@code outputPath}.
+     * @throws IOException if {@code is} refers to an invalid path.
+     */
+    public static void unzip(InputStream is, Path outputPath) throws IOException {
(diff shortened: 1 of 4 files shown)


src/main/java/reposense/util/FileUtil.java
```diff
@@ -104,9 +104,18 @@ public static void zip(Path sourcePath, Path outputPath, String... fileTypes) {
      * Unzips the contents of the {@code zipSourcePath} into {@code outputPath}.
      */
     public static void unzip(Path zipSourcePath, Path outputPath) {
```
eugenepeh: why do you need this method

src/main/java/reposense/util/FileUtil.java
```diff
@@ -104,9 +104,18 @@ public static void zip(Path sourcePath, Path outputPath, String... fileTypes) {
...
-        ZipEntry entry;
         try (
```
AdityaA1998: no need for line break when only 1 statement is within the try-with
Make it something like:
```
try(InputStream is = Files.newInputStream(zipSourcePath)) {
} 
```

PR #198, opened 2018-07-16
Fix ramp chart positioning #197

### LINKED ISSUE #197: Ramp chart display does not correspond to date range
Date range:

image

Ramp chart:
image

The date range ends on 14/04/2018 but the last ramp in the ramp chart, which is dated on 13/04/2018, is not positioned at the end of the ramp chart.

### LINKED ISSUE #201: Ramp chart does not display information inclusive of the since and until date
image

Date range : `since` : 04/10/2017 & `until` : 11/10/2017

Commits were made on both - `since` and `until` date; however, the ramp chart excludes information about both the days. 

It is more intuitive to display information about both `since` and `until` dates as well.

ongspxm:
fixes #197, fixes #201 

```
When the dashboard loads two projects of different date range, some of
the dashboard attributes will be messed up and this affects the
rendering of the ramp charts.

More specifically, the `sliceWidth` and `sliceCount` will be affected
by the date range and will cause the dashboard to incorrectly "thinks"
that it has more slices then it actually does, and hence causing it to
be incorrectly displayed squashed up together.

Let's remove the attributes in `v_summary` and generate the width and
position of the slices from the actual array of slices instead.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -94,8 +94,8 @@ html
                   v-bind:href="getSliceLink(user, slice)", target="_blank",
                   v-bind:class="'summary-chart__ramp__slice--color'+j%5",
                   v-bind:style="{\
-                    borderLeftWidth: getWidth(slice)+'em',\
-                    right: (sliceCount-j-1)*sliceWidth + '%'\
+                    borderLeftWidth: getWidth(slice) + 'em',\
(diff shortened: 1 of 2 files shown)


AdityaA1998:
Maybe you can try and fix #201 in this pr as well.

ongspxm:
tired of waiting, gonna put #166 here as well

ongspxm:
@damithc ready for review

PR #205, opened 2018-07-17
Authorship: incorrect line authorship when analyzing date range #203

### LINKED ISSUE #203: Authorship: incorrect line authorship when analyzing date range
```
When a date range is supplied, the FileInfoAnalyzer will only
git blame on the range of commits that fall within the given
date range.

This may cause the authorship of certain lines that have not
been touched at all during the date range to be incorrectly
authored to a random author that commited in that date range.

Let's update the git blame command to run the git blame from
initial to the latest commit of that date range. In addition, let's
also update FileInfoAnalyzer to only analyze files that have been
modified during that commit range.
```

yamidark:
Fixes #203 
Depends on #212 
```
When a date range is supplied, the FileInfoAnalyzer will only
git blame on the range of commits that fall within the given
date range.

This may cause the authorship of certain lines that have not
been touched at all during the date range to be incorrectly
authored to a random author that commited in that date range.

Let's update the git blame command to run the git blame from
initial to the latest commit of that date range. In addition, let's
also update FileInfoExtractor to only track lines that were added
within the date range, and add them into the final authorship count.
```

Below was edited by @eugenepeh, for reference purpose
```diff
diff --git a/docs/UserGuide.md b/docs/UserGuide.md
index 6f0b634..a8af6df 100644
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -18 +18 @@
-Usage: `java -jar RepoSense.jar -config CSV_CONFIG_FILE_PATH [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY]`
+Usage: `java -jar RepoSense.jar -config CSV_CONFIG_FILE_PATH [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...]`
@@ -25 +25 @@ Sample usage:
-$ java -jar RepoSense.jar -config CSV_path.csv -output output_path/ -since 01/10/2017 -until 01/11/2017
+$ java -jar RepoSense.jar -config CSV_path.csv -output output_path/ -since 01/10/2017 -until 01/11/2017 -formats java adoc js
@@ -31,2 +31,3 @@ Argument List:
-- since : Optional. start date of analysis. Format: `DD/MM/YYYY`
-- until : Optional. end date of analysis. Format: `DD/MM/YYYY`
+- since : Optional. The start date of analysis. Format: `DD/MM/YYYY`
+- until : Optional. The end date of analysis. Format: `DD/MM/YYYY`
+- formats : Optional. The file formats to analyse. Formats: `alphanumerical file formats`. If not provided, the following file formats will be used. `adoc, cs, css, fxml, gradle, html, java, js, json, jsp, md, py, tag, xml`
```

### DIFF
--- a/src/main/java/reposense/authorship/FileInfoExtractor.java
+++ b/src/main/java/reposense/authorship/FileInfoExtractor.java
@@ -199,7 +199,6 @@ public class FileInfoExtractor {
         Matcher filePathMatcher = FILE_PATH_PATTERN.matcher(fileDiffResult);
 
         if (!filePathMatcher.find()) {
-            logger.severe(fileDiffResult);
             throw new AssertionError("Should not have error matching file path pattern inside file diff result!");
         }
 
--- a/src/main/java/reposense/authorship/FileInfoExtractor.java
+++ b/src/main/java/reposense/authorship/FileInfoExtractor.java
@@ -34,13 +34,14 @@ public class FileInfoExtractor {
     private static final String FILE_DELETED_METADATA = "deleted file mode 100644\n";
     private static final String LINE_CHUNKS_SEPARATOR = "\n@@ ";
     private static final String LINE_INSERTED_SYMBOL = "+";
+    private static final String STARTING_LINE_NUMBER_GROUP_NAME = "startingLineNumber";
+    private static final String FILE_CHANGED_GROUP_NAME = "filePath";
 
-    private static final int CHUNK_HEADER_LINE_INDEX = 0;
-    private static final int CHUNK_STARTING_LINE_GROUP_INDEX = 2;
-    private static final int FILE_PATH_GROUP_INDEX = 2;
+    private static final int LINE_CHANGED_HEADER_INDEX = 0;
 
-    private static final Pattern CHUNK_HEADER_PATTERN = Pattern.compile("-(\\d+)[,\\d]* \\+(\\d+).*");
-    private static final Pattern FILE_PATH_PATTERN = Pattern.compile("\n(\\+\\+\\+ b/)(.*)\n");
+    private static final Pattern STARTING_LINE_NUMBER_PATTERN = Pattern.compile(
+            "-(\\d)+(,)?(\\d)* \\+(?<startingLineNumber>\\d+)(,)?(\\d)* @@");
+    private static final Pattern FILE_CHANGED_PATTERN = Pattern.compile("\n(\\+){3} b/(?<filePath>.*)\n");
 
     /**
      * Extracts a list of relevant files given in {@code config}.
@@ -73,14 +74,14 @@ public class FileInfoExtractor {
     public static List<FileInfo> getEditedFileInfos(RepoConfiguration config, String lastCommitHash) {
         List<FileInfo> fileInfos = new ArrayList<>();
 
-        String diffResult = CommandRunner.diffCommit(config.getRepoRoot(), lastCommitHash);
+        String fullDiffResult = CommandRunner.diffCommit(config.getRepoRoot(), lastCommitHash);
 
         // no diff between the 2 commits, return an empty list
-        if (diffResult.isEmpty()) {
+        if (fullDiffResult.isEmpty()) {
             return fileInfos;
         }
 
-        String[] fileDiffResultList = diffResult.split(DIFF_FILE_CHUNK_SEPARATOR);
+        String[] fileDiffResultList = fullDiffResult.split(DIFF_FILE_CHUNK_SEPARATOR);
 
         for (String fileDiffResult : fileDiffResultList) {
             // file deleted or is a binary file, skip it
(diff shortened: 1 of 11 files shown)


yamidark:
@yong24s @AdityaA1998 @ongspxm @eugenepeh For your reviews

yamidark:
@eugenepeh Do mark this for v.11 release as well

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -35,19 +52,101 @@
...
-        ArrayList<FileInfo> fileInfos = new ArrayList<>();
-        getAllFileInfo(config, Paths.get(config.getRepoRoot()), fileInfos);
+        List<FileInfo> fileInfos = new ArrayList<>();
+        String lastCommitHash = CommandRunner.getCommitHashBeforeDate(
+                config.getRepoRoot(), config.getBranch(), config.getSinceDate());
+
+        if (!lastCommitHash.isEmpty()) {
+            fileInfos = getEditedFileInfos(config, lastCommitHash);
+        } else {
+            getAllFileInfo(config, Paths.get(config.getRepoRoot()), fileInfos);
```
eugenepeh: Seems really weird that both `getAllFileInfo` and `getEditedFileInfos` are getters, but one has return type and the other don't.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -35,19 +52,101 @@
...
+        // no diff between the 2 commits, return an empty list
+        if (diffResult.isEmpty()) {
+            return fileInfos;
+        }
+
+        String[] fileDiffResultList = diffResult.split(DIFF_FILE_CHUNK_SEPARATOR);
+
+        for (String fileDiffResult : fileDiffResultList) {
+            String[] fileDiffResultLines = fileDiffResult.split("\n");
+            String[] filesCompared = fileDiffResultLines[COMPARED_FILE_LINE_INDEX].split(COMPARED_FILE_PATH_SEPARATOR);
```
eugenepeh: don't really get this part. how do you get `files` from `file`?

yamidark:
>If the PR is incorrect line authorship when analyzing date range why are noDateRange test expected results affected?

Since we removed the `copy and move detection`, the result for `noDateRange` got affected as well.

src/main/java/reposense/authorship/FileInfoExtractor.java
yamidark: `getAllFileInfo` is a recursive function, which makes it slightly harder to have a return type (would have to keep appending all recursive called results).
Perhaps I should just make `getEditedFileInfos` not have a return type as well (pass in the `fileInfos` to be appended)?

src/main/java/reposense/authorship/FileInfoExtractor.java
yamidark: Hmm, the `files` here is for the 'before' and 'after' for a single `file` change, since the file from 'before' may have been renamed into the 'after'.
Maybe I can just mix these with the below line, since we are only interested in the 'after' file name?

eugenepeh: Also what you want is the first line here... what if the original directory contains `b/`? E.g. `blob/`

```
diff --git a/docs/UserGuide.md b/docs/UserGuide.md
index 6f0b634..a8af6df 100644
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -18 +18 @@
-Usage: `java -jar RepoSense.jar -config CSV_CONFIG_FILE_PATH [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY]`
+Usage: `java -jar RepoSense.jar -config CSV_CONFIG_FILE_PATH [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...]`
@@ -25 +25 @@ Sample usage:
-$ java -jar RepoSense.jar -config CSV_path.csv -output output_path/ -since 01/10/2017 -until 01/11/2017
+$ java -jar RepoSense.jar -config CSV_path.csv -output output_path/ -since 01/10/2017 -until 01/11/2017 -formats java adoc js
@@ -31,2 +31,3 @@ Argument List:
-- since : Optional. start date of analysis. Format: `DD/MM/YYYY`
-- until : Optional. end date of analysis. Format: `DD/MM/YYYY`
+- since : Optional. The start date of analysis. Format: `DD/MM/YYYY`
+- until : Optional. The end date of analysis. Format: `DD/MM/YYYY`
+- formats : Optional. The file formats to analyse. Formats: `alphanumerical file formats`. If not provided, the following file formats will be used. `adoc, cs, css, fxml, gradle, html, java, js, json, jsp, md, py, tag, xml`

```

eugenepeh: Perhaps use regex matcher and capturing group instead to make it cleaner? Also, it seems that there are some left over texts after `DIFF_FILE_CHUNK_SEPARATOR` which cause the following method to skip the first index.

src/main/java/reposense/authorship/FileInfoExtractor.java
eugenepeh: nvm, lets leave it for now

eugenepeh:
> Since we removed the copy and move detection, the result for noDateRange got affected as well.

Can we have that change in a separate PR? As a practice, its generally not good to have 2 test breaking changes in a single PR.

yamidark:
>Can we have that change in a separate PR? As a practice, its generally not good to have 2 test breaking changes in a single PR.

Noted, opened #212 for that.

yamidark:
@eugenepeh For your review again.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -35,19 +51,93 @@
...
+    /**
+     * Analyzes the {@code fileDiffResult} and marks each {@code LineInfo} in {@code FileInfo} on whether they were
+     * inserted in between the commit range.
+     */
+    private static void setLinesToTrack(FileInfo fileInfo, String fileDiffResult) {
+        String[] lineChangedChunks = fileDiffResult.split(LINE_CHUNKS_SEPARATOR);
+        List<LineInfo> lineInfos = fileInfo.getLines();
+        int fileLinePointer = 0;
+
+        for (int chunkIndex = 1; chunkIndex < lineChangedChunks.length; chunkIndex++) {
```
eugenepeh: maybe can add a comment, e.g. // skip the header, index starts from 1

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -26,6 +29,19 @@
...
+    private static final String BINARY_FILE_SYMBOL = "\nBinary files ";
+    private static final String FILE_DELETED_METADATA = "deleted file mode 100644\n";
+    private static final String LINE_CHUNKS_SEPARATOR = "\n@@ ";
+    private static final String LINE_INSERTED_SYMBOL = "+";
+
+    private static final int CHUNK_HEADER_LINE_INDEX = 0;
+    private static final int CHUNK_STARTING_LINE_GROUP_INDEX = 2;
+    private static final int FILE_PATH_GROUP_INDEX = 2;
+
+    private static final Pattern CHUNK_HEADER_PATTERN = Pattern.compile("-(\\d+)[,\\d]* \\+(\\d+).*");
```
eugenepeh: instead of chunk, would it be better to name it as `LINES_CHANGED_HEADER`?

yamidark:
>Also for the regex, I believe it can be done more elegantly using repeated patterns and name capturing group.

Whoops, forgot about those :sweat_smile: 
So something like this?
```Java
    private static final Pattern LINES_CHANGED_PATTERN = Pattern.compile(
            "-(\\d)+(,)?(\\d)* \\+(?<lineNumber>\\d+)(,)?(\\d)* @@");
    private static final Pattern FILE_CHANGED_PATTERN = Pattern.compile("\n(\\+){3} b/(?<filePath>.*)\n");
```

yamidark:
@eugenepeh For your review again.

Issue #206, opened 2018-07-19
Same Repository different branch overriding each other


yamidark:
In the `.csv` config file, we may supply the same organization and repository,
but different branches for different authors.

In `CsvParser`, these will be separated these into two different `RepoConfiguration`s
,and analysis will be run separately, generating separate reports.

However, as we set the display name for each `RepoConfiguration` to only be the 
'organization' and 'repository', these different `RepoConfiguration`s will override each
other's report file as they are being generated into the same folder.

yamidark:
[teammates.txt](https://github.com/reposense/RepoSense/files/2208142/teammates.txt)
The above file (converted to `.txt`) has 1 author `dalessr` in the `recycling-bin` branch, and 6 other authors in the `master` branch.

However, the report generated from this config file (as shown below) only shows the stats for the 6 authors in `master` branch, since the `recycling-bin` branch report has been overridden.
![screenshot from 2018-07-19 11-52-15](https://user-images.githubusercontent.com/18352498/42920628-3f03cd62-8b4a-11e8-9864-471762445f16.png)

yamidark:
@eugenepeh @damithc  For your acknowledgement.

damithc:
> @damithc For your acknowledgement.

ACK. I didn't notice it before. Good catch. Yes, let's go for a simple solution now (even if not efficient)

yamidark:
Closed via #207

PR #210, opened 2018-07-19
Allow location to be supplied into CSV Config File

yong24s:
```
Representing a repository with organization and repository name in 
the config csv file may only satisfies basic usage.

By generalizing it to location allows cloning from disk locations or using
of more complex .git links.
i.e. https://username@github.com/organization/repository.git. 

Let's use location in the config csv file to make git clone more versatile.
```

### DIFF
--- a/sample.csv
+++ b/sample.csv
@@ -1,9 +1,9 @@
 Repository's Location,Branch,Author's GitHub ID,Author's Display Name,Author's Git Author Name
-https://github.com/testrepo-Beta/main.git,master,nbriannl,Nbr,
-https://github.com/testrepo-Beta/main.git,master,zacharytang,Zac,Zachary Tang
-https://github.com/testrepo-Beta/main.git,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting
-https://github.com/testrepo-Beta/main.git,master,CindyTsai1,Cin,YuHsuan
-https://github.com/testrepo-Delta/main.git,master,lithiumlkid,Ahm,
-https://github.com/testrepo-Delta/main.git,master,codeeong,Cod,
-https://github.com/testrepo-Delta/main.git,master,jordancjq,Jor,
-https://github.com/testrepo-Delta/main.git,master,lohtianwei,Loh,
+https://github.com/reposense/testrepo-Beta.git,master,nbriannl,Nbr,
+https://github.com/reposense/testrepo-Beta.git,master,zacharytang,Zac,Zachary Tang
+https://github.com/reposense/testrepo-Beta.git,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting
+https://github.com/reposense/testrepo-Beta.git,master,CindyTsai1,Cin,YuHsuan
+https://github.com/reposense/testrepo-Delta.git,master,lithiumlkid,Ahm,
+https://github.com/reposense/testrepo-Delta.git,master,codeeong,Cod,
+https://github.com/reposense/testrepo-Delta.git,master,jordancjq,Jor,
+https://github.com/reposense/testrepo-Delta.git,master,lohtianwei,Loh,
--- a/sample.csv
+++ b/sample.csv
@@ -1,9 +1,9 @@
 Repository's Location,Branch,Author's GitHub ID,Author's Display Name,Author's Git Author Name
-https://github.com/CS2103AUG2017-W09-B1/main.git,master,nbriannl,Nbr,
-https://github.com/CS2103AUG2017-W09-B1/main.git,master,zacharytang,Zac,Zachary Tang
-https://github.com/CS2103AUG2017-W09-B1/main.git,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting
-https://github.com/CS2103AUG2017-W09-B1/main.git,master,CindyTsai1,Cin,YuHsuan
-https://github.com/CS2103JAN2018-F14-B1/main.git,master,lithiumlkid,Ahm,
-https://github.com/CS2103JAN2018-F14-B1/main.git,master,codeeong,Cod,
-https://github.com/CS2103JAN2018-F14-B1/main.git,master,jordancjq,Jor,
-https://github.com/CS2103JAN2018-F14-B1/main.git,master,lohtianwei,Loh,
+https://github.com/testrepo-Beta/main.git,master,nbriannl,Nbr,
+https://github.com/testrepo-Beta/main.git,master,zacharytang,Zac,Zachary Tang
+https://github.com/testrepo-Beta/main.git,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting
+https://github.com/testrepo-Beta/main.git,master,CindyTsai1,Cin,YuHsuan
+https://github.com/testrepo-Delta/main.git,master,lithiumlkid,Ahm,
+https://github.com/testrepo-Delta/main.git,master,codeeong,Cod,
+https://github.com/testrepo-Delta/main.git,master,jordancjq,Jor,
+https://github.com/testrepo-Delta/main.git,master,lohtianwei,Loh,
--- a/sample.csv
+++ b/sample.csv
@@ -1,4 +1,4 @@
-Repo location,branch,Author's GitHub ID,Author's Display Name,Author's Git Author Name
+Repository's Location,Branch,Author's GitHub ID,Author's Display Name,Author's Git Author Name
 https://github.com/CS2103AUG2017-W09-B1/main.git,master,nbriannl,Nbr,
 https://github.com/CS2103AUG2017-W09-B1/main.git,master,zacharytang,Zac,Zachary Tang
 https://github.com/CS2103AUG2017-W09-B1/main.git,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting
--- a/sample.csv
+++ b/sample.csv
@@ -1,9 +1,9 @@
-Organization,Repository,branch,Author's GitHub ID,Author's Display Name,Author's Git Author Name
-CS2103AUG2017-W09-B1,main,master,nbriannl,Nbr,
-CS2103AUG2017-W09-B1,main,master,zacharytang,Zac,Zachary Tang
-CS2103AUG2017-W09-B1,main,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting
-CS2103AUG2017-W09-B1,main,master,CindyTsai1,Cin,YuHsuan
-CS2103JAN2018-F14-B1,main,master,lithiumlkid,Ahm,
-CS2103JAN2018-F14-B1,main,master,codeeong,Cod,
-CS2103JAN2018-F14-B1,main,master,jordancjq,Jor,
-CS2103JAN2018-F14-B1,main,master,lohtianwei,Loh,
+Repo location,branch,Author's GitHub ID,Author's Display Name,Author's Git Author Name
+https://github.com/CS2103AUG2017-W09-B1/main.git,master,nbriannl,Nbr,
+https://github.com/CS2103AUG2017-W09-B1/main.git,master,zacharytang,Zac,Zachary Tang
+https://github.com/CS2103AUG2017-W09-B1/main.git,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting
+https://github.com/CS2103AUG2017-W09-B1/main.git,master,CindyTsai1,Cin,YuHsuan
+https://github.com/CS2103JAN2018-F14-B1/main.git,master,lithiumlkid,Ahm,
+https://github.com/CS2103JAN2018-F14-B1/main.git,master,codeeong,Cod,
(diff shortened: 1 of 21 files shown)


yong24s:
@eugenepeh @yamidark 

Ready for review, please.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -66,7 +82,7 @@ public boolean equals(Object other) {
 
     @Override
     public int hashCode() {
-        return Objects.hash(this.organization, this.repoName, this.branch);
+        return Objects.hash(location);
```
eugenepeh: perhaps hash displayname since its supposed to be unique? location can be different e.g. http and https etc.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -126,4 +137,22 @@ private static void setAliases(String[] elements, RepoConfiguration config, Auth
             config.setAuthorAliases(author, aliases);
         }
     }
+
+    private static RepoConfiguration buildRepoConfig(String location, String branch) throws ParseException {
+        if (PATH_PATTERN.matcher(location).matches()) {
```
eugenepeh: is there a need to use pattern for this? 
You can just use `Path.get` or `new URL(...)` if `Path` doesn't work isn't it?
Otherwise, we would have to test the Patterns as well.

yong24s: I'll go verify if `URL` is suffice for .git links.

Anyway, I believe these checks would be more appropriate if located in RepoConfiguration` constructor. Should I change it?

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -27,15 +30,22 @@
     /**
      * Positions of the elements of a line in the user-supplied CSV file
      */
-    private static final int ORGANIZATION_POSITION = 0;
-    private static final int REPOSITORY_NAME_POSITION = 1;
-    private static final int BRANCH_POSITION = 2;
-    private static final int GITHUB_ID_POSITION = 3;
-    private static final int DISPLAY_NAME_POSITION = 4;
-    private static final int ALIAS_POSITION = 5;
+    private static final int LOCATION = 0;
```
yamidark: `LOCATION_POSITION`

src/main/java/reposense/builder/ConfigurationBuilder.java
```diff
@@ -9,8 +9,8 @@
 public class ConfigurationBuilder {
     private RepoConfiguration config;
 
-    public ConfigurationBuilder(String organization, String repoName, String branch) {
-        config = new RepoConfiguration(organization, repoName, branch);
+    public ConfigurationBuilder(String location, String branch) {
```
yamidark: `location` sounds abit ambigious (what location?)
maybe can be more specific `repoLocation`?

yong24s: Ah.. the `location` is from a csv config file.
Should I keep it faithful to the column header of the csv config file?

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -190,4 +196,43 @@ public void setAuthorAliases(Author author, String... aliases) {
...
+
+    public String getRepoName() {
+        return repoName;
+    }
+
+    /**
+     * Verifies {@code location} can be presented as a {@code URL} or {@code Path}.
+     * @throws ParseException if otherwise.
+     */
+    private void verifyLocation(String location) throws ParseException {
```
yamidark: Actually, can't we change this to `isPathLocation` and `isUrlLocation` instead, and put them into `CsvParser#processlines`, and make them call different `RepoConfiguration` constructors?
1 which takes a `Path`, and 1 which takes a `URL`?

src/main/java/reposense/commits/CommitInfoExtractor.java
```diff
@@ -21,7 +21,7 @@
      * Extracts out and returns the raw information of each commit for the repo in {@code config}.
      */
     public static List<CommitInfo> extractCommitInfos(RepoConfiguration config) {
-        logger.info("Extracting commits info for " + config.getDisplayName() + "...");
+        logger.info("Extracting commits info for " + config.getLocation() + "...");
```
yamidark: Update `FileInfoExtractor#extractFileInfos`  as well.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -28,11 +40,25 @@
...
+        verifyLocation(location);
+        Matcher matcher = GIT_REPOSITORY_LOCATION_PATTERN.matcher(location);
+
+        if (matcher.matches()) {
+            String organization = matcher.group("org");
+            repoName = matcher.group("repoName");
+            displayName = organization + "_" + repoName + "_" + branch;
+        } else {
+            repoName = Paths.get(location).getFileName().toString();
+            displayName = repoName + "_" + branch;
```
yamidark: Hmm, this may give rise to the same problem mentioned in #206, whereby we give 2 different locations but the folder name is same, and use the same branch, end up overriding each other's report.
Perhaps we can temporarily set the repoName to be the absolute path, just replace the file separators with `_`?

src/main/java/reposense/model/RepoConfiguration.java
eugenepeh: > Actually, can't we change this to isPathLocation and isUrlLocation instead, and put them into CsvParser#processlines, and make them call different RepoConfiguration constructors?
1 which takes a Path, and 1 which takes a URL?

I think this solution is fine, it would be a little redundant to convert String > Path > String again.
However, instead of throwing `ParseException`, perhaps we can make an `InvalidLocationException` instead.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -190,4 +196,43 @@ public void setAuthorAliases(Author author, String... aliases) {
...
+        try {
+            Path pathLocation = Paths.get(location);
+            isPathLocation = Files.exists(pathLocation);
+        } catch (InvalidPathException ipe) {
+            // Ignore exception
+        }
+
+        try {
+            new URL(location);
+            isGitLocation = location.endsWith(GIT_LINK_SUFFIX);
```
eugenepeh: maybe `isValidGitUrl` would be more explanatory

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -190,4 +196,43 @@ public void setAuthorAliases(Author author, String... aliases) {
...
+     * Verifies {@code location} can be presented as a {@code URL} or {@code Path}.
+     * @throws ParseException if otherwise.
+     */
+    private void verifyLocation(String location) throws ParseException {
+        boolean isPathLocation = false;
+        boolean isGitLocation = false;
+
+        try {
+            Path pathLocation = Paths.get(location);
+            isPathLocation = Files.exists(pathLocation);
```
eugenepeh: similarly, maybe `isValidFilePath`? Also, if we're standardizing with url, perhaps you want to add `&& location.endsWith(GIT_LINK_SUFFIX);`

src/main/java/reposense/model/RepoConfiguration.java
eugenepeh: Hmm, is `getFileName()` correct in the first place? Aren't we making user entering `.git` at the end by default?

src/test/java/reposense/util/TestConstants.java
```diff
@@ -5,8 +5,13 @@
 public class TestConstants {
     public static final String TEST_ORG = "reposense";
     public static final String TEST_REPO = "testrepo-Alpha";
+    public static final String TEST_REPO_DISPLAY_NAME = TEST_ORG + "_" + TEST_REPO + "_" + "master";
+    public static final String TEST_REPO_GIT_LOCATION = "https://github.com/" + TEST_ORG + "/" + TEST_REPO + ".git";
+
+    public static final String DISK_REPO_DISPLAY_NAME = "DISK_TEST_REPO";
```
eugenepeh: I think all these constants should be in `GitTestTemplate`?

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -96,10 +96,10 @@ public static String getCommitHashBeforeDate(String root, String branchName, Dat
         return runCommand(rootPath, revListCommand);
     }
 
-    public static String cloneRepo(String org, String repoName) throws IOException {
-        Path rootPath = Paths.get(Constants.REPOS_ADDRESS, org);
+    public static String cloneRepo(String location, String displayName) throws IOException {
+        Path rootPath = Paths.get(Constants.REPOS_ADDRESS, displayName);
```
eugenepeh: Let's have here use `getRepoDirectory`, shift `REPOS_ADDRESS` to `FileUtil.java` and destroy the `Constants` class.

src/main/java/reposense/git/GitDownloader.java
```diff
@@ -15,12 +15,12 @@
 
     private static final Logger logger = LogsManager.getLogger(GitDownloader.class);
 
-    public static void downloadRepo(String organization, String repoName, String branchName)
+    public static void downloadRepo(String location, String displayName, String repoName, String branchName)
```
eugenepeh: a bit strange that downloadRepo isn't using RepoConfig directly, alot of the code within it can be simplified.
E.g. `GitChecker.checkout(config.getRepoRoot())`

sample.csv
```diff
@@ -1,9 +1,9 @@
...
-CS2103JAN2018-F14-B1,main,master,lohtianwei,Loh,
+Repository's Location,Branch,Author's GitHub ID,Author's Display Name,Author's Git Author Name
+https://github.com/CS2103AUG2017-W09-B1/main.git,master,nbriannl,Nbr,
+https://github.com/CS2103AUG2017-W09-B1/main.git,master,zacharytang,Zac,Zachary Tang
+https://github.com/CS2103AUG2017-W09-B1/main.git,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting
+https://github.com/CS2103AUG2017-W09-B1/main.git,master,CindyTsai1,Cin,YuHsuan
+https://github.com/CS2103JAN2018-F14-B1/main.git,master,lithiumlkid,Ahm,
+https://github.com/CS2103JAN2018-F14-B1/main.git,master,codeeong,Cod,
+https://github.com/CS2103JAN2018-F14-B1/main.git,master,jordancjq,Jor,
+https://github.com/CS2103JAN2018-F14-B1/main.git,master,lohtianwei,Loh,
```
eugenepeh: btw, since we're changing the format, I would like to take this opportunity to replace them with our organization as well.
`CS2103AUG2017-W09-B1` -> `testrepo-Beta`
`CS2103JAN2018-F14-B1` -> `testrepo-Delta`

src/main/java/reposense/model/RepoConfiguration.java
yong24s: Will open a issue for this.

eugenepeh:
Windows can have space in the directory, please cater for that special case in the git clone

src/main/java/reposense/parser/InvalidLocationException.java
```diff
@@ -0,0 +1,7 @@
+package reposense.parser;
+
+public class InvalidLocationException extends Exception {
```
yamidark: JavaDocs

src/test/java/reposense/git/GitCheckerTest.java
```diff
@@ -9,37 +10,49 @@
...
+        RepoConfiguration diskConfig = new RepoConfiguration(diskLocation.toString(), "master");
+        GitDownloader.downloadRepo(diskConfig);
+        Path clonedDiskLocation = Paths.get(FileUtil.REPOS_ADDRESS, DISK_REPO_DISPLAY_NAME).toAbsolutePath();
+        TestUtil.compareDirectories(diskLocation, clonedDiskLocation);
+    }
+
     @Test
     public void checkoutBranchTest() {
-        Path branchFile = Paths.get(TestConstants.LOCAL_TEST_REPO_ADDRESS, "inTestBranch.java");
+        Path branchFile = Paths.get(LOCAL_TEST_REPO_ADDRESS, "inTestBranch.java");
```
yamidark: `LOCAL_TEST_REPO_ADDRESS` -> `config.getRepoRoot()`?
same for all below

src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -21,25 +19,36 @@
...
-import reposense.util.TestConstants;
 
 public class GitTestTemplate {
-
-    protected RepoConfiguration config;
+    protected static final String TEST_ORG = "reposense";
+    protected static final String TEST_REPO = "testrepo-Alpha";
+    protected static final String TEST_REPO_GIT_LOCATION = "https://github.com/" + TEST_ORG + "/" + TEST_REPO + ".git";
+    protected static final String LOCAL_TEST_REPO_ADDRESS = FileUtil.REPOS_ADDRESS
+            + File.separator + TEST_ORG + "_" + TEST_REPO + "_" + "master" + File.separator + TEST_REPO;
```
yamidark: Some of these can be directly get from `config`, remove them.

PR #212, opened 2018-07-19
CommandRunner#blameRaw: remove copy and move detection

yamidark:
Part of #203 
```
The git blame command in CommandRunner#blameRaw(String, String)
uses copy and move detection[1] for its line authorship analysis.

These may cause an author's commit contribution and
his line contribution count on the dashboard to be inconsistent,
as a the author may have copied or moved lines from another
author, which will cause those lines to be authored to the other
author instead.

If necessary, users may use collate if they wish to identify if
certain codes belong to them, instead of us trying
to identify the 'original' author by ourselves.

Let's remove the copy and move detection in
CommandRunner#blameRaw(String, String) git blame command.

[1]: Docs on the git blame copy and move detection:
https://git-scm.com/docs/git-blame#git-blame--Mltnumgt

### DIFF
--- a/src/functional/resources/dateRange/expected/reposense_testrepo-Charlie_master/commits.json
+++ b/src/functional/resources/dateRange/expected/reposense_testrepo-Charlie_master/commits.json
@@ -1676,10 +1676,10 @@
     ]
   },
   "authorFinalContributionMap": {
-    "charlesgoh": 948,
-    "jeffreygohkw": 1849,
-    "Esilocke": 3121,
-    "wangyiming1019": 1769
+    "charlesgoh": 1029,
+    "jeffreygohkw": 2191,
+    "Esilocke": 3408,
+    "wangyiming1019": 2411
   },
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: src/functional/resources/dateRange/expected/reposense_testrepo-Charlie_master/commits.json [hunk 1]
```diff
@@ -1676,10 +1676,10 @@
...
   },
   "authorFinalContributionMap": {
-    "charlesgoh": 948,
-    "jeffreygohkw": 1849,
-    "Esilocke": 3121,
-    "wangyiming1019": 1769
+    "charlesgoh": 1029,
+    "jeffreygohkw": 2191,
+    "Esilocke": 3408,
+    "wangyiming1019": 2411
```
damithc: Looks like those flags do make a significant difference. We'll seriously consider bringing them back in a future version.

PR #216, opened 2018-07-21
Extracting info from git diff is not working as expected #215

### LINKED ISSUE #215: Extracting info from git diff is not working as expected
Following #205, running with since date result in following error.

```
java -jar RepoSense.jar -config interns.csv -since 15/05/2018 -formats java adoc js md css jsp tag xml fxml gradle
03:10:08 - Cloning reposense/reposense...
03:10:14 - Cloning completed!
03:10:14 - Extracting commits info for reposense_reposense_master...
03:10:15 - Extracting relevant file infos reposense_reposense_master...
Exception in thread "main" java.lang.AssertionError: Should not have error matching file path pattern inside file diff result!
        at reposense.authorship.FileInfoExtractor.getFilePathFromDiffPattern(FileInfoExtractor.java:205)
        at reposense.authorship.FileInfoExtractor.getEditedFileInfos(FileInfoExtractor.java:92)
        at reposense.authorship.FileInfoExtractor.extractFileInfos(FileInfoExtractor.java:60)
        at reposense.authorship.AuthorshipReporter.generateAuthorshipSummary(AuthorshipReporter.java:21)
        at reposense.report.ReportGenerator.generateReposReport(ReportGenerator.java:53)
        at reposense.RepoSense.main(RepoSense.java:27)
```

yamidark:
Fixes #215:
```
FileInfoExtractor#getFilePathFromDiffPattern(String) checks
for the group pattern '+++ b/path/to/file'
within the git diff result in order to get the file path.

However, when a new and also empty file is added in between the
commit range to analysis, the file will appear in the git diff result,
but since there are no lines to display, the above group does not
get printed, thus 
FileInfoExtractor#getFilePathFromDiffPattern(String) is unable
to match the pattern, causing the program to crash.

Let's add a check to skip these new and empty files in 
the git diff result, since an empty file has no lines to 
attribute any contribution.
```

### DIFF
--- a/src/main/java/reposense/authorship/FileInfoExtractor.java
+++ b/src/main/java/reposense/authorship/FileInfoExtractor.java
@@ -32,7 +32,6 @@ public class FileInfoExtractor {
 
     private static final String DIFF_FILE_CHUNK_SEPARATOR = "\ndiff --git a/.*\n";
     private static final String BINARY_FILE_SYMBOL = "\nBinary files ";
-    private static final String FILE_DELETED_METADATA = "deleted file mode 100644\n";
     private static final String LINE_CHUNKS_SEPARATOR = "\n@@ ";
     private static final String LINE_INSERTED_SYMBOL = "+";
     private static final String STARTING_LINE_NUMBER_GROUP_NAME = "startingLineNumber";
@@ -45,8 +44,10 @@ public class FileInfoExtractor {
             "-(\\d)+(,)?(\\d)* \\+(?<startingLineNumber>\\d+)(,)?(\\d)* @@");
     private static final Pattern FILE_CHANGED_PATTERN = Pattern.compile("\n(\\+){3} b/(?<filePath>.*)\n");
 
-    private static final Predicate<String> NEW_EMPTY_FILE_PREDICATE = Pattern.compile("^new file mode 100644\n"
-            + "index [a-zA-Z0-9]{7}\\.{2}[a-zA-Z0-9]{7}$").asPredicate();
+    private static final Predicate<String> FILE_DELETED_PREDICATE = Pattern.compile(
+            "^deleted file mode [\\d]{6}\n").asPredicate();
+    private static final Predicate<String> NEW_EMPTY_FILE_PREDICATE = Pattern.compile(
+            "^new file mode [a-zA-Z0-9\n. ]*$").asPredicate();
 
     /**
      * Extracts a list of relevant files given in {@code config}.
@@ -90,7 +91,7 @@ public class FileInfoExtractor {
 
         for (String fileDiffResult : fileDiffResultList) {
             // file deleted, is binary file or is new and empty file, skip it
-            if (fileDiffResult.startsWith(FILE_DELETED_METADATA) || fileDiffResult.contains(BINARY_FILE_SYMBOL)
+            if (FILE_DELETED_PREDICATE.test(fileDiffResult) || fileDiffResult.contains(BINARY_FILE_SYMBOL)
                     || NEW_EMPTY_FILE_PREDICATE.test(fileDiffResult)) {
                 continue;
             }
--- a/src/main/java/reposense/authorship/FileInfoExtractor.java
+++ b/src/main/java/reposense/authorship/FileInfoExtractor.java
@@ -37,7 +37,7 @@ public class FileInfoExtractor {
     private static final String LINE_INSERTED_SYMBOL = "+";
     private static final String STARTING_LINE_NUMBER_GROUP_NAME = "startingLineNumber";
     private static final String FILE_CHANGED_GROUP_NAME = "filePath";
-    private static final String MATCH_GROUP_FAIL_FORMAT = "Failed to match the %s group for:%s\n";
+    private static final String MATCH_GROUP_FAIL_MESSAGE_FORMAT = "Failed to match the %s group for:\n%s";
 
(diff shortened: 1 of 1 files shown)


yamidark:
@eugenepeh For your review.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -84,8 +89,9 @@
         String[] fileDiffResultList = fullDiffResult.split(DIFF_FILE_CHUNK_SEPARATOR);
 
         for (String fileDiffResult : fileDiffResultList) {
-            // file deleted or is a binary file, skip it
-            if (fileDiffResult.startsWith(FILE_DELETED_METADATA) || fileDiffResult.contains(BINARY_FILE_SYMBOL)) {
+            // file deleted, is binary file or is new and empty file, skip it
+            if (fileDiffResult.startsWith(FILE_DELETED_METADATA) || fileDiffResult.contains(BINARY_FILE_SYMBOL)
+                    || NEW_EMPTY_FILE_PREDICATE.test(fileDiffResult)) {
```
eugenepeh: actually, why don't do it similarly as `FILE_DELETED_METADATA`?

yamidark: For files that are deleted,the `git diff` result will still print out all the lines in the file (as deleted)
A new file also does the same, printing out all the lines in the file (as added).
However, what we are checking here is a special case where if we add a new file, but that file is empty, thus not having the header group we are trying to find, or any lines to print as well.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -36,13 +37,17 @@
...
+    private static final String MATCH_GROUP_FAIL_MESSAGE_FORMAT = "Failed to match the %s group for:\n%s";
 
     private static final int LINE_CHANGED_HEADER_INDEX = 0;
 
     private static final Pattern STARTING_LINE_NUMBER_PATTERN = Pattern.compile(
             "-(\\d)+(,)?(\\d)* \\+(?<startingLineNumber>\\d+)(,)?(\\d)* @@");
     private static final Pattern FILE_CHANGED_PATTERN = Pattern.compile("\n(\\+){3} b/(?<filePath>.*)\n");
 
+    private static final Predicate<String> NEW_EMPTY_FILE_PREDICATE = Pattern.compile("^new file mode 100644\n"
+            + "index [a-zA-Z0-9]{7}\\.{2}[a-zA-Z0-9]{7}$").asPredicate();
```
eugenepeh: `\\.` seems like an additional `\` here.
But anyway, let's make use of the `+++` and `---` functionality and simplify this
-> `^new file mode [a-zA-Z0-9\n\. ]*$`

eugenepeh:
for the `deleted file mode` please simplify it to exclude the digit as well. I think the number represents file permission, so if someone adds the file with other permission levels, it may bypass this test.

src/main/java/reposense/authorship/FileInfoExtractor.java
yamidark: >\\. seems like an additional \ here.

There is a need for 2 `\`, since the 1st `\` is meant to tell that the 2nd `\` is the actual backslash rather than an escape character.
Anyway, within the square brackets, there is no need for `\` :sweat_smile:.

eugenepeh: I see, I am used to using online regex test so I kinda forgot the syntax for java.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -90,7 +91,7 @@
 
         for (String fileDiffResult : fileDiffResultList) {
             // file deleted, is binary file or is new and empty file, skip it
-            if (fileDiffResult.startsWith(FILE_DELETED_METADATA) || fileDiffResult.contains(BINARY_FILE_SYMBOL)
+            if (FILE_DELETED_PREDICATE.test(fileDiffResult) || fileDiffResult.contains(BINARY_FILE_SYMBOL)
```
eugenepeh: Wouldn't this fail if you replace `startsWith` with predicate test?

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -84,8 +90,9 @@
         String[] fileDiffResultList = fullDiffResult.split(DIFF_FILE_CHUNK_SEPARATOR);
 
         for (String fileDiffResult : fileDiffResultList) {
-            // file deleted or is a binary file, skip it
-            if (fileDiffResult.startsWith(FILE_DELETED_METADATA) || fileDiffResult.contains(BINARY_FILE_SYMBOL)) {
+            // file deleted, is binary file or is new and empty file, skip it
+            if (FILE_DELETED_PREDICATE.test(fileDiffResult) || fileDiffResult.contains(BINARY_FILE_SYMBOL)
```
eugenepeh: Wouldn't this fail if you replace `startsWith` with predicate test?

yamidark: >Wouldn't this fail if you replace startsWith with predicate test?

The predicate has `^` at the start to indicate that the `String` must start with this pattern, thus its the same as `startsWith`, just that `startWith` must be an exact `String`, while we need to check for  a pattern.

PR #227, opened 2018-07-23
Dashboard: determine date range only after all repo data are loaded

ongspxm:
**proposed message**
```
The dashboard as of v1.1.1 is loaded dynamically after each
`commits.json` is loaded.

If the filter date range is not defined, the range will be determined
by the first and last commits of that particular repository. This start
and end dates will then be set as the filter dates to be used for later
rendering of the other repositories.

However, this leads to a problem when the repositories are of a
different date range, if the earlier repository is loaded first, the
date range will not be able to display the ramps in the later
repository, vice versa.

To fix this issue, let's only determine the date ranges after all the
`commits.json` of the repositories is loaded. However, in doing this,
we would only be able to display the ramps after all the repositories
are loaded.

Assuming a steady and rather fast connection to the data server, this
probably wouldn't be an issue, since we are able to load the files
rather quickly, so let's do this anyways.
```

### DIFF
--- a/frontend/src/static/css/style.scss
+++ b/frontend/src/static/css/style.scss
@@ -73,6 +73,8 @@ header{
         height:100%;
         overflow-y:scroll;
         flex-grow:1;
+
+        text-align:center;
     }
 
     #tabs-wrapper{
--- a/frontend/src/static/js/api.js
+++ b/frontend/src/static/js/api.js
@@ -5,47 +5,46 @@ window.enquery = (key, val) => `${key}=${encodeURIComponent(val)}`;
 const REPORT_DIR = '.';
 
 // data retrieval functions //
-function loadJSON(fname, fn) {
-  const err = () => window.alert('unable to get file');
-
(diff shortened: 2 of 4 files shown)


AdityaA1998:
@ongspxm Not being able to preview Netlify

yong24s:
Don't tag `reposense/devs`. Tag us individually first so you don't prematurely involve the senior devs in this PR.

LGTM, but please also include why the change from `callback` to `promises`.

ongspxm:
Oh, cos in order to set the right date range, we will have to wait for all
the `commit.json` to have complete loading before we get absolute start and
end dates.

We could do that by updating the count of a global counter through the xhr
callbacks, but it may generate inconsistent result when multiple callbacks
try to alter the same global variable.

A cleaner way to ensure all the xhr response are handled will be to convert
the xhr calls into a promise which then can be managed through
`Promise.all`.

ongspxm:
@eugenepeh ready for merge?

PR #229, opened 2018-07-23
Add support for file exclusion patterns #218

### LINKED ISSUE #218: Back end support for file exclusion patterns
Objective: to be able to exclude certain files/folders based on a list of glob patterns supplied.

* To be done at a per-repo basis

yamidark:
#Fixes #218 
```
RepoSense only accepts an inclusive list of file formats to analyze.

However, users may wish to specifically exclude analyze of some
files or folders, such as large data files or test folders, as they may
not be meaningful to analyze.

Let's add back-end support to exclude analyzing of files of specified
glob patterns, which can be done in git log by using pathspec[1].

[1] Intro to Git's pathspec:
https://kgrz.io/git-intro-to-pathspec.html
```

### DIFF
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -80,7 +80,7 @@ Branch | The branch to analyse in the target repository
 Author's GitHub ID | GitHub ID of the target contributor in the repository
 Author's Display Name | Optional Field. The value of this field, if not empty, will be displayed in the dashboard instead of author's GitHub ID.
 [Optional] Author's Git Author Name | Detailed explanation below
-[Optional] Ignore Global List | The list of file path globs to ignore analyze. The glob list should be specified on the first author of the repository. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
+[Optional] Ignore Global List | The list of file path globs to ignore during analysis. The glob list should be specified on the first author of the repository. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
 
 #### Git Author Name
 `Git Author Name` refers to the customizable author's display name set in the local `.gitconfig` file.
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -80,6 +80,7 @@ Branch | The branch to analyse in the target repository
 Author's GitHub ID | GitHub ID of the target contributor in the repository
 Author's Display Name | Optional Field. The value of this field, if not empty, will be displayed in the dashboard instead of author's GitHub ID.
 [Optional] Author's Git Author Name | Detailed explanation below
+[Optional] Ignore Global List | The list of file path globs to ignore analyze. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
 
 #### Git Author Name
 `Git Author Name` refers to the customizable author's display name set in the local `.gitconfig` file.
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -80,7 +80,7 @@ Branch | The branch to analyse in the target repository
 Author's GitHub ID | GitHub ID of the target contributor in the repository
 Author's Display Name | Optional Field. The value of this field, if not empty, will be displayed in the dashboard instead of author's GitHub ID.
 [Optional] Author's Git Author Name | Detailed explanation below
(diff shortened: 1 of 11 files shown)


src/main/java/reposense/parser/CsvParser.java
```diff
@@ -122,8 +125,22 @@ private static void setAliases(String[] elements, RepoConfiguration config, Auth
...
+            String[] aliases = elements[ALIAS_POSITION].split(AUTHOR_ALIAS_AND_GLOB_SEPARATOR);
             config.setAuthorAliases(author, aliases);
         }
     }
+
+    /**
+     * Sets the list of globs to ignore inside {@code config} for the file analysis.
+     */
+    private static void setIgnoreGlobList(String[] elements, RepoConfiguration config) {
+        boolean areIgnoreGlobListInElements = elements.length > IGNORE_GLOB_LIST_POSITION
```
yong24s: `isIgnoreGlobListInElements`? sound more intuitive?

src/functional/resources/sample.csv
```diff
@@ -1,9 +1,9 @@
...
-reposense,testrepo-Beta,master,CindyTsai1,Cin,YuHsuan
-reposense,testrepo-Charlie,master,charlesgoh,Cha,Charles Goh
-reposense,testrepo-Charlie,master,Esilocke,Esi,
-reposense,testrepo-Charlie,master,jeffreygohkw,Jef,Jeffrey Goh
-reposense,testrepo-Charlie,master,wangyiming1019,Wan,ACER\kyle;Wang Yiming
+Organization,Repository,branch,Author's GitHub ID,Author's Display Name,Author's Git Author Name,IgnoreGlobList
+reposense,testrepo-Beta,master,nbriannl,Nbr,,
+reposense,testrepo-Beta,master,zacharytang,Zac,Zachary Tang,
+reposense,testrepo-Beta,master,April0616,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting,
+reposense,testrepo-Beta,master,CindyTsai1,Cin,YuHsuan,
```
eugenepeh: Quick question: how do you plan to handle the setting where different set of IgnoreGlobLists entered for each individual author from same repository?

Currently it seems that the latest author will always override the existing `IgnoreGlobList`

yamidark: Hmmm, as we currently are unable to set a 'repo level' setting in the current `.csv` format, we probably just need the user to ensure the `IgnoreGlobList` is set the same for all users, or for the last user.
This should be resolved when we shift over to `config.json` in #232, so maybe we can instead postpone adding this to the `config.csv` for now?

damithc: 1. We need to add support for repo-config.csv soon. 
2. How about this? If something is specified both at repo level and author level, author level _modifies_ (not _replaces_) repo level (cumulative).

yamidark: Since we don't have the repo-config.csv yet, for this PR we will change it to do the analysis and file exclusions by author level?

damithc: I guess you can do that. Alternatively, as the csv file is not a backend matter, you can _assume_ there is a way to specify file exclusions at repo level and provide backend support for it. That way, @yong24s can add the support for repo-config.csv without needing any further changes from you. Otherwise the two of you will have to work together to implement support for file exclusion at repo level.

yamidark: The thing is if we are to support file exclusion at author level, we would need to modify the `FileInfo` extraction/analysis and `CommitInfo` extraction to be done at the author level as well (which is currently at repo level).
I think that for this PR, we will just assume and stick at repo level exclusion, and modify the `FileInfo` extraction and `CommitInfo` extraction to be at author level in another PR?

docs/UserGuide.md
```diff
@@ -80,6 +80,7 @@ Branch | The branch to analyse in the target repository
 Author's GitHub ID | GitHub ID of the target contributor in the repository
 Author's Display Name | Optional Field. The value of this field, if not empty, will be displayed in the dashboard instead of author's GitHub ID.
 [Optional] Author's Git Author Name | Detailed explanation below
+[Optional] Ignore Global List | The list of file path globs to ignore analyze. The glob list should be specified on the first author of the repository. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
```
damithc: `ignore analyze` -> `ignore during analysis`

PR #233, opened 2018-07-24
Dashboard: default view: sort by name within repo #223

### LINKED ISSUE #223: Dashboard: default view: sort by name within repo
v1.1.1a

Current: the default view is grouped by repo and sorted by contribution within the group.

Suggested: group by repo but sort by name within the group

Reason: Sort by contribution puts an unnecessary focus on the contribution LoC which may not be a true reflection of the true contribution.

ongspxm:
fixes #223 

```
The current dashboard as of v1.1.1a sorts the users in the repository
by the number of contributions they have made for the project in that
time period.

This however puts an unnecessary focus on the contribution LoC which
may not be a true reflection of the users' contribution. Let's change
the default view to sort the users by their names instead.
```

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -41,7 +41,7 @@ window.vSummary = {
     return {
       filtered: [],
       filterSearch: '',
-      filterSort: 'totalCommits',
+      filterSort: 'displayName',
       filterSortReverse: false,
       filterGroupRepos: true,
       filterGroupWeek: false,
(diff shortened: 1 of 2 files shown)


damithc:
I thought we decided it is better to sort based on the order in the config file instead?

ongspxm:
the repos will be ordered in config order, but within each repo, the users will be sorted by date

damithc:
> the repos will be ordered in config order, but within each repo, the users will be sorted by date

Why not follow the same sorting order within the repo?

What order do we follow when `Group by repo` is unticked?

I was thinking we should follow the same sorting order (i.e., follow config file) for consistency.

ongspxm:
> Why not follow the same sorting order within the repo?

The dashboard gets the repo order from `summary.json`, which is possible because its an array, hence, there is an order to follow. This is the order that we follow when we the `group by repo` is checked.

However, the `commits.json` (the report where we get the relevant ramp information) is delivered to the dashboard as a key value mapping, hence, there is no default sorting order.

And out of the various fields, displayName makes sense to be the default sorting criteria.

damithc:
Alright, let's go with that then.

Issue #234, opened 2018-07-24
Modify analysis of files and commits to be at Author level


yamidark:
RepoSense only uses Repoconfiguration for the 
analysis of files and commits, causing these analysis
to be on the repo level.

As we will be implementing configuration settings for each
individual author of a repo, the analysis of files and commits
will also need to be on the author level.

Let's update the analysis of files and commits to be on the
author level.

yamidark:
Resolved in #236

PR #236, opened 2018-07-24
 Modify analysis of files and commits to be at Author level #234

yamidark:
```
RepoSense only uses Repoconfiguration for the
analysis of files and commits, causing these analysis
to be on the repo level.

As we will be implementing configuration settings for each
individual author of a repo, the analysis of files and commits
will also need to be on the author level.

Let's update the analysis of files and commits to be on the
author level.
```


yamidark:
The changes to the functional test files for `noDateRange/testrepo-Charlie` is due to some of the commits not being in chronological order previously, while we are sorting them this time. Will post a more detailed investigation later.

yamidark:
screenshot from 2018-07-25 14-35-11
There is a range of commits that are dated from `06/11-07/11`, but in between the commits of the day `08/11`
May have occurred due to amending and then force pushing the commits.

src/test/java/reposense/commits/CommitInfoExtractorTest.java
```diff
@@ -1,20 +1,23 @@
...
 import reposense.commits.model.CommitInfo;
+import reposense.model.Author;
 import reposense.template.GitTestTemplate;
 import reposense.util.TestUtil;
 
 public class CommitInfoExtractorTest extends GitTestTemplate {
 
     @Test
     public void withContentTest() {
+        config.setAuthorList(Collections.singletonList(new Author(".*")));
```
yong24s: Seems like you already have `CommandRunnerTest#getNoFilterAuthor()` to create `new Author(.*)`. 

Why not reuse it for clarity?

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -176,6 +182,25 @@ private static String convertToGitDateRangeArgs(Date sinceDate, Date untilDate)
         return gitDateRangeArgs;
     }
 
+    /**
+     * Returns the {@code String} command to specify the authors to analyze for `git log` command.
+     */
+    private static String convertToFilterAuthorArgs(Author author) {
+        StringBuilder filterAuthorArgsBuilder = new StringBuilder();
+        // the author name which git log checks against will also contain the email enclosed in <> brackets
```
eugenepeh: perhaps just simplify, `ignore check against email`?

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -176,6 +182,25 @@ private static String convertToGitDateRangeArgs(Date sinceDate, Date untilDate)
...
+        filterAuthorArgsBuilder.append(" --author=\"^").append(author.getGitId()).append(" <.*>$");
+        final String cmdFormat = "\\|^%s <.*>$";
```
eugenepeh: hmm, why don't you use format for above as well? would be more readable.

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -176,6 +182,26 @@ private static String convertToGitDateRangeArgs(Date sinceDate, Date untilDate)
...
+        StringBuilder filterAuthorArgsBuilder = new StringBuilder(" --author=\"");
+        // ignore check against email
+        final String cmdFormat = "\\|^%s <.*>$";
+
+        filterAuthorArgsBuilder.append(String.format(cmdFormat, author.getGitId()).substring(2));
+
+        // git author names may contain regex meta-characters, so we need to escape those
+        author.getAuthorAliases().stream()
+                .map(authorAlias -> String.format(
+                        cmdFormat, escapeSpecialRegexChars(authorAlias.replace("\\", "\\\\\\"))))
```
eugenepeh: Perhaps the `.replace("\\", "\\\\\\")` can be included into `escapeSpecialRegexChars` as well

yamidark: If `\\` is also placed into `escapeSpecialRegexChars`, only 1 extra backslash will be added to it, but the `git` author regex pattern requires 2 backslashes, which is why I needed to do a separate escape for it

src/main/java/reposense/model/Author.java
```diff
@@ -40,5 +77,14 @@ public int hashCode() {
     public String toString() {
         return gitId;
     }
+
+    /**
+     * Called after a new ignore glob list is set.
+     * Updates the {@code PathMatcher} to the new ignore glob list set.
```
damithc: Overview statement should be the first line of the header comment

PR #243, opened 2018-07-26
Fix empty ramps when overriding RepoConfiguration

yamidark:
```
When a repo has a _reposense/config.json configuration file,
the RepoConfiguration will be reset to the new config.json file,
and all authors are reset as well.

However, these reset authors does not have their author aliases
or ignore glob list set inside them. This may cause some of the
commit info to be missing as we are filtering commits by author
names and aliases.

Let's fix this by updating the
RepoConfiguration#update(StandAloneConfig) to set the
author aliases and ignore glob list for each author inside the
config.json file.
```

### DIFF
--- a/_reposense/config.json
+++ b/_reposense/config.json
@@ -4,27 +4,32 @@
 		{
 			"githubId": "AdityaA1998",
 			"displayName": "Aditya",
-			"authorNames": ["AdityaA1998"]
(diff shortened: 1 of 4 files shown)


yamidark:
screenshot from 2018-07-26 18-13-46

PR #247, opened 2018-07-27
Update data for testrepo-Delta

yamidark:
```
RepoSense/testrepo-Delta has been updated with the author
aliases for each author for testing if author aliases will be
added correctly from the _reposense/config.json config file.

Let's update the data for unit and functional test to consider those
aliases as well.
```


eugenepeh:
Nits in proposed message:
```
RepoSense/testrepo-Delta has been updated with the author
aliases for each author for testing if author aliases will be
added correctly from the _reposense/config.json config file.
```
=> 
```
testrepo-Delta contains its own standalone configuration,
it has been updated with the author aliases.

Let's update our expected results in the unit and functional test
to match the recent update.
```

Issue #245, opened 2018-07-26
Support a minimal repo-config.csv


damithc:
Modify `-config` CLI parameter to accept a config folder rather than a config file. The config folder can contain `repo-config.csv` and `author-config.csv` files.

PR #249, opened 2018-07-28
Change -config to use a configuration folder #245

yong24s:
Part of #245

```
RepoSense uses a CSV config file for all repository information.

Following #234, analysis can be done at author level.
To make author level analysis available to the users, the CSV config
file has to be broken down into repo-config.csv and author-config.csv.

Let's start by changing cli argument -config to accept a folder to
support multiple CSV config files. 
```


yong24s:
@eugenepeh 
Netlify has to be updated to use `./gradlew clean build && ./gradlew run -Dargs="-config _reposense/repo-config.csv"`
as `-config` is changed to use `folder`

eugenepeh:
@yong24s I believe for `_reposense/repo-config.csv`, `repo-config.csv` shouldn't be part of the parameter.

eugenepeh:
Actually, its quite a bad practice to write into the sibling folder when we specify the `_reposense` folder.
This is because the user has neither given the implicit nor explicit permission for our application to do that.
Perhaps, we should specify the folder before `_reposense` instead.

damithc:
My understanding was:

The user should be able to specify any folder as the config folder, as long as it contains `repo-config.csv`, `repo-config.csv` files. It need not be inside a `_reposense` folder. If the config parameter is not specified, we can look for the two files in `./_reposense`. 

Would that work?

Reason. We assume the command is run in the folder containing RepoSense.jar but the user might have multiple dashboards whose config files are in different locations. e.g.,
`java -jar RepoSense -config interns/config_folder -output interns/report`
`java -jar RepoSense -config cs2103/info -output cs2103/report`

yong24s:
>In that case, if the config filename is fixed, I think its fine to traverse the specified directory to look for the repo-config.csv. I don't think we should look into the cwd unless -config is not specified.
This is to ensure that our application match what the user wants and not over-interpreting it. <br/>
The output will be the directory specified in -output or docs/ in the cwd otherwise

That is what I am doing in this PR.

I have placed `repo-config.csv` in `_reposense` to consolidate all config files into one place. Sorry if this caused confusion.

**The following summaries what this PR intends to solve, to ensure that everyone is on the same page**:

`-config` now only accepts **a** `folder` that contains `repo-config.csv`. It does not validate the `folder name`.

In the `documentation` and `cli usage`, we have mentioned that output files will be at `cwd` if `-output` is not specified. 

`RepoSense` will always only write to `output`. 

In any event, when RepoSense actually writes to a `sibling folder` of `_reposense` then the `cwd` must be at `./_reposense`

src/main/java/reposense/parser/ConfigFolderArgumentType.java
```diff
@@ -0,0 +1,36 @@
...
+        for (String file : REQUIRED_FILES_IN_CONFIG_FOLDER) {
+            matchCount += (Files.exists(Paths.get(value, file))) ? 1 : 0;
+        }
+
+        if (matchCount != REQUIRED_FILES_IN_CONFIG_FOLDER.size()) {
+            throw new ArgumentParserException(
+                    String.format(PARSE_EXCEPTION_MESSAGE_MISSING_REQUIRED_CONFIG_FILES, value), parser);
+        }
+
+        return new File(value);
```
yamidark: Perhaps we can just return a `Path` straight?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -22,21 +22,23 @@
...
+            .getResource("output").getFile()).toPath();
     private static final Path OUTPUT_DIRECTORY_ABSOLUTE = new File(ArgsParserTest.class.getClassLoader()
             .getResource("output").getFile()).toPath();
-    private static final Path CONFIG_FILE_RELATIVE = PROJECT_DIRECTORY.relativize(CONFIG_FILE_ABSOLUTE);
+    private static final Path CONFIG_FOLDER_RELATIVE = PROJECT_DIRECTORY.relativize(CONFIG_FOLDER_ABSOLUTE);
     private static final Path OUTPUT_DIRECTORY_RELATIVE = PROJECT_DIRECTORY.relativize(OUTPUT_DIRECTORY_ABSOLUTE);
-    private static final String DEFAULT_MANDATORY_ARGS = "-config " + CONFIG_FILE_ABSOLUTE + " ";
+    private static final String DEFAULT_MANDATORY_ARGS = "-config " + CONFIG_FOLDER_ABSOLUTE + " ";
+    private static final Path REPO_COFIG_CSV_FILE =
+            Paths.get(CONFIG_FOLDER_ABSOLUTE.toString(), CsvParser.REPO_CONFIG_FILENAME);
```
yamidark: `CONFIG_FOLDER_ABSOLUTE.resolve(CsvParser.REPO_CONFIG_FILENAME)`

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -41,15 +44,17 @@
      *
      * @throws IOException if user-supplied csv file does not exists or is not readable.
      */
-    public static List<RepoConfiguration> parse(Path configFilePath) throws IOException {
-        assert (configFilePath != null);
+    public static List<RepoConfiguration> parse(Path configFolderPath) throws IOException {
```
yamidark: Feels weird that a `CsvParser` is passed a folder to parse, instead of the `csv` file directly.
Perhaps get the `RepoSense` main class to resolve and pass the `csv` file path here?

src/main/java/reposense/parser/ConfigFolderArgumentType.java
```diff
@@ -0,0 +1,36 @@
...
+            "Unable to find " + CsvParser.REPO_CONFIG_FILENAME;
+    private static final List<String> REQUIRED_FILES_IN_CONFIG_FOLDER = Arrays.asList(CsvParser.REPO_CONFIG_FILENAME);
+
+    @Override
+    public File convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
+        Arguments.fileType().verifyExists().verifyIsDirectory().verifyCanRead().convert(parser, arg, value);
+        int matchCount = 0;
+
+        for (String file : REQUIRED_FILES_IN_CONFIG_FOLDER) {
+            matchCount += (Files.exists(Paths.get(value, file))) ? 1 : 0;
```
yamidark: Maybe instead of this `matchCount`, we just can throw the `ArgumentParserException` straight if we can't find the file in the folder, which also lets the user know which mandatory config file is missing when we have multiple mandatory files.
That is if we are doing a single folder, and not a directory where we need to recursively search child folders 😅

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -46,9 +46,9 @@ private static ArgumentParser getArgumentParser() {
                 .action(new HelpArgumentAction());
 
         mutexParser.addArgument("-config")
-                .type(Arguments.fileType().verifyExists().verifyIsFile().verifyCanRead())
+                .type(new ConfigFolderArgumentType())
                 .metavar("PATH")
-                .help("The path to the CSV config file to read.");
+                .help("The path to the config folder containing repo-config.csv.");
```
yamidark: `containing the config files.` in general, instead of specifically `repo-config.csv`, since we are adding more in the future.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -163,23 +165,29 @@ public void missingMandatoryConfigArg_throwsParseException() throws ParseExcepti
...
         String input = String.format("-config %s", relDirectory);
         ArgsParser.parse(translateCommandline(input));
     }
 
     @Test(expected = ParseException.class)
-    public void parse_absoluteConfigFileNotExists_throwsParseException() throws ParseException {
-        String absConfigFile = Paths.get(PROJECT_DIRECTORY.toString(), "somefile_not_existent.csv").toString();
-        String input = String.format("-config %s", absConfigFile);
+    public void parse_notExistsConfigFolder_throwsParseException() throws ParseException {
+        String absConfigFolder = Paths.get(PROJECT_DIRECTORY.toString(), "non_existing_random_folder").toString();
```
yamidark: `PROJECT_DIRECTORY.resolve(...)`

src/main/java/reposense/parser/ConfigFolderArgumentType.java
yong24s: Ah.. I used `File` because I didn't want to change too much. Originally `-config` and `-output` both used a helper method that returns `File`.

Should I write a `ArgumentType` for `-output`so both can use `Path`?

src/main/java/reposense/parser/ArgsParser.java
yong24s: I intend to print the filename explicitly as RepoSense will fast-fail if otherwise.

Also, the upcoming specs will only add support for `repo-config.csv` and `author-config.csv`.

src/main/java/reposense/parser/ConfigFolderArgumentType.java
yamidark: >Should I write a ArgumentType for -outputso both can use Path?

Yep, can do that. Then we won't need have to do `null` check.

src/main/java/reposense/parser/ArgsParser.java
yamidark: >I intend to print the filename explicitly as RepoSense will fast-fail if otherwise.

fast-fail? We are just updating the help message here right?

yong24s: **Correction for clarity**: I intend to print the filename explicitly as RepoSense will fast-fail ~if otherwise~ as now the `-config` expects a folder that contains a `repo-config.csv`.

src/main/java/reposense/parser/ConfigFolderArgumentType.java
```diff
@@ -0,0 +1,29 @@
...
+
+import net.sourceforge.argparse4j.impl.Arguments;
+import net.sourceforge.argparse4j.inf.Argument;
+import net.sourceforge.argparse4j.inf.ArgumentParser;
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+import net.sourceforge.argparse4j.inf.ArgumentType;
+
+public class ConfigFolderArgumentType implements ArgumentType<Path> {
+    private static final String PARSE_EXCEPTION_MESSAGE_MISSING_REQUIRED_CONFIG_FILES =
+            "Unable to find " + CsvParser.REPO_CONFIG_FILENAME;
```
eugenepeh: `The required config file %s is not found in the specified folder.`

PR #251, opened 2018-07-30
 Some browsers show extra element #209

### LINKED ISSUE #209: Some browsers show extra element
image

This can be fixed by setting the `mui-select->label`
    `overflow-x` property to `visible`

e.g.
```
.mui-select>label {
    position: absolute;
    top: 0;
    display: block;
    width: 100%;
    color: rgba(0,0,0,.54);
    font-size: 12px;
    font-weight: 400;
    line-height: 15px;
    overflow-x: visible;
    text-overflow: ellipsis;
    white-space: nowrap;
}
```

ongspxm:
fixes #209 

```In some browsers (chrome, edge), the height of the content in the
"label" tag seems to be greater than the height allocated to it by the
render engine. This results in the "scrollbar" arrows popping up to
show the users the "full content".

This can be distracting for users when they are trying to navigate the
site. Let's remove this scrollbar arrows by applying the "overflow-y:
hidden" CSS setting to the filter labels.
```

### DIFF
--- a/frontend/src/static/css/style.scss
+++ b/frontend/src/static/css/style.scss
@@ -150,6 +150,10 @@ header{
     &__date:first-child{
         margin-right:0.5rem;
     }
+
+    label{
+        overflow-y:hidden;
+    }
 }
 
 .summary-charts{


damithc:
I still get it when I hover over `Since` and `Until`. Do you want to fix it in this PR or as a separate PR?
image

ongspxm:
Let me look into it, to see if its a small fix.

eugenepeh:
> I still get it when I hover over Since and Until. Do you want to fix it in this PR or as a separate PR?

Actually those button are pretty useful compared to the useless overflow button from before.
The `up` and `down` allows fine date adjustment and the `x` button resets the date back to the limit.

ongspxm:
i guess we should keep it, since, its part of the default implementation?

PR #255, opened 2018-07-31
 Large ramp blocks interaction of ramps under it #189

### LINKED ISSUE #189: Large ramp blocks interaction of ramps under it
image

In the above image, the large magenta ramp is blocking clicks to ramps that is covered by it.

Let's use a smaller Z index for larger ramp, so overlapping ramps are clickable.

ongspxm:
fixes #189 

```
Ramps representing very huge commits will be rendered as a very
humongous triangle. This display blocks up the commits in the earlier
periods, hence causing them to be unclickable.

To fix this, let's render the ramps such that the earlier ramps will
have a larger z-index value to place it on top of the more recent
ramps. This ensures that users will be able to interact with all the
ramps.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -92,6 +92,7 @@ html
                   v-bind:href="getSliceLink(user, slice)", target="_blank",
                   v-bind:class="'summary-chart__ramp__slice--color'+j%5",
                   v-bind:style="{\
+                    zIndex: user.commits.length - j,\
                     borderLeftWidth: getWidth(slice) + 'em',\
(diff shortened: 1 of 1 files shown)


damithc:
> Should we do something about the ramp size as well?
> I am thinking of changing it to log scale.

Let's keep it in a normal scale for now. Log scale is not intuitive to the human eye. RS should show the 'truth' no matter how ugly it is. Attempts to prettify should not distort the truth. Having said that, we _can_ have prettified alternative views as long as we have the 'ugly truth' version too. That's why I'm keen to have a per-commit view as well because that's the ultimate truth. per-day and per-week views are prettified versions that distort the truth.

PR #256, opened 2018-07-31
Dashboard: able to search for multiple keywords

ongspxm:
```
Current search function filters the users based on whether the search
string is a match for the particular user. However, it might be useful
for users to be able to select ramp charts from multiple projects.

Let's allow the search filter to pick out users based on search params
separated by spaces.
```


frontend/src/static/js/api.js
```diff
@@ -59,7 +59,7 @@ window.api = {
           };
 
           const searchParams = [
-            repo.organization, repo.repoName,
+            repo.displayName,
             obj.displayName, author,
```
eugenepeh: if we use repo.displayName, do we still need obj.displayName?

ongspxm: right right, let me go check

ongspxm: yep, still need, on line 51 on (paiseh, dun know how link), obj.displayName is the displayName of the particular author, (eg Cin, all those).

in case there is confusion, author refers to their github username.

eugenepeh: ic... we need to clean up the variable names one of these days to make it clearer.

PR #260, opened 2018-07-31
Remove unused ConfigurationBuilder.java

yong24s:


eugenepeh:
As mentioned in #288,

>On a side note, we should clean up the repoconfig setup one day. Seems like an awful number of lines to setup the expected config.

This could be one of the possible solution, hence we should probably revamp it instead of removing it.

eugenepeh:
Closing this PR for now.

PR #248, opened 2018-07-28
Add repo file exclusion patterns to standalone config #244

yong24s:
Fixes #244

```
Following #218, file exclusion glob patterns can be supplied in the
csv config file.

Let's decentralized this setting to the _reposense/config.json so
repository owners can apply it to their repositories.
```

PR #265, opened 2018-08-02
Ignore glob list does not propagate to code view #262

### LINKED ISSUE #262: Ignore glob list does not propagate to code view
Following #218 and #248, file exclusion glob patterns can be supplied in the
`csv config files` and `_reposense/config.json`.

Files excluded by the glob patterns only apply to ramps but not code snippet in code view.

In the below diagram, `**.java` is excluded. 
Notice that exclusion works on ramps but not code view.

codee

yamidark:
Fixes #262 
```
Each author has their own ignoreGlobList, which should ignore
any of their contributions to any files within that glob list.

However, while the commits information are correctly ignored,
the authorship information is not, as some files that should be
in the glob list still appeared for the author's contribution.
This is due to AnnotatorAnalyzer not ignoring the files in the
glob list, and overriding the original git blame results.

Let's resolve this by updating AnnotatorAnalyzer to remove
author contribution if the file being analyzed is inside their
glob list.
```

### DIFF
--- a/src/functional/resources/noDateRange/expected/reposense_testrepo-Delta_master/authorship.json
+++ b/src/functional/resources/noDateRange/expected/reposense_testrepo-Delta_master/authorship.json
@@ -25600,1012 +25600,6 @@
     }
   },
   {
-    "path": "src/main/java/seedu/address/commons/events/ui/ChangeTagColourEvent.java",
-    "lines": [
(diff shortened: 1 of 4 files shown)


yamidark:
Will wait till #248 is merged before updating the functional test results.

PR #279, opened 2018-08-03
Add tests to RepoConfiguration#merge

yong24s:
```
AuthorConfigCsvParser#getRepoConfiguration should return existing
RepoConfiguration from a list of RepoConfigurations. Otherwise it adds
a newly created RepoConfiguration into the list and returns it.

However, due to a bug, any RepoConfigurations existing or not are added
to the list of RepoConfigurations causing duplicates.

Let's add test to RepoConfiguration#merge to prevent bugs. 
```


src/main/java/reposense/parser/CsvParser.java
```diff
@@ -106,7 +106,7 @@ protected String getValueInElement(final String[] elements, int position) {
         }
 
         String manyValue = getValueInElement(elements, position);
-        // Wrap with new ArrayList<> to make the created List is mutable.
+        // Wrap with new ArrayList<> to make list resizable.
```
eugenepeh: to support all list functionality.

PR #281, opened 2018-08-03
Add clickable link to repository #226

### LINKED ISSUE #226: Add clickable link to repository
To view the repository, we need to manually open the link to navigate to it.
Let's simplify this by adding a redirect link in the dashboard itself.

eugenepeh:
Fix #226 

To view the repository, we need to manually open the link to navigate to it.

Let's simplify this by adding a redirect link to the remote repository in the 
dashboard itself.

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -76,7 +76,10 @@ html
               span group by repo
         #summary-charts
           .summary-charts(v-for="repo of filtered", v-if="repo.length>0")
-            .summary-charts--title(v-if="filterGroupRepos") {{ repo[0].repoPath }}
+            a.summary-charts__title(
+              v-if="filterGroupRepos",
+              v-bind:href="repo[0].location", target="_blank"
+            ) {{ repo[0].repoPath }}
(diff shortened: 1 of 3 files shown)


frontend/src/index.jade
```diff
@@ -76,7 +76,8 @@ html
               span group by repo
         #summary-charts
           .summary-charts(v-for="repo of filtered", v-if="repo.length>0")
-            .summary-charts--title(v-if="filterGroupRepos") {{ repo[0].repoPath }}
+            .summary-charts--title(v-if="filterGroupRepos")
+              a.summary-chart__title--url(v-bind:href="repo[0].location", target="_blank") {{ repo[0].repoPath }}
```
ongspxm: can do directly on the title, dun need to create another a tag inside

```
a.summary-charts__title(
    v-if="filterGroupRepos", 
    v-bind:href="repo[0].location", 
    target="_blank") {{ repo[0].repoPath }}
```

rational: usually, we will keep the html as light as possible, only necessary stuff

also side note, my mistake before that, should be `.summary-charts__title`, not `.summary-charts--title`, so if css need to change, then kam sia also.

PR #282, opened 2018-08-03
Add total number of lines contributed in code panel #264

### LINKED ISSUE #264: Add total number of lines contributed in code panel
Code panel could use more useful information such as number of lines contributed for each files.
This would give the reviewer a rough idea how many lines the author has contributed to each component.

ongspxm:
fixes #264 

```
Code panel could use more useful information such as number of lines
contributed for each file. This would give the reviewer a rough idea
how many lines the author has contributed to each component.

Let's add the line of code information into the display.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -111,7 +111,7 @@ html
 
         .files(v-if="isLoaded")
           .file(v-for="file in files")
-            .title(onclick="toggleNext(this)") {{ file.path }} ({{ file.lineCount }} LOC)
+            .title(onclick="toggleNext(this)") {{ file.path }} ({{ file.lineCount }} lines)
             pre.hljs.java
               code
                 template(v-for="segment in file.segments")
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -111,7 +111,7 @@ html
 
         .files(v-if="isLoaded")
           .file(v-for="file in files")
-            .title(onclick="toggleNext(this)") {{ file.path }}
+            .title(onclick="toggleNext(this)") {{ file.path }} ({{ file.lineCount }} LOC)
             pre.hljs.java
               code
                 template(v-for="segment in file.segments")
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -137,6 +137,8 @@ window.vAuthorship = {
         }
       });
 
+      res.sort((a, b) => b.lineCount - a.lineCount);
+
       this.files = res;
       this.isLoaded = true;
     },
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -121,9 +121,11 @@ window.vAuthorship = {
       const res = [];
 
       files.forEach((file) => {
-        if (file.authorContributionMap[this.info.author]) {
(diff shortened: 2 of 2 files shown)


damithc:
IIRC, the code view used to be sorted by the number of lines modified. Rationale: Major contributions are at the top of the list. Files at the bottom of this list can be ignored. 

It seems that is not the case anymore. Need to bring that feature back @ongspxm

ongspxm:
> contains 30 lines from manual calculation instead of just 28 lines shown. (need investigation)

this is kinda a feature, to remove all the untouched code segments that are too tiny, like one or two lines. we dump them to the next touched code instead to be displayed together.

https://github.com/reposense/RepoSense/blob/62c0f68866b02222e90016c5f102a7bbcb55783b/frontend/src/static/js/v_authorship.js#L88-L107

ongspxm:
> It seems that is not the case anymore. Need to bring that feature back

roger that :+1:

damithc:
> this is kinda a feature, to remove all the untouched code segments that are too tiny, like one or two lines. we dump them to the next touched code instead to be displayed together.

I wish this kind of "features" are run by me first :-p This feature is problematic because it distorts the truth and can result in students getting penalized for code they didn't write.

ongspxm:
this was part od the old ui as well, was ported over, using the same 5 line
count

damithc:
> this was part od the old ui as well, was ported over, using the same 5 line count.

That's quite possible. I wouldn't have noticed even now if it wasn't mentioned in the comment, but students are likely to notice it when we use it on their code. I don't even mind having it as a feature but not as the default. For now, let's create an issue for it and try to fix it as soon as we can.

ongspxm:
so make it a optional feature, disabled by default?

damithc:
> so make it a optional feature, disabled by default?

Yup, that's the idea.

damithc:
> so make it a optional feature, disabled by default?

In fact, it can even be the default if there is a way to disable it.

ongspxm:
activated by checkbox? not this pr?

damithc:
> activated by checkbox? not this pr?

Not this PR.
Is this optimization done totally on the front-end? In that case I guess it can be a check-box, which means we need a settings panel on top of the code panel as well.

eugenepeh:
>activated by checkbox? not this pr?

Why even need a checkbox...? how about letting them be not encapsulated by default?

E.g. we encapsulate them in the expand box only if untouched code > 5 lines but we still differentiate them using color irregardless.

eugenepeh:
Opened issue #289

damithc:
> E.g. we encapsulate them in the expand box only if untouched code > 5 lines but we still differentiate them using color irregardless.

That might be even better 👍

ongspxm:
> you missed out this comment of mine.

paiseh, paiseh, do now

ongspxm:
> LGTM, just add a space in between will do

donez

damithc:
Feature looks good 👍 
Not sure what's going on here though. Says 1 line, but shown more than 1 line. Unrelated bug?
image

eugenepeh:
> Not sure what's going on here though. Says 1 line, but shown more than 1 line. Unrelated bug?

Think its the same bug as discussed earlier, described in #289, as the additional lines are still lesser than minimum lines.

PR #286, opened 2018-08-06
Dashboard: optimize settings bar to fit various screen sizes #258

### LINKED ISSUE #258: Dashboard: optimize settings bar in 'code view' format
v1.1.2

Current:
image

Suggested: optimize the layout so that settings appear neater and doesn't take so much space.
e.g., use smaller fonts for values

ongspxm:
fixes #258 

```
Layout not very optimized, settings looks messy and takes up a lot of
space.

Let's reduce the font size for the values to make the settings look
neater.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -55,6 +55,11 @@ html
               option(value="displayName") Author Name
               option(value="searchPath") Team Name
             label sort by
+          .mui-select.summary-picker__input
+            select(v-model="filterTimeFrame")
+              option(value="day") Day
+              option(value="week") Week
+            label granularity
           .summary-picker__input
             .mui-textfield.summary-picker__date
               input(type="date", v-model="filterSinceDate")
@@ -62,11 +67,6 @@ html
             .mui-textfield.summary-picker__date
               input(type="date", v-model="filterUntilDate")
               label until
-          .mui-select.summary-picker__input
-            select(v-model="filterTimeFrame")
-              option(value="day") Day
-              option(value="week") Week
-            label granularity
           .mui-checkbox.summary-picker__checkboxes.summary-picker__input
             label
               input(type="checkbox", v-model="filterSortReverse").summary-picker__checkbox
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -45,34 +45,35 @@ html
     vuetemplate#v_summary
       #summary.wrapper
         form.summary-picker.mui-form--inline(v-on:submit.prevent="getFiltered();")
-          .mui-textfield.summary-picker__input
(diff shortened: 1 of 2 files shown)


ongspxm:
image

@damithc look okay?

damithc:
> @damithc look okay?

Yes. Check if it degrades gracefully for slightly smaller screen sizes. We don't have to cater for mobile devices, but it's nice if we can support a low end laptop (or even a decent size tablet).

damithc:
> Check if it degrades gracefully for slightly smaller screen sizes. We don't have to cater for mobile devices, but it's nice if we can support a low end laptop (or even a decent size tablet).

If not easy to do, responsiveness can be left out as a separate issue.

ongspxm:
consider it done.

it will look something like that on ipad air mini (according to firefox devtools)
screen shot 2018-08-06 at 17 06 28

damithc:
> it will look something like that on ipad air mini (according to firefox devtools)

Nice 👍

eugenepeh:
image

still doesn't seems to degrade gracefully, I think its partially because the `since` and `until` are packed as single div, whereas the others are one div on their own.

Perhaps group the `search`, `sort` and `granularity` as single div will balance it better?

ongspxm:
> still doesn't seems to degrade gracefully,

mobile too small dah... but okay, i put the day thingy in front see look like wut

eugenepeh:
> mobile too small dah... but okay, i put the day thingy in front see look like wut

perhaps make them align to the left in their own div? The main div can still be aligned to the center.

ongspxm:
> perhaps make them align to the left in their own div

its not really grouped together, its a flexbox, all separated, only the dates grouped together.
the flexbox currently stretch across 100%, if i align left, will align left everything, can?

eugenepeh:
> its not really grouped together, its a flexbox, all separated, only the dates grouped together.
the flexbox currently stretch across 100%, if i align left, will align left everything, can?

uh that's why i propose them to be enclosed separately in an div each earlier.

```
Perhaps group the search, sort and granularity as single div will balance it better?
```

If its too much of a change, we can do it in future.

ongspxm:
> uh that's why i propose them to be enclosed separately in an div each earlier.

paiseh, too much angmo, might have been confused and missed it out.

i will give it a try, see how it looks

ongspxm:
best i can do is size based alignment, if smaller than a certain size, align left, and things like that.

unless you dun mind having the large screen one align left as well.

because there isn't any way for the rendering engine to tell the css that the flexbox is being wrapped. the css wont know when to change to `text-align:left`

ongspxm:
@eugenepeh imma just gonnna do the wrap in div for now, let me know if you need the align left to be in too

Issue #259, opened 2018-07-31
Dashboard: add a 'show code' button


damithc:
v1.1.2

Current:
![image](https://user-images.githubusercontent.com/1673303/43456689-063e8ad8-94f7-11e8-842e-0309c2bd1a13.png)

Suggested: We can add a small `show code` button (or an icon with a tooltip) so that new users can intuitively know how to view code.

On a related note, I'm wondering if we should put the contribution bar above the ramp chart because the clicking on the name (or the proposed button) shows the code related to the bar, not the ramp chart. It's more intuitive if the two appear together.

PR #287, opened 2018-08-06
 Dashboard: add an indicator to code view

ongspxm:
part of #259 

```
New users are confused as to how to view the code from an author.

Let's add a tooltip to the ramp chart title so that new users can
intuitively know how to view code.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -84,9 +84,11 @@ html
               .summary-chart__title(
                 v-on:click="$emit('view-authorship', {author:user.name, repo:user.repoName, name:user.displayName})"
               )
+                i.summary-chart__title--tooltip.fas.fa-code
                 .summary-chart__title--index {{ i+1 }}
                 .summary-chart__title--repo(v-if="!filterGroupRepos") {{ user.repoPath }}
                 .summary-chart__title--name {{ user.displayName }}
+                .summary-chart__title--tooltip view author's code
               .summary-chart__ramp
                 a.summary-chart__ramp__slice(
                   v-for="(slice, j) in user.commits",
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -80,7 +80,7 @@ html
               v-if="filterGroupRepos",
               v-bind:href="repo[0].location", target="_blank"
             ) {{ repo[0].repoPath }}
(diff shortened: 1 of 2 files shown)


ongspxm:
this tooltip is similar to github's header link in the markdown view. please check if appropriate.

image
the link icon available on hover

ongspxm:
@damithc do give a check on netlify to see if okay

damithc:
> @damithc do give a check on netlify to see if okay

1. I prefer to have it as a proper tool tip if that is possible. That is, the tip is a callout _from/near_ the icon, or at least together with the icon rather than split the icon and the text.
Example:
image

2. If not possible to have both _before_ the name, can consider having both _after_ the name.

ongspxm:
image

in that case, i will be using the browser default title tag for this feature, it will look something like that

damithc:
> couldn't get the cursor in the screenshot, but it is at the title of of the ramp chart
> 
> in that case, i will be using the browser default title tag for this feature, it will look something like that

Sure. 
I have to say I find it more intuitive if the icon comes after the name, but that could be just me.
Also, I prefer if the icon is visible all the time, as opposed to appearing on hover. The whole purpose of this feature is to tell the user that there is a way to see the code, without user having to read the user guide. If the icon appears only on hover, we leave it to the chance that user will discover the feature by accident.

ongspxm:
> I have to say I find it more intuitive if the icon comes after the name

i could do that, but it looks a little ugly, because the icon will then be unaligned.
image

damithc:
> i could do that, but it looks a little ugly, because the icon will then be unaligned.

Let's do that in this version. We can tweak in a future version if required.

ongspxm:
> Also, I prefer if the icon is visible all the time, as opposed to appearing on hover.
image
Also possible, but the icon will look like a bullet point instead, okay?

damithc:
> Also possible, but the icon will look like a bullet point instead, okay?

Not applicable as we'll show the icon at the end of the name.

ongspxm:
donez

eugenepeh:
Proposed message tweak:

```
New users may not know all the functionalities that are available to
them, hence hidden functionality such as code view may go unnoticed.

To enhance our user friendliness, let's add a tool-tip and icon to the
ramp chart title so that the code view feature stands out more by
intuitive.
```

PR #288, opened 2018-08-06
Add functionality to ignore standalone config #272

### LINKED ISSUE #272: Allow author-config.csv to step in when config.json is missing
e.g., `repo-config.csv` has 2 repos `foo`, `bar`
* `foo` has provided a `config.json`. 
* `bar` has not provided a `config.json`

We should be able to provide author data of repo `bar` in `author-config.csv` while using `config.json` for `foo`.

Justification: to cater for cases where student team hasn't added the json file on time.

yong24s:
Fixes #272

```
RepoSense offers decentralization of configuration by allowing repository
owners to set up their own standalone config.

While this enhance the ease of use, it does not offer our users the flexibility
to overrule the standalone configuration with their own.

Let's add an additional configuration column in repo-config.csv to
forcibly use the user supplied configuration by ignoring
the one defined by the repository owner.

```


src/main/java/reposense/parser/JsonParser.java
```diff
@@ -20,16 +20,17 @@
 
     /**
      * Converts json file from the given {@code path} into an object and returns it.
-     * @throws FileNotFoundException if {@code path} is invalid.
+     * @throws IOException if {@code path} is invalid.
```
eugenepeh: is this change necessary?

src/functional/resources/repo-config.csv
```diff
@@ -1,4 +1,4 @@
-Repository's Location,Branch,Ignore Glob List
-https://github.com/reposense/testrepo-Beta.git,master,docs**
-https://github.com/reposense/testrepo-Charlie.git,master,
-https://github.com/reposense/testrepo-Delta.git,master,
+Repository's Location,Branch,Ignore Glob List,Ignore standalone config.json
```
eugenepeh: Since this is like an enabling/disabling option, perhaps `Disable standalone config` would be more appropriate.
Another reason is that it avoids confusion with `Ignore Glob List`

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -8,13 +8,15 @@
 
 public class RepoConfigCsvParser extends CsvParser<RepoConfiguration> {
     public static final String REPO_CONFIG_FILENAME = "repo-config.csv";
+    private static final String IGNORE_STANDALONE_CONFIG_JSON_KEYWORD = "overrule";
```
eugenepeh: lets keep it simple, maybe just yes or no will do like most cli boolean option.

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -31,8 +41,10 @@
             Arrays.asList("collated**", "*.aa1", "**.aa2", "**.java");
     private static final List<String> SECOND_AUTHOR_GLOB_LIST = Arrays.asList("collated**", "**[!(.md)]");
 
-    @Test
-    public void configJson_overridesRepoConfig_success() throws InvalidLocationException, GitDownloaderException {
+    private static RepoConfiguration REPO_DELTA_STANDLONE_CONFIG;
```
eugenepeh: make it static final, can just initialize here.

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -44,28 +56,85 @@ public void configJson_overridesRepoConfig_success() throws InvalidLocationExcep
...
         expectedConfig.setAuthorDisplayName(FIRST_AUTHOR, "Ahm");
-        expectedConfig.setAuthorDisplayName(SECOND_AUTHOR, "Cod");
-        expectedConfig.setAuthorDisplayName(THIRD_AUTHOR, "Jor");
-        expectedConfig.setAuthorDisplayName(FOURTH_AUTHOR, "Loh");
 
         expectedConfig.setIgnoreGlobList(REPO_LEVEL_GLOB_LIST);
 
-        RepoConfiguration actualConfig = new RepoConfiguration(TEST_REPO_DELTA, "master");
+        Path testFolderPath = new File(CsvParserTest.class.getClassLoader()
+                .getResource("RepoConfigurationTest/repoconfig_ignore_test").getFile()).toPath();
```
eugenepeh: make such value constant for maintainability.

src/main/java/reposense/parser/JsonParser.java
```diff
@@ -20,16 +20,17 @@
...
         return fromJson(new Gson(), path, getType());
     }
 
-    protected T fromJson(Gson gson, Path path, Type type) throws FileNotFoundException {
-        JsonReader jsonReader = new JsonReader(new FileReader(path.toString()));
-        return gson.fromJson(jsonReader, type);
+    protected T fromJson(Gson gson, Path path, Type type) throws IOException {
+        try (JsonReader jsonReader = new JsonReader(new FileReader(path.toString()))) {
+            return gson.fromJson(jsonReader, type);
+        }
```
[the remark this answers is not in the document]
yong24s: It is necessary as config.json file has to be closed after read, in order to ensure that `repos` folder can be deleted, to pass the tests.

src/main/java/reposense/parser/JsonParser.java
yong24s: The exception signature has changed after wrapping `JsonReader` with `try-with-resources ` block
i.e. `try (JsonReader jsonReader = new JsonReader(...))`. 

Reason of change is in the reply to @yamidark.

src/test/java/reposense/parser/RepoConfigurationTest.java
yong24s: As discussed, this nit will be ignored. 

There is no direct way to make it a `static final`, solutions may decrease code quality.

docs/UserGuide.md
```diff
@@ -100,6 +100,7 @@ Column Name | Explanation
 Repository's Location | The `GitHub URL` or `Disk Path` to the git repository
 Branch | The branch to analyse in the target repository
 [Optional] Ignore Global List | The list of file path globs to ignore during analysis for each author. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
+[Optional] Disable standalone config | Overrule the usage of standalone config in repositories. Leaving it as blank or any values other than `yes` will be interpreted as `no`.
```
eugenepeh: avoid using the word overrule as its not very appropriate for this scenario. 
you're not overruling anybody, you're just simply choosing not to use it.

eugenepeh: repositories -> repository.

eugenepeh: To disable the usage of standalone config in target repository, please enter `yes` in this column. Otherwise, the configuration in the target repository will be used by default.

eugenepeh:
> Generally looks good. The term Ignore is more appropriate here than Disable? What do you guys think?

I am okay with Ignore. Chose `Disable` in first place because I do prefer not to confuse the users by using similar wording as `Ignore Glob List`.

docs/UserGuide.md
```diff
@@ -100,6 +100,7 @@ Column Name | Explanation
 Repository's Location | The `GitHub URL` or `Disk Path` to the git repository
 Branch | The branch to analyse in the target repository
 [Optional] Ignore Global List | The list of file path globs to ignore during analysis for each author. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
+[Optional] Ignore standalone config | Ignore the usage of standalone config in repository. To ignore the usage of standalone config in target repository, please enter **`yes`** in this column. Otherwise, the configuration in the target repository will be used by default.
```
eugenepeh: Ignore the ~usage~ `presence` of standalone config in `target` repository. ~To ignore the usage of standalone config in target repository~ `To do so`,

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -41,8 +43,19 @@ protected void processLine(List<RepoConfiguration> results, String[] elements) t
         String location = getValueInElement(elements, LOCATION_POSITION);
         String branch = getValueInElement(elements, BRANCH_POSITION);
         List<String> ignoreGlobList = getManyValueInElement(elements, IGNORE_GLOB_LIST_POSITION);
+        String ignoreStandaloneConfig = getValueInElement(elements, IGNORE_STANDALONE_CONFIG_POSITION);
 
-        RepoConfiguration config = new RepoConfiguration(location, branch, ignoreGlobList);
+        if (!ignoreStandaloneConfig.equalsIgnoreCase(IGNORE_STANDALONE_CONFIG_KEYWORD)
+                && !ignoreStandaloneConfig.isEmpty()) {
+            logger.warning(
+                    "Ignoring unknown value " + ignoreStandaloneConfig + " used to ignore standalone config.");
```
yamidark: "Ignoring unknown value " + ignoreStandaloneConfig + " **in** ignore standalone config **column**.", since it technically isn't being used.

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -41,8 +43,19 @@ protected void processLine(List<RepoConfiguration> results, String[] elements) t
...
+            ignoreStandaloneConfig = "";
+        }
+
+        boolean isStandaloneConfigIgnored = !ignoreStandaloneConfig.isEmpty();
```
yamidark: can we set boolean earlier, then for the error logging, we check that `(!isStandaloneConfigIgnored && !ignoreStandAloneConfig.isEmpty())`

eugenepeh:
Proposed message nit:

>RepoSense offers decentralization of configuration to repostories in a
form of a config.json file.

RepoSense offers decentralization of configuration by allowing repository 
owners to set up their own standalone config.

> As the config.json is owned by repositories owners and RepoSense will
use it when available, this does not allow ease of analysis with
different configurations.

While this enhance the ease of use, it does not offer our users the flexibility
to overrule the standalone configuration with their own.

>Let's add an additional column in repo-config.csv to overrule
config.json, to allow analysis with other configurations.

Let's add an additional configuration column in repo-config.csv to 
forcibly use the user supplied configuration by ignoring
the one defined by the repository owner.

eugenepeh:
> Just a nit. I'm sorry I didn't raise this before, but is Repo-provided more self-explanatory than Standalone? Reason: while the config in concern is used for stand-alone mode, it can be used for other modes too. Repo-provided is more direct and definitive. what do you guys think?

Yep sounds good. Or maybe we can just name it as `repo-config` similar to git terminology.

The `.csv` config file we should rename it as `main-config` instead.

damithc:
> The .csv config file we should rename it as main-config instead.

`main` is a bit ambiguous though. The user has to know which one _we_ consider as the `main` (that's an extra bit of knowledge the user needs to have).

However, we can use `config.json` (instead of `repo-provided`) because the file name uniquely identifies what we are referring to. e.g., the column name can be `Ignore config.json`

eugenepeh:
>However, we can use config.json (instead of repo-provided) because the file name uniquely identifies what we are referring to. e.g., the column name can be Ignore config.json

It would be best if we provide a terminology to it though. We've been referring as `Standalone Configuration` in our [documentations](https://github.com/reposense/RepoSense/blob/master/docs/UserGuide.md#standalone-configuration). Users who are not a repository owner may not know that `config.json` == `Standalone Configuration`.

How about to the user, we refer it as `Remote repo-config`?
To the repository owner, we can refer it as `repo-config`?

The `.csv` file we refer it as `RepoSense config`? Hence rename to `RepoSense.csv`?

damithc:
hmmm... I'd rather not introduce additional terms as the three files names are unique already -- but the current three file names may not be good enough for that purpose. Still, now may not be the time to tackle this issue. Let's revisit this in December. For now, we can go ahead with the current state of the PR.

PR #284, opened 2018-08-06
Dashboard: update code base to align with vue style guide

ongspxm:

PR #290, opened 2018-08-07
Reduce memory cache in code view #267

### LINKED ISSUE #267: Fix memory leak in code view
Current: The browser memory keep increasing as a user keeps going through code of various authors.

Suggested: Limit the cache of code view data to a fixed number of authors. Can even be 1.

ongspxm:
fixes #267 

```
Currently, the authorship data is cached in the `window.REPOS` to
reduce unnecessary loading because `authorship.json` is a very big
file. However, this causes the the browser memory keep increasing as a
user keeps going through code of various authors.

Let's limit the cache of code view data to a fixed number of authors to
5.
``

### DIFF
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -37,7 +37,7 @@ window.vAuthorship = {
     initiate() {
       const repo = window.REPOS[this.info.repo];
 
-      if (repoCache.length === 5) {
+      if (repoCache.length === 3) {
         const toRemove = repoCache.shift();
         if (toRemove !== this.info.repo) {
           delete window.REPOS[toRemove].files;
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -22,6 +22,7 @@ window.toggleNext = function toggleNext(ele) {
   }
 };
 
+const repoCache = [];
 window.vAuthorship = {
   props: ['info'],
   template: window.$('v_authorship').innerHTML,
@@ -35,6 +36,14 @@ window.vAuthorship = {
   methods: {
     initiate() {
       const repo = window.REPOS[this.info.repo];
+
+      if (repoCache.length === 5) {
+        const toRemove = repoCache.shift();
+        if (toRemove !== this.info.repo) {
+          delete window.REPOS[toRemove].files;
+        }
+      }
+      repoCache.push(this.info.repo);
 
       if (repo.files) {
         this.processFiles(repo.files);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ongspxm** [comment]
oh yah, this seems too small to be in a new pr. its the empty state thingy,
i realized it when i tried coddee

**Comment by eugenepeh** [comment]
The change is small  but it is still targeting a potential issue. Can separate it into another PR to make it easier to search in future.

File: frontend/src/static/js/main.js [hunk 1]
```diff
@@ -72,6 +75,7 @@ window.app = new window.Vue({
       });
       return full;
     },
+
```
eugenepeh: All the changes in this file here seems unnecessary.

ongspxm: yeah, just some minor cleanups i missed before that

ongspxm:
> The change is small but it is still targeting a potential issue. Can separate it into another PR to make it easier to search in future.

removed and shifted to new PR

eugenepeh:
> removed and shifted to new PR

This PR doesnt seems to be updated.

frontend/src/static/js/main.js
eugenepeh: > yeah, just some minor cleanups i missed before that

Such changes should only be done if the changes touch file or in a separate PR. 
Otherwise it will complicate things should we need to revert the PR in future.

ongspxm: i removed the part for the empty state already, so i removed the alignment as well?

ongspxm:
@eugenepeh removed the blank lines alignment thingy, check plz

eugenepeh:
> can i stuff it into #284?

Can since its almost like a clean up

ongspxm:
@eugenepeh  donez

damithc:
Also, do some manual tests to confirm the memory use is stable after this fix.

eugenepeh:
> Also, do some manual tests to confirm the memory use is stable after this fix.

Tested, doesn't seems to have any memory leak issues. However, loading a large contribution e.g. 400k lines still strains the system esp on the UI side.

I wonder if we can simply load the title first (collapses them) to reduce that strain. @ongspxm do you think its possible?

eugenepeh:
message tweak:

>Currently, the authorship data is cached in the `window.REPOS` to
reduce unnecessary loading because `authorship.json` is a very big
file. However, this causes the the browser memory keep increasing as a
user keeps going through code of various authors.

>Let's limit the cache of code view data to a fixed number of authors to
5.

```
Code view caches the past loaded repository's authorship data to avoid
repetitive loading which increases the response time.

However, this cause the memory usage to keep increasing as
more repository's authorship data gets stored as cache.

As users are unlikely to revisit repositories that were loaded long
ago, let's preserve only most recent 2 repository's authorship data
as cache.
```

PR #292, opened 2018-08-08
Code view: tweak code packing algorithm #289 

### LINKED ISSUE #289: Code view: tweak code packing algorithm
Untouched code segments that are smaller than minimum lines are displayed together with the touched code.

This distorts the truth as these code do not belong to the author and may lead to unnecessary disputes.

ongspxm:
fixes #289 

```
Untouched code segments that are smaller than minimum lines are
displayed together with the touched code. This distorts the truth as
these code do not belong to the author and may lead to unnecessary
disputes.

Let's encapsulate untouched code in the expand box only if >5 lines but
still differentiate them using color irregardless.
```


ongspxm:
> The toggle view shouldn't be there if its small changes.
yep, done

eugenepeh:
image

Why is there a gap for this part?

ongspxm:
image

donez

ongspxm:
as of previous comment, letting files be closed by default

ongspxm:
> Also, on a side note (not that important but would be nice if it can be fixed), the original file has a new line in between here (highlighted below). Any idea why it is gone?

fixed this as well

damithc:
> as of previous comment, letting files be closed by default

Just to be clear, in the code view, files should be expanded by default. Otherwise the user has to click hundreds of times to view files. In future we can collapse files by default if the total diff is too big to load at once. We can also consider collapse-by-default if there is a 'expand/collapse all' button.

eugenepeh:
>Just to be clear, in the code view, files should be expanded by default. Otherwise the user has to click hundreds of times to view files. In future we can collapse files by default if the total diff is too big to load at once. We can also consider collapse-by-default if there is a 'expand/collapse all' button.

Oops sorry, my comment in the other PR #290 was only meant as a question/suggestion for future enhancement. 
@ongspxm don't suddenly add features not related to the PR >_<

On a side note, making it collapse by default doesn't fully resolve the rendering issue. May need to look into Lazy Loading.

PR #295, opened 2018-08-12
Support specifying of file formats at repo level #271

### LINKED ISSUE #217: Support a minimal project.json
The idea is to allow `{target repo}/_reponses/project.json` to provide RepoSense with some repo/author config data

More info in the [google doc](https://docs.google.com/document/d/17qDXIJ4jKKWlMuKcPmtOoLjr98hkel-l2Dt66tyK7vA/edit?usp=sharing)

To keep thing simple, we can allow project.json to override RS's own config files. We'll add an 'ignore repo-supplied config' flag in a future version.

We can download the files directly from the repo using an http request. We can look for a better way to get the files in a later version.

### LINKED ISSUE #271: Support specifying file types at repo-level
Current: file types are specified as a CLI parameter

Suggested: support specifying file types as a column in repo-config.csv

If both are specified, repo-level overrides the cli parameter (?)

yong24s:
Fixes #271 

```
RepoSense allows specifying of file formats for analysis through CLI.

However, this sets all repositories to the same file formats.

Let's make file formats configurable in CSV and standalone configs
to improve usage on repositories with different file formats
and codebases.
```


src/main/java/reposense/parser/CsvParser.java
```diff
@@ -15,7 +15,7 @@
 import reposense.system.LogsManager;
 
 public abstract class CsvParser<T> {
-    protected static final String AUTHOR_ALIAS_AND_GLOB_SEPARATOR = ";";
+    protected static final String MANY_VALUES_SEPARATOR = ";";
```
yamidark: maybe `COLUMN_VALUES_SEPARATOR` instead?

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -96,7 +96,7 @@ protected String getValueInElement(final String[] elements, int position) {
 
     /**
      * Gets the value of {@code position} in {@code elements}.
-     * Returns the value of {@code position} in an array, delimited by {@code AUTHOR_ALIAS_AND_GLOB_SEPARATOR}
+     * Returns the value of {@code position} in an array, delimited by {@code MANY_VALUES_SEPARATOR}
      * if it is in {@code element} and not empty.
      * Otherwise returns an empty array.
```
yamidark: Forgot that this returns a `List` now instead of an `Array`. do update header comments

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -134,4 +144,45 @@ public void repoConfig_wrongKeywordUseStandaloneConfig_success()
 
         TestUtil.compareRepoConfig(REPO_DELTA_STANDALONE_CONFIG, actualConfig);
     }
+
+    @Test
+    public void repoConfigWithFormats_doesNotUseCliFormats_success() throws ParseException, IOException {
```
yamidark: should be `repoConfig_withFormats_ignoreCliFormats`?
same for the rest below

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -326,4 +336,23 @@ private void verifyLocation(String location) throws InvalidLocationException {
...
+     */
+    private static boolean isValidFormat(String value) {
+        return value.matches(FORMAT_VALIDATION_REGEX);
+    }
+
+    /**
+     * Checks that all the strings in the {@code formats} are alphanumeric.
+     * @throws IllegalArgumentException if any of the values do not meet the criteria.
+     */
+    private void validateFormats(List<String> formats) {
```
eugenepeh: shouldn't this be static as well?
Also, ~alphanumeric~ -> `valid formats`

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -96,18 +96,17 @@ protected String getValueInElement(final String[] elements, int position) {
...
      */
     protected List<String> getManyValueInElement(final String[] elements, int position) {
         if (!containsValueAtPosition(elements, position)) {
             return Collections.emptyList();
         }
 
         String manyValue = getValueInElement(elements, position);
-        // Wrap with new ArrayList<> to support all list functionalities.
-        return new ArrayList<>(Arrays.asList(manyValue.split(AUTHOR_ALIAS_AND_GLOB_SEPARATOR)));
+        return Arrays.stream(manyValue.split(COLUMN_VALUES_SEPARATOR)).map(String::trim).collect(Collectors.toList());
```
eugenepeh: is there a need to change this? `Stream` has slower performance compare to the previous.

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -55,6 +57,8 @@ protected void processLine(List<RepoConfiguration> results, String[] elements) t
         RepoConfiguration config =
                 new RepoConfiguration(location, branch, ignoreGlobList, isStandaloneConfigIgnored);
 
+        config.setFormats(formats);
```
eugenepeh: should probably make formats a constructor param as well.

src/test/resources/StandaloneConfigTest/invalidFormats_config.json
```diff
@@ -0,0 +1,30 @@
+{
+	"ignoreGlobList": ["collated**"],
+	"formats": ["java", " && rm -rf /"],
```
eugenepeh: please don't use malicious code in test... although it doesn't work in this scenario, but what if it works?

src/test/resources/RepoConfigurationTest/repoconfig_keyword_test/repo-config.csv
```diff
@@ -1,2 +1,2 @@
-Repository's Location,Branch,Ignore Glob List,Ignore standalone config
-https://github.com/reposense/testrepo-Delta.git,master,collated**,yesss
+Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config
+https://github.com/reposense/testrepo-Delta.git,master,,collated**,yesss
```
eugenepeh: make it the same as above

src/test/resources/RepoConfigurationTest/repoconfig_ignore_test/repo-config.csv
```diff
@@ -1,2 +1,2 @@
-Repository's Location,Branch,Ignore Glob List,Ignore standalone config
-https://github.com/reposense/testrepo-Delta.git,master,collated**,yes
+Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config
+https://github.com/reposense/testrepo-Delta.git,master,java;adoc;md,collated**,yes
```
eugenepeh: this file's folder should be renamed to ignoreStandAlone

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -133,6 +138,10 @@ public void update(StandaloneConfig standaloneConfig) {
         authorDisplayNameMap.clear();
         ignoreGlobList = standaloneConfig.getIgnoreGlobList();
 
+        if (!standaloneConfig.getFormats().isEmpty()) {
```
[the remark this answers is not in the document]
yong24s: If `empty` list overrides then the resulting formats set will be larger than the `default` formats used by RS.

src/main/java/reposense/parser/CsvParser.java
yong24s: I realize I should trim spaces. `i.e. java   ;   adoc`.

src/main/java/reposense/model/RepoConfiguration.java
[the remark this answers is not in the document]
eugenepeh: standalone config is meant to override the RS config.
if the user doesn't want to be override, he/she should ignore sa config

src/functional/resources/repo-config.csv
```diff
@@ -1,5 +1,5 @@
-Repository's Location,Branch,Ignore Glob List,Ignore standalone config
-https://github.com/reposense/testrepo-Alpha.git,master,,
-https://github.com/reposense/testrepo-Beta.git,master,docs**,yes
-https://github.com/reposense/testrepo-Charlie.git,master,,
-https://github.com/reposense/testrepo-Delta.git,master,,
+Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config
+https://github.com/reposense/testrepo-Alpha.git,master,,,
+https://github.com/reposense/testrepo-Beta.git,master,java;adoc,docs**,yes
+https://github.com/reposense/testrepo-Charlie.git,master,java;adoc,,
+https://github.com/reposense/testrepo-Delta.git,master,java;adoc,,
```
eugenepeh: is your feature working? how is it that adding the formats `java;adoc` to 3 of the test repos, yet the functional still pass without change to the generated files?

docs/StandaloneConfiguration.md
```diff
@@ -27,6 +28,11 @@
 ```
 "ignoreGlobList": ["**.dat", "**.js"]  <-- Repository level's list of file formats to ignore.
```
eugenepeh: We should update the example here to demonstrate how both can use in conjunction with each other.
Let's change the examples to folder ignore instead.

src/functional/resources/repo-config.csv
yong24s: Because functional tests was using `java` and `adoc` as formats.

docs/StandaloneConfiguration.md
```diff
@@ -25,14 +26,20 @@
...
+"formats": ["html", "css"]                         <-- Repository level's file formats to analyse.      
+                                                       If not provided, the following file formats will be used.
+                                                       adoc, cs, css, fxml, gradle, html, java, js, json, jsp,
+                                                       md, py, tag, xml.
 
-"githubId": "bob"                      <-- Only githubId is mandatory.
-                                           Optional information can be left out to prevent clutter.
+"ignoreGlobList": ["**.css"]                       <-- Author level's ignoreGlobList adds on to the Repository
+                                                       level's. Thus, the actual ignoreGlobList for alice would
+                                                       contain css, dat and js.
```
eugenepeh: description not updated.

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -119,19 +147,58 @@ public void repoConfig_ignoresStandaloneConfig_success()
...
+        RepoConfiguration.setFormatsToRepoConfigs(actualConfigs, cliArguments.getFormats());
+
+        Assert.assertEquals(1, actualConfigs.size());
+        Assert.assertEquals(CONFIG_FORMATS, actualConfigs.get(0).getFormats());
+    }
+
+    @Test
+    public void repoConfig_withoutFormats_ignoreCliFormats() throws ParseException, IOException {
+        String formats = String.join(" ", CLI_FORMATS);
+        String input = String.format("-config %s -formats %s", WITHOUT_FORMATS_TEST_CONFIG_FILES, formats);
```
eugenepeh: Actually, now that we have added the `formats` into config file, should we remove `-formats` from cli arguments?

@damithc your opinion on this?

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -119,19 +147,58 @@ public void repoConfig_ignoresStandaloneConfig_success()
...
+        List<RepoConfiguration> actualConfigs =
+                new RepoConfigCsvParser(((ConfigCliArguments) cliArguments).getRepoConfigFilePath()).parse();
+        RepoConfiguration.setFormatsToRepoConfigs(actualConfigs, cliArguments.getFormats());
+
+        Assert.assertEquals(1, actualConfigs.size());
+        Assert.assertEquals(CONFIG_FORMATS, actualConfigs.get(0).getFormats());
+    }
+
+    @Test
+    public void repoConfig_withoutFormats_ignoreCliFormats() throws ParseException, IOException {
```
eugenepeh: this should be `_useCliFormats` isn't it?

docs/StandaloneConfiguration.md
```diff
@@ -25,14 +26,20 @@
...
-                                           css, dat and js.
+"formats": ["html", "css"]                         <-- Repository level's file formats to analyse.      
+                                                       If not provided, the following file formats will be used.
+                                                       adoc, cs, css, fxml, gradle, html, java, js, json, jsp,
+                                                       md, py, tag, xml.
 
-"githubId": "bob"                      <-- Only githubId is mandatory.
-                                           Optional information can be left out to prevent clutter.
+"ignoreGlobList": ["**.css"]                       <-- Author level's ignoreGlobList adds on to the Repository
+                                                       level's. Thus, the actual ignoreGlobList for alice would
```
eugenepeh: should be `for bob`

src/functional/java/reposense/Entry.java
```diff
@@ -73,7 +71,7 @@ private void generateReport(String inputDates) throws IOException, URISyntaxExce
 
         RepoConfiguration.merge(repoConfigs, authorConfigs);
 
-        RepoConfiguration.setFormatsToRepoConfigs(repoConfigs, TESTING_FILE_FORMATS);
+        RepoConfiguration.setFormatsToRepoConfigs(repoConfigs, ArgsParser.DEFAULT_FORMATS);
```
eugenepeh: shouldn't this use cliArguments? Also, add a `-formats` arg in e.g. java, adoc

src/functional/resources/repo-config.csv
```diff
@@ -1,5 +1,5 @@
-Repository's Location,Branch,Ignore Glob List,Ignore standalone config
-https://github.com/reposense/testrepo-Alpha.git,master,,
-https://github.com/reposense/testrepo-Beta.git,master,docs**,yes
-https://github.com/reposense/testrepo-Charlie.git,master,,
-https://github.com/reposense/testrepo-Delta.git,master,,
+Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config
+https://github.com/reposense/testrepo-Alpha.git,master,java;adoc,,
+https://github.com/reposense/testrepo-Beta.git,master,java;adoc,docs**,yes
```
eugenepeh: following the functional cli change, let's make all empty, other than beta. Beta should use a diff value from cli arg.

damithc:
I think cli argument can still serve a purpose. cli argument applies to all repos unless overridden by the repo config which can then be overridden by author config, right?

eugenepeh:
>I think cli argument can still serve a purpose. cli argument applies to all repos unless overridden by the repo config which can then be overridden by author config, right?

Yup, was mislead-ed for a moment. Sorry about that >_<

docs/StandaloneConfiguration.md
yong24s: Fixed as discussed.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -118,10 +122,11 @@ public static void merge(List<RepoConfiguration> repoConfigs, List<RepoConfigura
     }
 
     /**
-     * Sets all {@code RepoConfiguration} in {@code configs} to have {@code formats} set.
+     * Sets {@code formats} to {@code RepoConfiguration} in {@code configs} if {@code formats} is empty.
```
eugenepeh: `if {@code formats} is empty.` this part is wrong. Shouldn't use `{@code ...}` for internal implementation.

Perhaps `if its formats is empty.` will do.

repo-config.csv
```diff
@@ -1,6 +1,5 @@
-Repository's Location,Branch,Ignore Glob List,Ignore standalone config
-https://github.com/reposense/testrepo-Alpha.git,master,,
-https://github.com/reposense/testrepo-Beta.git,master,docs**,yes
-https://github.com/reposense/testrepo-Delta.git,master,,
-https://github.com/reposense/RepoSense.git,master,,
-https://github.com/reposense/RepoSense.git,nonExistentBranch,,
+Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config
+https://github.com/reposense/testrepo-Alpha.git,master,,,
+https://github.com/reposense/testrepo-Beta.git,master,java;adocs,docs**,yes
+https://github.com/reposense/testrepo-Delta.git,master,html;css,,
```
eugenepeh: try to keep it similar to functional as possible. Also, you've accidentally left out the last line.

docs/StandaloneConfiguration.md
eugenepeh: > Fixed as discussed.

Seems like nothing has changed though

repo-config.csv
yong24s: As discussed, this `repo-config.csv` is used for both `netlify` and `user guide` so we should split it.
Keeping it similar to the `functional` copy may be confusing to those who read the `user guide`.

As such, I will make a simpler `sample` `repo-config.csv` under `docs` folder for `user guide` and keep the current one similar to `functional` so we can visualize the output with `netlify`.

docs/UserGuide.md
```diff
@@ -103,7 +103,7 @@ gradlew run -Dargs="-view output_path/reposense-report"
 ## CSV Config Files
 #### Repository configuration file
 to configure the list of repositories to analyze and the respective repository level options. <br/>
-[repo-config.csv](../repo-config.csv) is an example of a configuration file setup. It should contain the following columns:
+[repo-config.csv](repo-config.csv) is an example of a configuration file setup. It should contain the following columns:
```
eugenepeh: we should create a example copy of author as well for docs.

repo-config.csv
```diff
@@ -1,5 +1,5 @@
 Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config
 https://github.com/reposense/testrepo-Alpha.git,master,,,
-https://github.com/reposense/testrepo-Beta.git,master,java;adocs,docs**,yes
-https://github.com/reposense/testrepo-Delta.git,master,html;css,,
-https://github.com/reposense/RepoSense.git,master,,,
+https://github.com/reposense/testrepo-Beta.git,master,fxml,docs**,yes
+https://github.com/reposense/testrepo-Delta.git,master,,,
+https://github.com/reposense/testrepo-Delta.git,nonExistentBranch,,,
```
eugenepeh: Please add back the RepoSense as we want to continuously visualize a `live` repo.
Others can all follow the properties of functional config.

docs/UserGuide.md
```diff
@@ -103,19 +103,20 @@ gradlew run -Dargs="-view output_path/reposense-report"
...
+[Optional] File formats | The file formats to analyse in `alphanumerical`. If not provided, the following file formats will be used. `adoc, cs, css, fxml, gradle, html, java, js, json, jsp, md, py, tag, xml`
+[Optional] Ignore Global List | The list of file path globs to ignore during analysis for each author. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
 [Optional] Ignore standalone config | Ignore the presence of standalone config in target repository. To do so, enter **`yes`** in this column. Otherwise, the configuration in the target repository will be used by default.
 [Optional] Ignore Commit List | The list of commits to ignore during analysis. For accurate results, the commits should be provided with their full hash.
 
 #### Author configuration file [Optional]
 to configure the list of authors to analyse and the options. <br/>
-[author-config.csv](../author-config.csv) is an example of a configuration file setup. It should contain the following columns:
+[author-config.csv](author-config.csv) is an example of a configuration file setup. It should contain the following columns:
```
eugenepeh: the docs version of author and repo config need to include the recently added commits ignore as well.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -96,18 +96,17 @@ protected String getValueInElement(final String[] elements, int position) {
 
     /**
      * Gets the value of {@code position} in {@code elements}.
-     * Returns the value of {@code position} in an array, delimited by {@code AUTHOR_ALIAS_AND_GLOB_SEPARATOR}
+     * Returns the value of {@code position} in a {@code List}, delimited by {@code COLUMN_VALUES_SEPARATOR}
```
eugenepeh: `in` -> `as`

eugenepeh: `the value of {@code element} at {@code position}`

src/main/java/reposense/model/StandaloneConfig.java
```diff
@@ -23,6 +24,14 @@
         return ignoreGlobList;
     }
 
+    public List<String> getFormats() {
+        if (formats == null) {
+            return Collections.emptyList();
```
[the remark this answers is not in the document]
yong24s: I had implemented `RepoConfiguration#update method to set the default formats if the list is empty.`

> eugenepeh 
standalone config is meant to overwrite the RS config.
if the user doesn't want to be overwritten, he/she should ignore sa config

I think we just need to handle the case where format list is `null` -> use `default` list

docs/UserGuide.md
```diff
@@ -103,19 +103,20 @@ gradlew run -Dargs="-view output_path/reposense-report"
...
-[repo-config.csv](../repo-config.csv) is an example of a configuration file setup. It should contain the following columns:
+[repo-config.csv](repo-config.csv) is an example of a configuration file setup. It should contain the following columns:
 
 Column Name | Explanation
 ----------- | -----------
 Repository's Location | The `GitHub URL` or `Disk Path` to the git repository
 Branch | The branch to analyse in the target repository
-[Optional] Ignore Glob List | The list of file path globs to ignore during analysis for each author. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
+[Optional] File formats | The file formats to analyse in `alphanumerical`. If not provided, the following file formats will be used. `adoc, cs, css, fxml, gradle, html, java, js, json, jsp, md, py, tag, xml`
+[Optional] Ignore Global List | The list of file path globs to ignore during analysis for each author. More details on the Java glob standard [here](https://javapapers.com/java/glob-with-java-nio/)
```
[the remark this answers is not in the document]
yong24s: Not sure, the `merge` reverted it.

src/main/java/reposense/model/StandaloneConfig.java
```diff
@@ -26,7 +28,7 @@
 
     public List<String> getFormats() {
         if (formats == null) {
-            return Collections.emptyList();
+            return ArgsParser.DEFAULT_FORMATS;
```
eugenepeh: hmm I guess this was the reason why our contribution bar was empty earlier.

Btw I think we should customize our standalone config's formats as well now that its in place.
Perhaps let's add the following to our formats?

.md, .java, .js, .jade, .scss, .gradle, .yml 

Let me know if i missed out any.

src/main/java/reposense/model/StandaloneConfig.java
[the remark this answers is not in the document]
yamidark: >I had implemented RepoConfiguration#update method to set the default formats if the list is empty.

From what I see, the `RepoConfiguration#update` method only calls `this.setFormats(standaloneConfig.getFormats()`, thus it sets the empty list previously when the `formats` is not specified in `repo-config`.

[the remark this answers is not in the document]
eugenepeh: > I had implemented RepoConfiguration#update method to set the default formats if the list is empty.

Yes. This is on the account that the csv configuration may have set the formats for RepoConfiguration. Hence, we cannot just reuse the preset formats in the RepoConfiguration.

src/main/java/reposense/model/StandaloneConfig.java
eugenepeh: nvm, leave the _reposense config alone for now.

yong24s: Will use `["gradle", "jade", "java", "js", "md", "scss", "yml"]` in `_reposense/config.json`

eugenepeh:
> LGTM. I wonder if there is a better term we can use in place of formats though.

How about `ext`? Short form for filename extensions. Would be shorter for user to type in command line too.

damithc:
> How about ext? Short form for filename extensions. Would be shorter for user to type in command line too.

too cryptic though. I can't think of a better one though. Anyway, we can tackle that at a later time. `formats` is not that bad.

PR #302, opened 2018-08-15
Report folder not generated when failed to analyze repo #301

### LINKED ISSUE #301: Report folder not generated when failed to analyze repo
When we fail to clone/checkout a config's repository, we will
skip analyzing that repository, thus not generating its report folder.

However, as the config is still inside the list, we will still add that
'failed' config into `summary.json`. This will cause the dashboard to
crash as it tries to load a report folder which does not exist.

Let's update ReportGenerator to generate the report folder with 
'empty reports, so as to allow the dashboard to be viewed, and also
allow the user to identify errors in which repository config info given.

yamidark:
Fixes #301 
```
When we fail to clone/checkout a config's repository, we will
skip analyzing that repository, thus not generating its report folder.

However, as the config is still inside the config list, we will still add
that 'failed' config into summary.json. This will cause the dashboard
to crash as it tries to load a report folder which does not exist.

Let's update ReportGenerator to generate the report folder with
'empty' reports when failing to clone/checkout the repository, so as
to allow the dashboard to be viewed, and also allow the user to identify
errors in which repository config info given.
```


src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -49,10 +50,11 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
             try {
-                GitDownloader.downloadRepo(config);
                 FileUtil.createDirectory(repoReportDirectory);
+                GitDownloader.downloadRepo(config);
             } catch (GitDownloaderException gde) {
                 logger.log(Level.WARNING, "Exception met while trying to clone the repo, will skip this one", gde);
+                generateEmptyRepoReport(repoReportDirectory.toString());
                 continue;
             } catch (IOException ioe) {
                 logger.log(Level.WARNING, "Error while creating repo directory, will skip this repo.", ioe);
```
yamidark: Should we also handle 'failing to create the directory' exception, by removing the config from the `configs` list?

src/main/java/reposense/report/CommitReportJson.java
```diff
@@ -18,6 +20,27 @@
     private final Map<Author, Float> authorContributionVariance;
     private final Map<Author, String> authorDisplayNameMap;
 
+    /**
+     * Empty constructor to construct an empty commit report.
+     */
+    public CommitReportJson() {
```
eugenepeh: seems okay for a temporary solution, but this adds on to the maintenance work.
I think it would probably be better if we handle this on the ui side.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -49,10 +50,11 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
         for (RepoConfiguration config : configs) {
             Path repoReportDirectory = Paths.get(outputPath, config.getDisplayName());
             try {
-                GitDownloader.downloadRepo(config);
                 FileUtil.createDirectory(repoReportDirectory);
+                GitDownloader.downloadRepo(config);
```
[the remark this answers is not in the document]
yamidark: We need to create the report directory before we can create the report files inside it, and if `GitDownloader.downloadRepo(config)` fails, the directory would not be created earlier, causing the 'empty' report file creation to fail.

src/main/java/reposense/report/CommitReportJson.java
yamidark: This solution is to handle the UI(dashboard) side, where we show in the dashboard that we failed to download this repo.
Also, currently there is logging when `GitDownloader` fails, perhaps we can add on to that logged error message that we are creating an empty report folder for them?

eugenepeh: > *Edit: oh, perhaps you meant if we failed to find the folder in the frontend side, then handle it over there? This would also handle the IOException if we fail to create the report files for whatever reason.

Yes, we can create and attach a feedback field on the RepoConfig, so that when it fails to load on the frontend, we can show that feedback on the frontend by default. 

It would be nice to not treat the 2 exception too differently.

For now, shall we just remove the repoconfig from the list (so that it won't be generated in the summary) since user would've gotten the error message from terminal?

yong24s:
Another possible way for front-end to know about the failure is to mark failed `RepoConfiguration` as with a `fail` field in `summary.json`.

**summary.json**
```
...
{
      "location": "https://github.com/reposense/testrepo-Delta.git",
      "organization": "reposense",
      "repoName": "testrepo-Delta",
      "branch": "nonExistentBranch",
      "displayName": "reposense_testrepo-Delta_nonExistentBranch",
      "sinceDate": "2017-09-01",
      "untilDate": "2017-10-30",    <--- In this case, some of the fields can be removed too.

      "fail": "Sorry. The supplied branch does not exists."  <--- Suggested approach.
}
...
```
In this approach, we can do a `RepoConfiguration#setFailureMessage(mesg)` on the `backend` and for the `frontend`, we just need to check if `fail` exists.

Also this way allow RepoSense to show error messages to users who are not familiar with `CLI` or `logs`.

damithc:
In general, try to have more logic in the backend and less in the frontend, on account of backend being easier to test compared to frontend code.

eugenepeh:
> In general, try to have more logic in the backend and less in the frontend, on account of backend being easier to test compared to frontend code.

Agree. However in this case, if we are going to display the message in `frontend`, I personally think it will be better for the `frontend` to handle it gracefully. 

Reason: 
- adding a constructor that we normally doesn't use adds on to the maintenance work which may cause more lead time
- GitDownloaderException and IOException were handled differently in this case which prevents IOException error message to be display on the dashboard. 
- This helps to cover more scenarios which helps to prevent more unexpected breakdowns

src/main/java/reposense/report/CommitReportJson.java
```diff
@@ -18,6 +20,27 @@
...
+        authorFinalContributionMap = new HashMap<>();
+        authorFinalContributionMap.put(emptyAuthor, 0);
+
+        authorContributionVariance = new HashMap<>();
+        authorContributionVariance.put(emptyAuthor, (float) 0.0);
+
+        authorDisplayNameMap = new HashMap<>();
+        authorDisplayNameMap.put(emptyAuthor, "FAILED TO CLONE OR CHECKOUT THIS REPOSITORY");
+    }
+
```
ongspxm: so are we gonna to remove this part? to actually write out a blank hashmap with no keys?

i can use that to detect when to show the "error" in the UI

eugenepeh: can refer to @yong24s description above. So the repos in summary.json will now have additional one more field to hold the message.

PR #303, opened 2018-08-17
hotfix. fix commit[0] undefined error

ongspxm:
```
there is some error, we wanna fix it
```

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -191,13 +191,15 @@ window.vSummary = {
       this.filtered.forEach((repo) => {
         repo.forEach((user) => {
           const { commits } = user;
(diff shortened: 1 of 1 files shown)


ongspxm:
@eugenepeh test pls

eugenepeh:
image

Seems like error still persist in current netlify.

ongspxm:
@eugenepeh looks like some browser don't return false for empty array, i am using `commits.length` instead. it should work for you now. give it a check.

eugenepeh:
image

seems like another bug. First commit and since date doesn't match

eugenepeh:
>seems like another bug. First commit and since date doesn't match

On second look, doesn't seems to be caused by this PR; seems like another legacy issue.

PR #306, opened 2018-08-17
Rename functional to system test #85

### LINKED ISSUE #85: Rename functional to system test
Functional folder is a skeleton of an end-to-end testing infrastructure.

However the name is not intuitive enough to imply it so.

Let's rename functional to system test instead.

yamidark:
Fixes #85 
```
Functional folder is a skeleton of an end-to-end testing infrastructure.

However the name is not intuitive enough to imply it so. Also, the functional
test is placed in a separate package of the 'src' module instead of the 'test' 
module, making IntelliJ unable to find the classes and resources to run it.

Let's rename functional to system test instead, and move it to the test
module as well.
```


yamidark:
Since 'functional/systemTests' is now inside `test`, `build` will also run the 'functional/systemTests' now.
Will do a cleanup of the `build.gradle` in another PR to 'separate' the 'unitTests' and 'systemTests' and other stuff.
We can now also finally run/debug the `SystemTest` now on IntelliJ!

yamidark:
Since this is a test class/package, I believe it should be placed inside the 'test' folder to define a clear line between the production code.

>I would prefer having a clear line, separating unit test and system test. Because in an edit-test cycle, the faster you receive your feedback, the more productive you are. Hence in most cases, developers only execute unit tests in that scenario.

Yes, I do plan to do this in the cleaning up of `build.gradle` in another PR, but I can bring it forward here if needed as well. Or, we could just place it in a different module under 'test' folder.

>Also, having their own resource folder, we can avoid the test and system test resources being used interchangeably which enhance the clear distinction.

Actually, this whole time the 'test' and 'system test resources' are built into the same folder, since the gradle configuration sets:
```
test {
    resources.srcDir file('src/functional/resources')
  }
```
instead of placing it inside 'functional' 💦 

>Was it because of poor gradle configuration which cause us unable to debug it on intellij?

Not exactly, the gradle configuration sets the classpath of 'functional' to include 'main' and 'test', and IntelliJ  is able to read it and thus not giving any code errors. However, for some reason while building, IntelliJ is unable to get the 'test' classpath and thus keeps failing at importing `TestUtil` (at least on my side).

eugenepeh:
> Since this is a test class/package, I believe it should be placed inside the 'test' folder to define a clear line between the production code.

True, however there are also practices where developers separate system test module from unit test to avoid reusing any of the test code and resources. 

If there is no additional benefit to incorporate everything into test module, let's keep it as it is.
We can consider the option when there is no other way to reactivate debugging for system test.

> Actually, this whole time the 'test' and 'system test resources' are built into the same folder, since the gradle configuration sets:

In this case, I think its more of a configuration issue rather than an architecture problem. Let's resolve it by cleaning up the build.gradle instead.

eugenepeh:
To merge after #295

yamidark:
@eugenepeh For your review again, rebased to master

PR #307, opened 2018-08-18
Code view: add controls to expand/collapse all files #36

### LINKED ISSUE #36: Code view: give a button to expand/collapse all
It can be useful to be able to expand/collapse all code panels

eugenepeh:
Fix #36 

```
Files in code view are shown in open state by default.

With the addition of #282, it can be useful to show only the file
headers as it provides a rough impression of the author's focus area
of contribution and pattern.

Let's add controls to allow user to expand or collapse all the files in
code view at their will.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -28,13 +28,10 @@ html
           .tab-content
             .tab-panes
               #tab-authorship.tab-pane(v-if="isTabAuthorship")
-                ul.authorship__toolbar
-                  li.dropdown
-                    a.dropdown-button
-                      i.fas.fa-lg.fa-bars
-                    .dropdown-content
-                      a(onclick="expandAll()") Expand all
-                      a(onclick="collapseAll()") Collapse all
+                .toolbar
+                  a(onclick="expandAll()") Expand all
+                  label &nbsp;/&nbsp;
+                  a(onclick="collapseAll()") Collapse all
                 v_authorship(
                   v-bind:key="generateKey(tabInfo.tabAuthorship)",
                   v-bind:info="tabInfo.tabAuthorship")
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -28,6 +28,13 @@ html
           .tab-content
             .tab-panes
               #tab-authorship.tab-pane(v-if="isTabAuthorship")
+                ul.authorship__toolbar
+                  li.dropdown
+                    a.dropdown-button
+                      i.fas.fa-lg.fa-bars
+                    .dropdown-content
+                      a Expand all
+                      a Collapse all
                 v_authorship(
                   v-bind:key="generateKey(tabInfo.tabAuthorship)",
                   v-bind:info="tabInfo.tabAuthorship")
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -33,8 +33,8 @@ html
                     a.dropdown-button
                       i.fas.fa-lg.fa-bars
                     .dropdown-content
-                      a Expand all
-                      a Collapse all
+                      a(onclick="expandAll()") Expand all
+                      a(onclick="collapseAll()") Collapse all
                 v_authorship(
                   v-bind:key="generateKey(tabInfo.tabAuthorship)",
                   v-bind:info="tabInfo.tabAuthorship")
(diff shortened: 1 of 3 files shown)


ongspxm:
Gud :+1:

damithc:
In my comp it appears as a hamberger menu in black.; feels like it takes too much attention/space. Can it follow a similar style to that of the control panel on the chart view? i.e., just a small button in a white background? We could be adding more controls such as sort/filter criteria in future, which might not fit a menubar.

eugenepeh:
### Normal:
image

### Hover:
image

How it supposed to look like, yup its a  hamberger menu in black.

image

^ I wanted to turn it into something like this in the future, so the search bar will be on the right end.
I make it looks like a navigation bar because adding a button will sort of takes up the entire horizontal space too (so might as well).

### Sticky to the top:
image

Furthermore, it looks nicer when sticky (just a personal gut feeling). 
Making it similar to the chart view might not looks nice to be on sticky, also not effective use of space.

I can try changing it to something else though, if you prefer something else other than the navigation bar design style.

damithc:
My guess is the control panels of the two panels can be similar -- some consistency between the two is not a bad thing after all. 
* They both can be sticky if we want but probably non-sticky is better to avoid them taking screen space permanently. But we can consider a minimized sticky version (e.g., an hamburger icon in the corner) so that the control panel can be accessed easily when it is out of screen.
* 'Appear on hover' is an idea we can consider in future, but again, for for both sides. 

For now, let's not go for the menu-bar style as it has certain limitations e.g., not suitable for tick-box controls which we might need later for various flags e.g., hide-untouched code

eugenepeh:
@damithc 

>They both can be sticky if we want but probably non-sticky is better to avoid them taking screen space permanently. But we can consider a minimized sticky version (e.g., an hamburger icon in the corner) so that the control panel can be accessed easily when it is out of screen.

In this case, would you prefer 
- fixed toolbar at the top + a floating `back to top` button (at bottom right of the screen)? 
- or simply a sticky toolbar?

> For now, let's not go for the menu-bar style as it has certain limitations e.g., not suitable for tick-box controls which we might need later for various flags e.g., hide-untouched code

For tick-box controls actually my plan is to do a dropdown list like the one below (to be more organized)
Let me know which do you prefer.

image

>My guess is the control panels of the two panels can be similar -- some consistency between the two is not a bad thing after all.

Similar is good for re-usability and maintainability which is a good thing. However, it can also make the design looks dull, boring and repetitive. 

Also, due to some nature of the 2 different view, some part of the design needs to have some significant difference. The reason for my current design was because:
1. The screen size is different. Code view's tool bar needs to be more compact for the smaller screen ratio.
2. Content length is different. Chart view often does not require as much scrolling as Code view does.

If you have a rough illustration or came across a design that you like, do let us know so that we can see how to fit that into the future planning and produce the ideal outcome.

damithc:
Let's discuss when me meet tomorrow.

frontend/src/static/js/v_authorship.js
```diff
@@ -15,6 +15,28 @@ function toggleNext(ele) {
...
+}
+
+function expandAll() {
+  const targetClass = 'file';
+  const renameValue = 'file active';
+
+  var files = document.getElementsByClassName(targetClass);
+  for (var i = 0; i < files.length; i++) {
+    if (files[i].className === targetClass) {
+      files[i].className = renameValue;
```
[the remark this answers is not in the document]
eugenepeh: > I would suggest writing just a generic toggleAll func then pass the initial and final className in during the calling. Up to you

Sounds good. Maybe I will just name it expandAll(true/false).

eugenepeh:
@damithc , I've made the change. Please let me know if any further changes needed.

image

frontend/src/static/js/v_authorship.js
```diff
@@ -15,6 +15,15 @@ function toggleNext(ele) {
   parent.className = classes.join(' ');
 };
 
+function expandAll(isActive) {
+  const renameValue = isActive ? 'file active' : 'file';
+
+  const files = document.getElementsByClassName('file');
+  Array.from(files).forEach((file) => {
```
ongspxm: i think you can just use `files.forEach` without the `Array.from`, if i ain't wrong.

frontend/src/index.jade
```diff
@@ -28,6 +28,10 @@ html
           .tab-content
             .tab-panes
               #tab-authorship.tab-pane(v-if="isTabAuthorship")
+                .toolbar
+                  a(onclick="expandAll(true)") Expand all
+                  label &nbsp;/&nbsp;
```
ongspxm: use a span instead, label by convention for inputs

PR #309, opened 2018-08-19
Clean up build.gradle

yamidark:
```
build.gradle contains some unused dependencies, inconsistent styles,
outdated version plugins etc.

Let's fix them.
```


yamidark:
@eugenepeh 

> I've tested it, it seems that debug still doesn't work with this clean up.

Hmm, the  running of functional/systemtest in IntelliJ? Yep, still doesn't work. The current `build.gradle` setup is the same as shown in the guide, and IntelliJ does show that we are depending on the `test` module, but somehow the `test` module is still not added to the compilation/runtime configurations, thus still unable to read `TestUtil`.
Anyway, I went to investigate into this more, and found [this solution](https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000045279-cannot-find-symbol-integration-test-class-depends-on-unit-test-class) which seems to work on my side. Could you try this at your side as well (pushed to upstream)?

>On a side note, I think we should rename myZip. Perhaps just zip alone will do?

We can be more specific, mention `zipReport` or `zipJson`?

yamidark:
>Perhaps its because of the sourceSets? Previously resources.srcDir was under test.

`resources.srcDir` under test just places the resouces in the test directory instead of functional, but the IntelliJ running/debugging problem is due to unable the `test` module not inside the classpath (even after we specify it in `build.gradle`)
I tested removing the `idea` plugin and it fails again. Most probably is just a bug in IntelliJ, maybe different versions will cause the issue? You can try reimporting your whole project and check if it works.

>I am thinking whether we should make it totally independent from unit test in future, e.g. we can duplicate compareFileContents to put in functional test.

Hmm, that would lead to duplicate code though, since the `TestUtil#compareDirectories(...)` which depends on `TestUtil#compareFileContents(...)` is being used by `GitChecker`, and we may require this method when adding the missing unit tests in the future though.

eugenepeh:
>resources.srcDir under test just places the resouces in the test directory instead of functional, but the IntelliJ running/debugging problem is due to unable the test module not inside the classpath (even after we specify it in build.gradle)
I tested removing the idea plugin and it fails again. Most probably is just a bug in IntelliJ, maybe different versions will cause the issue? You can try reimporting your whole project and check if it works.

In that case, let's just go with this PR. Main focus is to ensure it works for all developers anyway.

>Hmm, that would lead to duplicate code though, since the TestUtil#compareDirectories(...) which depends on TestUtil#compareFileContents(...) is being used by GitChecker, and we may require this method when adding the missing unit tests in the future though.

Hmm. I guess its fine for now as long as the dependency is one way.

eugenepeh:
@yong24s can you verify that debug on functional works for you as well?

yong24s:
Yes. It works! @yamidark 👍 🥇 🍪 

@eugenepeh I think you need to `File > Invalidate Caches/Restart...` for changes to take effect.
At least I had to do that before I can debug `Entry.java` tests.

build.gradle
```diff
@@ -1,32 +1,57 @@
...
 }
 
 dependencies {
-    implementation  group: 'commons-collections', name: 'commons-collections', version: '3.2'
-    implementation  'com.github.javaparser:javaparser-core:3.0.1'
-    implementation  'com.google.code.gson:gson:2.6.2'
-    implementation  group: 'org.json', name: 'json', version: '20180130'
-    implementation  group: 'net.sourceforge.argparse4j', name: 'argparse4j', version: '0.8.1'
+    implementation  group: 'com.github.javaparser', name: 'javaparser-core', version: '3.0.1'
+    implementation  group: 'com.google.code.gson' ,name: 'gson', version:'2.6.2'
```
yong24s: Should we also change to the latest `gson` version `2.8.2`?

https://github.com/google/gson/blob/master/CHANGELOG.md
**2.6.2** `2016-02-26`
**2.8.5** `2018-05-21`

build.gradle
```diff
@@ -1,32 +1,57 @@
...
+  functionalRuntime.extendsFrom testRuntime
 }
 
 dependencies {
-    implementation  group: 'commons-collections', name: 'commons-collections', version: '3.2'
-    implementation  'com.github.javaparser:javaparser-core:3.0.1'
-    implementation  'com.google.code.gson:gson:2.6.2'
-    implementation  group: 'org.json', name: 'json', version: '20180130'
-    implementation  group: 'net.sourceforge.argparse4j', name: 'argparse4j', version: '0.8.1'
+    implementation  group: 'com.github.javaparser', name: 'javaparser-core', version: '3.0.1'
```
yong24s: Note we may want to use a newer version **in due time** to support Java 10/11 syntax.
https://github.com/javaparser/javaparser/blob/master/changelog.md

build.gradle
```diff
@@ -37,46 +62,37 @@ shadowJar {
     destinationDir = file("${buildDir}/jar/")
 }
 
-task myZip(dependsOn: 'npmRunSpuild', type: Zip) {
+task zipReport(dependsOn: 'npmRunSpuild', type: Zip) {
     from 'frontend/build/'
     baseName = 'templateZip'
     destinationDir = file('src/main/resources')
 }
```
eugenepeh: In addition to the change suggested by @yong24s , 
let's remove this new line as well so that the task and the dependency appears as a module.

PR #314, opened 2018-08-25
code view: differentiate untouched code more #311

### LINKED ISSUE #311: code view: differentiate untouched code more
Now that we auto-show small untouched code, they can pass of as touched code (i.e., give the viewer an impression that the author changed more code than she actually did) because the grey line in the gutter is the only way to differentiate the two types of code.

image

Options:
* Go back to not showing untouched code by default even if they are small fragments
* Do more to distinguish between the two types of code options:
  * Use a grey font for untouched code (preferred)
  * Use a slighty grey background for untouched code
  * Use a slight green background for touched code

eugenepeh:
Fix #311 

```
Untouched code uses the same background and font color as written
code.

This allows the possibility of it being passed off as written code
(i.e., give the viewer an impression that the author changed more code
than she actually did) because the only way to differentiate the two
types of code is the color of the line in the gutter.

Let's differentiate the two code types more by adding a light green 
background to written code.
```

### DIFF
--- a/frontend/src/static/css/style.scss
+++ b/frontend/src/static/css/style.scss
@@ -252,7 +252,11 @@ header{
 
             &.active{
                 pre{
-                    display:block;
+                    display:grid;
+                    .hljs{
+                        // overwrite hljs library
+                        display:inherit;
+                    }
                 }
             }
         }
@@ -271,14 +275,16 @@ header{
         .code{
             padding-left:1rem;
             border-left:0.25rem solid mui-color('green');
+            background-color:mui-color('green', '50');
         }
 
         .untouched{
-            $gray: mui-color('grey', '400');
+            $grey:mui-color('grey', '400');
 
             &.active{
                 .code{
-                    display:block;
+                    display:inherit;
+                    background-color:mui-color('white');
                 }
 
                 .closer{
@@ -295,7 +301,8 @@ header{
 
             .code{
                 display:none;
-                border-left:0.25rem solid $gray;
+                border-left:0.25rem solid $grey;
+                background-color:mui-color('white');
             }
 
             .closer{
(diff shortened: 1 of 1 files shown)


damithc:
This looks better than before 👍 
What about the option of using one/two background colors instead? That way, we don't lose the syntax highlighting for untouched code.

eugenepeh:
> What about the option of using one/two background colors instead? That way, we don't lose the syntax highlighting for untouched code.

Because of the hljs library properties, background color may not be able extend to the entire block.
I can look into ways to do that; may need a significant change to existing structure.
So shall I use light green background color for written code and plain white background + grey font for untouched code? (because if both background and font is grey, the contrast wouldn't be that good)

As for the syntax highlighting, you prefer it to stay there for the untouched code?

frontend/src/static/css/style.scss
```diff
@@ -274,11 +274,15 @@ header{
...
-            $gray: mui-color('grey', '400');
+            $grey: mui-color('grey', '400');
 
             &.active{
                 .code{
                     display:block;
+                    color:$grey;
+                    [class*='hljs-']{
+                      color:inherit;
+                    }
```
ongspxm: dun need to write this, the css on line 302 will take care of it

eugenepeh: when calling the highlight function in authorship.js, it will add the hljs element which overwrites the parent's implementation. hence this is necessary.

but I can probably leave this code out if prof prefers the syntax highlight to remain on the untouched code.

damithc:
> So shall I use light green background color for written code and plain white background + grey font for untouched code? (because if both background and font is grey, the contrast wouldn't be that good)

Yes, that might work. No need for grey text for untouched code. 

> As for the syntax highlighting, you prefer it to stay there for the untouched code?

Yes, that will help a reader who is trying to understand the code, not merely trying to quantify the contribution.

eugenepeh:
@damithc 

> Yes, that might work. No need for grey text for untouched code.

Switched to adding just a green background color.

PR #322, opened 2018-08-28
Add repo as alias to repos in CLI 

### LINKED ISSUE #317: Support alias 'repo' for CLI parameter 'repos'
As most new users will use RS for one repo at the start, perhaps we should support a `repo` alias for the parameter `repos`. i.e., the first command they type can be
`java -jar RepoSense.jar -repo https://github.com/reposense/RepoSense.git`

Justification: the above command feels more intuitive. Besides, a user might miss the `s` in `repos` as they don't expect that to be plural in the context of this command.

yong24s:
Fixes #317 

```
RepoSense accepts repository locations with argument -repos.

The plurality is not intuitive for first-time users, who wish
to use one repository.

Let's add repo as an alias to repos to make the CLI more inituive.
```


src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -51,8 +51,9 @@ private static ArgumentParser getArgumentParser() {
                 .metavar("PATH")
                 .help("The directory containing the config files.");
 
-        mutexParser.addArgument("-repos")
+        mutexParser.addArgument("-repo", "-repos")
                 .nargs("+")
+                .dest("repos")
```
yong24s: `.dest("repos")` is added so that the value can be retrieved by `repos` key.

Prior to this change, it implicitly use the argument `-repos` as the key `repos` but it does not work well with `multiple` aliases.

damithc:
Hang on, there are no changes to docs?

docs/UserGuide.md
```diff
@@ -42,7 +42,7 @@ Argument List:
 - formats : Optional. The file formats to analyse. Formats: `alphanumerical file formats`. If not provided, the following file formats will be used. `adoc, cs, css, fxml, gradle, html, java, js, json, jsp, md, py, tag, xml`
 
 ### Using repository location(s)
-Usage: `java -jar RepoSense.jar -repos REPO_PATH... [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...]`
+Usage: `java -jar RepoSense.jar -repo/-repos REPO_PATH... [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...]`
```
eugenepeh: this change should be on line 53 instead.
Adding it in the usage adds unnecessary confusion.

yong24s:
@damithc 

> Adding all these equals method is related to this feature? If not, it should be a separate commit in the master branch, right?

The equals methods are used for testing.

Also, `User Guide` is updated to include `-repo` as an argument.

PR #326, opened 2018-08-31
Dashboard: tweak footer #321

### LINKED ISSUE #321: Dashboard: tweak footer
Current: 
image

Suggested:
* No need for `Last`
* Use smaller font and may be lighter color
* Center-align?
* Include a link to reposense
* Include version number if possible

e.g.,
>Generated by [**RepoSense** v1.3.1]() on {date/time}

ongspxm:
Fix #321 
```
Current footer doesn't contain much information. Let's include a link 
to reposense and include the version number.
````


eugenepeh:
@ongspxm 

If its not possible to do it with gradle, perhaps we can do fill it with json, similarly to how we do it for timestamp?

ongspxm:
but then it means the report must contain the value leh... means we will
need to make reposense get the value ah?

eugenepeh:
> but then it means the report must contain the value leh... means we will
need to make reposense get the value ah?

Wait... on a second thought, I think that wouldn't work because we will need the `tag` info from the `git` repo which is not found in jar release.

Gradle is still probably the best way to go about doing it.
Do you remember when I suggested that you encode a special constant in the file and replace only that constant in the file? (for the timestamp PR)
We can probably do that in this scenario. Custom a gradle task to run a java static method, read and overwrite the file before zipping it.

damithc:
We can leave out the version number for a future issue as it is not that important.

ongspxm:
roger tat, removing the version number from the footer

PR #327, opened 2018-09-01
README: update url to UserGuide

eugenepeh:
```
Documentation link in landing page points to the User Guide in
production.

This can result in users reading the User Guide that is ahead of
release, looking at features irrelevant to the jar they are using.

As users may not how to look for user guide pertaining to their
downloaded version, let's have the url in README to redirect user to
the latest released User Guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ The features of the report includes:
 * Filtering and sorting of authors
 
 ## Documentations
-* [**User Guide**](docs/UserGuide.md)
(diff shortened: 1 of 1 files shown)


eugenepeh:
@damithc I ended up decided to name the branch as `release` because it would be more intuitive to users who build their own release, to know that they are reading the wrong version of the user guide.

Do you think its fine?

damithc:
> Do you think its fine?

Yes, that's a good idea.

PR #339, opened 2018-09-18
Dashboard crashes when modify date to be out of range #336

### LINKED ISSUE #336: Dashboard crashes when modify date to be out of range
When we try to modify the date of the Since or Until filter in the dashboard,
if the date specified is out of range, like setting year 2020 for until date,
or any year for the since date (since typing the 1st number will set it to be that year 1st), 
the dashboard will hang, and then crash.
This can also occur while shifting the numbers using the arrow keys.

ongspxm:
fixes #336

```
When we try to modify the date of the Since or Until filter in the
dashboard, if the date specified is out of range, like setting year
2020 for until date, or any year for the since date (since typing the
1st number will set it to be that year 1st), the dashboard will hang,
and then crash.

Let's fix this by using a global minDate and maxDate, then checking if
the filter dates fit within the range.
```


yamidark:
Seems like the loading of the dashboard as I change the date will become longer and longer if the date range difference gets larger, and may still cause the thing to crash when it takes too long to load?

ongspxm:
Hahaha, sorry, not quite sure what you mean. Like a valid but very huge
date range might cause the thing to crash?

yamidark:
>Like a valid but very huge date range might cause the thing to crash?

Yep.
Basically, I can set the until year to be around around `2025` and the dashboard will start taking a noticeable amount of time to load, and if this date range keeps increasing, the amount of time to load increases, which may eventually cause the dashboard to crash(?)

ongspxm:
> Basically, I can set the until year to be around around 2025 and the dashboard will start taking a noticeable amount of time to load, and if this date range keeps increasing, the amount of time to load increases, which may eventually cause the dashboard to crash(?)

fixed

eugenepeh:
is it not possible to set the datepicker selectable range?

ongspxm:
Yeah, that be better. i go change it

eugenepeh:
image

seems like there's still some error, the browser hangs when i try to modify the date by typing and editing

eugenepeh:
> oh, looks like using the datepicker min and max only serves to invalid the date, doesn't actually change the value. looks like im gonna to rollback those changes and make use of vue to update the datepicker value instead.

The date picker is good, I think just the backend also need to verify as well.

ongspxm:
Manual edit? Like typing it in?



So a tempMinDate and tempMaxDate together with e actual minDate and
maxDate? It adds an additional layer of variables between the picker and
the actual vue object?

eugenepeh:
> Manual edit? Like typing it in?

Yup

> So a tempMinDate and tempMaxDate together with e actual minDate and
maxDate? It adds an additional layer of variables between the picker and
the actual vue object?

If it works, I'm ok with that.

eugenepeh:
image

Seems to work well. just wondering, how come only since date gets flag whereas until doesnt?

ongspxm:
Right, i skipped the check where untilDate > sinceDate. Okay, i will add it
in

PR #346, opened 2018-09-25
Fix inconsistent length of contribution bar #342

### LINKED ISSUE #342: Lone search result with no contribution shown as full-length bar
image

Steps:
Go to a dashboard (e.g., https://nus-cs2103-ay1819s1.github.io/cs2103-dashboard) and search for the name of an author who doesn't have any contribution yet. Choose the keyword in such a way only that author will appear in search results.

Expected: empty contribution bar

Actual: full-length contribution bar

eugenepeh:
Fix #342 

```
Length of the contribution bars varies, dependent on the set of
displayed authors.

This should not be the case. The scale used for the bar should be fixed
for a given period irrespective of which authors are displayed. Search
filters should simply filter out certain authors without changing
anything else.

Let's fixed this by performing the calculations on all authors'
calculation instead of filtered.
Also, since this value should not change in subsequent events, let's
only perform this calculation once and store the value in a global
constant.
```


frontend/src/static/js/v_summary.js
```diff
@@ -91,17 +91,20 @@ window.vSummary = {
...
     avgContributionSize() {
-      let totalLines = 0;
-      let totalCount = 0;
-      this.filtered.forEach((repo) => {
-        repo.forEach((user) => {
-          if (user.totalCommits > 0) {
-            totalCount += 1;
-            totalLines += user.totalCommits;
-          }
+      if (typeof meanContributionSize === 'undefined') {
```
ongspxm: why need to do this, cannot just do everything without the if and `return totalLines/totalCount;`?

eugenepeh: > why need to do this, cannot just do everything without the if and return totalLines/totalCount;?

To improve the response time. Since its a value that do not change, we should not compute the value every single instance.

ongspxm:
If that is the case, should write value into data, instead of using a vue
func, can chnage.

eugenepeh:
> If that is the case, should write value into data, instead of using a vue
func, can chnage.

Tested by putting an `alert(...);` into the if condition.
It doesn't activate more than once.

eugenepeh:
@damithc for your review.

PR #344, opened 2018-09-23
Reorganize UserGuide.md to improve the flow

damithc:
The main objective was to separate the parts that need to be read by cs2103 students from the rest. In the process, I tried to reduce duplication and forward referencing too -- i.e., the user guide can be read mostly from top to bottom.

Please check for problems, in particular, factual errors.

EDIT: Suggested commit message
```
In the User Guide, sections relevant to those using RepoSense on a single repo
are scattered.

Let's reorganize the flow to bring such sections to the top of the document.
At the same time, let us reduce forward referencing and duplication so that the
User Guide can be read mostly top to bottom in logical order.
```

Issue #341, opened 2018-09-22
If author info is not specified, show only those who contributed during analysis period


damithc:
Current: if author info is not specified, RepoSense shows all contributors for the entire history of the repo

Suggested: show only those who contributed during the analysis period as that is more likely to be the user's intention.

eugenepeh:
> Suggested: show only those who contributed during the analysis period as that is more likely to be the user's intention.

A little tricky here because a simple solution will be simply drop the authors who have no contribution at the end of the analysis. However because user can add authors who have zero contribution in the author-config, we can't do that.

I guess we can add an attribute to the Author model to indicate whether it is compulsory to include in the report, and drop the others who have no contribution afterward.

damithc:
> A little tricky here because a simple solution will be simply drop the authors who have no contribution at the end of the analysis. However because user can add authors who have zero contribution in the author-config, we can't do that.

This issue is about the case where no author config is provided.  In that case we can drop everyone who doesn't have a contribution? If the author config is provided, we have to show all authors mentioned irrespective of whether they have contributions.

damithc:
In a future version, we can also add a parameter to indicate if non-contributors should be listed.

eugenepeh:
Hmm... nvm, found an easier way to do this.

`git log --pretty=format:"%an" --after="2017-02-31" --until="2017-03-10"`

PR #347, opened 2018-09-25
Allow non-alpha-numeric characters in GitHub usernames #340

### LINKED ISSUE #340: Allow non-alpha-numeric characters in GitHub usernames
Error message:
```
12:41:22 - CS2103-AY1819S1-W10-2_main_master has no authors specified, using all authors by default.
Exception in thread "main" java.lang.IllegalArgumentException: The provided Git ID, Aiden Low (Yew Woei), contains illegal characters.
        at reposense.model.Author.<init>(Author.java:35)
        at reposense.git.GitShortlog.lambda$extractAuthorsFromLog$0(GitShortlog.java:24)
        at java.util.stream.ReferencePipeline$3$1.accept(Unknown Source)
        at java.util.Spliterators$ArraySpliterator.forEachRemaining(Unknown Source)
        at java.util.stream.AbstractPipeline.copyInto(Unknown Source)
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(Unknown Source)
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(Unknown Source)
        at java.util.stream.AbstractPipeline.evaluate(Unknown Source)
        at java.util.stream.ReferencePipeline.collect(Unknown Source)
        at reposense.git.GitShortlog.extractAuthorsFromLog(GitShortlog.java:26)
        at reposense.report.ReportGenerator.updateAuthorList(ReportGenerator.java:112)
        at reposense.report.ReportGenerator.generateReposReport(ReportGenerator.java:66)
        at reposense.RepoSense.main(RepoSense.java:51)

```

yamidark:
Fixes #340 
```
The Author class restricts the github id and author names to only contain
alphanumeric characters, plus some other symbols, for security purposes as
this values are used inside the OS terminal.

As we are allowed to change out git author name to be anything as it
does not have any restrictions, having this restrictions for our Author
class will prevent some users from using our app to analyze their repo
if the author name they wish to use contain non alphanumeric characters.

Also, special non-UTF-8 characters may also be used for the author name,
which may cause MalformedInputException[1] when trying to read the
author-config.csv file.

As the --author param field of the git log command uses regex pattern
to perform the author filtration, we can swap out any of these computer
symbols which may have different uses in the OS terminal into a . period
instead to indicate any character instead, allowing us to still get the correct
author filter without conflicting with the OS terminal. However, this may
cause the accuracy of the analysis to be reduced if there are multiple authors
with very similar names in the repo.

Let's
- remove validation of github id and names in the Author class
- add a method in StringUtil to swap computer symbols in the author
name to the . period
- add a note in the UserGuide warning users who have author names
that are very similar and have symbols
- update the CsvParser to use the java.io BufferedReader constructor for
reading of the csv files instead.

[1]: Similar issue when reading files previously resolved
https://github.com/reposense/RepoSense/commit/1f222c0fd93fd8c5399a27c18aa03c03871f7cee
```


yamidark:
@damithc Could you try testing this version on those repositories with the problems?

damithc:
Try running against markdown-here repo
I'll send you the repo list of the current batch to try.

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -24,7 +24,7 @@
     private static final String AUTHOR_NAME_PATTERN = "^%s <.*>$";
     private static final String OR_OPERATOR_PATTERN = "\\|";
 
-    private static final Pattern SPECIAL_REGEX_CHARS = Pattern.compile("[{}()\\[\\].+*?^$|]");
+    private static final Pattern SPECIAL_SYMBOLS = Pattern.compile("[;:&/\\\\!<>{}%#\"\\-='()\\[\\].+*?^$|]");
```
eugenepeh: can probably do like 
>!-/:-@[-\`

making use of range, but i'm okay with anything.

yamidark:
>perhaps can add unit test for replaceSpecialSymbols

Its a private helper method though, don't think we want to make it public just for testing. 
Have also already added 1 test for it with the `gitlog` command as well.

yamidark:
@eugenepeh Add the tests, and also added a note in UserGuide for users using names with computer symbols in them (will rebase after #344 is merged)

@damithc This version works for the markdown-here repo.
It also works for the current list of student repos, but the Chrome/Firefox browser isn't able to load all of it due to insufficient memory. I think this may be because we are finding all authors for the repo by not specifying a `author-config.csv`, so each of them will have the all the past contributors to AB4).

damithc:
> It also works for the current list of student repos, but the Chrome/Firefox browser isn't able to load all of it due to insufficient memory. I think this may be because we are finding all authors for the repo by not specifying a `author-config.csv`, so each of them will have the all the past contributors to AB4).

Thank for checking. This is why I requested for #341 :-)

yamidark:
>This is why I requested for #341 :-)

Will push a fix for it.

yamidark:
@eugenepeh Rebased, for your review again

docs/UserGuide.md
```diff
@@ -180,6 +180,8 @@ git config --global user.name "YOUR_AUTHOR_NAME”
 ```
 RepoSense expects the Git Author Name to be the same as author's GitHub username. If an author's `Git Author Name` is different from her `GitHub ID`, the `Git Author Name` needs to be specified in the standalone config file. If the author has more than one `Git Author Name`, multiple values can be entered too.
 
+> Note: As we use the OS terminal git command for analyse of repos, computer symbols (e.g. `"`, `!`, `/` etc.) inside the name will be replaced with a period `.` during analysis, which may reduce the accuracy of the analysis.
```
eugenepeh: I think a bit too much info. Perhaps just a `Note: using symbols such as... in your author name may reduce the accuracy of the analysis if 2 names are approximately similar`. will do

yamidark: Fixed

src/test/resources/StandaloneConfigTest/specialCharacterAuthor_config.json
```diff
@@ -0,0 +1,22 @@
...
+	[
+		{
+			"githubId": "‘Processed�‘Cooked�",
+			"displayName": "‘Processed�‘Cooked�",
+			"authorNames": ["‘Processed�‘Cooked�"]
+		},
+		{
+			"githubId": "Aiden Low (Yew Woei)",
+			"displayName": "Aiden Low (Yew Woei)",
+			"authorNames": ["Aiden Low (Yew Woei)"]
```
damithc: Best not to use real names

damithc:
Thanks for the fix @yamidark

PR #348, opened 2018-09-26
-output param does not allow nonexistent folder #345

### LINKED ISSUE #345: -output param does not allow nonexistent folder
In most software, the program will create the folder that user has specified if its found to be nonexistent, however this functionality was not extended to the -output param.

apoorva17:
Fixes #345

```
If the output path specified by the user does not exist, the program
exits with a 'folder not found' error message.

Let's automatically create the output folder for such cases
as it is more convenient to the user and is the norm in most software.
```

### DIFF
--- a/src/test/java/reposense/parser/ArgsParserTest.java
+++ b/src/test/java/reposense/parser/ArgsParserTest.java
@@ -267,18 +267,16 @@ public class ArgsParserTest {
         ArgsParser.parse(translateCommandline(input));
     }
 
-    @Test(expected = ParseException.class)
-    public void file_forOutputDirectory_throwsParseException() throws ParseException {
-        String file = PROJECT_DIRECTORY.resolve("parser_test.csv").toString();
-        String input = DEFAULT_MANDATORY_ARGS + String.format("-output %s", file);
-        ArgsParser.parse(translateCommandline(input));
-    }
-
-    @Test(expected = ParseException.class)
-    public void nonExistentDirectory_forOutputDirectory_throwsParseException() throws ParseException {
+    @Test
+    public void outputPath_nonExistentDirectory_success() throws ParseException, IOException {
         String nonExistentDirectory = PROJECT_DIRECTORY.resolve("some_non_existent_dir/").toString();
+        Path expectedRelativeOutputDirectoryPath = Paths.get(nonExistentDirectory)
+                .resolve(ArgsParser.DEFAULT_REPORT_NAME);
         String input = DEFAULT_MANDATORY_ARGS + String.format("-output %s", nonExistentDirectory);
-        ArgsParser.parse(translateCommandline(input));
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+        Assert.assertTrue(cliArguments instanceof ConfigCliArguments);
+        Assert.assertTrue(Files.isSameFile(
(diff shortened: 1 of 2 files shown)


apoorva17:
@eugenepeh there's a test case which does not allow filename such as parser_test.csv to be used as a folder name

eugenepeh:
> @eugenepeh there's a test case which does not allow filename such as parser_test.csv to be used as a folder name

If its allowed as a folder name, the most reasonable rationale will be, why not? 
We should try to be less restrictive whenever possible.

apoorva17:
> > As in, there was already a test case in the original implementation which does not allow filename such as parser_test.csv to be used as a folder name. If you are saying that it is allowed as a folder name, why was there such a test case present in the first place? What was the rationale behind the existing test case?
> 
> I see, in that case it was probably something implemented long ago or we've missed it.
> Can just remove that test and fix it accordingly.

Ok, i've removed that test, ready for review.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -267,18 +267,16 @@ public void parse_missingConfigValue_throwsParseException() throws ParseExceptio
...
-    public void file_forOutputDirectory_throwsParseException() throws ParseException {
-        String file = PROJECT_DIRECTORY.resolve("parser_test.csv").toString();
-        String input = DEFAULT_MANDATORY_ARGS + String.format("-output %s", file);
-        ArgsParser.parse(translateCommandline(input));
-    }
-
-    @Test(expected = ParseException.class)
-    public void nonExistentDirectory_forOutputDirectory_throwsParseException() throws ParseException {
+    @Test
+    public void nonExistentDirectory_forOutputDirectory_success() throws ParseException, IOException {
```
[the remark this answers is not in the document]
apoorva17: test method name has been updated

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -267,18 +267,16 @@ public void parse_missingConfigValue_throwsParseException() throws ParseExceptio
...
-    public void file_forOutputDirectory_throwsParseException() throws ParseException {
-        String file = PROJECT_DIRECTORY.resolve("parser_test.csv").toString();
-        String input = DEFAULT_MANDATORY_ARGS + String.format("-output %s", file);
-        ArgsParser.parse(translateCommandline(input));
-    }
-
-    @Test(expected = ParseException.class)
-    public void nonExistentDirectory_forOutputDirectory_throwsParseException() throws ParseException {
+    @Test
+    public void nonExistentDirectory_forOutputDirectory_createDirectory() throws ParseException, IOException {
```
eugenepeh: misaligned to @yamidark suggested nit

>Slight nit:
outputPath_nonExistentDirectory_success()
The naming convention for test methods we use is featureUnderTest_testScenario_expectedBehaviour()

eugenepeh: Here our 
featureUnderTest -> `outputPath
testScenario -> `nonExistentDirectory`
expectedBehaviour/outcome -> `success`

apoorva17: ok, test method name has been updated accordingly

apoorva17:
@damithc commit message has been updated!

PR #355, opened 2018-10-02
 granularity#week: fix url not using correct time gap #354

### LINKED ISSUE #354: granularity#week: url link should show contribution for entire week
When the granularity is set to `week` in report, the url link is still using a single day time gap.
This should be updated accordingly to reflect the commits done in the week.

eugenepeh:
Fix #354 

```
When the granularity is set to week in report, the url link is still
using a single day time gap.

This should be updated accordingly to show users the commits that were
done in the week.

Let's update the url function to do a conditional check prior to the
generation of hyper link.
```

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -34,6 +34,11 @@ function dateRounding(datestr, roundDown) {
   return getDateStr(datems);
 }
 
+function addDays(dateStr, numDays) {
+  const date = new Date(dateStr);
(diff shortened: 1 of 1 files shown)


eugenepeh:
> Very good. But if i never use it anywhere else, i would probably do a addWeek instead of a addDay(6)

Hmm... but 6 days doesn't actually make up one week, which is why i named the function `addDays`. If I name it `addWeek` but uses 6 days in the backend, it may appear as a typo instead.

PR #356, opened 2018-10-04
Fix error occurred from file change due to permission bits modification

eugenepeh:


src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -97,9 +99,10 @@
         String[] fileDiffResultList = fullDiffResult.split(DIFF_FILE_CHUNK_SEPARATOR);
 
         for (String fileDiffResult : fileDiffResultList) {
-            // file deleted, renamed, is binary file or is new and empty file, skip it
+            // file deleted, renamed, permission changed, is binary file or is new and empty file, skip it
             if (fileDiffResult.contains(SIMILAR_FILE_RENAMED_SYMBOL) || fileDiffResult.contains(BINARY_FILE_SYMBOL)
-                    || FILE_DELETED_PREDICATE.test(fileDiffResult) || NEW_EMPTY_FILE_PREDICATE.test(fileDiffResult)) {
+                    || FILE_DELETED_PREDICATE.test(fileDiffResult) || NEW_EMPTY_FILE_PREDICATE.test(fileDiffResult)
+                    || FILE_PERMISSION_CHANGED_PREDICATE.test(fileDiffResult)) {
```
yamidark: Seems like we have alot of checks for no file change `diff` =O.
Maybe we can abstract this out to another method `isValidDiffFile`?

PR #359, opened 2018-10-06
 Report#datepicker: default value not filled on initial load #358

### LINKED ISSUE #358: Report#datepicker: default value not filled on initial load
image

ongspxm:
Fix #358 

```
The default value of the datepicker is not set on load.

Let's set the default value to minimum and maximum dates of the generated report.
```


eugenepeh:
@ongspxm thanks for the fixed, still have some bugs though,

When we load with the hashparam, it seems that the default values were not reflected as well.
https://deploy-preview-359--reposense.netlify.com/#=undefined&search=&sort=displayName&since=2017-10-28&until=2018-10-05&timeframe=day&reverse=false&repoSort=true

eugenepeh:
@ongspxm seems like this change has brought back the original problem. It hanged 🤣

eugenepeh:
> wait to clarify, when does the hanging seem to take place? after the date is changed? cos i dun seem to get any lag 🤔

When try to manually type the dates

eugenepeh:
@ongspxm 

image

Still have a little bug, but most of the concerned issues seem to be resolved.

ongspxm:
@eugenepeh done

Issue #360, opened 2018-10-06
java.lang.AssertionError: Should not have error matching file path pattern inside file diff result


damithc:
v1.5 fails with the following message
Env: Windows 10, for [this config](https://github.com/nus-cs2103-AY1819S1/cs2103-dashboard/tree/master/config)

```
07:40:10 - Extracting relevant file infos https://github.com/CS2103-AY1819S1-F11-4/main.git...
07:40:10 - Failed to match the file path group for:
diff --git a/.attach_pid11612 b/.attach_pid11612
new file mode 100644
index 00000000..e69de29b
Exception in thread "main" java.lang.AssertionError: Should not have error matching file path pattern inside file diff result!
        at reposense.authorship.FileInfoExtractor.getFilePathFromDiffPattern(FileInfoExtractor.java:233)
        at reposense.authorship.FileInfoExtractor.getEditedFileInfos(FileInfoExtractor.java:106)
        at reposense.authorship.FileInfoExtractor.extractFileInfos(FileInfoExtractor.java:76)
        at reposense.authorship.AuthorshipReporter.generateAuthorshipSummary(AuthorshipReporter.java:21)
        at reposense.report.ReportGenerator.generateReposReport(ReportGenerator.java:69)
        at reposense.RepoSense.main(RepoSense.java:51)
```

PR #361, opened 2018-10-06
FileInfoExtractor: explore cleaner alternative to extract code change #357

### LINKED ISSUE #357: FileInfoExtractor: explore cleaner alternative to extract code change
#356 has broke out recently due to some out of ordinary scenario that was not covered in the `FileInfoExtractor`. We should try to avoid such incidents by exploring cleaner options. 

E.g. instead of blacklisting, we should use white list instead.

yamidark:
Fixes #357 
```
FileInfoExtractor#getEditedFileInfos(...) first checks for each file
whether they in a blacklist of 'uncommon' file changes which does not
follow the line patterns that we check for, and skips those files as they
are all unchanged files that does not have any difference to check for.

As there are many of such 'uncommon' file changes, some of these changes
may not have been added in our blacklist yet, which may cause the program
to crash unexpectedly, e.g #215 #300 #356.

Instead of trying to find out each of this 'uncommon' file changes and
blacklisting each 1 of them, it is easier and cleaner to whitelist only
those files that have valid diff file changes which follow the line patterns
we are checking for.

Let's change the blacklist for valid diff files to be a whitelist instead,
and check for the exception if this file is deleted.
```

### DIFF
--- a/src/main/java/reposense/authorship/FileInfoExtractor.java
+++ b/src/main/java/reposense/authorship/FileInfoExtractor.java
@@ -36,13 +36,14 @@ public class FileInfoExtractor {
     private static final String LINE_INSERTED_SYMBOL = "+";
     private static final String STARTING_LINE_NUMBER_GROUP_NAME = "startingLineNumber";
     private static final String FILE_CHANGED_GROUP_NAME = "filePath";
+    private static final String FILE_DELETE_SYMBOL = "/dev/null";
     private static final String MATCH_GROUP_FAIL_MESSAGE_FORMAT = "Failed to match the %s group for:\n%s";
 
     private static final int LINE_CHANGED_HEADER_INDEX = 0;
 
     private static final Pattern STARTING_LINE_NUMBER_PATTERN = Pattern.compile(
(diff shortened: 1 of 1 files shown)


yamidark:
>Is it possible to check whether there are any positive line change instead of looking for the LINE_CHUNKS_SEPARATOR tag?

For positive line change, we can check for the `+` sign at the start of the line.
However, this conflicts with the `+++ b/file/path` pattern above, hence I switched to using ` LINE_CHUNKS_SEPARATOR` which also ensures there are line changes in the file.
Perhaps we can instead search for `+++ b/file/path` pattern (instead of throwing the `AssertionError` if we can't find it), since this marker would only appear if the file has line changes too (would still need to separately handle a deleted file though).

yamidark:
@damithc Tried the config in #360 (`-since 01/08/2018`), fixed with this PR as well.

eugenepeh:
As discussed in chat:
I've added the error case to the testrepo-Alpha, need to update the following to pass the test case:
GitShortlogTest -> Line 24 :
       expectedAuthorList.add(new Author("Eugene Peh"));
GitTestTemplate -> Line 35:
       hash change to b28dfac5bd449825c1a372e58485833b35fdbd50

PR #364, opened 2018-10-10
Make the commits info analysis case insensitive #363

### LINKED ISSUE #363: Make the commits info analysis case insensitive
In certain reports, some authors do not have any ramps shown in their ramp chart, even though their code view shows line contributions.

This is because unlike authorship analyzer, commits info analysis uses case sensitive comparison.

eugenepeh:
Fix #363 

```
CommitInfoExtractor uses git log to extract users' commits info.
However, by default, the author matching pattern used in git log is
case sensitive.

This can cause the CommitInfoAnalyzer to miss out some or all their
commits in the analysis. As a result, some authors do not have any
ramps shown in their ramp chart, even though their code view shows line
contributions.

To resolve this problem, let's set the git log regex patterns to be
case insensitive.
```


eugenepeh:
image

PR #368, opened 2018-10-12
Fix NullPointerException when accessing empty displayName in StandaloneConfig #366

### LINKED ISSUE #366: NullPointerException occurs when utilizing remote configurations
image

yong24s:
Fixes #366

```
The display name of Author in StandaloneConfig is optional.
Thus, it is null when not provided.

However, no null check was performed when using display name,
which may result in a NullPointerException.

Let's fix this by returning an empty string instead of null as
the default value of display name and including unit tests for
StandaloneConfig parser.
```


src/test/java/reposense/parser/StandaloneConfigJsonParserTest.java
```diff
@@ -0,0 +1,96 @@
...
+        Assert.assertEquals(Collections.emptyList(), config.getIgnoreGlobList());
+        Assert.assertEquals(Collections.emptyList(), config.getIgnoreCommitList());
+        Assert.assertEquals(ArgsParser.DEFAULT_FORMATS, config.getFormats());
+        Assert.assertEquals(1, config.getAuthors().size());
+
+        StandaloneAuthor author = config.getAuthors().get(0);
+        Assert.assertEquals("yong24s", author.getGithubId());
+        Assert.assertEquals("", author.getDisplayName());
+        Assert.assertEquals(Collections.emptyList(), author.getAuthorNames());
+        Assert.assertEquals(Collections.emptyList(), author.getIgnoreGlobList());
```
eugenepeh: Perhaps we create the expected RepoConfig and do the comparison in an extract method?

It helps to prevent the developers from missing out any fields in the comparison.

src/test/java/reposense/parser/StandaloneConfigJsonParserTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+    private static final Path STANDALONE_MALFORMED_CONFIG = new File(
+            StandaloneConfigJsonParserTest.class.getClassLoader()
+            .getResource("StandaloneConfigJsonParserTest/standaloneConfig_malformedJson.json").getFile()).toPath();
+
+    private static final Path STANDALONE_UNKNOWN_PROPERTY_CONFIG = new File(
+            StandaloneConfigJsonParserTest.class.getClassLoader().getResource(
+                    "StandaloneConfigJsonParserTest/standaloneConfig_unknownPropertyInJson.json").getFile()).toPath();
+
+
```
eugenepeh: unnecessary extra line here.

src/test/java/reposense/parser/StandaloneConfigJsonParserTest.java
```diff
@@ -0,0 +1,107 @@
...
+        StandaloneConfig config = new StandaloneConfigJsonParser().parse(STANDALONE_CONFIG_GITHUBID_ONLY);
+        isExact(EXPECTED_GITHUBID_ONLY_REPOCONFIG, config);
+    }
+
+    @Test(expected = JsonSyntaxException.class)
+    public void standaloneConfig_malformedJsonFile_throwsIoException() throws IOException {
+        new StandaloneConfigJsonParser().parse(STANDALONE_MALFORMED_CONFIG);
+    }
+
+    private void isExact(RepoConfiguration expectedRepoConfig, StandaloneConfig actualStandaloneConfig)
```
eugenepeh: Not sure if `isExact` sounds correct because exact usually refers to 2 reference pointing to the same object.
Also, since this method does not return a boolean result, we probably shouldn't prefix it with an `is`.

Perhaps `assertSame` would be better to imply the duplication and internal assertion.

eugenepeh:
Amendment to the commit message:
```
In our user guide, we allow most of the fields under the authors'
configuration to be optional other than their GitHub Id. Particularly
`Display Name`, which is suppose to use GitHub Id value when it is not
provided. However, when the display name is not filled in the
standalone config, the program crashes due to a NullPointerException.

This is because standalone config adopts the json format. When a field
is not filled in, in the json file, it will be initialized as null when
deserialized into a java object. This was unforeseen and was left
unchecked when accessing the variable.

Let's fix this by returning an empty string as the default value of
display name in the event of a null case, and also include unit tests
for StandaloneConfig parser.
```

PR #371, opened 2018-10-14
InvalidPathException thrown during run time due to Illegal char <:> in file name #370

### LINKED ISSUE #370: InvalidPathException thrown during run time due to Illegal char <:> in file name
image

yamidark:
Fixes #370 
```
Windows disallows certain characters for the file names, such as `/\:*?"<>|`,
while other OSes like Linux allows almost any characters for their file names.
This may result in some files with such `banned` characters in the repository
to not be copied/cloned into a local copy for Windows OS.

As we use the `git diff` command to find file differences for analyzing
repositories with a date range, such files with `banned` characters may appear
in the `git diff` result, which result in an InvalidPathException being thrown
when we try to create a Path to this file name, which is invalid in Windows.
This causes the program to crash as the exception is not handled anywhere.

Let's update FileInfoExtractor to handle files with invalid file names by
skipping those files.
```

### DIFF
--- a/src/main/java/reposense/authorship/FileInfoExtractor.java
+++ b/src/main/java/reposense/authorship/FileInfoExtractor.java
@@ -106,17 +106,13 @@ public class FileInfoExtractor {
             }
 
             if (isFormatInsideWhiteList(filePath, config.getFormats())) {
-                FileInfo currentFileInfo;
-
                 try {
-                    currentFileInfo = generateFileInfo(config.getRepoRoot(), filePath);
+                    FileInfo currentFileInfo = generateFileInfo(config.getRepoRoot(), filePath);
+                    setLinesToTrack(currentFileInfo, fileDiffResult);
+                    fileInfos.add(currentFileInfo);
                 } catch (InvalidPathException ipe) {
                     logger.warning(String.format(INVALID_FILE_PATH_MESSAGE_FORMAT, filePath));
-                    continue;
                 }
-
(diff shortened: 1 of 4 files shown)


src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -104,7 +106,15 @@
...
+                try {
+                    currentFileInfo = generateFileInfo(config.getRepoRoot(), filePath);
+                } catch (InvalidPathException ipe) {
+                    logger.warning(String.format(INVALID_FILE_PATH_MESSAGE_FORMAT, filePath));
+                    continue;
+                }
+
                 setLinesToTrack(currentFileInfo, fileDiffResult);
                 fileInfos.add(currentFileInfo);
             }
```
yong24s: Not sure if the `continue` in the `catch` block is necessary.

Try simplify it like this?

```
if (isFormatInsideWhiteList(filePath, config.getFormats())) {    
    try {
        // generateFileInfo may throw InvalidPathException 
        FileInfo currentFileInfo = generateFileInfo(config.getRepoRoot(), filePath);
        setLinesToTrack(currentFileInfo, fileDiffResult);
        fileInfos.add(currentFileInfo);
    } catch (InvalidPathException ipe) {
        logger.warning(String.format(INVALID_FILE_PATH_MESSAGE_FORMAT, filePath));
    }
}
```

Issue #379, opened 2018-10-20
Use copy instead of clone for local repository


eugenepeh:
As people tend to perform trial and error for the json config file before committing to the repository. However, the json config file will not be included when we perform the clone on the local repository.

This means that we should probably use copy instead of clone for local repository.

Note: along with this change, unstaged files will also be carried along to the analysis, we probably need to include the `-f` flag for the checkout command for this change.

damithc:
Oh, I didn't realize this. That means our instructions are wrong. Their new json file is not used (unless it is committed) when they run RepoSense locally. Can we do a quick fix?

damithc:
If it is going to take time to fix, we might have to update the user guide to ask users to commit changes (possibly in a branch) when they are testing RepoSense. The next week project progress recommendation specifically asks students to make their code RepoSense-compatible.

yamidark:
>This means that we should probably use copy instead of clone for local repository.

Did some tests with this, some problems with copying all the files instead of clone, is that we will copy all the files inside `.gitignore` as well. As `FileInfoExtractor#getAllFileInfo(...)` will generate all `FileInfos` through traversing the directory, the app will crash if some of those files that has their file formats is inside our whitelist, but we are unable to run 'git blame' analysis on them.

One solution for this would be to like `FileInfoExtractor#getEditedFileInfo(...)`, for `FileInfoExtractor#getAllFileInfo(...)` should also get all files to be tracked and analyzed by using 'git' commands instead. This will probably take quite some time to research and implement, so for now might just want to update the UserGuide 1st.

damithc:
Perhaps we should abort this approach and ask users to commit changes (possibly in a separate branch) before running RepoSense locally?

eugenepeh:
>Perhaps we should abort this approach and ask users to commit changes (possibly in a separate branch) before running RepoSense locally?

Actually they can do `git commit --amend` when doing trial and error with the `config.json`, just that I'm not sure if students know how to do that in source tree.

damithc:
> Actually they can do `git commit --amend` when doing trial and error with the `config.json`, just that I'm not sure if students know how to do that in source tree.

Yes. Using a separate branch (something they should know how to do) is the worst case scenario.

damithc:
> Let's do this quickly while we look for an alternative?

[...]

PR #381, opened 2018-10-21
UserGuide: update json Config File instructions

eugenepeh:
```
As people tend to perform trial and error for the json config file
before committing to the repository. However, the json config file will
not be included when we perform the clone on the local repository.

Let's add the instructions to request user to commit the json config
file.
```

temporary fix for #379


docs/UserGuide.md
```diff
@@ -116,7 +116,7 @@ Repo owners can provide the following additional information to RepoSense using
 * which git and github usernames belong to which authors
 * the display of an author
 
-To use this feature, add a `_reposense/config.json`  to the root of your repo using the format in the example below ([another example](../_reposense/config.json)):
+To use this feature, add a `_reposense/config.json`  to the root of your repo using the format in the example below ([another example](../_reposense/config.json)) and **commit it**:
```
damithc: `and **commit it** (reason: RepoSense can see committed code only):`?

damithc: Should we mention this under `@@author` section too?

eugenepeh: Hmm... but its not really relevant because `@@author` can typically work with/without json file.

damithc: But the tags will not affect the results unless they are committed first, right?

eugenepeh: oh... okay

PR #384, opened 2018-10-24
Code panel: make the expand/collapse button a toggle button #376

### LINKED ISSUE #376: Code panel: make the expand/collapse button a toggle button
Current:
image

Suggested: make it one link showing only `Expand all` or `Collapse all` (whichever the one that is releveant)

Justification: to be consistent with how usually expand/collapse buttons work

apoorva17:
Fixes #376 

```
Code panel shows both Expand and Collapse buttons.

This is not consistent with how expand/collapse buttons normally work.

Let's make it a single link and only display the Expand or Collapse
button, whichever one is relevant.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -32,10 +32,9 @@ html
           .tab-content
             .tab-panes
               #tab-authorship.tab-pane(v-if="isTabAuthorship")
-                .toolbar
-                  a(onclick="expandAll(true)") Expand all
-                  span &nbsp;/&nbsp;
-                  a(onclick="expandAll(false)") Collapse all
+                .toolbar(v-if="tabInfo.tabAuthorship.totalCommits > 0")
+                    a(v-if="isCollapsed", v-on:click="expand(true)") Expand all
+                    a(v-else, v-on:click="expand(false)") Collapse all
                 v_authorship(
                   v-bind:key="generateKey(tabInfo.tabAuthorship)",
                   v-bind:info="tabInfo.tabAuthorship")
@@ -90,8 +89,9 @@ html
             .summary-chart(v-for="(user, i) in repo")
               .summary-chart__title(
                 title="click to view author's code",
-                v-on:click="$emit('view-authorship', "
-                    + "{author:user.name, repo:user.repoName, name:user.displayName, minDate:minDate, maxDate:maxDate})"
+                v-on:click="$emit('view-authorship', \
+                    {author:user.name, repo:user.repoName, name:user.displayName, \
+                        minDate:minDate, maxDate:maxDate, totalCommits:user.totalCommits})"
               )
                 .summary-chart__title--index {{ i+1 }}
                 .summary-chart__title--repo(v-if="!filterGroupRepos") {{ user.repoPath }}
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -23,6 +23,7 @@ window.app = new window.Vue({
     loadedRepo: 0,
     userUpdated: false,
 
+    isCollapsed: false,
     isTabActive: false,
     isTabAuthorship: false,
     tabInfo: {},
@@ -82,6 +83,13 @@ window.app = new window.Vue({
 
       this.isTabActive = true;
       this.isTabAuthorship = true;
+      this.isCollapsed = false;
+    },
+
+    /*global expandAll*/
+    expand(isActive) {
+      this.isCollapsed = !isActive;
+      expandAll(isActive);
     },
 
     generateKey(dataObj) {


apoorva17:

frontend/src/index.jade
```diff
@@ -90,7 +90,8 @@ html
             .summary-chart(v-for="(user, i) in repo")
               .summary-chart__title(
                 title="click to view author's code",
-                v-on:click="$emit('view-authorship', {author:user.name, repo:user.repoName, name:user.displayName})"
+                v-on:click="$emit('view-authorship', " +
+                "{author:user.name, repo:user.repoName, name:user.displayName, totalCommits:user.totalCommits})"
```
eugenepeh: `+` operator should be on this line, also line wrap should include double the indentation from previous line.

eugenepeh:
> @eugenepeh is it necessary to make the button appear after code view is generated? this is because the Collapse All button is not clickable until the code view is generated anyways, and it would be an unnecessary exchange of information between the code panel component and parent component which can be avoided.

Hmm I've managed to clicked it several times though. Anyway I think this should put on hold for a while because if we going to introduce the search bar, which is going to be one of the highest priority, there will be some relocation and changes. Because of that, I don't think it is worth maintaining this for the time being. 

@damithc do you think it is reasonable to put this on hold?

damithc:
> @damithc do you think it is reasonable to put this on hold?

I guess it is better to finish this. Search bar is unlikely to happen in this semester.

As @eugenepeh mentioned, it is better if the button can appear only after the code is loaded. But that can be done as a separate feature, possibly after the search bar is implemented.

eugenepeh:
hmm okay.

@apoorva17 can just fix the coding standard and request for the next iteration.

apoorva17:
@eugenepeh coding standard fixes have been made

frontend/src/index.jade
```diff
@@ -90,7 +90,8 @@ html
             .summary-chart(v-for="(user, i) in repo")
               .summary-chart__title(
                 title="click to view author's code",
-                v-on:click="$emit('view-authorship', {author:user.name, repo:user.repoName, name:user.displayName})"
+                v-on:click="$emit('view-authorship', "
+                    + "{author:user.name, repo:user.repoName, name:user.displayName, totalCommits:user.totalCommits})"
```
eugenepeh: why is this change necessary?

frontend/src/static/js/main.js
```diff
@@ -23,6 +23,7 @@ window.app = new window.Vue({
     loadedRepo: 0,
     userUpdated: false,
 
+    isCollapsed: false,
```
eugenepeh: what if we let it be null until its loaded? that way we should be able to make the toggle appears only when its ready isnt it?

frontend/src/index.jade
apoorva17: @eugenepeh this change ensures that the button is not displayed when there are no files i.e. total contribution = 0.

frontend/src/static/js/main.js
apoorva17: in order for to set the value to null until loaded, i need to retrieve the information from the code panel component on whether the code view is loaded. Anyways, as mentioned in the above discussion, this can be done as a separate feature.

apoorva17: @eugenepeh ```this.isCollapsed = false``` is necessary to ensure that the button is reset to ```Collapse all``` when the code panel is closed and re-opened for another author. Otherwise, if the button was ```Expand all``` when the code panel is closed, when it is opened again for another author, button remains as ```Expand all``` and is not reset to ```Collapse all```.

eugenepeh: I see. Good point.

eugenepeh:
No more comments from me. 
@ongspxm can you please help to take a quick look? Thanks!

apoorva17:
@eugenepeh can this PR be merged and closed? or are there any other changes/review required?

frontend/src/index.jade
ongspxm: do it like this instead:

```
.toolbar(v-if="tabInfo.tabAuthorship.totalCommits > 0")
    a(v-on:click="isCollapsed=false", onclick="expandAll(true)") Expand all
    a(v-on:click="isCollapsed=true", onclick="expandAll(false)") Collapse all
```

frontend/src/index.jade
```diff
@@ -32,10 +32,10 @@ html
...
             .tab-panes
               #tab-authorship.tab-pane(v-if="isTabAuthorship")
-                .toolbar
-                  a(onclick="expandAll(true)") Expand all
-                  span &nbsp;/&nbsp;
-                  a(onclick="expandAll(false)") Collapse all
+                .toolbar(v-if="isCollapsed && tabInfo.tabAuthorship.totalCommits > 0")
+                  a(onclick="expandAll(true)", v-on:click="isCollapsed=false") Expand all
+                .toolbar(v-if="!isCollapsed && tabInfo.tabAuthorship.totalCommits > 0")
+                  a(onclick="expandAll(false)", v-on:click="isCollapsed=true") Collapse all
```
ongspxm:
```
.toolbar(v-if="tabInfo.tabAuthorship.totalCommits > 0")
    a(v-on:click="isCollapsed=false", onclick="expandAll(true)") Expand all
    a(v-on:click="isCollapsed=true", onclick="expandAll(false)") Collapse all
```

apoorva17: @ongspxm If ```isCollapsed``` is defined in ```v_authorship.js``` instead of ```main.js```, how do i retrieve that information in toolbar section?

apoorva17:
@ongspxm requested changes have been made

eugenepeh:
Tweak to the proposed merge message
```
Code panel shows both Expand All and Collapse All buttons.

This is inconsistent with how expand/collapse buttons normally work.

Let's replace it with a toggle button that alternates between the two.
```

eugenepeh:
Good work @apoorva17, do try working a higher priority issue on your next PR.

PR #401, opened 2018-11-12
Fix invalid file path occurrence in git log

eugenepeh:


eugenepeh:
>`A pathspec that begins with a colon : has special meaning.
>
>In the short form, the leading colon : is followed by zero or more magic signature letters (which optionally is terminated by another colon :), and the remainder is the pattern to match against the path.`
the `/` is some kind of magic signature letter apparently, so it also get consumed, then result in relative path which worked

Quoted from @yamidark

Issue #409, opened 2018-12-08
Gradle task npmRunSpuild is not working as intended


eugenepeh:
@ongspxm , seems like you've recently updated your spuild library and the npmRunSpuild task in gradle is not working the same as before. Could you please fix it?

ongspxm:
paiseh, didnt thought it will affect, okay i go now

eugenepeh:
@ongspxm thanks! do close this issue when its fixed.

eugenepeh:
Thanks @ongspxm !

PR #410, opened 2018-12-08
Gradle task npmRunSpuild is not working as intended #409

ongspxm:
```
seems like after the recent update on the spuild library, the npmRunSpuild
task in gradle is not working the same as before.

Let's fix the version of spuild to use for the rendering of the dashboard.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "homepage": "https://github.com/reposense/RepoSense#readme",
   "dependencies": {
-    "spuild": "^1.0.2"
+    "spuild": "1.0.2"
   },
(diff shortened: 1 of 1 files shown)


ongspxm:
can help restart build pls?

eugenepeh:
Minor nits to the proposed message:

```
The recent update on the spuild library has caused the gradle task,
npmRunSpuild, to deviate from our expected functionality and no longer
building our frontend files as before.

Let's fix this by removing the caret sign in the package.json to retain
the last working version of spuild we used for the rendering of the
dashboard.
```

PR #412, opened 2018-12-10
Make StandaloneConfig foolproof to exceptions caused by trailing commas #402

### LINKED ISSUE #402: Make StandaloneConfig foolproof from NullPointerException
```Exception in thread "main" java.lang.NullPointerException
        at reposense.model.Author.<init>(Author.java:34)
        at reposense.model.RepoConfiguration.update(RepoConfiguration.java:152)
        at reposense.report.ReportGenerator.updateRepoConfig(ReportGenerator.java:98)
        at reposense.report.ReportGenerator.generateReposReport(ReportGenerator.java:68)
        at reposense.RepoSense.main(RepoSense.java:51)

```
image

When the a json list is terminated with a `,`, it introduce an additional null element to the list when we load it back as a java object. This leads to an unhandled NullPointerException. To prevent this from happening, we should make the StandaloneConfig class fool proof from such configuration errors.

yong24s:
Fix #402

```
Extra trailing commas is non-compliant to JSON standards.

This causes gson to improperly parse a list, creating additional
null elements, resulting NullPointerException in RepoConfiguration#update.

Let's prevent NullPointerException by removing null elements in lists.
```


src/main/java/reposense/parser/StandaloneConfigJsonParser.java
```diff
@@ -27,6 +27,13 @@ public Type getType() {
      */
     @Override
     public StandaloneConfig parse(Path path) throws IOException {
-        return fromJson(path);
+        StandaloneConfig config = fromJson(path);
+
+        // Checks that config is properly parsed
+        if (config != null) {
+            config.initialize();
```
eugenepeh: whats the benefit of using `initialize();` over performing the null check in the getters itself?
The latter seems to be more resilience to me because dev/reviewer may missed out adding the field into the `initialize()` method.

yong24s: Good point. Since `StandaloneConfig` is a single use object, we can do null checks in the getters for simplicity.

Issue #394, opened 2018-11-03
Search bar allows lower case only


damithc:
Current: the search bar allows lowercase inputs only (i.e., auto-changes to lowercase)

Suggested: this is unnecessary as it can confuse the user albeit momentarily. If converting is needed, it can be done internally, hidden from the user.

PR #420, opened 2018-12-12
Fix appearance of search text and hide auto-changes to lowercase from user.

RonakLakhotia:
Fix [#394](https://github.com/reposense/RepoSense/issues/394)


frontend/src/static/js/v_summary.js
```diff
@@ -168,7 +168,8 @@ window.vSummary = {
     getFilterHash() {
       const { addHash } = window;
 
-      this.filterSearch = this.filterSearch.toLowerCase();
+
+      this.filterSearch = this.filterSearch;
```
yamidark: This line doesn't do anything anymore right?

frontend/src/static/js/v_summary.js
```diff
@@ -256,12 +257,12 @@ window.vSummary = {
...
         // filtering
         repo.users.forEach((user) => {
-          const toDisplay = this.filterSearch
+          const toDisplay = this.filterSearch.toLowerCase()
             .split(' ').filter(param => param)
             .map(param => user.searchPath.search(param) > -1)
             .reduce((curr, bool) => curr || bool, false);
 
-          if (!this.filterSearch || toDisplay) {
+          if (!this.filterSearch.toLowerCase() || toDisplay) {
```
yamidark: No need to convert lower case here, rather if `this.filterSearch` is `null` then we would get an error instead.

frontend/src/static/js/v_summary.js
RonakLakhotia: Yes it is a redundant line, will remove it.

frontend/src/static/js/v_summary.js
RonakLakhotia: Oh right, makes sense. 👍

RonakLakhotia: @yamidark requesting review

PR #421, opened 2018-12-12
[#394] Search bar allows lower case only

fzdy1914:
Delete the default translation of case in style.scss
Will keep work on for the auto-change of case after return key is entered


ongspxm:
Purpose of doing this?

ongspxm:
duplicate of #210

PR #422, opened 2018-12-12
Sort order name: rename from Team to Repo/Branch #30

### LINKED ISSUE #30: Sort order: team name --> repo
It's more accurate to say sort by `Repo` ?

hzf1174:
As mentioned in #30, the sorting behavior now looks like sorting by the Repo/Branch name instead of the Team name. 

Using the term `Team` may cause confusion to users. Renaming to `Repo/Branch` can also keep consistent with `group by repo/branch`.

Let's rename the sort order name from `Team` to `Repo/Branch`.


hzf1174:
@reposense/devs Ready for review.

hzf1174:
@reposense/devs Changes are made. Ready for review.

damithc:
> 😕 Repo isn't correct either because we don't group 2 different branches of the same repo together. I think we should probably retain it as team unless there's a better proposition.

That's true. `Repo/Branch`?

eugenepeh:
>That's true. Repo/Branch?

Sounds good to me. @hzf1174 please proceed with that.

Also, for the `group by repo` check box, can we have it relabeled correspondingly as well?

hzf1174:
> > That's true. Repo/Branch?
> 
> Sounds good to me. @hzf1174 please proceed with that.
> 
> Also, for the `group by repo` check box, can we have it relabeled correspondingly as well?

so we use `group by repo/branch` right?

eugenepeh:
> so we use group by repo/branch right?

yup

hzf1174:
Hi @eugenepeh, changes are made. Ready for review.

eugenepeh:
Nits to the proposed message:

```
In our tool bar, we use the term `Team` to define the group of authors
that from the same repository and branch. 

However, the use of this term can cause confusion to users as it is not
a common terminology among the Git users.

To improve the clarity of our selectable options, let's rename the word
Team to Repo/Branch.

PR #425, opened 2018-12-12
Split repo configuration #316

### LINKED ISSUE #316: Split RepoConfiguration into more classes
RepoConfiguration represents the analysis configuration for a repository.

However, as more analysis parameters being added, RepoConfiguration
becomes larger and harder to maintain.

This should be resolved by splitting some of the parameters to other classes,
such as location, glob list and etc as the validation of these params can be 
delegated to their own dedicated classes.

JasonChong96:
Fixes #316

`RepoConfiguration` has many analysis parameters and as more are added,
it becomes larger and harder to maintain.

Let's delegate some of the parameters to their own classes, this includes
`Location`, `Format` and `Commit`.

These parameters were delegated as currently only formats, ignored
commits and location are being validated.

### DIFF
--- a/src/main/java/reposense/RepoSense.java
+++ b/src/main/java/reposense/RepoSense.java
@@ -13,6 +13,7 @@ import reposense.model.CliArguments;
 import reposense.model.ConfigCliArguments;
 import reposense.model.LocationsCliArguments;
 import reposense.model.RepoConfiguration;
+import reposense.model.RepoLocation;
 import reposense.model.ViewCliArguments;
 import reposense.parser.ArgsParser;
 import reposense.parser.AuthorConfigCsvParser;
@@ -84,9 +85,9 @@ public class RepoSense {
      */
     public static List<RepoConfiguration> getRepoConfigurations(LocationsCliArguments cliArguments) {
         List<RepoConfiguration> configs = new ArrayList<>();
-        for (String location : cliArguments.getLocations()) {
+        for (String locationString : cliArguments.getLocations()) {
             try {
-                configs.add(new RepoConfiguration(location));
+                configs.add(new RepoConfiguration(new RepoLocation(locationString)));
             } catch (InvalidLocationException ile) {
                 logger.log(Level.WARNING, ile.getMessage(), ile);
             }
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -1,12 +1,6 @@
 package reposense.model;
 
 import java.io.File;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.nio.file.Files;
-import java.nio.file.InvalidPathException;
-import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -18,10 +12,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.TreeMap;
 import java.util.logging.Logger;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
-import reposense.parser.InvalidLocationException;
 import reposense.system.LogsManager;
 import reposense.util.FileUtil;
 
@@ -31,17 +22,11 @@ public class RepoConfiguration {
     private static final String MESSAGE_ILLEGAL_FORMATS = "The provided formats, %s, contains illegal characters.";
     private static final String FORMAT_VALIDATION_REGEX = "[A-Za-z0-9]+";
 
-    private static final String GIT_LINK_SUFFIX = ".git";
-    private static final Pattern GIT_REPOSITORY_LOCATION_PATTERN =
-            Pattern.compile("^.*github.com\\/(?<org>.+?)\\/(?<repoName>.+?)\\.git$");
     private static final String COMMIT_HASH_REGEX = "^[0-9a-f]+$";
     private static final String INVALID_COMMIT_HASH_MESSAGE =
             "The provided commit hash, %s, contains illegal characters.";
 
-
-    private String location;
-    private String organization;
-    private String repoName;
+    private RepoLocation location;
     private String branch;
     private String displayName;
     private Date sinceDate;
@@ -58,25 +43,16 @@ public class RepoConfiguration {
     private transient boolean isStandaloneConfigIgnored;
     private transient List<String> ignoreCommitList;
 
-    /**
-     * @throws InvalidLocationException if {@code location} cannot be represented by a {@code URL} or {@code Path}.
-     */
-    public RepoConfiguration(String location) throws InvalidLocationException {
+    public RepoConfiguration(RepoLocation location) {
         this(location, DEFAULT_BRANCH);
     }
 
-    /**
-     * @throws InvalidLocationException if {@code location} cannot be represented by a {@code URL} or {@code Path}.
-     */
-    public RepoConfiguration(String location, String branch) throws InvalidLocationException {
+    public RepoConfiguration(RepoLocation location, String branch) {
         this(location, branch, Collections.emptyList(), Collections.emptyList(), false, Collections.emptyList());
     }
 
-    /**
-     * @throws InvalidLocationException if {@code location} cannot be represented by a {@code URL} or {@code Path}.
-     */
-    public RepoConfiguration(String location, String branch, List<String> formats, List<String> ignoreGlobList,
-            boolean isStandaloneConfigIgnored, List<String> ignoreCommitList) throws InvalidLocationException {
+    public RepoConfiguration(RepoLocation location, String branch, List<String> formats, List<String> ignoreGlobList,
+            boolean isStandaloneConfigIgnored, List<String> ignoreCommitList) {
         this.location = location;
         this.branch = branch;
         this.ignoreGlobList = ignoreGlobList;
@@ -86,15 +62,12 @@ public class RepoConfiguration {
         validateIgnoreCommits(ignoreCommitList);
         this.ignoreCommitList = ignoreCommitList;
 
-        verifyLocation(location);
-        Matcher matcher = GIT_REPOSITORY_LOCATION_PATTERN.matcher(location);
+        String organization = location.getOrganization();
+        String repoName = location.getRepoName();
 
-        if (matcher.matches()) {
-            organization = matcher.group("org");
-            repoName = matcher.group("repoName");
+        if (organization != null) {
             displayName = organization + "_" + repoName + "_" + branch;
         } else {
-            repoName = Paths.get(location).getFileName().toString().replace(GIT_LINK_SUFFIX, "");
             displayName = repoName + "_" + branch;
         }
     }
@@ -169,10 +142,10 @@ public class RepoConfiguration {
     }
 
     public String getRepoRoot() {
-        String path = FileUtil.REPOS_ADDRESS + File.separator + repoName + File.separator;
+        String path = FileUtil.REPOS_ADDRESS + File.separator + getRepoName() + File.separator;
 
-        if (!repoName.isEmpty()) {
-            path += repoName + File.separator;
+        if (!getRepoName().isEmpty()) {
+            path += getRepoName() + File.separator;
         }
 
         return path;
@@ -328,43 +301,20 @@ public class RepoConfiguration {
         return displayName;
     }
 
-    public String getLocation() {
+    public String getRepoName() {
+        return location.getRepoName();
+    }
+
+    public RepoLocation getLocation() {
         return location;
     }
 
-    public String getRepoName() {
-        return repoName;
+    public String getOrganization() {
+        return location.getOrganization();
     }
 
     public boolean isStandaloneConfigIgnored() {
         return isStandaloneConfigIgnored;
-    }
-
-    /**
-     * Verifies {@code location} can be presented as a {@code URL} or {@code Path}.
-     * @throws InvalidLocationException if otherwise.
-     */
-    private void verifyLocation(String location) throws InvalidLocationException {
-        boolean isValidPathLocation = false;
-        boolean isValidGitUrl = false;
-
-        try {
-            Path pathLocation = Paths.get(location);
-            isValidPathLocation = Files.exists(pathLocation);
-        } catch (InvalidPathException ipe) {
-            // Ignore exception
-        }
-
-        try {
-            new URL(location);
-            isValidGitUrl = location.endsWith(GIT_LINK_SUFFIX);
-        } catch (MalformedURLException mue) {
-            // Ignore exception
-        }
-
-        if (!isValidPathLocation && !isValidGitUrl) {
-            throw new InvalidLocationException(location + " is an invalid location.");
-        }
     }
 
     /**
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -20,10 +20,6 @@ public class RepoConfiguration {
     public static final String DEFAULT_BRANCH = "HEAD";
     private static final Logger logger = LogsManager.getLogger(RepoConfiguration.class);
 
-    private static final String COMMIT_HASH_REGEX = "^[0-9a-f]+$";
-    private static final String INVALID_COMMIT_HASH_MESSAGE =
-            "The provided commit hash, %s, contains illegal characters.";
-
     private RepoLocation location;
     private String branch;
     private String displayName;
@@ -39,7 +35,7 @@ public class RepoConfiguration {
     private transient TreeMap<String, Author> authorAliasMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
     private transient Map<Author, String> authorDisplayNameMap = new HashMap<>();
     private transient boolean isStandaloneConfigIgnored;
-    private transient List<String> ignoreCommitList;
+    private transient List<CommitHash> ignoreCommitList;
 
     public RepoConfiguration(RepoLocation location) {
         this(location, DEFAULT_BRANCH);
@@ -50,14 +46,12 @@ public class RepoConfiguration {
     }
 
     public RepoConfiguration(RepoLocation location, String branch, List<Format> formats, List<String> ignoreGlobList,
-            boolean isStandaloneConfigIgnored, List<String> ignoreCommitList) {
+            boolean isStandaloneConfigIgnored, List<CommitHash> ignoreCommitList) {
         this.location = location;
         this.branch = branch;
         this.ignoreGlobList = ignoreGlobList;
         this.isStandaloneConfigIgnored = isStandaloneConfigIgnored;
         this.formats = formats;
-
-        validateIgnoreCommits(ignoreCommitList);
         this.ignoreCommitList = ignoreCommitList;
 
         String organization = location.getOrganization();
@@ -127,8 +121,9 @@ public class RepoConfiguration {
             aliases.add(author.getGitId());
             aliases.forEach(alias -> newAuthorAliasMap.put(alias, author));
         }
+
         Format.validateFormats(standaloneConfig.getFormats());
-        validateIgnoreCommits(standaloneConfig.getIgnoreCommitList());
+        CommitHash.validateCommits(standaloneConfig.getIgnoreCommitList());
 
         // only assign the new values when all the fields in {@code standaloneConfig} pass the validations.
         authorList = newAuthorList;
@@ -136,7 +131,7 @@ public class RepoConfiguration {
         authorDisplayNameMap = newAuthorDisplayNameMap;
         ignoreGlobList = newIgnoreGlobList;
         formats = Format.convertStringsToFormats(standaloneConfig.getFormats());
-        ignoreCommitList = standaloneConfig.getIgnoreCommitList();
+        ignoreCommitList = CommitHash.convertStringsToCommits(standaloneConfig.getIgnoreCommitList());
     }
 
     public String getRepoRoot() {
@@ -221,11 +216,11 @@ public class RepoConfiguration {
         this.ignoreGlobList = ignoreGlobList;
     }
 
-    public List<String> getIgnoreCommitList() {
+    public List<CommitHash> getIgnoreCommitList() {
         return ignoreCommitList;
     }
 
-    public void setIgnoreCommitList(List<String> ignoreCommitList) {
+    public void setIgnoreCommitList(List<CommitHash> ignoreCommitList) {
         this.ignoreCommitList = ignoreCommitList;
     }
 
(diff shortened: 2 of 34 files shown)


JasonChong96:
Ready for review

src/main/java/reposense/model/Location.java
```diff
@@ -0,0 +1,78 @@
...
+        try {
+            Path pathLocation = Paths.get(location);
+            isValidPathLocation = Files.exists(pathLocation);
+        } catch (InvalidPathException ipe) {
+            // Ignore exception
+        }
+
+        try {
+            new URL(location);
+            isValidGitUrl = location.endsWith(RepoConfiguration.GIT_LINK_SUFFIX);
```
eugenepeh: Since this `Represents a repository location.`, the `GIT_LINK_SUFFIX` should probably belongs to the `Location` class.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -25,7 +25,6 @@
     public static final String GIT_LINK_SUFFIX = ".git";
     private static final Logger logger = LogsManager.getLogger(RepoConfiguration.class);
     private static final String MESSAGE_ILLEGAL_FORMATS = "The provided formats, %s, contains illegal characters.";
```
eugenepeh: this should be removed since it has been moved over to formats.

src/main/java/reposense/model/IgnoreCommitsList.java
```diff
@@ -0,0 +1,54 @@
...
+            "The provided commit hash, %s, contains illegal characters.";
+
+    private List<String> ignoreCommits = new ArrayList<>();
+
+    public List<String> getIgnoreCommitsList() {
+        return ignoreCommits;
+    }
+
+    public void setIgnoreCommitsList(List<String> formats) {
+        validateIgnoreCommits(formats);
```
eugenepeh: This is doing double job. Although the effect is marginal, I would prefer to keep a clear separation between validation and set so that future developers will not attempt to set any of the values before all fields are validated in the `update` method.

eugenepeh:
can tag `@reposense/devs` when youre ready for next round of review.

JasonChong96:
@reposense/devs Ready for next round of review

src/main/java/reposense/model/FormatList.java
```diff
@@ -0,0 +1,59 @@
+package reposense.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class FormatList {
```
eugenepeh: Can do with a class header

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -77,24 +65,25 @@ public RepoConfiguration(String location, String branch) throws InvalidLocationE
...
+        this.location = new Location(location);
         this.branch = branch;
         this.ignoreGlobList = ignoreGlobList;
         this.isStandaloneConfigIgnored = isStandaloneConfigIgnored;
-        this.formats = formats;
 
-        validateIgnoreCommits(ignoreCommitList);
-        this.ignoreCommitList = ignoreCommitList;
+        FormatList.validateFormats(formats);
+        this.formats.setFormats(formats);
```
eugenepeh: actually... on a second look, why isn't this using a constructor?

src/test/java/reposense/model/FormatListTest.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class FormatListTest {
+    @Test
+    public void validateFormats_alphaNumeric_success()
+            throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+        Method m = FormatList.class.getDeclaredMethod("validateFormats", List.class);
+        m.setAccessible(true);
+        m.invoke(null, Arrays.asList("java", "7z"));
+    }
+    @Test(expected = IllegalArgumentException.class)
```
eugenepeh: add a line break to improve readability.

eugenepeh:
>@eugenepeh I don't think outsiders can tag the team (?)

I see... seems like GitHub still haven't implement that 🤔

src/main/java/reposense/model/RepoConfiguration.java
JasonChong96: Good point

JasonChong96:
@eugenepeh ready for review

JasonChong96:
Ready for next round of review

src/main/java/reposense/report/RepoConfigurationJson.java
```diff
@@ -0,0 +1,28 @@
+package reposense.report;
+
+import java.util.Date;
+
+import reposense.model.RepoConfiguration;
+
+/**
+ * Represents the structure of a repository's information in summary.json.
+ */
+public class RepoConfigurationJson {
```
eugenepeh: quick question: why is this necessary?

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -76,25 +64,22 @@ public RepoConfiguration(String location, String branch) throws InvalidLocationE
      * @throws InvalidLocationException if {@code location} cannot be represented by a {@code URL} or {@code Path}.
      */
     public RepoConfiguration(String location, String branch, List<String> formats, List<String> ignoreGlobList,
-            boolean isStandaloneConfigIgnored, List<String> ignoreCommitList) throws InvalidLocationException {
-        this.location = location;
+                boolean isStandaloneConfigIgnored, List<String> ignoreCommitList) throws InvalidLocationException {
```
eugenepeh: similarly, this indentation seems unnecessary.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -76,25 +64,22 @@ public RepoConfiguration(String location, String branch) throws InvalidLocationE
...
-        verifyLocation(location);
         Matcher matcher = GIT_REPOSITORY_LOCATION_PATTERN.matcher(location);
 
         if (matcher.matches()) {
             organization = matcher.group("org");
             repoName = matcher.group("repoName");
             displayName = organization + "_" + repoName + "_" + branch;
         } else {
-            repoName = Paths.get(location).getFileName().toString().replace(GIT_LINK_SUFFIX, "");
+            repoName = Paths.get(location).getFileName().toString().replace(Location.GIT_LINK_SUFFIX, "");
```
eugenepeh: I think the `Paths.get(location).getFileName().toString().replace(Location.GIT_LINK_SUFFIX, "");` can be replaced with a method in Location? That way we don't need to reference the `Location.GIT_LINK_SUFFIX`.

yong24s:
You may want to also consider these in another **PR** as there may be substantial `loc` changes:

Since `Location` is a class, it can be verified outside of `RepoConfiguration`.

`i.e.` RepoConfiguration(~String~ Location location ... ) ~throws InvalidLocationException~ {

src/main/java/reposense/report/RepoConfigurationJson.java
yamidark: I believe its because we convert the `String` location variable to a `Location` class, but it is a non-transient variable thus it is printed out in the json, so end up the format in the json output changes, which caused the tests to fail.
In this case, we can just update the expected json files to the new format instead of creating 1 whole new class for this.

JasonChong96:
@yamidark @yong24s Those implementations do make more sense, I'll redo this accordingly.

src/main/java/reposense/report/RepoConfigurationJson.java
JasonChong96: I was thinking that the json file would be more readable(/editable?) by keeping the fields consistent as it seems unintuitive to have only one field nested.

eugenepeh: True but let's say if its nested, we can include additional information such as `Local` or `Remote`. 

That would be a benefit to the existing implementation as we can fixed the link to local repository that currently doesn't work for local analysis.

JasonChong96:
Redid most of the code.
Ready for review

JasonChong96:
Redid the class delegation
Ready for review

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -41,12 +44,13 @@ public RepoConfigCsvParser(Path csvFilePath) throws IOException {
...
     protected void processLine(List<RepoConfiguration> results, String[] elements) throws InvalidLocationException {
-        String location = getValueInElement(elements, LOCATION_POSITION);
+        Location location = new Location(getValueInElement(elements, LOCATION_POSITION));
         String branch = getValueInElement(elements, BRANCH_POSITION, RepoConfiguration.DEFAULT_BRANCH);
-        List<String> formats = getManyValueInElement(elements, FILE_FORMATS_POSITION);
+        List<Format> formats = Format.convertStringsToFormats(getManyValueInElement(elements, FILE_FORMATS_POSITION));
         List<String> ignoreGlobList = getManyValueInElement(elements, IGNORE_GLOB_LIST_POSITION);
         String ignoreStandaloneConfig = getValueInElement(elements, IGNORE_STANDALONE_CONFIG_POSITION);
-        List<String> ignoreCommitList = getManyValueInElement(elements, IGNORE_COMMIT_LIST_CONFIG_POSITION);
+        List<Commit> ignoreCommitList = Commit.convertStringsToCommits(
```
yong24s: `InvalidArgumentException` may be thrown by `Format#convertStringsToFormats()` and `Commit#convertStringsToCommits()`, if given inputs are invalid. 

Also, please do a failing test to check it is handled 😃.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -110,12 +114,16 @@ public static CliArguments parse(String[] args) throws ParseException {
             Path outputFolderPath = results.get("output");
             Optional<Date> sinceDate = results.get("since");
             Optional<Date> untilDate = results.get("until");
-            List<String> formats = results.get("formats");
-            List<String> locations = results.get("repos");
+            List<Format> formats = Format.convertStringsToFormats(results.get("formats"));
+            List<String> locationStrings = results.get("repos");
```
yong24s: Any reason not to use this convention? Its pretty neat.

`List<Location> locations = Location.convertStringsToLocations(results.get("repos"));`

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -156,23 +120,23 @@ public void update(StandaloneConfig standaloneConfig) {
             aliases.add(author.getGitId());
             aliases.forEach(alias -> newAuthorAliasMap.put(alias, author));
         }
-        validateFormats(standaloneConfig.getFormats());
-        validateIgnoreCommits(standaloneConfig.getIgnoreCommitList());
+        Format.validateFormats(standaloneConfig.getFormats());
+        Commit.validateCommits(standaloneConfig.getIgnoreCommitList());
```
yong24s: Add a empty line before 123 for readability.

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -98,28 +99,9 @@ public void cleanRepoDirectory() throws IOException {
...
-            m.invoke(null, Arrays.asList(".java"));
-        } catch (InvocationTargetException ite) {
-            throw ite.getCause();
-        }
-    }
-
     @Test
     public void repoConfig_usesStandaloneConfig_success() throws InvalidLocationException, GitDownloaderException {
-        RepoConfiguration actualConfig = new RepoConfiguration(TEST_REPO_DELTA, "master");
+        RepoConfiguration actualConfig = new RepoConfiguration(new Location(TEST_REPO_DELTA), "master");
```
yong24s: Why not make it a Location?

`i.e.` `private static final Location TEST_REPO_DELTA = new Location(...)`

or if the string is required for validation

`private static final String TEST_REPO_DELTA = "..."`
`private static final Location TEST_REPO_DELTA_LOCATION = new Location(TEST_REPO_DELTA)`

src/main/java/reposense/parser/RepoConfigCsvParser.java
JasonChong96: Should it throw a `ParseException` when an `IllegalArgumentException` is caught or should it just log the error and use an empty list instead? Since as opposed to not being able to proceed without a valid location, the parsing can continue even without formats/ignore commits.

src/main/java/reposense/parser/ArgsParser.java
JasonChong96: Good point

src/main/java/reposense/parser/RepoConfigCsvParser.java
eugenepeh: Since we been using `IllegalArgumentException` before, let's just keep it as it is for now. We can fix it in another PR unless the exception is not handled anywhere.

JasonChong96: IllegalArgumentException is not handled anywhere currently (the application stops and the stack trace is shown to the user, 
`Exception in thread "main" java.lang.IllegalArgumentException: A
        at reposense.parser.RepoConfigCsvParser.processLine(RepoConfigCsvParser.java:49)
        at reposense.parser.CsvParser.parse(CsvParser.java:60)
        at reposense.RepoSense.getRepoConfigurations(RepoSense.java:68)
        at reposense.RepoSense.main(RepoSense.java:42)`

I'll just catch it in the CsvParser#parse and handle it similarly to ParseException for now. (Show a message telling the user it is invalid and stop there).

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -156,23 +120,24 @@ public void update(StandaloneConfig standaloneConfig) {
             aliases.add(author.getGitId());
             aliases.forEach(alias -> newAuthorAliasMap.put(alias, author));
         }
-        validateFormats(standaloneConfig.getFormats());
-        validateIgnoreCommits(standaloneConfig.getIgnoreCommitList());
+
+        Format.validateFormats(standaloneConfig.getFormats());
+        Commit.validateCommits(standaloneConfig.getIgnoreCommitList());
```
yamidark: Was thinking we no longer need this `validateFormats`/`validateCommits` method anymore, since when we convert the `String` -> `Commit`/`Format` we will do the validation check in the constructor itself.

However, since our current behaviour is that we completely ignore the `StandaloneConfig` if there are any invalid values inside, we can leave this here for now. Might want to remove this in the future, and allow using of the `StandaloneConfig` with only the valid values inside.

src/test/java/reposense/model/FormatTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.junit.Test;
+
+public class FormatTest {
+    @Test
+    public void validateFormats_alphaNumeric_success()
+            throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+        Method m = Format.class.getDeclaredMethod("validateFormats", List.class);
+        m.setAccessible(true);
+        m.invoke(null, Arrays.asList("java", "7z"));
```
yamidark: Rather than using Reflection for the test (which is considered bad practice since we are breaking encapsulation), we can use the class constructor for the validation tests here right?
Also, add some tests for valid `Commits` as well.

src/main/java/reposense/model/Commit.java
```diff
@@ -0,0 +1,87 @@
+package reposense.model;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Represents a git commit in {@code RepoConfiguration}.
+ */
+public class Commit {
```
yamidark: This may be confused with the `CommitInfo` we are using as well.
Perhaps rename to `IgnoreCommit` or `CommitHash`, to be more specific (update the header docs as well)?

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -62,8 +62,8 @@ public CsvParser(Path csvFilePath) throws IOException {
             }
         } catch (IOException ioe) {
             throw new IOException(MESSAGE_UNABLE_TO_READ_CSV_FILE, ioe);
-        } catch (ParseException pe) {
-            logger.log(Level.WARNING, pe.getMessage(), pe);
+        } catch (ParseException | IllegalArgumentException e) {
+            logger.log(Level.WARNING, e.getMessage(), e);
```
yamidark: We can have a separate catch block and logging message for both exceptions, also easier for us to trace which error occur.

JasonChong96: Not sure if I understood this correctly, they currently already have different messages, i.e MESSAGE_ILLEGAL_FORMATS and INVALID_COMMIT_HASH_MESSAGE.

yamidark:
@JasonChong96 

Do update your branch (resolve the conflicts), and let us know when it is ready for the next iteration of review.

JasonChong96:
@reposense/devs Ready for next round of review

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -55,13 +55,22 @@
...
+    private static final List<Format> CONFIG_FORMATS = Format.convertStringsToFormats(Arrays.asList(
+            "java", "adoc", "md"));
     private static final List<String> CLI_FORMATS = Arrays.asList("css", "html");
 
     private static RepoConfiguration REPO_DELTA_STANDALONE_CONFIG;
 
+    static {
+        try {
+            TEST_REPO_DELTA = new Location("https://github.com/reposense/testrepo-Delta.git");
+        } catch (InvalidLocationException e) {
```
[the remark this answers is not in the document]
yong24s: My intention of creating `TEST_REPO_DELTA` was to simplify tests.

`i.e.` `Assert.assertEquals(new Location(TEST_REPO_BETA_LOCATION), config.getLocation());` -> `Assert.assertEquals(TEST_REPO_BETA, config.getLocation());`

However, I have overlooked that an ugly `try-catch` block is required. 

@yamidark 
What do you think of this, should we stick the to previous way?
`Assert.assertEquals(new Location(TEST_REPO_BETA_LOCATION), config.getLocation());`
Since there are some tests that still haven't got refactored.

src/main/java/reposense/model/Location.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import reposense.parser.InvalidLocationException;
+
+/**
+ * Represents a repository location.
+ */
+public class Location {
```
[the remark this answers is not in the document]
JasonChong96: My bad

JasonChong96:
@reposense/devs Ready for next round of review

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -18,83 +12,53 @@
...
 
-        verifyLocation(location);
-        Matcher matcher = GIT_REPOSITORY_LOCATION_PATTERN.matcher(location);
-
-        if (matcher.matches()) {
-            organization = matcher.group("org");
-            repoName = matcher.group("repoName");
+        String organization = location.getOrganization();
+        String repoName = location.getRepoName();
+        if (organization != null) {
```
yong24s: Add an empty line before this.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -0,0 +1,120 @@
...
+    private String organization;
+
+    /**
+     * @throws InvalidLocationException if {@code location} cannot be represented by a {@code URL} or {@code Path}.
+     */
+    public RepoLocation(String location) throws InvalidLocationException {
+        verifyLocation(location);
+        this.location = location;
+        Matcher matcher = GIT_REPOSITORY_LOCATION_PATTERN.matcher(location);
+        if (matcher.matches()) {
```
yong24s: Add an empty line before this.

src/main/java/reposense/model/RepoConfiguration.java
JasonChong96: Thanks for the tips!

JasonChong96:
@reposense/devs Ready for next round of review

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -225,13 +228,10 @@ public void parse_repoAliases_sameResult() throws ParseException, IOException {
         Assert.assertEquals(repoAliasCliArguments, reposAliasCliArguments);
     }
 
-    @Test
-    public void parse_invalidRepoLocation_emptyRepoConfigurationList() throws ParseException, IOException {
+    @Test(expected = InvalidLocationException.class)
+    public void parse_invalidRepoLocation_emptyRepoConfigurationList() throws ParseException {
```
yamidark: `parse_invalidRepoLocation_throwsInvalidLocationException`

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public int hashCode() {
+        return location.hashCode();
+    }
+
+    /**
+     * Converts all the strings in {@code locations} into {@code RepoLocation} objects.
+     * Returns null if {@code locations} is null.
+     * @throws IllegalArgumentException if any of the strings are in invalid formats.
```
yamidark: Should be `@throws InvalidLocationException`

src/main/java/reposense/RepoSense.java
```diff
@@ -84,12 +84,8 @@ public static void main(String[] args) {
...
     public static List<RepoConfiguration> getRepoConfigurations(LocationsCliArguments cliArguments) {
         List<RepoConfiguration> configs = new ArrayList<>();
-        for (String location : cliArguments.getLocations()) {
-            try {
-                configs.add(new RepoConfiguration(location));
-            } catch (InvalidLocationException ile) {
-                logger.log(Level.WARNING, ile.getMessage(), ile);
-            }
+        for (RepoLocation location : cliArguments.getLocations()) {
+            configs.add(new RepoConfiguration(location));
```
yamidark: Sorry, just noticed this from the changed test earlier, the behaviour of our program should not be changing here, from filtering out only invalid 'repo locations', to completely abandoning the process if any1 of the 'repo locations' is invalid.

So in `ArgsParser` and `LocationCliArguments`, we should make them maintain a list of `String`, and we do the conversion to `RepoLocation` here instead before passing it into `RepoConfiguration` (or we can do the conversion within `RepoConfiguration` constructor itself).

@yong24s What you think of this?

yong24s: > So in ArgsParser and LocationCliArguments, we should make them maintain a list of String, and we do the conversion to RepoLocation here instead before passing it into RepoConfiguration (or we can do the conversion within RepoConfiguration constructor itself).

Yes, this implementation is direct and simple.

In the future, we should look into using `ConfigurationBuilder`, where we can delay the validation until `build()` is invoked.

JasonChong96:

Issue #396, opened 2018-11-05
Provide a link to repo even when not grouped by repo


damithc:
Current: a link to the repo is only available when the chart view is grouped by repo.

![image](https://user-images.githubusercontent.com/1673303/47981150-9ffc5380-e105-11e8-9975-9cbbe7acac23.png)

Suggested: also provide a way to reach the repo when not grouped by repo.

yamidark:
Yes, closing as issue is resolved in #426.

damithc:
@fzdy1914 remember to put the correct `Fixes #1234` in the PR so that the issue gets auto-closed when the PR is merged.

PR #426, opened 2018-12-12
#396 [Enhancement] Provide a link to repo even when not grouped by repo

fzdy1914:
Add a link to repo page at the personal detail page.
image
More enhancement can be done if needed to add a direct link on the main page.

### DIFF
--- a/frontend/src/static/css/style.scss
+++ b/frontend/src/static/css/style.scss
@@ -232,6 +232,7 @@ header{
         .repoName{
             font-size:1.5rem;
             font-weight:bold;
+            cursor:pointer;
         }
         .contribution{
             margin-top:0.1rem;
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -90,7 +90,7 @@ html
               .summary-chart__title(
                 title="click to view author's code",
                 v-on:click="$emit('view-authorship', \
-                    {author:user.name, repo:user.repoName, name:user.displayName, \
+                    {author:user.name, repo:user.repoName, name:user.displayName, location:repo[0].location, \
                         minDate:minDate, maxDate:maxDate, totalCommits:user.totalCommits})"
               )
                 .summary-chart__title--index {{ i+1 }}
@@ -122,7 +122,10 @@ html
     vuetemplate#v_authorship
       #authorship
         .title
-          .repoName {{ info.repo }}
+          a.repoName(
+            v-bind:href="info.location", target="_blank",
+            v-bind:title="'Click to open the repo'"
+          ) {{ info.repo }}
           .author {{ info.name }} ({{ info.author }}) contribution from {{ info.minDate }} to {{ info.maxDate }}
           .contribution(v-if="isLoaded")
             span Total LoC contributed:&nbsp;


damithc:
@fzdy1914 I prefer a more subtle approach e.g., an octocat icon linked to the repo

fzdy1914:
I have added the octocat icon with the link
image

damithc:
> I have added the octocat icon with the link

Nice, but doesn't go with the other icon `</>` which is grey. :-p

fzdy1914:
Have edited to make it all gray.
image

fzdy1914:
> Better 👍
> No need for the `ReposeSense_Link` right?

It's the group repo name actually, and I screenshotted it when it is still grouped by repo. It will disappear when is not grouped by repo.

frontend/src/static/css/style.scss
```diff
@@ -183,6 +184,11 @@ header{
         &--tooltip{
             color:mui-color("grey");
         }
+
+        &--octocat{
```
eugenepeh: instead of octocat, probably can rename to repourl-button or something.
do change the code view as well so that they're standardized.

yamidark:
>It's the group repo name actually, and I screenshotted it when it is still grouped by repo. It will disappear when is not grouped by repo.

I think what @damithc meant was that there is no longer a need for the group repo name to also have a link to the github repo anymore.

frontend/src/static/css/style.scss
fzdy1914: Can I know what is the meaning of "change the code view as well so that they're standardized"?

fzdy1914: Do you mean that adding an icon at the title of code view page?

yamidark:
Do tag @reposense/devs when you are ready for the next review.

fzdy1914:

frontend/src/static/css/style.scss
eugenepeh: Hi @fzdy1914 

> change the code view as well so that they're standardized

Yes if possible, I think it would be better to contain the code view link within the icon itself as well for standardization. Then we can probably change the clicking of title to hide the ramp chart or something in the future. @damithc your opinion on this?

frontend/src/static/css/style.scss
```diff
@@ -180,8 +179,10 @@ header{
             padding-right:0.25rem;
         }
 
-        &--tooltip{
+        &--tooltip, &--repo-url-button{
```
eugenepeh: `tooltip` should probably be changed to `codeview-button`

fzdy1914: Have changed

frontend/src/static/css/style.scss
```diff
@@ -176,8 +175,10 @@ header{
             padding-right:0.25rem;
         }
 
-        &--tooltip{
+        &--code-view-button, &--repo-url-button{
```
ongspxm: if the style is the same, we will try to use the same modifier.

in this case, both should be `--button`

fzdy1914:
@ongspxm Have edited the css. Please review.

damithc:
Doesn't this require any updates to user guide?

yamidark:
>Doesn't this require any updates to user guide?

Yes, @fzdy1914 do update the Chart Panel diagrams in the UserGuide with the new UI.

fzdy1914:
Sorry for not updating the diagram. Because I am now in China and I get problem directly cloning the repo from github, which takes a quite long time. I will update the diagram as long as the repo are cloned correctly.

fzdy1914:
UG updated, please review.

eugenepeh:
Nits to the proposed message:

```
Link to the repository is only available in the chart view when the
`group by repo/branch` is selected, represented by a embedded hyperlink
in the group title.

This cause inconvenience to the users who want to visit the repository
when the authors are listed individually as the link to repository
cannot be found. 

Let's move the hyperlink in the group title to the individual author's
code view, and also add a button next to every author's name in the
chart view that will direct the users to the author's repository, to
ensure consistent navigability to the repositories.
```

Issue #405, opened 2018-11-26
code view: auto-scroll to top when loading code of a new person


damithc:
Current: the code panel holds the scroll position of the previous view even after loading code of a new person

Suggested: auto-scroll to the top

eugenepeh:
similar issue to https://github.com/reposense/RepoSense/issues/190, closing the other...

yamidark:
>This issue is resolved, right?

Yes, closing this issue as it is resolved in #428.

PR #428, opened 2018-12-13
[#405] Code view: reset to top when loading new code view

fzdy1914:
As mentioned in #405, the code panel holds the scroll position of the previous view even after loading code of a new person.

A user may need to scroll the code panel to the top by himself which is inconvenient.

Let's add auto scroll to top when loading code of a new person.


fzdy1914:
It is weird that the test fails.
If click the details, it says the deployment is successful
image

fzdy1914:
Added check for the first time load of tab wrapper.

fzdy1914:
@eugenepeh Can I know what aspect can I improve?  Sorry for not getting your point.

eugenepeh:
>@eugenepeh Can I know what aspect can I improve? Sorry for not getting your point.

Let's take `Ahm` and `Jor` as example. If you open the code view of `Ahm`, scroll down a bit then open `Jor`, you will realize that the code view for `Jor` is not reset to the top.

However, when you return to `Ahm`, you will realize that the code view for `Ahm` is still at the position where you left it to be (which is a good thing). However, if you scrolled the code view when you're on `Jor`, you will realized that `Ahm` code view gets shifted correspondingly (because they're in the same group).

fzdy1914:
@eugenepeh I believe this problem may due to the browser we use. For my chrome, it behaves well:
a
I will try to test on other browser to find out the reason.

eugenepeh:
@fzdy1914 hmm nvm, seems to be working for me now after updating the chrome browser.

Just one more concern though:
The url always gets updated to `https://deploy-preview-428--reposense.netlify.com/#tab-content` when we open a new code view, losing all the search and filter parameters. Is it possible to retain them as bookmarking is one of our intended functionality?

ongspxm:
> Actually, the URL will auto jump to the correct location in a very short time, so it should not be a big problem.

oh, on firefox it doesn't jump back at all, it just keeps at `#tabContent` until the user make some changes to the params, or when it get rewritten over.

in this case, you should check that `tabContent` exist before you do the jumping. so in full, the code wil be something like:

```
var ele = $('#tabContent');
if (ele){
    ele.parentElement.scrollTop=0;
}
```

ongspxm:
dun really need the `!=null` cos in javascript is default. but looks gud

eugenepeh:
Nits to the proposed message:

```
The code panel holds the scroll position of the previous view even
after loading the code of a new person.

This can cause inconvenience to the reviewers as they need to scroll
the panel back to the top after each review.

Let's reset the code panel back to the top when the content changes.
```

PR #429, opened 2018-12-13
[#392]Add Loading Page

### LINKED ISSUE #392: Show a 'loading...' message while loading the report for the first time
Current: When a Browser is loading a report for the first time, it shows the following:
image

Suggested: Show a 'loading ...' page instead

Reason: When loading a big report (which could take a while), the user might think the report is not loading at all and a zip file is needed. e.g., the [current cs2103-dashboard](https://nus-cs2103-ay1819s1.github.io/cs2103-dashboard/#=undefined&search=&sort=displayName&since=2018-09-12&until=2018-11-02&timeframe=day&reverse=false&repoSort=true) takes quite long to load and the initial page shown during that period is misleading.

hzf1174:
Fixes #392 
loading

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -44,6 +44,8 @@ html
 
         form#file-upload(onsubmit="return false;")
           input(type="file", v-on:change="updateReportZip")
+          .loader(v-if="isLoading")
+            .loading Loading zip file...
 
     vuetemplate#v_summary
       #summary.wrapper
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -45,7 +45,7 @@ html
         form#file-upload(onsubmit="return false;")
           input(type="file", v-on:change="updateReportZip")
           .loader(v-if="isLoading")
-            .loading Loading zip file...
+            .loading Loading resources...
 
     vuetemplate#v_summary
       #summary.wrapper
--- a/frontend/src/static/css/style.scss
+++ b/frontend/src/static/css/style.scss
@@ -23,6 +23,24 @@ vuetemplate{
     text-align:center;
 }
 
+.loader{
+    background:#000;
+    opacity:1;
+    position:fixed;
+    z-index:10;
+    top:0px;
+
+    width:100%;
+    height:1600px;
+
+    .loading{
+        color:#FFF;
+        width:120px;
+        height:24px;
+        margin:300px auto;
+    }
+}
+
 /** navigation bar **/
 header{
     z-index:100;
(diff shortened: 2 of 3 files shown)


hzf1174:
The files above are here @yamidark 
Invalid and Valid zip:
valid-and-invalid-zip
Server:
server

yamidark:
@hzf1174 

Seems like there are different cases for different invalid files.

In my case, I give a zip file from a successful run, but I removed some lines in my `summary.json`, which lead to the hanging in the loading page/window.

You can try this zip file [summary.zip](https://github.com/reposense/RepoSense/files/2684647/summary.zip) 
and this json file [summary.txt](https://github.com/reposense/RepoSense/files/2684655/summary.txt) (convert to back to `json`, since github doesn't allow it).

>For testing invalid files, I think it may beyond the topic of this PR.

I don't think I mentioned anything about testing here (we don't really have any tests for frontend yet anyway). What I meant was for a pop-up message or some indication to the user when he has failed to upload the file (invalid), so he has a better clue as to what is happening (as the loading page/window might not even appear if it failed from the start, and the user may not know that it has even failed).

But then again, this is probably a different issue altogether, so no need to work on this PR.

hzf1174:
@yamidark 
I checked the file summary.zip. In my opinion, the reason why the page is stuck is that the summary.zip file disguised as a valid file. As it provides a valid summary.json file, it is hard to judge whether it is a valid file, because we can not distinguish whether the commit.json file is taking a long time to load, or it does not exist at all.
If the user is not malicious and generates the archive.zip by using CLI, I think the current implementation is enough for the common situations. However, it could be improved in the future by completing the validation test of the file. What do you think about it?

ongspxm:
> @ongspxm Your input on this? I tried using .catch(err), which works in the situations above (my summary.zip), but seems to fail to catch the error when loading server and any of the repo folder is missing (failing at loadJson method, return new Promise() section, and the window.alert('unable to get file') gets called).

on [here](https://github.com/reposense/RepoSense/blob/dda75347ebbc305da0d2e20f7fa07423b352b52d/frontend/src/static/js/api.js#L12), add a reject callback with a unique text, like this:

```
return new Promise((resolve, reject) => {
    const xhr = new XMLHttpRequest();
    xhr.open('GET', fname);
    xhr.onload = function xhrOnload() {
      if (xhr.status === 200) {
        resolve(JSON.parse(xhr.responseText));
      } else {
        reject('unableToGetFile');
        window.alert('unable to get file');
      }
    };
    xhr.send(null);
});
```

then on your vue side, you check the error message, if it is `unableToGetFile` then you do the changes you want to make.

yamidark:
@hzf1174 

Could you try implementing the error detection both me and @ongspxm has mentioned?

yamidark:
@hzf1174 

Any updates on this PR?

yamidark:
Closing due to inactivitiy

emer7:
I'm taking over

PR #432, opened 2018-12-15
Undefined parameter in the URL #393

### LINKED ISSUE #393: Undefined parameter in the URL
Why do we have this? Can remove?
image

JasonChong96:
Fixes #393 

`window.location.hash` is initially empty and calling `split('&')` 
creates an array containing a single empty string. 

When trying to split this empty string into a key and value,
the result is an empty key and and `undefined` value which 
causes `=undefined` to appear in the URL.

Let's add a check to ensure that the key is not empty to 
prevent this empty string from being included when adding 
key value pairs to `window.hashParams`. This also prevents
any empty keys introduced as a result of user error, perhaps 
from editing the URL manually, from being processed.


JasonChong96:
Ready for review

frontend/src/static/js/v_summary.js
```diff
@@ -183,7 +183,7 @@ window.vSummary = {
       const params = window.location.hash.slice(1).split('&');
       params.forEach((param) => {
         const [key, val] = param.split('=');
-        window.hashParams[key] = decodeURIComponent(val);
+        if (key) { window.hashParams[key] = decodeURIComponent(val); }
```
eugenepeh: By our coding convention, we discourage the use of single line conditional because making changes to either (the condition or the statements) affect the other.

Please change to 

```
if (key) { 
    window.hashParams[key] = decodeURIComponent(val); 
}
```

JasonChong96: I'll change it. 
Since single line conditionals are currently used throughout the same method, should I change the other lines too in a seperate PR?

eugenepeh: > Since single line conditionals are currently used throughout the same method, should I change the other lines too in a seperate PR?

That one we can slowly fix them when there's a need to change them

eugenepeh:
Thanks for the PR! Great work and good proposed message! Keep it up!

PR #434, opened 2018-12-16
Show breakdown of blank line for total contribution in code panel #427

### LINKED ISSUE #427: Show breakdown of blank line for total contribution in code panel
Code Panel currently shows the total LOC, which includes both blank and non-blank lines. It would be good if the code panel could also show a breakdown of the blank lines and non-blank lines i.e. ```(Blank: 200, Non-Blank:1800)``` Such a breakdown provides a better understanding of the actual amount of code contributed by the author.
screen shot 2018-12-13 at 2 38 43 pm

apoorva17:
Fixes #427 
```
With the addition of #385, code view now shows a breakdown of total
contribution by filetype.

However, this does not provide any information on the blank LOC and
non-blank LOC.

Let's show a breakdown of total contribution by blank lines and
non-blank lines in the code view, to provide a better understanding of
the actual amount of code contributed by the author.

Let's also show this breakdown for each file type in the form of a
tooltip.
```
dashboard


apoorva17:

frontend/src/index.jade
```diff
@@ -126,11 +126,12 @@ html
...
           .contribution(v-if="isLoaded")
             span Total LoC contributed:&nbsp;
-            span.loc {{ totalLineCount }}
+            span.loc {{ totalLineCount }}&nbsp;
+            span.bloc (Blank: {{ totalBlankLineCount }}, Non-Blank: {{ totalLineCount - totalBlankLineCount }})
             br
             span.fileType(v-if="totalLineCount > 0") &rarr; &nbsp;
                 template(v-for="type in Object.keys(filesLinesObj)")
-                    span {{ type }}: {{ filesLinesObj[type] }}
+                    span(v-bind:title="getBlankLineCount(type)") {{ type }}: {{ filesLinesObj[type] }}
```
yamidark: Cant seem to see this portion on the netlify deploy?

frontend/src/static/js/v_authorship.js
```diff
@@ -129,6 +146,11 @@ window.vAuthorship = {
               ...acc, [key]: unsortedFilesInfoObj[key]
           }), {});
     },
+
+    getBlankLineCount(fileType) {
```
yamidark: This method name makes me think I'm just getting the blank line of the entire analysis counts as an `int`, but it returns a `string` instead with info on non-blank line count as well. Perhaps rename to `getFileBlankLineInfo`?

frontend/src/static/js/v_authorship.js
```diff
@@ -97,16 +110,20 @@ window.vAuthorship = {
           out.lineCount = lineCnt;
           this.addLineCountToFileType(file.path, lineCnt, filesInfoObj);
 
-          const segments = this.splitSegments(file.lines);
-          out.segments = segments;
+          var segmentInfo = this.splitSegments(file.lines);
```
yamidark: Any reason why this is changed to `var`?

damithc:

Sounds good.

frontend/src/index.jade
apoorva17: it appears when you hover over the filetype in the code view in netlify

apoorva17:
@eugenepeh @damithc so u would like to retain the current tooltip display which shows the breakdown of blank and non-blank line for each filetype, and also add in the blank lines in brackets like below? but in such a case how would the user know what the value in the brackets represent?

> Total LoC contributed: 2316 (401)
> →  gradle: 47 (8), java: 2143 (363), js: 4 (0), md: 120 (30), yml: 2 (0) 
> ```
> where the number in the bracket denotes the blank lines.

eugenepeh:
> @eugenepeh @damithc so u would like to retain the current tooltip display which shows the breakdown of blank and non-blank line for each filetype, and add in the blank lines in brackets in the code view like below right? but would it still be confusing for the user as to what the values in the brackets represent?

Since it should be colored in gray, its not important information. Users who would like to know what that means can either hover over it or refer to the user guide for more info. So it shouldn't be much of a problem.

frontend/src/static/js/v_authorship.js
apoorva17: ok, have renamed the method

frontend/src/static/js/v_authorship.js
apoorva17: have changed it back to ```const```

apoorva17:
> Since it should be colored in gray, its not important information. Users who would like to know what that means can either hover over it or refer to the user guide for more info. So it shouldn't be much of a problem.

Ok, i've added in the values in brackets accordingly, and i've also added a tooltip for total LOC blank and non-blank lines breakdown.

yong24s:
Some comments on the UI, else LGTM. 👍 

Is it more user-friendly if the tooltip appears when hover for the entire `adoc: 304 (71)` rather than just the bracket `(71)`?

Also for the tooltip content, is it more intuitive for the `Non-Blank` to be on the left since it of interest to the user? `i.e.`  `adoc: Non-Blank: 233, Blank: 71`

Let's hear from @damithc first. 😄

damithc:
@apoorva17 can post a screenshot of the latest UI?

apoorva17:
dashboard
The tooltip currently appears when user hovers over bracket value ```(1)```
> @apoorva17 can post a screenshot of the latest UI?

damithc:
I'm OK with the current UI.

> Is it more user-friendly if the tooltip appears when hover for the entire `adoc: 304 (71)` rather than just the bracket `(71)`?

Yes, that can be useful.

apoorva17:
I've updated the implementation with this
> Is it more user-friendly if the tooltip appears when hover for the entire `adoc: 304 (71)` rather than just the bracket `(71)`?

PR #446, opened 2018-12-20
Make -view a flag to launch the report automatically after processing #315

### LINKED ISSUE #315: Make -view a flag to launch the report automatically after processing
-view is currently a separate command which user has to enter after 
the report is generated to view the report.

This is a hassle to the user as it can be joined together with the main
command to form an atomic action.

The original functionality of -view can remain though.

myteo:
Fixes #315 based on suggestions made in pull request #437.

```
A separate -view command is required to view the report after
generating it. 

This is a hassle to the user as it can be joined together with the main
command to form an atomic action.

Let's make -view a flag for users to view the report automatically
after it has been generated.
```


myteo:
Work in progress

myteo:
Ready for review

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -59,7 +59,8 @@ private static ArgumentParser getArgumentParser() {
                 .metavar("LOCATION")
                 .help("The GitHub URL or disk locations to clone repository.");
 
-        mutexParser.addArgument("-view")
+        parser.addArgument("-view")
+                .nargs("*")
                 .metavar("PATH")
                 .type(new ReportFolderArgumentType())
                 .help("Starts a server to display the dashboard in the provided directory.");
```
yamidark: Should update this as well (same as what you wrote in the DeveloperGuide)

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -151,6 +179,39 @@ public void parse_configFolderAndOutputDirectory_success() throws ParseException
         Assert.assertTrue(Files.isSameFile(expectedAbsoluteOutputDirectoryPath, cliArguments.getOutputFilePath()));
     }
 
+    @Test
+    public void parse_configFolderAndLaunchAutomatically_success() throws ParseException, IOException {
```
yamidark: `parse_configFolderandViewWithouthArgs_returnsConfigCliArguments()`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -190,6 +251,35 @@ public void parse_validGitRepoLocations_repoConfigurationListCorrectSize() throw
         Assert.assertEquals(2, repoConfigs.size());
     }
 
+    @Test
+    public void parse_repoLocationsAndLaunchAutomatically_success() throws ParseException {
```
yamidark: `parse_repoLocationsAndView_returnsLocationCliArguments`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -190,6 +251,35 @@ public void parse_validGitRepoLocations_repoConfigurationListCorrectSize() throw
...
+        String input = String.format("-repos \"%s\" %s -view", TEST_REPO_REPOSENSE, TEST_REPO_DELTA);
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof LocationsCliArguments);
+        Assert.assertTrue(cliArguments.isAutomaticallyLaunching());
+    }
+
+    @Test
+    public void parse_repoLocationsAndNoLaunchAutomatically_success() throws ParseException {
```
yamidark: `parse_repoLocationsOnly_success()`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -127,6 +134,27 @@ public void parse_viewOnly_success() throws ParseException, IOException {
...
+        String input = String.format("-view %s %s", OUTPUT_DIRECTORY_ABSOLUTE, CONFIG_FOLDER_ABSOLUTE);
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof ViewCliArguments);
+        Assert.assertTrue(Files.isSameFile(
+                OUTPUT_DIRECTORY_ABSOLUTE, ((ViewCliArguments) cliArguments).getReportDirectoryPath()));
+    }
+
+    @Test
+    public void parse_viewOnlyWithoutArgs_resultsInConfigCliArguments() throws ParseException {
```
yamidark: `parse_viewOnlyWithoutArgs_returnsConfigCliArguments`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -190,6 +251,35 @@ public void parse_validGitRepoLocations_repoConfigurationListCorrectSize() throw
...
+    public void parse_repoLocationsAndNoLaunchAutomatically_success() throws ParseException {
+        String input = String.format("-repos \"%s\" %s", TEST_REPO_REPOSENSE, TEST_REPO_DELTA);
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof LocationsCliArguments);
+        Assert.assertFalse(cliArguments.isAutomaticallyLaunching());
+    }
+
+    @Test
+    public void parse_repoLocationsAndViewWithArgs_resultsInViewCliArguments() throws ParseException, IOException {
```
yamidark: `parse_repoLocationsAndViewWithArgs_returnsViewCliArguments`

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -59,7 +59,8 @@ private static ArgumentParser getArgumentParser() {
                 .metavar("LOCATION")
                 .help("The GitHub URL or disk locations to clone repository.");
 
-        mutexParser.addArgument("-view")
+        parser.addArgument("-view")
+                .nargs("*")
```
yong24s: Since the number of args is expected to be 0 or 1, also args >1 will be silently ignored, isn't `?` a better fit?
`.nargs("*")` -> `.nargs("?")`

docs/DeveloperGuide.md
```diff
@@ -69,7 +69,7 @@ Eslint and its accompaning modules can be installed through NPM, so do ensure th
 ### Building and running RepoSense from code
 
 1. Execute the following command on the OS terminal inside the project directory. <br/>
-Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | -repos REPO_PATH_OR_URL... | -view REPORT_FOLDER) [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...]"` <br/>
+Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | -repos REPO_PATH_OR_URL...) [-view REPORT_FOLDER] [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...]"` <br/>
```
yong24s: I realized `-repos REPO_PATH_OR_URL...` is not mandatory, please help to wrap it with `[` `]`.

src/main/java/reposense/model/ConfigCliArguments.java
```diff
@@ -17,14 +17,16 @@
     private Path authorConfigFilePath;
 
     public ConfigCliArguments(Path configFolderPath,
-            Path outputFilePath, Optional<Date> sinceDate, Optional<Date> untilDate, List<String> formats) {
+            Path outputFilePath, Optional<Date> sinceDate, Optional<Date> untilDate, List<String> formats,
+                              boolean isAutomaticallyLaunching) {
```
yong24s: Line break can be improved.
Move everything up a line then break after `sinceDate`

```
    public ConfigCliArguments(Path configFolderPath, Path outputFilePath, Optional<Date> sinceDate, 
            Optional<Date> untilDate, List<String> formats, boolean isAutomaticallyLaunching) {
```

src/main/java/reposense/model/LocationsCliArguments.java
```diff
@@ -12,12 +12,14 @@
     private List<String> locations;
 
     public LocationsCliArguments(List<String> locations,
-            Path outputFilePath, Optional<Date> sinceDate, Optional<Date> untilDate, List<String> formats) {
+            Path outputFilePath, Optional<Date> sinceDate, Optional<Date> untilDate, List<String> formats,
+                                 boolean isAutomaticallyLaunching) {
```
yong24s: Line break can be improved.

src/main/java/reposense/parser/ArgsParser.java
yamidark: @yong24s I tried this, but it seems if we just called `-view` without any parameter, it will give a `null` value as well, which is the same as not indicating the `-view` argument.

src/main/java/reposense/parser/ArgsParser.java
myteo: Do you mean update the help text?

myteo:
Edits have been made. Ready for review.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -151,6 +179,39 @@ public void parse_configFolderAndOutputDirectory_success() throws ParseException
...
+        Assert.assertTrue(cliArguments instanceof ConfigCliArguments);
+        Assert.assertTrue(Files.isSameFile(
+                REPO_CONFIG_CSV_FILE, ((ConfigCliArguments) cliArguments).getRepoConfigFilePath()));
+        Assert.assertTrue(Files.isSameFile(
+                AUTHOR_CONFIG_CSV_FILE, ((ConfigCliArguments) cliArguments).getAuthorConfigFilePath()));
+        Assert.assertTrue(cliArguments.isAutomaticallyLaunching());
+    }
+
+    @Test
+    public void parse_configFolderAndViewWithArgs_resultsInViewCliArguments() throws ParseException, IOException {
```
yamidark: `..._returnsViewCliArguments`

src/main/java/reposense/parser/ArgsParser.java
yamidark: >Do you mean update the help text?

Yep

src/main/java/reposense/model/ConfigCliArguments.java
```diff
@@ -17,14 +17,16 @@
     private Path authorConfigFilePath;
 
     public ConfigCliArguments(Path configFolderPath,
-            Path outputFilePath, Optional<Date> sinceDate, Optional<Date> untilDate, List<String> formats) {
+            Path outputFilePath, Optional<Date> sinceDate,
```
yamidark: This 2 can be appended with the line above

src/main/java/reposense/model/LocationsCliArguments.java
```diff
@@ -12,12 +12,14 @@
     private List<String> locations;
 
     public LocationsCliArguments(List<String> locations,
-            Path outputFilePath, Optional<Date> sinceDate, Optional<Date> untilDate, List<String> formats) {
+            Path outputFilePath, Optional<Date> sinceDate,
```
yamidark: Same as above

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -115,15 +118,19 @@ public static CliArguments parse(String[] args) throws ParseException {
...
+            if (reportFolderPath != null && !reportFolderPath.isEmpty()) {
+                return new ViewCliArguments(reportFolderPath.get(0));
             }
 
-            if (reportFolderPath != null) {
-                return new ViewCliArguments(reportFolderPath);
+            boolean isAutomaticallyLaunching = reportFolderPath != null;
+
+            if (locations != null) {
+                return new LocationsCliArguments(locations, outputFolderPath, sinceDate, untilDate, formats,
```
yamidark: Let's do a line break after `(` instead (all parameters in the same line)

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -115,15 +118,19 @@ public static CliArguments parse(String[] args) throws ParseException {
...
+                        isAutomaticallyLaunching);
             }
 
-            return new ConfigCliArguments(configFolderPath, outputFolderPath, sinceDate, untilDate, formats);
+            return new ConfigCliArguments(configFolderPath, outputFolderPath, sinceDate, untilDate, formats,
```
yamidark: Same as above

src/main/java/reposense/parser/ArgsParser.java
yong24s: Try this:

```
parser.addArgument("-view")
      .setConst(Paths.get(""))
      .nargs("?")
      .metavar("PATH")
      .type(new ReportFolderArgumentType())      
      .help("Starts a server to display the dashboard in the provided directory.");
```

`.setConst(Paths.get(""))` be used when no value is given.

yong24s: @myteo 

Please implement the above comment.

src/main/java/reposense/model/ConfigCliArguments.java
yong24s: @myteo 

Indentation for wrapped lines should be 8 spaces (i.e. twice the normal indentation of 4 spaces) more than the parent line. 
`From https://oss-generic.github.io/process/codingStandards/CodingStandard-Java.html#layout`

You can copy the above code snip exactly or refer to [this line](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/parser/AuthorConfigCsvParser.java#L78)

myteo: Got it. Thanks

myteo:
Changes made. Ready for review.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -151,6 +169,39 @@ public void parse_configFolderAndOutputDirectory_success() throws ParseException
...
+        Assert.assertTrue(Files.isSameFile(
+                REPO_CONFIG_CSV_FILE, ((ConfigCliArguments) cliArguments).getRepoConfigFilePath()));
+        Assert.assertTrue(Files.isSameFile(
+                AUTHOR_CONFIG_CSV_FILE, ((ConfigCliArguments) cliArguments).getAuthorConfigFilePath()));
+        Assert.assertTrue(cliArguments.isAutomaticallyLaunching());
+    }
+
+    @Test
+    public void parse_configFolderAndViewWithArgs_returnsViewCliArguments() throws ParseException, IOException {
+        String input = String.format("-config %s -view %s", CONFIG_FOLDER_ABSOLUTE, OUTPUT_DIRECTORY_ABSOLUTE);
```
yong24s: Hmm, this shouldn't be valid?

Line 197 implies that such command is allowed -> 
`java -jar RepoSense.jar -config path/to/configs -output . -view /etc/secrets`
Note: When `-output` is not provided, it will use `cwd` **.**

As -view is atomic, we can merge commands `1.` and `2.` to `3.`
1. `java -jar RepoSense.jar -config path/to/configs -output path/to/output`
2. `java -jar RepoSense.jar -view path/to/output`
3. `java -jar RepoSense.jar -config path/to/configs -output path/to/output -view`

Since  `1.` and `2.` are equivalent to `3.`, output and view should use the same `PATH`. 
It seems weird for `-view` to accept another `PATH`.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -190,6 +241,35 @@ public void parse_validGitRepoLocations_repoConfigurationListCorrectSize() throw
...
+        String input = String.format("-repos \"%s\" %s", TEST_REPO_REPOSENSE, TEST_REPO_DELTA);
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof LocationsCliArguments);
+        Assert.assertFalse(cliArguments.isAutomaticallyLaunching());
+    }
+
+    @Test
+    public void parse_repoLocationsAndViewWithArgs_returnsViewCliArguments() throws ParseException, IOException {
+        String input = String.format("-repos \"%s\" %s -view %s",
```
yong24s: Similar as above. 

`-view` should be a `flag` when used with `-config` and `-repos`.
When used alone, value `PATH` should be expected. `-view=PATH`

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -59,10 +59,14 @@ private static ArgumentParser getArgumentParser() {
                 .metavar("LOCATION")
                 .help("The GitHub URL or disk locations to clone repository.");
 
-        mutexParser.addArgument("-view")
+        parser.addArgument("-view")
+                .nargs("?")
                 .metavar("PATH")
                 .type(new ReportFolderArgumentType())
-                .help("Starts a server to display the dashboard in the provided directory.");
+                .setConst(Paths.get(""))
```
yong24s: `Paths.get("")` can be refactored to an expressive variable make `Line 122` more readable. :-P

src/test/java/reposense/parser/ArgsParserTest.java
yong24s: The main issue is `ViewCliArguments` is returned. So, there will not be any code analysis.

myteo: There is some issue with this because there is no easy way to differentiate these 2 commands:
1. `java -jar RepoSense.jar -config path/to/configs -view /etc/secrets`
2. `java -jar RepoSense.jar -view /etc/secrets`

This is because `-config` parameter has a default value which is the cwd. When we parse the command line arguments for both commands, the values for `-config` and `-view` are non-null. So it might be hard to detect when a command like the one below is given.
- `java -jar RepoSense.jar -config path/to/configs -output . -view /etc/secrets`

yong24s: @myteo 

Sorry I didn't follow the previous PR that closely, it seems that it was agreed to keep it like this. So, disregard this comment for now. 👍

src/test/java/reposense/parser/ArgsParserTest.java
myteo: > When used alone, value `PATH` should be expected. `-view PATH`

How about `java -jar RepoSense.jar -view`? Since running `java -jar RepoSense.jar` by default uses the config files in the cwd, is it reasonable for `java -jar RepoSense.jar -view` to have the same behaviour and in addition automatically launch the report?

myteo: In this case, it is possible to detect the following invalid command  ->
`java -jar RepoSense.jar -repos REPO_PATH_OR_URL -view /etc/secrets`
because `-repos` does not have a default value like `-config`.

Should I throw an exception in this case? On the other hand we might want the following 2 commands to exhibit similar behaviour:
1. `java -jar RepoSense.jar -repos repo_url -view /etc/secrets`
2. `java -jar RepoSense.jar -config path/to/configs -view /etc/secrets`

myteo: Sorry didn’t see your other comment:)

yong24s: > How about java -jar RepoSense.jar -view? Since running java -jar RepoSense.jar by default uses the config files in the cwd, is it reasonable for java -jar RepoSense.jar -view to have the same behaviour and in addition automatically launch the report?

Looks fine. Since they are equivalent `java -jar RepoSense.jar -config . -view` <=> `java -jar RepoSense.jar -view` 

> Should I throw an exception in this case? On the other hand we might want the following 2 commands to exhibit similar behaviour:

Don't worry about this comment, it can be done in another PR.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -190,6 +241,35 @@ public void parse_validGitRepoLocations_repoConfigurationListCorrectSize() throw
         Assert.assertEquals(2, repoConfigs.size());
     }
 
+    @Test
+    public void parse_repoLocationsAndView_returnsLocationCliArguments() throws ParseException {
+        String input = String.format("-repos \"%s\" %s -view", TEST_REPO_REPOSENSE, TEST_REPO_DELTA);
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof LocationsCliArguments);
+        Assert.assertTrue(cliArguments.isAutomaticallyLaunching());
```
yamidark: Let's do the assertion for `TEST_REPO_REPOSENSE` directory path here as well, since the test is both 'repo' and 'view'

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -190,6 +241,35 @@ public void parse_validGitRepoLocations_repoConfigurationListCorrectSize() throw
...
+    }
+
+    @Test
+    public void parse_repoLocationsOnly_success() throws ParseException {
+        String input = String.format("-repos \"%s\" %s", TEST_REPO_REPOSENSE, TEST_REPO_DELTA);
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof LocationsCliArguments);
```
yamidark: Same as above

src/test/java/reposense/parser/ArgsParserTest.java
myteo: I have updated the test case, please have a look.

docs/DeveloperGuide.md
```diff
@@ -69,7 +69,7 @@ Eslint and its accompaning modules can be installed through NPM, so do ensure th
 ### Building and running RepoSense from code
 
 1. Execute the following command on the OS terminal inside the project directory. <br/>
-Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | -repos REPO_PATH_OR_URL... | -view REPORT_FOLDER) [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...]"` <br/>
+Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | [-repos REPO_PATH_OR_URL...]) [-view REPORT_FOLDER] [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...]"` <br/>
```
eugenepeh: I think additional bracket is needed in `-view` to denote optional.
E.g. `[-view [REPORT_FOLDER]]`

eugenepeh: Should also add a sample usage in the dev guide, or at least append the `-view` to one of the sample usage.
Do also update the description such that the report will be open automatically.

docs/UserGuide.md
```diff
@@ -224,9 +227,11 @@ In addition, there are some _optional_ extra parameters you can use to customize
   Example:`-since 21/10/2017`
 * **`-formats LIST_OF_FORMATS`**: A space-separated list of file extensions that should be included in the analysis. Default: `adoc cs css fxml gradle html java js json jsp md py tag xml`<br>
   Example:`-formats css fxml gradle`
+* **`-view`**: A flag to launch the report automatically after processing. Note that if an argument is given to this parameter, no analysis will be performed and the report specified by the argument will be opened.<br>
```
eugenepeh: should also add in the arg for the `-view`.
Here is the breakdown of each fields, hence needs to be more descriptive.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -115,15 +120,19 @@ public static CliArguments parse(String[] args) throws ParseException {
 
             verifyDatesRangeIsCorrect(sinceDate, untilDate);
 
-            if (locations != null) {
-                return new LocationsCliArguments(locations, outputFolderPath, sinceDate, untilDate, formats);
+            if (reportFolderPath != null && !reportFolderPath.equals(EMPTY_PATH)) {
```
eugenepeh: not quite convinced by this... so if someone else type `-view .`, wouldn't it still runs the analysis since its pointing to CWD?

myteo: I thought so too at first and did some testing. Typing `-view .` does not cause `reportFolderPath.equals(EMPTY_PATH) == True`. I tried other inputs as well and didn't find any that is equals to `Paths.get("")`.

From https://docs.oracle.com/javase/8/docs/api/java/nio/file/Paths.html#get-java.lang.String-java.lang.String...- explaining the `Paths.get()` method:

> A Path representing an empty path is returned if first is the empty string and more does not contain any non-empty strings.

From https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html explaining empty Path:

> A Path is considered to be an empty path if it consists solely of one name element that is empty.

I think as long as there is an input, even if it is "." or cwd, the name element will not be empty. The documentation for `equals()` method of `Path` didn't specify exactly how `Path` instances are compared but it hints that that the hashCode is used. From https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#hashCode--:

> The hash code is based upon the components of the path

So I believe that any `Path` with non-empty name element will not be equals to an empty Path.

> Paths.get("") points to cwd

I believe this behaviour only occurs when accessing an empty path. From https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html:

> Accessing a file using an empty path is equivalent to accessing the default directory of the file system.

myteo: Basically it is true that Path.get("") points to CWD but I think Path.get("") is not equals to any non-empty path (eg. cwd or ".")

docs/DeveloperGuide.md
myteo: I have updated the developer guide, please have a look.

docs/UserGuide.md
myteo: This section of the UserGuide is under _Customizing the Analysis_, and is explaining optional extra parameters in addition to `-repo` parameter. Adding the description for the argument of `-view` here seems a bit out of place to me. There is another section describing the `-view` option called _Interpreting the Report_. Should I add it there instead?

eugenepeh: Sounds good to me.

src/main/java/reposense/parser/ArgsParser.java
eugenepeh: I see. In that case, let's add another test to keep this assumption in check (in case of api change in future).

docs/UserGuide.md
myteo: Actually this section [here](https://github.com/reposense/RepoSense/blob/release/docs/UserGuide.md#interpreting-the-report) already has some description for the argument of `-view`. Not quite sure what else I should add.

myteo:
Changes made. Ready for next review.

docs/UserGuide.md
eugenepeh: hmm, I think in that case, let's just add in this section for completeness sake because the following line mentions about the argument. So perhaps, let's put ```**`-view [REPORT_FOLDER ]`**```, and update the message :

Note that if ~an~ `the REPORT_FOLDER` argument is given to this parameter,

src/main/java/reposense/model/CliArguments.java
```diff
@@ -13,6 +13,7 @@
     protected Optional<Date> sinceDate;
     protected Optional<Date> untilDate;
     protected List<String> formats;
+    protected boolean isAutomaticallyLaunching;
```
yong24s: Please add this to `equals()`.

`this.isAutomaticallyLaunching == (otherCliArguments.isAutomaticallyLaunching)`

myteo: thanks for the catch, made the change.

src/main/java/reposense/RepoSense.java
```diff
@@ -53,6 +53,10 @@ public static void main(String[] args) {
                     formatter.format(ZonedDateTime.now(ZoneId.of("UTC+8"))));
 
             FileUtil.zip(cliArguments.getOutputFilePath().toAbsolutePath(), ".json");
+
+            if (cliArguments.isAutomaticallyLaunching()) {
+                DashboardServer.startServer(SERVER_PORT_NUMBER, (cliArguments.getOutputFilePath().toAbsolutePath()));
```
yong24s: Extra parentheses wrapping `(cliArguments.getOutputFilePath().toAbsolutePath())`

-> `DashboardServer.startServer(SERVER_PORT_NUMBER, cliArguments.getOutputFilePath().toAbsolutePath());`

yamidark:
Nits to proposed merge message:

```
In order for users to view the report using our app, they have
to use a separate -view command after generating it with a 
-config or -repo command. 

This is a hassle to the user as it can be joined together with the main
command to form an atomic action.

Let's update -view to be a flag for users to view the report
automatically after it has been generated using the main command,
while maintaining its original functionality.
```

PR #450, opened 2018-12-26
[#7][Enhancement] Generate collate files and allow users to download them

chelseyong:
Currently, RepoSense does not allow users to download their individual contribution. Graders may also have difficulty marking students' code because the original code is in json format, and additional processing may be needed for scripts which run quality checking.

Now, in addition to the commits and authorship json files, every repository folder contains a `collated` folder, which has all the individual collate-files (`name.md`) after file processing.

If the users themselves want to download a copy of the collate file, they may click on the download icon in the tabs section (below `Collapse All`).
downloadfile

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -37,8 +37,7 @@ html
                     a(v-else, v-on:click="expand(false)") Collapse all
                 v_authorship(
                   v-bind:key="generateKey(tabInfo.tabAuthorship)",
-                  v-bind:info="tabInfo.tabAuthorship",
-                  v-bind:path="getLink()")
+                  v-bind:info="tabInfo.tabAuthorship")
 
       template(v-else)
         .empty please enter a report directory or upload a report zip
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -37,7 +37,8 @@ html
                     a(v-else, v-on:click="expand(false)") Collapse all
                 v_authorship(
                   v-bind:key="generateKey(tabInfo.tabAuthorship)",
-                  v-bind:info="tabInfo.tabAuthorship")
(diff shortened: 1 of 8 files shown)


chelseyong:
Ready for review

eugenepeh:
Hi @chelseyong, thanks for making the effort to create this PR. 
However, an external fix has been implemented (https://github.com/reposense/convert-markdown) to convert the output files to markdown. Thus, not sure if Prof needs this feature anymore.

Do discuss with us before working on old issues, because the project has been evolving ever since, hence requirement may also have changed.

Anyway, for the feature to be useful, I think Prof would want all the markdown files to be generated at once and probably not one at a time.

@damithc please advise if this PR should be continued.

damithc:
> Do discuss with us before working on old issues, because the project has been evolving, hence some of the requirements may also have changed.

Yup.

> @damithc please advise if this PR should be continued.

While this feature is not a good fit for how we use collated files in CS2103 grading (we'll still need https://github.com/reposense/convert-markdown for that), this feature can be useful for some users.  So, I don't mind going ahead with it.

I wonder if this can be done using JS. For example, let's say user filtered out (or sorted) some files using some mechanism in the UI. Can the download button compile a file based on how the code appears currently in the UI?

eugenepeh:
>I wonder if this can be done using JS. For example, let's say user filtered out (or sorted) some files using some mechanism in the UI. Can the download button compile a file based on how the code appears currently in the UI?

If we dynamically create text file using the visible content in code view, i believe its possible. Hence, perhaps we should not stuff in the pre-created markdown files into the generated report folder, and create it dynamically.

chelseyong:
Okay noted. Will create a separate PR.

PR #448, opened 2018-12-22
CodeView: add selection to filter files by format #170

chelseyong:
#170 
What was suggested is to implement a search bar for users to type which file ext they prefer to see.
The number of file extensions written by a user is usually few, so I implemented checkboxes for user to choose the preferred file exts to view, this may be more intuitive to users.

Previously, the LOC count for each file type is implemented, I moved them nearer to the checkboxes so that it is more organized and less repetitive.
image

PR #451, opened 2018-12-29
[#450] Download files selected by user

chelseyong:
image
image

Fixes #450. This feature may be useful for some users who want to download certain files directly from the authorship tab.
User is allowed to select at least 1 checkbox beside the file chosen to download. After selecting them, press the download button at the top.
If no file is selected, the download button will be greyed out. 
File generated is .txt.


chelseyong:
Ready for review

frontend/src/static/js/v_authorship.js
```diff
@@ -156,6 +157,36 @@ window.vAuthorship = {
...
+      return text;
+    },
+
+    downloadAll() {
+      if (this.filesDownloaded.length === 0) {
+        return;
+      }
+      let text = '';
+      this.filesDownloaded.forEach((file) => {
+        text = this.injectText(text, file);
```
ongspxm: name this `injectFile` instead of injectText, cos `injectText` implies that all the formatting for the text is done in there, but the function more for text formatting for each file.

also, it would make more sense to create a `renderFile` function instead that returns the text for each file, then do all the appending in `downloadAll`, instead of passing the text to the function then passing it back out again

yamidark:
@chelseyong Do update your branch and let us know when you are ready for the next iteration of review by tagging @reposense/devs

chelseyong:

damithc:
> @damithc mentioned [here](https://github.com/reposense/RepoSense/pull/450#issuecomment-449974951) that it should compile a file based based on how the code appears currently in the UI.

yes, that's correct. It should download the code as shown in the current view. Is that hard to do?

chelseyong:
@damithc I don't get what you mean. Do you mean that if the hidden code is toggled, it should be included inside the collated file? Are non-authored lines included inside this file?

If so, every time the hidden code is toggled, I will have to recollate the file for users to download.

damithc:
> @damithc I don't get what you mean. Do you mean that if the hidden code is toggled, it should be included inside the collated file? Are non-authored lines included inside this file?

Actually, can you host a sample report with this feature for me to have a look? e.g., deploy to github pages of a repo

yamidark:
>Actually, can you host a sample report with this feature for me to have a look

@damithc We can use the netlify deploy on this PR to test this feature out.

damithc:
> @damithc I don't get what you mean. Do you mean that if the hidden code is toggled, it should be included inside the collated file? Are non-authored lines included inside this file?
> 
> If so, every time the hidden code is toggled, I will have to recollate the file for users to download.

Looks like the current behavior matches what I wanted 👍  We'll probably need to add a mechanism to _select/deselect all_ but that can be a separate PR.

frontend/src/index.jade
```diff
@@ -148,6 +153,8 @@ html
             .title(onclick="toggleNext(this)")
               span {{ file.path }}&nbsp;
               span.loc ({{ file.lineCount }} lines)
+            #downloadCheckbox
+              input(type="checkbox", :value="file", v-model="filesDownloaded")
```
eugenepeh: shouldn't `filesDownloaded` be `filesSelected`?

damithc:
> @damithc is your intention for the downloaded file to be in plain text or should we adopt a readable file format e.g. html or etc?

It's best to keep it as plain as possible,  but we still need some annotations to show which code belongs to which file, right?

frontend/src/index.jade
chelseyong: Are you referring to the naming of variable? `filesSelected` may conflict with PR #448 if merged.

eugenepeh: Yes. #448 can probably be `fileTypeSelected`? So there will be no conflict. 

Also, just a correction, since the value is `file`, it should correspondingly be `file...`

yamidark:
@damithc 

> but we still need some annotations to show which code belongs to which file, right?

Currently the codes between different files are separated using `########################################## file/path` at the top for each file's code.

However, each segment of code are all squeezed together (line 1-5 and 100-105 may be grouped together).

Should we have a way to separate segment of codes (like in `git diff` where each segment of code a certain distance away are separated, and with each section indicating the starting line number and number of lines), or leave it as it is?

@chelseyong Do resolve the conflicts and update your branch

damithc:
> Should we have a way to separate segment of codes (like in `git diff` where each segment of code a certain distance away are separated, and with each section indicating the starting line number and number of lines), or leave it as it is?

It would be good to have some indication of which line range is being cited. i.e., add some annotatation to separate each code segment

chelseyong:
Changes made:
If there are non-authored segments, `< Non-authored: Lines a-b >` will be added.
E.g.
```
Code

< Non-authored: Lines a-b >

Code
```

chelseyong:
Ready for review

damithc:
As the file contains only code authored by the person, it's enough to have something like this?

```
########################################## file/path
################# 24-45
....
################# 48-72
....
```

chelseyong:
Ready for review

chelseyong:
@yamidark oh, I may have misunderstood the suggestion. I placed the line numbers for those non-authored (white-coloured) segments instead.

chelseyong:
Ready for review

chelseyong:
@eugenepeh can you provide more info to the crashing? I can't seem to replicate the issue.

eugenepeh:
@chelseyong , you will need to try on someone with larger contribution. In my case, I did it on `Aditya` from `reposense_RepoSense_master`

fzdy1914:
> Had tried this several times, and result in crash every single time.
> 
> 1. Un-check and check a file type filter, e.g. java
> 2. Click the checkbox next to the files
> 3. Try to hide / unhide a file
> 
> The report crash every time when I tried to do this, can someone else please verify?
> Seems like the download and filter feature are not quite compatible with each other.

Tried and it crashs too.
My step is to hide the file all and clicking the file from the first to the last. After ticking some files, it will crash.

yong24s:
I'm still testing. I believe the following code-block is the reason of the crash, as vuejs went into endless rendering. 

At the moment, I have the it removed, but there's still highlight 🤔.

https://github.com/reposense/RepoSense/blob/ee21b0d8752adda7cffe547d490222c3dbd474a4/frontend/src/static/js/v_authorship.js#L209-L215

Test site: https://yong24s.github.io/publish-RepoSense/

chelseyong:
As this feature can create some unnecessary lagging and does not seem to be an important feature that has to be implemented in future releases, e.g. [convert-markdown](https://github.com/reposense/convert-markdown) is capable of converting code to markdown files already, this pull request will be closed.

eugenepeh:
Good effort, irregardless. It may be useful for future developers 👍

PR #477, opened 2019-01-14
Fix mismatch of file content #473

### LINKED ISSUE #473: File type doesn't match file content
Not sure what's going on here:

image

chelseyong:
Fixes #473 
```
When a file extension is unselected, the file content is not always updated 
despite a change in filename.

This is due to vuejs reusing the DOM for each file as a performance optimization. 

Let's add a key attribute to each file, so that vuejs can track every file correctly 
with proper reusing of the DOM. 
```


chelseyong:
Ready for review

PR #478, opened 2019-01-14
Open dashboard with code-view shown

### LINKED ISSUE #464: Open dashboard with code-view shown
Current: code-view is hidden at the initial loading of the report

Suggested: open the code view for the author at the top of the ramp chart

Reason: It gives a more representative view of the report, especially for someone seeing a RepoSense report for the first time.

apoorva17:
Fixes #464 
```
Code view is hidden upon initial loading of report.

Let's open the code view for the author at the top of the ramp chart
upon loading to provide a more representative view of the report,
especially for a user seeing a RepoSense report for the first time.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -38,6 +38,15 @@ html
                 v_authorship(
                   v-bind:key="generateKey(tabInfo.tabAuthorship)",
                   v-bind:info="tabInfo.tabAuthorship")
+              #tab-empty.tab-pane(v-else)
+                .title
+                  span To view the code attributed to a specific author, click the &nbsp;
+                  i.fas.fa-code
+                  span &nbsp; icon next to that author's name.
+                  br
+                  span To hide the code view, click the &nbsp;
+                  i.fas.fa-caret-right
+                  span &nbsp; icon on the left.
 
       template(v-else)
         .empty please enter a report directory or upload a report zip
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -24,7 +24,7 @@ window.app = new window.Vue({
     userUpdated: false,
 
     isCollapsed: false,
-    isTabActive: false,
+    isTabActive: true,
     isTabAuthorship: false,
     tabInfo: {},
     tabAuthorship: {},


damithc:
> wouldn't it be more preferable if its bind to the url instead?
> Reason: i can resume to the last author that i'm reviewing.

Yes, that is a good idea. Capturing code view configuration into the URL would be a separate feature though. The first step is to open the code view by default when launching the report for the first time, but show a dummy code view page to reduce loading time. The dummy page can have some instructions on how to view the code of a specific person.

apoorva17:
> Just a thought: to reduce loading time, we can load a dummy page for the code view at first rather than load the code of a specific person.

I am currently using information of the 1st author of the 1st repo which is already loaded at the time of the creation of the report, so I am not sure how much it would affect the loading time to use a dummy page instead.

damithc:
> I am currently using information of the 1st author of the 1st repo which is already loaded at the time of the creation of the report, so I am not sure how much it would affect the loading time to use a dummy page instead.

Are you sure? In the current version, loading the code view of a person takes an appreciable time even if it is the first person in the list; I'm not sure how the first one can be loaded by default without any additional delay.

apoorva17:
I've updated the implementation to show a dummy code view page with some instructions on how to view the code of a specific author.

apoorva17:

apoorva17:
@reposense/devs i've updated the branch ready for review

frontend/src/index.jade
```diff
@@ -38,6 +38,11 @@ html
                 v_authorship(
                   v-bind:key="generateKey(tabInfo.tabAuthorship)",
                   v-bind:info="tabInfo.tabAuthorship")
+              #tab-authorship.tab-pane(v-else)
+                .title
+                  span To get started, click on the &nbsp;
```
damithc: Change as follows?
~To get started ...~ To view the code attributed to a specific author, click the `</>` icon next to that author's name.

ongspxm: the empty tab shouldn't be id-ed `#tab-authorship`, maybe call it `#tab-empty` or something.

cos the idea is that, next time we can add new kinds of side display, say a diff view will have the id `#tab-commit`

eugenepeh:
Just came to my mind... now that the code view is opened by default, perhaps we should also inform the user that the code view can be hidden?

apoorva17:
> Just came to my mind... now that the code view is opened by default, perhaps we should also inform the user that the code view can be hidden?

Ok, I've added a line to inform the user that the code view can be hidden

Issue #468, opened 2019-01-07
Add per-commit granularity to the report


damithc:
Current: only `day` and `week` granularities are supported
![image](https://user-images.githubusercontent.com/1673303/50772699-9dff6e00-12c9-11e9-88b6-0de6da505779.png)

Suggestion: also support per commit granularity

I suspect this will need major changes in the backend

ongspxm:
:+1: i will make the changes to the frontend. so we will shift the consolidation of the day informaiton to frontend?

yamidark:
>so we will shift the consolidation of the day information to frontend?

I believe daily is still consolidated in the backend though, just that each day will contain an array of all the commits for that day.

ongspxm:
so it be like `[{
    sinceDate: 20190101,
    untilDate: 20190102
    insertion:0,
    deletion:0,
    commits:[{...}, {...}]
}]` ?

yamidark:
From my previous commit in the `per-commit-view`:
```
{
        "insertions": 408,
        "deletions": 96,
        "date": "2017-10-01",
        "commitResults": [
          {
            "hash": "353cbca2cca5671ddb0d358e4bf07ef838221a2a",
            "message": "[W09-B2] Added more Non Functional Requirements to the Developer Guide",
            "insertions": 7,
            "deletions": 2
          },
....
```

yamidark:
Closing this as the discussion moved to #468

PR #484, opened 2019-01-17
CommitsReporter: add support for per-commit-view

yamidark:
Updating the branch as we are working on #468


yamidark:
The dashboard for this currently doesn't work as we have changed the backend `json` report files without the frontend portion yet.

@ongspxm Can double check the new `commits.json` format if it is what you wish to work with, or if any changes need to be made.

yamidark:
>Looks fine to me, but why dont we continue using AuthorIntervalContribution.java and specify our own time / date range instead?

Hmmm, iirc, at that time after discussing with @ongspxm, we decided that the calculating of weekly interval contribution would be done in the frontend, so we went with this. If we plan to do shorter intervals, like for each 6 hours or so, we could also shift the calculation of that to the frontend, just we need to add the time as well to the `commit.json` report.

I also don't think we would want to store too much different intervals in the `commit.json` report, as it can get extremely large, which would make the loading of the page slower as well.

ongspxm:
okay, the format is something i can work with

ongspxm:
since its a side branch, im just gonna merge yah?

src/systemtest/resources/noDateRange/expected/reposense_testrepo-Alpha_master/commits.json
```diff
@@ -1,141 +1,134 @@
 {
-  "authorWeeklyIntervalContributions": {
+  "authorDailyContributionsMap": {
     "fakeAuthor": [
       {
         "insertions": 0,
         "deletions": 0,
-        "sinceDate": "2018-02-06",
```
ongspxm: keep the sinceDate, untilDate, its will be easier to generate the github link to show just that period

PR #486, opened 2019-01-18
[#463] Add sorting of ramp chart by repo contribution

### LINKED ISSUE #463: Ramp chart: allow sorting by total team contribution
Current: ramp chart can be sorted by individual contribution

image

Suggestion: also allow sorting by total team contribution so that more active teams can be identified easily

chelseyong:
Fixes #463.
```
Ramp charts can only be sorted by individual contribution.

Let's allow users to sort ramp charts by repo contribution so that more active repos 
can be identified.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -147,7 +147,7 @@ html
 
         .files(v-if="isLoaded")
           .empty(v-if="files.length===0") nothing to see here :(
-          .file.active(v-for="file in selectedFiles")
+          .file.active(v-for="file in selectedFiles", v-bind:key="file.path")
             .title(onclick="toggleNext(this)")
               span {{ file.path }}&nbsp;
               span.loc ({{ file.lineCount }} lines)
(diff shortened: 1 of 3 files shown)


eugenepeh:
Hmm not sure if this is what Prof means.

I think it should be sorted by repositories automatically when the `group` checkbox is checked?
need clarification from you, @damithc

damithc:
> I think it should be sorted by repositories automatically when the `group` checkbox is checked?
> need clarification from you, @damithc

This is a tricky situation. Normally, the sorting criterion is applied within the group, if `group` is ticked. As `Repo contribution` is applicable only when `group` is ticked, maybe auto-tick the `group` when that sort criterion is selected? I can't think of a clean solution because there are two sort levels (within group, and overall) and some sort criteria are linked to other settings.

chelseyong:
@damithc Currently, if it is not grouped by repo, sorting order will still be the same, the only difference being each ramp chart is labelled with the repo name.
image
So I think this is not very significant?
Another solution would be to disable the un-selection of the `group by repo` button when `repo contribution` is chosen as sorting criteria.

damithc:
@chelseyong is this PR still relevant?
There seems to be a way to sort groups by the contribution in the current version.
image

chelseyong:
Yes, it has been resolved in #639.

PR #488, opened 2019-01-19
ChartView: apply case-insensitive sorting sequence to `Author Name` #501

### LINKED ISSUE #501: Fix sorting of ramp charts by author names
Sorting by author names seems to be incorrect, as they are currently being sorted by capitalized display names, then those with non-capitalized names. This led to some author names starting with 'Z' to be placed in the middle.

E.g. When not grouped by repo, the sorting is as shown:
image

chelseyong:
Fixes #501.
```
Sorting by author is currently being sorted case-sensitively.

Let's apply case-insensitive sorting sequence to authors' name.
```


frontend/src/static/js/v_summary.js
```diff
@@ -382,7 +382,12 @@ window.vSummary = {
       });
 
       if (!this.filterGroupRepos) {
-        full[0].sort(comparator(ele => ele[this.filterSort]));
+        full[0].sort(comparator((ele) => {
+          if (typeof ele[this.filterSort] === 'string') {
+            return ele[this.filterSort].toLowerCase();
+          }
+          return ele[this.filterSort];
```
ongspxm: use inline if return and do the case on presence of `toLowerCase` (always call it if possible), using `typeof` to check is generally considered to be hacky.

```
ele => {
    const field = ele[this.filterSort];
    return field.toLowerCase ? field.toLowerCase() : field;
}
```

chelseyong:

yamidark:
@chelseyong Do open an issue 1st and give a full description of it 1st, before submitting a PR.

chelseyong:
> @chelseyong Do open an issue 1st and give a full description of it 1st, before submitting a PR.

Okay noted

yamidark:
Nits to proposed message:
```
ChartView's sorting option sorts the author's name in a
case-sensitive manner.

This causes author's name with upper-case characters to be
behind lower-case characters, which may not be intuitive to users
whom expect it to be in alphabetical order.

Let's apply case-insensitive sorting sequence for the author's name,
which will sort the author's name alphabetically instead.
```

@eugenepeh Can review the message?

eugenepeh:
@yamidark 

>Can review the message?

Looks fine to me, however the main reason should actually be 
```
By default, the display name in the report takes on the Author's GitHub
ID. As GitHub ID is case insensitively unique, there is no apparent
benefit in treating the uppercase and lowercase text as distinct.
```

PR #496, opened 2019-01-19
[#459] CLI: add args to choose any timezone 

### LINKED ISSUE #459: Allow user to choose any timezone
Current: Uses SGT timezone e.g., timestamp shown at the bottom of the report

image

Suggest: Allow user to choose any timezone

fzdy1914:
Fixes #459 

The generated time shown in the report is in the timezone of SGT
(UTC+8).

Due to time differences, it is inconvenient for global users to use the
timezone of Singapore in their report.

Let's set the default timezone to the system's one, and add an argument
in the command line to allow the users to choose a timezone to use.


yong24s:
[ZoneId.systemDefault().getId()](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html#getId--) may return `Z` for UTC (ISO-8601)

Resulting in this timestamp, see Netlify:
```
Generated by  RepoSense  on Sat Jan 19 08:55:46Z 2019
```

yong24s:
The reason of my previous comment was:
On the user expectations side, this change may be abrupt and unintuitive. 

Can we have a screenshot on how it will look like for `SGT/UTC+8`? 

Also how the input format like? Can you include a test?

fzdy1914:
I am still working on it.

fzdy1914:
The new time screenshot is like bellow:
image
Ignore the Chinese characters because my system language is Chinese.
In English version, it will be:
`Generated by  RepoSense  on Sun Jan 20 23:26:27+08:00 2019`

docs/DeveloperGuide.md
```diff
@@ -69,7 +69,7 @@ Eslint and its accompaning modules can be installed through NPM, so do ensure th
 ### Building and running RepoSense from code
 
 1. Execute the following command on the OS terminal inside the project directory. <br/>
-Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | [-repos REPO_PATH_OR_URL...]) [-view [REPORT_FOLDER]] [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...] [-isac | --ignore-standalone-config]"` <br/>
+Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | [-repos REPO_PATH_OR_URL...]) [-view [REPORT_FOLDER]] [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...] [-isac | --ignore-standalone-config] [-timezone UTC+/-HH]"` <br/>
```
yong24s: Following the specs of UTC from https://en.wikipedia.org/wiki/List_of_UTC_time_offsets.

It should be `-timezone UTC±[hh]:[mm], UTC±[hh][mm], or UTC±[hh]`

docs/DeveloperGuide.md
```diff
@@ -69,7 +69,7 @@ Eslint and its accompaning modules can be installed through NPM, so do ensure th
 ### Building and running RepoSense from code
 
 1. Execute the following command on the OS terminal inside the project directory. <br/>
-Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | [-repos REPO_PATH_OR_URL...]) [-view [REPORT_FOLDER]] [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...] [-isac | --ignore-standalone-config]"` <br/>
+Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | [-repos REPO_PATH_OR_URL...]) [-view [REPORT_FOLDER]] [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...] [-isac | --ignore-standalone-config] [-timezone UTC±[hh]]"` <br/>
```
yong24s: Strings like `UTC+0530` or `UTC+05:30` are valid as well.

So shall we use something like `-timezone UTC±hh[[:]mm]`?

src/main/java/reposense/parser/ZoneIdArgumentType.java
```diff
@@ -0,0 +1,26 @@
...
+
+import net.sourceforge.argparse4j.inf.Argument;
+import net.sourceforge.argparse4j.inf.ArgumentParser;
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+import net.sourceforge.argparse4j.inf.ArgumentType;
+
+/**
+ * Verifies and parses a string-formatted zone id to a {@code ZoneId} object.
+ */
+public class ZoneIdArgumentType implements ArgumentType<ZoneId> {
```
yong24s: If you move `private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("E MMM d HH:mm:ssXXX yyyy");` into this class, you can expose the argument as a String, `ArgumentType<String>`.

src/main/java/reposense/RepoSense.java
```diff
@@ -51,7 +50,7 @@ public static void main(String[] args) {
             RepoConfiguration.setFormatsToRepoConfigs(configs, cliArguments.getFormats());
             RepoConfiguration.setDatesToRepoConfigs(configs, cliArguments.getSinceDate(), cliArguments.getUntilDate());
             ReportGenerator.generateReposReport(configs, cliArguments.getOutputFilePath().toAbsolutePath().toString(),
-                    formatter.format(ZonedDateTime.now(ZoneId.of("UTC+8"))));
+                    formatter.format(ZonedDateTime.now(cliArguments.getZoneId())));
```
yong24s: Then you can simplify `formatter.format(ZonedDateTime.now(cliArguments.getZoneId())`
to `cliArguments.getZoneId()`.

However, I leave this as a **suggestion**. I'm okay as it is as this change may make testing harder.

docs/DeveloperGuide.md
fzdy1914: Good suggestion, it could be` -timezone UTC[±hh[mm]]`.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -442,4 +447,40 @@ public void parse_extraArgumentForIgnore_throwsParseException() throws ParseExce
...
         ArgsParser.parse(translateCommandline(input));
     }
+
+    @Test
+    public void parse_withTimezone_success() throws ParseException {
+        String input = String.format("-repos \"%s\" %s -timezone UTC+11", TEST_REPO_REPOSENSE, TEST_REPO_DELTA);
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof LocationsCliArguments);
+        Assert.assertEquals(ZoneId.of("UTC+11"), cliArguments.getZoneId());
```
yamidark: Extract the timezone String as a variable, and pass in into the input and assertion

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -442,4 +447,40 @@ public void parse_extraArgumentForIgnore_throwsParseException() throws ParseExce
...
+
+        input = String.format("-repos \"%s\" %s -timezone UTC+00", TEST_REPO_REPOSENSE, TEST_REPO_DELTA);
+        cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof LocationsCliArguments);
+        Assert.assertEquals(ZoneId.of("UTC"), cliArguments.getZoneId());
+    }
+
+    @Test
+    public void parse_withoutTimezone_success() throws ParseException {
```
yamidark: Don't think this is necessary, there are already lots of other tests without passing the timezone option in.
Instead, maybe we can put a test for adding timezone option but empty argument instead.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -442,4 +447,40 @@ public void parse_extraArgumentForIgnore_throwsParseException() throws ParseExce
...
+        String input = String.format("-repos \"%s\" %s", TEST_REPO_REPOSENSE, TEST_REPO_DELTA);
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof LocationsCliArguments);
+        Assert.assertEquals(ZoneId.systemDefault(), cliArguments.getZoneId());
+    }
+
+    @Test(expected = ParseException.class)
+    public void parse_incorrectTimezone_throwsParseException() throws ParseException {
+        String input = String.format("-config %s -timezone UTC+", CONFIG_FOLDER_ABSOLUTE);
```
yamidark: Let's use the `DEFAULT_MANDATORY_ARGS` for all this timezone tests as well

fzdy1914:
I would like to put this pr on hold until #526 is merged.

yamidark:
@fzdy1914 

>I would like to put this pr on hold until #526 is merged.

#526 has been merged, you may continue with this PR.

src/main/java/reposense/parser/ZoneIdArgumentType.java
```diff
@@ -0,0 +1,26 @@
...
+    private static final String MESSAGE_TIMEZONE_INVALID =
+            "The timezone provided is invalid, please use the format of UTC±[hh]";
+
+    @Override
+    public ZoneId convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
+        try {
+            return ZoneId.of(value);
```
yamidark: It is still possible for users to choose other timezone besides UTC (like GMT) though... Do we want to strictly enforce UTC (users don't need to type UTC at all)?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -500,4 +510,39 @@ public void parse_extraArgumentForIgnore_throwsParseException() throws ParseExce
...
+        Assert.assertEquals(ZoneId.of(zoneId), cliArguments.getZoneId());
+
+        zoneId = "UTC-1030";
+        input = DEFAULT_INPUT_BUILDER.addTimezone(zoneId).build();
+        cliArguments = ArgsParser.parse(translateCommandline(input));
+
+        Assert.assertTrue(cliArguments instanceof ConfigCliArguments);
+        Assert.assertEquals(ZoneId.of(zoneId), cliArguments.getZoneId());
+
+        input = DEFAULT_INPUT_BUILDER.addTimezone("UTC+00").build();
```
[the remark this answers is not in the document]
fzdy1914: Because the two string are different, one is `UTC+00` , another is `UTC`.

src/main/java/reposense/parser/ZoneIdArgumentType.java
fzdy1914: Actually, all zoneIds that can be recognized by `ZoneId` are OK, but we'd better just instruct our user to use `UTC`. Anyway, more users will not specify this arg and will use the default system time.

yamidark: >all zoneIds that can be recognized by `ZoneId` are OK, but we'd better just instruct our user to use `UTC`.

The thing is the UserGuide, DeveloperGuide and your help message as well all seems to suggest that only UTC timezone can be used, but in actual other timezones can be used as well.

fzdy1914: So, do you suggest that I need to ban all the non `UTC` usage of it?

fzdy1914: @yamidark 
Please kindly give your suggestion so I am able to progress with this pr.

docs/UserGuide.md
```diff
@@ -248,9 +248,11 @@ In addition, there are some _optional_ extra parameters you can use to customize
   Example:`--ignore-standalone-config` or `-i`
 * **`--view, -v [REPORT_FOLDER]`**: A flag to launch the report automatically after processing (`-v` as alias). Note that if the `REPORT_FOLDER` argument is given, no analysis will be performed and the report specified by the argument will be opened.<br>
 Example:`--view` or `-v`
+* **`--timezone, -t ZONE_ID`**: Indicates the timezone to be used for the generated report relative to UTC. Format: `UTC[±hh[mm]]` Default: system's default timezone.<br>
```
chelseyong: ...timezone relative to UTC which will be used for the generated report

docs/UserGuide.md
```diff
@@ -248,9 +248,11 @@ In addition, there are some _optional_ extra parameters you can use to customize
...
+  Example:`-timezone UTC+08`, `-timezone UTC-1030`
```
chelseyong: `--` here

src/main/java/reposense/parser/ZoneIdArgumentType.java
yamidark: @fzdy1914 Sorry I completely missed this. I think you could probably just update the docs/help messages to give `UTC` as an example, but allow them to choose whatever timezone they may want to give instead, as long as it is valid.

src/main/java/reposense/parser/ZoneIdArgumentType.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+public class ZoneIdArgumentType implements ArgumentType<ZoneId> {
+    private static final String MESSAGE_TIMEZONE_INVALID =
+            "The timezone provided is invalid, please use the format of UTC±[hh[mm]]";
+
+    @Override
+    public ZoneId convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
+        try {
+            if (!value.contains("UTC")) {
+                throw new DateTimeException(MESSAGE_TIMEZONE_INVALID);
```
eugenepeh: Seems like other timezone other than UTC were still banned.
I agree with @yamidark https://github.com/reposense/RepoSense/pull/496/files#r265014150 that we should allow other instances of timezone, and only mention the more popular ones such as UTC and GMT in our examples.

src/main/java/reposense/parser/ZoneIdArgumentType.java
```diff
@@ -0,0 +1,29 @@
...
+import net.sourceforge.argparse4j.inf.ArgumentParser;
+import net.sourceforge.argparse4j.inf.ArgumentParserException;
+import net.sourceforge.argparse4j.inf.ArgumentType;
+
+/**
+ * Verifies and parses a string-formatted zone id to a {@code ZoneId} object.
+ */
+public class ZoneIdArgumentType implements ArgumentType<ZoneId> {
+    private static final String MESSAGE_TIMEZONE_INVALID =
+            "The timezone provided is invalid, please use the format of UTC±[hh[mm]]";
```
eugenepeh: ... `please use a more widely accepted format. E.g. UTC±[hh[mm]]`

eugenepeh: On a side note, if `hh` is optional value, isn't `+-` the same?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -93,7 +97,7 @@ public void parse_help_throwsHelpScreenException() throws HelpScreenException, P
 
     @Test
     public void parse_allCorrectInputsAlias_success() throws ParseException, IOException, HelpScreenException {
-        String input = String.format("-c %s -o %s -s 01/07/2017 -u 30/11/2017 -f java adoc html css js -i -v",
+        String input = String.format("-c %s -o %s -s 01/07/2017 -u 30/11/2017 -f java adoc html css js -i -v -t UTC+08",
```
eugenepeh: for non-timezone unit tests, we can just use a standard value, e.g. `UTC+08`, and make it a constant.

fzdy1914:
> Hmm was expecting the value at the end of the dashboard to be more informative.
> image

The new timezone looks like below. 
image
If the user specifies the timezone to be `UTC-1030`, the timezone looks now:
image

yamidark:
@fzdy1914 Note the comment mentioned [here](https://github.com/reposense/RepoSense/pull/496#discussion_r265073874)

docs/DeveloperGuide.md
```diff
@@ -112,6 +112,10 @@ Named Arguments:
 --ignore-standalone-config, -i
                      A flag to ignore the standalone config file in
                      the repo.
+--timezone UTC[±hh[mm]], -t UTC[±hh[mm]]
```
yamidark: `--timezone ZONE_ID[±hh[mm]], -t ZONE_ID[±hh[mm]]`, since it is no longer just `UTC` alone

docs/DeveloperGuide.md
```diff
@@ -112,6 +112,10 @@ Named Arguments:
...
+                     The timezone to use for the generated report. 
+                     The timezone should be recognized by Java ZoneId class.
```
yamidark: The timezone should be recognized by Java ZoneId class. **e.g. UTC**

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -127,6 +129,15 @@ private static ArgumentParser getArgumentParser() {
                 .metavar("LOCATION")
                 .help("The GitHub URL or disk locations to clone repository.");
 
+        parser.addArgument(TIMEZONE_FLAGS)
+                .dest(TIMEZONE_FLAGS[0])
+                .metavar("UTC[±hh[mm]]")
```
yamidark: Update here as well `ZONE_ID[±hh[mm]]`

docs/DeveloperGuide.md
```diff
@@ -86,7 +86,7 @@ We use [Cypress](https://www.cypress.io/) for automated end-to-end front-end tes
 ### Building and running RepoSense from code
 
 1. Execute the following command on the OS terminal inside the project directory. <br/>
-Usage: `gradlew run -Dargs="([--config CONFIG_FOLDER] | [--repos REPO_PATH_OR_URL...]) [--view [REPORT_FOLDER]] [--output OUTPUT_DIRECTORY] [--since DD/MM/YYYY] [--until DD/MM/YYYY] [--formats FORMAT...] [--ignore-standalone-config]"` <br/>
+Usage: `gradlew run -Dargs="([--config CONFIG_FOLDER] | [--repos REPO_PATH_OR_URL...]) [--view [REPORT_FOLDER]] [--output OUTPUT_DIRECTORY] [--since DD/MM/YYYY] [--until DD/MM/YYYY] [--formats FORMAT...] [--ignore-standalone-config] [--timezone UTC[±hh[mm]]]"` <br/>
```
yamidark: Here as well `--timezone ZONE_ID[±hh[mm]]`

yamidark: This part hasn't been updated?

src/main/java/reposense/model/CliArguments.java
```diff
@@ -14,6 +15,11 @@
     protected Optional<Date> untilDate;
     protected List<Format> formats;
     protected boolean isAutomaticallyLaunching;
+    protected ZoneId zoneId;
```
yamidark: Add this to verify in the `equals` method as well

eugenepeh:
For the proposed message, try not to use words like `now`, `current` etc as it is all implied.

The second paragraph should emphasize more on the reason and rationale.
For example, you mentioned that 
```
This does not make sense for global users to use the timezone of Singapore
in their report.
```
It should be more explanatory and explains why. It can be due to inconvenience, time difference etc,

docs/DeveloperGuide.md
yamidark: Reminder here, `--timezone ZONE_ID[±hh[mm]]`

fzdy1914: Sorry for not seeing this comment. Have corrected it.

docs/DeveloperGuide.md
```diff
@@ -112,6 +112,10 @@ Named Arguments:
 --ignore-standalone-config, -i
                      A flag to ignore the standalone config file in
                      the repo.
+--timezone ZONE_ID[±hh[mm]], -t ZONE_ID[±hh[mm]]
+                     The timezone to use for the generated report. 
+                     The timezone should be recognized by Java ZoneId class. E.g. UTC
```
damithc: give a few more examples. An example should be _typical_; the one given is a special case.
The user is supposed to know ZoneId class? Let's know leak implementation details to the user. Is there any other way to specify the acceptable values?

fzdy1914: The zone-id class takes in four kinds of zone-ids which cannot be easily described in the UG or DG.
What I can do is give more examples. However, I would like to suggest the users use the UTC[±hh[mm]] because it is easier to use.

fzdy1914: It is also acceptable to ask the users to use UTC only.

eugenepeh: Perhaps instead of the second line `The timezone should be recognized by Java ZoneId class.`,
we can list some out for the users for e.g. `Example of ZONE_ID: UTC, GMT etc recognized by Java ZoneId class.`

damithc: Although four formats work, we don't have to explain all four. Just describe one? The other three can be a _hidden_ feature.

fzdy1914: @damithc Can you review the latest message?

docs/DeveloperGuide.md
```diff
@@ -112,6 +112,11 @@ Named Arguments:
 --ignore-standalone-config, -i
                      A flag to ignore the standalone config file in
                      the repo.
+--timezone ZONE_ID[±hh[mm]], -t ZONE_ID[±hh[mm]]
+                     The timezone to use for the generated report. 
+                     One kind of valid timezones is relative to UTC. 
+                     E.g. UTC, UTC+08, UTC-1030.
```
yamidark: @damithc Just to confirm, did you meant to give multiple example usages of `UTC` timezone, or to give multiple timezone examples that is accepted by us?

damithc: Multiple examples of UTC is enough 👍

yamidark:
Slight nits to message:
The generated time show**n** in the report is in the timezone of SGT
(UTC+8).

Due to ~~the~~ time difference**s**, it is inconvenient for global users to use the
timezone of Singapore in their report.

Let's set the default timezone to the system's one **,** and add an arg**ument** in **the**
command line to ~~give~~ **allow** the users ~~a choice~~ to choose a timezone to use.

PR #498, opened 2019-01-20
Show a 'loading...' message while loading the report for the first time #392

### LINKED ISSUE #392: Show a 'loading...' message while loading the report for the first time

emer7:
Fix #392 
Loading page should not be stuck again if file is invalid
```
RepoSense needs time to process the resources to generate report,
especially if there is large resources to generate large report.

However, RepoSense doesn't have any indication for user whether it
is loading or encountered error. This can cause confusion and frustation
for the user.

Let's implement a loading page to inform the user that the
application is loading and a pop-up to inform user if the application
fail to load the resources
```


emer7:
Ready for review

emer7:
Implemented pop-up, added `Fixes #Issue No.`, and also proposed git merge message

damithc:
Use a white background with a spinner? That way, the look is consistent with the other pages.

emer7:
> I got the pop-up when I loaded a zip file, but not when running a server though, can look into this?

I tested it and it shows. Perhaps trying hard reloading + clear cache will do? (I encountered it before when testing)
image

emer7:
> Use a white background with a spinner? That way, the look is consistent with the other pages.

Where do we implement a spinner? @ongspxm said we don't use it anymore?

yamidark:
```
VM120:7 Uncaught SyntaxError: Unexpected token } in JSON at position 178
    at JSON.parse (<anonymous>)
    at XMLHttpRequest.xhrOnload (api.js:20)
```

I removed a few lines in `summary.json` and got this error on the console, but no pop-up window. 
Probably need to catch the error when we fail to parse the `json` file.

frontend/src/static/js/api.js
```diff
@@ -6,17 +6,28 @@ const REPORT_DIR = '.';
...
-      .then(txt => JSON.parse(txt));
+    const zipObject = window.REPORT_ZIP.file(fname.slice(2));
+    if (zipObject) {
+      return zipObject
+          .async('text')
+          .then((txt) => {
+            try {
+              return JSON.parse(txt);
+            } catch (err) {
+              return Promise.reject(new Error('JSON file is invalid.'));
```
yamidark: Still don't see the pop-up though :thinking:
Seems like the code doesn't reach this `catch` section at all, perhaps we need a different way of catching it.

Anyway, since this is a `p.High` and we are taking too much time on this (and other delays), let's just remove this portion first and raise an issue afterwards to fix this error (can leave the other error handling portion).

emer7:
> I'm still getting a black background for loading? No spinner as well.
> Can try looking into [vue-spinner](https://madewithvuejs.com/vue-spinner) or [vue-loading-spinner](https://www.npmjs.com/package/vue-loading-spinner), or @ongspxm has some other suggestions for this?

Yeah I have not implemented the spinner. I am thinking of waiting for @ongspxm for the spinner thing

emer7:
Should I just go and implement the spinner, regardless there is spinner in the application now or not?

yamidark:
>Should I just go and implement the spinner, regardless there is spinner in the application now or not?

Yes, you can look into implementing 1 first so we can review it first

emer7:
Implemented spinner on white background
Review requested

frontend/src/static/js/api.js
yamidark: Remove this portion first, since it doesn't seem do anything right now (doesn't reach this portion of code apparently)

yamidark:
Nits to proposed mesesage:
```
RepoSense needs to load the resource files before it can generate
the dashboard, which can take a significant amount of time if the
resource files given are large.

However, RepoSense remains on the initial page, with the option to
upload a zip file, even after it has started loading these resource
files. This can be misleading to users as they may think they have
selected the wrong file and need to upload it again.

There is also no indication given if an invalid zip file was given,
which may make users wait indefinitely for the dashboard to load when
it won't.

Let's implement a loading page to inform the user that the application
is loading and a pop-up to inform user if the an invalid zip file is
given.
```
@yong24s Can help me review this?

PR #499, opened 2019-01-20
Split author level configurations from RepoConfiguration #457

### LINKED ISSUE #457: Split author level configurations from RepoConfiguration
Before #238, RepoSense only required 1 configuration file, which
holds all the repo and author level settings.

After splitting them into repo and author level configuration files in
#238, we still use the RepoConfiguration class to represent an author
configuration file, as most of the settings were still similar back
then.

However, after many changes to the configuration files, such as adding
commit exclusion settings in #261 and ignore standalone config
settings in #288 at repo level, the 2 configuration files now differ largely
from one another, making it harder to maintain the RepoConfiguration
class to represent them both, as any changes to 1 of the configuration file
will affect the other.

Let's create a new class AuthorConfiguration to represent the
author config file settings, splitting it from the RepoConfiguration.

myteo:
Fixes #457 

```
Repo and author level configuration files are both represented by
RepoConfiguration class.

This makes it harder to maintain as any changes to one of the
configuration file will affect the other.

Let's create a new class AuthorConfiguration to represent the
author config file settings, splitting it from the RepoConfiguration.
```


myteo:
This PR extracts author level config into `AuthorConfiguration`, `AuthorConfiguration` is still contained in `RepoConfiguration`. For the next step, I am thinking of further decoupling `RepoConfiguration` and `AuthorConfiguration` by creating a "container" class that contains `RepoConfiguration` and `AuthorConfiguration`. Move repo level config to `RepoConfiguration` and author level config to `AuthorConfiguration`.

myteo:
Ready for review.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -0,0 +1,87 @@
+package reposense.model;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+public class AuthorConfiguration {
```
yong24s: Can we have a class header to describe what this class represents?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+        addAuthorEmailsAndAliasesMapEntry(author, Arrays.asList(author.getGitId()));
+
+        addAuthorEmailsAndAliasesMapEntry(author, author.getAuthorAliases());
+        addAuthorEmailsAndAliasesMapEntry(author, author.getEmails());
+
+        setAuthorDisplayName(author, author.getDisplayName());
+    }
+
+    /**
+     * Propagate the {@code IgnoreGlobList} of {@code RepoConfiguration} to {@code author}.
```
yong24s: Propagate -> Propagates

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -47,6 +44,7 @@ public RepoConfiguration(RepoLocation location, String branch) {
 
     public RepoConfiguration(RepoLocation location, String branch, List<Format> formats, List<String> ignoreGlobList,
```
yong24s: Might as well do a class header for `RepoConfiguration` as well

myteo:
Changes made. Ready for review.

myteo:
> Also, on a related note, currently we initialize `author-config.csv` as a `RepoConfiguration` in the import phase. Can we possibly change to initialize as `AuthorConfiguration` and attach it to `RepoConfiguration` afterwards?

Should be possible. This means AuthorConfiguration will need to store `location` and `branch` as well. 

Eventually, I was thinking of separating `AuthorConfiguration` from `RepoConfiguration` because after this PR, `AuthorConfiguration` is still contained within `RepoConfiguration`. Maybe `AuthorConfiguration` and `RepoConfiguration` can merge to form a separate container class `Configuration`. Do you think this is an acceptable approach?

eugenepeh:
>Eventually, I was thinking of separating AuthorConfiguration from RepoConfiguration because after this PR, AuthorConfiguration is still contained within RepoConfiguration. Maybe AuthorConfiguration and RepoConfiguration can merge to form a separate container class Configuration. Do you think this is an acceptable approach?

Yes, that's fine. Since its meant to `Represents author configuration information from CSV config file for a single repository.`, it's totally okay to possess the location and branch info; it's a small cost to bear for better code quality.

myteo:
Made changes. Ready for review.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -77,25 +76,32 @@ public static void setDatesToRepoConfigs(
...
             }
 
-            int index = repoConfigs.indexOf(authorConfig);
+            RepoConfiguration matchingRepoConfig = null;
 
-            if (index == -1) {
+            for (RepoConfiguration repoConfig: repoConfigs) {
+                if (repoConfig.getLocation().equals(authorConfig.getLocation())
+                        && repoConfig.getBranch().equals(authorConfig.getBranch())) {
+                    matchingRepoConfig = repoConfig;
```
eugenepeh: perhaps extract this entire chunk as a static method that returns `matchingConfig` from List of `RepoConfiguration`?

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -105,6 +97,20 @@ public static void merge(List<RepoConfiguration> repoConfigs, List<AuthorConfigu
         }
     }
 
+    private static RepoConfiguration getMatchingRepoConfig(
+            List<RepoConfiguration> repoConfigs, AuthorConfiguration authorConfig) {
+        RepoConfiguration matchingRepoConfig = null;
```
eugenepeh: this is not necessary.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -105,6 +97,20 @@ public static void merge(List<RepoConfiguration> repoConfigs, List<AuthorConfigu
...
+
+        for (RepoConfiguration repoConfig: repoConfigs) {
+            if (repoConfig.getLocation().equals(authorConfig.getLocation())
+                    && repoConfig.getBranch().equals(authorConfig.getBranch())) {
+                matchingRepoConfig = repoConfig;
```
eugenepeh: can just return `repoConfig` to end the method, no need for break.

myteo: Oops sorry, my bad. Made the change.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+/**
+ * Represents author configuration information from CSV config file for a single repository.
+ */
+public class AuthorConfiguration {
+    public static final String DEFAULT_BRANCH = "HEAD";
+    private static final Logger logger = LogsManager.getLogger(AuthorConfiguration.class);
+
+    private RepoLocation location;
+    private String branch;
+    private String displayName;
```
yamidark: Is there a need for `AuthorConfiguration` to hold the `displayName`?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+
+    private transient List<Author> authorList = new ArrayList<>();
+    private transient TreeMap<String, Author> authorEmailsAndAliasesMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
yamidark: `TreeMap` -> `Map`, let's try to use interfaces as much as we can.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+
+        if (other == this) {
+            return true;
+        }
+
+        if (!(other instanceof AuthorConfiguration)) {
+            return false;
+        }
+
+        return hashCode() == ((AuthorConfiguration) other).hashCode();
```
yamidark: We can make `AuthorConfiguration` check for all of its value (instead of just location and branch, which is only used in the step to find the correct `RepoConfiguration` mapping)?

src/main/java/reposense/parser/AuthorConfigCsvParser.java
```diff
@@ -73,14 +73,14 @@ protected void processLine(List<RepoConfiguration> results, String[] elements)
...
      *
      * @throws InvalidLocationException if {@code location} is invalid.
      */
-    private static RepoConfiguration getRepoConfiguration(
-            List<RepoConfiguration> results, String location, String branch) throws InvalidLocationException {
-        RepoConfiguration config = new RepoConfiguration(new RepoLocation(location), branch);
+    private static AuthorConfiguration getAuthorConfiguration(
+            List<AuthorConfiguration> results, String location, String branch) throws InvalidLocationException {
+        AuthorConfiguration config = new AuthorConfiguration(new RepoLocation(location), branch);
         int index = results.indexOf(config);
```
yamidark: And here, instead of using `indexOf` to find an 'equal' `AuthorConfiguration`, we can just iterate through each `AuthorConfiguration` and check if the location and branch is the same, and return that config.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -76,28 +76,38 @@ public static void setDatesToRepoConfigs(
      * Merges a {@code RepoConfiguration} from {@code repoConfigs} with another from {@code authorConfigs}
      * if {@code location} and {@code branch} matches.
```
yamidark: Just noticed, `location` and `branch` is not mentioned anywhere in this method.
Can update this javadoc to `if their {@code Repolocation} and branch matches`.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -76,28 +76,38 @@ public static void setDatesToRepoConfigs(
...
-            if (index == -1) {
+            if (matchingRepoConfig == null) {
                 logger.warning(String.format(
                         "Repository %s is not found in repo-config.csv.", authorConfig.getLocation()));
                 continue;
             }
 
-            RepoConfiguration repoConfig = repoConfigs.get(index);
-            repoConfig.addAuthors(authorConfig.getAuthorList());
+            matchingRepoConfig.addAuthors(authorConfig.getAuthorList());
```
yamidark: Can change the `addAuthors` method to `setAuthorConfiguration`? 
Else we would be doing double work of setting all the values inside the `RepoConfiguration`'s `AuthorConfiguration` copy, when it was already done in `AuthorConfigCsvParser`.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -0,0 +1,205 @@
...
+            displayName = repoName + "_" + branch;
+        }
+    }
+
+    /**
+     * Clears authors information and use the information provided from {@code standaloneConfig}.
+     */
+    public void update(StandaloneConfig standaloneConfig) {
+        List<Author> newAuthorList = new ArrayList<>();
+        TreeMap<String, Author> newAuthorEmailsAndAliasesMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
yamidark: Likewise here, `TreeMap` -> `Map`

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -76,28 +76,38 @@ public static void setDatesToRepoConfigs(
...
         }
     }
 
+    private static RepoConfiguration getMatchingRepoConfig(
```
yamidark: Javadocs for this

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -16,6 +14,9 @@
 import reposense.system.LogsManager;
 import reposense.util.FileUtil;
 
+/**
+ * Represents configuration information from CSV config file for a single repository.
+ */
 public class RepoConfiguration {
```
yamidark: Can update `RepoConfiguration#equals(Object)` method as well to check for the `AuthorConfiguration`

src/main/java/reposense/model/RepoConfiguration.java
myteo: I think it is possible. One addition thing `addAuthors` does is that it propagates `IgnoreGlobList` to the authors so as long as `setAuthorConfiguration` does that it should be fine.

myteo: Another thing is that if `RepoConfiguration` already has existing authors then they will be lost in `setAuthorConfiguration` whereas `addAuthors` just adds on to the existing authors.

src/main/java/reposense/model/AuthorConfiguration.java
yamidark: Hmmm, maybe can make this portion print the repository location + branch, since `AuthorConfiguration` doesn't needs to have a displayName

src/main/java/reposense/model/RepoConfiguration.java
yamidark: > One addition thing addAuthors does is that it propagates IgnoreGlobList to the authors so as long as setAuthorConfiguration does that it should be fine.

We can have an additional call to the propogate `ignoreGlobList` method after setting the `authorConfig`.

>Another thing is that if RepoConfiguration already has existing authors then they will be lost in setAuthorConfiguration whereas addAuthors just adds on to the existing authors.

Thing is at this stage, there should not be any authors inside `RepoConfiguration` (just parsed from `repo-config.csv`) at all, which is why we should just be setting the `AuthorConfiguration` into it. This also ensures that each `RepoConfiguration` should only have a single `AuthorConfiguration` (something is wrong if `AuthorConfigurationCsvParser` returns 2 different`AuthorConfiguration` that matches with 1 `RepoConfiguration`).

myteo: I tried implementing this but ran into the issue below.

In `RepoConfiguration#merge`:
```
if (authorConfig.getLocation().isEmpty()) {
    for (RepoConfiguration repoConfig : repoConfigs) {
        repoConfig.addAuthors(authorConfig.getAuthorList());
    }
    continue;
}
```
If `AuthorConfiguration` has no location, the current logic will add the authors to all `repoConfigs`. I don't think we will be able to do this with `setAuthorConfiguration` because it will just override the authors.

`CsvParserTest#merge_emptyLocation_success` tests this behaviour as well so I think it is the intended behaviour.

myteo: We can work around by removing this portion of code and putting it in another for-loop after the current for-loop, so that `setAuthorConfiguration` will not override.

myteo:
Changes made accordingly. Ready for next round of review.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -169,7 +163,8 @@ public boolean equals(Object other) {
             return false;
         }
 
-        return hashCode() == ((RepoConfiguration) other).hashCode();
+        return hashCode() == ((RepoConfiguration) other).hashCode()
```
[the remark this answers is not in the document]
myteo: Also some non-transcient variables are not compared in `TestUtil#compareRepoConfig` while some transcient variables are compared. I changed by following exactly from `TestUtil#compareRepoConfig`.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -0,0 +1,187 @@
...
+
+    public AuthorConfiguration(RepoLocation location, String branch) {
+        this.location = location;
+        this.branch = location.isEmpty() ? DEFAULT_BRANCH : branch;
+    }
+
+    /**
+     * Clears authors information and use the information provided from {@code standaloneConfig}.
+     */
+    public void update(StandaloneConfig standaloneConfig) {
```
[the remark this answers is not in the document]
myteo: I am not sure if they are sufficient, but `RepoConfigurationTest` has some standalone config tests that tests `RepoConfiguration#update`, which calls `AuthorConfiguration#update`. It verifies the `AuthorConfiguration` as well.

myteo:
Ready for next review.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -0,0 +1,194 @@
...
+        return authorList.contains(author);
+    }
+
+    public void setAuthorList(List<Author> authorList) {
+        this.authorList = authorList;
+    }
+
+    /**
+     * Clears {@code authorEmailsAndAliasesMap} and {@code authorDisplayNameMap} and resets them with the details
+     * of authors in {@code authorList}
```
yamidark: `{code authorEmailsAndAliasesMap}` and `{@code authorDisplayNameMap}` and `{@code authorList}` is not visible to outside classes.
Can just mention that we are updating the author mapping information to its current author list (full stop at the end as well).

myteo: Updated.

PR #500, opened 2019-01-21
Add per-commit granularity to the report #468

### LINKED ISSUE #468: Add per-commit granularity to the report

ongspxm:
closes #468 

```
The dashboard only allows summary view for daily and weekly
granularities.

While the daily and weekly intervals give the reviewer the rough
impression of the authors' workload, it does not provide any insights
on the authors' commit frequency, commit organization and merged
changes. These information are important in helping users to further
understand the authors' workflow and contribution.

Let's implement summary view for per commit granularity. To do so,
let's update `commits.json` report to store the list of each
CommitResult inside the daily interval contribution data, and update
the frontend implement the per commit granularity using this data.

As the weekly interval view is now generated on the frontend based
on the daily view by #129, there is no longer a need to store the
weekly interval contribution data in commits.json.

Let's also remove the weekly interval contribution data in
`commits.json` report to reduce the size of the json files.
```


yamidark:
>Let's make the required changes to the report structure so as to
provide a summary view that gives a per commit granularity.

Can give more details on *what* are the required changes (my commit message already mention the backend changes, you can have a look at it)?

eugenepeh:
@ongspxm you need to rebase master n resolve the conflicts

yamidark:
Nits to proposed message:
```
The dashboard only allows summary view for daily and weekly
granularities.

However, users may also want to have a summary view for
per commit granularity as well.

Let's implement summary view for per commit granularity. To do so,
let's update `commits.json` report to store the list of each
CommitResult inside the daily interval contribution data, and update
the frontend implement the per commit granularity using this data.

As the weekly interval view is now generated on the frontend based
on the daily view by #129, there is no longer a need to store the
weekly interval contribution data in commits.json.

Let's also remove the weekly interval contribution data in
`commits.json` report to reduce the size of the json files.
```

.travis.yml
```diff
@@ -8,7 +8,7 @@ matrix:
 
 script:
   - ./config/travis/run-checks.sh
-  - cd ./frontend; npm run lint src/**/*js; cd ..
+  - cd ./frontend; npm run lint src/**/*js; cd ..;
```
yamidark: Why this change?

ongspxm: oh yeah, that was residue from the travis change, let me go fixit

yamidark:
@eugenepeh Can help review my message as well here?

frontend/src/index.jade
```diff
@@ -112,18 +113,33 @@ html
...
+                    template(v-for="(slice, j) in user.commits")
+                      a.summary-chart__ramp__slice(
+                        v-for="(commit, k) in slice.commitResults",
+                        v-if="commit.insertions>0",
+                        v-bind:title="`${slice.date} - ${commit.message}: ${commit.insertions} lines`",
+                        v-bind:class="'summary-chart__ramp__slice--color'+j%5",
+                        v-bind:style="{\
+                          zIndex: user.commits.length - j,\
+                          borderLeftWidth: getWidth(commit) + 'em',\
+                          right: ((getSlicePos(j, user.commits.length) + (getSlicePos(k, slice.commitResults.length)/user.commits.length))*100) + '%'\
```
eugenepeh: space around the operators?

frontend/src/static/js/v_summary.js
```diff
@@ -132,25 +132,26 @@ window.vSummary = {
       const newSize = 100 * (slice.insertions / this.avgCommitSize);
       return Math.max(newSize * this.rampSize, 0.5);
     },
-    getSliceTitle(slice) {
-      return `contribution on ${slice.sinceDate}: ${slice.insertions} lines`;
+    getSlicePos(i, total){
```
eugenepeh: space before curly brace `{`

eugenepeh:
image
image

Perhaps standardize the tool tip message format?
E.g. 
[2018-05-07] Daily contribution: 1 lines
[2018-05-07] Weekly contribution: 1 lines
[2018-05-07] COMMIT MESSAGE: 1 lines

eugenepeh:
>@eugenepeh Can help review my message as well here?

LGTM

ongspxm:
> Perhaps standardize the tool tip message format?

#516

yamidark:
@ongspxm Need to resolve your eslint errors as well (in #516)

yamidark:
@ongspxm Sorry just remembered, we were supposed to provide link to the GitHub commit for each ramp/commit right?

ongspxm:
That one i gonna add another sidebar view for it. Like click already, that
day's commit will be list in the side bar view.

ongspxm:
the side will show more than one, it is more of like a listing of the same
colored slices

ongspxm:
grouped by day

eugenepeh:
>the side will show more than one, it is more of like a listing of the same
colored slices

🤔 have you consulted prof regarding that? because if its a pure listing, I believe we can do that easily with github url functionality too. Example: https://github.com/reposense/testrepo-Alpha/commits/master?author=eugenepeh&since=2d87a431fcbb8f73a731b6df0fcbee962c85c250

damithc:
Based on my recent discussion with @ongspxm:

My basic requirement is to be able to click on a ramp and see the diff of that commit.

@ongspxm pointed out that it may be hard to click on a specific commit due to closes spacing of ramps (aka a 'hairy' ramp chart, to use @ongspxm's terminology :-p) and suggested a two-stage approach where we present the user with a list of commits first and the user chooses which one to view. In a future date, we can optimize such that the two-stage approach is used only when the ramp chart is hairy. In other cases, the user can click on the ramp to go to the diff view directly.

eugenepeh:
>@ongspxm pointed out that it may be hard to click on a specific commit due to closes spacing of ramps (aka a 'hairy' ramp chart, to use @ongspxm's terminology :-p) and suggested a two-stage approach where we present the user with a list of commits first and the user chooses which one to view. In a future date, we can optimize such that the two-stage approach is used only when the ramp chart is hairy. In other cases, the user can click on the ramp to go to the diff view directly.

I see. I thought we're going to solve that with a `magnifying glass`, `drag` or `zoom` view.

>My basic requirement is to be able to click on a ramp and see the diff of that commit.

It is unlikely we're able to do so without tapping on GitHub functionality. In that case, wouldn't it be easier for us to `present the user with a list of commits` using the GitHub functionality in the first place?

damithc:
> I see. I thought we're going to solve that with a `magnifying glass`, `drag` or `zoom` view.

Yes, that's the plan. But that feature is unlikely to be implemented soon, right? I suspect it is not easy to implement.

> It is unlikely we're able to do so without tapping on GitHub functionality. In that case, wouldn't it be easier for us to `present the user with a list of commits` using the GitHub functionality in the first place?

Yes, we need to look into that aspect. `show commit diff` is a basic functionality we should have one way or other.

eugenepeh:
>Yes, we need to look into that aspect. `show commit diff` is a basic functionality we should have one way or other.

The reason I mentioned that it is unlikely unless we tap on GitHub functionality is because this will requires us to retain all the repositories for the javascript to utilize in order to be able to retrieve code diff of any commits. So if our NFR is to be able to support > 50 big repos, we can't possibly host all that repositories on the GH Pages anymore as it would be too huge. 

Also, it might take extensive amount of time to do the research, trial and testing to get it to work using e.g. [isomorphic-git](https://github.com/isomorphic-git/isomorphic-git). Hence, I propose that we tap on the GitHub functionality for now and replace it with our own implementation in the future.

damithc:
> Also, it might take extensive amount of time to do the research, trial and testing to get it to work using e.g. [isomorphic-git](https://github.com/isomorphic-git/isomorphic-git). Hence, I propose that we tap on the GitHub functionality for now and replace it with our own implementation in the future.
> 
> So as for the `present the user with a list of commits`, perhaps we should also use the GitHub functionality? Modifying the container to show the `list` and removing it subsequently afterward (when the better alternatives are implemented) doesn't seems to be efficient.

Sounds good. I'm fine with using GitHub functionality if that is possible.

> The reason I mentioned that it is unlikely unless we tap on GitHub functionality is because this will requires us to retain all the repositories for the javascript to utilize in order to be able to retrieve code diff of any commits. So if our NFR is to be able to support > 50 big repos, we can't possibly host all that repositories on the GH Pages anymore as it would be too huge.

Note that we are only interested in the commits done by the authors in our list during the period we are analyzing. It is unlikely they wrote 50GB worth of code.

ongspxm:
@eugenepeh do we add that in now? the link for each commit to the github diff?

eugenepeh:
>@eugenepeh do we add that in now? the link for each commit to the github diff?

Don't have to be each commit since your proposition is to show all the commits within the day, can do it the similar way using GitHub (refer to above conversation). If it's not a major change, can add that in now.

PR #507, opened 2019-01-23
merging master into per-commit-view before #500 can go in

ongspxm:


eugenepeh:
@ongspxm please add the latest commit in too. It has a test-breaking change.

yamidark:
This should be going into the `per-commit-view` branch instead right?

PR #511, opened 2019-01-24
[#380] Prevent generate FileInfo from directory with valid name

### LINKED ISSUE #380: FileInfoExtractor#getAllFileInfo(...): may generate FileInfo for directories
https://github.com/reposense/RepoSense/blob/be94eedde7dd797793cd29b2af425a5d2e69f486/src/main/java/reposense/authorship/FileInfoExtractor.java#L166-L172

If we have a directory that has is name within the format white list, e.g. `.gradle` (mostly it is inside `.gitignore` so we didnt have this problem so far), it will actually have its `FileInfo` generated (as seen in our current code implementation above), which will cause the app to crash later when we try to `git blame` this directory.

We could append the 'else-if' white listed formats check to the directory check, so directories
won't have their `FileInfo` generated, or we could make `FileInfoExtractor#generateFileInfo(...)`
return a 'false' value so that we will not add any files which we have error reading into
the list to be analyzed.

emer7:
Fix #380 

```
RepoSense uses a list of file format as a whitelist to
filter files for analysis.

However, we use `String#endsWith` to test if both directories and
file's name is inside this whitelist to run the git-blame analysis.
Directories that are named with a file format inside this whitelist
are also added into the analysis, which causes RepoSense to crash as
git-blame fails when it is executed on a directory.

Let's fix this bug by excluding directories from the whitelist test.
```


src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -167,13 +167,13 @@ private static void getAllFileInfo(RepoConfiguration config, Path directory, Lis
...
                     getAllFileInfo(config, filePath, fileInfos);
-                }
-
-                if (Format.isInsideWhiteList(relativePath, config.getFormats())) {
-                    try {
-                        fileInfos.add(generateFileInfo(config.getRepoRoot(), relativePath));
-                    } catch (InvalidPathException ipe) {
-                        logger.warning(String.format(INVALID_FILE_PATH_MESSAGE_FORMAT, filePath));
+                } else {
+                    if (Format.isInsideWhiteList(relativePath, config.getFormats())) {
```
yong24s: Does it work if you use `continue` in 
```
if (Files.isDirectory(filePath)) {
    getAllFileInfo(config, filePath, fileInfos);	                    getAllFileInfo(config, filePath, fileInfos);
}
```
Otherwise, use `elseif`.

Also, let's have a test. You can add a branch that contains the required test files in `reposense/testrepo-Alpha` 😃

emer7: @yong24s It works.

And also added a branch https://github.com/reposense/testrepo-Alpha/tree/directory-with-valid-whitelisted-name and it passes the test

emer7:
Ready for review again

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -167,8 +167,8 @@ private static void getAllFileInfo(RepoConfiguration config, Path directory, Lis
                 String relativePath = filePath.toString().substring(config.getRepoRoot().length());
                 if (Files.isDirectory(filePath)) {
                     getAllFileInfo(config, filePath, fileInfos);
+                    continue;
                 }
-
```
yong24s: Unnecessary change

yong24s:
Since this is your first PR, let me help you a bit 😀 .

```
RepoSense accepts a list of file formats, which acts as a whitelist to
filter files for analysis.

However, we use `String#endsWith` to test if a directory/file's format
is in the whitelist. Directories that are named like a file are added
into the analysis, which cause git-blame to fail when it is executed
on a directory.

Let's fix this bug by excluding directories from whitelist test.
```

yong24s:
@emer7 

Make it a habit to keep your first post updated with the latest/final proposed message.

PR #513, opened 2019-01-25
ChartView: update code for list-rendering

chelseyong:
```
During list rendering, `v-for` is used with `v-if` which can increase
the file processing time, due to additional checks.

Let's follow the vue style guide[1] by replacing these instances with
computed properties, where necessary. Data is only updated when
changed, instead of being generated every time.

[1]: Vue style guide for avoiding use of v-if with for
https://vuejs.org/v2/style-guide/#Avoid-v-if-with-v-for-essential
```


chelseyong:

chelseyong:
Tried implementing your suggestion by hiding the files using `v-show` which uses _CSS-based toggling_ so data will remain in the DOM. However, it seems to lag more. :/

After some searching, storing `selectedFiles` inside computed properties seems to be a better option. `selectedFiles` will be cached and only recomputed when user selects/deselects checkboxes. This can improve the efficiency of the filtering files process.

Also, Vuejs discourages `v-for` and `v-if` used together (current implementation), as this is more inefficient and alike to mapping that iterates through every file for checking.

eugenepeh:
I see, great work for trying out alternatives. 

>Also, Vuejs discourages v-for and v-if used together (current implementation), as this is more inefficient and alike to mapping that iterates through every file for checking.

In that case, could multiple vue instances be the causation of the crash in #451?

chelseyong:
> In that case, could multiple vue instances be the causation of the crash in #451?

There are certainly some examples of `v-if` used with `v-for`, but these are inside the `v-summary` instance:
https://github.com/reposense/RepoSense/blob/c0f5bfd8603b6b6b8c766933b5e28c5abd2ce016/frontend/src/index.jade#L84
https://github.com/reposense/RepoSense/blob/c0f5bfd8603b6b6b8c766933b5e28c5abd2ce016/frontend/src/index.jade#L107-L108
Hence, they are not directly related to the crashing of the download feature. Perhaps I can open an issue about this, to improve vue's efficiency?

------
On my side, I have changed the files rendering to use computed properties. However, I noticed that using computed properties does not speed up the selection process significantly. In fact, if `v-for` is used with `v-if` (current implementation), the processing is actually faster. 

I can push these latest changes up for you to compare the speed difference.

eugenepeh:
Yes, it seems to be slower after each iteration and eventually crashed on the 4th. 

image
image
image

yong24s:
Can try to move the following code-block out of `updated()`?

The crash seems to be caused by vuejs going into an endless update.

https://github.com/reposense/RepoSense/blob/ee21b0d8752adda7cffe547d490222c3dbd474a4/frontend/src/static/js/v_authorship.js#L211-L213

Call it only when code view is open for a new author? 

It seems that the above code-block also tries to highlight code of other users as well 🤔.

yong24s:
I have removed `highlightjs` and its really fast now.
Test site: https://yong24s.github.io/RepoSense

yong24s:
> Nice, tested and barely feel any lag right now 👍.

But I haven't find out which variable to `watch` to invoke the `hightlightjs` 😅

yong24s:
If you toggle code views of different contribution, `highlightjs` will render after the first.

Even `updated()` has been removed https://github.com/yong24s/RepoSense/blob/gh-pages/static/js/v_authorship.js 🤔

yong24s:
My attempt to fix the lag, with comments: https://github.com/yong24s/RepoSense/commit/141a8e675aaef01b4f3fe38d7a1ee58240b7fd0c

Test-site: https://yong24s.github.io/RepoSense

Probably need to show `loading...` while `hljs` is rendering. 😅

eugenepeh:
>I have removed highlightjs and its really fast now.

`highlightjs` is meant to give context to the code... without that it is much harder to make sense out of jumbled codes.

yong24s:
> `highlightjs` is meant to give context to the code... without that it is much harder to make sense out of jumbled codes.

I meant to say I moved `hljs` out of the `Vue#updated`. 😓

chelseyong:
@yong24s By "hljs rendering", do you mean the toggling of hidden (or non-authored) code?
Because from the test-site preview, the highlighting seemed to work, except for the toggling.

yong24s:
> @yong24s By "hljs rendering" problems, do you mean the toggling of hidden (or non-authored) code?
> Because from the test-site preview, the highlighting seemed to work, except for the toggling.

It don't always work in the code view. I'm not sure where to best put the `hljs` code 😅.

ongspxm:
> It don't always work in the code view. I'm not sure where to best put the hljs code :sweat_smile:

one other way is to wrap the whole code thing into a component of its own and put the hljs inside the component, so one call will just be in charge of one "file"

yong24s:
@chelseyong

You want to work on the hljs optimization? I can create a new issue and assign it to you.

Let this PR be code view optimization only.

chelseyong:
@yong24s Okay can

yamidark:
@chelseyong Any progress on this PR?

yamidark:
Don't really see a different a response time when using the file type filter though.
Currently, hiding of the file types is very fast (almost instantaneous), but loading of the file types may take up some time (depending on the size).

From testing of Eugene in reposense_RepoSense_master (loading of file types):

Current master:
untitled

This PR:
untitled

Don't see much of a time difference between the two (and this PR actually may be slower!), and both still takes up a long time for scripting. 
Can try to look into how we can speed up the loading time of file types?

chelseyong:
@yamidark Okay, do you mean the loading in tab? 
image

yamidark:
>@yamidark Okay, do you mean the scripting in the loading of tab?

Nope, I meant the loading of all the files in the code panel, when after we hide all the files types and re-show them (uncheck and check the `All` box), which is supposed to be what this PR is working on right (according to the title)?
untitled

chelseyong:
Noted, the message was not clear before.

yamidark:
@chelseyong Well if you could improve the time for initial loading of all the codeview files as well, that will be good too 😄

chelseyong:
@eugenepeh Okay. May I know what do you refer to as cleaning up?

eugenepeh:
>@eugenepeh Okay. May I know what do you refer to as cleaning up?

just pretty much whatever you're doing now, following suggestion given in https://vuejs.org/v2/style-guide etc.

Let us know when you're done.

chelseyong:
@eugenepeh PR is ready

yamidark:
```
During list rendering, `v-for` is used with `v-if` which can increase
the file processing time, due to additional checks.

Let's follow the vue style guide[1] by replacing these instances with
computed properties, where necessary. Data is only updated when
changed, instead of being generated every time.

[1]: Vue style guide for avoiding use of v-if with for
https://vuejs.org/v2/style-guide/#Avoid-v-if-with-v-for-essential
```

@chelseyong Do add 'footnotes' in your message for when you want to provide additional info through a link

chelseyong:
@yamidark okay noted

Issue #502, opened 2019-01-22
Enforce ESLint check on travis


eugenepeh:
ESLint doesn't seems to be enforcing style check on any of the ci platforms. @ongspxm can you look into it?

ongspxm:
hao bah, let me check

emer7:
Just wondering, should we use CRLF or LF?

eugenepeh:
>Just wondering, should we use CRLF or LF?

I believe all our files use LF, there are shell script enforcing this in the repo.

PR #514, opened 2019-01-26
Enable ESLint JavaScript linter on travis #502

ongspxm:
```
ESLint doesn't seems to be enforcing style check on any of the ci
platforms. This is due to the glob that was passed into eslint, which
skips the actual js files.

Let's modify the glob string to run eslint on the *.js files.
```


yong24s:
Should we take this chance to write a simple script to test if ESLint is being enforced? 

For example, run ESLint on a JavaScript file with code quality issues then compare output, so we know if ESLint should stop working 😁.

```
$ eslint bad.js > actual && diff expected actual
```

ongspxm:
> Should we take this chance to write a simple script to test if ESLint is being enforced?

like the warnings? we cannot do errors, cos that might cos the travis job to go out.

but the problem previously was a glob problem, there is no point unless you plan to put `bad.js` into `frontend/src` which is kinda missing the point, cos not supposed to be in production.

any other ways?

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -38,12 +38,13 @@ public void setUp() throws IOException {
...
     }
 
+    /*
     @Test
     public void testNoDateRange() throws IOException, URISyntaxException, ParseException {
         generateReport();
         Path actualFiles = Paths.get(getClass().getClassLoader().getResource("noDateRange/expected").toURI());
         verifyAllJson(actualFiles, FT_TEMP_DIR);
-    }
+    }*/
```
eugenepeh: why is this test commented?

frontend/src/static/js/v_authorship.js
```diff
@@ -152,10 +150,10 @@ window.vAuthorship = {
 
     sortFileTypeAlphabetically(unsortedFilesInfoObj) {
       return Object.keys(unsortedFilesInfoObj)
-          .sort()
-          .reduce((acc, key) => ({
-              ...acc, [key]: unsortedFilesInfoObj[key]
-          }), {});
+        .sort()
```
eugenepeh: hmm... this doesn't seems to be right as its linewrapped

ongspxm: i put `MemberExpression: 2` already, gives me the same thing :disappointed:

yong24s:
> like the warnings? we cannot do errors, cos that might cos the travis job to go out.
> 
> but the problem previously was a glob problem, there is no point unless you plan to put `bad.js` into `frontend/src` which is kinda missing the point, cos not supposed to be in production.

As a good practice, we should check that our code quality tools are being enforced. May be not in this PR.

frontend/src/static/js/v_authorship.js
eugenepeh: Does the `MemberExpression` even affects it? If its not then probably its other rules.

eugenepeh:
Also, why do we have 2 copy of `.eslintrc.json` file? One at the root, and the other in the frontend folder. Do we need both?

eugenepeh:
@ongspxm do fix the conflict n request for review when ready

eugenepeh:
Nits to proposed message:

```
ESLint has been implemented, however it doesn't seems to be enforcing
style check on any of the CI platforms. 

This is due to the glob that was passed into eslint was incorrect which
causes it to skip the js files.

Let's modify the glob string to enable the javascript linter, and also
fixed the violations that were found.
```

PR #525, opened 2019-01-31
Update JavaScript dependencies used in the frontend #519

### LINKED ISSUE #519: Update JavaScript dependencies used in the frontend
The frontend depends on various JavaScript libraries namely, [vuejs](https://github.com/vuejs/vue/releases)   and [highlightjs](https://highlightjs.org/), which are often updated with improvements and fixes.

Let's update the JavaScript dependencies to the latest, so that the frontend performs optimally.

emer7:
Fix #519

```
The frontend depends on various JavaScript libraries,
namely, vuejs and highlightjs, which are often updated
with improvements and fixes.

Let's update the JavaScript dependencies to the latest
version, so that the frontend performs optimally.
```


ongspxm:
eh, most cdn got the "latest" link, but problem is sometimes those
dependencies might break how we use them.

emer7:
We can write a script that only take minor update according to semver from the CDN

can check this out https://api.cdnjs.com/libraries/vue?output=human

eugenepeh:
>edit: or write a script that only takes the latest one if it is a minor update according to semver

I guess easiest solution will be to download the latest minor as local resource, then our jade can just target local resources?

frontend/src/index.jade
```diff
@@ -1,18 +1,18 @@
...
+    link(rel="stylesheet", href="https://cdn.jsdelivr.net/npm/normalize.css@8/normalize.min.css")
+    link(rel="stylesheet", href="https://cdn.jsdelivr.net/npm/muicss@0/dist/css/mui.min.css")
+    link(rel="stylesheet", href="https://cdn.jsdelivr.net/npm/vue-loading-spinner@1/dist/vue-loading-spinner.css")
     link(rel="stylesheet", href="static/css/style.css")
+    link(rel="stylesheet", href="static/css/color-brewer.css")
 
     script(src="https://use.fontawesome.com/releases/v5.0.13/js/all.js", defer=True)
-    script(src="https://cdnjs.cloudflare.com/ajax/libs/vue/2.5.16/vue.min.js")
-    script(src="https://cdnjs.cloudflare.com/ajax/libs/jszip/3.1.5/jszip.min.js")
+    script(src="https://cdn.jsdelivr.net/npm/vue@2/dist/vue.min.js")
```
yong24s: Can we use point release? Like 2.6.X?

yong24s:
Also, let's [download a copy of highlightjs](https://highlightjs.org/download/) for our own use, the one hosted on the CDNs only has 24 commonly used languages, but we use additional languages like `adocs`, `gradle`, `plaintext`, `scss`, `yaml`.

@eugenepeh 
highlightjs supports 185 languages, but using all may slow down the frontend. 
Should we keep the it small and only use languages that are in our default formats? 🤔

Finally, let's not drag this PR.

eugenepeh:
>Should we keep the it small and only use languages that are in our default formats? 🤔

Preferably not, unless we can do it in a continuous integrate-able way. E.g. a self version updating url that allows us to download in every build.

emer7:
I just found out they provide cdn for each language that is not in the default list

yong24s:
@emer7 

Can pull/rebase master?

yong24s:
> Preferably not, unless we can do it in a continuous integrate-able way. E.g. a self version updating url that allows us to download in every build.

So for simplicity we download all languages to a local js? 🙆‍♂️

eugenepeh:
>So for simplicity we download all languages to a local js? 🙆‍♂️

As long as the download is scripted, I am fine with it.

emer7:
> > As long as the download is scripted, I am fine with it.
> 
> The most straightforward way requires us to add over 100+ lines to index.jade.
> 
> ```
> script(src="https://cdn.jsdelivr.net/gh/highlightjs/highlight.js@9.14/src/languages/1c.min.js")
> ...
> script(src="https://cdn.jsdelivr.net/gh/highlightjs/highlight.js@9.14/src/languages/zephir.min.js")
> ```
> Full list at https://cdn.jsdelivr.net/gh/highlightjs/highlight.js@9.14/src/languages/

oh wait we are really including everything?

yong24s:
@eugenepeh 

For confirmation, please.

eugenepeh:
Instead of doing that, perhaps do something like downloading all the files from the directory https://cdn.jsdelivr.net/gh/highlightjs/highlight.js@9.14/src/languages/, or look for a Gradle plugin that does that.

emer7:
> Instead of doing that, perhaps do something like downloading all the files from the directory https://cdn.jsdelivr.net/gh/highlightjs/highlight.js@9.14/src/languages/, or look for a Gradle plugin that does that.

I found gradle plugin to download all the files from that directory, but I just realize that aren't we need to load it one by one to index.jade?

eugenepeh:
>I found gradle plugin to download all the files from that directory, but I just realize that aren't we need to load it one by one to index.jade?

Hmm, the main highlight script does not handle the language importation?

emer7:
Nope

From https://highlightjs.org/usage/
> The CDN-hosted package doesn't have all the languages. Otherwise it'd be too big. If you don't see the language you need in the "Common" section, it can be added manually:
`<script charset="UTF-8" src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.12.0/languages/go.min.js"></script>`

eugenepeh:
then nvm as that would be too much work for this scope. In that case, just the common 24 and those listed in our default formats will do. 

Do arrange the scripts by domain to make it neater.

yong24s:
@emer7 

highlightjs stopped working after one of the last few commits. 😱

emer7:
> @emer7
> 
> highlightjs stopped working after one of the last few commits. 😱

What I found:
1. `cdn.jsdelivr.net/npm/highlight.js@9.14/lib/languages/xxx.js` doesn't work because it uses CommonJS (module.exports)
2. `cdn.jsdelivr.net/gh/highlightjs/highlight.js@9.14/src/languages/xxx.js` works, but it doesn't have minified version
3. `cdnjs.cloudflare.com/ajax/libs/highlight.js/9.14.2/languages/xxx.min.js` works, but it doesn't allow version range (9.14.2 instead of 9.14)

so which one to use?

yong24s:
😢 I still use 2nd option with `.min`. 

Maybe we can open an issue upstream at https://github.com/terser-js/terser/.

emer7:
@yong24s found a working CDN!

Issue #479, opened 2019-01-14
ArgsParser: Extract magic strings into constants


myteo:
Mentioned in https://github.com/reposense/RepoSense/pull/437#discussion_r242410159: in ArgsParser.java, ``"config"`` (and other similar strings) is a magic string repeated in multiple places and should be defined in a central location as a constant.

yong24s:
Edited issue to include original comment.

fzdy1914:
Close via #526

PR #526, opened 2019-01-31
ArgsParser: extract magic strings and follow a widely-used standard for CLI

### LINKED ISSUE #449: ArgsParser: Should follow a widely-used standard for CLI
```
As RepoSense is designed for CLI usage, it is a good idea to follow a
widely-used standard for CLI to match user expectations.

According to [1], `long-named options` should be prefix with double
`--` and should have an equivalent `single-letter options`;

i.e.  --config and its equivalent -c
      --repo and its equivalent -r

For completeness to GNU Standards [1], we may want to add options
`--verbose` and `--version`.

[1] GNU Coding Standards: Command Line Interfaces
https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html
```

fzdy1914:
Fix #449, #479 

In ArgsParser.java, the strings of flags for CLI are magic strings repeated in 
multiple places. Meanwhile, the standard of CLI argument we use now does 
not follow a widely-used standard.

We should avoid the uses of magic strings and follow a widely-used 
standard for CLI to match user expectations.

Let's define the magic strings in a central location as constants and uses 
`--` prefix for `long-named flags` and `-` for its equivalent `single-letter alias`.


src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -31,6 +31,30 @@
...
+    private static final String OUTPUT_FLAG = "output";
+    private static final String OUTPUT_FLAG_ALIAS = "o";
+    private static final String SINCE_FLAG = "since";
+    private static final String SINCE_FLAG_ALIAS = "s";
+    private static final String UNTIL_FLAG = "until";
+    private static final String UNTIL_FLAG_ALIAS = "u";
+    private static final String FORMAT_FLAG = "formats";
+    private static final String FORMAT_FLAG_ALIAS = "f";
+    private static final String IGNORE_FLAG = "ignore-standalone-config";
+    private static final String IGNORE_FLAG_ALIAS = "isac";
```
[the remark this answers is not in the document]
fzdy1914: As we may have much more CLI arguments in the future,  we may have some args with same single letter alias, maybe we just use -isac instead of -i because it is already an abstraction?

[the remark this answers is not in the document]
yong24s: Depends on how closely and faithfully are we sticking to POSIX standards. 🤔 

In POSIX, the usage of single letter can be together `i.e.` `-abcd` is equivalent as `-a -b -c -d`.

You won't run out of letters to use, but this may be the reason why single letter options are non-intuitive. 🤣

docs/DeveloperGuide.md
```diff
@@ -69,7 +69,41 @@ Eslint and its accompaning modules can be installed through NPM, so do ensure th
...
 1. Execute the following command on the OS terminal inside the project directory. <br/>
-Usage: `gradlew run -Dargs="([-config CONFIG_FOLDER] | [-repos REPO_PATH_OR_URL...]) [-view [REPORT_FOLDER]] [-output OUTPUT_DIRECTORY] [-since DD/MM/YYYY] [-until DD/MM/YYYY] [-formats FORMAT...] [-isac | --ignore-standalone-config]"` <br/>
+Usage: `gradlew run -Dargs="([--config CONFIG_FOLDER] | [--repos REPO_PATH_OR_URL...]) [--view [REPORT_FOLDER]] [--output OUTPUT_DIRECTORY] [--since DD/MM/YYYY] [--until DD/MM/YYYY] [--formats FORMAT...] [--ignore-standalone-config]"` <br/>
+Named Arguments: 
+```
+--help, -h           Show help message.
+--view [PATH], -v [PATH]
+                     Starts a server to  display  the  dashboard in the
+                     provided directory.If  used  as  a  flag  (with no
+                     argument), generates  a  report  and automatically
```
yamidark: Extra whitespaces (more than 1) between words throughout each description (missing a whitespace in `directory.If` here though

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -47,8 +47,33 @@
 
     @Test
     public void parse_allCorrectInputs_success() throws ParseException, IOException {
-        String input = String.format("-config %s -output %s -since 01/07/2017 -until 30/11/2017 "
-                + "-formats java adoc html css js --ignore-standalone-config -view",
+        String input = String.format("--config %s --output %s --since 01/07/2017 --until 30/11/2017 "
+                + "--formats java adoc html css js --ignore-standalone-config --view",
```
yamidark: Still using magic strings throughout the test though?

fzdy1914: I am not so sure if it is better to substitute all the string here. Because, if we do so, if there is a typo in the flags, we are unable to find it out by tests. Also, it is not so intuitive from my perspective.

yamidark: >Because, if we do so, if there is a typo in the flags, we are unable to find it out by tests.

If we use the variable as given in `ArgsParser` why would there be a typo when they are exactly the same? Rather, now since we are still using magic Strings here, if in the future we do change the option name in `ArgsParser` we would need to update it here as well (like what you are doing now, converting from single `-` to double `--`), which is inconvenient.

>Also, it is not so intuitive from my perspective.

How would it become not intuitive? We could separate the creation of each option and its argument, and then finally join them together using `String.join`, before running the parsing test, which I think would make it more readable as well.

@yong24s Your opinion on this?

yong24s: Clearly, like you say, using constants would be easier to maintain. Also, we are unlikely to make  typos since we have all the peer reviews. 👀

But we are unlikely to do such a major change to the flags again. 🤔 Also the only benefit of using magic string is its easier to read off the tests.

Either way are fine for me. Ask @eugenepeh too. 😬

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -42,52 +52,59 @@ private static ArgumentParser getArgumentParser() {
...
                 .setDefault(EMPTY_PATH.toAbsolutePath())
-                .help("The directory containing the config files."
+                .help("The directory containing the config files. "
                         + "If not provided, the config files will be obtained from the current working directory.");
 
-        mutexParser.addArgument("-repo", "-repos")
+        mutexParser.addArgument(REPO_FLAGS)
                 .nargs("+")
-                .dest("repos")
+                .dest(REPO_FLAGS[0])
```
eugenepeh: doesn't this change the `repos` to `--repos`?

src/test/java/reposense/parser/ArgsParserTest.java
eugenepeh: If that's the concern, let's make a test util which builds the command string.

src/main/java/reposense/parser/ArgsParser.java
fzdy1914: Yes, it does. However, the way we get the result is also changed. 
`List<String> locations = results.get(REPO_FLAGS[0]);`
This is used to avoid the magic string.

src/test/java/reposense/parser/ArgsParserTest.java
fzdy1914: Because there are many combinations of args, can I know what kind of function do you expect to have in test Util? For example, input and output? 
I am considering the suggestion by yamidark:
```
How would it become not intuitive? We could separate the creation of each option and its argument, and then finally join them together using `String.join`, before running the parsing test, which I think would make it more readable as well.
```

yamidark: @fzdy1914 Can you try my suggestion 1st, maybe for the new tests 1st, then we can review and see if we want to change the others as well.

eugenepeh: >Because there are many combinations of args, can I know what kind of function do you expect to have in test Util? For example, input and output?

E.g. `new InputBuilder().addConfig("...").addSince("01/07/2017").build()` ... etc. Its basically a StringBuilder in the backend with concept similar to the builders in [AB4](https://github.com/se-edu/addressbook-level4/tree/master/src/test/java/seedu/address/testutil)

>@fzdy1914 Can you try my suggestion 1st, maybe for the new tests 1st, then we can review and see if we want to change the others as well.

We can do that too, whichever provides better readability.

fzdy1914:
I have finished the InputBuilder part and all the test are updated except the ones with the alias.
There are two ways to write the function. First one is to just add another method called `addConfigAlias`. The second is to use method overload that adds another method with the same name but to have arg to indicate with FlAG to use, i.e. `addConfig(PATH, 1)` will use the second flag in CONFIG_FLAGS.

src/test/java/reposense/parser/ArgsParserTest.java
fzdy1914: I have implemented the InputBuilder part, please kindly give a review and give suggestions about the alias method mentioned below.
Also, please kindly give a suggestion to improve the test below:
https://github.com/reposense/RepoSense/pull/526/files#diff-aca26ed6b596a7c94f290b9c15f01597R109

eugenepeh: since its a testutil, doesn't hurt to add a method such as `addWhiteSpace(int)` etc to help you with that test.

fzdy1914: @eugenepeh How about the alias builder?

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -65,7 +66,7 @@ private void generateReport(String inputDates) throws IOException, URISyntaxExce
         Path configFolder = Paths.get(getClass().getClassLoader().getResource("repo-config.csv").toURI()).getParent();
 
         String formats = String.join(" ", TESTING_FILE_FORMATS);
-        String input = String.format("-config %s -formats %s ", configFolder, formats) + inputDates;
+        String input = new InputBuilder().addConfig(configFolder).addFormats(formats).add(inputDates).build();
```
yamidark: Let's have some line breaks here after adding each option and arugment, to make it more readable:
```
new InputBuilder().addConfig(configFolder)
        .addFormats(formats)
        .add(inputDates)
        .build();
```

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -166,7 +167,7 @@ public void merge_twoRepoConfigs_success() throws ParseException, IOException {
         expectedConfig.addAuthorEmailsAndAliasesMapEntry(SECOND_AUTHOR,  Arrays.asList("Zachary Tang"));
         expectedConfig.setIgnoreGlobList(REPO_LEVEL_GLOB_LIST);
 
-        String input = String.format("-config %s", TEST_CONFIG_FOLDER);
+        String input = new InputBuilder().addConfig(TEST_CONFIG_FOLDER).build();
```
yamidark: Can leave these which only add in 1 option and argument

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        }
+        return this;
+    }
+
+    public InputBuilder addOutput(Path path) {
+        input += ArgsParser.OUTPUT_FLAGS[0] + WHITESPACE + path + WHITESPACE;
+        return this;
+    }
+
+    public InputBuilder addSince(String date) {
```
yamidark: `addSinceDate`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        input += ArgsParser.SINCE_FLAGS[0] + WHITESPACE + date + WHITESPACE;
+        return this;
+    }
+
+    public InputBuilder addUntil(String date) {
```
yamidark: `addUntilDate`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        input += ArgsParser.UNTIL_FLAGS[0] + WHITESPACE + date + WHITESPACE;
+        return this;
+    }
+
+    public InputBuilder addFormats(String formats) {
+        input += ArgsParser.FORMAT_FLAGS[0] + WHITESPACE + formats + WHITESPACE;
+        return this;
+    }
+
+    public InputBuilder addIgnore() {
```
yamidark: `addIgnoreStandaloneConfig`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,92 @@
...
+        return input;
+    }
+
+    public InputBuilder addHelp() {
+        input += ArgsParser.HELP_FLAGS[0] + WHITESPACE;
+        return this;
+    }
+
+    public InputBuilder addConfig(Path path) {
+        input += ArgsParser.CONFIG_FLAGS[0] + WHITESPACE + path + WHITESPACE;
```
yamidark: Hmm, this actually allows us to add the config option multiple times (and for the other options below).
Maybe we can make `InputBuilder` hold a `boolean` flag to indicate if an option has been 'set', then only add it in when we call `build` if it has been 'set', else we don't

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -202,7 +204,7 @@ public void repoConfig_withFormats_ignoreCliFormats() throws ParseException, IOE
     @Test
     public void repoConfig_withoutFormats_useCliFormats() throws ParseException, IOException {
         String formats = String.join(" ", CLI_FORMATS);
-        String input = String.format("-config %s -formats %s", WITHOUT_FORMATS_TEST_CONFIG_FILES, formats);
+        String input = new InputBuilder().addConfig(WITHOUT_FORMATS_TEST_CONFIG_FILES).addFormats(formats).build();
```
yamidark: This as well:
```
new InputBuilder().addConfig(WITHOUT_FORMATS_TEST_CONFIG_FILES)
        .addFormats(formats)
        .build();
```

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -172,7 +173,8 @@ public void repoConfig_ignoresStandaloneConfigInCli_success() throws ParseExcept
     public void repoConfig_wrongKeywordUseStandaloneConfig_success()
             throws ParseException, GitCloneException, IOException {
         String formats = String.join(" ", CLI_FORMATS);
-        String input = String.format("-config %s -formats %s", IGNORE_STANDALONE_KEYWORD_TEST_CONFIG_FILES, formats);
+        String input =
+                new InputBuilder().addConfig(IGNORE_STANDALONE_KEYWORD_TEST_CONFIG_FILES).addFormats(formats).build();
```
yamidark: This as well:
```
new InputBuilder().addConfig(IGNORE_STANDALONE_KEYWORD_TEST_CONFIG_FILES)
       .addFormats(formats)
       .build();
```

src/test/java/reposense/parser/RepoConfigurationTest.java
```diff
@@ -188,7 +190,7 @@ public void repoConfig_wrongKeywordUseStandaloneConfig_success()
     @Test
     public void repoConfig_withFormats_ignoreCliFormats() throws ParseException, IOException {
         String formats = String.join(" ", CLI_FORMATS);
-        String input = String.format("-config %s -formats %s", FORMATS_TEST_CONFIG_FILES, formats);
+        String input = new InputBuilder().addConfig(FORMATS_TEST_CONFIG_FILES).addFormats(formats).build();
```
yamidark: This as well:
```
new InputBuilder().addConfig(FORMATS_TEST_CONFIG_FILES)
        .addFormats(formats)
        .build();
```

src/test/java/reposense/parser/ArgsParserTest.java
eugenepeh: its fine to leave it as it is, no point fine graining it

src/test/java/reposense/util/InputBuilder.java
fzdy1914: I believe we are writing a test util class which will not be used in the main code. If there are misuses of this InputBuilder,  ParseException will be raised when executing the input. So I believe there is no need of making sure InputBuilder returns a valid string.

fzdy1914:
Ready for review, need to mention that the content in ` ArgsParserTest.java` is not loaded by default.

src/test/java/reposense/util/InputBuilder.java
yamidark: Yes, this is a test util class. But by ensuring it gives a valid output, we minimize errors on our part (especially for newer developers), such as using the same `InputBuilder` to build 2 strings but the 2nd string giving an incorrect string, and ensures that any test failures that occurs is due to a bug in the main code, and not misusing our test code.

fzdy1914: I want to clarify one point. If a user adds the same flag multiple times, do you expect it to throw an exception, all just output the latest one that is added?

fzdy1914: If it throws an exception, do I need to write a new exception class for it?

yong24s: This behavior should be handled by the shell. So, I don't think our program would know if a flags is repeated. 🤔 

`i.e.` `./gradlew run -Dargs="-repo . -repo . -repo . -view -view -view"` or `ls -l -l -l -la -l`.

Also, we use Apache's `translateCommandline()` to mimic shell-like behavior in `ArgsParserTest`. Does it not work?

yamidark: >I want to clarify one point. If a user adds the same flag multiple times, do you expect it to throw an exception, all just output the latest one that is added?

Hmm, did some testing with `-config` option, seems like it will parse all the options, then give you the last 1 specified. However, I'm not sure if we want to add tests for this.
@yong24s @eugenepeh Your opinions on this?

yamidark: @fzdy1914 Anyway, since our `ArgsParser` allows for an option to be specified multiple times, we can keep the `InputBuilder` as it is now

yong24s: I was about to say that previous comment may not be actually correct. 

But running these on `Bash` yields different result than yours. 🤷‍♂️ 
`./gradlew run -Dargs="-repo a.git -repo b.git -repos c.git"` -> takes `c.git`
`./gradlew run -Dargs="-config a -config b"` -> takes `a`

If is behavior is caused by the shell then there's no point testing it?

yamidark: >If is behavior is caused by the shell then there's no point testing it?

Whoops, forgot to test with different options. But yea if it gives different results each time, then there's no point in testing this.
@fzdy1914 In this case, ignore my previous comment, we should ensure `InputBuilder` only allow adding of each option once. No need for another exception if user add flags more than once for now.

fzdy1914: How can we achieve this without using exceptions? I am not able to figure out an elegant way to achiece this effect.

fzdy1914: How can we achieve this without using exceptions? I am not able to figure out an elegant way to achiece this effect.

fzdy1914: How can we achieve this without using exceptions? I am not able to figure out an elegant way to achiece this effect.

yamidark: >How can we achieve this without using exceptions? I am not able to figure out an elegant way to achiece this effect.

Why do we need an exception for this? If we call 'set' for any option, we should just overwrite the existing value if it exists, ensure that each option can only be called once.

fzdy1914: I get your point. So we change add to set, correct?

yamidark: >I get your point. So we change add to set, correct?

Yes

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,122 @@
...
+    private String help;
+    private String config;
+    private String repos;
+    private String view;
+    private String output;
+    private String sinceDate;
+    private String untilDate;
+    private String formats;
+    private String ignoreStandaloneConfig;
+    private String extraString;
```
eugenepeh: These changes are not necessary.
Reason: Test util shouldn't interfere with the parsing, e.g. double params such as `--config ... --config ... ` because this is handled by the ArgParser.

eugenepeh: Also use a `StringBuilder` instead. It is more efficient than concatenation in loops.

fzdy1914: > These changes are not necessary.
> Reason: Test util shouldn't interfere with the parsing, e.g. double params such as `--config ... --config ... ` because this is handled by the ArgParser.

@yamidark So I revert the latest commit?

yamidark: >These changes are not necessary.
Reason: Test util shouldn't interfere with the parsing, e.g. double params such as `--config ... --config ... ` because this is handled by the ArgParser.

@eugenepeh From our discussion [here](https://github.com/reposense/RepoSense/pull/526#discussion_r256039060), from testing by @yong24s, using double params or multiple params gives undefined behaviour, so I thought it would be good to ensure that we only use single params for testing though.

src/test/java/reposense/util/InputBuilder.java
eugenepeh: Hmm, 

>`./gradlew run -Dargs="-repo a.git -repo b.git -repos c.git"` -> takes `c.git`

Isn't that caused by the different in flag? e.g. `-repo` -> `-repos` may have caused the ArgParser to treat it differently.

I don't think terminal does anything to the args, they should be fed into the program as a whole string and parsed into string array to the main.
https://github.com/reposense/RepoSense/blob/f86a52617ee8ab13cdf696c0e196fdc90cf5f778/src/main/java/reposense/RepoSense.java#L34
image

yamidark: >`./gradlew run -Dargs="-config a -config b"` -> takes `a`

@yong24s Used 2 `-config` params and it took the 1st one, while when I used 2 `-config` params it took the last one instead, which is why we thought the behaviour is undefined/unpredictable depending on how to user uses it

src/test/java/reposense/util/InputBuilder.java
eugenepeh: continuing the discussion from that thread:

>@yong24s Used 2 `-config` params and it took the 1st one, while when I used 2 `-config` params it took the last one instead, which is why we thought the behaviour is undefined/unpredictable depending on how to user uses it

Hmm I would prefer keep it as similar as the terminal input as possible. The bad thing about this change is that it keeps the params in a rigid order and can cause `extraString` to be treated as input of any param. 

If there's an issue with the `undefined/unpredictable` input, we can just avoid testing multiple same param as a whole; since we have environment for 3 different os, it should have low possibly of passing the tests if thats the case.

yamidark: >we can just avoid testing multiple same param as a whole

Well, I asked for this change to prevent 'accidentally' setting up any tests for this (e.g. using the same `InputBuilder` for 2 tests within a single test method) 😅. But then again, I think this can be prevented by making `build` to call `reset` after building everything.

>The bad thing about this change is that it keeps the params in a rigid order and can cause `extraString` to be treated as input of any param.

Hmmm, thats true as well, params must be in a specific order (`-help` comes first, than `-config` and so on) which prevents certain tests as well. I guess we can revert to the old method, but add Javadocs to warn developers that they should only be called once for each `build`?

eugenepeh: >Hmmm, thats true as well, params must be in a specific order (-help comes first, than -config and so on) which prevents certain tests as well. I guess we can revert to the old method, but add Javadocs to warn developers that they should only be called once for each build?

Yeah, javadocs should be good enough.

yamidark: @fzdy1914 Could you revert back to the old version, this time add Javadocs as mentioned. Sorry for the trouble caused.

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     * Add a help flag to the input.
+     * This method should only be called once in one build.
+     */
+    public InputBuilder addHelp() {
+        input.append(ArgsParser.HELP_FLAGS[0] + WHITESPACE);
+        return this;
+    }
+
+    /**
+     * Add a config flag with the config folder path to the input.
```
yamidark: `Adds the config flag with the {@code path} as argument.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @param path The config folder path.
+     */
+    public InputBuilder addConfig(Path path) {
+        input.append(ArgsParser.CONFIG_FLAGS[0] + WHITESPACE + path + WHITESPACE);
+        return this;
+    }
+
+    /**
+     * Add a repo flag with the several repo paths to the input.
```
yamidark: `Adds the repo flag with the {@code paths} as arguments.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+    public InputBuilder addRepos(String... paths) {
+        input.append(ArgsParser.REPO_FLAGS[0] + WHITESPACE);
+        for (String path : paths) {
+            input.append(path + WHITESPACE);
+        }
+        return this;
+    }
+
+    /**
+     * Add a view flag with the view folder path to the input.
```
yamidark: `Adds the view flag with the {@code path} as argument.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     * Add a view flag only to the input.
+     * This method should only be called once in one build.
+     */
+    public InputBuilder addView() {
+        input.append(ArgsParser.VIEW_FLAGS[0] + WHITESPACE);
+        return this;
+    }
+
+    /**
+     * Add a output flag with the output folder path to the input.
```
yamidark: `Adds the output flag with the {@code path} as argument.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @param path The output folder path.
+     */
+    public InputBuilder addOutput(Path path) {
+        input.append(ArgsParser.OUTPUT_FLAGS[0] + WHITESPACE + path + WHITESPACE);
+        return this;
+    }
+
+    /**
+     * Add a since flag with the since date to the input.
```
yamidark: `Adds the since flag with the {@code date} as argument.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @param date The since date.
+     */
+    public InputBuilder addSinceDate(String date) {
+        input.append(ArgsParser.SINCE_FLAGS[0] + WHITESPACE + date + WHITESPACE);
+        return this;
+    }
+
+    /**
+     * Add a until flag with the until date to the input.
```
yamidark: `Adds the until flag with the {@code path} as argument.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     * Add a ignoreStandaloneConfig flag to the input.
+     * This method should only be called once in one build.
+     */
+    public InputBuilder addIgnoreStandaloneConfig() {
+        input.append(ArgsParser.IGNORE_FLAGS[0] + WHITESPACE);
+        return this;
+    }
+
+    /**
+     * Add extra content to the input.
```
yamidark: `Adds {@code content} to the input.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     */
+    public InputBuilder add(String content) {
+        input.append(content + WHITESPACE);
+        return this;
+    }
+
+    /**
+     * Add {@code num} white spaces to the input.
```
yamidark: Add {@code num} **of** white spaces to the input.

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     */
+    public InputBuilder addWhiteSpace(int num) {
+        for (int i = 0; i < num; i++) {
+            input.append(WHITESPACE);
+        }
+        return this;
+    }
+
+    /**
+     * Reset the {@code InputBuilder}.
```
yamidark: `Clears all input and flags given.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+     *
+     * @param date The until date.
+     */
+    public InputBuilder addUntilDate(String date) {
+        input.append(ArgsParser.UNTIL_FLAGS[0] + WHITESPACE + date + WHITESPACE);
+        return this;
+    }
+
+    /**
+     * Add a format flag with the formats to the input.
```
yamidark: `Adds the format flag with {@code formats} as argument.`

src/test/java/reposense/util/InputBuilder.java
```diff
@@ -0,0 +1,159 @@
...
+public class InputBuilder {
+    private static final String WHITESPACE = " ";
+
+    private StringBuilder input;
+
+    public InputBuilder() {
+        this.input = new StringBuilder();
+    }
+
+    public String build() {
```
yamidark: Javadocs for this as well.

yamidark:
Nits to proposed message:
```
In ArgsParser, the string flags used for the CLI are magic strings
repeated in multiple places. Meanwhile, the standard of CLI argument
used does not follow a widely-used standard.

We should avoid the use of magic strings, instead defining them in a
central location. As RepoSense is designed for CLI usage, it is also
a good idea to follow a widely-used standard for CLI to match user
expectations. 

Let's define the magic strings in a central location as constants and
follow the GNU Coding Standard for the CLI, using `--` prefix for
`long-named flags` and `-` for its equivalent `single-letter alias`.

[1] GNU Coding Standards: Command Line Interfaces
https://www.gnu.org/prep/standards/html_node/Command_002dLine-Interfaces.html
```

PR #529, opened 2019-01-31
CodeView: Collapse/Expand All button not updated correctly after hiding all files #521

### LINKED ISSUE #521: Code view: expand/collapse button not updated correctly when hidden files are unhide
expand-collapse

chelseyong:
Fixes #521.
```
If all files are collapsed, then the select-all box is unchecked, and checked,
all files are now expanded, but the button shows `Expand All`.

Since files are automatically expanded when they are rendered,
let's change the button to `Collapse All` whenever select all button
is clicked.
```


chelseyong:
Ready for review

frontend/src/static/js/v_authorship.js
```diff
@@ -174,9 +175,11 @@ window.vAuthorship = {
       if (this.fileTypes.length === this.selectedFileTypes.length) {
         this.selectedFiles = this.files;
         this.isSelectAllChecked = true;
+        window.app.isCollapsed = false;
       } else if (this.selectedFileTypes.length === 0) {
         this.selectedFiles = [];
         this.isSelectAllChecked = false;
+        window.app.isCollapsed = false;
       } else {
         this.selectedFiles = this.files.filter((file) => this.selectedFileTypes.includes((file.path.split('.').pop())));
```
ongspxm: Just a qns. How come you dont want isCollapsed to be set when the files are partially filtered?

chelseyong: The user may have collapsed some files, so its not true that all the files are expanded, vice versa.

yamidark:
As a side note, if I manually collapse/expand all the files 1 by 1, should the `Collapse all` and `Expand all` button be changed as well?

chelseyong:
> As a side note, if I manually collapse/expand all the files 1 by 1, should the `Collapse all` and `Expand all` button be changed as well?

Yes

chelseyong:
The files-rendering `div` has `.file .active` as class, causing there to be 1 extra `active file` class. However, when files are manually toggled, the files-rendering `div` does not change accordingly.  So it is now replaced as a `template` with no class.
This bug happened occasionally, and I didn't find it previously :sweat_smile:

chelseyong:
> Hmm, seems like the recent change has changed content and title mismatch again

Fixed

chelseyong:
> Hmm, seems like the recent change has changed content and title mismatch again

Fixed.
Also moved the changing of `collapse/expand all` button into `updated()`

yamidark:
Nits to proposed message:
```
If the select-all box for file formats in the code panel is unchecked,
the 'Collapse/Expand All' button will change to become 'Expand All' as
all files are hidden. However, when the select-all box is checked
again, all the files rendered will be expanded, but the
'Collapse/Expand All' button will remain at 'Expand All'.

This is due to files being automatically expanded when they are
rendered, after we check any of the file format checkboxes.

Let's update the 'Collapse/Expand All' button to also track files
that are newly rendered that are expanded whenever a file format 
checkbox is clicked.
```
@yong24s Can help me nit?

PR #530, opened 2019-01-31
build.gradle: Fix checkstyle plugin failure in JDK11

### LINKED ISSUE #523: Gradle CheckStyle plugin fails in jdk11
As title.

fzdy1914:
Fixes #523 

The `checkstyle` plugin of Gradle fails in JDK11.

The main reason is that, in JDK11, the `user.dir` cannot be reset by gradle.
Also, the `FileContentHolder` module are deprecated.

Let's add a variable suggested by Gradle to solve this problem.


fzdy1914:
Following are the reference for the solution。
https://github.com/gradle/gradle/issues/8286
https://github.com/gradle/gradle/issues/5187
https://github.com/checkstyle/checkstyle/issues/5187

yong24s:
Try not to tag **devs**, now that you can add labels. 

Quick question:
Why does checkstyle `8.16` fails? 

Does it only work with jdk11 because you didn't update `travis.yml` and `appveyor.yml` like you did in `addressbook-level4`. 🤔

yong24s:
Should we take this chance to update our jdk version in CI as well? The version we use has already EOL. 😢 

image

fzdy1914:
> Try not to tag **devs**, now that you can add labels.
> 
> Quick question:
> Why does checkstyle `8.16` fails?
> 
> Does it only work with jdk11 because you didn't update `travis.yml` and `appveyor.yml` like you did in `addressbook-level4`. 🤔

It works with all JDK I believe, you can see that Travis and AppVoyer uses JDK9 and it passes. What I have done does not use any of JDK11 feature.

fzdy1914:
> In this case, we need use **JDK8** and **JDK11** as the Java versions to test in our CI?

Do you mean that I need to change the script?

yong24s:
@eugenepeh @yamidark 

> The purpose of this PR is to ensure that users and devs can run or build RepoSense in a **JDK11+** environment.
> 
> But the CS2103/CS2113 users using addressbook-level4 would be using **JDK8** right? 🤔
> 
> In this case, we need use **JDK8** and **JDK11** as the Java versions to test in our CI?

Can share your opinion? 

I believe the goal of this PR should be preparing RepoSense for **JDK11+** enviroment.
If not, there is no need to fix checkstyle in Gradle. 🤔

eugenepeh:
>I believe the goal of this PR should be preparing RepoSense for **JDK11+** enviroment.

Yes. 

>In this case, we need use **JDK8** and **JDK11** as the Java versions to test in our CI?

Not necessary. Unless there are deprecated packages which latest jdk do not support, we don't have to test using JDK11. 

>If not, there is no need to fix checkstyle in Gradle. 🤔

It would be good for us to fix it, so users who have latest jdk can continue to use our product without switching versions.

@yong24s can you verify that it works with jdk11?

yong24s:
> @yong24s can you verify that it works with jdk11?

Let me verify using CI first, https://github.com/yong24s/RepoSense/pull/6.

yong24s:
> > In this case, we need use **JDK8** and **JDK11** as the Java versions to test in our CI?
> 
> Not necessary. Unless there are deprecated packages which latest jdk do not support, we don't have to test using JDK11.

My thoughts are reverse of yours 😂. What if we unknowingly used new features not in JDK8?

eugenepeh:
Then it wouldn't be able to execute with Gradle lol, since compatibility is set to jdk8

yamidark:
>What if we unknowingly used new features not in JDK8?

We should have already set the JDK compatibility in our Intellij IDE to use JDK8, so we shouldn't be able to use/add any features not in JDK8.

yong24s:
@yamidark 
Fails on macOS with JDK 11. 
```
Task :checkstyleMain FAILED
```

PR #487, opened 2019-01-18
Show breakdown of file type in Contribution Bar

apoorva17:
Fixes #456 

```
Contribution Bar is currently a single red bar which represents total contribution of 
the author.

Let's reflect the breakdown of filetype in Contribution Bar. 

The user can hover over a specific section of the Contribution Bar to see information 
about the filetype, total lines contributed and analysis period in the tooltip. In order 
to deal with the 'too many filetypes' scenario, the Contribution Bar only 
shows the breakdown for the first 5 filetypes.

Since this filetype information is currently being processed in the front-end in code 
view (v_authorship), let's move the processing of filetype information to the backend 
in order to be able to use the pre-processed filetype information for both chart 
view (v_summary) and code view (v_authorship). Let's also refactor the filetype
processing in code-view (v_authorship) to match the changes being made.
```

PR #532, opened 2019-02-01
Code view: support glob filters

### LINKED ISSUE #504: Code view: support glob filters
Current: we can filter the code view by file types, using tick boxes
Suggested: support a more powerful filter box where the user can type a glob e.g., `**/test/*.java`

apoorva17:
Fixes #504 
```
With the addition of #170, users can filter the code view by filetype
by selecting tick boxes.

Let's support a more powerful search box where users can filter files
by glob e.g. **/test/*.java.
```


apoorva17:
Since I could not find any frontend Javascript library for glob pattern matching, i have used a node-module called ```minimatch```. Therefore, I have used ```browserify``` and ```vueify``` to use ```minimatch``` in the frontend. 

Since the build javascript file produced is automated, it does not conform to ```ESLint``` style checks, therefore I have excluded it from the ```travis.yml``` file.

apoorva17:

apoorva17:
@ongspxm kindly review

apoorva17:
> This feature doesn't work together with the current file type filter checkboxes:
> Uncheck all file types
> Set glob filter to any of the filetypes (e.g. `**.java`)
> All `.java` files will appear, when the checkbox for `.java` is still unchecked.

@yamidark this is fixed now

apoorva17:
> @apoorva17 I intend to have the file types selection to tap on the search bar functionality when its complete.
> E.g. clicking on the file formats check box will update the search bar, letting the search bar to perform the filtration, so that we can remove the redundant code that does the same thing. Can you look into it?

This can be looked at in a separate PR. @damithc your opinion?

damithc:
> This can be looked at in a separate PR. @damithc your opinion?

I'm OK to do it as a separate PR. 

@eugenepeh Actually, I'm not entirely sure about that feature. Auto-updating the search bar as the user ticks checkboxes is nice, but complications will arise when the user types in the search bar and tick checkboxes in tandem. If we go in that direction, we may have to disable checkboxes the moment the user starts typing in the search bar.
Alternatively, instead of auto-updating the search bar, we can simply consider the combination of the two implicitly. i.e., if the documentation group is ticked and the user types `*foo*.adoc` in the search bar, we filter out files that meet both criteria.

eugenepeh:
>@eugenepeh Actually, I'm not entirely sure about that feature. Auto-updating the search bar as the user ticks checkboxes is nice, but complications will arise when the user types in the search bar and tick checkboxes in tandem. If we go in that direction, we may have to disable checkboxes the moment the user starts typing in the search bar.

The reason I propose this is because it is not going to be easy to get the checkbox and search bar work well with each other. The complication you mentioned, still exist in current implementation. E.g. toggling the checkbox ignores the search bar values and shows results corresponding only to the checkboxes.

I propose that we replace all the checkboxes with buttons that append or replace the search bar value so that both work in coherent. E.g. `[clear]`, `[gradle]`, `[java]` etc and the lines summary can be moved to the search results which updates along with the search condition. E.g. `All: 2741 (443) gradle: 14 (2) jade: 75 (9) java: 1438 (258) js: 477 (72) md: 129 (34) scss: 571 (60) yml: 37 (8)`. 

>Alternatively, instead of auto-updating the search bar, we can simply consider the combination of the two implicitly. i.e., if the documentation group is ticked and the user types `*foo*.adoc` in the search bar, we filter out files that meet both criteria.

The advantage of my proposition over this alternative is that:
1. Easier to manage the url bookmarking. Only need to parse a single field.
2. Eliminate race condition and concurrent actions. If we have both checkbox and search bar which operates on a single list of files concurrently, it can result in crash.
3. Currently, there is no atomic way to show a specific file type. E.g. java. User would have to hide all files first, follow by selecting the java files which result in unnecessary waiting time to hide all files.
With this idea, we can just click on `java` and the search bar will be updated accordingly to show only `java` in an atomic action. To hide all `java` files, we can probably implement middle mouse click to show the inverse.
4. Lower complexity of implementation, and reduce redundant codes.
5. LoC update with changes.
6. In tandem with my suggestion in #547

damithc:
> @damithc , what do you think?

Good points. 
I kind of dislike the idea of auto-updating a field that is also meant to be editable by the user. Users are unpredictable, for example, they can edit the auto-generated text in arbitrary ways.
In [here](https://github.com/reposense/RepoSense/pull/547#issuecomment-462016200), I'm proposing to keep tick boxes and filter box independent. i.e., only one will be used at any time. Would that reduce the complications?

eugenepeh:
>I kind of dislike the idea of auto-updating a field that is also meant to be editable by the user. Users are unpredictable, for example, they can edit the auto-generated text in arbitrary ways.

I see. Though, another advantage of auto-generated text is that it helps user understand how the glob works .

>In [here](https://github.com/reposense/RepoSense/pull/547#issuecomment-462016200), I'm proposing to keep tick boxes and filter box independent. i.e., only one will be used at any time. Would that reduce the complications?

It does helps to reduce the complexity, though not any helpful to the bookmarking because it is still 2 separate field. However, if users are more comfortable with the ui in https://github.com/reposense/RepoSense/pull/547#issuecomment-462016200, I am all in for it as my main concerns are mainly the complexity and potential bugs getting the 2 working together.

damithc:
> I see. Though, another advantage of auto-generated text is that it helps user understand how the glob works .

We can show the auto-generated glob under the tick boxes as a read-only label so that the user can copy it over to the filter box, if he wants.

> It does helps to reduce the complexity, though not any helpful to the bookmarking because it is still 2 separate field.

That's true. In this case, we'll have to encode filter text or tickbox configuration, whichever the chosen option. On a general note, it maybe good to define a data structure that can represent all report parameters, and functions to encode it into a URL (and decode it from the URL). perhaps something based on JSON?

 > However, if users are more comfortable with the ui in [#547 (comment)](https://github.com/reposense/RepoSense/pull/547#issuecomment-462016200), I am all in for it as my main concerns are mainly the complexity and potential bugs getting the 2 working together.

Yes, getting the two to work together will be hard. Let's keep them separate.

package.json
```diff
@@ -8,7 +8,8 @@
   },
   "scripts": {
     "lint": "eslint",
-    "spuild": "spuild frontend"
+    "spuild": "spuild frontend",
+    "build": "browserify -t vueify -e frontend/src/static/js/v_authorship.js -o frontend/src/build/v_authorship.js"
```
ongspxm: add the build part to spuild as well?, so that dun need to run 2 commands?

apoorva17: @ongspxm how do i add it in?

apoorva17: Thanks @ongspxm !

apoorva17:
@damithc  i've updated the implementation as per [#547 (comment)](https://github.com/reposense/RepoSense/pull/547#issuecomment-462016200)

package.json
apoorva17: Seem to be getting an error with ```appveyor`` upon adding this tho @ongspxm

damithc:
The view doesn't seem to update when I switch back from glob filter to tick boxes:

image

apoorva17:
> The view doesn't seem to update when I switch back from glob filter to tick boxes:

@damithc this has been fixed

damithc:
> @damithc this has been fixed

Still not working. Steps:
1. Load the code of one person
2. Use a glob filter to filter out some files
3. Switch back to tick boxes
   Expected: Show all files (as the `All` is ticked)
   Actual: the view shown in the step 2 remains

apoorva17:
> 
> Still not working. Steps:
> 
> 1. Load the code of one person
> 2. Use a glob filter to filter out some files
> 3. Switch back to tick boxes
>    Expected: Show all files (as the `All` is ticked)
>    Actual: the view shown in the step 2 remains

@damithc I have updated the implementation such that if a user switches to/from search bar and tickboxes, all files are shown as per the default when the code view is first opened.

damithc:
> @damithc I have updated the implementation such that if a user switches to/from search bar and tickboxes, all files are shown as per the default when the code view is first opened.

The better approach is to retain (instead of reset) the settings when switching between the two modes. But I'm OK with the current approach as the retaining of settings can be added later.

I think resetting upon losing is better than resetting upon gaining focus. The former gives a false impression that the setting will be retained for future. But this is a minor thing.

apoorva17:
> I think resetting upon losing is better than resetting upon gaining focus. The former gives a false impression that the setting will be retained for future. But this is a minor thing.

@damithc I've updated the implementation to reset upon losing instead of upon gaining focus.

apoorva17:
@ongspxm ```appveyor``` is giving me the following error for adding ```"spuild": "npm run build; spuild frontend"```: ```npm ERR! missing script: build;``` although the ```build``` script is present, how do I fix this error?

apoorva17:
@reposense/devs changes have been made pls proceed to review this PR.

apoorva17:
> Instead of disabling the deselected option, can we possibly make interaction switch the selection?
> 
> E.g. if I change any of the checkbox, automatically switch the radio button focus to checkbox and perform the filter? Same for when attempting to type in the search box.

I think that it is clearer for the users that both the tick boxes and search box are independent if we disable the deselected option. @damithc your opinion on this?

eugenepeh:
>I think that it is clearer for the users that both the tick boxes and search box are independent if we disable the deselected option.

In this scenario, the radio button serves nothing more than an "indication" rather than a selection. Changing option doesn't do anything until the corresponding controls are modified. Hence, it is rather redundant for users to manually switch the selection.

eugenepeh:
Also, it is less confusing when the selection is tied to the display result rather than the user's selection.

damithc:
> I think that it is clearer for the users that both the tick boxes and search box are independent if we disable the deselected option. @damithc your opinion on this?

As @eugenepeh pointed out, clicking the radio button seems to be an unnecessary extra step, but I don't mind going ahead with the current UX as those optimizations can be done as future PRs. Best to get an MVP version merged first and refine further as separate PRs.

On a separate note, the feature doesn't seem to be working correctly. 
Steps:
* Apply a glob filter that shows nothing
* Switch back to tick boxes. 
  * Expected: show all files
  * Actual:
image

apoorva17:
> On a separate note, the feature doesn't seem to be working correctly.
> Steps:
> 
> * Apply a glob filter that shows nothing
> * Switch back to tick boxes.
>   
>   * Expected: show all files
>   * Actual:
>     image

@damithc this should be fixed now.

damithc:
> @damithc this should be fixed now.

Yes, seems to be working

apoorva17:
> As @eugenepeh pointed out, clicking the radio button seems to be an unnecessary extra step, but I don't mind going ahead with the current UX as those optimizations can be done as future PRs. Best to get an MVP version merged first and refine further as separate PRs.

Yes, I would like to get the MVP version merged first, and add the interaction-based switch selection in a separate PR. @reposense/devs you may kindly proceed with the review for this PR.

frontend/src/index.jade
```diff
@@ -215,5 +222,5 @@ html
 
     script(src="static/js/api.js")
     script(src="static/js/v_summary.js")
-    script(src="static/js/v_authorship.js")
+    script(src="build/v_authorship.js")
```
yamidark: Why this change?

.travis.yml
```diff
@@ -11,7 +11,7 @@ matrix:
 
 script:
   - ./config/travis/run-checks.sh
-  - npm run lint frontend/src/**/*.js
+  - npm run lint frontend/src/static/*.js
```
yamidark: Why this change?

apoorva17: The ```build.js``` file is auto-generated by browersify and does not conform to the eslint rules so I have excluded it here.

frontend/src/index.jade
apoorva17: The build file generated by browersify is located in the ```build``` path.

eugenepeh: Apparently this change has caused the CI to failed. Please ensure that the CI passed the test before requesting for review.

apoorva17: > Apparently this change has caused the CI to failed. Please ensure that the CI passed the test before requesting for review.

No, the CI was failing because of [this](https://github.com/reposense/RepoSense/pull/532#discussion_r255380348), which I have also mentioned [here](https://github.com/reposense/RepoSense/pull/532#issuecomment-464006057).

eugenepeh: >No, the CI was failing because of [this](https://github.com/reposense/RepoSense/pull/532#discussion_r255380348), which I have also mentioned [here](https://github.com/reposense/RepoSense/pull/532#issuecomment-464006057).

My bad, was looking through phone during my lunch time earlier.

Anyway, its bad to have the JS all over the place. Instead of modifying it while its still in the source directory, can you modify it after its build? Or move it to the build directory directly?

E.g. `browserify -t vueify -e frontend/src/static/js/v_authorship.js -o frontend/build/static/js/v_authorship.js`
or  `browserify -t vueify -e frontend/build/static/js/v_authorship.js -o frontend/build/static/js/v_authorship.js`
whichever works.

apoorva17: > E.g. `browserify -t vueify -e frontend/src/static/js/v_authorship.js -o frontend/build/static/js/v_authorship.js`
> or `browserify -t vueify -e frontend/build/static/js/v_authorship.js -o frontend/build/static/js/v_authorship.js`
> whichever works.

I tried both of this, it doesn't work.

eugenepeh: >I tried both of this, it doesn't work, not sure why.

I think the `build` has to be executed after `spuild` for it to work. Can you try if changing the order on `package.json` helps?

```
    "spuild": "spuild frontend",
    "build": ...
```

apoorva17: > I think the `build` has to be executed after `spuild` for it to work. Can you try if changing the order on `package.json` helps?
> 
> ```
>     "spuild": "spuild frontend",
>     "build": ...
> ```

Tried this, but it still doesn't work. Maybe the file needs to be modified while it is in the source directly before the build so that browersify can bundle up the 'require' dependencies during build?

yong24s: @apoorva17 

Can fill me in? From what I read, I believe you are trying to add `build` after `spuild`?

apoorva17: @yong24s I'm using ```browserify``` to generate a modified v_authorship.js file which bundles up dependencies for using ```require``` in browser. I'm currently using the command ```browserify -t vueify -e frontend/src/static/js/v_authorship.js -o frontend/src/build/v_authorship.js```, which creates and puts the modified v_authorship.js file in ```src/build```. However, @eugenepeh would like to modify it after its build/move modified file directly to build directly using ```browserify -t vueify -e frontend/src/static/js/v_authorship.js -o frontend/build/static/js/v_authorship.js```. I tried doing this, but it is not working, which is why I tried adding ```build``` after ```spuild```. But it still doesn't seem to work.

apoorva17: > 
> [yong24s#12](https://github.com/yong24s/RepoSense/pull/12)
> https://ci.appveyor.com/project/yong24s/reposense/builds/22623766#L42
> 
> ^ Hmm, it seems like the CI are already doing this automagically 🤔

Yeah, this seems to show that ```spuild``` is already being executed before ```build``` regardless of ordering in ```package.json```, so it shouldn't be the cause of the above problem.

yong24s: According to the [Travis logs in your builds](https://travis-ci.org/reposense/RepoSense/jobs/497467138#L716), the tasks order are like this `npmRunSpuild -> zipReport -> npmRunBuild`.

I made [changes to the `build.gradle`](https://github.com/yong24s/RepoSense/pull/12) to reorder it to `npmRunSpuild -> npmRunBuild -> zipReport`.

But I found out that `npmRunBuild` has a hidden gotcha, where it depends on `processResources`. This breaks the correct order we define in #552.

```
$ ./gradlew run --dry-run

:processResources SKIPPED <-- Moved to first
:downloadNode SKIPPED
:npmInstall SKIPPED
:npmRunSpuild SKIPPED
:npmRunBuild SKIPPED
:zipReport SKIPPED
:compileJava SKIPPED
:classes SKIPPED
:run SKIPPED
```

I have exhaustively tried to the correct its order but it creates a circular dependency. After having given some thoughts, I tried to change its name and it works. 🎉 

```
$ ./gradlew run --dry-run

:downloadNode SKIPPED
:npmInstall SKIPPED
:npmRunSpuild SKIPPED
:npmRunBrowserify SKIPPED
:zipReport SKIPPED
:compileJava SKIPPED
:processResources SKIPPED
:classes SKIPPED
:run SKIPPED
```

Not sure if this is what you needed, you can check my changes [here](https://github.com/yong24s/RepoSense/pull/12).

apoorva17:
@chelseyong Due to the changes in #513 , ```checkbox.checked``` is not working which is needed in this PR. Please look into this.

chelseyong:
@apoorva17 the selecting/unselecting of all the checkboxes seems functional. Can you elaborate on the problem or provide snapshots?

apoorva17:
@chelseyong a couple of things:

1) I was using your ```this.selectedFiles``` to filter the files for my glob filter, since it is now moved to ```computed()```, how do i use that to do the filtering of the files now?

2) In my methods, I am accessing the checkbox elements via ```document.getElementsByClassName``` and doing ```checkbox.checked``` and checkbox.click()``` to reset files upon switching between glob filter and checkbox which I now no longer seem to be able to modify.

chelseyong:
1. The filtering can be done in `selectedFiles`, which will automatically update the files shown.

2. If you want the checkboxes to be unselected, the chosen file types can be removed from `selectedFileTypes` in `v_authorship`.

frontend/src/index.jade
apoorva17: @yong24s i've implemented the changes u suggested and it works locally, but it doesn't seem to work when i push it to netlify.

apoorva17:
> 2. If you want the checkboxes to be unselected, the chosen file types can be removed from `selectedFileTypes` in `v_authorship`.

Due to this change, I had to move ```enableSearchBar()``` and ```enableCheckBoxes()``` into ```methods``` in Vue to be able to access ```fileTypes```, ```selectedFileTypes```, and ```isSelectAllChecked```, as I am unable to reset and tick all the checkboxes otherwise.

frontend/src/index.jade
apoorva17: @yong24s this worked thanks!

apoorva17:
@reposense/devs requested changes have been made, you can proceed to review the PR.

eugenepeh:
@apoorva17 Just to clarify, upon switching the radiobox selection, the checkbox and search box values are reset, is that an intentional behaviour?

apoorva17:
> @apoorva17 Just to clarify, upon switching the radiobox selection, the checkbox and search box values are reset, is that an intentional behaviour?

Yes, as mentioned [here](https://github.com/reposense/RepoSense/pull/532#issuecomment-464005385)

apoorva17:
> Do update the report image in the UserGuide as well, only DeveloperGuide updated

Ok, I've updated it

frontend/src/static/js/v_authorship.js
```diff
@@ -188,6 +191,46 @@ window.vAuthorship = {
       }
     },
 
+    updateFilterSearch(evt) {
+      this.filterSearch = (evt.target.value.length !== 0) ? evt.target.value : '*';
+    },
+
+    getFilteredFiles() {
+    },
```
ongspxm: this is here because?

damithc:
On a separate note, is there a glob syntax to specify multiple conditions? e.g., having `Test` or `Manager` in the file name.
@apoorva17 a follow-up PR you can do is to support more complex conditions in the glob filter, such as `OR` `AND`

frontend/src/static/js/v_authorship.js
apoorva17: I tried to do ```v-on:submit.prevent``` for glob filter without specifying any method but it wasn't working so i had to add this method in.

apoorva17:
> On a separate note, is there a glob syntax to specify multiple conditions? e.g., having `Test` or `Manager` in the file name.
> @apoorva17 a follow-up PR you can do is to support more complex conditions in the glob filter, such as `OR` `AND`

Not currently, I'm using [minimatch](https://github.com/isaacs/minimatch) library which does not seem to support such a glob syntax.

yong24s:
> > On a separate note, is there a glob syntax to specify multiple conditions? e.g., having `Test` or `Manager` in the file name.
> 
> Not currently, I'm using [minimatch](https://github.com/isaacs/minimatch) library which does not seem to support such a glob syntax.

Can try [multimatch]( https://github.com/sindresorhus/multimatch). It extends minimatch library to support multiple conditions.

apoorva17:
> Can try [multimatch](https://github.com/sindresorhus/multimatch). It extends minimatch library to support multiple conditions.

Sure, I can look into this in a follow-up PR.

eugenepeh:
@ongspxm just a reminder, you have an unresolved conversation https://github.com/reposense/RepoSense/pull/532#pullrequestreview-209869904

frontend/src/static/js/v_authorship.js
ongspxm: sflr; pls add a comment here if needed. but try this first, it should work, using vanilla form functions:

```
form(onsubmit="return false")
```

ongspxm:
@apoorva17 sry, i pressed wrong, dun know how to unapprove, but just that change will do

apoorva17:
@ongspxm I've made the requested change.

apoorva17:
@yong24s getting same frontend error like [here](https://github.com/reposense/RepoSense/pull/487#issuecomment-469580574)

yamidark:
@apoorva17 Restarted it a few times and it passes now, seems like cypress will fail the tests once in awhile (happens to be locally as well). Will raise an issue for this.

eugenepeh:
Nits to proposed message:

```
The addition of #170 provides users with the convenience of filtering
the files in code view by formats using tick boxes. However, this does
not solve the problem of finding specific file(s) by name, folder or
path.

As a solution, let's add a search box to provide users with an
additional option of filtering files by glob patterns. E.g.
**/test/*.java.
```

Issue #527, opened 2019-01-31
Code view: highlightjs is triggered multiple times causing report to lag/crash


yong24s:
hightlightjs (`hljs`) is used for highlighting of codes in code view giving it context and readability. It can be triggered to render an element with ```window.hljs.highlightBlock(element)```. 

However, it seems to render regardless of trigger for any code in `<pre><code>` [1]. This can be verified by removing ```window.hljs.highlightBlock(element)```.

Furthermore, in our frontend, `vuejs` will trigger `hljs` in `Vue#update`, when there is a data change, causing DOM to be re-rendered [2].
https://github.com/reposense/RepoSense/blob/ee21b0d8752adda7cffe547d490222c3dbd474a4/frontend/src/static/js/v_authorship.js#L209-L215

`hljs` operations are expensive and stateless, where it re-renders every time even on a rendered DOM;
Also, multiple `hljs` operations can be triggered at once, causing the report to lag and crash [3]. This can be verified by spam-clicking on the same button to view author's code after it has been loaded.

Let's 

* fix [1] and [2] by enclosing the code segment with a different tag and only trigger `hljs` when code-view is opened. 

* fix [3] by ensuring only one `hljs` operation runs at a time, by terminating any operation before starting a new one.

yong24s:
@chelseyong 
I decide to work on this issue as it is a **P.High**. It should be cleared ASAP before other PRs can be merged.

Let me know if you have already worked on it.

PR #533, opened 2019-02-02
highlightjs is triggered multiple times causing report to lag/crash #527

yong24s:


yong24s:
@ongspxm 

Can help me with the ESLint errors? 😢

yong24s:
@eugenepeh

For your preliminary testing, please.

frontend/src/static/js/main.js
```diff
@@ -64,6 +64,17 @@ window.deregisterMouseMove = () => {
   window.$('app-wrapper').style['user-select'] = 'auto';
 };
 
+/* global Vue hljs */
+Vue.directive('hljs', {
+  inserted(ele, binding) {
+    const element = ele;
+    const fileExtension = binding.value.split('.').pop();
+
+    element.className = fileExtension;
```
ongspxm: Minor nit, just set element.className to binding.value.split dun need make new var

emer7:
I think HighlightJS doesn't highlight the syntax for subsequent contributor now:

Let's say there are 2 contributor A and B. Upon opening's A code panel it highlights the syntax but it doesn't highlight B's code panel.

Is it caused by this PR?

eugenepeh:
@emer7 I am unable to replicate this. Would need you to provide more details. Note that `adoc` doesn't have much syntax highlighted as the other file types. If you suspect that this is caused by the PR, do use netlify to verify. To do so, visit the deploy url for this commit n the one before e.g. `https://deploy-preview-533--reposense.netlify.com/`, replace the `533` with target PR number.

emer7:
Found it, turns out that problem itself is fixed by this PR. Sorry for the false alarm

PR #490, opened 2019-01-19
Code view: give link to history and blame views of each file

fzdy1914:
Fixes #466 

```
Reviewers may have the need to investigate on a file to see who else
has edited the same file, however, this feature is not available in the
code view.

Let's add direct links in the code view to visit the history view and
blame view of the files (on GitHub).
```

Issue #390, opened 2018-10-29
Revamp the override option


eugenepeh:
As discussed earlier with Akshay, there doesn't seems to be a clean way to override just a specific field for the json config.

Some ideas that we can possibly revamp this:
1. Make the cli param highest order which can override any single field
2. Make single field override-able option, to override some fields of the json config (if any), in the csv config instead of override the entire thing.

Can discuss other alternatives.

eugenepeh:
Another way is to when ignore SA config (can be rename to override SA) is selected, 
override only values that are not empty in the config. 
However, as there are probably fields that we wish to set it as empty, we probably need a `empty` or `default` key to keep it that way.

damithc:
Would this work?
If a field value in the csv file is prefixed with `override:`, the value in the SA config is ignored.

eugenepeh:
>If a field value in the csv file is prefixed with override:, the value in the SA config is ignored.

Sounds good to me

fzdy1914:
The #602 fixes part of the problem but not fully.

fzdy1914:
The priority of CLI is not decided.

fzdy1914:
Close as #430 deals with the same issue.

PR #536, opened 2019-02-03
[#534] Fixes untouched code button failure

### LINKED ISSUE #534: Code View: Show untouched code button does not work after #490
Broke in this commit: https://deploy-preview-490--reposense.netlify.com/#search=&sort=displayName&since=2017-10-09&until=2019-01-22&timeframe=day&reverse=false&repoSort=true

A commit before: https://deploy-preview-505--reposense.netlify.com/#search=&sort=displayName&since=2017-10-09&until=2019-01-22&timeframe=day&reverse=false&repoSort=true

fzdy1914:
Fixes #534 

In #490, the `toggleNext()` are edited to limit the toggle option to the path of the title.

However, the untouched code button also uses this function. So this button does not work after #390 is merged.

Let's return the `toggleNext()` to its original state, but edit the way to call this function.

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -166,7 +166,7 @@ html
           .empty(v-if="files.length===0") nothing to see here :(
           .file.active(v-for="file in selectedFiles", v-bind:key="file.path")
             .title
-              span.path(onclick="toggleNext(this)") {{ file.path }}&nbsp;
+              span.path(onclick="toggleNext(this.parentNode)") {{ file.path }}&nbsp;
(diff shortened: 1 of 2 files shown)


eugenepeh:
Nits to the proposed message:

```
The toggleNext() function were edited in #490 to limit the toggle
option to the path of the title.

However, as the untouched code button also taps on this function, it
has caused the show/hide untouched code button loses its
functionality. 

Let's restore the toggleNext() to the state before #490, and changes
the way to this function is called at the title to perform the required
result in #490.
```

PR #539, opened 2019-02-03
RampChart: show up to report generation date, not last commit date #518

### LINKED ISSUE #518: Ramp chart: show up to report generation date, not last commit date
When a report is generated with no `-until` flag, the ramp chart should show up to _today_, not the last commit date.

Example:
https://nus-cs3281.github.io/2019-dashboard (the three dates don't match)
image
image
image

emer7:
Fix #518 

```
In the UI user can choose to what date to show the report. However,
user can only choose up to the most recent commit, despite the user
may generate the report way beyond the most recent commit. This may
cause confusion to the user.

Let's fix this bug by showing up until the report generation date in the UI
instead.
```


emer7:
> The date range doesn't seem to be working now, I set Since Date 01/01/2019 to Until Date 02/10/2019, but the ramp chart doesnt' change, still showing commits from year 2018
> untitled

Fixed

emer7:
> Hmm, current work doesn't seems to have taken `-until` flag into consideration.
> 
> > When a report is generated with no `-until` flag, the ramp chart should show up to _today_, not the last commit date.
> 
> E.g. if an `until` flag is set, user would expect to see just the range entered instead of up to _today_

I don't think current implementation (not including this PR) takes `-until` flag into consideration for displaying the `until date`. Current implementation only shows up until the latest commit date in the range.

frontend/src/static/js/v_summary.js
```diff
@@ -223,12 +224,17 @@ window.vSummary = {
...
+
+      if (window.app.untilDate) {
+        maxDate = window.app.untilDate;
+      } else {
+        const creationDateArray = window.app.creationDate.split(' ');
+        const year = creationDateArray.pop();
+        creationDateArray.pop();
+        maxDate = window
+            .moment(`${creationDateArray.join(' ')} ${year}`, 'ddd MMM D HH:mm:ss YYYY')
+            .format('YYYY-MM-DD');
```
eugenepeh: Hmm... instead of doing this, perhaps we can just add the values to summary.json such that the `window.app.untilDate` will always be set. Better maintainability and one less library to import.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -66,7 +66,7 @@ public static void setDatesToRepoConfigs(
             List<RepoConfiguration> configs, Optional<Date> sinceDate, Optional<Date> untilDate) {
         for (RepoConfiguration config : configs) {
             config.setSinceDate(sinceDate.orElse(null));
-            config.setUntilDate(untilDate.orElse(null));
+            config.setUntilDate(untilDate.orElse(new Date()));
```
yamidark: Can add a comment above this, that the `new Date()` is for setting the `untilDate` to be the current/report generation date?

emer7:
> Changing status, as Travis and Appveyor are failing (seems like the `untilDate` is added to `summary.json`, but the system test files are not updated)

Do you have suggestion how to include the ever changing untilDate in summary.json in expected file?

eugenepeh:
>Do you have suggestion how to include the ever changing untilDate in summary.json in expected file?

Hmm... perhaps configure the system time on ci?

emer7:
> > Do you have suggestion how to include the ever changing untilDate in summary.json in expected file?
> 
> Hmm... perhaps configure the system time on ci?

@eugenepeh appveyor cannot change the system time though 😕

emer7:
> > > Do you have suggestion how to include the ever changing untilDate in summary.json in expected file?
> > 
> > 
> > Hmm... perhaps configure the system time on ci?
> 
> @eugenepeh appveyor cannot change the system time though 😕

@ongspxm any idea how to solve this?

ongspxm:
as discussed, move all the since and until date out to make it global so that you can do some magic string processing, or you could do it the same way as `dashboardGeneratedDate`

emer7:
> as discussed, move all the since and until date out to make it global so that you can do some magic string processing, or you could do it the same way as `dashboardGeneratedDate`

I think the magic string processing is better compared to doing it the same way as `dashboardGeneratedDate`, because if we are to insert it as how we insert `dashboardGeneratedDate`, I feel it will just defeat the purpose of testing.

eugenepeh:
> @eugenepeh appveyor cannot change the system time though 😕

Hmm, it may the jvm is not using the system time. Try make a change to print the system time from the code base and see what's the reason. If all else fails, try setting up a mock time for testing purposes.

emer7:
> > @eugenepeh appveyor cannot change the system time though 😕
> 
> Hmm, it may be that the jvm is not using the system time. Try make a change to print the system time from the code base and see what's the reason. If all else fails, try setting up a mock time for testing purposes.

It's appveyor's issue, it doesn't allow changing of sytem time

eugenepeh:
>It's appveyor's issue, it doesn't allow changing of sytem time

We can't be the first that face this issue since appveyor is a CI platform. 
Note that it is sort of a taboo to say that `this won't work` without exhausting the options, present the negative results or having the original developer of the software/platform to agree with you, unless its logically sound that it is impossible.

Anyway like I mentioned earlier, if it all else fails, try setting up a mock time for testing purposes.

eugenepeh:
On the other hand, as mentioned on the issue, it would be best to implement this after https://github.com/reposense/RepoSense/issues/542 is rectified.

Focus on https://github.com/reposense/RepoSense/issues/542 first

emer7:
> Just found a very questionable thing here:
> 
> Why is it that this has never triggered any error?
> image

Because this one has always been tested with mock time.
The structure of the report is such that `dashboardGeneratedTime` can be passed as a variable easily when writing a file, but `untilDate` is quite nested and harder to be mocked.

emer7:
> On the other hand, as mentioned on the issue, it would be best to implement this after #542 is rectified.
> 
> Focus on #542 first

Yes PR is coming soon

emer7:
Waiting for #596 to be merged

eugenepeh:
Some changes in the requirement: https://github.com/reposense/RepoSense/issues/610#issuecomment-474423709
Do take note of that when resuming this PR.

emer7:
@reposense/stage2-reviewers ready for review
Just realized the prerequisite PR has been merged

frontend/src/static/js/v_summary.js
```diff
@@ -273,22 +273,18 @@ window.vSummary = {
...
       let minDate = window.app.sinceDate;
-      let maxDate = '';
       this.filtered.forEach((repo) => {
         repo.forEach((user) => {
           const { commits } = user;
           if (commits.length) {
             const date1 = commits[0].date;
-            const date2 = commits[commits.length - 1].date;
             if (!minDate || minDate > date1) {
               minDate = date1;
```
eugenepeh: think this entire chunk can be removed as well?

emer7: @eugenepeh cannot because the commits[0] date for each user in each repo varies. `window.app.sinceDate` is also not always defined.

eugenepeh: I see 👍

emer7:
@yamidark @yong24s ready for review 🙏

yamidark:
Nits to proposed message:
```
In the UI, the user can choose the date range to show the report, which
is set to be until the date of the most recent commit, if no `--until`
flag is given.

However, the report may be generated at a date beyond the most recent
commit, and users would want to view all contributions up to the
report generation date instead.

Let's fix this bug by showing up until the report generation date in
the UI instead.
```

PR #541, opened 2019-02-04
CLI: -view should not accept a PATH when used with -config or -repos #476

### LINKED ISSUE #476: CLI: -view should not accept a PATH when used with -config or -repos
Following #446, `-view` can be used together with `-config` or `-repos`, in an atomic fashion, to launch the generated report after analysis.

In the discussion https://github.com/reposense/RepoSense/pull/446#discussion_r243760364, it was raised that `-view` when used together with `-config` or `-repos` should not be expecting an argument.

As it may unintuitive and insecure, `i.e.` `java -jar RepoSense.jar -config path/to/configs -output path/to/output -view /etc` will **not** execute analysis but the report will be launched at given **PATH** `/etc`.

myteo:
Fixes #476.

```
When -view option is given with an argument and is used together
with -config or -repos option, no analysis will be performed, but a
server will be launched at the path specified by the -view option.

This may be unintuitive and insecure for users as they might expect an
analysis to be performed since a -config or -repos option was given.

Lets fix this issue by performing the analysis when -view option is
given with an argument and used together with -config or -repos, and
launching a server to view the report generated by the analysis
afterwards.
```


myteo:
Made requested changes.

yamidark:
```
When -view option is given with an argument and is used together
with -config or -repos option, no analysis will be performed, but a
server will be launched at the path specified by the -view option.

This may be unintuitive and insecure for users as they might expect an
analysis to be performed since a -config or -repos option was given.

Lets fix this issue by performing the analysis when -view option is
given with an argument and used together with -config or -repos, and
launching a server to view the dashboard generated by the analysis
afterwards.
```
@yong24s  Can help me reivew/nit?

yong24s:
@yamidark 

Looks good. We may need to be consistent usage of `dashboard` and `report`. 🤔

yamidark:
> @yamidark
> 
> Looks good. We may need to be consistent usage of `dashboard` and `report`. 🤔

Hmm, guess I will switch over to `report` for now as specified in the UserGuide.

PR #524, opened 2019-01-31
[#465] Bookmark codeview uri

chelseyong:
Fixes #465.
```
Users can only bookmark/save the state of the main dashboard into url 
which can be shared with other people.

Currently, it does not allow user to bookmark the code view and
share it with the others.

Let's encode the code view into the url.
```

PR #544, opened 2019-02-07
Add highlighting to ramps

ongspxm:
```
To be able to open up a "zoomed-in" view of the ramp, the user must
first be able to select part of the ramp to focus on.

Let's add a way for the user to highlight the range of the ramp to
focus on. This will later translate into the date range for the
"zoomed-in" view to display the relevant commits.

In this particular implementation, we use a global `drags` object as a
way of preventing user from highlighting on multiple ramp charts. E.g.
if a user mousedown on one ramp chart and mouseup on the other, the
second chart will be highlighted with the positions defined between the
two mouse events.
```


ongspxm:
@yamidark, @yong24s, @eugenepeh  if i were to use the tab functions in #524, should i just implement the ramp selection here (the ui part) and do the actual display of the information in the tab as another pr?

eugenepeh:
>@yamidark, @yong24s, @eugenepeh if i were to use the tab functions in #524, should i just implement the ramp selection here (the ui part) and do the actual display of the information in the tab as another pr?

no need. both should be part of this feature.

ongspxm:
@yamidark can create a new branch for me to merge into instead? i think this gonna be pretty big

yamidark:
>@yamidark can create a new branch for me to merge into instead? i think this gonna be pretty big

Done, `ramp-chart-zoom` branch created

PR #547, opened 2019-02-08
Custom grouping of files

### LINKED ISSUE #520: Custom grouping of files
This is an end-to-end feature to allow users to specify custom groupings of files in a repository e.g. ```test, docs, etc.```.

A new column called ```Groups``` can be added to ```repo-config.csv```, which allows users to specify groupings of files through a text file. The text file can contain the group(s), and filepaths belonging to that group e.g. ```test: src/test/java/Test.java```

This information can be propagated to the front-end. When groupings are specified by users, these groups can be added as a checkbox option in the Code View, such that you can only view files belonging to that group in the Code View when it is selected. 

Since this is an end-to-end feature, it could be done incrementally, by implementing the backend first, followed by front-end.

apoorva17:
Fixes #520 

```
With the addition of #385, users are able to view a breakdown by filetype. 

Let's allow users to specify their own custom groups for files in a repository to allow
them to view a breakdown by their specified groups in the dashboard.

The user can specify a config file which contains the groups and file path globs to be 
included in that group in 'repo-config.csv' e.g. 'repoName-groups-config.csv' for a given 
repository.

If no group is assigned to a file, it is assigned the default group: `other`.

If groups are specified by the user:
* Code view will show a breakdown by groups instead of filetypes. Users can tick and 
filter files by group.
* Contribution Bar will show a breakdown by groups when the 'breakdown by filetype\group'
option is selected. 
```


apoorva17:
This is an end-to-end feature which will include both changes to the backend and frontend. 
I've currently implemented only the backend, and would appreciate initial feedback on the implementation from @reposense/devs. System tests have not yet been updated, so please ignore the failing tests for now.

apoorva17:
For the frontend, should the user only view breakdown by groups if he specifies the groupings for a repository, or should he be able to view both breakdown by filetype and breakdown by groups @damithc @reposense/devs ?

damithc:
> For the frontend, should the user only view breakdown by groups if he specifies the groupings for a repository, or should he be able to view both breakdown by filetype and breakdown by groups @damithc @reposense/devs ?

Let's limit it just to groups.

How are you planning to deal with overlapping groups?

apoorva17:
> How are you planning to deal with overlapping groups?
Each file should only have 1 group, so I will assign the later group to the file.

damithc:
> Each file should only have 1 group, so I will assign the later group to the file.

That's a bit too restrictive. A file may need to be categorized as 'test files' as well as 'documentation'.  Is it possible to have no such restriction? It can be added as a separate PR.

apoorva17:
> > Each file should only have 1 group, so I will assign the later group to the file.
> 
> That's a bit too restrictive. A file may need to be categorized as 'test files' as well as 'documentation'. Is it possible to have no such restriction? It can be added as a separate PR.

I can add support for multiple groups in this PR, but wouldn't that double-count the lines contributed to both the groups? For instance, if a file is classified as 'test' and documentation', then the lines contributed for that file would count for both 'test' and 'documentation'.

damithc:
> I can add support for multiple groups in this PR, but wouldn't that double-count the lines contributed to both the groups? For instance, if a file is classified as 'test' and documentation', then the lines contributed for that file would count for both 'test' and 'documentation'.

Yes, that's fine, as long as it doesn't distort the total. If the user defines overlapping groups, he should be aware that group totals will not add up to the grand total.

eugenepeh:
A bit confused with this `grouping` concept. How are you going to present it on the frontend and how does it help the user than `search bar` already has?

A simpler alternative that I can think of to implement this would be allowing user to set `Quick keys` for the search bar. E.g. Some buttons such as `[**.java]`, `[**.css|**.fxml|**.html]` where clicking it automatically update the `Search bar` filter and show the desired results. In such case, user don't have to create a dedicated `csv` file and attach it to each repository. This can easily handle the overlapping groups and not cause any drastic changes to the code view. The group totals can simply be the summary of the search result just like google search engine e.g. `About 17,850,000,000 results (0.82 seconds)`. 

@damithc what do you think?

damithc:
> A bit confused with this `grouping` concept. How are you going to present it on the frontend and how does it help the user than `search bar` already has?

The _grouping_ concept is useful when differentiating by file type is too primitive to fit the situations. For example, users might want to divide the work into `functional code`, `test code`, and `documentation` (or `front-end` and `back-end`). Each group will have a name and a glob that indicates which files it contains. These globs need to be predefined by the person who sets up the reposense; we can't expect readers of the report to type in these globs into the search bar. 

How about the following, to keep things simpler?
* If the user has defined groups, we use groups. If not, we use file types. 
* Keep the filter bar separate from tick boxes. i.e., when one is used, the other is disabled.

image

apoorva17:
I've updated the implementation for the front-end for groups. The code view shows the groups instead of filetypes when groups are specified by user for a repository. Upon selecting ```breakdown by filetype/group```, the contribution bar will reflect a breakdown by groups if groups are specified, otherwise it should show the breakdown by filetype once #487 is merged.

apoorva17:
There is a lot of overlapping code from #487 which I will optimise once #487 is merged.

apoorva17:
@reposense/devs kindly provide your review for this PR.

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -22,9 +25,17 @@
...
+    private static final int GROUPS_CONFIG_POSITION = 6;
+
+    private static String configFolderPath;
 
     public RepoConfigCsvParser(Path csvFilePath) throws IOException {
         super(csvFilePath);
+        if (csvFilePath.getParent() != null) {
+            this.configFolderPath = csvFilePath.getParent().toString() + '/';
+        } else {
+            this.configFolderPath = "./";
```
eugenepeh: Why is this change necessary?

Also, this is not a very clean way to do it, seems rather OS dependent. Can you please do it the same way as how the `-config` is parsed?

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -51,6 +62,14 @@ protected void processLine(List<RepoConfiguration> results, String[] elements) t
         String ignoreStandaloneConfig = getValueInElement(elements, IGNORE_STANDALONE_CONFIG_POSITION);
         List<CommitHash> ignoreCommitList = CommitHash.convertStringsToCommits(
                 getManyValueInElement(elements, IGNORE_COMMIT_LIST_CONFIG_POSITION));
+        List<Group> groups = Collections.emptyList();
```
eugenepeh: why is the group field a list? Is it even meant to be a multi-value field?

apoorva17:
@eugenepeh If you click the ```breakdown by filetyp/group``` option in netlify, you should be able to see the breakdown for repositories where grouping is specified, which is ```reposense_testrepo-Delta-master``` and ```reposense_RepoSense_master```. As for the other repositories which still show default, I'm waiting for #487 to merge the changes in so that it will reflect breakdown by filetype instead.

src/main/java/reposense/parser/RepoConfigCsvParser.java
apoorva17: Each repository can have multiple groups, which is extracted from the groups-config.csv file using ```GroupConfigCsvParser``` and stored in this list.

src/main/java/reposense/parser/RepoConfigCsvParser.java
apoorva17: I need the ```configFolderpath``` in order to retrieve the ```groups-config.csv``` file, which contains the groupings specification, and pass it to ```GroupConfigCsvParser```.

apoorva17: Sorry, I'm not sure how to do it as ```-config``` is parsed since it makes use of ```Argument``` and ```ArgumentParser```. Is there a way I can retrieve the configFolder parsed in the arguments and use it here?

eugenepeh: >Sorry, I'm not sure how to do it as -config is parsed since it makes use of Argument and ArgumentParser. Is there a way I can retrieve the configFolder parsed in the arguments and use it here?

There are several ways to go about doing that, I leave it to you to figure it out as I do not have much time recently to launch my development pc. 
For example, your `groups-config.csv` needs to be on the same folder as `repo-config.csv` and both of them use fixed names. So why don't you just pass in the `configFolderPath` in the first case?

apoorva17: > For example, your `groups-config.csv` needs to be on the same folder as `repo-config.csv` and both of them use fixed names. So why don't you just pass in the `configFolderPath` in the first case?

The problem is, ```groups-config.csv``` is not a fixed name since there can be multiple ```groups-config.csv``` for each repository. For instance, for ```reposense-Alpha```, the filename can be ```alpha-groups-config.csv```, and for ```reposense-Delta```, the filename can be ```delta-groups-config.csv```.

eugenepeh: Mistake on my part of comment there, but u only need the folder path don't you?
In that case, since `repo-config.csv` is fixed, you can do so.

eugenepeh:
By the way, whats the rationale that we don't do it like the `author-config.csv` style such that we will only need 1 `groups-config.csv`?

E.g. 
`Repository's Location, Branch, Group Name, Globs`
Branch actually can be omitted because it doesn't make sense if the files organization is structured differently.

apoorva17:
> By the way, whats the rationale that we don't do it like the `author-config.csv` style such that we will only need 1 `groups-config.csv`?
> 
> E.g.
> `Repository's Location, Branch, Group Name, Globs`
> Branch actually can be omitted because it doesn't make sense if the files in the organization are structured differently.

Yes, I suppose it could be done in a way similar to `author-config.csv`. @damithc your opinion on this?

docs/UserGuide.md
```diff
@@ -290,6 +290,17 @@ Repository's Location | The `GitHub URL` or `Disk Path` to the git repository e.
...
+#### `groups-config.csv`
+
+Optionally, you can provide a `groups-config.csv`(which should be in the same directory as `repo-config.csv` file) to provide more details about any groupings of files in the repository. It should contain the following columns:
+
+Column Name | Explanation
+----------- | -----------
+Group Name | Name of the group e.g.,`test`
+Globs | The list of file path globs to include for specified group. e.g.,`**/test/*;**.java`
+
+Note that each file can only have 1 group. If more than 1 group is specified for a given file, the latter group is set for the file.
```
damithc: This sentence is misleading. It sounds like each csv file can have only one group.

damithc:
> Yes, I suppose it could be done in a way similar to `author-config.csv`. @damithc your opinion on this?

Yes, let's do that. There will be only one `groups-config.csv` but we should still be able to specify different groups for different repos, right?

docs/UserGuide.md
apoorva17: I've updated the User Guide for `groups-config.csv`, hopefully it is clearer now.

apoorva17:
> By the way, whats the rationale that we don't do it like the `author-config.csv` style such that we will only need 1 `groups-config.csv`?
> 
> E.g.
> `Repository's Location, Branch, Group Name, Globs`
> Branch actually can be omitted because it doesn't make sense if the files in the organization are structured differently.

@reposense/devs i've updated the implementation to use a single `groups-config.csv` file, similar to `author-config.csv`. 

I've also integrated the changes in #487 into this PR. Now, upon clicking the `breakdown by file format/group` option, you should be able to view the breakdown by groups in Contribution Bar if user has specified any groups for that repository, otherwise breakdown by file format is shown. 

The checkboxes in the code view shows groups, if user has specified groups, otherwise file formats. Note that the glob filter does not support any form of filtering by groups, it only filters by file glob paths.

apoorva17:
@reposense/devs ready for review.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -39,12 +40,14 @@ public RepoConfiguration(RepoLocation location) {
     }
 
     public RepoConfiguration(RepoLocation location, String branch) {
-        this(location, branch, Collections.emptyList(), Collections.emptyList(), false, Collections.emptyList());
+        this(location, branch, Collections.emptyList(), Collections.emptyList(), false,
+            Collections.emptyList());
```
eugenepeh: unnecessary change

src/main/java/reposense/model/Group.java
```diff
@@ -0,0 +1,89 @@
...
+    @Override
+    public int hashCode() {
+        return groupName.hashCode();
+    }
+
+    /**
+     * Converts all the strings in {@code groups} into {@code Group} objects. Returns null if {@code groups} is null.
+     * @throws IllegalArgumentException if any of the strings are in invalid formats.
+     */
+    public static List<Group> convertStringsToGroups(List<String> groups) throws IllegalArgumentException {
```
eugenepeh: This seems to be used for test case and not meant to be part of main code. 
Do not mix the 2 of them, if its only used for test case, this code should be in test folder.

src/main/java/reposense/authorship/model/AuthorshipSummary.java
```diff
@@ -15,40 +16,71 @@
...
+        // Add file format contribution count
         Map<String, Integer> fileFormatContributionMap = authorFileFormatContributionMap.get(author);
         String fileFormat = filePath.substring(filePath.lastIndexOf('.') + 1);
         fileFormat = fileFormat.isEmpty() ? "others" : fileFormat;
         fileFormatContributionMap.put(fileFormat, fileFormatContributionMap.getOrDefault(fileFormat, 0) + 1);
+
+        // Add group contribution count
+        if (group != null) {
+            Map<String, Integer> groupContributionMap = authorGroupContributionMap.get(author);
+            groupContributionMap.put(group, groupContributionMap.getOrDefault(group, 0) + 1);
```
eugenepeh: It seems like on the UI, you only utilizes one of them; if the group exists, you don't show the formats. 
In that case, why do you still need to calculate formats for the repo if you're not utilizing them?
As a benefit, you wouldn't even need to modify the frontend that much if that is the case.

src/main/java/reposense/model/Group.java
apoorva17: I added this method to `Group`, as it is done similarly for `Format (convertStringsToFormats)`. Should I move this method to `RepoConfigurationTest` then?

src/main/java/reposense/authorship/model/AuthorshipSummary.java
apoorva17: File format information is still needed for the checkboxes in the Code Panel when the `breakdown by file format/group` option is not selected, as the default view does not make use of the breakdown by grouping.

apoorva17: Wouldn't it be better to retain the file format information for such repos as well, in case the user specifies it in `repo-config.csv`? @damithc your opinion?

src/main/java/reposense/model/Group.java
eugenepeh: >I added this method to Group, as it is done similarly for Format (convertStringsToFormats). 

I see. Probably missed out in the reviews. All non-functional test code should be in test folder.

>Should I move this method to RepoConfigurationTest then?

Can move it to a relevant TestUtil class or create one, we might need it in other area in the future.

src/main/java/reposense/authorship/model/AuthorshipSummary.java
apoorva17: @eugenepeh there are 2 maps being passed to front-end now: `authorFileFormatContributionMap` and `authorGroupContributionMap`. If we are merging it into 1 map to be passed, what should it be renamed to? `authorBreakdownContributionMap`? References in front-end will also have to be changed accordingly e.g. `isSelectedFileTypes` and `isSelectedGroups` to `isSelectedBreakdown`

eugenepeh: If let's say they were to be mutually exclusive, we can just simplify the 2 into a single element such that there is no distinction between `java` and `code` on the frontend. This helps to enhance maintainability and computation. 

Do consult @damithc 's opinion on the ideal outcome though.

damithc: Do what is optimal for the current feature set. Let's not complicate the code for the sake of possible future features.

src/main/java/reposense/model/Group.java
apoorva17: Ok, i've moved this method to `TestUtil`.

src/main/java/reposense/authorship/model/AuthorshipSummary.java
apoorva17: > If let's say they were to be mutually exclusive, we can just simplify the 2 into a single element such that there is no distinction between `java` and `code` on the frontend. This helps to enhance maintainability and computation.
> 
> Do consult @damithc 's opinion on the ideal outcome though.

@eugenepeh I've updated the implementation accordingly, to use `format` to refer to both `fileFormat` and `group`. So now, there is no distinction between `java` and `code`.

apoorva17:

apoorva17:
> To avoid the confusion between `format` and `file format`, I think we can name it something else. E.g. `fileType`.
> 
> I think the code can be simplify even further such that there is no distinction between group and format in the backend.
> 
> E.g. both `group` and `format` can be stored in a single list.
> You just have to append `**.` to the format so that they become globs as well.
> When there's group set to the repo, can replace the `formats glob` with the `group glob`.

I need the distinction in the backend for `AuthorshipSummary.java` for initialising the `authorFormatContributionMap` with the defaults for file format/groups, as the default file formats and default groups are different.

src/systemtest/resources/dateRange/expected/reposense_testrepo-Beta_master/authorship.json
```diff
@@ -1,6 +1,7 @@
 [
   {
     "path": "src/main/resources/view/MainWindow.fxml",
+    "format": "fxml",
```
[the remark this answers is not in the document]
apoorva17: ok, this has been updated accordingly.

eugenepeh:
>I need the distinction in the backend for AuthorshipSummary.java for initialising the authorFileTypeContributionMap with the defaults for file format/groups, as the default file formats and default groups are different.

Do you mean this part?

```
        if (!groups.isEmpty()) {
            // group contribution
            authors.forEach((author) -> {
                LinkedHashMap<String, Integer> defaultGroupContribution = new LinkedHashMap<>();
                for (Group group : groups) {
                    defaultGroupContribution.put(group.toString(), 0);
                }
                defaultGroupContribution.put(Group.DEFAULT_GROUP, 0);
                authorFileTypeContributionMap.put(author, defaultGroupContribution);
            });
        } else {
            authors.forEach((author) -> authorFileTypeContributionMap.put(author, new LinkedHashMap<>()));
            // file format contribution
            authors.forEach((author) -> {
                LinkedHashMap<String, Integer> defaultFileFormatContribution = new LinkedHashMap<>();
                for (Format format : formats) {
                    defaultFileFormatContribution.put(format.toString(), 0);
                }
                authorFileTypeContributionMap.put(author, defaultFileFormatContribution);
            });
        }
```

apoorva17:
> Do you mean this part?
> 
> ```
>         if (!groups.isEmpty()) {
>             // group contribution
>             authors.forEach((author) -> {
>                 LinkedHashMap<String, Integer> defaultGroupContribution = new LinkedHashMap<>();
>                 for (Group group : groups) {
>                     defaultGroupContribution.put(group.toString(), 0);
>                 }
>                 defaultGroupContribution.put(Group.DEFAULT_GROUP, 0);
>                 authorFileTypeContributionMap.put(author, defaultGroupContribution);
>             });
>         } else {
>             authors.forEach((author) -> authorFileTypeContributionMap.put(author, new LinkedHashMap<>()));
>             // file format contribution
>             authors.forEach((author) -> {
>                 LinkedHashMap<String, Integer> defaultFileFormatContribution = new LinkedHashMap<>();
>                 for (Format format : formats) {
>                     defaultFileFormatContribution.put(format.toString(), 0);
>                 }
>                 authorFileTypeContributionMap.put(author, defaultFileFormatContribution);
>             });
>         }
> ```

Yes

eugenepeh:
wouldn't it be the same if both are treated as globs? if none match the glob, place it in `others`.

apoorva17:
@eugenepeh so you want me to combine `Format` and `Group` located in `model`, to `FileType`? But the parameters accepted by both are different, constructor for `Format` is `Format(String format)`, and constructor for `Group` is `Group(String groupName, List<String> filePaths)`. Also, there are many usages of `Format` throughout the code which needs to be refactored. Since this PR is already a big PR, i feel that such a simplification can be considered in a separate future PR instead.

eugenepeh:
> constructor for Format is Format(String format), and constructor for Group is Group(String groupName, List<String> filePaths)

Both can reach a consensus such that format can have its equivalent file path to be `java` -> `**.java`, so that is a minor issue. 

>Also, there are many usages of Format throughout the code which needs to be refactored. Since this PR is already a big PR, i feel that such a simplification can be considered in a separate future PR instead.

Refactoring can be done in separate PR, but logic and code wise should aim to target it within this one. Since this is one of your last PR before the wrap up of your FYP, try to make it a good one. A big PR like this can raise a good discussion, discussing the various techniques to tackle the problem. You can emphasize strongly on the point that due to the similarity, both can be combined to have better/efficient  code utilization and maintainability.

apoorva17:
> Refactoring can be done in separate PR, but logic and code wise should aim to target it within this one. 

The logic and code are already targeting this, as the distinction has already been removed from the front-end. What additional changes are you specifically looking for in the back-end for this PR?

eugenepeh:
>The logic and code are already targeting this, as the distinction has already been removed from the front-end. What additional changes are you specifically looking for in the back-end for this PR?

As mentioned above, the integration of group n format into a single component. So that there isn't any redundant codes to perform distinct classification between the 2.

apoorva17:
> As mentioned above, the integration of group n format into a single component. So that there isn't any redundant codes to perform distinct classification between the 2.

As I have also mentioned above, integrating group and file format into a single component requires a lot of refactoring for format, therefore it should be done in a separate PR. If the refactoring is to be done in a separate PR, what additional changes are you looking for?

eugenepeh:
>As I have also mentioned above, integrating group and file format into a single component requires a lot of refactoring for format, therefore it should be done in a separate PR. If the refactoring is to be done in a separate PR, what additional changes are you looking for?

What I meant was the refactoring of variable names can be done in a separate PR, but redundant code should therefore be removed in this. For example the code i pointed out earlier:

```
        if (!groups.isEmpty()) {
            // group contribution
            authors.forEach((author) -> {
                LinkedHashMap<String, Integer> defaultGroupContribution = new LinkedHashMap<>();
                for (Group group : groups) {
                    defaultGroupContribution.put(group.toString(), 0);
                }
                defaultGroupContribution.put(Group.DEFAULT_GROUP, 0);
                authorFileTypeContributionMap.put(author, defaultGroupContribution);
            });
        } else {
            authors.forEach((author) -> authorFileTypeContributionMap.put(author, new LinkedHashMap<>()));
            // file format contribution
            authors.forEach((author) -> {
                LinkedHashMap<String, Integer> defaultFileFormatContribution = new LinkedHashMap<>();
                for (Format format : formats) {
                    defaultFileFormatContribution.put(format.toString(), 0);
                }
                authorFileTypeContributionMap.put(author, defaultFileFormatContribution);
            });
        }
```

Its a pretty straight forward concept, turn the formats to use the `Group` structure so that there is only a single `List` that maintains the 2 in the repository.
When there is a line in group config targeting specific repository, clear the list and fill it with the group.

damithc:
On the one hand we don't want to deteriorate the code base quality for the sake of adding a feature. If the dev disappears after merging the feature, someone else has to clean up after, which is not nice. On the other hand we don't want to make the PR grow too big either. How about we open a long-lived feature branch? First merge this feature to the new branch. Then, clean up the code as a new PR and merge that to the new branch. After that we can merge the new branch back to master.

eugenepeh:
>On the one hand we don't want to deteriorate the code base quality for the sake of adding a feature. If the dev disappears after merging the feature, someone else has to clean up after, which is not nice. On the other hand we don't want to make the PR grow too big either. How about we open a long-lived feature branch? First merge this feature to the new branch. Then, clean up the code as a new PR and merge that to the new branch. After that we can merge the new branch back to master.

Created the branch, `feature-grouping`. Can merge this towards that branch for to consolidate the changes towards the feature.

apoorva17:
> On the one hand we don't want to deteriorate the code base quality for the sake of adding a feature. If the dev disappears after merging the feature, someone else has to clean up after, which is not nice. On the other hand we don't want to make the PR grow too big either. How about we open a long-lived feature branch? First merge this feature to the new branch. Then, clean up the code as a new PR and merge that to the new branch. After that we can merge the new branch back to master.

@damithc I view what @eugenepeh is suggesting as an extension to this PR, and not as cleaning up of the code. The scope of this PR was to introduce groups, which I have done. If @eugenepeh wants to change the way `Format` works (change `java` to `**.java`), then this is a completely separate issue as I'm not touching `Format` at all in this PR.  Therefore, I believe that code base quality has not deteriorated.

Also, in my earlier PRs, when I found bugs in the existing implementation, not related to the issue I was working on in the PR, I have been told by the devs that "It is a courtesy to fix things that are found broken when doing your own work; software engineering is never about individualism, its a collaborative effort. " If I am expected to clean up after other developer's code, even in circumstances when I was not familiar with what was being used in that section, then I believe such a team spirit should be expected from all the developers.

eugenepeh:
>@damithc I view what @eugenepeh is suggesting as an extension to this PR, and not as cleaning up of the code. The scope of this PR was to introduce groups, which I have done. If @eugenepeh wants to change the way Format works (change java to **.java), then this is a completely separate issue as I'm not touching Format at all in this PR. Therefore, I believe that code base quality has not deteriorated.

The suggestion was made due to the code was unnecessary complicated, which will have to be eventually removed or rectify by another developer who need to enhance / rework / touch the targeted portion.

>Also, in my earlier PRs, when I found bugs in the existing implementation, not related to the issue I was working on in the PR, I have been told by the devs that "It is a courtesy to fix things that are found broken when doing your own work; software engineering is never about individualism, its a collaborative effort. " If I am expected to clean up after other developer's code, even in circumstances when I was not familiar with what was being used in that section, then I believe such a team spirit should be expected from all the developers.

This is because everyone has the responsibility to maintain the code base. When a bug is found when working on a portion done by someone else, it is also the responsibility of others to clean it up because no one catches the bug when the work is under reviewed in the first place. Hence, everyone is equivalent responsible for causing it.

damithc:
Say the software has feature X. It was implemented without thinking that there will be a feature Y later.
Then we add feature Y without touching feature X. But now, the whole design is sub optimal state because there is a better way to implement X+Y which requires changes to X as well. Is that what's happening here? I'm asking because I haven't been following the details of this discussion.

eugenepeh:
>Say the software has feature X. It was implemented without thinking that there will be a feature Y later.
Then we add feature Y without touching feature X. But now, the whole design is sub optimal state because there is a better way to implement X+Y which requires changes to X as well. Is that what's happening here? I'm asking because I haven't been following the details of this discussion.

Yes, the similarity between the two allows better coexistence which can leads to more efficient code and performance increase on the front end.

damithc:
> Yes, the similarity between the two allows better coexistence which can leads to more efficient code and performance increase on the front end.

I see. That means there are two ways to implement this feature. 
* The current approach: takes less effort, less modifications, but results in a technical debt because the resulting code is more complex than necessary. 
* The other approach: requires changes to the existing code (i.e., bigger blast radius), takes more work, but avoids the technical debt, and as claimed, improves performance.

Is this your understanding too @apoorva17 ?

apoorva17:
> I see. That means there are two ways to implement this feature.
> 
> * The current approach: takes less effort, less modifications, but results in a technical debt because the resulting code is more complex than necessary.
> * The other approach: requires changes to the existing code (i.e., bigger blast radius), takes more work, but avoids the technical debt, and as claimed, improves performance.
> 
> Is this your understanding too @apoorva17 ?

@damithc, as per my understanding, making any modifications to `Format` is not within the scope of this PR. The existing implementation is already in a sub optimal state, due to the inefficient implementation of `Format` by a previous developer. The introduction of groups has only surfaced this existing inefficiency. Therefore, this should be resolved as a completely separate issue, as this PR does not touch `Format` at all. Moreover, I have removed the distinction for file formats and groups in the frontend, improving performance on the frontend.

damithc:
> The existing implementation is already in a sub optimal state, due to the inefficient implementation of `Format` by a previous developer. The introduction of groups has only surfaced this existing inefficiency.

Noted @apoorva17. Please create a new PR against the `feature-grouping` using the same code. We'll merge it to the branch (which indicates the acceptance of this code) and leave it there until someone fixes the technical debt in concern, after which we'll merge the branch back to master and release the feature.

apoorva17:
> Noted @apoorva17. Please create a new PR against the `feature-grouping` using the same code. We'll merge it to the branch (which indicates the acceptance of this code) and leave it there until someone fixes the technical debt in concern, after which we'll merge the branch back to master and release the feature.

@damithc thank you for your understanding. I've created a new PR against `feature-grouping` using the same code as requested.

eugenepeh:
@apoorva17 , the base has been updated to match master. 
Please resolve the conflicts and ping me / yamidark for merge afterward.

apoorva17:
@eugenepeh has any changes been made to the way file format is computed? In the `master` branch, when i select `breakdown by file format`, the total length of the contribution bar (with breakdowns) no longer matches the total length of contribution bar (in default view without breakdown). Therefore, it seems one of the changes added to master has touched this functionality?

eugenepeh:
>@eugenepeh has any changes been made to the way file format is computed? In the master branch, when i select breakdown by file format, the total length of the contribution bar (with breakdowns) no longer matches the total length of contribution bar (in default view without breakdown). Therefore, it seems one of the changes added to master has touched this functionality? If no changes were made to file format, may i know why this has changed?

@apoorva17 this seems to be broken by the recent update but not by your feature. Feel free to put up an issue to explain and highlight on that.

eugenepeh:
opened issue #637 on that

PR #551, opened 2019-02-11
Remove redundant commands in Travis config

yong24s:
```
In the Travis config file, we specify our Linux environment to use
JDK9 with `jdk: oraclejdk9` and `addons.apt.packages key`, but the
latter is redundant. Similarly, the command used to install frontend npm
dependencies can be simpified. The cd command does not change the
install location of the npm dependencies as frontend folder does not
contain a package.json file.

Let's remove the redundant commands for clarity and maintainability.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,28 +2,23 @@ language: java
 
 matrix:
   include:
-    - os: linux
+    - name: Linux Ubuntu 14.04 JDK 9
+      os: linux
       jdk: oraclejdk9
-    - os: osx
+    - name: MacOS 10.13 JDK 9
+      os: osx
 
 script:
   - ./config/travis/run-checks.sh
   - npm run lint frontend/src/**/*.js
   - time travis_retry ./gradlew clean checkstyleMain checkstyleTest test systemTest
 
-addons:
-  apt:
-    packages:
-      - oracle-java9-installer
-
-before_install:
-  - cd ./frontend; npm install --only=dev; cd ..
-
 before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
 
 cache:
+  npm: true,
   directories:
     - $HOME/.gradle/caches/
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,28 +2,24 @@ language: java
 
 matrix:
   include:
-    - os: linux
+    - name: Ubuntu 14.04 JDK 9
+      os: linux
       jdk: oraclejdk9
```
yamidark: Shall we change this back to oraclejdk8, which is the Java compatibility we should we using (as per Gradle and our DeveloperGuide)

yong24s: I believe in #530, @eugenepeh wants us to move to JDK11.

As we target `JDK8` in Gradle, the Jar should be JDK8 compliant. 🤔

yamidark: Yes we should plan to move to JDK11 soon. But currently we are still at JDK8, so our tests environment should be running in JDK8 as well.

yong24s:
@ongspxm 

`cd ./frontend; npm install --only=dev; cd ..` does not install dependencies in `frontend/` but at project directory. Is this intended?

```
$ tree RepoSense
RepoSense
├── frontend
└── package.json

$ (cd RepoSense/frontend/ && npm install ---only=dev)
npm notice created a lockfile as package-lock.json. You should commit this file.
...

$ tree RepoSense -L 1
RepoSense
├── frontend
├── node_modules
├── package.json
└── package-lock.json

```

ongspxm:
yep, the npm commands assumes root at the project root

.travis.yml
yong24s: Switching of JDK versions should be done in another PR? 
Since it involves AppVeyor too as it is using JDK9 `JAVA_HOME: C:\Program Files\Java\jdk9`.

yamidark: >Switching of JDK versions should be done in another PR?
Since it involves AppVeyor too as it is using JDK9 `JAVA_HOME: C:\Program Files\Java\jdk9`.

Hmm, yea probably can raise a different issue for that

PR #556, opened 2019-02-13
Downgrade to JDK8 in CIs

yong24s:
```
RepoSense is advertised to run on JDK 1.8.0_60 and later.

However, we use JDK 9 in our CIs.

Let's downgrade to use JDK 8 in CIs to validate this non-functional
requirement.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,17 +9,22 @@ matrix:
     - name: macOS 10.13.6 JDK 11.01
       os: osx
       osx_image: xcode10.1
+      before_install:
+        - export HOMEBREW_NO_AUTO_UPDATE=1; # Prevent time-consuming updating
+        - brew install cask;
+      install:
+        - brew cask install caskroom/versions/java8
+        - java -version
 
 script:
   - ./config/travis/run-checks.sh
   - npm run lint frontend/src/**/*.js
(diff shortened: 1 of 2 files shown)


fzdy1914:
For downgrading to jdk8, the add-on that has been deleted may be necessary to make sure that the jdk8 used is its latest version.

yamidark:
In Travis, the 2 different OSes seems to be using different JDK 8 version though:
Ubuntu:
untitled

macOs:
untitled

Can check to see if we can force both of them to use the same version (our UG and DG states `JDK 1.8.0_60 or later`)

yong24s:
@yamidark 

I don't think AppVeyor offers a way to update their JDK. It comes preinstalled as `JDK 1.8 Update 162`.

eugenepeh:
Doesn't the Mac os clash with one of the ongoing PR?

yong24s:
@eugenepeh 

This PR just downgrades to JDK 8.

After this PR gets merged, then in #559, we can test if we can use the latest macOS image for the latest git and downgrade to JDK 8 or just have another environment for the check scripts.

yamidark:
@yong24s #559 has been merged, do resolve the conflicts and continue with this PR.

PR #558, opened 2019-02-14
Refactor window.location.hash update methods

ongspxm:
in response to #524 

```
`window.location.hash` is used and updated in multiple places.
This results in duplication of code for writing to and using of
hash parameters.

Let's refactor the way we update the location hash by creating
functions for accessing and using hashes: `addHash`, `removeHash`,
`encodeHash` and `decodeHash`.

All these functions will be implemented with `window.hashParams` as the
main way of access and writing to `location.hash`.
```


chelseyong:
Should `updateHash` be implemented here as well? If not, I can do it in #524 as it is needed to do a one-time hash update.

ongspxm:
cos i making the chnages for v_summary as well. tbh, you can just pull into
yours and merge haha.

frontend/src/static/js/main.js
```diff
@@ -3,16 +3,32 @@ window.REPOS = {};
...
 
-  const hash = [];
-  const enquery = (key, val) => `${key}=${encodeURIComponent(val)}`;
-  Object.keys(hashParams).forEach((hashKey) => {
-    hash.push(enquery(hashKey, hashParams[hashKey]));
-  });
+  window.location.hash = Object.keys(hashParams)
+      .map((key) => (hashParams[key]
+        ? `${key}=${encodeURIComponent(hashParams[key])}` : ''))
+      .filter((x) => x).join('&');
```
chelseyong: Why not filter out removed params in `rmHash`? Since this also means that no key can contain a `false` boolean value.

ongspxm: gd catch

ongspxm:
@chelseyong check pls

frontend/src/static/js/main.js
```diff
@@ -3,16 +3,31 @@ window.REPOS = {};
 
 window.hashParams = {};
 window.addHash = function addHash(newKey, newVal) {
+  window.hashParams[newKey] = newVal;
+};
+window.rmHash = function rmHash(key) {
```
yamidark: Perhaps rename as `removeHash`?

yamidark:
Nits to proposed message:
```
`window.location.hash` is used and updated in multiple places.
This results in duplication of code for writing to and using of
hash parameters.

Let's refactor the way we update the location hash by creating
functions for accessing and using hashes: `addHash`, `removeHash`,
`encodeHash` and `decodeHash`.

All these functions will be implemented with `window.hashParams` as the
main way of access and writing to `location.hash`.
```

PR #559, opened 2019-02-14
Travis: check scripts do not work in Travis macOS environment #557

### LINKED ISSUE #557: Travis: check scripts do not work in Travis macOS environment
Following #414, the line to reinstall `git` with `pcre2` was removed in Travis macOS environment. As running `brew reinstall --with-pcre2 git` would caused Travis to fail. Furthermore, `pcre2` now ships with `git`.

Since its removal, check scripts, that depends on `pcre2`, do not work but Travis macOS still passes. This is due to the use of an older version of git which does not comes with `pcre2`.

Let's upgrade git in Travis macOS environment to ensure check script runs correctly.

```
$ ./config/travis/run-checks.sh
fatal: cannot use Perl-compatible regexes when not compiled with USE_LIBPCRE
fatal: cannot use Perl-compatible regexes when not compiled with USE_LIBPCRE
The command "./config/travis/run-checks.sh" exited with 0.
```

yong24s:
Fixes #557 

```
Following #414, the line to reinstall `git` with `pcre2` was removed in
Travis macOS environment. As running `brew reinstall --with-pcre2 git`
would caused Travis to fail. Furthermore, `pcre2` now ships with
`git >= 2.19`.

Since its removal, check scripts, that depends on `pcre2`, do not work
but Travis macOS still passes. This is due to the use of an older
version of git which does not comes with `pcre2`.

Let's use the latest version of Travis macOS image, that comes with git
2.19.1, to ensure check scripts run correctly.
```


yong24s:
@eugenepeh 

`brew update` takes up most of the time, which is implicitly called by `brew upgrade git`.

Without updating `brew`, we cannot upgrade to the latest version of `git` that ships with `pcre2`.

However, we are able to use `brew reinstall --with-pcre2 git;` on an older version of git, which recompiles `git` with `pcre2`. This approach is slightly faster but will fail on `git >= 2.20`.

Alternatively, we can move all code quality enforcing scripts to a single environment 🤔.

eugenepeh:
>Alternatively, we can move all code quality enforcing scripts to a single environment 🤔.

Yes... if there are no faster substitute-able method, lets do that. No point wasting 5 mins just for brew update. 
Alternatively, we can just leave the issue open and see if Travis will update its mac os with later versions.

yong24s:
@eugenepeh 

Travis latest macOS image `osx_image: xcode10.1` comes with `git 2.19.0`, which has `pcre2`.

Since #556 is downgrading to `JDK8` our macOS environment, we can upgrade to use this latest macOS image. 🎉

.travis.yml
```diff
@@ -6,8 +6,9 @@ matrix:
       os: linux
       jdk: oraclejdk9
 
-    - name: macOS 10.13 JDK 10
+    - name: macOS
```
eugenepeh: can update the name accordingly. It seems to be using Mac OS X 10.13.6 JDK 11.01

yong24s: I want to wait for #556 first, before testing which approach to take. 🤐

yong24s:
@yamidark 

@eugenepeh wants this to go in before #556 👌

Issue #337, opened 2018-09-15
Remove multiple clones when analyzing multiple branch of a repository


eugenepeh:
When we analyzing multiple branches of a repository, RepoSense will clone that same repository as many time as the number of the branches. This is inefficient as all the analysis are performed against the same repo. Hence, there shouldn't be a necessity to clone it multiple time.

Issue #378, opened 2018-10-18
Better implementation of repoRoot


apoorva17:
#332 has updated ```getRepoRoot()``` to use ```repoName``` instead of ```displayName```, which causes an error when testing the functionality of ```git checkout``` with a local disk location, specifically ```checkout_fromDiskLocation_success()``` in ```GitCheckerTest``` 

A better implementation of ```getRepoRoot()``` is required to ensure that this functionality is not affected i.e. ```repoRoot``` can be constructed in the ```RepoConfiguration``` constructor to be kept as a constant value instead of being updated when ```displayName``` is updated.

PR #560, opened 2019-02-14
Perform cloning in parallel with analyzing #540

### LINKED ISSUE #540: Perform cloning in parallel with analyzing
Cloning takes significant amount of time if the target repo is big even with a relatively fast internet. However, it does not takes much processing power, perform cloning in parallel with analyzing shouldn't increase much cpu load. As analyzing contribution can take significantly long and no bandwidth is being utilized in this state, we can speed up the process by doing both in parallel. 

Let's clone the next repo to be analyzed during the analysis to essentially remove cloning from waiting time.

myteo:
Fixes #540.

```
Repos are processed sequentially, one at a time.

As cloning does not takes much processing power and analysis does not
take any network bandwidth, the two can be done in parallel to reduce
the total processing time.

Here are some test results to support the above hypothesis:
The test was performed using CS2103 AY1819S1 project repos on my local
machine. The average time taken to generate the report was measured
across 10 runs. Current code took 17 min 21s while new implementation
took 11 min 55s.

Let's clone the next repo in the list while the current repo is being
analyzed.
```


myteo:
Could you guys take a first look at the general approach and implementation?

I tried measuring the performance gain using CS3281 configs (https://github.com/nus-cs3281/2019-dashboard/tree/master/configs). The current implementation take an average (across 5 runs) of 375 seconds while the new implementation is around 340 seconds. I think not as much gain as expected.

damithc:
> The current implementation take an average (across 5 runs) of 375 seconds while the new implementation is around 340 seconds. I think not as much gain as expected.

You can try this one too https://github.com/nus-cs2103-AY1819S1/cs2103-dashboard

myteo:
Tried to make code less ugly😅 Ready for review.

myteo:
> You can try this one too https://github.com/nus-cs2103-AY1819S1/cs2103-dashboard

Gains are more significant for this. ~17mins vs ~12mins

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -48,28 +48,85 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
+     * Performs analysis of each repository in parallel with the cloning of the next repository.
+     */
+    private static void cloneAndAnalyzeRepos(List<RepoConfiguration> configs, String outputPath) throws IOException {
+        boolean isPreviousRepoCloned = false;
+
+        for (int i = 0; i < configs.size(); i++) {
+            RepoConfiguration config = configs.get(i);
+            boolean isCurrentRepoCloned = spawnCloneProcess(outputPath, config);
+
+            if (isPreviousRepoCloned) {
```
eugenepeh: Seems better, but we can improve it further. 
Let's have another class, say `RepoCloner` to take care of the cloning and cleaning up.

This `RepoCloner` can have hold up to 2 `RepoConfiguration` and always be responsibled for cleaning up the older one. 
So the instance variables will have something like this.
```
    RepoConfiguration[] configs = new RepoConfiguration[2];
    int index = 0;
    int prevIndex = 0;
``` 
So if we do something like `repoCloner.clone(config)` the backend should do something like this:
```
void clone(RepoConfiguration config) {
    prevIndex = index;
    if (config.getLocation().equals(configs[index])) {
        return;
    }
    index = index + 1 % configs.length;
    FileUtil.deleteDirectory(configs[index].getRepoRoot());
    spawnCloneProcess(outputPath, config);
}
```

We can then have `repoCloner` return the previous repo with `repoCloner.getClonedRepo()`
So backend would be something like this:
```
RepoConfiguration getClonedRepo() {
    waitForCloneProcess(configs[prevIndex]);
    Git.checkout(configs.getBranch()); 
    return configs[prevIndex];
}
```

Note that with this new change, we need to remember the default branch of the cloned repo. This is because it can result in error if the order of configs are something like this:|
```
reposense, release
reposense, empty -> default branch (HEAD)
```
This will result in the second configuration not analyzing the correct branch.

fzdy1914:
This PR also fixes #337, correct?

myteo:
> This PR also fixes #337, correct?

Thanks for pointing it out. Currently in this PR, cloning will be skipped if the previous repo is the same repo. But if the same repos are not next to each other in the list then they will still be cloned twice.

I think will be better to fix that in a separate PR.

eugenepeh:
btw @myteo , do use the request for reviewer on the top right
image

or include us in your comment as well on top of changing the PR status when requesting for review in case we didn't notice.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -78,14 +98,16 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
             AuthorshipSummary authorshipSummary = AuthorshipReporter.generateAuthorshipSummary(config);
             generateIndividualRepoReport(commitSummary, authorshipSummary, repoReportDirectory.toString());
 
-            try {
-                FileUtil.deleteDirectory(FileUtil.REPOS_ADDRESS);
-            } catch (IOException ioe) {
-                logger.log(Level.WARNING, "Error deleting report directory.", ioe);
+            if (shouldDeleteDirectory) {
+                FileUtil.deleteDirectory(config.getRepoRoot());
```
eugenepeh: as mentioned earlier, clean up can be on the cloner side.
It is easier for the cloner to determine whether the directory can be deleted.
E.g. if the array slot is going to be used to clone new repo, old one can be deleted. 
when all the analyzing is done, can call for clean up too.

src/main/java/reposense/system/CommandRunnerProcess.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * Waits for the previously spawned process to finish executing before returning the exit value.
+     * Should only be called after {@code spawnProcess} has been called.
+     */
+    public int waitForProcess() throws InterruptedException {
+        int exit = process.waitFor();
+        outputGobbler.join();
+        errorGobbler.join();
```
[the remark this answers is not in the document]
myteo: I maintained the same logic as the current code. Not exactly sure what I should cleanup also

src/main/java/reposense/git/GitClone.java
```diff
@@ -41,22 +45,86 @@ public static void clone(RepoConfiguration repoConfig)
...
     }
 
-    private static void clone(RepoLocation location, String repoName) throws IOException {
-        Path rootPath = Paths.get(FileUtil.REPOS_ADDRESS, repoName);
-        Files.createDirectories(rootPath);
-        runCommand(rootPath, "git clone " + addQuote(location.toString()));
+    /**
+     * Checks out the branch specified in {@code repoConfig}.
+     */
+    public static void checkoutBranch(RepoConfiguration repoConfig) throws GitCloneException {
```
eugenepeh: This should go under `GitCheckout`

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -48,44 +48,69 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
-                logger.log(Level.SEVERE, "Error has occurred during analysis, will skip this repo.", rte);
-                continue;
-            }
+        cloneAndAnalyzeRepos(configs, outputPath);
 
-            // preprocess the config and repo
-            updateRepoConfig(config);
-            updateAuthorList(config);
+        try {
+            FileUtil.deleteDirectory(FileUtil.REPOS_ADDRESS);
```
eugenepeh: instead of cleaning up here, let's make `cleanupDirectories()` a public method and call it at the end of `cloneAndAnalyzeRepos`.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -0,0 +1,117 @@
...
+    private void generateEmptyRepoReport(String outputPath, RepoConfiguration config) throws IOException {
+        Path repoReportDirectory = Paths.get(outputPath, config.getDisplayName());
+        FileUtil.createDirectory(repoReportDirectory);
+        ReportGenerator.generateEmptyRepoReport(repoReportDirectory.toString());
+    }
+
+    /**
+     * Deletes cloned repo directories that are not in use anymore.
+     */
+    private void cleanupDirectories() {
```
eugenepeh: why is this called `cleanupDirectories()` when it only cleans up 1 repo?

src/main/java/reposense/git/GitClone.java
myteo: This is actually a "wrapper" around `GitCheckout.checkout()`, to catch `RuntimeException` and throw `GitCloneException`. Might be weird if I shift this and `GitCheckout` throws `GitCloneException`.

But I do agree its weird to have a function called `checkoutBranch` under `GitClone`. Will think about how to improve it.

src/main/java/reposense/report/ReportGenerator.java
myteo: Here I'm deleting the entire `repos` folder, but `cleaupDirectories()` is meant for cleaning up and deleting the folder containing the previous repo which has been analyzed.

src/main/java/reposense/report/RepoCloner.java
myteo: Good point. I originally used plural because it also deletes the folder of repos that failed to clone, but I think thats not a good use of the function.

src/main/java/reposense/git/GitClone.java
```diff
@@ -59,4 +60,65 @@ private static void clone(RepoLocation location, String repoName) throws IOExcep
...
+            throw new GitCloneException(e);
+        }
+        crp = null;
+        updateRepoConfigBranch(repoConfig);
+    }
+
+    /**
+     * Updates the branch of {@code repoConfig} to the current branch if default branch is specified.
+     */
+    public static void updateRepoConfigBranch(RepoConfiguration repoConfig, String currentBranch) {
```
eugenepeh: This method should go to `RepoConfiguration`.
It shouldn't be here in first place since it is being called from `RepoCloner`.

src/main/java/reposense/git/GitClone.java
```diff
@@ -59,4 +60,65 @@ private static void clone(RepoLocation location, String repoName) throws IOExcep
...
+        if (repoConfig.getBranch().equals(RepoConfiguration.DEFAULT_BRANCH)) {
+            repoConfig.setBranch(currentBranch);
+        }
+    }
+
+    private static void updateRepoConfigBranch(RepoConfiguration repoConfig) throws GitCloneException {
```
eugenepeh: Perhaps make this a member method of `RepoConfiguration` too.

src/main/java/reposense/report/ReportGenerator.java
eugenepeh: As in, we should have a clean up method afterward to ensure all the folders are deleted.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -48,44 +46,61 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
+        } catch (IOException ioe) {
+            logger.log(Level.WARNING, "Error deleting report directory.", ioe);
+        }
+        FileUtil.writeJsonFile(new SummaryReportJson(configs, generationDate), getSummaryResultPath(outputPath));
+        logger.info("The report is generated at " + outputPath);
+    }
+
+    /**
+     * Clones and analyzes repositories in {@code configs}.
+     * Performs analysis of each repository in parallel with the cloning of the next repository.
```
eugenepeh: Seems to have missed out the report generation details in the javadocs.

src/main/java/reposense/report/ReportGenerator.java
myteo: Not sure I understand. All repos are cloned into the `repos` folder, so deleting the `repos` folder would ensure all the folders are deleted.

src/main/java/reposense/git/GitClone.java
```diff
@@ -59,4 +63,53 @@ private static void clone(RepoLocation location, String repoName) throws IOExcep
...
+            logger.info("Cloning of " + repoConfig.getLocation() + " completed!");
+        } catch (RuntimeException | CommandRunnerProcessException e) {
+            crp = null;
+            logger.log(Level.SEVERE, "Error encountered in Git Cloning, will attempt to continue analyzing", e);
+            throw new GitCloneException(e);
+        }
+        crp = null;
+
+        try {
+            if (repoConfig.getBranch().equals(RepoConfiguration.DEFAULT_BRANCH)) {
```
eugenepeh: perhaps extract a method for these lines and place it as a member method in RepoConfiguration.

```
        try {
            if (repoConfig.getBranch().equals(RepoConfiguration.DEFAULT_BRANCH)) {
                String currentBranch = GitBranch.getCurrentBranch(repoConfig.getRepoRoot());
                repoConfig.setBranch(currentBranch);
            }
            logger.log(Level.SEVERE, "Branch does not exist! Analysis terminated.", e);
            throw new GitCloneException(e);
        }
```

Can call it `renewBranch` or `fetchBranch` etc.

myteo: Ok makes sense. It might be weird for `RepoConfiguration` to throw `GitCloneException`, so I'll handle the exception at a higher level.

myteo:
> also you've missed out my comments earlier [#560 (comment)](https://github.com/reposense/RepoSense/pull/560#discussion_r261147627)

Yeah about that comment, I am not sure I understand it correctly. You might also have missed my reply haha (https://github.com/reposense/RepoSense/pull/560#discussion_r261180561).

src/main/java/reposense/git/GitClone.java
eugenepeh: You can create its own exception, e.g. `BranchNotFoundException`. The respective function that calls it can catch it and decide what to do afterward. e.g. catch `BranchNotFoundException` and throw `
GitCloneException` etc

eugenepeh:
>Yeah about that comment, I am not sure I understand it correctly. You might also have missed my reply haha ([#560 (comment)](https://github.com/reposense/RepoSense/pull/560#discussion_r261180561)).

Oops, yeah I missed that. 

What I meant is that we can package that chuck of code into `RepoCloner` to make it responsible for cleaning up to make the responsibilities are more clearly defined and separated.

```
try {
            FileUtil.deleteDirectory(FileUtil.REPOS_ADDRESS);
...
```

myteo:
Ready for review. @reposense/devs

src/main/java/reposense/git/GitClone.java
```diff
@@ -59,4 +60,50 @@ private static void clone(RepoLocation location, String repoName) throws IOExcep
...
+        } catch (RuntimeException | CommandRunnerProcessException e) {
+            crp = null;
+            logger.log(Level.SEVERE, "Error encountered in Git Cloning, will attempt to continue analyzing", e);
+            throw new GitCloneException(e);
+        }
+        crp = null;
+
+        try {
+            repoConfig.updateBranch();
+        } catch (RuntimeException | BranchNotFoundException e) {
```
eugenepeh: Can remove the `RuntimeException` if it won't occur anymore

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -48,44 +46,57 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
+                analyzeRepo(outputPath, clonedRepo);
             }
+            clonedRepo = repoCloner.getClonedRepo(outputPath);
         }
-        FileUtil.writeJsonFile(new SummaryReportJson(configs, generationDate), getSummaryResultPath(outputPath));
-        logger.info("The report is generated at " + outputPath);
+        if (clonedRepo != null) {
+            analyzeRepo(outputPath, clonedRepo);
+        }
+        repoCloner.deleteDirectory(FileUtil.REPOS_ADDRESS);
```
eugenepeh: Let's encapsulate this in a method call `cleanup` in RepoCloner to improve the abstraction and save the explanation here on why this line here is necessary. Can return `deleteDirectory` back to a private method afterward.

src/main/java/reposense/git/GitClone.java
```diff
@@ -59,4 +60,50 @@ private static void clone(RepoLocation location, String repoName) throws IOExcep
         Files.createDirectories(rootPath);
         runCommand(rootPath, "git clone " + addQuote(location.toString()));
     }
+
+    /**
+     * Spawns a process to clone repo specified in {@code repoConfig}. Does not wait for process to finish executing.
+     * Can only handle a maximum of one spawned process at any time and subsequent calls are ignored if current process
+     * is still running.
+     */
+    public static void spawnCloneProcess(RepoConfiguration repoConfig) throws GitCloneException {
```
eugenepeh: Let's shift this 2 methods to `RepoCloner` as well, since its only utilized by it. That way we can save `crp` from being a static object as well. 

By the way, if only a maximum of one spawned process should be called at any point of time, we should use Assertion here to prevent multiple calls.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -79,32 +85,54 @@ private boolean isPreviousRepoDifferent() {
         return prevIndex == index || !configs[prevIndex].getLocation().equals(configs[index].getLocation());
     }
 
+    /**
+     * Spawns a process to clone repo specified in {@code repoConfig}. Does not wait for process to finish executing.
+     * Can only handle a maximum of one spawned process at any time and subsequent calls are ignored if current process
```
eugenepeh: This line can just simplify to `Should only handle a maximum of one spawned process at any time.`

myteo:
@reposense/devs In case you missed this.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -0,0 +1,159 @@
...
+import reposense.model.RepoConfiguration;
+import reposense.system.CommandRunnerProcess;
+import reposense.system.CommandRunnerProcessException;
+import reposense.system.LogsManager;
+import reposense.util.FileUtil;
+
+public class RepoCloner {
+    private static final Logger logger = LogsManager.getLogger(RepoCloner.class);
+
+    private RepoConfiguration[] configs = new RepoConfiguration[2];
```
yong24s: Can make `2` a constant for easier maintenance.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -0,0 +1,159 @@
...
+     */
+    public RepoConfiguration getClonedRepo(String outputPath) throws IOException {
+        if (isCurrentRepoCloned && isPreviousRepoDifferent()) {
+            isCurrentRepoCloned = waitForCloneProcess(outputPath, configs[index]);
+        }
+
+        if (!isCurrentRepoCloned) {
+            deleteDirectory(configs[index].getRepoRoot());
+            return null;
+        } else if (isPreviousRepoDifferent()) {
```
yong24s: Since the above is returning early.

This can be

```
 <--- Line break
if (isPreviousRepoDifferent()) {
```

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -48,44 +46,57 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
+            if (clonedRepo != null) {
+                analyzeRepo(outputPath, clonedRepo);
             }
+            clonedRepo = repoCloner.getClonedRepo(outputPath);
         }
-        FileUtil.writeJsonFile(new SummaryReportJson(configs, generationDate), getSummaryResultPath(outputPath));
-        logger.info("The report is generated at " + outputPath);
+        if (clonedRepo != null) {
+            analyzeRepo(outputPath, clonedRepo);
+        }
```
yong24s: Why is this block repeated?

```
if (clonedRepo != null) {
    analyzeRepo(outputPath, clonedRepo);
}
```
Is it possible to restructure the control flow?

myteo: In the for loop, the current repo is cloned and the previous repo is analyzed. We loop for n times where n is length of configs, thats why the last repo is analyzed outside the for loop.

One way to avoid the repeated block is to loop n+1 times:
```
for (int i = 0; i <= configs.size(); i++) {
            if (i != configs.size())
                repoCloner.clone(outputPath, configs[i]);

            if (clonedRepo != null) {
                analyzeRepo(outputPath, clonedRepo);
            }
            if (i != configs.size())
                clonedRepo = repoCloner.getClonedRepo(outputPath);
        }
```

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -0,0 +1,162 @@
...
+            config.updateBranch();
+        } catch (BranchNotFoundException e) {
+            logger.log(Level.SEVERE, "Branch does not exist! Analysis terminated.", e);
+            generateEmptyRepoReport(outputPath, config);
+            return false;
+        }
+        return true;
+    }
+
+    private void generateEmptyRepoReport(String outputPath, RepoConfiguration config) throws IOException {
```
yamidark: Hmmm, we don't really need to know that we need to generate an empty report within `RepoCloner` if we fail to clone, perhaps rename this method to `handleCloningFailed`?

src/main/java/reposense/system/CommandRunnerProcessException.java
```diff
@@ -0,0 +1,8 @@
+package reposense.system;
+
+
+public class CommandRunnerProcessException extends Exception {
```
yamidark: Likewise, javadocs to give a brief description of this exception.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -0,0 +1,162 @@
...
+import reposense.git.GitBranch;
+import reposense.git.GitCheckout;
+import reposense.git.exception.BranchNotFoundException;
+import reposense.model.RepoConfiguration;
+import reposense.system.CommandRunnerProcess;
+import reposense.system.CommandRunnerProcessException;
+import reposense.system.LogsManager;
+import reposense.util.FileUtil;
+
+public class RepoCloner {
```
yamidark: Javadocs description of this class

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -151,6 +173,15 @@ public String getRepoRoot() {
         return path;
     }
 
+    public String getRepoFolderName() {
```
yamidark: Sorry, can explain why this is needed?

myteo: Sorry should have probably explained somewhere.

Currently, we use `repoName` of `RepoLocation` as the name of the folder that repos are cloned into. However, 2 different repos can have the same `repoName`, for example all CS2103 projects will have "main" as their  `repoName`. So when we clone and analyze two repos, the second one will override the first one.

`getRepoFolderName()` uses `organization` of `RepoLocation` as well to ensure that the folder is unique across different repos.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -141,8 +143,28 @@ public void update(StandaloneConfig standaloneConfig) {
         ignoreCommitList = CommitHash.convertStringsToCommits(standaloneConfig.getIgnoreCommitList());
     }
 
+    /**
+     * Updates branch if default branch is specified.
+     */
+    public void updateBranch(String defaultBranch) {
+        if (branch.equals(RepoConfiguration.DEFAULT_BRANCH)) {
+            setBranch(defaultBranch);
```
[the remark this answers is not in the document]
eugenepeh: I think we needed branch to possess a non-empty value when use in conjunction with parameters. Not sure if that is still the case but take note when amending this field.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -0,0 +1,162 @@
...
+     */
+    private boolean isPreviousRepoDifferent() {
+        return prevIndex == index || !configs[prevIndex].getLocation().equals(configs[index].getLocation());
+    }
+
+    /**
+     * Spawns a process to clone repo specified in {@code repoConfig}. Does not wait for process to finish executing.
+     * Should only handle a maximum of one spawned process at any time.
+     */
+    private boolean spawnCloneProcess(String outputPath, RepoConfiguration config) throws IOException {
```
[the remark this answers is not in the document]
eugenepeh: The reason I requested him to make it a member method because he need to keep track of the `crp` in order to check its status. Moving it to elsewhere would delegate that duty / add unnecessary overhead, let's just keep it within RepoCloner since it is its main responsibility.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * Deletes previously cloned repo directories that are not in use anymore.
+     */
+    private void cleanupPrevRepoFolder() {
+        if (isPreviousRepoDifferent() && prevIndex != index) {
+            deleteDirectory(configs[prevIndex].getRepoRoot());
+        }
+    }
+
+    private void deleteDirectory(String root) {
```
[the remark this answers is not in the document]
myteo: This method calls the `FileUtil` method, it just surrounds it with a try-catch.

src/main/java/reposense/model/RepoConfiguration.java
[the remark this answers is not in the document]
myteo: I think I will change `defaultBranch` to `currentBranch` to make it less confusing. One of the tests fails when I change `DEFAULT_BRANCH` to `EMPTY_BRANCH`.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -0,0 +1,162 @@
...
+            logger.log(Level.WARNING,
+                    "Exception met while trying to clone the repo, will skip this repo.", e);
+            generateEmptyRepoReport(outputPath, config);
+            return false;
+        }
+        crp = null;
+
+        try {
+            config.updateBranch();
+        } catch (BranchNotFoundException e) {
```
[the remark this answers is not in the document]
myteo: You are right, removed this.

myteo:
I removed `BranchNotFoundException` because I realized `GitBranch#getCurrentBranch` shouldn't throw an exception because its just running `git branch`. It was originally in a try-catch so I thought it throws exception. My bad..

src/main/java/reposense/model/RepoConfiguration.java
yamidark: I see.

Anyway, to prevent the same issue as #378, let's just create a new `final` variable `repoFolderName`, which is set to this value within the constructor.

src/main/java/reposense/report/RepoCloner.java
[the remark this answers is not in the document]
yamidark: Oh... whoops

src/main/java/reposense/model/RepoConfiguration.java
myteo: Done.

damithc:
Suggested changes to commit message:
```
When analyzing the current repository, cloning of the next repository
is not done in parallel.
```
This sounds like reverse engineering the answer to form the question. Suggested:
```
Repos are processed sequentially, one at a time.
```
The rest:
```
As cloning does not takes much processing power and analysis
does not take any network bandwidth, the two can be done in
parallel to reduce the total processing time.

Here are some test results to support the above hypothesis:
{add your test results here}

Let's clone the next repo in the list while the current repo is
being analyzed.
```

myteo:

PR #565, opened 2019-02-21
[#503] CLI: add --version tag to show version of RepoSense

### LINKED ISSUE #503: Add --version tag to show the version of the jar of RepoSense
As, title. The users may not know the jar file they download is the correct version of RepoSensse or not and they cannot verify it.

Let's add a -version tag to show the version of jar file of RepoSense.

fzdy1914:
Fixes #503 

Users may want to figure out whether the jar file they download is
the correct version of RepoSense they want to use.

The users cannot verify it because we do not have a 'version' tag 
and the generated report do not contain such information.

Let's add a -version tag to show the version of jar file of RepoSense and
include it in the generated report.

The version can be passed in to RepoSense using `-Pversion` for both run
and shadowJar. Otherwise, latest git tag will be used.


fzdy1914:
I have added the function code with `--version` tag and the report.
image
However, for the alias for `--version`, there is already a tag `--view` have the same tag `-v`. Please kindly give suggestion about the alias of it.

yong24s:
> However, for the alias for `--version`, there is already a tag `--view` have the same tag `-v`. Please kindly give suggestion about the alias of it.

`-V`? In general, use uppercase. 😂

yong24s:
Comments:

Not sure how the `Version` class would be useful. 🤔 While you have comparators, there's no interface to use it. Also, keeping it simple, comparing of versions should be done at script level.

src/main/java/reposense/RepoSense.java
```diff
@@ -102,4 +104,9 @@ public static void main(String[] args) {
 
         return configs;
     }
+
+    private static String getVersion() {
+        String version = RepoSense.class.getPackage().getImplementationVersion();
+        return  version == null ? "v1.7" : version;
```
yong24s: Not sure if it will be null. It seems to be `Unspecified` when run without `-Pversion`.

yong24s: Instead of using `v1.7` is it possible to leave it as an empty string?

fzdy1914:
> When run as Jar, there are single quotes around version number `i.e.` `Generated by RepoSense 'v1.7'`, which breaks link to GitHub.

The single quotes to pass the version is not necessary, i.e:
`./gradlew shadowJar -Pversion=v1.7` is enough.

> It will be complete if we can also push version for `gradlew run ...`. 🤔 Not sure if it is easy tho.

Maybe System property can help, but not sure about it.

src/main/java/reposense/RepoSense.java
fzdy1914: And also broke the link.

fzdy1914: Maybe we can ask other's opinion of `default version` ?

fzdy1914:
@emer7 @ongspxm @chelseyong Please kindly give your review.

fzdy1914:
It seems that no stage1-reviewer wants to review this pr, I will just request review from stage2-reviewer.

src/main/java/reposense/RepoSense.java
```diff
@@ -27,6 +27,9 @@
 import reposense.util.FileUtil;
 
 public class RepoSense {
+    public static final String VERSION = getVersion();
```
emer7: How about getting the version with the static method instead of static variable? I feel this is quite weird, unless there is reason behind it.

frontend/src/index.jade
```diff
@@ -31,8 +31,10 @@ html
           )
           .timestamp-footer
             span Generated by &nbsp;
-            a(href="https://github.com/reposense/RepoSense")
-              strong RepoSense
+            a(
```
emer7: Does this follow the style guide?

emer7:
> @emer7 @ongspxm @chelseyong Please kindly give your review.

Oops I didn't saw this

frontend/src/index.jade
fzdy1914: The eslint in CI does not give warning, so I assume it is correct.

yamidark:
@fzdy1914 Do resolve the conflicts, and request review from us again

src/main/java/reposense/RepoSense.java
```diff
@@ -28,6 +28,8 @@
 import reposense.util.FileUtil;
 
 public class RepoSense {
+    public static final String DEFAULT_VERSION = "v1.8rc";
```
yong24s: Use a generic value so we don't have to maintain this.

The only time the value will be used is when the Jar is built without `-Pversion`.

fzdy1914: Can you suggest one?

fzdy1914: Or just empty string?

yong24s:
@fzdy1914 

Can consider Eugene's [comment](https://github.com/reposense/RepoSense/issues/503#issuecomment-466638399)? 

This way, we can remove the need to specify version with `-Pversion`. 💯 🎉

src/main/java/reposense/RepoSense.java
yong24s: Empty string is good.

Having a version may be confusing on Netlify previews.

fzdy1914:
> @fzdy1914
> 
> Can consider Eugene's [comment](https://github.com/reposense/RepoSense/issues/503#issuecomment-466638399)?
> 
> This way, we can remove the need to specify version with `-Pversion`. 💯 🎉

Then, we still need an argument to decide whether to use the tag or hash right?

eugenepeh:
>Or just print the version and the hash all together?

Since its meant for the release, let's stick with just version for now.

yong24s:
@eugenepeh 

By default, shadowJar will set the `version` to the latest commit hash.
If `-Pversion` is provided, then use that value instead.

This should be easier for all of us right? 🤔

eugenepeh:
>By default, shadowJar/run tasks will set the version to the latest commit hash.
If -Pversion is provided, then use that value instead.

Yes, that is fine too. As long as the version tag is there e.g. `v1.8` by default, it is fine.

fzdy1914:
> Yes, that is fine too. As long as the version tag is there e.g. `v1.8` by default, it is fine.

Do we need to print the version number on the report?
Because if we have this feature, we have to pass this information to the JSON. Then, our system test resources will contain this version information also. Then, for each new tag, the system test json needed to be changed as well.

eugenepeh:
>Do we need to print the version number on the report?
Because if we have this feature, we have to pass this information to the JSON. Then, our system test resources will contain this version information also. Then, for each new tag, the system test json needed to be changed as well.

Wouldn't that be the same if we have the commit hash as well? If that is a problem, we can choose to output to separate json file or have the program modify the output html file etc.

fzdy1914:
@eugenepeh This plugin seems not to work well with netlify deploy. It works locally. But it shows v1.0 in the preview anyway, which is quite weird.

src/main/java/reposense/RepoSense.java
```diff
@@ -105,4 +106,14 @@ public static void main(String[] args) {
 
         return configs;
     }
+
+    public static String getVersion() {
+        String version = RepoSense.class.getPackage().getImplementationVersion();
+
+        if (version == null || "unspecified".equals(version)) {
```
yamidark: Abstract out "unspecified" as a constant variable.
Also, since we already did a `version == null` check before, change the order of `equals` -> `version.equals("unspecified")` instead (sounds more logically correct)

src/main/java/reposense/RepoSense.java
```diff
@@ -28,6 +28,7 @@
 import reposense.util.FileUtil;
 
 public class RepoSense {
+
```
yamidark: Unnecessary change

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -63,9 +66,14 @@ private static ArgumentParser getArgumentParser() {
 
         // Boolean flags
         parser.addArgument(HELP_FLAGS)
-                .dest(HELP_FLAGS[0])
```
yamidark: Why this removed?

src/main/java/reposense/report/SummaryReportJson.java
```diff
@@ -8,11 +8,13 @@
  * Represents the structure of summary.json file in reposense-report folder.
  */
 public class SummaryReportJson {
+    private final String repoSenseVersion;
     private final String reportGeneratedTime;
     private final List<RepoConfiguration> repos;
 
-    public SummaryReportJson(List<RepoConfiguration> repos, String reportGeneratedTime) {
+    public SummaryReportJson(List<RepoConfiguration> repos, String reportGeneratedTime, String version) {
```
yamidark: Keep the variables same name, `repoSenseVersion`

frontend/src/static/js/main.js
```diff
@@ -175,6 +175,10 @@ window.app = new window.Vue({
     generateKey(dataObj) {
       return JSON.stringify(dataObj);
     },
+
+    getRepoSenseLink() {
+      return `https://github.com/reposense/RepoSense/commits/${window.app.repoSenseVersion}`;
```
eugenepeh: instead of commits, can we point to `tree`?

src/main/java/reposense/model/ConfigCliArguments.java
```diff
@@ -10,7 +10,7 @@
 import reposense.parser.RepoConfigCsvParser;
 
 /**
- * Represents command line arguments user supplied when running the program with mandatory field -config.
+ * Represents command line arguments user supplied when running the program with mandatory field --config.
```
eugenepeh: since this PR doesn't modify these files, let's not fix these in this PR.

src/main/java/reposense/parser/ArgsParser.java
fzdy1914: Because this is not necessary, the help argument will only execute the action and nothing will be pasted through this flag.

frontend/src/static/js/main.js
fzdy1914: I have edited it, but pass by hash feature will break this link.

src/main/java/reposense/RepoSense.java
```diff
@@ -31,6 +31,7 @@
     private static final Logger logger = LogsManager.getLogger(RepoSense.class);
     private static final int SERVER_PORT_NUMBER = 9000;
     private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("E MMM d HH:mm:ss 'SGT' yyyy");
+    private static String VERSION_UNSPECIFIED = "unspecified";
```
eugenepeh: u have forgotten to use `final`

fzdy1914:
> Do raise this issue to the library owner, maybe he/she can provide us some insight / help on this.

I have raised the issue. One workaround is to specify the version in the build script of Netlify.

eugenepeh:
>I have raised the issue. One workaround is to specify the version in the build script of Netlify. fzdy1914#1

Yes that is fine. Do link the issue to this PR so that we can be kept updated.

fzdy1914:
image
I suspect that the `depth` may be the source of the problem.

fzdy1914:
> image
> I suspect that the `depth` may be the source of the problem.

I have tried and found that this may cause the information with tag lost.

eugenepeh:
>I have tried and found that this may cause the information with tag lost.

@fzdy1914 I have reconfigured the netlify to perform `git fetch --unshallow`. Try to retrigger a build to see if it works.

fzdy1914:
@eugenepeh It seems that it is not the reason for the bug.

`fatal: --unshallow on a complete repository does not make sense`

eugenepeh:
>@eugenepeh It seems that it is not the reason for the bug.

I see. Well, at least we eliminated 1 possibility.

fzdy1914:
> > @eugenepeh It seems that it is not the reason for the bug.
> 
> I see. Well, at least we eliminated 1 possibility.

By the inspiring of your try, I have found out how to fix it. It seems that it just doesn't fetch the tag.
image
image

eugenepeh:
>By the inspiring of your try, I have found out how to fix it. It seems that it just doesn't fetch the tag.

Great work, I have updated the build command.

fzdy1914:
Weird, same pr and same command give different results.

My [deploy](https://app.netlify.com/sites/reposense-fzdy/deploys/5c98dd1a3f137c00089d82e4)

Reposense [deploy](https://app.netlify.com/sites/reposense/deploys/5c98dd1978a9f7000899a7f0)

eugenepeh:
I have noticed that the image of the 2 build are different. I have changed the configuration to use the latest available image.

image

fzdy1914:
@eugenepeh The preview works fine now.

build.gradle
```diff
@@ -92,6 +99,15 @@ test {
 shadowJar {
     archiveName = 'RepoSense.jar'
     destinationDir = file("${buildDir}/jar/")
+
+    def repoSenseVersion = project.property("version")
+    if (repoSenseVersion.equals("unspecified")) {
+        repoSenseVersion = versionDetails().lastTag
+    }
```
yong24s:
```
    if (repoSenseVersion.equals("unspecified")) {
        if (versionDetails().commitDistance == 0) {
            repoSenseVersion = versionDetails().lastTag
        }
        else {
            repoSenseVersion = versionDetails().gitHash
        }
    }
```
This way the version can be tag or hash. If the `HEAD` is pointing tag, then tag will be used; else the hash will be used.

eugenepeh: if its a simple change, i think no harm adding that

fzdy1914: > if its a simple change, i think no harm adding that

This will result that all the netlify preview will have the hash shown in the bottom of report. If it is acceptable, I will start editing the code.

eugenepeh: Yes that is acceptable to me, thanks @fzdy1914

fzdy1914: Edited as request

build.gradle
```diff
@@ -50,7 +51,17 @@ wrapper {
 
 run {
     //the second arguments indicates the default value associated with the property.
+    def repoSenseVersion = project.property("version")
+    if (repoSenseVersion.equals("unspecified")) {
+        if (versionDetails().commitDistance == 0) {
+            repoSenseVersion = versionDetails().lastTag
+        } else {
+            repoSenseVersion = versionDetails().gitHash
+        }
```
damithc: Can use a `.. ? .. : ..` ?

build.gradle
```diff
@@ -92,6 +103,19 @@ test {
...
     destinationDir = file("${buildDir}/jar/")
+
+    def repoSenseVersion = project.property("version")
+    if (repoSenseVersion.equals("unspecified")) {
+        if (versionDetails().commitDistance == 0) {
+            repoSenseVersion = versionDetails().lastTag
+        } else {
+            repoSenseVersion = versionDetails().gitHash
+        }
+    }
```
damithc: This code fragment seems to be duplicated. Any way to avoid that duplication?

yamidark:
@fzdy1914 Do resolve the conflicts, as well as investigate why during cypress test, it is unable to load the files (stuck at loading page)

fzdy1914:
@yamidark The bug has been identified. The reason for this bug is that the version information is not passed into the `startServerInBackground` task. Also, to ensure that RepoSense will not break even the version information is missed, the default link is added.

fzdy1914:
@yamidark I have resolved the conflict.

PR #566, opened 2019-02-21
Refactor toolbar into v-authorship tab

ongspxm:
might affect #524 and #544 

```
Currently the toolbar for the "expand/collapse all" text is outside of
the v-authorship tab. This shouldn't be the case as it makes it
difficult to extend and include other kinds of tabs.

Let's shift the toolbar into the v-authorship tab, and make some
corresponding changes to how the toggle state is changed.

This will keep the toolbar within the authorship tab, which makes sense
as the button is only relevant for the authorship tab. In light that we
are going to have other kinds of tabs to extend the use of the side tab
in the future.
```


ongspxm:
@apoorva17 could you check if there is any test case i missed?

ongspxm:
@apoorva17 could you check if there is any test case i missed?

frontend/src/index.jade
```diff
@@ -166,6 +163,9 @@ html
 
     vuetemplate#v_authorship
       #authorship
+        .toolbar(v-if="info.totalCommits > 0")
```
chelseyong: This forces `totalCommits` to be present in `info`. However in #524, `totalCommits` has to be derived from repo commits, as it cannot be inside the url.

ongspxm: means i take from `window.REPOS` instead of passing it in?

chelseyong: Yes, which I have implemented in #524.

ongspxm: i saw the implementation, a func to open up `window.REPOS`. the implementation is gud.

but i think i shouldn't pull that change here. because that would require me to pull the function from the javascript file as well, there might be more conflicts.

i put it as it is, but add a comment, marking it to be removed in during merging with #524?

ongspxm:
> The `Expand/Collapse All` tab does not seem to reflect the changes

@chelseyong  which one is this? i tested everything seems fine

chelseyong:
Same as #521.

ongspxm:
> Same as #521.

right, good catch

ongspxm:
@chelseyong check pls

chelseyong:
Still doesn't work. Play around with the checkboxes and try.

ongspxm:
> Still doesn't work. Play around with the checkboxes and try.

let me confirm the error
- load tab authorship (text: collapse all)
- click "collapse all" (text: expand all)
- uncheck "all" (text: expand all) by default
- check "all" (text: collapse all)

this is state right?

frontend/src/static/js/v_authorship.js
```diff
@@ -171,9 +169,11 @@ window.vAuthorship = {
       if (!this.isSelectAllChecked) {
         this.selectedFileTypes = this.fileTypes;
         this.selectedFiles = this.files;
+        this.activeFiles = this.files.length;
```
chelseyong: Change `activeFiles` to `activeFileCount`?

frontend/src/index.jade
```diff
@@ -166,6 +163,10 @@ html
 
     vuetemplate#v_authorship
       #authorship
+        // TODO: remove in #524
+        .toolbar(v-if="info.totalCommits > 0")
+            a(v-if="activeFiles==0", v-on:click="expandAll(true)") Expand all
```
chelseyong: Use `===` instead

ongspxm:
> What's the difference btw isTabActive and tabActive ?

isTabActive is like a flag that triggers the tab pane re-rendering

tabActive is used to tell the tab pane which kind of panes to load. Here only got authhorship, we gonna have zoomin and others later on

eugenepeh:
>tabActive is used to tell the tab pane which kind of panes to load. Here only got authhorship, we gonna have zoomin and others later on

Perhaps give a less confusing name? E.g. `tabType`, `activeTabType` etc.
If there isn't any `inactiveTabs`, I dont think there's a need to include active in its name.

frontend/src/static/js/main.js
```diff
@@ -99,10 +99,11 @@ window.app = new window.Vue({
 
     isLoading: false,
     isCollapsed: false,
+
+    // isTabActive used to force tab wrapper to load
     isTabActive: true,
```
eugenepeh: perhaps inline the comments? E.g. `isTabActive: true,  // to force tab wrapper to load`

frontend/src/static/js/v_authorship.js
```diff
@@ -73,6 +54,21 @@ window.vAuthorship = {
...
+      const files = document.getElementsByClassName('file');
+      Array.from(files).forEach((file) => {
+        file.className = renameValue;
+      });
+
+      this.activeFileCount = isActive ? this.selectedFiles.length : 0;
+    },
+
+    updateCount() {
+      this.activeFileCount = document.getElementsByClassName('file active').length;
```
eugenepeh: activeFile`s`Count

frontend/src/index.jade
```diff
@@ -37,17 +37,14 @@ html
 
         #tabs-wrapper(v-if="isTabActive")
           #tab-resize(onmousedown="registerMouseMove()")
-          .tab-close(v-on:click="isTabActive=false")
+          .tab-close(v-on:click="isTabActive=false;")
```
eugenepeh: this change doesn't seems to match this PR purpose.

frontend/src/index.jade
```diff
@@ -41,13 +41,10 @@ html
...
             .tab-panes
-              #tab-authorship.tab-pane(v-if="isTabAuthorship")
-                .toolbar(v-if="tabInfo.tabAuthorship.totalCommits > 0")
-                    a(v-if="isCollapsed", v-on:click="expand(true)") Expand all
-                    a(v-else, v-on:click="expand(false)") Collapse all
-                v_authorship(
-                  v-bind:key="generateKey(tabInfo.tabAuthorship)",
-                  v-bind:info="tabInfo.tabAuthorship")
+              v_authorship#tab-authorship.tab-pane(
+                v-if="tabType=='authorship'"
```
eugenepeh: space around operator, also, why not follow existing style to use `===`?

frontend/src/index.jade
```diff
@@ -166,6 +163,10 @@ html
 
     vuetemplate#v_authorship
       #authorship
+        // TODO: remove in #524
+        .toolbar(v-if="info.totalCommits > 0")
+            a(v-if="activeFileCount===0", v-on:click="expandAll(true)") Expand all
```
eugenepeh: space around operator ===

frontend/src/static/js/main.js
```diff
@@ -99,10 +99,10 @@ window.app = new window.Vue({
 
     isLoading: false,
     isCollapsed: false,
-    isTabActive: true,
-    isTabAuthorship: false,
+
+    isTabActive: true, // to force tab wrapper to load
```
eugenepeh: can leave the booleans grouped together

ongspxm:
@yamidark Merge?

yamidark:
@ongspxm We just merged in #487, seems like there is a small conflict, can help to resolve it?

frontend/src/index.jade
```diff
@@ -199,6 +196,10 @@ html
 
     vuetemplate#v_authorship
       #authorship
+        // TODO: remove in #524
+        .toolbar(v-if="info.totalCommits > 0")
+            a(v-if="activeFileCount === 0", v-on:click="expandAll(true)") Expand all
```
yamidark: Untitled

Seems like this `activeFileCount` variable hasn't been refactored to latest master?

ongspxm:
sha? let me merge again

PR #568, opened 2019-02-22
FAQ: add another entry to propose email as alternative to GitHub ID #546

### LINKED ISSUE #546: FAQ: add another entry to propose email as alternative to GitHub ID
as per title

yamidark:
Fixes #546 
```
In the UserGuide FAQ section, we mention that commit contribution
may not appear in the ramp chart if the user did not associate the
correct author name aliases to their GitHub ID, and to double check
their author name aliases.

After #493 though, RepoSense allows user to associate their GitHub
email to their GitHub ID as well, which is more accurate as it is a
fixed value, compared to author name aliases which can have many
different values depending on their local git settings and on different
machines. As such, it is better to ask users to use email instead of
author name aliases if they face such problems of their contribution
not being tracked by RepoSense.

Let's update FAQ to propose email as an alternative to associate to the
user's GitHub ID instead.

The command given in the `A Note About Git Author Name` is not
exactly accurate, as it only changes the git author name settings for
future initiated or cloned git repositories, but not for the current
git repository, which is what most users who are reading the FAQ would
want.

Let's also update the `A Note About Git Author Name` section to
include the correct command for changing the git author name settings
for the current git repository as well.
```

### DIFF
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -187,13 +187,17 @@ Date:   Fri Feb 9 19:13:13 2018 +0800
  ...
 ```
 `ActualGitHubId` and `ConfiguredAuthorName` are both `Git Author Name` of the same author.<br>
-To find the author name that you are currently using, run the following command at your analyzed git repository:
+To find the author name that you are currently using for your current git repository, run the following command within your git repository:
 ```
 git config user.name
 ```
-To set the author name to the value you want (e.g., to set it to your GitHub username) for the analyzed git repository, you can use the following command ([more info](https://www.git-tower.com/learn/git/faq/change-author-name-email)):
+To set the author name to the value you want (e.g., to set it to your GitHub username) for the current git repository, you can use the following command ([more info](https://www.git-tower.com/learn/git/faq/change-author-name-email)):
 ```
 git config user.name "YOUR_AUTHOR_NAME”
+```
+To set the author name to the value you want for the future git repository, you can use the following command:
+```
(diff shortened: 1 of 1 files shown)


docs/UserGuide.md
```diff
@@ -335,7 +335,8 @@ The ignore glob list can be specified through the [standalone config file](#prov
 #### Q: My commit contributions does not appear in the ramp chart (despite appearing in the contribution bar and code panel)?
 **A:** This is probably a case of giving an incorrect author name alias (or github ID) in your [author-config file](#author-configcsv).<br>
 Please refer to [A Note About Git Author Name](#a-note-about-git-author-name) above on how to find out the correct author name you are using, and how to change it.<br>
-Also ensure that you have added all author name aliases that you may be using (if you are using multiple computers or have previously changed your author name).
+Also ensure that you have added all author name aliases that you may be using (if you are using multiple computers or have previously changed your author name).<br>
```
eugenepeh: Perhaps add the command `git config user.name` as well which helps them find the author name?

yamidark: >Perhaps add the command `git config user.name` as well which helps them find the author name?

I have already referenced to the `A Note About Git Author Name` section of the UserGuide at the FAQ, but seems like it only mentions changing of the author name and not just for finding it. Well add to that section how to find the author name.

yamidark: @eugenepeh I have also updated the current command for setting of git author names for current git repository, as I tested that adding the `--global` flag may not affect the settings for the current git repository.

PR #570, opened 2019-02-24
RampChart: add group by authors option

### LINKED ISSUE #543: Implement grouping of ramp charts by authors
Current: Main page allows grouping of the ramp charts by repository.
Suggested: Group ramp charts by authors, so that it is easier to check on a particular user's progress.

chelseyong:
Fixes #543.
```
Users are only allowed to group the ramp charts by repository.

While grouping by repository allows users to have a rough comparison of
contributions between authors of the same repository, it does not
provide any insight into an author's contribution data across all
repositories he has contributed to.

Let's add the grouping of authors as an option for users to select, and
this can work towards the goal of having a toolbar[1] at the top of the
report.

[1]: Sorting options graphical goal:
https://github.com/orgs/reposense/teams/devs/discussions/1
```


frontend/src/static/js/v_summary.js
```diff
@@ -362,33 +361,88 @@ window.vSummary = {
...
-      });
-
-      if (!this.filterGroupRepos) {
-        full[0].sort(comparator((ele) => {
-          const field = ele[this.filterSort];
-          return field.toLowerCase ? field.toLowerCase() : field;
-        }));
+      let full = [];
+      if (this.filterGroupSelection === 'groupByNone') {
+        full[0] = this.filterByNone(this.filtered);
```
ongspxm: use `groupedByXXX` instead of `filterByXXX`

frontend/src/static/js/v_summary.js
```diff
@@ -362,33 +361,88 @@ window.vSummary = {
...
+      });
+      sortedRepos.sort(comparator((ele) => {
+        const field = ele[this.filterSort];
+        return field.toLowerCase ? field.toLowerCase() : field;
+      }));
+      return sortedRepos;
+    },
+    filterByAuthors(repos) {
+      const authorMap = {};
+      const authorData = {};
```
ongspxm: if you never use `authorData` until later, put it tgt below, where you actually use it. makes code easier to trace

frontend/src/static/js/v_summary.js
```diff
@@ -362,33 +361,88 @@ window.vSummary = {
...
+        });
+      });
+      // If filterSort is related to variance & contribution, sum them up
+      if (repos[0] && !Number.isNaN(repos[0][0][this.filterSort])) {
+        Object.keys(authorMap).forEach((author) => {
+          authorData[author] = {};
+          authorData[author][this.filterSort] = [0, 0];
+          authorMap[author].forEach((repo) => {
+            authorData[author][this.filterSort][0] += 1;
+            authorData[author][this.filterSort][1] += repo[this.filterSort];
```
ongspxm: what does this `[0,0]` do? is it total commits VS num of repo? can add a comment? `// [numRepos, numCommits]`

chelseyong:
Hmm I see, but sorting of the ramp charts is done in the select-menu, where the user can sort by `author name`, `contribution` etc. So changing this to `sort by` may be confusing.
@damithc your opinion on this?

damithc:
> Hmm I see, but sorting of the ramp charts is done in the select-menu, where the user can sort by `author name`, `contribution` etc. So changing this to `sort by` may be confusing.

I think he meant that we can use a drop-down instead of radio buttons, to follow the style of `sort by`.

damithc:
At one point we mentioned about aiming for something like this. So, it does make sense to make it a drop-down.
image

fzdy1914:
> I think he meant that we can use a drop-down instead of radio buttons, to follow the style of `sort by`.

@chelseyong Sorry for not explaining clearly. Using a drop-down is what I mean.

eugenepeh:
Looks good 👍 
However, I think need to discuss with @damithc (if you haven't) about the desired sorting order (e.g. contribution, variance etc) on how it should be presented when its `grouped by author`.

When it is `grouped by repo`, the order of groups doesn't change unlike `grouped by author`. We should come up with a fixed standard for this case. If the order of groups are meant to adhere to the sorting criteria, we should do the same when its `grouped by repo` too.

damithc:
> However, I think need to discuss with @damithc (if you haven't) about the desired sorting order (e.g. contribution, variance etc) on how it should be presented when its `grouped by author`.

I'm hoping to reach something like the below eventually. In the intermediate, we can do whatever is convenient. For example, we can leave the sort order 'undefined' when grouped by author (or use a fixed order e.g., alphabetically).

image

chelseyong:
@eugenepeh so leave the repos `grouped by author` sorted alphabetically by default? Should the other sorting parameters be allowed when `grouped by author`?

eugenepeh:
> @eugenepeh so leave the repos grouped by author sorted alphabetically by default? Should the other sorting parameters be allowed when grouped by author?

Yes, just leave it as default for now. The sorting of groups should be tackled in another PR.

frontend/src/index.jade
```diff
@@ -100,19 +106,16 @@ html
...
+            .summary-charts__title(v-if="filterGroupSelection === 'groupByRepos'") {{ repo[0].repoPath }}
+            .summary-charts__title(v-if="filterGroupSelection === 'groupByAuthors'") {{ repo[0].displayName }}
             .summary-chart(v-for="(user, i) in repo")
               .summary-chart__title
                 .summary-chart__title--index {{ i+1 }}
-                .summary-chart__title--repo(v-if="!filterGroupRepos") {{ user.repoPath }}
-                .summary-chart__title--name {{ user.displayName }}
+                .summary-chart__title--repo(v-if="filterGroupSelection === 'groupByNone'") {{ user.repoPath }}
+                .summary-chart__title--author-repo(v-if="filterGroupSelection === 'groupByAuthors'") {{ user.repoPath }}
+                .summary-chart__title--name(v-if="filterGroupSelection !== 'groupByAuthors'") {{ user.displayName }}
```
yamidark: Shouldn't this be `filterGroupSelection === 'groupByRepo'`?

chelseyong: The user's display name has to be shown when grouped `by none` and `by repo`.

frontend/src/index.jade
```diff
@@ -71,12 +71,15 @@ html
...
+                option(value="groupByRepos") Repo/Branch
+                option(value="groupByAuthors") Author
+              label group by
             .mui-select
               select(v-model="filterSort")
-                option(value="totalCommits") Contribution
-                option(value="variance") Variance
-                option(value="displayName") Author Name
-                option(value="searchPath") Repo/Branch Name
+                option(v-for="option in filterSortOptions" v-bind:value="option.value") {{ option.text }}
```
eugenepeh: just wondering, is it not possible to use `v-if` in such scenario?

chelseyong: Yes, that is possible..

chelseyong:
Hmm.. interesting indeed.
For these 2 repositories, despite both having the same `username`, they have different `display name`.
When the order is reversed, since the code takes the first repository's `display name`, the author's name will now be different.

However, the situation that 2 similar users have different display name but same username is quite rare.
One possible alternative is to display the username beside the author (in brackets) to reduce user's confusion. Eg: `Eugene (eugenepeh)`

chelseyong:
codacy kept failing.. The logic seems simple though.

yamidark:
>codacy kept failing.. The logic for updateSortSelection seems simple though.

If you are unable to resolve this Codacy issue, we can just ignore it for now, not much of an issue.

eugenepeh:
>However, the situation that 2 similar users have different display name but same username is quite rare.
One possible alternative is to display the username beside the author (in brackets) to reduce user's confusion. Eg: Eugene (eugenepeh)

Yes, that's a good way to resolve this problem.

chelseyong:
@eugenepeh thanks! Can this be merged soon to proceed with #601?

frontend/src/index.jade
```diff
@@ -127,8 +129,9 @@ html
             .summary-chart(v-for="(user, i) in repo")
               .summary-chart__title
                 .summary-chart__title--index {{ i+1 }}
-                .summary-chart__title--repo(v-if="!filterGroupRepos") {{ user.repoPath }}
-                .summary-chart__title--name {{ user.displayName }}
+                .summary-chart__title--repo(v-if="filterGroupSelection === 'groupByNone'") {{ user.repoPath }}
+                .summary-chart__title--author-repo(v-if="filterGroupSelection === 'groupByAuthors'") {{ user.repoPath }}
```
ongspxm: is this just a difference in a styling? cos i dun see any stylesheet changes?

why need to have two `{{ user.repoPath }}`?

frontend/src/static/js/v_summary.js
```diff
@@ -430,26 +433,21 @@ window.vSummary = {
...
-      if (!this.filterGroupRepos) {
-        full[0].sort(window.comparator((ele) => {
-          const field = ele[this.filterSort];
-          return field.toLowerCase ? field.toLowerCase() : field;
-        }));
+    },
+    sortFiltered() {
+      let full = [];
+      if (this.filterGroupSelection === 'groupByNone') {
+        full[0] = this.groupByNone(this.filtered);
```
ongspxm: comment that this is making everything "under one group", not apparent on first glance

frontend/src/index.jade
chelseyong: `.summary-chart__title--repo` is bolded by default

yamidark:
Nits to proposed message:

~~Currently,~~ **U**sers are **only** allowed to group the ramp charts by repository.

**While grouping by repository allows users to have a rough comparison of contributions
between authors of the same repository, it does not provide any**
~~They are not able to group them by authors, which can provide greater~~ insight
into an author's contribution data, across all repositories **he has contributed to**. 

Let's add the grouping of authors as an option for users to select, and this can
work towards the goal of having a toolbar[1] at the top of the dashboard.

[1]: ~~Sorting toolbar for reference~~**Sorting options graphical goal**:
https://github.com/orgs/reposense/teams/devs/discussions/1

PR #574, opened 2019-02-24
refactoring the way we handle sidebar opening events

ongspxm:
requires #566 to be merged in

```
The way `v_summary` handles the loading of the sidebar
assumes there is only one type of sidebar view.

In light of the new "zoomin" view, we would have to extend and
generalize the way we load the sidebar, so as to be able to load both
the authorship and zoomin view using the same mechanism.

The design of the "event bus" is as such. The actual loading of the
tabs is done in the main app, and those functions usually have the form
`updateTabXXX()`, which write the corresponding information object into
`this.tabInfo` and does `this.tabActive='XXX';`. This will then load
the right tab in the sidebar.

To load the sidebar from without another component, the VueJS event
emitting mechanism is used. A function of the form  `openTabXXX()` is
used to call the component's `$emit()`. The main application then
handles the emitted message using the corresponding `updateTabXXX()`.
```


yamidark:
@ongspxm Is this PR ready?

yamidark:
@ongspxm Can't seem to open the codeview now at all though, it only opens the sidebar without showing the codeview?

eugenepeh:
>@ongspxm Can't seem to open the codeview now at all though, it only opens the sidebar without showing the codeview?

Same

ongspxm:
its some merge error, got it fixed :+1:

PR #575, opened 2019-02-26
WIP: Implementing the zoomin tab

ongspxm:
to clean up after #574 merged in

- [x] `update summary view` -> `show ramp chart for this period`?
- [x] On a related note, we should provide an easy way to bring the ramp chart back to the full period. A button in the ramp chart toolbar?
- [x] The ui seems to interpret it as I am trying to drag and drop ui elements.
- [x] Looks great, however there are some instances where I can't drag select the date range.
- [x] Reset drag areas when resized
- [x] Higher contrast for highlighted areas
- [x] The lines are a little bit too thick.
- [x] I tried to netlify preview (Chrome). I can draw a box in the ramp chart but nothing happens after that.


ongspxm:
@damithc @eugenepeh 

first draft of UI
image

damithc:
Nice. 
* `update summary view` -> `show ramp chart for this period`?
* On a related note, we should provide an easy way to bring the ramp chart back to the full period. A button in the ramp chart toolbar?

eugenepeh:
Looks great, however there are some instances where I can't drag select the date range.
e.g. when the ramp chart is too compact. 
The ui seems to interpret it as I am trying to drag and drop ui elements.

ongspxm:
- [x] `update summary view` -> `show ramp chart for this period`?
- [x] On a related note, we should provide an easy way to bring the ramp chart back to the full period. A button in the ramp chart toolbar?
- [x] Looks great, however there are some instances where I can't drag select the date range.

@eugenepeh fixed it on my end, see if that works for you, if it doesn't tell me the browser version

yamidark:
Tried on netlify, seems like still have the same problem as @eugenepeh mentioned, sometimes I am unable to drag and select the range.

When I click the 'Show ramp chart for this period` multiple times, it seems to keep restricting the date range repeatedly, probably because the same region is still sensed to be highlighted but the date range has become smaller, eventually date range becomes a single date. Can probably make this function only work once for a single highlighted area?

Also, can we make the highlighted area have a darker/more contrast colour so it is clearer for us to see which area is zoomed in, or have lines to indicate the left start and right end range?

ongspxm:
> Can probably make this function only work once for a single highlighted area?
right, gd catch, i will reset the draggings

ongspxm:
- [ ] Looks great, however there are some instances where I can't drag select the date range.
- [ ] Reset drag areas when resized
- [ ] Higher contrast for highlighted areas

yamidark:
> @yamidark means you will start dragging the triangles instead?

Nope, I just tried dragging from an 'empty' area within the ramp chart, but as I tried to drag the cursor changed to a black 🚫  symbol instead, and fails to drag anymore (after release does nothing, have to redrag)

yamidark:
> > @yamidark means you will start dragging the triangles instead?
> 
> Nope, I just tried dragging from an 'empty' area within the ramp chart, but as I tried to drag the cursor changed to a black 🚫 symbol instead, and fails to drag anymore (after release does nothing, have to redrag)

@ongspxm May be a problem with Google Chrome, does seem to get this problem in Firefox Browser.

ongspxm:
Yeah, i tested on chrome, it seems to be a chrome only problem, it seems like the mouseup event is not fired. 

I couldnt find the cause of it tho :sad:

ongspxm:
@eugenepeh ah, if the feature requires the user to double click before dragging, then how ah? do we add the text on the UI interface itself? or write in the docs

eugenepeh:
>@eugenepeh ah, if the feature requires the user to double click before dragging, then how ah? do we add the text on the UI interface itself? or write in the docs

Docs is probably better. Any better ideas other than double click?

ongspxm:
> Docs is probably better. Any better ideas other than double click?

its more of a "if you happen to get stuck, double clicking will solve your problem" thing

eugenepeh:
>its more of a "if you happen to get stuck, double clicking will solve your problem" thing

I see, so the original drag click will still work?

ongspxm:
yep, the original drag will still work

ongspxm:
yep, it should work, i merged master and everything alr

damithc:
I tried to netlify preview (Chrome). I can draw a box in the ramp chart but nothing happens after that.

eugenepeh:
>I tried to netlify preview (Chrome). I can draw a box in the ramp chart but nothing happens after that.

same

ongspxm:
> I tried to netlify preview (Chrome). I can draw a box in the ramp chart but nothing happens after that.

fixed

damithc:
Preview seems to work now 👍

yong24s:
Doesn't seem to work well with ramps near the edge.

The "zoom-in" becomes empty when the selection box is beyond or at the edge of the rampchart? 🤔 

image

https://deploy-preview-575--reposense.netlify.com/#search=&sort=displayName&since=2017-10-09&until=2019-03-25&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=false&tabAuthor=April0616&tabRepo=reposense_testrepo-Beta_master

yong24s:
> Any better ideas other than double click?

Hold `Shift` + `Left Click` to insert the a "dotted marker"? 

Reason: Its easier to use an additional modifier key to easily distinguish between normal clicks to the ramp or intention to insert a "dotted marker". Also dragging may be buggy or difficult (imprecise unless a mouse is used).

ongspxm:
shift click implemented, check pls

chelseyong:
Is there a way to exit the selected drag-box? Like pressing `Esc`.
Since the drag-box can only be selected when pressing shift, perhaps the click pointer can remain the same? i.e. not change to the "plus" pointer.

yong24s:
Works well on the same rampchart :+1:.

Things to consider:
1. **Left click should clear any dotted markers**
Reason: To avoid buggy situations, say you place a dotted marker then resize code view then insert another :boom: 
2. **Clear any preceding dotted markers if its not on the same rampchart**
Reason: To avoid buggy situations, say you place a dotted marker on a rampchart then insert another on another rampchart... *not sure feature or bug* :thinking:

eugenepeh:
>Since the drag-box can only be selected when pressing shift, perhaps the click pointer can remain the same? i.e. not change to the "plus" pointer.

Yes if it only selected when pressing shift, then the click pointer should remain normal. Else, retain the left click drag.

ongspxm:
> Else, retain the left click drag.

currently, i removed the left click drag, do you still want left click drag?

eugenepeh:
>currently, i removed the left click drag, do you still want left click drag?

Either is fine, but the `+` cursor shouldn't be there when its not a left click drag.

ongspxm:
all done. ready for review

eugenepeh:
image
Seems to have some bugs 🤔

ongspxm:
Everything good to go

ongspxm:
Everything good to go

PR #582, opened 2019-03-04
README.md: add new team members

fzdy1914:
Three new team members have been newly joined RepoSense from Jan 2019.

Let's add their information in README.md.


eugenepeh:
@myteo @apoorva17 can both of you provide @fzdy1914 your full names to be added as well?

PR #583, opened 2019-03-05
Split repo csv configurations from RepoConfiguration #548

### LINKED ISSUE #548: Decouple AuthorConfiguration from RepoConfiguration
Continuation of issue #457.

```
RepoConfiguration represents all the configurations of a repository and
AuthorConfiguration is contained in RepoConfiguration class.

This makes it harder to maintain as any changes to one of the
configuration file will affect the other.

Let's create a new class ReposenseConfiguration which contains
RepoConfiguration and AuthorConfiguration, with RepoConfiguration
representing configs from repo-config.csv and AuthorConfiguration
representing configs from author-config.csv.
```

myteo:
Fixes #548 

```
RepoConfiguration represents all configuration information for a single
repository, for both the case of command line parameters and CSV config
file. In the case of CSV config file, configuration from 
repo-config.csv is also represented by RepoConfiguration.

This makes it harder to maintain as RepoConfiguration violates the
Single Responsibility Principle.

Let's create a new class RepoCsvConfiguration to represent the
repo config file settings, splitting it from the RepoConfiguration.
```

### DIFF
--- a/src/main/java/reposense/RepoSense.java
+++ b/src/main/java/reposense/RepoSense.java
@@ -14,6 +14,7 @@ import reposense.model.CliArguments;
 import reposense.model.ConfigCliArguments;
 import reposense.model.LocationsCliArguments;
 import reposense.model.RepoConfiguration;
+import reposense.model.RepoCsvConfiguration;
 import reposense.model.RepoLocation;
 import reposense.model.ViewCliArguments;
 import reposense.parser.ArgsParser;
@@ -71,8 +72,11 @@ public class RepoSense {
      * @throws IOException if user-supplied csv file does not exists or is not readable.
      */
     public static List<RepoConfiguration> getRepoConfigurations(ConfigCliArguments cliArguments) throws IOException {
-        List<RepoConfiguration> repoConfigs = new RepoConfigCsvParser(cliArguments.getRepoConfigFilePath()).parse();
+        List<RepoCsvConfiguration> repoCsvConfigs =
+                new RepoConfigCsvParser(cliArguments.getRepoConfigFilePath()).parse();
         List<AuthorConfiguration> authorConfigs = null;
+
+        List<RepoConfiguration> repoConfigs = RepoConfiguration.getRepoConfigurationList(repoCsvConfigs);
 
         try {
             authorConfigs = new AuthorConfigCsvParser(cliArguments.getAuthorConfigFilePath()).parse();
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -2,7 +2,6 @@ package reposense.model;
 
 import java.io.File;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -13,7 +12,7 @@ import reposense.system.LogsManager;
 import reposense.util.FileUtil;
 
 /**
- * Represents configuration information from CSV config file for a single repository.
+ * Represents configuration information for a single repository.
  */
 public class RepoConfiguration {
     public static final String DEFAULT_BRANCH = "HEAD";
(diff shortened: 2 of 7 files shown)


myteo:
> i.e. the format and since/until date passed from the command line is actually set in `RepoCsvConfiguration`

Yes format is set in `RepoCsvConfiguration` and I agree this part seems not very clean. The since/until date is not set in `RepoCsvConfiguration` though.

> This refactor seems not so meaningful?

The purpose of the PR and issue was to extract configuration from `repo-config.csv` into a separate class but seems like `format` is shared by both `repo-config.csv` and command line `-formats` argument. @yamidark, do you think this PR and issue is needed?

fzdy1914:
> Yes format is set in `RepoCsvConfiguration` and I agree this part seems not very clean. The since/until date is not set in `RepoCsvConfiguration` though.

Sorry, only `formats` and `isStandaloneConfigIgnored` are shared.

Also, this will cause the information of location and branch duplicate in `RepoConfiguration`.

One possible way is to make `RepoCsvConfiguration` as well as `RepoCmdConfiguration` subclasses of `RepoConfiguration`  just like the relation between `LocationsCliArguments` and `ConfigCliArguments`.

But it is only my suggestion, let's ask the mentor's opinion.

yamidark:
>@yamidark, do you think this PR and issue is needed?

Sorry, missed this.

Anyway, I think we can still continue with this PR, just that we can separate the 'repo'-level configurations into `RepoConfiguration`, instead of it being a pure 'representation' of `repo-config.csv`. ( `AuthorConfiguration` also contains some 'repo-level' and 'cli-level' configurations passed over). 

Then we can perhaps rename the original `RepoConfiguration` to `RepoSenseConfiguration` or `ReportConfiguration` instead, that contains both (and the 'cli' configurations that are the same throughout all repos like `sinceDate` and `untilDate`)? 

@yong24s @eugenepeh Your opinion on this?

fzdy1914:
> Then we can perhaps rename the original `RepoConfiguration` to `RepoSenseConfiguration` or `ReportConfiguration` instead, that contains both (and the 'cli' configurations that are the same throughout all repos like `sinceDate` and `untilDate`)?

Maybe not `ReportConfiguration`. We are going to allow our users to customize some of details of the generated report including title, icon, heading, etc. The config of this may be more suitable to have this name.

eugenepeh:
>@yong24s @eugenepeh Your opinion on this?

I am okay with this, it can be useful in the future if we need to define any form of hierarchical settings.
But I can't really recall why was there a need for Author and Repo config to be of equal standing, can remind me on that please?

yamidark:
>But I can't really recall why was there a need for Author and Repo config to be of equal standing, can remind me on that please?

Do you mean why both of the same settings are in the same `RepoConfiguration`? I believe this was because when we split the config settings into 2 files, 'author' and 'repo' level, they were still largely similar then, so we kept the same `RepoConfiguration` to store them both for 'merging' later on. 

Then after large number of changes to both of then, it become very messy to do this merging purpose when `RepoConfiguration` is also used to store all the 'author' level settings, so we split them in #499.

eugenepeh:
Hmm in that case, we should create the hierarchy such that
`Global config > repo config > author config`
if there isn't a reason for author config to be of same level as repo config.

yamidark:
> Hmm in that case, we should create the hierarchy such that
> `Global config > repo config > author config`
> if there isn't a reason for author config to be of same level as repo config.

Yea, I believe we wanted the 'cli' arguments (which are global) to override all the other levels, but for some reason (can't remember), we didn't really implement them.

eugenepeh:
>Yea, I believe we wanted the 'cli' arguments (which are global) to override all the other levels, but for some reason (can't remember), we didn't really implement them.

This change may potentially help us simplify that process then.

yamidark:
@myteo 

Do continue this PR based on the suggestions given (remake the original `RepoConfiguration` to `RepoSenseConfiguration` (or any other name you think is suitable), and move 'repo-level' settings to `RepoConfiguration`.

fzdy1914:

PR #586, opened 2019-03-06
edit the address of DTD file used for checkstyle

fzdy1914:
We are currently using the DTD files from `http://checkstyle.sourceforge.net/`
and `http://puppycrawl.com/` in our config file of checkstyle.

However, due to security reason, checkstyle decided to remove DTDs from
above websites and ask users to use the DTD files from `https://checkstyle.org/`[1]

Let's update the address of our location of DTD files correspondingly.

[1] https://github.com/checkstyle/checkstyle/issues/6478

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE module PUBLIC
-    "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
-    "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+    "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+    "https://checkstyle.org/dtds/configuration_1_3.dtd">
 
 <!--
     This configuration file enforces rules for a modified version of the module's code standard at
--- a/config/checkstyle/suppressions.xml
+++ b/config/checkstyle/suppressions.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
 
 <!DOCTYPE suppressions PUBLIC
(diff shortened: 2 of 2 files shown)


emer7:
LGTM

For the commit message, if not mistaken avoid using the word `currently` as it is implicitly so

yamidark:
@fzdy1914 

From  https://checkstyle.org/config_filters.html#SuppressionFilter_Examples, I don't seem to find anything that mentions we should be using version 1.2 though (it just gives an example usage)? Did you give the wrong link, or is there anywhere specific I should look at?

Also, do give a short summary of the link you gave, example in this [commit](https://github.com/reposense/RepoSense/commit/069fa7908112311795a1ad2eb14ddb5819d8bf17).

fzdy1914:
> @fzdy1914
> 
> From https://checkstyle.org/config_filters.html#SuppressionFilter_Examples, I don't seem to find anything that mentions we should be using version 1.2 though (it just gives an example usage)? Did you give the wrong link, or is there anywhere specific I should look at?
> 
> Also, do give a short summary of the link you gave, example in this [commit](https://github.com/reposense/RepoSense/commit/069fa7908112311795a1ad2eb14ddb5819d8bf17).

The example is what I am referring to. Because it is the only place that checkstyle mention about the version of its `suppression.xml`. Since it is an example that gives instructions on how to write this kind of file, so I interpreted it as the suggested version. Maybe you can suggest a way to rephrase it?

yamidark:
>Since it is an example that gives instructions on how to write this kind of file, so I interpreted it as the suggested version. Maybe you can suggest a way to rephrase it?

If this is the latest version available, can just mention that instead.

yamidark:
>Also, do give a short one line summary of any link you gave, example in this [commit](https://github.com/reposense/RepoSense/commit/069fa7908112311795a1ad2eb14ddb5819d8bf17).  

Remember to add this in as well.

Issue #585, opened 2019-03-06
Non-linear ramp chart cause by invalid date range


eugenepeh:
Notice how the last ramps show different dates

![image](https://user-images.githubusercontent.com/19277206/53892322-074ff300-4067-11e9-87c2-8501af7a48bb.png)
![image](https://user-images.githubusercontent.com/19277206/53892339-12a31e80-4067-11e9-8b9a-41d881d976ca.png)

PR #591, opened 2019-03-07
Non-linear ramp chart cause by invalid date range #585

ongspxm:
```
Currently, the ramp chart's filter is updated when the checks have
passed. The checks only accounts for cases where the dates are valid,
checking that they are within range. When dates such as "2019-02-31" is
given, the filter is updated anyways. This causes the ramp charts to be
generated up till the repository's latest commit.

Let's add an additional check to ensure that a valid temporary date is
given before we update the filter fields.
```

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -77,13 +77,13 @@ window.vSummary = {
       this.getFiltered();
     },
     tmpFilterSinceDate() {
-      if (this.tmpFilterSinceDate >= this.minDate) {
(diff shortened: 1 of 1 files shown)


eugenepeh:
Nits to the proposed message:

```
Checks are performed on the dates prior to every update of the ramp
chart's filter.

However, the checks only accounts for cases where the dates are valid,
checking that they are within the given range. When an invalid date
such as "2019-02-31" is given, the checks are bypassed, causing
ramp charts of different repositories to have different end dates and
linearly unsynchronous.

Let's add an additional check to ensure that valid dates were
given before we update the filter fields.
```

PR #592, opened 2019-03-08
GitLog: handle paths in config that exist outside the repo [#398]

### LINKED ISSUE #398: Invalid file path leads to unhandled exception in Git Log
```11:00:19 - Extracting commits info for https://github.com/CS2113-AY1819S1-F10-1/main.git...
Exception in thread "main" java.lang.RuntimeException: Error returned from command git log --no-merges -i  --since="2018-09-12T00:00:00+08:00" --pretty=format:"%H|%aN|%ad|%s" --date=iso --shortstat --author="^meisbokai <.*>$\|^Neow Bo Kai <.*>$\|^meisbokai <.*>$" -- "*.adoc" -- "*.cs" -- "*.css" -- "*.fxml" -- "*.gradle" -- "*.html" -- "*.java" -- "*.js" -- "*.json" -- "*.jsp" -- "*.md" -- "*.py" -- "*.tag" -- "*.xml" ":(exclude)*.class" ":(exclude)addressbook.txt" ":(exclude)backup_data.txt" ":(exclude)exams.txt" ":(exclude)statistics.txt" ":(exclude)*.jar" ":(exclude)*.war" ":(exclude)*.ear" ":(exclude).idea/" ":(exclude)*.iml" ":(exclude)out/" ":(exclude)/bin/" ":(exclude)/data/" ":(exclude)publish.sh" ":(exclude).gradle/" ":(exclude)build/"on path repos\main\main :
fatal: Invalid path '/bin': No such file or directory

        at reposense.system.CommandRunner.runCommand(CommandRunner.java:169)
        at reposense.system.CommandRunner.gitLog(CommandRunner.java:38)
        at reposense.commits.CommitInfoExtractor.extractCommitInfos(CommitInfoExtractor.java:32)
        at reposense.commits.CommitsReporter.generateCommitSummary(CommitsReporter.java:19)
        at reposense.report.ReportGenerator.generateReposReport(ReportGenerator.java:71)
        at reposense.RepoSense.main(RepoSense.java:51)```

### PR DESCRIPTION
**Pull request by chelseyong**
Fixes #398
```
If users try to put folders existing outside of the git repository 
inside the ignore-glob-list, it will lead to a RuntimeException as the
`git log` command fails.

Let's check that the path is inside the repository, before writing it
into the ignore-glob-list and running the `git log` command.
```

### CODE REVIEW

**Comment by fzdy1914** [comment]
CI is failing...

**Comment by chelseyong** [comment]
> CI is failing...

Still fixing

File: src/main/java/reposense/git/GitUtil.java [hunk 1]
```diff
@@ -80,14 +86,51 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
+            File parentFile = canonicalRoot;
+            while (parentFile != null) {
+                if (canonicalChild.equals(parentFile)) {
+                    return true;
+                }
+                parentFile = parentFile.getParentFile();
+            }
+        } catch (IOException ex) {
+            logger.log(Level.WARNING,
+                    "File/Directory in ignore glob list cannot be accessed.", ex);
```
fzdy1914: You are also ignoring these kinds of invalid files. So, another warning should be given when the file is ignored.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -80,14 +86,51 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
+        try {
+            File canonicalRoot = repoRoot.getCanonicalFile();
+            File canonicalChild = child.getCanonicalFile();
+
+            File parentFile = canonicalRoot;
+            while (parentFile != null) {
+                if (canonicalChild.equals(parentFile)) {
+                    return true;
+                }
+                parentFile = parentFile.getParentFile();
```
fzdy1914: The logic here is a little bit weird. Why we are comparing canonicalRoot and its parents with canoicalChild?
Shouldn't we compare canonicalChild and its parents with canonicalRoot?

chelseyong: The code here is trying to trace that the `child` does not exist outside the `root` folder:
image

fzdy1914: Why a child outside the root must be a parent of root?

fzdy1914:
Do note the codacy failure.

src/main/java/reposense/git/GitUtil.java
fzdy1914: For example, the parent root is A/B, the child root is A/C.

Wouldn't it be more straight forward to just check if the child is inside the root?

fzdy1914: i.e. The parent root is A/B/C and the child root is A/B/C/D/E/F, we just check if A/B/C/D/E/F or A/B/C/D/E or A/B/C/D or A/B/C equals the parent root A/B/C?

chelseyong: > Why a child outside the root must be a parent of root?

That's to handle special situations and prevent runtime exception.

chelseyong: > i.e. The parent root is A/B/C and the child root is A/B/C/D/E/F, we just check if A/B/C/D/E/F or A/B/C/D/E or A/B/C/D or A/B/C equals the parent root A/B/C?

The example is not valid for this situation.

fzdy1914: > > Why a child outside the root must be a parent of root?
> 
> That's to handle special situations and prevent runtime exception.

Do you mean that the child passed in the function is either a parent of the root or inside the root? Can you explain why is so?

chelseyong: Yes.
Special situations like if users try to put folders existing outside of the git repository 
inside the ignore-glob-list, e.g. `../bin`, this will lead to `git log` creating a runtime exception.

fzdy1914: I have tested the globs mentioned in https://github.com/reposense/RepoSense/issues/398#issuecomment-470583555 and the problem is not solved. i.e. add `../nonExisting` will still result in runtime error.

fzdy1914: The problem is solved after I change the code to below:
```
    static boolean pathExistsInRepo(File repoRoot, String path) {
        String validPath = path;
        if (path.contains("/*")) {
            validPath = path.substring(0, path.indexOf(("/*")));
        } else if (path.contains("*")) { // not in directories
            return true;
        }
        return childIsInsideRepo(repoRoot, new File(repoRoot, validPath));
    }

    static boolean childIsInsideRepo(File repoRoot, File child) {
        try {
            File rootFile = repoRoot.getCanonicalFile();
            File childFile = child.getCanonicalFile();

            while (childFile != null) {
                if (childFile.equals(rootFile)) {
                    return true;
                }
                childFile = childFile.getParentFile();
            }
        } catch (IOException ex) {
            logger.log(Level.WARNING, "File system queries are needed.", ex);
            return false;
        } catch (SecurityException ex) {
            logger.log(Level.WARNING, "File cannot be accessed.", ex);
            return false;
        }
        return false;
    }
```

fzdy1914: > Yes.
> Special situations like if users try to put folders existing outside of the git repository
> inside the ignore-glob-list, e.g. `../bin`, this will lead to `git log` creating a runtime exception.

`../bin` will go the folder of with name `bin` in the same parent with the root but itself is not the parent of root.

You should try to add ignore glob list in author-csv to test.

fzdy1914: image
This is the repo csv I am using.

chelseyong: I get what you mean now. 
Didn't check specifically for `../bin` in the test case, but `../*java` passes the test.

fzdy1914: You may also add a `.exist()` to test whether the child exists. But not so sure if it is necessary. (Don't know how git glob works)

src/main/java/reposense/git/GitUtil.java
fzdy1914: > You are also ignoring these kinds of invalid files. So, another warning should be given when the file is ignored.

What I mean here is that you should give the user warning that there is an invalid file in ignore-glob but not just ignoring it.

chelseyong:
> Do note the codacy failure.

Not sure whether to standardize the coding style by not putting the access modifier for the function:
https://github.com/reposense/RepoSense/blob/038f645dd8bb0940009bee122f053ae74b89d0a5/src/main/java/reposense/git/GitUtil.java#L120

Or put access modifier to fix codacy.

The frontend tests are also failing, though it was not touched here.

@reposense/devs pls advise :)

fzdy1914:
> > Do note the codacy failure.
> 
> Not sure whether to standardize the coding style by not putting the access modifier for the function:
> [RepoSense/src/main/java/reposense/git/GitUtil.java](https://github.com/reposense/RepoSense/blob/038f645dd8bb0940009bee122f053ae74b89d0a5/src/main/java/reposense/git/GitUtil.java#L120)
> 
> Line 120 in [038f645](/reposense/RepoSense/commit/038f645dd8bb0940009bee122f053ae74b89d0a5)
> 
>  static boolean childIsInsideRepo(File repoRoot, File child) { 
> Or put access modifier to fix codacy.
> 
> The frontend tests are also failing, though it was not touched here.
> 
> @reposense/devs pls advise :)

From my perspective, putting modifier to the static method is fine.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -80,14 +86,54 @@ static String convertToGitFormatsArgs(List<Format> formats) {
     /**
      * Returns the {@code String} command to specify the globs to exclude for `git log` command.
      */
-    static String convertToGitExcludeGlobArgs(List<String> ignoreGlobList) {
+    public static String convertToGitExcludeGlobArgs(File root, List<String> ignoreGlobList) {
         StringBuilder gitExcludeGlobArgsBuilder = new StringBuilder();
         final String cmdFormat = " " + addQuote(":(exclude)%s");
         ignoreGlobList.stream()
-                .filter(item -> !item.isEmpty())
+                .filter(item -> !item.isEmpty() && isValidPath(root, item))
```
fzdy1914: It seems that my explanation in this https://github.com/reposense/RepoSense/pull/592#discussion_r264026681 is not clear enough.

I mean that this filter will automatically exclude the invalid path which may not throw exceptions. So, when ignoring these kinds of invalid paths, it is better to give a log message to warn the user that, there is an invalid path in the ignoreGlobList and this path is skipped. Having this message, the user will realize that there may be a mistake in the csv file.

eugenepeh:
```
Or put access modifier to fix codacy.

The frontend tests are also failing, though it was not touched here.

@reposense/devs pls advise :)
```
Yes, put the modifier.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -80,14 +86,59 @@ static String convertToGitFormatsArgs(List<Format> formats) {
     /**
      * Returns the {@code String} command to specify the globs to exclude for `git log` command.
```
eugenepeh: perhaps explain the use of `root` in the javadocs?

src/main/java/reposense/git/GitUtil.java
```diff
@@ -80,14 +86,59 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
-                .filter(item -> !item.isEmpty())
+                .filter(item -> !item.isEmpty() && isValidPath(root, item))
                 .map(ignoreGlob -> String.format(cmdFormat, ignoreGlob))
                 .forEach(gitExcludeGlobArgsBuilder::append);
 
         return gitExcludeGlobArgsBuilder.toString();
     }
+
+    /**
+     * Returns true if the {@code String} path is inside the current repository
```
eugenepeh: same here, should include in javadocs that logging will be performed.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -80,14 +86,59 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
+        if (!isChildInsideRepo) {
+            logger.log(Level.WARNING, path + " will be skipped as this file is outside the repo.");
+        }
+        return isChildInsideRepo;
+    }
+
+    /**
+     * Returns true if the child path is inside repository folder
+     * @throws IOException if file system queries are needed
+     * @throws SecurityException if file permission rights are needed
```
eugenepeh: this method already handled the error, so nothing is being thrown.

src/main/java/reposense/system/CommandRunner.java
```diff
@@ -50,7 +50,7 @@ public static String runCommand(Path path, String command) {
             return outputGobbler.getValue();
         } else {
             String errorMessage = "Error returned from command ";
-            errorMessage += command + "on path ";
+            errorMessage += command + " on path ";
```
eugenepeh: doesnt seems to fit the purpose of this PR

src/main/java/reposense/git/GitUtil.java
```diff
@@ -79,15 +85,60 @@ static String convertToGitFormatsArgs(List<Format> formats) {
 
     /**
      * Returns the {@code String} command to specify the globs to exclude for `git log` command.
+     * Also checks that every file in {@code String} file is within the given repository {@code File} root
```
eugenepeh: When citing the parameters, can just use the parameter name e.g. `{@code ignoreGlobList}`.
The `{@code String}` in the above header was used to described the return value as there was only a given type.

src/main/java/reposense/git/GitUtil.java
eugenepeh: Same for here `{@code String}`. Missed that out earlier.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -79,15 +85,60 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
         return gitExcludeGlobArgsBuilder.toString();
     }
+
+    /**
+     * Returns true if the {@code path} is inside the current repository
+     * Produces log messages when the invalid file path is skipped
+     */
+    public static boolean isValidPath(File repoRoot, String path) {
+        String validPath = path;
+        if (path.startsWith("/") || path.startsWith("\\")) {
```
eugenepeh: instead of doing this, is it possible to use PathMatcher?
https://stackoverflow.com/questions/24677662/match-path-string-using-glob-in-java

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -77,6 +77,47 @@ public void gitLog_authorIgnoreAllJavaFiles_success() {
...
+        content = GitLog.get(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberExpectedCommitsToGitLogLines(8, content));
+
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("collate.md"));
+        content = GitLog.get(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberExpectedCommitsToGitLogLines(8, content));
+
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("./java"));
+        content = GitLog.get(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberExpectedCommitsToGitLogLines(8, content));
```
eugenepeh: Doesn't seems quite right to have all the expected result to be the same when its working.
In order word, this doesn't seems to be the best way to test it.

Could you create a new testutil to sum and compared the number of files changed instead?

src/test/java/reposense/util/TestUtil.java
```diff
@@ -141,6 +142,29 @@ public static boolean compareNumberExpectedCommitsToGitLogLines(int expectedNumb
...
+    /**
+     * Returns true if the {@code expectedNumberFilesChanged} is equal to the actual number of files changed in
+     * {@code gitLogResult}
+     */
+    public static boolean compareNumberFilesChanged(int expectedNumberFilesChanged, String gitLogResult) {
+        // if git log result is empty, then there are no files changed
+        if (gitLogResult.isEmpty()) {
+            return expectedNumberFilesChanged == 0;
+        }
+        // 2nd line of each commit has info of the changed file
```
yamidark: This comment, and what the segment below doing doesn't seem to the same thing?

src/main/java/reposense/git/GitUtil.java
```diff
@@ -79,15 +89,51 @@ static String convertToGitFormatsArgs(List<Format> formats) {
 
     /**
      * Returns the {@code String} command to specify the globs to exclude for `git log` command.
+     * Also checks that every file in {@code ignoreGlobList} is within the given repository {@code root}
```
eugenepeh: I think `checks that every glob in {@code ignoreGlobList} only targets files within the given repository {@code root}` is more correct?

src/main/java/reposense/git/GitUtil.java
```diff
@@ -79,15 +89,51 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
      */
-    static String convertToGitExcludeGlobArgs(List<String> ignoreGlobList) {
+    public static String convertToGitExcludeGlobArgs(File root, List<String> ignoreGlobList) {
```
eugenepeh: need a few unit test on this method.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -79,15 +89,51 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
                 .forEach(gitExcludeGlobArgsBuilder::append);
 
         return gitExcludeGlobArgsBuilder.toString();
     }
+
+    /**
+     * Returns true if the {@code path} is inside the current repository
+     * Produces log messages when the invalid file path is skipped
+     */
+    public static boolean isValidPath(File repoRoot, String path) {
```
eugenepeh: seems like this method is only used internally, can set to private in that case.

src/main/java/reposense/git/GitUtil.java
chelseyong: Is it appropriate to put these unit tests in `GitLogTest`?

src/main/java/reposense/git/GitLog.java
```diff
@@ -10,10 +10,12 @@
...
 /**
  * Contains git log related functionalities.
- * Git log is responsible to obtain the commit logs and the authors' info.
  */
 public class GitLog {
 
+    /**
+     * Responsible for obtaining the commit logs and the authors' info.
+     */
     public static String get(RepoConfiguration config, Author author) {
```
yamidark: The Javadocs for methods, should be more of explaining what the method is doing, rather than just saying what it is responsible for. Like in this case, it should be saying something like:
`Returns the git commit log info of {@code Author}, in the repository specified in {@code config}`

src/test/java/reposense/util/TestUtil.java
```diff
@@ -141,6 +142,29 @@ public static boolean compareNumberExpectedCommitsToGitLogLines(int expectedNumb
...
+    public static boolean compareNumberFilesChanged(int expectedNumberFilesChanged, String gitLogResult) {
+        // if git log result is empty, then there are no files changed
+        if (gitLogResult.isEmpty()) {
+            return expectedNumberFilesChanged == 0;
+        }
+        String[] changesLogged = gitLogResult.split("\n");
+        HashSet<String> filesChanged = new HashSet<>();
+        // Checks the 2nd line of each commit which contains info of the changed file
+        for (int i = 1; i < changesLogged.length; i += 4) {
+            String log = changesLogged[i];
```
yamidark: Can we start `int i = 0`, and get the `String log = changesLogged[i + 1]`, easier for us to know that we are getting the 2nd line in each iteration.

src/main/java/reposense/git/GitLog.java
```diff
@@ -10,10 +10,12 @@
 
 /**
  * Contains git log related functionalities.
- * Git log is responsible to obtain the commit logs and the authors' info.
```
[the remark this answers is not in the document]
chelseyong: Originally is because this line seems to belong to the `get` method instead, now that there are 2 methods in this class, it may seem inappropriate to put it there. 
Will put it back.

src/main/java/reposense/git/GitUtil.java
eugenepeh: Uh no, when the method under test is from the GitUtil, it would be best to put it under GitUtilTest.

chelseyong: Noted. Have made the changes.

src/main/java/reposense/git/GitLog.java
```diff
@@ -14,6 +14,10 @@
  */
 public class GitLog {
 
+    /**
+     * Returns the git commit log info of {@code Author},
+     * in the repository specified in {@code config}
```
yamidark: This can be in a single line (max 120 chars in single line for `java` files). Also, no need to split into newlines if they are not separate points/sentences.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -79,15 +89,51 @@ static String convertToGitFormatsArgs(List<Format> formats) {
 
     /**
      * Returns the {@code String} command to specify the globs to exclude for `git log` command.
+     * Also checks that every glob in {@code ignoreGlobList} only targets files within the given repository {@code root}
```
yamidark: Index `.` missing at end of sentence.

src/test/java/reposense/util/TestUtil.java
```diff
@@ -141,6 +142,29 @@ public static boolean compareNumberExpectedCommitsToGitLogLines(int expectedNumb
         return expectedNumberCommits * 3 - 1 == gitLogResult.split("\n").length;
     }
 
+    /**
+     * Returns true if the {@code expectedNumberFilesChanged} is equal to the actual number of files changed in
+     * {@code gitLogResult}
```
yamidark: Index `.` missing here as well

src/test/java/reposense/git/GitUtilTest.java
```diff
@@ -0,0 +1,46 @@
...
+    public void gitUtil_convertToGitExcludeGlobArgs_success() {
+        File repoRoot = new File(config.getRepoRoot());
+        final String cmdFormat = " " + addQuote(":(exclude)%s");
+        final String emptyResult = "";
+
+        String result = convertToGitExcludeGlobArgs(repoRoot, Collections.EMPTY_LIST);
+        Assert.assertEquals(result, emptyResult);
+
+        result = convertToGitExcludeGlobArgs(repoRoot, Collections.singletonList("**.js"));
+        Assert.assertEquals(result, String.format(cmdFormat, "**.js"));
```
yamidark: By convention, `expected` is the 1st argument, and `actual` is the 2nd argument when we use `assertEquals`.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -79,15 +89,51 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
      */
-    static String convertToGitExcludeGlobArgs(List<String> ignoreGlobList) {
+    public static String convertToGitExcludeGlobArgs(File root, List<String> ignoreGlobList) {
         StringBuilder gitExcludeGlobArgsBuilder = new StringBuilder();
         final String cmdFormat = " " + addQuote(":(exclude)%s");
         ignoreGlobList.stream()
-                .filter(item -> !item.isEmpty())
+                .filter(item -> !item.isEmpty() && isValidPath(root, item))
```
yamidark: Let's add the `empty` string check inside the `isValidPath` method as well, so we don't need to have 2 different boolean value here.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -79,15 +89,51 @@ static String convertToGitFormatsArgs(List<Format> formats) {
...
                 .map(ignoreGlob -> String.format(cmdFormat, ignoreGlob))
                 .forEach(gitExcludeGlobArgsBuilder::append);
 
         return gitExcludeGlobArgsBuilder.toString();
     }
+
+    /**
+     * Returns true if the {@code path} is inside the current repository
+     * Produces log messages when the invalid file path is skipped
```
yamidark: Missing index `.` at end of each sentence.

src/main/java/reposense/git/GitLog.java
chelseyong: The whole line is actually more than 120 chars.

yamidark: Whoops, then break the line at the word which passes the limit.

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,69 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+
+        // Existing file is ignored
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("annotationTest.java"));
+        String content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(5, content));
+
+        // Non-existing files that are valid
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("**collate"));
+        content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(6, content));
```
eugenepeh: lets make all these test more relevant (show some changes) to the alpha repo. 
https://github.com/reposense/testrepo-Alpha

E.g. **collate is not found in testrepo alpha, prob use **11612 or something.

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,69 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
 
-        String content = GitLog.get(config, ignoreAllJavaFilesAuthor);
-        Assert.assertTrue(TestUtil.compareNumberExpectedCommitsToGitLogLines(2, content));
+        String content = GitLog.getWithFiles(config, ignoreAllJavaFilesAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(0, content));
+    }
+
+    @Test
+    public void gitLog_authorIgnoreValidFilePath_success() {
+        config.setFormats(Collections.singletonList(new Format("java")));
```
eugenepeh: lets not set any formats so that we can utilize more files for this test

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,69 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+
+        ignoreInvalidFileAuthor.setIgnoreGlobList(Collections.singletonList("../bin"));
+        String content = GitLog.getWithFiles(config, ignoreInvalidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(6, content));
+
+        ignoreInvalidFileAuthor.setIgnoreGlobList(Collections.singletonList("../*.java"));
+        content = GitLog.getWithFiles(config, ignoreInvalidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(6, content));
+
+        ignoreInvalidFileAuthor.setIgnoreGlobList(Collections.singletonList("/collate"));
```
eugenepeh: same for these, can set to `/newPos` instead

src/test/java/reposense/git/GitLogTest.java
chelseyong: `.attach_pid11612` is not registered in the list of files. Perhaps because it is not in the list of `DEFAULT_FORMATS`.
I will use another example. Actually some of these tests are for files that are not found.

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,68 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("**Test**"));
+        content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(5, content));
+
+        // Files that do not exist
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("**blameTest"));
+        content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(7, content));
+
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("nonExisting.txt"));
```
eugenepeh: this one i think can use `README.md`?

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,68 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+        content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(7, content));
+
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("./java"));
```
eugenepeh: this can use any value file, prefix with the `./` to demonstrate that it is valid and working.
e.g. `./newPos` to show the clear difference between this and the invalid example

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,68 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+        content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(7, content));
+
+    }
+
+    @Test
+    public void gitLog_authorIgnoreInvalidFilePath_success() {
+        config.setFormats(Collections.singletonList(new Format("java")));
```
eugenepeh: do remove this so that the result is visually comparable from the other test.

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,68 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("annotationTest.java"));
+        String content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(6, content));
+
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("**Test**"));
+        content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(5, content));
+
+        // Files that do not exist
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("**blameTest"));
```
eugenepeh: I think this can just use `**.java` or something.

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,68 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
         ignoreAllJavaFilesAuthor.setIgnoreGlobList(Collections.singletonList("*.java"));
 
-        String content = GitLog.get(config, ignoreAllJavaFilesAuthor);
-        Assert.assertTrue(TestUtil.compareNumberExpectedCommitsToGitLogLines(2, content));
+        String content = GitLog.getWithFiles(config, ignoreAllJavaFilesAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(0, content));
+    }
+
+    @Test
+    public void gitLog_authorIgnoreValidFilePath_success() {
```
eugenepeh: Sorry, i didn't identify this earlier.
`authorIgnoreValidFilePath` -> `validIgnoreGlobs`.

The author can be removed since this applies to the global ignore glob as well.

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,68 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+        Author ignoreValidFileAuthor = getAlphaAllAliasAuthor();
```
eugenepeh: can just name it `author` since this is not a specially created author for this test.

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,68 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(7, content));
+
+        ignoreValidFileAuthor.setIgnoreGlobList(Collections.singletonList("./java"));
+        content = GitLog.getWithFiles(config, ignoreValidFileAuthor);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(7, content));
+
+    }
+
+    @Test
+    public void gitLog_authorIgnoreInvalidFilePath_success() {
```
eugenepeh: similarly, here should be `gitLog_invalidIgnoreGlobs_filtered`

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -64,17 +64,68 @@ public void gitLog_includeAllJavaFilesAuthorIgnoreMovedFile_success() {
...
+        config.setFormats(Collections.singletonList(new Format("java")));
+        Author ignoreInvalidFileAuthor = getAlphaAllAliasAuthor();
+
+        ignoreInvalidFileAuthor.setIgnoreGlobList(Collections.singletonList("../bin"));
```
eugenepeh: make this `../testrepo-Alpha` because thats the folder name when u cloned the repo

eugenepeh:
just add 1 non existence glob for the success case will do.

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -79,53 +79,50 @@ public void gitLog_authorIgnoreAllJavaFiles_success() {
...
+        content = GitLog.getWithFiles(config, author);
+        Assert.assertTrue(TestUtil.compareNumberFilesChanged(6, content));
 
     }
 
     @Test
-    public void gitLog_authorIgnoreInvalidFilePath_success() {
-        config.setFormats(Collections.singletonList(new Format("java")));
-        Author ignoreInvalidFileAuthor = getAlphaAllAliasAuthor();
+    public void gitLog_invalidIgnoreGlobs_success() {
```
eugenepeh: last word should be filtered, because thats the expected behavior

`gitLog_invalidIgnoreGlobs_filtered`

yamidark:
Slights nits to proposed message:
```
If users try to put folders existing outside of the git repository 
inside the ignore-glob-list, it will lead to a RuntimeException as the
`git log` command fails.

Let's check that the path is inside the repository, before writing it
into the ignore-glob-list and running the `git log` command.
```

PR #597, opened 2019-03-11
Remove unused checkstyle analysis feature

### LINKED ISSUE #510: Remove unused CheckStyle plugin
@damithc , tanghao seems to have sort of planned to integrate checkstyle into RepoSense but doesn't seems to be working nor complete. 

Do you still require this feature or should we remove it from repository for now?

fzdy1914:
Fix #510 

The checkstyle jar file is added to the code base in the early stage of RepoSense
when there was some hope of running static analyzers over the code written by 
each user. 

This feature has not been used/commented out since early stages (v1.0), and 
it unnecessarily complicates our codebase.

Let's remove unused checkstyle analysis feature in our code base.


yamidark:
The `checkstyle` jar file was initially used also for analysis of the repositories (not our own code quality), marking each LOC with any checkstyle issues. However, we have since removed this feature as it seems to take up lots of space, and also wasn't very useful, IIRC. 

@damithc Can we double-check if we still wish to have this `checkstyle` additional analysis on each LOC?

If not,  @fzdy1914 we can also remove this unneeded LOC as well:
https://github.com/reposense/RepoSense/blob/7c06d6928ebf4ca1e1d662c60060c485b2903622/src/main/java/reposense/system/CommandRunner.java#L11-L17

damithc:
> @damithc Can we double-check if we still wish to have this `checkstyle` additional analysis on each LOC?

I have no recollection of this. What does it do?

yamidark:
>I have no recollection of this. What does it do?

damithc:
Thanks for checking @yamidark 
Yes, we can remove it.
Background: at the early stages of the project there was some hope of running static analyzers over the code written by each student. I don't think that feature is in our immediate plans anymore.

yamidark:
@fzdy1914 You may proceed to remove this checkstyle feature, such as removing `CheckStyleParser` as shown above, the `needCheckStyle` boolean in `RepoConfiguration`, the `checkStyleRaw` command in `CommandRunner`, and etc.

yamidark:
@fzdy1914 Don't think the images should be removed, they are for configuring the checkstyle settings for our local RepoSense project (in the [Using CheckStyle](https://github.com/reposense/RepoSense/blob/master/docs/UsingCheckstyle.md) docs), not the added checkstyle analysis on other repositories using RepoSense.

yamidark:
https://github.com/reposense/RepoSense/blob/daaed304969cb9694e002f0fa4ff7b6a46d4215e/src/main/java/reposense/authorship/model/LineInfo.java#L15-L22

Do remove the `issues` variable and the commented out portion, as well as any method refering to that.
The `IssueInfo` class should also be removed, since it is no longer used as well.

yamidark:
For your proposed message, while this checkstyle feature is indeed not in our immediate plan, the main reason why we are removing it is because it has not been used/commented out since early stages (v1.0), and thus it unnecessarily complicates the codebase.

yamidark:
@fzdy1914 Sorry, I meant to add on the reason I gave and not completely remove the other reason 😓

Nits to proposed message:
```
The checkstyle analysis feature was added to the code base in the
early stage of RepoSense, when there was some hope of running static
analysis over the code written by authors of the repositories.

As this feature has not been used since early stages(v1.0), and it
is also not in our immediate plans anymore, leaving it will
unnecessarily complicate the codebase.

Let's clean up the code base by removing this unused checkstyle
analysis feature.
```

PR #608, opened 2019-03-19
[#581] CodeView#filter: add interactive selection switch

### LINKED ISSUE #581: CodeView#filter: add interactive selection switch
As discussed in https://github.com/reposense/RepoSense/pull/532#pullrequestreview-204250162, `the filter selection's radio button serves nothing more than an "indicator"`. Changing option doesn't do anything until the corresponding controls are modified. Hence, it is rather redundant for users to manually switch the selection. To simplify things in #532, this has been discussed to be implemented separately in another PR. 

Hence, we should improve the users' satisfaction by implementing this change after the addition of #532. That is to `Instead of disabling the deselected option, let's make interaction switch the selection`.

fzdy1914:
Fixes #581 

The filter selection's radio button serves nothing more than an "indicator". 
Changing option doesn't do anything until the corresponding controls are
modified. 

It is redundant and inconvenient for users to manually switch the selection.

Instead of disabling the deselected option, let's allow interaction switch 
the selection.


fzdy1914:
> Perhaps the indicated selection (i.e. radio button) can update when the user presses but have not entered anything into the filter box. This can provide more user feedback by showing that the selection has changed.

Have edited it as request.

fzdy1914:
> I notice the filter box is not "disabled" (looks grey) after switching. Is it intended this way?

It should be the intended behavior mentioned in #581 .

damithc:
hmm... after playing around with the feature, it feels like the previous state was more intuitive although it takes an extra click. What do you guys think?
Perhaps we should allow both ways of switching between the two modes?

fzdy1914:
> hmm... after playing around with the feature, it feels like the previous state was more intuitive although it takes an extra click. What do you guys think?
> Perhaps we should allow both ways of switching between the two modes?

@eugenepeh @yamidark @yong24s Your opinions on this?

eugenepeh:
>@eugenepeh @yamidark @yong24s Your opinions on this?

Sounds good to me

yamidark:
>@eugenepeh @yamidark @yong24s Your opinions on this?

Sounds good to me as well

fzdy1914:
@damithc I have changed as request, please review the latest commit.

frontend/src/index.jade
```diff
@@ -211,13 +211,15 @@ html
...
               form.file-picker.mui-form--inline(onsubmit="return false")
-                input#search(type="search", placeholder="Filter by glob", v-on:change="updateFilterSearch", disabled)
-                button#submit-button(type="submit", disabled) Filter
+                input#search(type="search", placeholder="Filter by glob", v-on:change="updateFilterSearch",
+                    v-on:click="indicateSearchBar")
+                button#submit-button(type="submit", v-on:click="indicateSearchBar") Filter
             .fileformat
-              input.radio-button--checkbox(type="radio", name="filter", value="checkboxes", v-on:click="enableCheckBoxes()",
-                  checked)
+              input.radio-button--checkbox(type="radio", name="filter", value="checkboxes", :checked="isCheckBoxes",
```
ongspxm: use the full version, instead of the shortform `:checked`, sticking to convention of the code repo

PR #611, opened 2019-03-19
[#599] CodeView: fix `All` checkbox indication failure

### LINKED ISSUE #599: Select-all checkbox is not updating itself when everything else is unchecked
image

fzdy1914:
Fixes #599 

If the user unchecks all the checkboxes of single file type in code view, 
the select-all checkbox of filetype remains checked, which is a bug because
it should be unchecked automatically.

This feature was broken in #487, where the author accidentally split 
`selectFileFormat` into two functions, which results that the status of select-all 
checkbox is not updated when checking/unchecking other checkboxes.

This feature is also related to a bug where the `fileFormats` in the frontend 
contains formats with 0 line, which results that the formats displayed may not
be equal to the formats stored in `fileFormats`.

Meanwhile, the logic of the check-all of select-all checkbox is flawed. If the 
user selects a checkbox after deselecting all of them, the select-all checkbox 
will not update correspondingly.

Let's fix the `fileFormats` bug, revert the separation made in #487 and also 
change the logic of select-all checkbox.


fzdy1914:
> The original purpose of having the select-all checkbox remaining unticked even after one of the checkboxes become unticked (assume all checkboxes are now ticked), is for the user to deselect all the checkboxes easily. If the user intends to deselect all the checkboxes, the current situation will require the user to press the select-all checkbox twice.

Maybe we can ask mentor's opinion on this.

> Also, there seems to be a bug that only happens sometimes... I was spam-pressing the select-all checkbox and found this:

Sorry, I am unable to reproduce this.

yamidark:
>The original purpose of having the select-all checkbox remaining unticked even after one of the checkboxes become unticked (assume all checkboxes are now ticked), is for the user to deselect all the checkboxes easily.

Sorry, I'm not too sure I follow this. You mentioned that after 1 checkbox is unticked, and then assume all checkboxes are now ticked, so which is it? I'm guessing you meant if all boxes are ticked, then we untick 1 box?

Anyway, I don't think this is much of an issue, as long as the `All` button remains consistent, such that it is either ticked/unticked when some (but not all) of the checkboxes are ticked (like the `Collapse/Expand All` button), I'm okay with it, since its quite 'easy' to tick or untick all by simply clicking twice.

chelseyong:
> I'm guessing you meant if all boxes are ticked, then we untick 1 box?

Yes

Here is the bug:
bug

fzdy1914:
> Here is the bug:

Quite weird, it does not happen for Eugene but happens for Aditya.

fzdy1914:
@chelseyong Any idea of the source of this bug? I tried a while and cannot locate it.

yamidark:
>Any idea of the source of this bug? I tried a while and cannot locate it.

This bug seems to occur for any author which has a 'small' amount of file formats (<5) or so.

chelseyong:
`this.filesLinesObj` contains file formats that are not supposed to be there, i.e. the author have extra file formats inside `fileFormats` (in v_authorship). This caused the number of selected file formats to not always match the total no of file formats.

The problem is actually propagated from the backend, where the file format contributions are generated in `AuthorshipSummary`:

https://github.com/reposense/RepoSense/blob/25c0f18bbffdbe16c87d3bfc351aba1f49f5671d/src/main/java/reposense/authorship/model/AuthorshipSummary.java#L26-L32

It seems extra file formats are added into the author's `fileFormatContributions` here.

chelseyong:
Btw, the `Expand/Collapse All` is not working, which is related to this (checking the number of formats), so fixing this will fix that too.

fzdy1914:
> `this.filesLinesObj` contains file formats that are not supposed to be there, i.e. the author have extra file formats inside `fileFormats` (in v_authorship). This caused the number of selected file formats to not always match the total no of file formats.
> 
> The problem is actually propagated from the backend, where the file format contributions are generated in `AuthorshipSummary`:
> 
> [RepoSense/src/main/java/reposense/authorship/model/AuthorshipSummary.java](https://github.com/reposense/RepoSense/blob/25c0f18bbffdbe16c87d3bfc351aba1f49f5671d/src/main/java/reposense/authorship/model/AuthorshipSummary.java#L26-L32)
> 
> Lines 26 to 32 in [25c0f18](/reposense/RepoSense/commit/25c0f18bbffdbe16c87d3bfc351aba1f49f5671d)
> 
>  authors.forEach((author) -> { 
>      LinkedHashMap<String, Integer> defaultFileFormatContribution = new LinkedHashMap<>(); 
>      for (Format format : formats) { 
>          defaultFileFormatContribution.put(format.toString(), 0); 
>      } 
>      authorFileFormatContributionMap.put(author, defaultFileFormatContribution); 
>  }); 
> It seems extra file formats are added into the author's `fileFormatContributions` here.

Thanks for your suggestion, will look into it.

fzdy1914:
In authorship.json, the file formats are correct.

fzdy1914:
image
The bug has identified, it seems that `fileFormats` contains all the format of repo but not all the formats of the author.

yamidark:
@fzdy1914 

As @damithc mentioned, avoid phrases such as `In our previous logic`, and what you renamed `the previous implementation` as that is implied.
Instead, we can just mention directly what the current implementation is: `The logic of the check-all of select-all checkbox is not so sound.`

Also, be more specific where the 'select-all checkbox' is at, even if there is only 1 currently, so it is easier for us to know where to find and test it (if needed).

eugenepeh:
@fzdy1914 

I don't quite understand your proposed message, can you refine it abit? Here's some feedback for you to improve it on:
```
The select-all checkbox of filetype in code view should be unchecked
when everything else is unchecked.
```
^ state the problem instead
```
This feature is related to a bug where the fileFormats in the frontend
contains formats with 0 lines. And this feature was broken in #487, where
the author accidentally split selectFileFormat into two functions.
```
^ elaborate more, what happens when `selectFileFormat` is split into 2
```
Also, the logic of the check-all of select-all checkbox is not so sound.
If the user unchecks all and checks on a single file, the select-all checkbox
will not update correspondingly.
```
^ instead of using `not so sound` perhaps can state that its flawed. Also, the second sentence can improve further. E.g. `leaving all but 1 checkbox checked` or `selecting a checkbox after deselected all of them`. 
```
Let's fix the fileFormats bug, revert the mistake and also change the logic
of select-all checkbox.
```
^ the above paragraph didn't really explicitly state the mistake

fzdy1914:
@eugenepeh I have edited my proposed message, please review.

eugenepeh:
Minor nits to the message:
```
The `All` checkbox of file formats remains checked even after the user
have performed the deselection of other checkboxes. This is a bug as
the `All` checkbox should have unchecked itself automatically.

This bug originates from #487, where the author have accidentally split
selectFileFormat into two functions. This cause the status of
`All` checkbox not being updated as the separated function was never
executed.

It is also related to a bug where the `fileFormats` in the frontend
contains formats with 0 lines, which results in the formats displayed
may not be equal to the formats stored in `fileFormats`.

Additionally, the logic of the `All` checkbox is flawed. When the user
selects a checkbox after deselecting all of them, the checkbox will not
update itself correspondingly.

To restore the intended functionality, let's revert the separation made
in #487 and correct the logic of the `All` checkbox.
```

Issue #590, opened 2019-03-07
Give a way to set website properties


damithc:
Current:
The deployed dashboard doesn't have a name, favicon, etc.
![image](https://user-images.githubusercontent.com/1673303/53924951-1b244500-40b9-11e9-9427-84216317be00.png)

Suggestion: Give a way to specify such website properties.

Items to do:
- [ ] A way to set favicon
- [ ] A way to set page title

fzdy1914:
Related to #467. I am considering to add a new config in our codebase with serves the role to configure the generated report. Including title, icon, intro or anything else which may need in the future to configure the report.

damithc:
> Related to #467. I am considering to add a new config in our codebase with serves the role to configure the generated report. Including title, icon, intro or anything else which may need in the future to configure the report.

Good idea. Note that a config file not may be the best format to hold HTML content, to be used in the intro, unless you restrict the intro to plain text only. The alternative is to expect these info to be in a specific files e.g., `/site_config/favicon.ico`  `/site_config/intro.html` etc.

Tejas2805:
@damithc What else do we have left for this? Anything else left or can we close this?

damithc:
> @damithc What else do we have left for this? Anything else left or can we close this?

I updated issue description to add the required tasks

PR #612, opened 2019-03-19
Add default title and icon for generated report

fzdy1914:
Part of #590 

We do not have a title and icon set for the generated report.

This causes the title to be automatically set to the url of the page
instead, which does not provide any useful information to the user.

Let's add a default title and icon for the generated report.


damithc:
In future, we should give users a way to set the page title to whatever they wish.

yamidark:
Nits to proposed message:

We do **not** have **a** ~~default~~ title and icon **set** for the generated report.

~~It is quite weird to have a long and meaningless title and empty
icon for the generated report.~~
**This causes the title to be automatically set to the url of the page instead,
which does not provide any useful information to the user.**

Let's add **a** default title and icon for the generated report.

PR #596, opened 2019-03-11
Page unresponsive when loading report of many repos with huge date gap #542

emer7:
Fix #542 

```
RepoSense is unresponsive if we load many repos
with huge date gap, making it nearly unusable.

This is caused by the report generated having a lot of
empty commits per day, in which the frontend needs to
sift through to display the report.

Let's fix this issue by removing all those empty commits,
including only commits that are necessary.
```

PR #615, opened 2019-03-20
summary.json: move sinceDate and untilDate become direct element #607

### LINKED ISSUE #607: summary.json: move sinceDate and untilDate become direct element
From #596

`summary.json` has `repos` field with `sinceDate` and `untilDate` fields in it.

This will results in duplicated `sinceDate` and `untilDate` in `repos`

Let's move `sinceDate` and `untilDate` into the "upper level"

emer7:
Fix #607

```
summary.json has multiple identical sinceDate and untilDate
in each repo that is analyzed, hence redundant.

This is caused by sinceDate and untilDate being instance variable
of repoConfig

Let's rectify this problem by moving sinceDate and untilDate
out of repoConfig into summaryReportJson
```

This PR will affect PR #596 and PR #539


emer7:
@reposense/stage1-reviewers Ready for review!

src/main/java/reposense/commits/CommitInfoExtractor.java
```diff
@@ -38,8 +43,8 @@
     }
 
     /**
-     * Parses the {@code gitLogResult} into a list of {@code CommitInfo} and returns it.
-     */
+    * Parses the {@code gitLogResult} into a list of {@code CommitInfo} and returns it.
```
chelseyong: Spaces are missing here.

src/main/java/reposense/RepoSense.java
```diff
@@ -50,9 +50,10 @@ public static void main(String[] args) {
             }
 
             RepoConfiguration.setFormatsToRepoConfigs(configs, cliArguments.getFormats());
-            RepoConfiguration.setDatesToRepoConfigs(configs, cliArguments.getSinceDate(), cliArguments.getUntilDate());
             ReportGenerator.generateReposReport(configs, cliArguments.getOutputFilePath().toAbsolutePath().toString(),
-                    formatter.format(ZonedDateTime.now(ZoneId.of("UTC+8"))));
+                    formatter.format(ZonedDateTime.now(ZoneId.of("UTC+8"))),
+                    cliArguments.getSinceDate().orElse(null),
```
fzdy1914: I believe the previous authors use Optional for some reason. Maybe due to avoid using null in our code. So I suggest you should replace all null with Optional in the code.

emer7: In the original code the previous authors also resort to `orElse(null)`

emer7:
@reposense/stage2-reviewers ready for review

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -24,8 +22,6 @@
     private RepoLocation location;
     private String branch;
     private String displayName;
-    private Date sinceDate;
-    private Date untilDate;
```
yamidark: Why are we removing this 2 variables from `RepoConfiguration`? 
These are part of the configuration settings, and should still remain here.
Removing this 2 ends up causing lots of redundant changes throughout the code, as we have to manually pass this 2 values down alongside the `RepoConfiguration`, which is unnecessary.

If we don't need these 2 to be printed out in the 'json' file, we can just set them both as `transient` instead.

yamidark:
I believe the frontend portion also needs to be edited to take the `sinceDate` and `untilDate` correctly from `summary.json`.

src/main/java/reposense/model/RepoConfiguration.java
emer7: @yamidark fixed

emer7:
> I believe the frontend portion also needs to be edited to take the `sinceDate` and `untilDate` correctly from `summary.json`.

No need because sinceDate and untilDate is not used so far (it is used in my other 2 PR)

emer7:
@reposense/stage2-reviewers ready for review

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -42,13 +43,15 @@
...
-            String generationDate) throws IOException {
+            String generationDate, Date sinceDate, Date untilDate) throws IOException {
         InputStream is = RepoSense.class.getResourceAsStream(TEMPLATE_FILE);
         FileUtil.copyTemplate(is, outputPath);
 
         cloneAndAnalyzeRepos(configs, outputPath);
 
-        FileUtil.writeJsonFile(new SummaryReportJson(configs, generationDate), getSummaryResultPath(outputPath));
+        FileUtil.writeJsonFile(new SummaryReportJson(
+                configs, generationDate, sinceDate, untilDate),
```
yamidark: This can be put together on the line above right (without passing the 120 characters limit)?

emer7:
@reposense/stage2-reviewers fixed last nit spotted by Jun An

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -86,7 +86,8 @@ private void generateReport(String inputDates)
         RepoConfiguration.setDatesToRepoConfigs(
                 repoConfigs, cliArguments.getSinceDate(), cliArguments.getUntilDate());
 
-        ReportGenerator.generateReposReport(repoConfigs, FT_TEMP_DIR, TEST_REPORT_GENERATED_TIME);
+        ReportGenerator.generateReposReport(repoConfigs, FT_TEMP_DIR, TEST_REPORT_GENERATED_TIME,
+                cliArguments.getSinceDate().orElse(null), cliArguments.getUntilDate().orElse(null));
```
eugenepeh: if not provided, should retrieve the since date from the configs, such as the earliest commit of all repos.

emer7: It is handled automatically by other parts of the code I believe. The JSON file all this time does not have sinceDate if it is not specified

emer7:
> seems like nothing from the js has incorporate this change?

@eugenepeh yes because it has never been used all this time until PR #596 and #539

src/systemtest/java/reposense/ConfigSystemTest.java
eugenepeh: Yes, its handled by the frontend. We should move it to the back as browser has limited ram and processing power.

emer7:
@reposense/stage2-reviewers ready for review

emer7:
There is big change to the expected file in systemTest because turns out the revised one should be the right analysis result. I compared it by running no date range and date range using same the same date range (we can get the implicit date range from running no date range, and I used that implicit date range explicitly in date range)

eugenepeh:
Doesn't seems quite right to me


Git blame from github 
image

The new changes
image

https://deploy-preview-615--reposense.netlify.com/#search=&sort=displayName&since=2017-10-09&until=2019-03-25&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=false&tabAuthor=lohtianwei&tabRepo=reposense_testrepo-Delta_master

src/test/java/seedu/address/logic/parser/AddressBookParserTest.java

emer7:
> Doesn't seems quite right to me
> 
> Git blame from github
> image
> 
> The new changes
> image
> 
> https://deploy-preview-615--reposense.netlify.com/#search=&sort=displayName&since=2017-10-09&until=2019-03-25&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=false&tabAuthor=lohtianwei&tabRepo=reposense_testrepo-Delta_master
> 
> src/test/java/seedu/address/logic/parser/AddressBookParserTest.java

@eugenepeh can I know what is netlify arguments?

eugenepeh:
>@eugenepeh can I know what is netlify arguments?

Its just a plain `gradlew run`

>Because running
--since 01/03/2018 --until 15/04/2018 --config ./src/systemtest/resources on master branch
will yield the same result with running
--since 01/03/2018 --until 15/04/2018 --config ./src/systemtest/resources on 607-move-date branch
and running
--config ./src/systemtest/resources on 607-move-date branch (no date)

What config are you using and which repo is showing difference? Need to be more specific.
Are you running Delta alone? What are the files showing difference?

emer7:
> What config are you using and which repo is showing difference? Need to be more specific.

I am using the config in src/systemtest/resources/repo-config.csv and src/systemtest/resources/author-config.csv
The repo that is showing difference is testrepo-Delta

> Are you running Delta alone?

The `repo-config` has other repo beside Delta, but the repo that is showing difference is Delta given the date range of `01/03/2018` until `15/04/2018`.

Why I am using that date? Because if we do not supply any date argument when running RepoSense, it will automatically analyze the repo, and by debuggin the app, the date range that I get for `testrepo-Delta` is `01/03/2018` until `15/04/2018`. This is also why `commits.json` of running RepoSense without date arguments is the similar with running Reposense with date range (the only difference is the `authorFinalContributionMap`).

> What are the files showing difference?

It is the same file you pointed out.
`src/test/java/seedu/address/logic/parser/AddressBookParserTest.java`

eugenepeh:
@emer7 

The questions I asked above is things that you should include and investigate when reporting a bug / error. 

Main problem is still that the new changes doesn't seems to be correct. If you think that there is an underlying problem, you will need to investigate and report it as an issue.

emer7:
@eugenepeh 
This is what I get after investigating:

Running `RepoSense` without any date argument, upon reaching
```
// CommitInfoExtractor.java#extractCommitInfos:33
String gitLogResult = GitLog.get(config, author);
```
It will get the git log, but it seems to get it arbitrarily. In the case of `testrepo-delta`, it gets all the commits from 01/03/2018 to 15/04/2018.
But when it reaches `CommitInfoExtractor.java#extractFileInfos`, it will get all the file info regardless of date range.
This is what causing running
`--config ./src/systemtest/resources --since 01/03/2018 --until 15/04/2018` on master branch
and
`--config ./src/systemtest/resources` on master branch
to have *almost the same* `commits.json` but different `authorship.json` I believe

So is this the intended design? (To have git log get the log of commits with arbitrary date range) or is this a bug?

eugenepeh:
>But when it reaches CommitInfoExtractor.java#extractFileInfos, it will get all the file info regardless of date range if it is not provided.

It is meant to retrieve all the affected files in the given date range. 

>So is this an intended design or a bug?

If this only cause an issue when a date range is issued, then most likely it is a bug. 

>Almost the same as in the commits list are the same in commits.json, but there is slight difference in contribution, etc. (will investigate further).

Yes, please do investigate further. It will help you understand the back end architecture better if you haven't already.

emer7:
@eugenepeh fixed and worked already. No bug

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -227,4 +238,16 @@ private static String getIndividualAuthorshipPath(String repoReportDirectory) {
     private static String getIndividualCommitsPath(String repoReportDirectory) {
         return repoReportDirectory + "/commits.json";
     }
+
+    public static void setEarliestSinceDate(Date newEarliestSinceDate) {
+        if (earliestSinceDate == null || earliestSinceDate.after(newEarliestSinceDate)) {
```
eugenepeh: I think normally when people look for the earliest date/object, they usually compare who comes before, instead of after, in multiple comparison context. To standardize with that convention, let's change it to use `before` instead

emer7: @eugenepeh done

yamidark:
@emer7 Do resolve the conflicts

emer7:
@yamidark resolved

yamidark:
```
summary.json has multiple identical sinceDate and untilDate
in each repo that is analyzed, which is redundant.

This is caused by sinceDate and untilDate being instance variable
of RepoConfiguration.

Let's rectify this problem by moving sinceDate and untilDate
out of RepoConfiguration into SummaryReportJson.
```

PR #616, opened 2019-03-22
Console output: make logs clearer after cloning in parallel #609

### LINKED ISSUE #609: Cloning in parallel makes some of log message unclear
As, titile.

See the log message below, we can not determine which repo is the underlined log message belonged to.

image

It is better to add the repo detail to all the log messages.

myteo:
Fixes #609.

```
After #560, log messages during analysis of a repository became unclear
as there can be more than one repository being cloned and analyzed at
one time.

This can be confusing and reduce the usefulness of the log messages.

Lets add repository information (including location and branch) to make
the logs clearer.
```


fzdy1914:
> Lets add a log message at the beginning of the analysis of each
> repository so it is clear that all subsequent log messages refer to
> that particular repository.

This is one possible solution, but I am not this is intuitive enough.

myteo:
> Is there any logging for cloning too?

Yup there is logging for cloning.

myteo:
> This is one possible solution, but I am not this is intuitive enough.

Ok, I will add repo detail to the log messages where possible. Some of the areas that we log are unaware of repo detail, for example the commit hash log message that you showed in the issue.

myteo:
Added log message after repo is analyzed to fix issue that @damithc mentioned in Slack channel. Also I realized that when analyzing multiple different branches of the same repository, the log message does not contain information about the branch, so I added that in as well.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -53,7 +53,8 @@
      * Extracts a list of relevant files given in {@code config}.
      */
     public static List<FileInfo> extractFileInfos(RepoConfiguration config) {
-        logger.info("Extracting relevant file info from " + config.getLocation() + "...");
+        logger.info(String.format("Extracting relevant file info from %s (%s)...",
```
eugenepeh: let's create the corresponding constants for these strings as well.

yamidark:
Nits to proposed message:

After #560, log messages during analysis of a repository became unclear
as there can be **another repository being cloned at the same time.** 
~~more than one repository being cloned and analyzed at
one time.~~

**This causes the log messages of the two repositories to interleave with one other, 
~~This can be~~ which can be** confusing, and reduces the usefulness of the log messages.

Lets add repository information (including location and branch) to **the log messages
to** make ~~the logs~~ **it** clearer **which repository is being cloned or analyzed**.

PR #617, opened 2019-03-22
author-config#email: include + operator infinite email matcher #613

### LINKED ISSUE #613: author-config#email: include + operator infinite email matcher
In some cases, instead of allocating the simple default email to the user using just the GitHub ID, GitHub allocates them prefixed with numbers and the `+` operator. For e.g. `35282962+githubid@users.noreply.github.com`. The `+` operator is commonly used in mail servers, such as Google, as aliases or means to allow users to possess infinite number of emails while having them remain identifiable. 

By adding this pattern into our email matcher, this will increase more number of authors that can be identified by default, hence increase the convenience of our users.

myteo:
Fixes #613.

```
In some cases, instead of allocating the simple default email to the
user using just the GitHub ID, GitHub allocates them prefixed with
numbers and the + operator. For e.g.
35282962+githubid@users.noreply.github.com. The + operator is commonly
used in mail servers, such as Google, as aliases or means to allow
users to possess infinite number of emails while having them remain
identifiable.

By adding this pattern into our email matcher, this will increase
more number of authors that can be identified by default, hence
increase the convenience of our users.

Lets add this pattern into our email matcher.
```


myteo:
Hmm not so straightforward to add test cases for this. Might have to add a commit to testrepo-Alpha by an author with such an email. Email should be easy to set, but adding commit will probably cause some existing test cases to fail.

fzdy1914:
Maybe you can look into commit 5ab03229f29254659ec48752178d5b156e900ef3 to see if some of the test cases can be changed.

myteo:
Thanks for the suggestion!

src/main/java/reposense/git/GitUtil.java
```diff
@@ -64,6 +69,19 @@ static String convertToFilterAuthorArgs(Author author) {
...
+    /**
+     * Converts email to regex, adding optional regex groups for better matching.
+     */
+    private static String convertToEmailRegex(String email) {
+        String[] emails = email.split("@");
+        String authorEmailRegex = EMAIL_PREFIX_REGEX_GROUP;
+        authorEmailRegex += StringsUtil.replaceSpecialSymbols(emails[0], ".");
+        authorEmailRegex += EMAIL_SUFFIX_REGEX_GROUP;
+        authorEmailRegex += ".";
+        authorEmailRegex += StringsUtil.replaceSpecialSymbols(emails[1], ".");
```
eugenepeh: here seems to assume that the email will definitely have the "@", which might not be the case. 
Note that all these are user input and tend to have misconfiguration, internationally or not.

eugenepeh: Also, there isn't actually any need to `improve the matching`. The `StringsUtil.replaceSpecialSymbols` is just a security feature to ensure theres are no malicious injection.

myteo: Oh no, I might have misunderstood the issue. I thought we were trying to improve matching of emails like `35282962+githubid@users.noreply.github.com`? What I did was to improve matching so that user can just specify `githubid@users.noreply.github.com` and `git log` will match emails with + operator prefixed or suffixed.

myteo: > here seems to assume that the email will definitely have the "@", which might not be the case.
> Note that all these are user input and tend to have misconfiguration, internationally or n

We do a validation of the email when Author is created => `Author#validateEmails`, which checks that there is a "@" in the email.

eugenepeh: >Oh no, I might have misunderstood the issue. I thought we were trying to improve matching of emails like 35282962+githubid@users.noreply.github.com? What I did was to improve matching so that user can just specify githubid@users.noreply.github.com and git log will match emails with + operator prefixed or suffixed.

Yes you're right but there isn't a need to have a suffix, isnt it?

myteo: Oh i see. I was trying to match emails like `mingyi+123@gmail.com`. I shall remove it.

myteo:
> LGTM
> 
> Btw @myteo sorry I should've provide an example, also the format i provided you earlier was wrong because I forgotten that the method name would contain underscore.
> Can you update the branch name to `617-FileAnalyzerTest-analyzeFile_emailWithAdditionOperator_success`

No worries, changed the branch name👍

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -185,4 +189,24 @@ public void addAuthorEmailsAndAliasesMapEntry(Author author, List<String> values
     public RepoLocation getLocation() {
         return location;
     }
+
+    /**
+     * Attempts to find matching {@code Author} given a name and an email.
+     * If no matching {@code Author} is found, {@code UNKNOWN_AUTHOR} is returned.
```
yamidark: `{@code Author#UNKNOWN_AUTHOR}`

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -162,6 +162,14 @@ public void update(StandaloneConfig standaloneConfig) {
         authorConfig.update(standaloneConfig, ignoreGlobList);
     }
 
+    /**
+     * Attempts to find matching {@code Author} given a name and an email.
+     * If no matching {@code Author} is found, {@code UNKNOWN_AUTHOR} is returned.
```
yamidark: Same here `{@code Author#UNKNOWN_AUTHOR}`

yamidark:
Updated proposed message 2nd paragraph to follow the one in the issue

PR #618, opened 2019-03-23
Redesigning the sorting toolbar

### LINKED ISSUE #601: Improve UI for toolbar of sorting options
As mentioned in https://github.com/orgs/reposense/teams/devs/discussions/1, 
we are working towards the toolbar goal.

Many types of sorting methods have been implemented already.
Let's add in the 2 middle columns to allow users to clearly define their sorting options.

image

chelseyong:
Fixes #601 
```
There are many sorting and grouping options for users to choose from, 
e.g. group by author/repo/none, sort by name/repo/contribution.

When grouped by repo, sorting is done within the group. Whereas, when 
grouped by none, sorting is done without any group.

Let's make it clearer for users to choose each sorting option (within
groups or by groups) by having 2 drop-down select menus.
```

Contribution and variance is not allowed for sorting by repos/author, because additional methods are required to add up the total contribution/variance for each group. Perhaps this can be continued from #486?

Note: `Name` sorting is performed on `usernames`, not `display names`.


chelseyong:
> Why do we have a sort within group by repo/branch when group by none and sort group byrepo/branch ?

To make it clearer, I have disabled `sort within group by`.

> This part should have reverse too if you decided to delete reverse checkbox.

Reversed options are already available here:
image

eugenepeh:
actually what I meant was, instead of just binding a single value, is there anyway we can bind 2 in those lines
`option(value="name asc", v-if="filterGroupSelection !== 'groupByRepos'") &uarr; name`

chelseyong:
> actually what I meant was, instead of just binding a single value, is there anyway we can bind 2 in those lines
> option(value="name asc", v-if="filterGroupSelection !== 'groupByRepos'") &uarr; name

With `v-model`, the isDsc value can be binded, but this has to be done for every option available, which will create a lot of variables. Can be quite messy.
With `v-bind`, query selectors are required to retrieve the selected option from the DOM, which would require tagging additional class names or ids onto each option. Messy css.
Whereas `computed props` from Vue actually does a good job in handling these more efficiently, they are only updated when the values change.

chelseyong:
`Computed props` may seem redundant in this case.. but it helps to cut down on a few lines of code.

eugenepeh:
>Computed props may seem redundant in this case.. but it helps to cut down on a few lines of code.

I see, would doing something like `value={ foo: foo, bar: bar }` simplify things further?
@ongspxm any ideas that can simplify her implementation further?

ongspxm:
commented

frontend/src/static/js/v_summary.js
```diff
@@ -456,6 +444,7 @@ window.vSummary = {
       return null;
     },
     updateSortSelection() {
+      this.getOptionWithOrder();
       if (this.filterGroupSelection === 'groupByAuthors') {
         if (!this.sortWithinGroupSelection || this.sortingWithinOption === 'name') {
           this.sortWithinGroupSelection = 'searchPath asc';
```
ongspxm: all these dun need change ah?

chelseyong: Referring to line 450? 
`sortWithinGroupSelection` has to be updated directly. Updating `sortingWithinOption` or `isSortingWithinDsc` will not update the UI, as its not bound 2-way.

ongspxm: makes sense, add comment `\\update UI`

frontend/src/static/js/v_summary.js
```diff
@@ -44,9 +44,10 @@ window.vSummary = {
     return {
       filtered: [],
       filterSearch: '',
-      filterSort: 'displayName',
       filterSortReverse: false,
       filterGroupSelection: 'groupByRepos',
+      sortGroupSelection: 'searchPath asc',
+      sortWithinGroupSelection: 'name asc',
```
ongspxm: not setting the options in here?

chelseyong: `getOptionWithOrder` is called within `sortFiltered`, which will be called upon initiation.

ongspxm: as in the intialization should be done here as well, just put an empty string

chelseyong: Okay added.

frontend/src/static/js/v_summary.js
```diff
@@ -428,13 +442,34 @@ window.vSummary = {
...
+        if (this.sortingOption !== 'searchPath') {
+          this.sortGroupSelection = 'searchPath asc';
+        }
+      } else if (this.filterGroupSelection === 'groupByNone') {
+        this.sortWithinGroupSelection = '';
       }
     },
+    getOptionWithOrder() {
+      [this.sortingOption, this.isSortingDsc] = this.sortGroupSelection.split(' ');
+      this.isSortingDsc = this.isSortingDsc === 'dsc';
```
eugenepeh: doesn't seems very nice to reassign the variable to become boolean.
Would something like this work?

`name` and `name dsc`

if just `name`, the `this.isSortingDsc` will automatically be null,
hence can just use `if (this.isSortingDsc)` to do the logic check.

chelseyong: Yes, that will be much simpler.

frontend/src/index.jade
```diff
@@ -79,13 +79,29 @@ html
...
+            .mui-select.sort-group
+              select(v-model="sortGroupSelection")
+                option(value="name", v-if="filterGroupSelection !== 'groupByRepos'") &uarr; name
+                option(value="name dsc", v-if="filterGroupSelection !== 'groupByRepos'") &darr; name
+                option(value="searchPath", v-if="filterGroupSelection !== 'groupByAuthors'") &uarr; repo/branch
+                option(value="searchPath dsc", v-if="filterGroupSelection !== 'groupByAuthors'") &darr; repo/branch
+                option(value="totalCommits", v-if="filterGroupSelection === 'groupByNone'") &uarr; contribution
+                option(value="totalCommits dsc", v-if="filterGroupSelection === 'groupByNone'") &darr; contribution
+                option(value="variance", v-if="filterGroupSelection === 'groupByNone'") &uarr; variance
+                option(value="variance dsc", v-if="filterGroupSelection === 'groupByNone'") &darr; variance
```
eugenepeh: just noticed this.

It seems that variance and totalCommits are only enabled when authors wasn't grouped. 
Was this the supposed functionality?

chelseyong: Originally intended to continue off from #486, but I can implement it here as well. Is sorting by variance is taken at average?

eugenepeh: >Originally intended to continue off from #486, but I can implement it here as well. Is sorting by variance is taken at average?

I see. I guess we can continue in #486.

frontend/src/static/js/v_summary.js
```diff
@@ -41,9 +41,14 @@ window.vSummary = {
...
       filterSearch: '',
-      filterSort: 'displayName',
       filterSortReverse: false,
       filterGroupSelection: 'groupByRepos',
+      sortGroupSelection: 'searchPath',
+      sortingOption: '',
+      isSortingDsc: '',
+      sortWithinGroupSelection: 'name',
+      sortingWithinOption: '',
+      isSortingWithinDsc: '',
```
ongspxm: minor nit, right now quite hard to know which is used in the code, which is use for UI

you can either group all those UI variables together, use the 'filter' keywork, add comments
choose any you think can fit, i think grouping and adding comments will be the best.

chelseyong: Okay changed.

chelseyong:
@eugenepeh can this be merged soon, to continue with #486?

eugenepeh:
Nits to the proposed message:

```
When the ramp charts are grouped together, the sorting is only
performed within the group. This may differs from the expectation of
some users, who may wish to sort them by groups e.g. to list them
accordingly to the group's combined contribution. 

Also, the `reverse` option is not well-presented nor intuitive to the
users enough to understand that it is meant to reverses the sorted
listing. 

Let's fix the mentioned problems with sorting function by adding
another drop-down select menu to allow the sorting of groups, and
include direction arrows in menu items to replace the need of reverse
option.
```

Note that you should mention the problems more precisely.

PR #623, opened 2019-03-24
CsvParser#parse: fix error handling of `processLine` #622

### LINKED ISSUE #622: CsvParser: fix error handling in `processLine`
When processing a line that is illegal or erroneous, `processLine` will throw an exception to the caller, `parse`, and the process gets terminated.

This result in the rise of an issue in https://github.com/nus-cs3281/2019-dashboard/commit/ef0791fa41c2718f61990db22b64ad237ceb12e5 where a misconfigured line in author-config.csv cause the subsequent lines of authors not being parsed and added to the analysis.

eugenepeh:
Fix #622 

```
When processing a line that is illegal or erroneous, `processLine` will
throw an exception to the caller, `parse`, and the process gets
terminated.

This result in the rise of an issue in
nus-cs3281/2019-dashboard@ef0791f where a misconfigured line in
author-config.csv causes the subsequent lines of authors not being
parsed and added to the analysis.

As the lines in our csv config are meant to be separate entities, let's
ignore any misconfigurations and move on to parsing the subsequent
lines.
```


src/main/java/reposense/parser/CsvParser.java
```diff
@@ -43,27 +46,27 @@ public CsvParser(Path csvFilePath) throws IOException {
...
                 if (line.isEmpty() || isLineMalformed(elements, lineNumber, line)) {
                     continue;
                 }
 
-                processLine(results, elements);
-                lineNumber++;
+                try {
+                    processLine(results, elements);
+                } catch (ParseException pe) {
+                    logger.warning(String.format(MESSAGE_LINE_PARSE_EXCEPTION_FORMAT, lineNumber, pe.getMessage()));
```
myteo: Maybe we can add the line to the log message as well, similar to the malformed line log message.

myteo: I realize my phrasing might be quite hard to understand. What I mean is this:

```
private static final String MESSAGE_LINE_PARSE_EXCEPTION_FORMAT =
    "Warning! Error parsing line %d in configuration file.\n"
    + "Contents: %s"        
    + "Error: %s";
```

```
logger.warning(String.format(MESSAGE_LINE_PARSE_EXCEPTION_FORMAT, lineNumber, line, pe.getMessage()));
```

PR #625, opened 2019-03-24
Use single clone to analyze multiple branches of a repository #337

### LINKED ISSUE #337: Remove multiple clones when analyzing multiple branch of a repository

myteo:
Fixes #337.

```
When analyzing multiple branches of a repository, RepoSense will clone
that same repository as many times as the number of the branches if
they are not listed consecutively in the config file.

This is inefficient as all the analysis were performed against the same
repo. Hence, there shouldn't be a necessity to clone it multiple times.

Lets use a single clone to analyze multiple branches of a repository.
```

### DIFF
--- a/src/main/java/reposense/report/RepoCloner.java
+++ b/src/main/java/reposense/report/RepoCloner.java
@@ -13,6 +13,7 @@ import java.util.logging.Logger;
 import reposense.git.GitBranch;
 import reposense.git.GitCheckout;
 import reposense.model.RepoConfiguration;
+import reposense.model.RepoLocation;
 import reposense.system.CommandRunnerProcess;
 import reposense.system.CommandRunnerProcessException;
 import reposense.system.LogsManager;
@@ -38,17 +39,15 @@ public class RepoCloner {
      */
     public void clone(String outputPath, RepoConfiguration config) throws IOException {
         configs[currentIndex] = config;
-        isCurrentRepoCloned = true;
-        if (isPreviousRepoDifferent()) {
-            isCurrentRepoCloned = spawnCloneProcess(outputPath, config);
-        }
+        isCurrentRepoCloned = spawnCloneProcess(outputPath, config);
     }
 
     /**
-     * Waits for current clone process to finish executing and returns the corresponding {@code RepoConfiguration}.
+     * Waits for current clone process to finish executing and returns the {@code RepoLocation} of the corresponding
+     * {@code RepoConfiguration}.
      */
-    public RepoConfiguration getClonedRepo(String outputPath) throws IOException {
-        if (isCurrentRepoCloned && isPreviousRepoDifferent()) {
+    public RepoLocation getClonedRepo(String outputPath) throws IOException {
+        if (isCurrentRepoCloned) {
             isCurrentRepoCloned = waitForCloneProcess(outputPath, configs[currentIndex]);
         }
 
@@ -57,23 +56,28 @@ public class RepoCloner {
             return null;
         }
 
-        if (isPreviousRepoDifferent()) {
-            currentRepoDefaultBranch = GitBranch.getCurrentBranch(configs[currentIndex].getRepoRoot());
-        }
-        configs[currentIndex].updateBranch(currentRepoDefaultBranch);
-
+        currentRepoDefaultBranch = GitBranch.getCurrentBranch(configs[currentIndex].getRepoRoot());
         cleanupPrevRepoFolder();
 
-        try {
-            GitCheckout.checkout(configs[currentIndex].getRepoRoot(), configs[currentIndex].getBranch());
-        } catch (RuntimeException e) {
-            logger.log(Level.SEVERE, "Branch does not exist! Analysis terminated.", e);
-            handleCloningFailed(outputPath, configs[currentIndex]);
-            return null;
-        }
         previousIndex = currentIndex;
         currentIndex = (currentIndex + 1) % configs.length;
-        return configs[previousIndex];
+        return configs[previousIndex].getLocation();
+    }
+
+    /**
+     * Prepares repo specified by {@code config} to be analyzed by updating and checking out its branch.
+     */
+    public boolean prepareToAnalyze(String outputPath, RepoConfiguration config) throws IOException {
+        config.updateBranch(currentRepoDefaultBranch);
+
+        try {
+            GitCheckout.checkout(config.getRepoRoot(), config.getBranch());
+            return true;
+        } catch (RuntimeException e) {
+            logger.log(Level.SEVERE, "Branch does not exist! Analysis terminated.", e);
+            handleCloningFailed(outputPath, config);
+        }
+        return false;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -46,34 +50,67 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
         logger.info("The report is generated at " + outputPath);
     }
 
     /**
-     * Clone, analyze and generate the report for repositories in {@code configs}.
+     * Groups {@code RepoConfiguration} with the same {@code RepoLocation} together so that they are only cloned once.
+     */
+    private static Map<RepoLocation, List<RepoConfiguration>> groupConfigsByRepoLocation(
+            List<RepoConfiguration> configs) {
+        Map<RepoLocation, List<RepoConfiguration>> repoLocationToBranch = new HashMap<>();
```
fzdy1914: The variable name seems not so suitable.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -46,34 +50,67 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
+    private static void cloneAndAnalyzeRepos(
+            Map<RepoLocation, List<RepoConfiguration>> repoLocationMap, String outputPath) throws IOException {
         RepoCloner repoCloner = new RepoCloner();
-        RepoConfiguration clonedRepo = null;
+        RepoLocation clonedRepoLocation = null;
 
-        for (RepoConfiguration config : configs) {
-            repoCloner.clone(outputPath, config);
+        for (RepoLocation location : repoLocationMap.keySet()) {
+            repoCloner.clone(outputPath, repoLocationMap.get(location).get(0));
```
fzdy1914: We already have a RepoLocation. Maybe we can use the location to clone the repo directly.

myteo: Ah, the reason I use RepoConfiguration instead of just RepoLocation is because during cloning I need some properties of RepoConfiguration, for example repoRoot folder, displayName, etc.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -57,23 +56,28 @@ public RepoConfiguration getClonedRepo(String outputPath) throws IOException {
...
 
+        previousIndex = currentIndex;
+        currentIndex = (currentIndex + 1) % configs.length;
+        return configs[previousIndex].getLocation();
+    }
+
+    /**
+     * Prepares repo specified by {@code config} to be analyzed by updating and checking out its branch.
+     */
+    public boolean prepareToAnalyze(String outputPath, RepoConfiguration config) throws IOException {
```
eugenepeh: doesn't seems to explain what does the boolean return type stand for, was it even used?

myteo: In this method, the branch of the config is checked out and the method returns false if it fails. This is so that ReportGenerator knows to skip the repo. Maybe throwing exception is the cleaner option.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -46,34 +50,75 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
+    private static void analyzeRepos(
+            String outputPath, List<RepoConfiguration> configs, RepoCloner repoCloner) throws IOException {
+
+        for (RepoConfiguration config : configs) {
+            try {
+                repoCloner.updateAndCheckoutBranch(config);
+            } catch (RuntimeException e) {
+                logger.log(Level.SEVERE, "Branch does not exist! Analysis terminated.", e);
+                Path repoReportDirectory = Paths.get(outputPath, config.getDisplayName());
+                FileUtil.createDirectory(repoReportDirectory);
```
yamidark: Would it be possible to move the creation of the folder up (think we can create a helper method for this to improve readability for this), and also do the error handling up here as well (if failed to create)? Then, we won't have to create the folder in the other method `analyzeRepo`.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -46,34 +50,75 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
+                generateEmptyRepoReport(repoReportDirectory.toString());
+                return;
+            }
+            analyzeRepo(outputPath, config);
+        }
+    }
+
     /**
      * Analyzes repo specified by {@code config} and generates the report.
      */
```
yamidark: I just noticed this, seems like `analyzeRepo` isn't catching the `RunTimeException` at the correct place, should be with the bottom chunk right? And also have to handle creating empty report for there?

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -57,38 +56,27 @@ public RepoConfiguration getClonedRepo(String outputPath) throws IOException {
...
-    public void cleanup() {
-        deleteDirectory(FileUtil.REPOS_ADDRESS);
+    public void updateAndCheckoutBranch(RepoConfiguration config) {
+        config.updateBranch(currentRepoDefaultBranch);
+        GitCheckout.checkout(config.getRepoRoot(), config.getBranch());
     }
 
     /**
-     * Returns true if current repo is different from the previously cloned repo.
+     * Cleans up after all repos have been cloned and analyzed
```
yamidark: index `.` at the end.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -38,17 +39,15 @@
...
     }
 
     /**
-     * Waits for current clone process to finish executing and returns the corresponding {@code RepoConfiguration}.
+     * Waits for current clone process to finish executing and returns the {@code RepoLocation} of the corresponding
+     * {@code RepoConfiguration}.
      */
-    public RepoConfiguration getClonedRepo(String outputPath) throws IOException {
-        if (isCurrentRepoCloned && isPreviousRepoDifferent()) {
+    public RepoLocation getClonedRepo(String outputPath) throws IOException {
```
yamidark: should be `getClonedRepoLocation` now?

src/main/java/reposense/report/ReportGenerator.java
myteo: Yeah this seems weird. I went to take a look at the history before cloning in parallel [here](https://github.com/reposense/RepoSense/blob/d426f09a3a5271d306c751803ec0d8ccabfdb87c/src/main/java/reposense/report/ReportGenerator.java).

For the bottom chunk, it seems like there wasn't any catch block for `RunTimeException` there all along. For the top chunk, the `RuntimeException` catch block was introduced in this [commit](https://github.com/reposense/RepoSense/commit/3d393c513a262d612629bf2eb6511cd04d6bad87) and so I just kept it there.

From the history, it also seems like in the case of `IOException` and `RuntimeException` we just skip the repo entirely and didn't create an empty report.

myteo: I can change it if you think its a mistake though.

src/main/java/reposense/report/ReportGenerator.java
myteo: Yup that sound better.

src/main/java/reposense/report/ReportGenerator.java
yamidark: >I can change it if you think its a mistake though.

This seems outside of the scope of this PR, so let's leave it for now. We can raise an issue and resolve this quickly if problems occur in CS2103/CS2113.

eugenepeh:
@myteo 

TODO: resolve the conflict

Issue #467, opened 2019-01-07
Give a way to add an intro to the generated report


damithc:
Suggestion: Provide a way for users to add some content at the top of the report e.g., a heading, some short intro text etc. to the report before publishing it. The content should remain through subsequent updates of the report.

PR #627, opened 2019-03-25
Model: Add ReportConfiguration

fzdy1914:
Related #590, #467 

Our user cannot edit the property of generated report.

Let's add a new model, ReportConfiguration, which serves the role to
configure the generated report.


fzdy1914:
@eugenepeh @yamidark 
As this is a totally new component, so I would like to ask for your opinion on the implementation of this pr.

For my implementation, I suppose that the report-config.json should under the same folder with the other two configs, which means, this feature only works with `ConfigCliArgument`. 

There is also another possible implementation. We add one more arg e.g. `--report-config` to specify the location of the folder and put all the file related to the report config in it.

eugenepeh:
>For my implementation, I suppose that the report-config.json should under the same folder with the other two configs, which means, this feature only works with ConfigCliArgument.

I think we can go with that for now. `--repo` is usually meant for few repositories, so its not that important to add support for it yet.

fzdy1914:
Put this PR on hold as there is no time to finish it.

fzdy1914:
Anyone who is willing to take over the PR can just go ahead.

fzdy1914:

PR #628, opened 2019-03-25
[#619] Remove Unused Class in Main Code

### LINKED ISSUE #619: Remove unused MethodInfo and ConfigurationBuilder
When running coverage for RepoSense, I found that there are two classes that are never used in our codebase, `MethodInfo` and `ConfigurationBuilder`.

If they are not in our immediate plan, we should remove them from our code.

fzdy1914:
Fixes #619 

The `MethodAnalyzer` and `MethodInfo`  classes were added to the
codebase in the early stage of RepoSense, when there was some 
hope of identifying authors of methods of Java file. The 
`ConfigurationBuilder` class was used by `GitGrader` which has been
remove from codebase and it is not used in the main code.

As these classes have not been used since early stages(v1.0), and they
are also not in our immediate plans anymore, leaving them will
unnecessarily complicate the codebase.

Let's clean up the code base by removing these unused classes.


yong24s:
From [a past comment](https://github.com/reposense/RepoSense/pull/260#issuecomment-412322145), we were looking into making `ConfigurationBuilder` more useful. 😅

PR #629, opened 2019-03-25
RepoConfigurationTest: move to reposense.model

fzdy1914:
`RepoConfigurationTest` is under package ` reposense.parser`.

The class tested is `RepoConfiguration` which is under package 
`reposense.model`. It is not reasonable to put this test under a
different package.

Let's move `RepoConfigurationTest` to `reposense.model` to 
solve the problem.


eugenepeh:
slight nit to second paragraph

```
However, the class under test is `RepoConfiguration` which is under
reposense.model package. Hence, it is not reasonable to put the test
file under a different package.
```

PR #632, opened 2019-03-27
[#621] LogsManager: shift the log file to output folder

### LINKED ISSUE #621: Shift the log file to output folder
In the event of a bug investigation, it would be more useful if the log file is tagged along with the generated report. Let's have the log file generated together in the designed output folder.

fzdy1914:
Fixes #621 

The log file of RepoSense is stored at a fixed folder `logs` in the 
current working directory.

In the event of a bug investigation, it would be more useful if the
log file is tagged along with the generated report. 

Let's have the log file generated together in the designed output
folder.

The side effect of this PR is that the log message generated during
`parse()` method in `ArgsParser` will not be recorded in the file as the 
output folder is not decided yet.


src/main/java/reposense/system/LogsManager.java
```diff
@@ -21,24 +21,30 @@
...
 
+    private static Path logFolderLocation;
     private static Level currentConsoleLogLevel = Level.INFO;
     private static Level currentFileLogLevel = Level.INFO;
     private static FileHandler fileHandler;
     private static ConsoleHandler consoleHandler;
 
     public static Logger getLogger(String name) {
         Logger logger = Logger.getLogger(name);
+        LOGGER_LIST.add(logger);
```
chelseyong: Should the adding to LOGGER_LIST be moved down, after the checking of `logFolderLocation` (line 46)?

fzdy1914: No need, logger is an object. LOGGER_LIST only stores the reference of the logger.

src/main/java/reposense/system/LogsManager.java
```diff
@@ -115,6 +122,14 @@ private static ConsoleHandler createConsoleHandler() {
         return consoleHandler;
     }
 
+    /** Set the {@code logFolderLocation } and add fileHandler
+     *  to all the {@code Logger} in {@code LOGGER_LIST}
```
yamidark: All the `{@code}` used here are all `private`, not visible outside of this class though.
Try to make the Javadoc not have any specific reference to any of those private variables.

fzdy1914:
> It seems when I run `./gradlew test`, a directory `some_non_existent_dir` gets created (which is completely empty), can help to check out whats happening?

···
    @Test
    public void outputPath_nonExistentDirectory_success() throws ParseException, IOException, HelpScreenException {
        Path nonExistentDirectory = PROJECT_DIRECTORY.resolve("some_non_existent_dir/");
        Path expectedRelativeOutputDirectoryPath = nonExistentDirectory.resolve(ArgsParser.DEFAULT_REPORT_NAME);
        String input = new InputBuilder().addOutput(nonExistentDirectory).build();
        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
        Assert.assertTrue(cliArguments instanceof ConfigCliArguments);
        Assert.assertTrue(Files.isSameFile(
                expectedRelativeOutputDirectoryPath, cliArguments.getOutputFilePath()));
    }
···
This causes the problem, however, it seems that I am unable to remove the directory during runtime becuase of .lck file. I will keep looking into it.

fzdy1914:
> This causes the problem, however, it seems that I am unable to remove the directory during runtime becuase of .lck file. I will keep looking into it.

I am able to delete this now.

fzdy1914:
@yamidark I have fixed the directory problem, can you help review again?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -57,6 +60,15 @@ public void before() {
         DEFAULT_INPUT_BUILDER.reset().addConfig(CONFIG_FOLDER_ABSOLUTE);
     }
 
+    @After
+    public void after() {
+        try {
+            FileUtil.deleteDirectory(PROJECT_DIRECTORY.resolve(NONEXISTENT_DIRECTORY).toString());
```
yamidark: Can just do the deletion of directory in the test method that creates it?

fzdy1914: Actually, we cannot. The reason is that when the log is being used, there will be a `.lck` file that prevents the log file to be deleted.

yamidark:
@fzdy1914 Sorry, I noticed in this log file created within the output folder, the same log message seems to be repeated many times. Can you look into this as well?

fzdy1914:
@yamidark This may due to you run the same command several times. Even in our previous implementation, when we have new log messages, we append the messages to the file instead of overwriting the previous log.

yamidark:
>@yamidark This may due to you run the same command several times. Even in our previous implementation, when we have new log messages, we append the messages to the file instead of overwriting the previous log.

What I meant was that there was repeated of the same messages next to each other, rather than the next running log messages appended to the previous run.


Anyway, I tried this again, don't seem to run into this issue anymore. Seems to happen during the `gradlew build` command only, so not really an issue for now.

PR #638, opened 2019-04-06
ContributionBar: length changes on breakdown option selected #637

### LINKED ISSUE #637: Contribution bar length changes on breakdown option selected
Seems like a bug introduced in #625, it only occurs after that PR

* https://deploy-preview-625--reposense.netlify.com/#search=eugene&sort=displayName&since=2017-10-09&until=2019-03-25&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=false
* https://deploy-preview-596--reposense.netlify.com/#search=eugene&sort=displayName&since=2017-10-09&until=2019-03-25&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=true

image
image

myteo:
Fixes #637

```
Contribution bar length changes when breakdown option is selected.

This happens because multi-line contribution bars are incorrectly
calculated for file formats. If a contribution bar is unable to fit
in the current line, a new line is added and the width of the new line
is subtracted from the previous contribution bar element. However, this
subtraction can lead to a negative number and it is left unchecked.
This will cause the contribution bar to be longer than expected because
the negative width was not accounted for (elements with negative widths
will just be rendered as having zero width).

Lets fix the bar width calculation by checking for negative widths and
rectifying it by removing the element with negative width and
subtracting that amount from the other remaining contribution bar
element.
```


myteo:
The reason for the bug:

In the `getFileFormatContributionBars(fileFormatContribution)` method, the widths of the respective file formats are computed. In the current logic, this can sometimes result in negative widths being computed. I added a check for such a case and rectify the width calculation.

eugenepeh:
Good catch 👍 

Can you also create a [cypress test](https://github.com/reposense/RepoSense/tree/master/frontend/cypress/tests) to ensure this functionality does not break again in the future?

myteo:
> Good catch 👍
> 
> Can you also create a [cypress test](https://github.com/reposense/RepoSense/tree/master/frontend/cypress/tests) to ensure this functionality does not break again in the future?

Yup sure

myteo:
> Can you also create a [cypress test](https://github.com/reposense/RepoSense/tree/master/frontend/cypress/tests) to ensure this functionality does not break again in the future?

I realize the bar widths might be different by a small amount due to floating point calculation.

Had trouble with Cypress as well because we have to sum up the widths of different bar elements and I couldn't find a clean way of doing that..

eugenepeh:
>I realize the bar widths before and after breakdown option is selected might be different by a small amount due to floating point calculation.

We can probably round off to a specific decimal or something.

>Had trouble with Cypress as well because we have to sum up the widths of the bar elements of the different formats to compare with the initial bar width, and I couldn't find a clean way of doing that..

Would grouping them into a div, if we haven't already, and sum the child elements help in that scenario?

myteo:
The widths might differ by 1 or more.

In Cypress, I can check the widths by retrieving the width from CSS style, and it is returned in pixels and returned as a whole number.

In our code, we calculate the bar widths for the different formats as percentages of the original bar width which will be a floating point number. So there will be some discrepancy.

myteo:
Personally I think its quite difficult to test this mainly because theres no exact correctness and also no strict error bound. For your consideration😬

eugenepeh:
Correct me if i am wrong,

All the contrib line of each ramp chart are stored as child elements in the corresponding `summary-chart__contribution` element.

Suppose we get the 2 sets of `summary-chart__contribution` in the 2 different view, we retrieve all the children of each ramp chart's contrib lines, sum up their width and compare with the corresponding ramp chart's in the other view. If the the 2 value are within 5% tolerance of each other, we can say that it is fairly accurate.

Can this not be done? If so, why?

myteo:
I wasn't so sure about the 5% tolerance.

Ok but this isn't a problem now, I realized I can get the percentage widths directly... My bad

myteo:
@reposense/devs Ready for review.

myteo:
> LGTM, add a propose message and its ready to go

Sorry missed that out. Done.

yamidark:
@myteo For your proposed message:
>This happens because contribution bar widths are incorrectly
calculated for file formats.

Could you explain more on **why** the contribution bar lengths are incorrectly calculated, and why this caused an issue for the inconsistent contribution bar length? 

>Lets fix the bar width calculation and add a Cypress test.

And also how exactly this was fixed?

myteo:
> @myteo For your proposed message:
> 
> > This happens because contribution bar widths are incorrectly
> > calculated for file formats.
> 
> Could you explain more on **why** the contribution bar lengths are incorrectly calculated, and why this caused an issue for the inconsistent contribution bar length?
> 
> > Lets fix the bar width calculation and add a Cypress test.
> 
> And also how exactly this was fixed?

Done.

PR #639, opened 2019-04-10
ChartView: Add remaining features missing from sorting of authors and repos

chelseyong:
Follow-up on #486 and #618.
```
Regarding the sorting of ramp charts, contribution and variance are
enabled only when there is no grouping.

Users may want to sort the authors or repos by contribution/variance
to look at more/less active groups at a glance.

Let's add the sorting of authors/repos by total contribution and total variance,
by collating these information when the report is loaded.
```


chelseyong:
Cypress tests are failing

frontend/src/index.jade
```diff
@@ -87,6 +87,14 @@ html
...
                 option(value="totalCommits dsc", v-if="filterGroupSelection === 'groupByNone'") &darr; contribution
+                option(value="repoCommits", v-if="filterGroupSelection === 'groupByRepos'") &uarr; repo contribution
+                option(value="repoCommits dsc", v-if="filterGroupSelection === 'groupByRepos'") &darr; repo contribution
+                option(value="repoVariance", v-if="filterGroupSelection === 'groupByRepos'") &uarr; repo variance
+                option(value="repoVariance dsc", v-if="filterGroupSelection === 'groupByRepos'") &darr; repo variance
+                option(value="authorCommits", v-if="filterGroupSelection === 'groupByAuthors'") &uarr; author contribution
+                option(value="authorCommits dsc", v-if="filterGroupSelection === 'groupByAuthors'") &darr; author contribution
+                option(value="authorVariance", v-if="filterGroupSelection === 'groupByAuthors'") &uarr; author variance
+                option(value="authorVariance dsc", v-if="filterGroupSelection === 'groupByAuthors'") &darr; author variance
```
eugenepeh: is it necessary to have `repoVariance`, `authorVariance` and `variance`? Seems like something that can be done in the backend

chelseyong:
Cypress has passed, Codacy failed on Cypress syntax

frontend/src/index.jade
```diff
@@ -85,10 +85,8 @@ html
                 option(value="name dsc", v-if="filterGroupSelection !== 'groupByRepos'") &darr; name
                 option(value="searchPath", v-if="filterGroupSelection !== 'groupByAuthors'") &uarr; repo/branch
                 option(value="searchPath dsc", v-if="filterGroupSelection !== 'groupByAuthors'") &darr; repo/branch
-                option(value="totalCommits", v-if="filterGroupSelection === 'groupByNone'") &uarr; contribution
-                option(value="totalCommits dsc", v-if="filterGroupSelection === 'groupByNone'") &darr; contribution
-                option(value="variance", v-if="filterGroupSelection === 'groupByNone'") &uarr; variance
-                option(value="variance dsc", v-if="filterGroupSelection === 'groupByNone'") &darr; variance
+                option(v-for="option in contributionSelection", v-bind:value="option.value") {{ option.text }}
+                option(v-for="option in varianceSelection", v-bind:value="option.value") {{ option.text }}
```
eugenepeh: No, what I mean was, why do you need to complex things and have different values shown (according to grouping) in the first place when its already categories under `sort within group`?

chelseyong: Do you mean that there are options to sort via contributions and variances in both `sort within group` and `sort group` ?
Contribution in `sort group` is to get all the authors/repos in the ascending order of contribution made. Same for variance.
Isn't that the whole point of this PR? :)

chelseyong: Forgot to mention that the contribution and variance values are now the total contribution/variances summed for each author/repo

eugenepeh: As in, there is no need to emphasize `author variance` vs `repo variance` when its all categories under the `sort groups by`; its already implied. It can all be named as `variance` to avoid complicating the code and confuse the user.

Less complexity also helps to reduce damage and chance of bugs occurs. E.g. occurrence of bug where group by shows `Author` and sort groups by shows `repo contribution`. It may confuse the users and they may not realize that it is a bug.

chelseyong: Okay, understood.

chelseyong:
CI is failing on java8.

yamidark:
>CI is failing on java8.

Resolved.

frontend/src/static/js/api.js
```diff
@@ -93,4 +101,26 @@ window.api = {
         });
   },
 
+  loadAuthorCommitsVariance() {
```
eugenepeh: Just to check, which/where is the repo's variant for this function?

chelseyong: It's inside `loadCommits`, lines 87-90

eugenepeh: Doesn't seems like a clean solution. E.g. the repo and author variant of calculating commits and variance are divided into 2 functions with diff name (`loadCommits` and `loadAuthorCommitsVariance`) and different variable names e.g. `User` vs `Author` which increase the amount of time needed to understand the code.

Perhaps try to extract it as a common method? E.g. `calcCommitsVariance(repos)`

frontend/src/static/js/api.js
```diff
@@ -93,4 +93,35 @@ window.api = {
         });
   },
 
+  calcCommitsVariance() {
```
eugenepeh: add some comments within this function to explain the flow

chelseyong:
@eugenepeh changes have been made.

frontend/src/static/js/api.js
```diff
@@ -93,4 +93,40 @@ window.api = {
...
+      let repoVariance = 0;
+      repos[repo].users.forEach((author) => {
+        repoCommits += author.totalCommits;
+        repoVariance += author.variance;
+        if (!Object.keys(authorCommits).includes(author)) {
+          authorCommits[author.name] = 0;
+          authorVariance[author.name] = 0;
+        }
+        // Accumulate author's commits and variance
+        authorCommits[author.name] += author.totalCommits;
```
eugenepeh: what prevents you from doing `author.authorCommits += author.totalCommits;` directly?

chelseyong: If totalCommits/variance is updated directly within the loop, the values that are changed only affects that repo, but not other repos that the author may have.

After reimplementing upon your suggestion, the sorting within groups became inconsistent, hence the bug.

eugenepeh: I see.

I am still not quite convinced by the need to have 2 separate schemes in order to do the sorting of variance and commits, it simply implies that the existing sorting function is flawed.

Perhaps the target area to solve this problem shouldn't be using `repos`, but rather the sorting of filtered in the grouping function itself?

https://github.com/reposense/RepoSense/blob/011977ffb6291fe2ac3c2fc1996a409e9b4134b3/frontend/src/static/js/v_summary.js#L487-L541

Especially this particular line (below):

https://github.com/reposense/RepoSense/blob/011977ffb6291fe2ac3c2fc1996a409e9b4134b3/frontend/src/static/js/v_summary.js#L536

chelseyong: It is possible to generate all total commits and variances from every author and repo, without `calcCommitsVariance()`. However this tabulating of values will be done every time the user sorts the ramp charts, inside the `groupBy...()` function.

Under situations where the repository is very big or there are many users, could there be a risk of slowing down the sorting process, thereby affecting user experience?

eugenepeh: The grouping and sorting is already done in the `groupBy` function anyway, I don't think even a 10000 sum operation can jeopardize the processing speed / operation.

chelseyong:
@eugenepeh I have removed authorCommits, repoCommits etc from the repos' data, and only tabulate them when the ramp charts are sorted.

frontend/src/static/js/v_summary.js
```diff
@@ -523,7 +539,12 @@ window.vSummary = {
         filtered.push(authorMap[author]);
       });
 
-      filtered.sort(window.comparator((ele) => ele[0][this.sortingOption]));
+      filtered.sort(window.comparator((ele) => {
```
eugenepeh: In both cases, filtered is a 2D array right? So why not you do something simpler like 
```
filtered.sort(window.comparator((group) => group.reduce(function(total, author) { 
  return total + author.sortingOption;
}));
```

Make it a side function so that it is neater and reusable for both cases.

frontend/src/static/js/v_summary.js
```diff
@@ -523,12 +528,21 @@ window.vSummary = {
         filtered.push(authorMap[author]);
       });
 
-      filtered.sort(window.comparator((ele) => ele[0][this.sortingOption]));
+      filtered.sort(window.comparator((author) => {
+        if (this.sortingOption === 'totalCommits' || this.sortingOption === 'variance') {
+          return author.reduce(this.groupCommitsVariance, 0);
```
ongspxm: the `groupCommitsVariance` now sounds like a property, make it more verby instead, maybe add a get in front.

eugenepeh:
Nits to proposed message:

```
Contribution and variance options are not available in `sort groups by`
when the `group by` option is selected.

Users may wish to sort the groups with this 2 options, ranking them in
terms of activity level, so that they can pay more attention to the
most/least achievers.

Let's allow the sorting of groups by these 2 options at all times.
```

Issue #640, opened 2019-04-12
CS3281 dashboard: wrong start date shown


damithc:
https://nus-cs3281.github.io/2019-dashboard
![image](https://user-images.githubusercontent.com/1673303/56010893-ffdfd180-5d17-11e9-8cf5-71c4fb1424a1.png)

Actual start date is different https://github.com/nus-cs3281/2019-dashboard/blob/master/run.sh

fzdy1914:
May be related to #603

PR #641, opened 2019-04-15
GitLog: update to use commit date instead of author date #640 

### LINKED ISSUE #603: Differ of Commit date and Author date confuses the github link in RampChart
The problem is like below:

From this [link](https://deploy-preview-29--ecstatic-goldwasser-4c5dd4.netlify.com/#search=&sort=displayName&since=2018-12-01&until=2019-03-16&timeframe=day&reverse=false&repoSort=true&breakdown=false), we are able to see 3 ramps in my contribution to checkstyle. However, only the last link opened in GitHub will show the commit info correctly.

The reason is that the commit date and the author date is not the same for my first two commits. 

See the following commit as an example:
image
image

Repo Sense caught my commit on 12/2, which is my author date. However, when using this date to filter on github, i.e. open the corresponding [link](https://github.com/checkstyle/checkstyle/commits/master?author=fzdy1914&since=2019-02-12%27T%2700:00:00+08:00&until=2019-02-12%27T%2723:59:59+08:00). Github will not return correct result.

ongspxm:
Fix #603 #640 

```
Currently we use the authored date as the date of the commit, that 
however the can cause some discrepancy when the commit is committed on 
a later date.

Let's use the commit date as the commit's date instead.
```


damithc:
So there are three dates: authored date, commit date, merge date?

ongspxm:
merge date is for merge commit only, i think its just commit date on merge commits

ongspxm:
its just commit and authored date, merge date is just the commit date of the merge commit. in this case, it was the original commit, so no merge commit created

damithc:
What's the definition of authored date? How does it differ from the commit date?

ongspxm:
https://stackoverflow.com/a/11857467 here is a discussion on the difference

damithc:
Whichever date we use, it should be used consistently across the application. The current bug indicates we use one date in some parts of the code and another date in other parts.

ongspxm:
Before that we always used the authored date. The difference only occurs
when we are rebasing, and reordering the original commit. thats why we only
notice it now, when the commit date is very far from the authored date.

if its alright, you can give it a quick look, then we can get it merged.

damithc:
> if its alright, you can give it a quick look, then we can get it merged.

I have no suggestions about the code; leaving it to the senior devs to decide if the fix is correct.

PR #656, opened 2019-05-14
[#567] Zip: prevent unrelated files and/or folders from getting zipped

### LINKED ISSUE #567: Generated archieve.zip contains non-relevant json file
As mentioned in the title, when generating the archieve.zip, we include all the json file in the output directory. However, if they are other directories that are left out in the output directory(e.g. analyze the other repos before), the json file will also be included in the archeve.zip. 

This behavior should not be the intended behavior of our project.

jylee-git:
```
Zipped files currently contain irrelevant files and/or folders. 

Some users may dump other irrelevant files/folders into 
the same directory as the directory for the repos to be 
cloned into. This results in irrelevant files and folders being 
included in the zip file, which is not intended to be.

Let's zip only the relevant repo folders and the 
corresponding .JSON files into the zip file. 

This method utilizes the existing class `RepoConfiguration` , to
get the list of repos report to be zipped.
```

Fixes #567


fzdy1914:
@jylee-git 
Please follow the following flow as specified in our [commit organization](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message) as specified in the [contribution guidelines](https://github.com/reposense/RepoSense/blob/master/docs/Process.md#how-to-contribute-to-the-reposense-repository):
```
{current situation} -- use present tense

{why it needs to change}

{what is being done about it} -- use imperative mood

{why it is done that way}

{any other relevant info}
```
Also, make each line within a word limit of 72 characters.

Meanwhile, avoid unnecessary changes to the codebase (i.e. insert unnecessary blank lines in this case).

Do request a review when the PR is ready for review.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -34,10 +37,52 @@
...
+
     private static final Logger logger = LogsManager.getLogger(FileUtil.class);
     private static final String GITHUB_API_DATE_FORMAT = "yyyy-MM-dd";
     private static final ByteBuffer buffer = ByteBuffer.allocate(1 << 11); // 2KB
 
+    /**
+     * Zips only the relevant .JSON files
+     * @param configs Utilizes the relevant repo folders that are required to be zipped
+     */
+    public static void zipRelevantJsonFiles(List<RepoConfiguration> configs, Path sourceAndOutputPath) {
```
fzdy1914: The classes in util package including FileUtil are mean to be *tool* classes, so we do not put such specific methods in these classes. You should put methods that can be reused here. 

So, maybe you can have a method named `zipRelativeFiles` which takes in `a list of relative paths` and `sourceAndOutputPath` and zip the file. Then, you can generate the list of relative paths in another place and call this new method. The new method is more general and can be served as a `tool` method.

Above is just my suggestion, the key idea is not to make the method in `Utils` too specific.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -94,8 +139,10 @@ public static void zip(Path sourcePath, Path outputPath, String... fileTypes) {
                 ZipOutputStream zos = new ZipOutputStream(fos)
         ) {
             Set<Path> allFiles = getFilePaths(sourcePath, fileTypes);
+
```
fzdy1914: Unnecessary blank line.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -94,8 +139,10 @@ public static void zip(Path sourcePath, Path outputPath, String... fileTypes) {
...
             for (Path path : allFiles) {
                 String filePath = sourcePath.relativize(path.toAbsolutePath()).toString();
+
```
fzdy1914: Unnecessary blank line.

fzdy1914:
For the title of PR, We usually put [#issue number] before the actual title to show the relationship with the issue.

src/main/java/reposense/util/FileUtil.java
jylee-git: Alright noted. Have made changes to the code base and will re-request for review soon.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +39,41 @@
...
+    public static void zipRelativeFiles(HashSet<Path> relativePaths, HashSet<Path> relevantFiles,
+                                        Path sourcePath, Path outputPath, String... fileTypes) {
+        try (
+                FileOutputStream fos = new FileOutputStream(outputPath + File.separator + ZIP_FILE);
+                ZipOutputStream zos = new ZipOutputStream(fos)
+        ) {
+            Set<Path> allFiles = getFilePaths(sourcePath, fileTypes);
+            for (Path path : allFiles) {
+                String filePath = sourcePath.relativize(path.toAbsolutePath()).toString();
+                if (relativePaths.contains(path.getParent()) || relevantFiles.contains(path.getFileName())) {
```
fzdy1914: Why this method needs both `relativePaths` and `relevantFiles`? It seems that you deal with them equally.

src/main/java/reposense/RepoSense.java
```diff
@@ -62,7 +67,11 @@ public static void main(String[] args) {
                     cliArguments.getSinceDate().orElse(null),
                     cliArguments.getUntilDate().orElse(null));
 
-            FileUtil.zip(cliArguments.getOutputFilePath().toAbsolutePath(), ".json");
+            HashSet<Path> relevantFolders = getRelativeFolders(cliArguments.getOutputFilePath(), configs);
+            HashSet<Path> relevantFiles = new HashSet<>();
+            relevantFiles.add(Paths.get(SummaryReportJson.SUMMARY_JSON_FILE_NAME));
+            FileUtil.zipRelativeFiles(relevantFolders, relevantFiles, cliArguments.getOutputFilePath().toAbsolutePath(),
```
fzdy1914: Similar to `zip()`, you can overload this method so you do not need to call `cliArguments.getOutputFilePath().toAbsolutePath()` twice.

src/main/java/reposense/RepoSense.java
```diff
@@ -76,6 +85,22 @@ public static void main(String[] args) {
         }
     }
 
+    /**
+     * Returns a list of relevant repo folders to be zipped up later
+     * @param sourcePath
+     * @param configs
+     * @return
+     */
+    private static HashSet<Path> getRelativeFolders(Path sourcePath, List<RepoConfiguration> configs) {
```
fzdy1914: I am not so sure it is a good place to put the method here.  I will leave this comment here to see whether the seniors have other suggestions.

src/main/java/reposense/util/FileUtil.java
fzdy1914: You may just add `summary.json` into the `relativePaths`.

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -5,22 +5,52 @@
...
+    private static final Path[] REPO_FOLDERS = {Paths.get("reposense_testrepo-Beta"),
+            Paths.get("reposense_testrepo-Charlie")};
+    private static final Path EXPECTED_RELEVANT_FOLDER_PATH = Paths.get(FILE_UTIL_TEST_DIRECTORY.toString(),
+            "expectedRelevantUnzippedFiles");
+
+    /**
+     * Ensures that only the specified files and folders with the corresponding file types get zipped.
+     */
+    @Test
+    public void zip_onlyRelevantFiles_success() throws IOException {
```
fzdy1914: Good job! You remember to write a unit test. It is important to ensure the code quality.

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -5,22 +5,52 @@
...
+        HashSet<Path> relevantFolders = new HashSet<>(Arrays.asList(REPO_FOLDERS));
+        HashSet<Path> relevantFiles = new HashSet<>();
+        relevantFolders.add(Paths.get(RELEVANT_REPO_DIRECTORY_PATH + File.separator
```
fzdy1914: The `relevantFolders` seems already have the addresses of these two folders?

fzdy1914: See
```
    private static final Path[] REPO_FOLDERS = {Paths.get("reposense_testrepo-Beta"),
            Paths.get("reposense_testrepo-Charlie")};
    HashSet<Path> relevantFolders = new HashSet<>(Arrays.asList(REPO_FOLDERS));
```

src/main/java/reposense/util/FileUtil.java
fzdy1914: My bad for missing `path.getParent()` in the first statement. But anyway, can you try to combine these two sets into one? Because from my perspective, it is unnecessary.

src/test/java/reposense/util/FileUtilTest.java
jylee-git: My bad; forgot to remove those

src/main/java/reposense/util/FileUtil.java
jylee-git: Noted; combined those two sets into one in my next commit.

src/main/java/reposense/RepoSense.java
fzdy1914: My rough idea is to write a new class called `ZipGenerator` to handle below sentences:
```
    HashSet<Path> relevantPaths = getRelativeFolders(cliArguments.getOutputFilePath(), configs);
    relevantPaths.add(Paths.get(cliArguments.getOutputFilePath().toAbsolutePath().toString(),
                SummaryReportJson.SUMMARY_JSON_FILE_NAME));
     FileUtil.zipRelativeFiles(relevantPaths, cliArguments.getOutputFilePath().toAbsolutePath(), ".json");
```
But let us to wait for the seniors to decide.

src/main/java/reposense/report/SummaryReportJson.java
```diff
@@ -9,6 +9,8 @@
  * Represents the structure of summary.json file in reposense-report folder.
  */
 public class SummaryReportJson {
+    public static final String SUMMARY_JSON_FILE_NAME = "summary.json";
```
fzdy1914: Is this the only appearance of "summary.json" in our codebase?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +39,49 @@
     private static final String GITHUB_API_DATE_FORMAT = "yyyy-MM-dd";
     private static final ByteBuffer buffer = ByteBuffer.allocate(1 << 11); // 2KB
 
+    /**
+     * Zips all the relative folders and relevant files of {@code fileTypes} contained in
+     * {@code sourceAndOutputPath} directory into the same folder.
+     */
+    public static void zipRelativeFiles(HashSet<Path> relativePaths,
+                                        Path sourceAndOutputPath, String... fileTypes) {
```
fzdy1914: No following our coding style.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +39,49 @@
...
+    }
+
+    /**
+     * Zips all the relevant files and relative folders
+     * @param relativePaths contains the relevant folders to be zipped.
+     * @param fileTypes contains the file types to be zipped. Only files which are of the type "fileTypes" will be
+     *                  zipped.
+     */
+    public static void zipRelativeFiles(HashSet<Path> relativePaths,
+                                        Path sourcePath, Path outputPath, String... fileTypes) {
```
fzdy1914: Same here, not following our coding style. We adopt 8 indentations from the previous line.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +39,49 @@
...
+        try (
+                FileOutputStream fos = new FileOutputStream(outputPath + File.separator + ZIP_FILE);
+                ZipOutputStream zos = new ZipOutputStream(fos)
```
fzdy1914: Weird style here too. Not sure if it is correct.

fzdy1914: Maybe our checkstyle is not checking this.

src/test/java/reposense/util/FileUtilTest.java
```diff
@@ -5,22 +5,48 @@
...
     private static final Path EXPECTED_UNZIPPED_DIRECTORY_PATH = Paths.get(FILE_UTIL_TEST_DIRECTORY.toString(),
             "expectedUnzip");
     private static final Path TEST_ZIP_PATH = Paths.get(FILE_UTIL_TEST_DIRECTORY.toString(), "testZip.zip");
     private static final Path UNZIPPED_DIRECTORY_PATH = Paths.get(FILE_UTIL_TEST_DIRECTORY.toString(),
             "UnzippedFolder");
+    private static final Path[] RELEVANT_FILE_FOLDER_PATHS = {
+            Paths.get(RELEVANT_REPO_DIRECTORY_PATH.toString(), "reposense_testrepo-Beta").toAbsolutePath(),
+            Paths.get(RELEVANT_REPO_DIRECTORY_PATH.toString(), "reposense_testrepo-Charlie").toAbsolutePath(),
+            Paths.get(RELEVANT_REPO_DIRECTORY_PATH.toString(), SummaryReportJson.SUMMARY_JSON_FILE_NAME)
+                    .toAbsolutePath()};
```
fzdy1914: The line break here. In this case, break after the `,` is better than before `.`.

src/main/java/reposense/report/SummaryReportJson.java
jylee-git: The "summary.json" is used twice: in source code `RepoSense.java` and the test code `FileUtilTest.java`

fzdy1914: I mean that, the `summary.json` here: 
https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/report/ReportGenerator.java#L238 can be replaced.

src/main/java/reposense/util/FileUtil.java
jylee-git: Aparently the checkstyle isn't picking up those, seemingly how my local checkstyle and travis dosen't respond to these issues. I assume those checkstyle can be configured to pick up these other checkstyle issues as well?

fzdy1914: Yes, maybe you can raise an issue about this. Before that, just follow the code style in your newly added part and leave the rest there.

src/main/java/reposense/report/SummaryReportJson.java
jylee-git: Oh ok noted. Will replace that with the static variable.

src/main/java/reposense/util/FileUtil.java
jylee-git: I believe this current indentation is correct (8 spaces from the previous line since it's a parentheses and not curly brackets)? The same code in the `zip` method in the master branch shows the same code

https://github.com/reposense/RepoSense/blob/81db2aaf93d06bbf19be2efc2372baa6b5106957/src/main/java/reposense/util/FileUtil.java#L91-L97

fzdy1914: The code style in the main branch may not be correct, too. I have mentioned that checkstyle may not check this part. I will do some search tonight. Before that, just leave this as it.

src/main/java/reposense/RepoSense.java
jylee-git: Alright noted

src/main/java/reposense/util/FileUtil.java
jylee-git: Ok noted

fzdy1914: This is a code use of try-with-resource and I do not find a code style from [oss-generic coding standards](https://oss-generic.github.io/process/docs/CodingStandards.html). That's what I found: [try with resources](https://stackoverflow.com/questions/30753996/what-is-the-code-convention-for-formatting-try-with-resources) basically. So, let the seniors decide which one is better.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +39,49 @@
     private static final String GITHUB_API_DATE_FORMAT = "yyyy-MM-dd";
     private static final ByteBuffer buffer = ByteBuffer.allocate(1 << 11); // 2KB
 
+    /**
```
fzdy1914: Javadoc missing the argument `relativePaths`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +39,49 @@
...
+     * Zips all the relative folders and relevant files of {@code fileTypes} contained in
+     * {@code sourceAndOutputPath} directory into the same folder.
+     */
+    public static void zipRelativeFiles(HashSet<Path> relativePaths,
+            Path sourceAndOutputPath, String... fileTypes) {
+        zipRelativeFiles(relativePaths, sourceAndOutputPath, sourceAndOutputPath, fileTypes);
+    }
+
+    /**
+     * Zips all the relevant files and relative folders
```
fzdy1914: Same here, missing `relativePaths`, `sourcePath`, `outputPath`.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +39,56 @@
...
+     * @param relativePaths contains the relevant folders to be zipped.
+     * @param sourcePath contains the directory where the relevant folders and files are located at
+     * @param outputPath contains the directory to be zipped to.
+     * @param fileTypes contains the file types to be zipped. Only files which are of the type "fileTypes" will be
+     *                  zipped.
+     */
+    public static void zipRelativeFiles(HashSet<Path> relativePaths,
```
yong24s: FileUtil#zip is similar to FileUtil#zipRelativeFiles and its not used anymore right?

src/main/java/reposense/RepoSense.java
```diff
@@ -76,6 +84,22 @@ public static void main(String[] args) {
...
```
yong24s: Can be place in `FileUtil`. Since its being used by a method in that class.

Also, these remove these trivial JavaDocs.
We don't use them like this. Use it like how you did in another method.
```
     * @param sourcePath
     * @param configs
     * @return
```

src/main/java/reposense/RepoSense.java
```diff
@@ -62,7 +67,10 @@ public static void main(String[] args) {
                     cliArguments.getSinceDate().orElse(null),
                     cliArguments.getUntilDate().orElse(null));
 
-            FileUtil.zip(cliArguments.getOutputFilePath().toAbsolutePath(), ".json");
+            HashSet<Path> relevantPaths = getRelativeFolders(cliArguments.getOutputFilePath(), configs);
+            relevantPaths.add(Paths.get(cliArguments.getOutputFilePath().toAbsolutePath().toString(),
+                    SummaryReportJson.SUMMARY_JSON_FILE_NAME));
+            FileUtil.zipRelativeFiles(relevantPaths, cliArguments.getOutputFilePath().toAbsolutePath(), ".json");
```
yong24s: Word choice suggestion. Try report folders/files?

You use both relative and relevant. Report folder/file is more consistent and it implies that they are relevant.

You may be thinking of naming this a more general method 🤔 but I think zipping is only used for report archival.

src/main/java/reposense/util/FileUtil.java
jylee-git: Yup, FuleUtil#zip method is no longer used anymore

fzdy1914: Then, we should remove this method from code base.

jylee-git: Ok, will make the changes

src/main/java/reposense/RepoSense.java
jylee-git: I renamed: -

- the method `getRelativeFolders` to `getReportFolders`,
- the method `zipRelativeFiles` to `zipReportFoldersAndFiles`,
- the variables `relevantPaths` and `repoFoldersAndFiles`.

Would these names be better?

yong24s: No need to embed source code, this should save you some time. 😃


>zipRelativeFiles to zipRelevantFoldersAndFiles

```public static void zipRelevantFoldersAndFiles(HashSet<Path> relevantPaths,```

If you specify paths (**with intention**), those are already relevant right?
Do developers need to be aware of *relevant* or *irrelevant* paths?

```public static void zipFoldersAndFiles(HashSet<Path> pathsToZip,``` may remove some confusion.

Comment (jylee-git) [reply src/main/java/reposense/RepoSense.java]: Ok noted

File: src/main/java/reposense/util/FileUtil.java [hunk 19]
```diff
@@ -38,6 +41,71 @@
...
+        }
+        return relevantFolders;
+    }
+
+    /**
+     * Zips all the relative folders and relevant files of {@code fileTypes} contained in
+     * {@code sourceAndOutputPath} directory into the same folder.
+     * @param pathsToZip contains the folders and files to be zipped.
+     * @param sourceAndOutputPath contains the directory where the source folders and files are located at and
+     *                           where to be zipped to.
```
eugenepeh: inconsistent wrapping, in fact if you can rename the variables and describe the method more accurately using `@code` in the paragraph, you probably wouldn't need to use `@param` tag

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +41,71 @@
...
+            for (Path path : allFiles) {
+                String filePath = sourcePath.relativize(path.toAbsolutePath()).toString();
+                if (pathsToZip.contains(path.getParent()) || pathsToZip.contains(path)) {
+                    String zipEntry = Files.isDirectory(path) ? filePath + File.separator : filePath;
+                    zos.putNextEntry(new ZipEntry(zipEntry.replace("\\", "/")));
+                    if (Files.isRegularFile(path)) {
+                        try (InputStream is = Files.newInputStream(path)) {
+                            int length;
+                            while ((length = is.read(buffer.array())) > 0) {
+                                zos.write(buffer.array(), 0, length);
```
eugenepeh: though its a legacy but since you're revamping it, do help to make the code more readable here, it is a rather huge [arrow head](https://blog.codinghorror.com/flattening-arrow-code/)

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +41,60 @@
...
+                    .toAbsolutePath());
+        }
+        return relevantFolders;
+    }
+
+    /**
+     * Zips all the relative folders and relevant files of {@code fileTypes} contained in
+     * {@code pathsToZip} directory into {@code sourceAndOutputPath}.
+     */
+    public static void zipFoldersAndFiles(HashSet<Path> pathsToZip,
```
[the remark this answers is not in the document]
jylee-git: Changed to interface `Set`. Was still used to the thought of strictly returning of the same exact type.. Will take note next time to return an interface whenever appropriate.

jylee-git:
@yamidark I've made the requested changes. Would appreciate if you'd continue reviewing. Thanks!

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +41,52 @@
...
     private static final ByteBuffer buffer = ByteBuffer.allocate(1 << 11); // 2KB
 
+    /**
+     * Returns a {@code set} of relevant repo paths from {@code repoConfigsList}.
+     */
+    public static Set<Path> getReportFolders(Path sourcePath, List<RepoConfiguration> repoConfigsList) {
+        Set<Path> relevantFolders = new HashSet<>();
+        for (RepoConfiguration repoConfiguration : repoConfigsList) {
+            relevantFolders.add(Paths.get(sourcePath + File.separator + repoConfiguration.getDisplayName())
+                    .toAbsolutePath());
```
eugenepeh: The readability of this line needs to improve

Either do this

```
            relevantFolders.add(
                    Paths.get(sourcePath + File.separator + repoConfiguration.getDisplayName()).toAbsolutePath());
```

or create a variable to concat and reuse the path
`sourcePath + File.separator + repoConfiguration.getDisplayName()`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +41,52 @@
     private static final String GITHUB_API_DATE_FORMAT = "yyyy-MM-dd";
     private static final ByteBuffer buffer = ByteBuffer.allocate(1 << 11); // 2KB
 
+    /**
+     * Returns a {@code set} of relevant repo paths from {@code repoConfigsList}.
```
eugenepeh: `set` dont have to highlighted with `{@code ...}` 
Also, the comment needs to improve. How do you determine `relevant`?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +41,52 @@
...
+        for (RepoConfiguration repoConfiguration : repoConfigsList) {
+            relevantFolders.add(Paths.get(sourcePath + File.separator + repoConfiguration.getDisplayName())
+                    .toAbsolutePath());
+        }
+        return relevantFolders;
+    }
+
+    /**
+     * Zips all files of type {@code fileTypes} that are in the directory {@code pathsToZip} into
+     * {@code sourceAndOutputPath}.
```
eugenepeh: into a single file and output it to `{@code sourceAndOutputPath}`.

src/main/java/reposense/RepoSense.java
```diff
@@ -62,7 +66,11 @@ public static void main(String[] args) {
                     cliArguments.getSinceDate().orElse(null),
                     cliArguments.getUntilDate().orElse(null));
 
-            FileUtil.zip(cliArguments.getOutputFilePath().toAbsolutePath(), ".json");
+            Set<Path> repoFoldersAndFiles = FileUtil.getReportFolders(cliArguments.getOutputFilePath(), configs);
```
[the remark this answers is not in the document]
jylee-git: The method only returns the set of folders, but I named the variable as `repoFoldersAndFiles` as the file `summary.json` will be added in as well at the next line of code.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -38,6 +41,52 @@
...
+    /**
+     * Returns a list of relevant repo paths from {@code repoConfigsList}.
+     */
+    public static Set<Path> getReportFolders(Path sourcePath, List<RepoConfiguration> repoConfigsList) {
+        Set<Path> relevantFolders = new HashSet<>();
+        for (RepoConfiguration repoConfiguration : repoConfigsList) {
+            relevantFolders.add(Paths.get(sourcePath + File.separator + repoConfiguration.getDisplayName())
+                    .toAbsolutePath());
+        }
+        return relevantFolders;
```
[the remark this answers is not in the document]
jylee-git: No duplicates. I used set due to the previous implementation of the code that will impede performance if list is used instead of set, which has already been re-implemented.

Have changed it to use list instead.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -41,6 +43,52 @@
...
+    /**
+     * Zips all files of type {@code fileTypes} that are in the directory {@code pathsToZip} into a single file and
+     * output it to {@code sourceAndOutputPath}.
+     */
+    public static void zipFoldersAndFiles(List<Path> pathsToZip, Path sourceAndOutputPath, String... fileTypes) {
+        zipFoldersAndFiles(pathsToZip, sourceAndOutputPath, sourceAndOutputPath, fileTypes);
+    }
+
+    /**
+     * Zips all files of type {@code fileTypes} that are in the directory {@code pathsToZip} into {@code outputPath}.
```
eugenepeh: `in the directory {@code pathsToZip}` seems to be an incorrect description.
The files are technically in `sourcePath`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -41,6 +43,52 @@
 
     private static final String MESSAGE_INVALID_FILE_PATH = "\"%s\" is an invalid file path. Skipping this directory.";
 
+    /**
+     * Returns a list of paths to the generated reports from {@code repoConfigs}.
```
eugenepeh: Do include `sourcePath` in your description. The comment is not really describing what it is doing.

In fact, it is possible to do without this method. Since these path were created in `ReportGenerator.generateReposReport`, you can simply tweak it to return the paths.

jylee-git: Integrated the method into `ReportGenerator#generateReposReport`.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -64,13 +65,13 @@
     private static Date earliestSinceDate = null;
 
     /**
-     * Generates the authorship and commits JSON file for each repo in {@code configs} at {@code outputPath}, as
-     * well as the summary JSON file of all the repos.
+     * Generates and returns a list of paths to the authorship and commits JSON file for each repo in {@code configs}
+     * at {@code outputPath}, as well as the summary JSON file of all the repos.
```
eugenepeh: A forcefully condensed message adds more confusion and is not advised.
Retain the original message and add another line `@return the list of filepaths that were generated` instead.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -83,9 +84,17 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
         FileUtil.writeJsonFile(
-                new SummaryReportJson(configs, generationDate, reportSinceDate, untilDate, RepoSense.getVersion()),
+                new SummaryJson(configs, generationDate, reportSinceDate, untilDate, RepoSense.getVersion()),
                 getSummaryResultPath(outputPath));
         logger.info(String.format(MESSAGE_REPORT_GENERATED, outputPath));
+
+        List<Path> reportFoldersAndFiles = new ArrayList<>();
+        for (RepoConfiguration config : configs) {
+            reportFoldersAndFiles.add(
+                    Paths.get(outputPath + File.separator + config.getDisplayName()).toAbsolutePath());
```
eugenepeh: Still not really the ideal solution because this assumes that the files were generated which might not be the case, but we can settle that in another issue. 

The ideal solution should the methods, that generate the file e.g. FileUtil.writeJsonFile, return the path of the files that were successfully generated. Only then, `generateReposReport` should consolidate them one by one and return it to the caller.

Can you please open an issue to target this problem?

jylee-git: Opened up a new issue on this.

PR #665, opened 2019-05-16
[#661] Tweak chart title format

### LINKED ISSUE #661: Tweak chart title format
current: `org_repo_branch`
image

suggested: `org/repo[branch]`

reason: the org/repo/branch names might have underscores themselves.

fzdy1914:
Fixes #661 
```
We use underscore to contact the org, repo, branch to be the title of
ramp chart.(i.e. org_repo_branch)

The org/repo/branch names might have underscores themselves, which leads
to confusions when trying to distinguish actual org/repo/branch names.

Let's tweak the format to be org/repo[branch] so it is clearer.
```

### DIFF
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -66,7 +66,7 @@ public class RepoConfiguration {
         String repoName = location.getRepoName();
 
         if (organization != null) {
-            displayName = organization + "/" + repoName + "[" + branch + "]";
+            displayName = organization + "_" + repoName + "_" + branch;
             repoFolderName = organization + "_" + repoName;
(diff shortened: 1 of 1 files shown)


fzdy1914:
I have tried the backend approach, however, the following error occurs.
image
So, I will use a front-end approach. But it may not be ideal. Because resource folders are still named with underscores.

damithc:
I'm not sure if using folder names to encode org/repo/branch names is foolproof because the characters allowed in folder names can vary based on the OS and can differ from what's allowed for org/repo/branch names. Does the current code works if the org/repo/branch name has underscores?

fzdy1914:
> Does the current code works if the org/repo/branch name has underscores?

I will have a try.

PR #668, opened 2019-05-17
[#647] Toolbar: rename values in `sort group by`

### LINKED ISSUE #647: Report#Toolbar: rename values in `sort groups by`
image

We can simply use `name`. There isnt a need to further emphasize it `repo/branch`.

jamessspanggg:
Fixes #647 

```
The `sort groups by` selection contains the option of `repo/branch` and
`name` when charts are grouped by Repo/Branch and Author respectively.

As it is a `sort group by` selection, there's is no need to use two
different options `repo/branch` and `name` as they both refer to the
`group name`

Let's rename both values to `group name` to standardize the values.
```


chelseyong:
It can get quite confusing when this happens:
image
image

When grouped by `author`, `repo/branch` still appears in `sort within group`, but not when grouped by `repo`.
Maybe stick to one for all groupings?

Also, when `grouped by none`, there seems to be duplicates of `name` sorting.

jamessspanggg:
> It can get quite confusing when this happens:
> image
> image
> 
> When grouped by `author`, `repo/branch` still appears in `sort within group`, but not when grouped by `repo`.
> Maybe stick to one for all groupings?
> 
> Also, when `grouped by none`, there seems to be duplicates of `name` sorting.

Shall we stick with `name` for all groupings?

damithc:
> Shall we stick with `name` for all groupings?

or `group name`?

Also, we can use `author` instead of `name` to mean to refer to the author name?

damithc:
UI looks OK.

eugenepeh:
if name is not a good choice, can rename both to `alphabetical`

eugenepeh:
@damithc, do you think `alphabetical`/`lexicographical` is a better replacement?
I think using a standardize name for all sorting controls, as well as in the same order, would be easier for users to get accustomed to it, especially when we have them in code view as well.

damithc:
> @damithc, do you think `alphabetical`/`lexicographical` is a better replacement?

I feel `... name` is safer because `name` usually implies _alphabetical_ but `alphabetical` doesn't imply _name_ . Put it in another way, `name`, `contribution`, `variance` etc. are attributes used to sort, whereas `alphabetical` is how the attribute is used in sorting. In theory, it is possible to sort by `contribution`, lexicographically.  Does that make sense?

eugenepeh:
true, that's a good point.

frontend/src/index.pug
```diff
@@ -90,19 +90,23 @@ html
...
-                option(value="name", v-if="filterGroupSelection !== 'groupByRepos'") &uarr; name
-                option(value="name dsc", v-if="filterGroupSelection !== 'groupByRepos'") &darr; name
-                option(value="searchPath", v-if="filterGroupSelection !== 'groupByAuthors'") &uarr; repo/branch
-                option(value="searchPath dsc", v-if="filterGroupSelection !== 'groupByAuthors'") &darr; repo/branch
+                option(value="name", v-if="filterGroupSelection === 'groupByAuthors'") &uarr; group name
+                option(value="name dsc", v-if="filterGroupSelection === 'groupByAuthors'") &darr; group name
+                option(value="searchPath", v-if="filterGroupSelection === 'groupByRepos'") &uarr; group name
+                option(value="searchPath dsc", v-if="filterGroupSelection === 'groupByRepos'") &darr; group name
+                option(value="name", v-if="filterGroupSelection === 'groupByNone'") &uarr; author
+                option(value="name dsc", v-if="filterGroupSelection === 'groupByNone'") &darr; author
```
eugenepeh: author -> author name

damithc:
> If we change from one to other in the `group by`, you will notice that if your selection is on `variance` or `contribution` in `sort within groups by`, the option wont reset. However if its on `Author` or `Repo/Branch`, it will resets back to ascending order (if its on descending order).

From my understanding, this resetting is due to a different reason: i.e., it doesn't make sense to have the same value in both dropdowns e.g., `group-by: author` and `sort-within-group-by: author`. Therefore, when the user chooses `group-by: author`, the entire `author` option is removed from the `sort-within-group-by` dropdown, not just changing from `descending` to `ascending`.

Is that the resetting you are referring to, or perhaps I understood wrongly?

eugenepeh:
>Is that the resetting you are referring to, or perhaps I understood wrongly?

Yes. Which is why if we use a universal name (without changing any of the elements) the list won't change.

E.g. Notice how changing the value of `group by` from `Repo\Branch` to `Author` doesn't reset the `sort within groups by` because that particular list item is not modified
image
image

But if the `sort within groups by` is on descending `author`, changing the value of `group by` will resets the `sort within groups by`, which also change the original descending to ascending order even though by the previous logic it should retain its last index.
image
image

This is one of the bug we're currently having and a simple fix for it would be to not modify the list and use a neutral word that represents `Repo\Branch` and `Author` for the `sort within groups by`.

damithc:
> This is one of the bug we're currently having and a simple fix for it would be to not modify the list and use a neutral word that represents `Repo\Branch` and `Author` for the `sort within groups by`.

I see. `chart title`?

jamessspanggg:
> > Is that the resetting you are referring to, or perhaps I understood wrongly?
> 
> Yes. Which is why if we use a universal name (without changing any of the elements) the list won't change.
> 
> E.g. Notice how changing the value of `group by` from `Repo\Branch` to `Author` doesn't reset the `sort within groups by` because that particular list item is not modified
> image
> image
> 
> But if the `sort within groups by` is on descending `author`, changing the value of `group by` will resets the `sort within groups by`, which also change the original descending to ascending order even though by the previous logic it should retain its last index.
> image
> image
> 
> This is one of the bug we're currently having and a simple fix for it would be to not modify the list and use a neutral word that represents `Repo\Branch` and `Author` for the `sort within groups by`.

I have added an extra check for descending values to fix that bug. This should not happen now.

eugenepeh:
>I have added an extra check for descending values to fix that bug. This should not happen now.

Yes @jamessspanggg that solves the problem "temporaily". In fact your work actually serves as a better illustration on why we should use a common identifier.

For e.g., this entire chunk of code shouldn't be there in the first place. There is no reason to have distinction between group by repo, author, none etc... All that is there to do is giving a set of items, and arrange them accordingly by the given comparator.
image

In my opinion, we should stop feeding the code that should be eventually removed; it would be easier to rectify an early mistake.

Right now, if we really want to (and if its worth to) support 2 transition names, we should either go with using index identifier. Or we can just apply a common and ambiguous name such as `group title` and `title` for the `sort group by` and `sort within groups by` respectively.

jamessspanggg:
> In my opinion, we should stop feeding the code that should be eventually removed; it would be easier to rectify an early mistake.
> 
> Right now, if we really want to (and if its worth to) support 2 transition names, we should either go with using index identifier. Or we can just apply a common and ambiguous name such as `group title` and `title` for the `sort group by` and `sort within groups by` respectively.

Noted. Thanks for the advice.

jamessspanggg:
> Right now, if we really want to (and if its worth to) support 2 transition names, we should either go with using index identifier. Or we can just apply a common and ambiguous name such as `group title` and `title` for the `sort group by` and `sort within groups by` respectively.

I have now used a common name `group title` for  `sort group by` and `title` for `sort within groups by`. 

I have also disabled `sort within groups by` when `group by none` is selected. (previously, N/A was displayed)

Now all `group by repo, author, none` have common identifiers.

PR #669, opened 2019-05-17
[#635] CodeView: add line number to code

### LINKED ISSUE #635: Add line number to code view
It is hard to associate a specific line from code view with the actual code without the line number.

E.g. to instruct anyone on a specific change from the review of the code view, we can only provide the corresponding code. It may takes the author / other reviewer a while to find the specific line, especially if its has multiple duplication.

jamessspanggg:
Fixes #635 

```
The code view displays only the code without the line number.

It is hard to associate a specific line from the code view with the
actual code without the line number.

Let's update the code view to display the line numbers by adding an
array in each segment to keep track of the line numbers of the code.
```

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -287,8 +287,8 @@ html
                         span.tooltiptext.tooltiptext--close Show untouched code
                         span.tooltiptext.tooltiptext--open Hide untouched code
                       .code(v-for="(line, index) in segment.lines")
-                        div.line-number {{ segment.lineNumbers[index] + "\n" }}
-                        div.line-content {{ line + "\n" }}
+                        .line-number {{ segment.lineNumbers[index] + "\n" }}
+                        .line-content {{ line + "\n" }}
         .empty(v-else) loading...
 
     script(src="static/js/api.js")
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -286,7 +286,9 @@ html
                         i.fas.fa-ellipsis-h.icon
                         span.tooltiptext.tooltiptext--close Show untouched code
                         span.tooltiptext.tooltiptext--open Hide untouched code
-                      .code {{ segment.lines.join("\n") }}
+                      .code(v-for="(line, index) in segment.lines")
+                        div.line-number {{ segment.lineNumbers[index] + "\n" }}
+                        div.line-content {{ line + "\n" }}
         .empty(v-else) loading...
 
     script(src="static/js/api.js")
--- a/frontend/src/static/css/style.scss
+++ b/frontend/src/static/css/style.scss
@@ -426,9 +426,9 @@ header{
         }
 
         .line-number{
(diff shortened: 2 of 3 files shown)


frontend/src/static/js/v_authorship.js
```diff
@@ -128,6 +128,7 @@ window.vAuthorship = {
       let lastId = -1;
       const segments = [];
       let blankLineCount = 0;
+      let lineCount = 0;
 
       lines.forEach((line) => {
```
ongspxm: forEach can be used with an index, use that instead

yong24s:
@jamessspanggg 

The line number shouldn't start from 0 right?

frontend/src/index.jade
```diff
@@ -286,7 +286,9 @@ html
                         i.fas.fa-ellipsis-h.icon
                         span.tooltiptext.tooltiptext--close Show untouched code
                         span.tooltiptext.tooltiptext--open Hide untouched code
-                      .code {{ segment.lines.join("\n") }}
+                      .code(v-for="(line, index) in segment.lines")
+                        .line-number {{ segment.lineNumbers[index] + "\n" }}
+                        .line-content {{ line + "\n" }}
```
eugenepeh: hmm... this `lineNumbers` array shouldn't be necessary if jade can perform this functionality of pug.
https://pugjs.org/language/iteration.html#while

Can you please try it out?

jamessspanggg: yep am on it.

jamessspanggg: > hmm... this `lineNumbers` array shouldn't be necessary if jade can perform this functionality of pug.
> https://pugjs.org/language/iteration.html#while
> 
> Can you please try it out?

I think the `lineNumbers` array is necessary as there are un-authored and authored segments. We wouldn't know at which line the authored segments start without the array since there are un-authored segments which are not displayed unless you toggle them.

frontend/src/static/css/style.scss
```diff
@@ -425,6 +425,31 @@ header{
             background-color:mui-color('green', '50');
         }
 
+        .line-number{
+            color:mui-color("grey");
```
eugenepeh: please follow existing style to use single quote for string literals

frontend/src/static/css/style.scss
```diff
@@ -425,6 +425,31 @@ header{
...
+            float: left;
+            width: 2rem;
```
eugenepeh: there shouldn't be space after colon

jamessspanggg:
> minor nits

Done fixing the code styles.

eugenepeh:
Nits to the last paragraph:

Let's update the code view to display the line number`s` ~that associates to the specific
line in the actual code~ by ~having~ `adding` an array in each segment to keep~s~ track of ~the~ 
line numbers ~corresponding to the lines of~ `of the` code.

Don't need to repeat the previous paragraph, it is implied

PR #670, opened 2019-05-17
[#660] ChartView: remove ramp slices not within date range

### LINKED ISSUE #660: Ramp slices not within date range are still shown
Screen Shot 2019-05-16 at 4 20 34 PM

After adjusting the dates, some ramp slices that are no longer within the date range are still shown.

jamessspanggg:
Fixes #660 

```
After adjusting the date range, ramp slices that are not within the 
date range still shows in the ramp row.

Users would only want to see the ramp slices that are within the date 
range.

Let's remove the ramp slices that are not within date range by 
checking if the position of the ramp slices are within the width of 
the ramp row (if it is not within the width of the ramp row, the ramp
slice is not within date range) before adding the ramp slices to the 
ramp row.
```


fzdy1914:
Sorry for my undetailed feedback. The approach you take is at rendering time, which does not fix the inner cause of the problem.

To find the inner cause of the problem, you need to look at the source code detailly.

The first problem to figure out is, what does the date shown in slice mean when granularity is set to be `week`.
image

See: https://github.com/reposense/RepoSense/blob/master/frontend/src/static/js/v_summary.js#L373-L388. It actually represents the start date of the week. 

I agree with that, this representation is vague and should be improved in the future. But we should not just ignore the slice. We need to find out why this is being included in the commit list.

See: https://github.com/reposense/RepoSense/blob/master/frontend/src/static/js/v_summary.js#L393-L400. This code seems to have no problem and should work fine.

Then, we want to have a look at how `startOfWeekMs` field is generated. We will found that it is related to a function, `dateRounding`,https://github.com/reposense/RepoSense/blob/master/frontend/src/static/js/v_summary.js#L373.

After taking a look at this function, https://github.com/reposense/RepoSense/blob/master/frontend/src/static/js/v_summary.js#L18-L30. We have actually located the source of the problem. We set `roundDown` to be 1. So the date generated representing the week is actually before the since date.

fzdy1914:
After finding the inner cause of the problem, then we can try to find out how to fix it. Following are some possible solution. 
1, Instead of using startDate of a week to represent the whole week, we use the endDate of a week.
2, Instead of roundDown, we use roundUp. 
3, ...
However, these approaches may cause some other regressions to happen and we need to figure out how to avoid them.

So, the key point is to find the inner cause of the problem and fix the bug at the root.

jamessspanggg:
> After finding the inner cause of the problem, then we can try to find out how to fix it. Following are some possible solution.
> 1, Instead of using startDate of a week to represent the whole week, we use the endDate of a week.
> 2, Instead of roundDown, we use roundUp.
> 3, ...
> However, these approaches may cause some other regressions to happen and we need to figure out how to avoid them.
> 
> So, the key point is to find the inner cause of the problem and fix the bug at the root.

Noted. Thanks for the very detailed explanation. I really do appreciate that.

I was thinking of the following solution:
roundUp the startDate. Have an extra check whether the filterSinceDate is equals to the startDate. If they are not equal, then I will add commits from the filterSinceDate until the startDate. Then I will have the for-loop of adding all the commits from the startDate until the untilDate.

Would this solution be viable?

fzdy1914:
This is a workable solution. 

Can you also edit how we present the Weekly contribution? Instead of using the start date to present the commits, use a period, (i.e. 2018-10-7 - 2018-10-14). If not doing so, we should state clearly in UG that, we use the start date of a week to represent a week.

jamessspanggg:
> This is a workable solution.
> 
> Can you also edit how we present the Weekly contribution? Instead of using the start date to present the commits, use a period, (i.e. 2018-10-7 - 2018-10-14). If not doing so, we should state clearly in UG that, we use the start date of a week to represent a week.

I think the first option may be clearer for the users.

I have tried the following by stating the start date (Sunday) until the end of the week (Saturday). Would this be alright?
Screen Shot 2019-05-19 at 6 02 37 PM

fzdy1914:
> I have tried the following by stating the start date (Sunday) until the end of the week (Saturday). Would this be alright?

Do remember to consider the special case, (`filterSinceDate` to `sinceDate`). Go ahead.

frontend/src/static/js/v_summary.js
```diff
@@ -385,23 +397,23 @@ window.vSummary = {
           insertions: 0,
           deletions: 0,
           date: getDateStr(startOfWeekMs),
+          endDate: getDateStr(startOfWeekMs + durationMs - DAY_IN_MS),
```
fzdy1914: Is it worthwhile to use endDate for all week contributes just for one special case?

jamessspanggg: Since for ramps, we are now displaying the period of the week instead of only the start date, I guess it would be easier to have an endDate property?

frontend/src/static/js/v_summary.js
```diff
@@ -385,23 +397,23 @@ window.vSummary = {
...
         };
 
         // commits are not contiguous, meaning there are gaps of days without
         // commits, so we are going to check each commit's date and make sure
         // it is within the duration of a week
         while (commits.length > 0
-            && (new Date(commits[0].date)).getTime() < startOfWeekMs + WEEK_IN_MS) {
+        && (new Date(commits[0].date)).getTime() < startOfWeekMs + durationMs) {
```
eugenepeh: indentation

frontend/src/static/js/v_summary.js
```diff
@@ -385,23 +397,23 @@ window.vSummary = {
...
           week.insertions += commit.insertions;
           week.deletions += commit.deletions;
         }
 
         res.push(week);
       }
-
-      user.commits = res;
     },
+
```
eugenepeh: this newline is not necessary considering the existing style of the file

damithc:
> Looks good to me, however do check with @damithc if he prefers the week to always end on Sunday or should the start of the week always be from the since date

Shall we use Monday as the start of the week?
image

jamessspanggg:
> Shall we use Monday as the start of the week?

Sure.

fzdy1914:
Can the Codacy issue be fixed?

PR #676, opened 2019-05-17
[#667] Report: fix search bar not displaying value provided by url parameter

### LINKED ISSUE #667: Encoding of search field in URL is not complete
We have encoded the search field in URL. 

However, when we open such a URL directly, it displays nothing in the search field, which is quite confusing.

Sample: [https://nus-cs3281.github.io/2019-dashboard/#search=jacob&sort=searchPath&sortWithin=name&since=2018-12-01&until=2019-05-14&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=false](https://nus-cs3281.github.io/2019-dashboard/#search=jacob&sort=searchPath&sortWithin=name&since=2018-12-01&until=2019-05-14&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=false)

We can see that `search=jacob` is encoded in the link. But when we open it, the search field is empty.

image

jylee-git:
```
Search bar component is not displaying the value that is provided 
by the URL parameter.

This confuses users into thinking that only the listed authors and 
repos are all that's captured by the system, when there are others 
which are filtered out.

Let's fix this bug by adding a two-way binding that binds the search 
field and the corresponding JavaScript filterSearch variable.
```

Fixes #667

### DIFF
--- a/frontend/src/index.jade
+++ b/frontend/src/index.jade
@@ -73,7 +73,7 @@ html
         form.summary-picker.mui-form--inline(v-on:submit.prevent="getFiltered();")
           .summary-picker__section
             .mui-textfield
-              input.summary-picker__search(type="text", v-on:change="updateFilterSearch")
+              input.summary-picker__search(type="text", v-on:change="updateFilterSearch", v-model="filterSearch")
               label search
             .mui-select.grouping
               select(v-model="filterGroupSelection")


yong24s:
Also, move the issue number behind your PR title.

`[#667] Fixed bug ...` -> `Fixed bug ... #667` 

Then make your PR title shorter too or it will get truncated off 😁.

fzdy1914:
> Also, move the issue number behind your PR title.
> 
> `[#667] Fixed bug ...` -> `Fixed bug ... #667`

I believe now we put the issue number before the PR title. You can see the commit in master.
image

fzdy1914:
The tool is available [here](https://chrome.google.com/webstore/detail/vuejs-devtools/nhdogjmejiglipccpnnnanhbledajbpd)

jylee-git:
> LGTM.
> 
> Nits on proposed message.
> 
> For the first paragraph, try simplifying by only describing the problem itself.
> This helps us (and you) to understand the issue.
> Optionally (maybe time consuming), you may find want to find how we call it `https://github.com/reposense/RepoSense/pull/160`
> 
> ```
> The report view state is not properly restored as the search field is empty.
> ```
> 
> For the second paragraph,
> 
> This confuses users into thinking that only the listed authors and
> repos are all that's captured by the system, when there are others
> which are filtered out ~(as set in the URL).Users will then need to type something in the search field, press `ENTER`, clear the search field, and press `ENTER` again, in order to see the unfiltered list of authors and repos~.
> 
> You describe the issue well. I don't think it is necessary to mention how the users would resolve the issue. 🤔
> 
> For third paragraph,
> Lets -> Let's
> 
> Tip: Just make sure the third paragraph directly address to the issue. If you are doing more, then your PR is too "big" as it is fixing things outside of your issue.
> 
> Other tips:
> 
> * Practice using imperative mood in your commit message.
>   i.e. Fixed bug where encoded search… -> Fix bug...
> * Practice using issue number in naming your branch.
>   i.e. Fix-Name-not-encoded-from-URL-to-input-search-text-box -> 667-fix-search...

Noted. Thanks for guiding me through!

eugenepeh:
Title needs to be more critical; the `bug` can be omitted as it is implied when you use the word `fix`.

jylee-git:
> Title needs to be more critical; the `bug` can be omitted as it is implied when you use the word `fix`.
> Also `view state` is too abstract, can simply mention the component that is not working as intended.
> 
> Do reword your first paragraph accordingly.
> Good work on implementing a simple and effective fix.

Noted and thanks for the feedback!

PR #679, opened 2019-05-17
[#678] Report: use development version of vue.js

### LINKED ISSUE #678: Use vue.js instead of vue.min.js to allow using Vue developer tool
Vue has provided a plugin for chrome and firefox to help debug website using Vue.

Our dashboard cannot use this developer tool because it uses the compressed version for `vue.js`, which is `vue.min.js`.

Let's use vue.js instead of vue.min.js to allow using Vue developer tool. 

Sample tool usage:
image

fzdy1914:
Fixes #678 

```
Vue has a plugin for Chrome and Firefox for debugging Vue.js
applications. However, the minified version of Vue in our frontend is a
production version, so we are unable to use the plugin to debug
RepoSense report.

Let's use the development version of Vue.js to allow the use of Vue
developer tool.
```


yong24s:
@ongspxm 

Is there any reason why we used the production version of vue?

ongspxm:
Its the standard to do so, using production. Plus, if you work on vue long
enough, the developer tool might not really be helpful, especially for our
report which is rather huge, so the tool may be laggy.

I havent really had to use the tool for anything yet, but if any of the
developers find that it helps, i have no objections, hahah.

fzdy1914:
> Plus, if you work on vue long enough, the developer tool might not really be helpful.

I agree with that. But as a newcomer of vue, it is better to have a tool that can help us see the data within vue component.

yong24s:
Let's use it to speed up our development.

If there's a performance penalty, we can use production or development version of Vue depending on how the report is generated, by Jar or Gradlew respectively.

eugenepeh:
>Is there any reason why we used the production version of vue?

The rationale is because no point making consumer download unnecessary data; the minified version is the same, just without the readability. Not sure why the plugin wouldn't work with the minified version, but we can compromise for now since the impact is small.

yong24s:
@eugenepeh 

The minified version is a production version, so development features are stripped off. 🤔

eugenepeh:
>The minified version is a production version, so development features are stripped off. 🤔

That shouldn't be the case, for development version, there should be an equivalent minified version as well. Technically, minified should only contains the pure compressed correspondence and not any additional feature 🤔

Seems like it works differently for Vue because it has its own build tool https://vuejs.org/v2/guide/installation.html#Development-vs-Production-Mode

@ongspxm does spuild possess this functionality?

ongspxm:
Cos we only using vue in browser, we dun actually compile vue ourselves,
instead we use the distributed js files.

eugenepeh:
@fzdy1914 can add this development tool under the setting up as well? As optional.

fzdy1914:
@eugenepeh @yong24s I have added the setup link, please review.

docs/DeveloperGuide.md
```diff
@@ -80,6 +81,17 @@ We use [Cypress](https://www.cypress.io/) for automated end-to-end front-end tes
 1.  At project root, run `gradlew frontendTest`
 > If you encountered an invalid browser error, please ensure that you have `Chrome` installed in the default installation directory. Otherwise, follow the instructions [here](https://docs.cypress.io/guides/guides/debugging.html#Launching-browsers) to create symbolic links so Cypress can locate `Chrome` in your system.
 
+### (Optional) Using Vue.js devtools for frontend debugging
```
eugenepeh: on Chrome

docs/DeveloperGuide.md
```diff
@@ -80,6 +81,17 @@ We use [Cypress](https://www.cypress.io/) for automated end-to-end front-end tes
...
+1. View and download [Vue.js devtools](https://chrome.google.com/webstore/detail/vuejs-devtools/nhdogjmejiglipccpnnnanhbledajbpd) extention on Chrome web store.
```
eugenepeh: exten~t~`s`ion 
~View and download~ Using Chrome, visit the website ... and add the extension to your Chrome.

docs/DeveloperGuide.md
```diff
@@ -80,6 +81,17 @@ We use [Cypress](https://www.cypress.io/) for automated end-to-end front-end tes
...
+1. Go the detail page of this extension in Chrome's extension management panel and tick `Allow access to file URLs`. If you are unable to locate it, paste and go to the link: `chrome://extensions/?id=nhdogjmejiglipccpnnnanhbledajbpd` in the address bar of chrome.
```
eugenepeh: ~tick~ `select`
~paste and go to~ **copy** the link ... and visit it on your Chrome.

yong24s:
I think its clearer to mention the `production` and `development` version of Vue.js instead of the `minified`

```
Vue has a plugin for Chrome and Firefox for debugging Vue.js
applications. However, the minified version of Vue in our frontend is a
production version, so we are unable to use the plugin to debug
RepoSense report.

Let's use the development version of Vue.js to allow the use of Vue
developer tool.
```

PR #684, opened 2019-05-21
[#672] Toolbar: set 'commit' as default for granularity

### LINKED ISSUE #672: Use 'commit' as default value of granularity
Current: `Day` is the default value for granularity
image

Suggested: use `Commit` instead

Reason: Commit view is closer to reality. The user can choose a more summarized view at her discretion.

jamessspanggg:
Fixes #672 

```
The default value of granularity is `Day`.

`Commit` as granularity is more closer to reality because users are more likely to 
observe per commits rather than per day/week. Users can choose a more 
summarized view at their discretion.

Let's change the default value for granularity to `Commit`.
```

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -49,7 +49,7 @@ window.vSummary = {
       isSortingDsc: '',
       sortingWithinOption: '',
       isSortingWithinDsc: '',
-      filterTimeFrame: 'day',
+      filterTimeFrame: 'commit',
       filterBreakdown: false,
       tmpFilterSinceDate: '',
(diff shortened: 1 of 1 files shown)


fzdy1914:
The netlify preview is down. Can you re-trigger the build?

frontend/src/static/js/v_summary.js
```diff
@@ -432,7 +432,7 @@ window.vSummary = {
     },
     updateSortSelection() {
       this.getOptionWithOrder();
-      // Update UI selection to change all illegal options
+      // Update UI selection to change all the illegal options
```
fzdy1914: No need for this extra line.

PR #685, opened 2019-05-21
[#664] Toolbar: expand width for `breakdown by file format`

### LINKED ISSUE #664: Expand width for `breakdown by file format`
The width of `breakdown by file format` is restricted so it will be displayed in two lines even there is enough room for it to display in one line.

Let's expand the width for it.

image

jamessspanggg:
Fixes #664 

```
The width of label `breakdown by fileformat` is set to 110px which
displays in two lines even though there are enough room for it to
display in a single line.

This will cause the label to not look as organised as compared to the
label being displayed in a single line.

Let's expand the width for `breakdown by file format` by setting the
width as fit-content to ensure that the label is displayed in a single
line.
```


frontend/src/static/css/style.scss
```diff
@@ -174,7 +174,8 @@ header{
     label{
         text-align:left;
         overflow-y:hidden;
-        width:110px;
+        width:fit-content;
+        font-size: 0.75rem;
```
yong24s: Extra space :-P

`font-size: 0.75rem;` -> `font-size:0.75rem;`

jamessspanggg: opps my bad

frontend/src/static/css/style.scss
```diff
@@ -279,8 +280,8 @@ header{
 /* Authorship */
 #tab-authorship{
     .toolbar{
-        float: right;
-        cursor: pointer;
+        float:right;
+        cursor:pointer;
```
eugenepeh: avoid changing lines that are not part of your PR next time,
it complicate things in the event that PR needs to be reverted.

jamessspanggg: noted

PR #686, opened 2019-05-21
[#673] Modify slice url to commit details page when granularity is `Commit`

### LINKED ISSUE #673: Open the GitHub page with commit details when a ramp is clicked
Current: when a ramp is clicked, RepoSense opens the GitHub page with a list of commits
image

Suggested: when a ramp represents only one commit (for example, when the granularity of the view is `Commit`), it makes sense to open the github page showing the details of the page, thus saving one click.

image

jamessspanggg:
Fixes #673 

```
When value of granularity is `Commit`, the slice's url points to the GitHub page 
showing a list of commits even though there is only one commit.

Users will need the extra effort to click on the commit in the Github page and 
navigate to the commit details page.

Let's update the slice's url to point directly to the commit details page to save
an extra click from the user.
```


frontend/src/static/js/v_summary.js
```diff
@@ -166,11 +166,13 @@ window.vSummary = {
 
       return `http://github.com/${
         REPOS[user.repoId].location.organization}/${
-        REPOS[user.repoId].location.repoName}/commits/${
-        REPOS[user.repoId].branch}?`
-                + `author=${user.name}&`
-                + `since=${slice.date}'T'00:00:00+08:00&`
-                + `until=${untilDate}'T'23:59:59+08:00`;
+        REPOS[user.repoId].location.repoName}${
```
fzdy1914: Really weird indentation here.

fzdy1914: Can you make it consistent?

jamessspanggg: I have updated the code due to an error where commits of the same date points to the same commit page link.

PR #687, opened 2019-05-21
[#683] Default "since date" and "until date" to 1/1/1970 and today's date respectively

### LINKED ISSUE #683: "Until date" defaults to year 1970 and "Since date" defaults to year 2050
There are 2 bugs with the similar behavior. 

1. In CLI, if only since date is specified where the date is after the last commit date, the until date will be defaut to the year 1970.
e.g.: `java -jar RepoSense.jar -r https://github.com/reposense/RepoSense.git -s 21/04/2019 -v`. The last commit in this case is on the 20/04/2019.
image

2. Similarly, if only the until date is specified where the until date is before the first commit date, the since date will be default to the year 2050.
e.g.: `java -jar RepoSense.jar -r https://github.com/reposense/RepoSense.git -u 01/01/2015 -v`.
The first commit date in this case is on 3 May 2018.
image

I believe this issue is related to #677 . Therefore, my suggestion is to make the following changes: -

- Set the upper-bound of "Until date" to today's date, rather than the last commit date
- Remove the lower-bound of "Since date", which is currently set to the first commit date
- Use a checker, such that when parsing CLI arguments, ensure that the "since date" and "until date" is no later than today's date. Return an error if so.

jylee-git:
```
Setting "since date" to be after the latest commit date and omitting 
"until date" in CLI input defaults the "until date" to 1970. Also, 
setting "until date" to be before the first commit date and omitting 
"since date" in CLI input defaults the "since date" to 2050.

This results in users being unable to change the "since date" and 
"until date" in the browser to the desired dates.

Let's set the default "since date" and "until date" to be 1/1/1970 and 
today's date respectively should the user specify only one date and 
not the other.
```

Resolves #683

### DIFF
--- a/src/main/java/reposense/commits/CommitResultAggregator.java
+++ b/src/main/java/reposense/commits/CommitResultAggregator.java
@@ -129,9 +129,7 @@ public class CommitResultAggregator {
     }
 
     private static Date getStartDate(List<CommitResult> commitInfos) {
-        Calendar cal = Calendar.getInstance();
-        cal.set(Calendar.YEAR, 2050);
-        Date min = cal.getTime();
+        Date min = new Date(1970);
         if (!commitInfos.isEmpty()) {
             min = commitInfos.get(0).getTime();
         }
@@ -139,9 +137,7 @@ public class CommitResultAggregator {
     }
 
     private static Date getUntilDate(List<CommitResult> commitInfos) {
-        Calendar cal = Calendar.getInstance();
-        cal.set(Calendar.YEAR, 1970);
-        Date max = cal.getTime();
+        Date max = new Date();
         if (!commitInfos.isEmpty()) {
             max = commitInfos.get(commitInfos.size() - 1).getTime();
         }
--- a/src/main/java/reposense/parser/ArgsParser.java
+++ b/src/main/java/reposense/parser/ArgsParser.java
@@ -2,7 +2,6 @@ package reposense.parser;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.time.LocalDate;
 import java.time.ZoneId;
 import java.util.Date;
 import java.util.List;
--- a/src/main/java/reposense/parser/ArgsParser.java
+++ b/src/main/java/reposense/parser/ArgsParser.java
@@ -2,6 +2,7 @@ package reposense.parser;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.LocalDate;
 import java.time.ZoneId;
 import java.util.Date;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -214,4 +217,26 @@ private static void verifyDatesRangeIsCorrect(Optional<Date> sinceDate, Optional
...
+
+    /**
+     * Verifies that {@code untilDate} is no later than today's date.
+     * @throws ParseException if {@code untilDate} is later than today's date.
+     */
+    private static void verifyUntilDateIsValid(Optional<Date> untilDate) throws ParseException {
+        long todayDate = new Date().getTime();
+        if (untilDate.isPresent() && untilDate.get().getTime() > todayDate) {
+            throw new ParseException(String.format(MESSAGE_DATE_OUT_OF_BOUND, "Until date"));
+        }
```
yong24s: Why is `verifyUntilDateIsValid` the same as `verifySinceDateIsValid`?

damithc:
> Reason: Configurations can be done ahead of time. A possible use case is to monitor CS2103 repos until week 12.

Yes, one should be able to give dates in future.

jylee-git:
> > Reason: Configurations can be done ahead of time. A possible use case is to monitor CS2103 repos until week 12.
> 
> Yes, one should be able to give dates in future.

Does it mean that I (as the user) can set since and until date to future dates beyond today's date? If so, I understand the use case here. 

But what if the user were to set a "since date" to a future date but without an "until date"? When should we default the "until date" to? 

Currently, if the user were to set the "since date" to a future date, the "until date" will be default to 1970, which is earlier than the "since date". So this PR is trying to address this issue.

src/main/java/reposense/parser/ArgsParser.java
jylee-git: My bad here. My thought was to make two methods with the same functionality but different logging messages. I should have made a single common method here to verify if the dates are after today's date.

But I'll remove these two methods as we are taking into account where users will want to set the configs to capture the repos ahead of time.

damithc:
> But what if the user were to set a "since date" to a future date but without an "until date"? When should we default the "until date" to?

since_date + 30 days?

yong24s:
> since_date + 30 days?

@jylee-git 

Give it a try and see if this creates a large whitespace in the rampchart.

fzdy1914:
Can this be addressed together with https://github.com/reposense/RepoSense/pull/690?

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -129,19 +129,15 @@ private static Date getStartOfDate(Date current) {
     }
 
     private static Date getStartDate(List<CommitResult> commitInfos) {
-        Calendar cal = Calendar.getInstance();
-        cal.set(Calendar.YEAR, 2050);
-        Date min = cal.getTime();
+        Date min = new Date(1970);
```
yong24s: The since date will be defaulted to 1970 if it is set to a future date right?

Not sure if this value is useful. Try comparing it with today in the report?

jylee-git: The "since date" will be defaulted to 1970 if user sets "until date" to any date before the first commit

jylee-git:
> Can this be addressed together with #690?

I think it's possible, and it's better to solve these two issues together. Should I close this PR and address everything in #690 ?

yong24s:
> I think it's possible, and it's better to solve these two issues together. Should I close this PR and address everything in #690 ?

Don't close PRs. We want to keep the conversation/comments together in a PR, so it can be traced.

You shouldn't worry about merging PRs unless they are trivial. Also #690 requirements are still being discussed, it may become a much larger PR.

yong24s:
@jylee-git 

I previously commented not close PR as I thought you would open a new PR addressing to the same issue with the same code.

Its fine to close this PR then now #690 is merged.

PR #690, opened 2019-05-22
[#688] Limit range of commits to one month if "since date" is not specified

### LINKED ISSUE #683: "Until date" defaults to year 1970 and "Since date" defaults to year 2050

### LINKED ISSUE #688: Limiting range of commits captured by the program to one month
As discussed with @damithc , when parsing in the CLI arguments: -

- If the user did not specify both "since date" and "until date", then commits ranging one month from today's date till now, will be captured by the program.
- If the user specified the "since date" but not the "until date", then commits from "since date" all the way till now will be captured by the program (which is currently the case)
- If user specified the "until date" but not the "since date", then only the commits that were made one month from today's date will be captured by the program.
- If the user specified both the "since date" and the "until date", then commits ranging from "since date" up to "until date" will be captured by the program (as usual).

This is to ensure consistent performance, especially when the user wants to capture large repos.

jylee-git:
```
"since date", when not specified, defaults to the date of the first
commit. 

Analyzing repos with large number of commits can take up a
long time, which may leave a bad impression for first time users.

Let's,
* limit the range of commits to be captured to one month, if
"since date" is not specified.
* allow `--since *` as a shorthand to specify the date of the first
commit without knowing the actual date.
```

Resolves #683 
Resolves #688 

Related PR #687


fzdy1914:
This PR need to be stalled until @eugenepeh add `--until` flag in Netlify preview.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -75,11 +76,23 @@ public RepoConfiguration(RepoLocation location, String branch, List<Format> form
     }
 
     public static void setDatesToRepoConfigs(
```
fzdy1914: Under the situation now, do we really need a `sinceDate` and `untilDate` in each `RepoConfiguration`?

jylee-git: I believe it's required. When extracting commit information (in line 33): -

https://github.com/reposense/RepoSense/blob/27655c76427fd97a00b3ee2debcd348c8b94b047/src/main/java/reposense/commits/CommitInfoExtractor.java#L22-L33

`git log` will be called, which takes in `since date` and `until date` for each repo configurations: -
https://github.com/reposense/RepoSense/blob/27655c76427fd97a00b3ee2debcd348c8b94b047/src/main/java/reposense/git/GitLog.java#L17-L30

If `since date` or `until date` not provided, the first commit date and last commit date will be set respectively, which is what this PR is trying to address.

eugenepeh:
There's a problem we need to address prior to this PR though. What if the user wants to capture the entire range of the repo (not knowing the exact date range)? How can we still provide such functionality?

damithc:
> There's a problem we need to address prior to this PR though. What if the user wants to capture the entire range of the repo (not knowing the exact date range)? How can we still provide such functionality?

That's a good point.  The reason I thought of using 30 days as the default is so that the report can be generated faster, thus making a good impression. Also, to avoid cases of a repo with a long history creating a bad impression because the report contains more data than it can show in a decent way,

Perhaps introduce a special way for this case. e.g., if the parameter is provided without a value, we take it as 'since the beginning' but if the parameter is missing altogether, we show last 30 days. Does that make sense?

jylee-git:
> > There's a problem we need to address prior to this PR though. What if the user wants to capture the entire range of the repo (not knowing the exact date range)? How can we still provide such functionality?
> 
> That's a good point. The reason I thought of using 30 days as the default is so that the report can be generated faster, thus making a good impression. Also, to avoid cases of a repo with a long history creating a bad impression because the report contains more data than it can show in a decent way,
> 
> Perhaps introduce a special way for this case. e.g., if the parameter is provided without a value, we take it as 'since the beginning' but if the parameter is missing altogether, we show last 30 days. Does that make sense?

Maybe: -

- if the input is `java -jar ... -s -u -v` OR `java -jar ... -s -v` (with -s parameter but without value), then all commits made on any date will be captured
- if the input is `java -jar ... -s 1/1/2000 -u -v` OR `java -jar ... -s 1/1/2000 -v`, then any commits made from 1/1/2020 will be captured
- if the input is `java -jar ... -s -u 1/1/2020 -v`, then commits from the date of first commit all the way till 1/1/2020 will be captured
- if the input is `java -jar ... -u 1/1/2020 -v`, then commits within the range of 1 month before 1/1/2020 will be captured
- if the input is `java -jar ... -v` (no since&until date provided), then commits within the range of 1 month before today's date will be captured

How does that sound?

src/main/java/reposense/model/RepoConfiguration.java
fzdy1914: What I mean previously is to use cliSince and cliUntil directly when analyzing the repo. Because now, all the repoconfig will have same since and same until date. However, as the discussion below, we may still need them to be separated. So I will resolve the conversation.

eugenepeh:
>How does that sound?

Sounds good to me, but I think previously there's a limitation doing that with the argsparser; it can't tell if the flag is there or not if the value given is empty. 

If that is still the case, perhaps we should use another flag for it, e.g. `-sb` - since beginning?

fzdy1914:
> If that is still the case, perhaps we should use another flag for it, e.g. `-sb` - since beginning?

To follow our flag standard, let's use `--since-beginning` and `-S` for short.

yong24s:
@damithc 

I don't think the one month limit is universal.

Say you may have repos are not updated in the last month, which result in empty rampchart (may be undesirable) or one that too frequently updated, which is slow to generate. 

A better metric would be to use the last N commits, which we can explore and have different level of granularity.

jylee-git:
> @damithc
> 
> I don't think the one month limit is universal.
> 
> Say you may have repos are not updated in the last month, which result in empty rampchart (may be undesirable) or one that too frequently updated, which is slow to generate.
> 
> A better metric would be to use the last N commits, which we can explore and have different level of granularity.

I think N commits may have it's drawbacks also. Some people might make many commits consisting of teeny-tiny changes or "fix travis/checkstyle", whereas others may only make commits that are "big bang", if we were to only take the last N commits as the metric

eugenepeh:
>To follow our flag standard, let's use --since-beginning and -S for short.

Using capital flag is fine, however do note that it still belongs to the `since` family. If our `since` occupies both the lower and upper case of the `-s` flag, it will be a problem if we want to implement another functionality that begins with `s`, hence it's not encouraged.

eugenepeh:
Regarding the effectiveness of this measure, my input for this is

### In the worst case scenario: 
**You won't see any visible difference.** This does not apply to only the new repos, but to repos with long history like teammates as well. 

Learning point here for the newcomers, note that we have 2 different operation here: commit and authorship. Commit is fast because it runs somewhat like a single file analysis. Just because the file is big with like 100,000 commits, it is not likely to slow down the operation tremendously compare to a 1000 commits. 

Authorship on the other hand runs multi files analysis. The multi files refer to the files that the repo contains. This is a much slower operation than commit because repo almost always have lots of files that each can be heavier than commit itself.

In our code, we perform analysis on files that are touched within the given date range. This means that IF all the files were touched in the 1 month time, all the files will be marked for analysis, regardless of the number of lines have changed. Hence there will be no improvement in the timing of analysis in such scenario.

For a guaranteed improvement in time, can consider bringing back multithreading.

damithc:
Indeed, both N days or N commits have certain drawbacks. Let's go with 30 days as that would give a reasonable result in _most_ cases. We shouldn't spend too much time on this as it is not how RepoSense is supposed to be used in normal cases.

As @yong24s mentioned, we can provide a separate option to specify a number of days (but probably not commits? -- commit counting is a bit odd when there are multiple repos involved). That option, coupled with travis automation, could be useful when a user wants to keep monitoring a moving window of last N days.

Should we use `--since *` (or something like that) to indicate _since the beginning_?

yong24s:
@jylee-git 

Sorry, one more comment for this PR. 😁

Let's retain the original test files. Modify the existing methods `ConfigSystemTest#testDateRange` and `ConfigSystemTest#testNoDateRange` to test since the beginning, you may also want to rename them.

Then create a new test for just the 30 days.

jylee-git:
I made changes to how the dates are handled: -

- `... -s * -u * ...` or `... -s * ...` : Commits from the first commit date to last commit date will be captured
- `... -s 1/1/2018 -u * ...` or `... -s 1/1/2018 ...` : Commits from 1/1/2018 to last commit date will be captured
- `... -u * ...` or `<no dates provided>` : Commits one month from date of report generation will be captured
- `... -u 1/1/2019 ...` : Commits one month before 1/1/2019 will be captured

jylee-git:
Ready for review

jylee-git:
Ready for review

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -130,18 +130,25 @@ private static Date getStartOfDate(Date current) {
...
-        Date max = cal.getTime();
+    private static Date getUntilDate(RepoConfiguration config, List<CommitResult> commitInfos) {
+        Date max;
+        if (config.getSinceDate() == null) {
+            max = new Date();
+        } else {
+            Calendar cal = Calendar.getInstance();
+            cal.setTime(config.getSinceDate());
+            cal.add(Calendar.MONTH, 1);
+            max = cal.getTime();
```
eugenepeh: why until is based on since date when its the opposite logic in the beginning?

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -75,11 +77,25 @@ public RepoConfiguration(RepoLocation location, String branch, List<Format> form
...
+        // set sinceDate to one month before untilDate if sinceDate is null.
+        if (sinceDate == null) {
+            Calendar cal = Calendar.getInstance();
+            cal.setTime(untilDate);
+            cal.add(Calendar.MONTH, -1);
+            sinceDate = cal.getTime();
+            logger.info(String.format(" As \"since date\" is not specified, only commits that were made form %s to %s "
+                    + "will be captured by the program.", sinceDate.toString(), untilDate.toString()));
+        } else if (sinceDate.equals(SinceDateArgumentType.ARBITRARY_SINCE_DATE)) {
+            sinceDate = null;
```
eugenepeh: the logic is questionable here, where you try to set a value to sinceDate when its null and try to set it null when it contains a specific value.

jylee-git: I can understand that the logic here is misleading. While I think of ways to restructure the code and the logic, these were my thoughts and approach when writing this code: -

> If the user were to specify `-s *`, the since date will ultimately be set to the first commit date. However, we have no idea when the first commit date is going to be. So my thought here was to set an arbitrary since date first, then set back to null.
> 
> Then, when the user does not specify the since date, the since date will ultimately be set to one month before until date. But this depends on the until date, and we don't know the until date until the until date has been parsed and processed. Hence I initially set sinceDate to null, and until I have processed the untilDate, then i set sinceDate to be one month before the untilDate.

jylee-git:
> I think you need to reconstruct some of your logic for this PR, such that sinceDate would never be null in the later part of the flow.

Noted. I'll think and refactor the code to see how the flow can be improved.

src/main/java/reposense/model/RepoConfiguration.java
eugenepeh: Yes logic is sound and the code works, not saying its bad, but rather it would be best if it can be more readable and less redundant code.

E.g. in the above scenario, the null should be assigned to what would've be null in the first place.
Also, in this PR, since the since date can be determined in the early stage of the program, we should avoid propagating the problem to the later part of the program.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -77,8 +80,17 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
         Map<RepoLocation, List<RepoConfiguration>> repoLocationMap = groupConfigsByRepoLocation(configs);
         cloneAndAnalyzeRepos(repoLocationMap, outputPath);
 
-        Date sinceDate = cliSinceDate == null ? earliestSinceDate : cliSinceDate;
-        Date untilDate = cliUntilDate == null ? latestUntilDate : cliUntilDate;
+        Date sinceDate = (cliSinceDate.isPresent()) ? cliSinceDate.get() : null;
```
fzdy1914: Can use `Option.getOrElse(value)`

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -77,8 +80,17 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
+        Date untilDate = (cliUntilDate.isPresent()) ? cliUntilDate.get() : new Date();
+
+        if (sinceDate == null) { // set sinceDate to one month before untilDate if sinceDate is null.
```
fzdy1914: This part is duplicated with the part in `RepoConfiguration.java`. Can this part be improved? i.e. Calculated the correct `since date` in `ArgsParser`?

jylee-git: https://github.com/reposense/RepoSense/blob/b72388535cc1a533bbc45e6fd23bd95a13a5cc5b/src/main/java/reposense/parser/ArgsParser.java#L104-L108

Should we `.setDefault(...)` to one month before report generation date here?

If so, should we also `.setDefault(...)` for untilDate in `ArgsParser` to the date of report generation as well?

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -203,15 +209,54 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
...
             throw new ParseException(MESSAGE_SINCE_DATE_LATER_THAN_UNTIL_DATE);
         }
     }
+
+    /**
+     * Verifies that {@code sinceDate} is no later than the date of report generation.
+     *
+     * @throws  ParseException if {@code sinceDate} supplied is later than date of report generation.
+     */
+    private static void verifySinceDateIsValid(Date sinceDate) throws ParseException {
```
fzdy1914: I believe this can be verified in `SinceDateArgumentType` you created, so you don't need an extra method to do this.

src/main/java/reposense/report/ReportGenerator.java
fzdy1914: I think it is a good idea to do so because it is a job of `ArgsParser` to get the final/real argument. The rest only need to respond to the date argument given.

src/main/java/reposense/parser/ArgsParser.java
jylee-git: Thought of doing that, but I'm not sure if doing so will violate LSP

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -77,11 +78,11 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
         Map<RepoLocation, List<RepoConfiguration>> repoLocationMap = groupConfigsByRepoLocation(configs);
         cloneAndAnalyzeRepos(repoLocationMap, outputPath);
 
-        Date sinceDate = cliSinceDate == null ? earliestSinceDate : cliSinceDate;
-        Date untilDate = cliUntilDate == null ? latestUntilDate : cliUntilDate;
+        Date reportSinceDate = (cliSinceDate.equals(SinceDateArgumentType.ARBITARY_FIRST_COMMIT_DATE))
+                ? earliestSinceDate : cliSinceDate;
```
eugenepeh: would this conditional ever be true?
Also seems like the variable on like 76 is unused.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -574,4 +573,30 @@ public void parse_timezoneWithoutArgument_throwsParseException() throws ParseExc
...
+     */
+    private Date getExpectedSinceDate(Date untilDate) {
+        Calendar cal = Calendar.getInstance();
+        cal.setTime(untilDate);
+        cal.set(Calendar.HOUR_OF_DAY, 0);
+        cal.set(Calendar.MINUTE, 0);
+        cal.set(Calendar.SECOND, 0);
+        cal.set(Calendar.MILLISECOND, 0);
+        cal.add(Calendar.MONTH, -1);
+        return cal.getTime();
```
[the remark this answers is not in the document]
jylee-git: But the solutions in that link do not explain how to set the time to midnight or 2359. So far this is the easiest that I can find to set the time to midnight or 23:59

src/main/java/reposense/report/ReportGenerator.java
jylee-git: Yes, if user sets `-s *`, `ARBITARY_FIRST_COMMIT_DATE` will be set to `cliSinceDate` first. Then after the repos are analyzed, `reportSinceDate` will be set to `earliestSinceDate`.

Removed the unused variable and methods.

src/test/java/reposense/parser/ArgsParserTest.java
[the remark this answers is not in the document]
eugenepeh: >But the solutions in that link do not explain how to set the time to midnight or 2359

I see... that's indeed a valid argument as we should validate the time as well. Can disregard my comment for this.

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -136,7 +136,7 @@ private static Date getStartOfDate(Date current) {
     }
 
     private static Date getStartDate(List<CommitResult> commitInfos) {
-        Date min = new GregorianCalendar(1970, 1, 1).getTime();
+        Date min = new GregorianCalendar(1970, Calendar.JANUARY, 1).getTime();
```
eugenepeh: Do also make the other `GregorianCalendar` implementation to use this method `Date date = new Date(Long.MIN_VALUE);`.

jylee-git: Made the changes

src/main/java/reposense/parser/ArgsParser.java
fzdy1914: I don't think LSP is something we must follow, we are already violating it in some sense (Like the `convert` in `SinceDateArgumentType` is overriding a non-abstract method). So, I still believe it is better to put it in `SinceDateArgumentType`. 
Your inputs? @eugenepeh @yong24s

eugenepeh:
Do help me investigate this issue, for some reason, the colors for the format icons disappeared in your PR

image

jylee-git:
> Do help me investigate this issue, for some reason, the colors for the format icons disappeared in your PR
> 
> image

Can tell me the steps to replicate this problem?

eugenepeh:
>Can tell me the steps to replicate this problem?

Enable the `breakdown by file format` alone should be sufficient

jylee-git:
> > Can tell me the steps to replicate this problem?
> 
> Enable the `breakdown by file format` alone should be sufficient

image


I can't seem to replicate the problem from here. You want to do a `gradlew run` again and try?

yong24s:
> I can't seem to replicate the problem from here. 

Use [netlify preview](https://deploy-preview-690--reposense.netlify.com/#search=&sort=groupTitle&sortWithin=title&since=2019-05-26&until=2019-06-26&timeframe=commit&groupSelect=groupByRepos&breakdown=true)? 


I have the same issue on Chrome/Firefox.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -203,15 +209,54 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
...
-        if (sinceDate.isPresent() && untilDate.isPresent() && sinceDate.get().getTime() > untilDate.get().getTime()) {
+        if (sinceDate.getTime() > untilDate.getTime()) {
             throw new ParseException(MESSAGE_SINCE_DATE_LATER_THAN_UNTIL_DATE);
         }
     }
+
+    /**
+     * Verifies that {@code sinceDate} is no later than the date of report generation.
+     *
+     * @throws  ParseException if {@code sinceDate} supplied is later than date of report generation.
```
yong24s: `@throws  ParseException`
Extra space within

src/main/java/reposense/parser/SinceDateArgumentType.java
```diff
@@ -0,0 +1,31 @@
...
+     */
+    public static final Date ARBITARY_FIRST_COMMIT_DATE = new Date(Long.MIN_VALUE);
+
+    /**
+     * Returns an arbitrary year 1/1/1 if user specifies * in {@code value}, or attempts to return
+     * the desired date otherwise.
+     */
+    @Override
+    public Optional<Date> convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
+        if ("*".equals(value)) {
```
yong24s: Not sure if this is a standard practice.

`value.equals("*")`

fzdy1914:
image

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -53,6 +54,8 @@
             "\"Since Date\" cannot be later than \"Until Date\"";
     private static final String MESSAGE_USING_DEFAULT_CONFIG_PATH =
             "Config path not provided, using current working directory as default.";
+    private static final String MESSAGE_SINCE_DATE_LATER_THAN_TODAY_DATE =
```
yong24s: Move it directly below `MESSAGE_SINCE_DATE_LATER_THAN_UNTIL_DATE `?

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -164,15 +167,18 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
             Path configFolderPath = results.get(CONFIG_FLAGS[0]);
             Path reportFolderPath = results.get(VIEW_FLAGS[0]);
             Path outputFolderPath = results.get(OUTPUT_FLAGS[0]);
-            Optional<Date> sinceDate = results.get(SINCE_FLAGS[0]);
-            Optional<Date> untilDate = results.get(UNTIL_FLAGS[0]);
+            Optional<Date> cliSinceDate = results.get(SINCE_FLAGS[0]);
+            Optional<Date> cliUntilDate = results.get(UNTIL_FLAGS[0]);
+            Date sinceDate = cliSinceDate.orElse(getSinceDate(cliUntilDate));
+            Date untilDate = cliUntilDate.orElse(getUntilDate());
```
yong24s: Why not do it in `SinceDateArgumentType`? 

If we don't use `Optional` anymore just change `SinceDateArgumentType` to return `Date`?

src/main/java/reposense/parser/SinceDateArgumentType.java
jylee-git: Codacy will complain about coding standard if i do it the other way

jylee-git:
This is interesting... I still can't replicate from here 
https://jylee-reposense-deploy-test.netlify.com/#search=&sort=groupTitle&sortWithin=title&since=2019-05-26&until=2019-06-26&timeframe=commit&groupSelect=groupByRepos&breakdown=false

Will get the others to try tomorrow and see how to proceed from there.

fzdy1914:
> This is interesting... I still can't replicate from here
> https://jylee-reposense-deploy-test.netlify.com/#search=&sort=groupTitle&sortWithin=title&since=2019-05-26&until=2019-06-26&timeframe=commit&groupSelect=groupByRepos&breakdown=true
> 
> Will get the others to try tomorrow and see how to proceed from there.

Use this [url](https://deploy-preview-690--reposense.netlify.com/). The official preview we provided.

src/main/java/reposense/parser/ArgsParser.java
jylee-git: Because if user did not specify `sinceDate`, we will need to rely on `untilDate` to generate the `sinceDate` that is one month before `untilDate`. For that to happen, `cliSinceDate` needs to be initially empty. 

I can't directly get the `sinceDate` from the `convert` method in `SinceDateArgumentType`.

jylee-git:
> > This is interesting... I still can't replicate from here
> > https://jylee-reposense-deploy-test.netlify.com/#search=&sort=groupTitle&sortWithin=title&since=2019-05-26&until=2019-06-26&timeframe=commit&groupSelect=groupByRepos&breakdown=true
> > Will get the others to try tomorrow and see how to proceed from there.
> 
> Use this [url](https://deploy-preview-690--reposense.netlify.com/). The official preview we provided.
> 
> Can check on this repo:
> image
> It is not displaying any info, too. Something goes wrong when analyzing this repo.

Thanks for this. Will find out the cause of this.

jylee-git:
Apparently for me, this problem is happening in the release `v1.9` and  `master branch` as well, if you run in the master branch with the following arguments: -

-  `-c ./ -s 1/6/2019 -u 27/6/2019 -v` (using the default repo-config and author-config csv files)
- `-r https://github.com/reposense/testrepo-Alpha.git https://github.com/reposense/testrepo-Delta.git -s 1/2/2019 -u 25/6/2019 -v`

Can someone help me verify if this is the case?

yong24s:
> Can someone help me verify if this is the case?

Verified not working on master using your parameters.

src/main/java/reposense/parser/ArgsParser.java
yong24s: > Because if user did not specify `sinceDate`, we will need to rely on `untilDate` to generate the `sinceDate` that is one month before `untilDate`. For that to happen, `cliSinceDate` needs to be initially empty.

Okay, since there is a dependency, we can only handle it after the inputs are parsed.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -203,15 +209,54 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
         }
     }
 
+    /**
+     * Returns a since date that is one month before {@code cliUntilDate} (if present) or one month before report
+     *  generation date otherwise.
+     */
+    private static Date getSinceDate(Optional<Date> cliUntilDate) {
```
yong24s: Rename this function to something it does. i.e. getDateAMonthBefore/Earlier/MinusAMonth?

Reason: You got too many `getSinceDate` references.

Also, extra space within  `*  generation`.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -203,15 +209,54 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
...
+        cal.set(Calendar.SECOND, 0);
+        cal.set(Calendar.MILLISECOND, 0);
+        cal.add(Calendar.MONTH, -1);
+        return cal.getTime();
+    }
+
+    /**
+     * Returns date of report generation with time set to 23:59:59.
+     */
+    private static Date getUntilDate() {
```
yong24s: Same for this. 
Rename it to better explain what it does.
Helps with code readability.

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -41,26 +41,42 @@ public void setUp() throws IOException {
...
     public void testDateRange() throws IOException, URISyntaxException, ParseException, HelpScreenException {
         generateReport(getInputWithDates("1/9/2017", "30/10/2017"));
         Path actualFiles = Paths.get(getClass().getClassLoader().getResource("dateRange/expected").toURI());
         verifyAllJson(actualFiles, FT_TEMP_DIR);
     }
 
-    private String getInputWithDates(String sinceDate, String untilDate) {
-        return String.format("--since %s --until %s", sinceDate, untilDate);
+    @Test
+    public void test30DaysFromUntilDate() throws URISyntaxException, HelpScreenException, ParseException, IOException {
```
yong24s: Can move this up before `testDateRange`?

Now report looks really empty.

yong24s:
@eugenepeh 

Can add a since date to netlify run Dargs?

LGTM on my side.

src/main/java/reposense/parser/SinceDateArgumentType.java
damithc: For you own knowledge, find out why coday recommends one over the other.

src/main/java/reposense/parser/SinceDateArgumentType.java
```diff
@@ -0,0 +1,31 @@
...
+            return Optional.of(ARBITARY_FIRST_COMMIT_DATE);
+        }
+        return super.convert(parser, arg, value);
+    }
```
damithc: You can use the ternary operator here.
`return ... ? ... : ...;`

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -27,9 +28,14 @@
...
     public static CommitContributionSummary aggregateCommitResults(
             RepoConfiguration config, List<CommitResult> commitResults) {
-        Date startDate = config.getSinceDate() == null ? getStartDate(commitResults) : config.getSinceDate();
+        Date startDate;
+        if (config.getSinceDate() == null
+                || config.getSinceDate().equals(SinceDateArgumentType.ARBITARY_FIRST_COMMIT_DATE)) {
+            startDate = getStartDate(commitResults);
+        } else {
+            startDate = config.getSinceDate();
+        }
```
damithc: Use ternary operator?
```java
boolean someMeaningfulName = config.getSinceDate() == null
                || config.getSinceDate().equals(SinceDateArgumentType.ARBITARY_FIRST_COMMIT_DATE);
startDate = ... ? ... : ... ;
```

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -203,15 +209,54 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
...
+     * Returns date of report generation with time set to 23:59:59.
+     */
+    private static Date getReportGenerationDate() {
+        Calendar cal = Calendar.getInstance();
+        cal.set(Calendar.HOUR_OF_DAY, 23);
+        cal.set(Calendar.MINUTE, 59);
+        cal.set(Calendar.SECOND, 59);
+        cal.set(Calendar.MILLISECOND, 0);
+        return cal.getTime();
+    }
```
damithc: This is actually, `getCurrentTime` as the method doesn't do anything related to the report.

To consider in future: This kind of utility methods can be moved to a utility class so that they can be reused from elsewhere.

src/main/java/reposense/parser/SinceDateArgumentType.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+public class SinceDateArgumentType extends DateArgumentType {
+    /*
+     * When user specifies *, arbitrary first commit date will be returned.
+     * Then, ReportGenerator will replace the arbitrary since date with the earliest commit date.
+     */
+    public static final Date ARBITARY_FIRST_COMMIT_DATE = new Date(Long.MIN_VALUE);
+
+    /**
+     * Returns an arbitrary year 1/1/1 if user specifies * in {@code value}, or attempts to return
```
damithc: Magic number in comment. `Returns {@link ARBITARY_FIRST_COMMIT_DATE} if ...`?

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -27,9 +28,10 @@
      */
     public static CommitContributionSummary aggregateCommitResults(
             RepoConfiguration config, List<CommitResult> commitResults) {
-        Date startDate = config.getSinceDate() == null ? getStartDate(commitResults) : config.getSinceDate();
+        Date startDate;
+        startDate = (config.getSinceDate().equals(SinceDateArgumentType.ARBITRARY_FIRST_COMMIT_DATE))
+                ? getStartDate(commitResults) : config.getSinceDate();
```
damithc: This slight change will make it easier to read (as per out coding standard).
```java
startDate = (config.getSinceDate().equals(SinceDateArgumentType.ARBITRARY_FIRST_COMMIT_DATE))
        ? getStartDate(commitResults)
        : config.getSinceDate();
```

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -41,26 +41,42 @@ public void setUp() throws IOException {
         FileUtil.deleteDirectory(FT_TEMP_DIR);
     }
 
+    /**
+     * System test with a specified until date and a wildcard since date (capture from the very first commit).
```
yong24s: First mention of `wildcard`,  another one in the pr message.

yong24s:
Changes:

- Simplified, try to be concise; avoid repeating terms
- Removed mention of `- until *`, I think it got removed

```
"since date", when not specified, defaults to the date of the first
commit. Analyzing repos with large number of commits can take up a
long time, which may leave a bad impression for first time users.

Let's,
* limit the range of commits to be captured to one month, if
"since date" is not specified.
* allow `--since *` as a shorthand to specify the date of the first
commit without knowing the actual date.
```

src/systemtest/java/reposense/ConfigSystemTest.java
jylee-git: I feel that specifying a certain character in `--since` flag is better than just specifying `--since` flag and nothing gets parsed in, as it's best to explicitly capture the user's intended date range (or any other critical settings).

Can call it `asterisk` if it's more appropriate.

yong24s: Sounds good.

yong24s: I don't think naming `*` is useful. It may cause confusion. i.e. You have a bracket clarification.

`System test with a specified until date and a * since date to capture from the first commit`?

jylee-git: Ok, will remove the naming convention of `*`

PR #693, opened 2019-05-22
[#506] Report: add checks to ensure that .zip file is uploaded

### LINKED ISSUE #506: Uncaught exception when user uploaded non .zip file
Current:
There is uncaught exception when user uploaded a non .zip file, while for now there is no any negative effect, but for better software engineering practice we should aim to catch it and do something with it (see Suggestion below).

Suggestion:
Implement a pop-up to remind the user to upload .zip file.
This will also be in line with currently being implemented pop-up if loading of resources failed.

jylee-git:
```
Uploaded files are not checked if the file type is of .zip file.

No alerts will be shown to the user and nothing will happen if they 
were to upload non .zip files to the web page.

Let's check if the uploaded file is .zip file, and show an alert to 
the user if the user uploaded a non .zip file to the web page.
```

Resolves #506


fzdy1914:
For futher reviewers, do take notice of the hovering text.

frontend/src/index.jade
```diff
@@ -60,10 +60,10 @@ html
                   span &nbsp; icon on the left.
 
       template(v-else)
-        .empty please enter a report directory or upload a report zip
+        .empty Please upload the .zip file that is generated by RepoSense.
```
fzdy1914: Please upload `a` .zip file that is generated by RepoSense.

jylee-git:
> I am wondering if you can integrate the checking of whether the zip file is valid or not in this? We should have some kind of [check](https://github.com/reposense/RepoSense/blob/master/frontend/src/static/js/api.js#L13) now, but it seems that it is not displaying.

This is the best I can do: -
https://github.com/reposense/RepoSense/blob/2aadb098a7fc689d0a60e9eda767f828a18590a7/frontend/src/static/js/main.js#L121-L124

 I can't shift the codes in here: -https://github.com/reposense/RepoSense/blob/27655c76427fd97a00b3ee2debcd348c8b94b047/frontend/src/static/js/api.js#L13

as uncaught exceptions will trigger if the user tried to upload non .zip files (it already assumes that the user actually uploaded a .zip file and tries to process the file, triggering uncaught exception), which I've have addressed the uncaught exception issue in the latest commit.

frontend/src/index.jade
```diff
@@ -60,10 +60,10 @@ html
                   span &nbsp; icon on the left.
 
       template(v-else)
-        .empty please enter a report directory or upload a report zip
+        .empty Please upload a .zip file that is generated by RepoSense.
 
         form#file-upload(onsubmit="return false;")
-          input(type="file", v-on:change="updateReportZip")
+          input(id="fileUpload", type="file", accept=".zip", v-on:change="updateReportZip")
```
ongspxm: why did you added in an `id` here, if not needed, can remove.

jylee-git: No longer needed.  The ID was before I integrated the checks with the existing .zip file checker. Removed the ID in the latest commit.

frontend/src/static/js/main.js
```diff
@@ -118,6 +118,9 @@ window.app = new window.Vue({
       window.JSZip.loadAsync(evt.target.files[0])
           .then((zip) => {
             window.REPORT_ZIP = zip;
+          }, () => {
+            window.alert('Either the .zip file is corrupted, or you did not upload a .zip file that is generated '
```
eugenepeh: >did not upload a .zip file

Would the alert appear if the user did not upload a file? If not change the wording abit such that
`you have uploaded a zip file that is not generated...`

frontend/src/index.jade
```diff
@@ -60,10 +60,10 @@ html
                   span &nbsp; icon on the left.
 
       template(v-else)
-        .empty please enter a report directory or upload a report zip
+        .empty Please upload a .zip file that is generated by RepoSense.
```
eugenepeh: can remove the ~that is~

PR #573, opened 2019-02-24
Report: Add zoom in feature for ramp charts

ongspxm:
- [x] highlighting of ramp charts #544 
- [x] refactoring ramp charts #572 
- [x] refactoring tab display (from master) #566 
- [x] add mechanism to load diff tabs #574 
- [x] the actual zoomin tab #575

```
The commit granularity setting for the ramp chart slice produces a ramp 
chart that may be too dense to reveal any substantial information. 

Let's add a side panel that provides a zoomed-in view of the ramp 
chart, selected based on a specified time frame.
```

PR #698, opened 2019-05-23
[#673] Modify slice url to commit details page when granularity is `Commit`

### LINKED ISSUE #673: Open the GitHub page with commit details when a ramp is clicked

jamessspanggg:
Fixes #673 

```
When value of granularity is `Commit`, the slice's url points to the GitHub page 
showing a list of commits even though there is only one commit.

Users will need the extra effort to click on the commit in the Github page and 
navigate to the commit details page.

Let's update the slice's url to point directly to the commit details page to save
an extra click from the user.
```

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -172,6 +172,14 @@ window.vSummary = {
                 + `since=${slice.date}'T'00:00:00+08:00&`
                 + `until=${untilDate}'T'23:59:59+08:00`;
     },
+    getSliceLinkCommit(user, slice, index) {
+      const { REPOS } = window;
+
+      return `http://github.com/${
+        REPOS[user.repoId].location.organization}/${
+        REPOS[user.repoId].location.repoName}/commit/${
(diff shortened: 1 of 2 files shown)


jamessspanggg:
Apologies again for the repeated PR.

frontend/src/static/js/v_summary.js
```diff
@@ -172,6 +172,14 @@ window.vSummary = {
                 + `since=${slice.date}'T'00:00:00+08:00&`
                 + `until=${untilDate}'T'23:59:59+08:00`;
     },
+    getSliceLinkCommit(user, slice, index) {
```
fzdy1914: One nit:
`getSliceCommitLink` seems to be better than `getSliceLinkCommit`.

jamessspanggg:
Fixed in PR #573, will close this PR.

PR #700, opened 2019-05-24
[#695] Report: allow reset of date controls

### LINKED ISSUE #695: Report: reset date to earliest "since date" or latest "until date" when user clicks on "X" button in "since date" or "until date" respectively
image

**Currently**: clicking on the "X" button just clears the date, and does nothing else after that, which isn't useful.

**Suggested**: clicking on the "X" button resets the date to the earliest "since date" and latest "until date" possible, which is what majority of the users will intend to do whenever they click on the "X" button.

jylee-git:
```
Pressing the "X" button in the date field only clears the 
date, with no useful functionality.

Users who wish to go back to the default dates will 
need to manually pick the dates from the date picker,
which can be troublesome.

Let's reset the "since date" and "until date" to the earliest and 
latest date respectively, should the user clear the date in the 
respective boxes.
```
Resolves #695

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -145,12 +145,11 @@ window.vSummary = {
     tmpFilterSinceDate() {
       if (this.tmpFilterSinceDate && this.tmpFilterSinceDate >= this.minDate) {
         this.filterSinceDate = this.tmpFilterSinceDate;
-        this.getFiltered();
       } else if (!this.tmpFilterSinceDate) { // If user clears the since date field
         this.filterSinceDate = this.minDate;
         this.tmpFilterSinceDate = this.filterSinceDate;
-        this.getFiltered();
       }
+      this.getFiltered();
     },
     tmpFilterUntilDate() {
       if (this.tmpFilterUntilDate && this.tmpFilterUntilDate <= this.maxDate) {
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -89,9 +89,7 @@ window.vSummary = {
       if (this.tmpFilterSinceDate && this.tmpFilterSinceDate >= this.minDate) {
         this.filterSinceDate = this.tmpFilterSinceDate;
         this.getFiltered();
-      }
-      // If user clears the since date field
-      else if (!this.tmpFilterSinceDate) {
+      } else if (!this.tmpFilterSinceDate) { // If user clears the since date field
         this.filterSinceDate = this.minDate;
         this.tmpFilterSinceDate = this.filterSinceDate;
         this.getFiltered();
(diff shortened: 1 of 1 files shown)


jylee-git:
I need opinions for this: -

By pressing the "X" button, besides resetting the filtered date, should we display the earliest and/or latest date possible (e.g.: user initially filtered the date 1/1/2018 to 20/4/2019, but clicks on the "X" button to reset the "since date" to 9/10/2017)?  
image

Or just make the input box blank without a date?
image

frontend/src/static/js/v_summary.js
```diff
@@ -89,12 +89,20 @@ window.vSummary = {
...
       }
     },
     tmpFilterUntilDate() {
       if (this.tmpFilterUntilDate && this.tmpFilterUntilDate <= this.maxDate) {
         this.filterUntilDate = this.tmpFilterUntilDate;
         this.getFiltered();
+      } else if (!this.tmpFilterUntilDate) { // If user clears the until date field
+        this.filterUntilDate = this.maxDate;
+        this.tmpFilterUntilDate = this.filterUntilDate;
+        this.getFiltered();
```
eugenepeh: can just move the existing `this.getFiltered();` to the end of the function, no need for 2 instance of it since it needs to be executed in both cases.

eugenepeh:
>Currently, pressing the "X" button in the date field only clears the 
date, with no useful functionality.

Currently is implied, its not needed in the commit message.

>Users who previously filtered the dates and want to reset the date 
filter will need to manually pick the dates from the date picker,
which can be troublesome.

`For users who wish to go back to the default dates...`

fzdy1914:
Notice that we have a `reset data range` which plays the same role now.

jylee-git:
> Notice that we have a `reset data range` which plays the same role now.

Similar, but "reset date range" resets both "since date" and "until date", which some users might not want to reset both dates simultaneously. Here, we can provide an option for users to reset dates individually.

frontend/src/static/js/v_summary.js
```diff
@@ -146,13 +146,20 @@ window.vSummary = {
       if (this.tmpFilterSinceDate && this.tmpFilterSinceDate >= this.minDate) {
         this.filterSinceDate = this.tmpFilterSinceDate;
         this.getFiltered();
+      } else if (!this.tmpFilterSinceDate) { // If user clears the since date field
+        this.filterSinceDate = this.minDate;
+        this.tmpFilterSinceDate = this.filterSinceDate;
+        this.getFiltered();
```
yong24s: Normally when we [comment](https://github.com/reposense/RepoSense/pull/700#discussion_r289009357), it applies to all the code. So here as well.

jylee-git: My bad, missed that out as well..

eugenepeh:
>Notice that we have a reset data range which plays the same role now.

Hmm, a overlook on our part, we have merged what @ongspxm has accidentally added that into his epic PR.
Let this PR plays a part in the backdating of that event as well as allow the cancel button to have some useful functionality too then.

PR #704, opened 2019-05-24
[#702] Fix url directing to the wrong branch

### LINKED ISSUE #702: Octacat icons always link to the master branch of the repository
The octacat icons always link to the master branch of the repository, even though the summary chart's branch is not master.

Screen Shot 2019-05-24 at 3 07 12 PM

jamessspanggg:
Fixes #702 

```
The Octocat icon always links to the master branch of a GitHub 
repository.

Octocat icons of summary charts that are not within the master 
branch will still point to the URL of the master branch.

Let's update the URL to point to their respective branch.
```


frontend/src/static/js/api.js
```diff
@@ -70,9 +70,8 @@ window.api = {
           ];
 
           obj.searchPath = searchParams.join('_').toLowerCase();
-          obj.repoPath = `${repo.displayName}`;
           obj.repoName = `${repo.displayName}`;
-          obj.location = `${repo.location.location}`;
+          obj.location = `https:/github.com/${repo.location.organization}/${repo.location.repoName}/tree/${repo.branch}`;
```
fzdy1914: I don't think it is a good idea to set `obj.location` to be related to GitHub only.  We also support analyzing local repos. This will break the location of local repo. Instead, edit the function of getting the github link.

frontend/src/index.jade
```diff
@@ -145,8 +145,8 @@ html
             .summary-chart(v-for="(user, i) in repo")
               .summary-chart__title
                 .summary-chart__title--index {{ i+1 }}
-                .summary-chart__title--repo(v-if="filterGroupSelection === 'groupByNone'") {{ user.repoPath }}
-                .summary-chart__title--author-repo(v-if="filterGroupSelection === 'groupByAuthors'") {{ user.repoPath }}
+                .summary-chart__title--repo(v-if="filterGroupSelection === 'groupByNone'") {{ user.repoName }}
+                .summary-chart__title--author-repo(v-if="filterGroupSelection === 'groupByAuthors'") {{ user.repoName }}
```
fzdy1914: We seem to have `repoName` and `repoPath` for the same value. If it is the case, remove one.

jamessspanggg: Yes I did remove repoPath.

frontend/src/static/js/api.js
jamessspanggg: Okay noted am on it.

frontend/src/index.jade
fzdy1914: What I mean is to remove all the occurrence of `repoPath` and this attribute together. We should not have duplicate arguments which may drag down our performance.

jamessspanggg: I removed the `repoPath` attribute and kept only the `repoName` attribute. After that, I replaced all occurrence of `repoPath` to `repoName` since they both refer to the same value. It is done in 0b13169d74ab86be4d699299ddab742f01ee2080. Is this what you meant?

frontend/src/static/js/api.js
fzdy1914: Sorry, I didn't see the change, my bad.

frontend/src/index.jade
fzdy1914: Sorry, I didn't see the change, my bad.

frontend/src/static/js/v_summary.js
```diff
@@ -172,6 +172,18 @@ window.vSummary = {
                 + `since=${slice.date}'T'00:00:00+08:00&`
                 + `until=${untilDate}'T'23:59:59+08:00`;
     },
+    getRepoLink(repo) {
+      const { REPOS } = window;
+      const urlString = repo.location;
+
+      if (urlString.indexOf('http://github.com/') === 0 || urlString.indexOf('https://github.com/') === 0) {
```
fzdy1914: Better use `startsWith()` method of String.

frontend/src/static/js/v_summary.js
```diff
@@ -172,6 +172,18 @@ window.vSummary = {
                 + `since=${slice.date}'T'00:00:00+08:00&`
                 + `until=${untilDate}'T'23:59:59+08:00`;
     },
+    getRepoLink(repo) {
+      const { REPOS } = window;
+      const urlString = repo.location;
+
+      if (urlString.startsWith('http://github.com/') || urlString.startsWith('https://github.com/')) {
```
eugenepeh: why not use contains?

jamessspanggg: Because I think that the remote URLs either starts with 'http://github.com/' or 'https://github.com/'. If it is a local repo, the URL will not start with these two.

When we use `startsWith`, it only checks for the keyword at the start of the string.
Whereas `includes` will go through the entire string to check if the keyword is within the
string. So `startsWith` is doing less work which should be faster.

damithc: Prefer code maintainability over efficiency in cases where the efficiency gain is negligible.

jamessspanggg: ok noted.

frontend/src/static/js/v_summary.js
```diff
@@ -176,7 +176,7 @@ window.vSummary = {
       const { REPOS } = window;
       const urlString = repo.location;
 
-      if (urlString.startsWith('http://github.com/') || urlString.startsWith('https://github.com/')) {
+      if (urlString.includes('http://github.com/') || urlString.includes('https://github.com/')) {
```
eugenepeh: actually what i meant was., can just do a `urlString.includes('github.com')`.
There is too many combination and you cant possibly cover all cases. 
E.g. it is also legit for ppl to misconfigure it with `www` prefix and RepoSense will still be able to clone that.

damithc: Should we also capture `://` in the search string? `github.com` can be a legit folder name.

eugenepeh: That make sense. My idea of differentiating local filepath and url (in another issue) is to actually indicate it in the json.

However, another way to go about it is to check if the `organization` is set. Local git has no organization.

frontend/src/static/js/v_summary.js
```diff
@@ -251,6 +251,19 @@ window.vSummary = {
       return res;
     },
 
+    getRepoLink(repo) {
+      const { REPOS } = window;
+      const urlString = repo.location;
+
+      if (urlString.includes('://github.com')) {
+        return `https://github.com/${
+          REPOS[repo.repoId].location.organization}/${
```
eugenepeh: the original indentation is actually correct; line wrap should have double indent

frontend/src/static/js/v_summary.js
```diff
@@ -251,6 +251,19 @@ window.vSummary = {
       return res;
     },
 
+    getRepoLink(repo) {
+      const { REPOS } = window;
+      const urlString = repo.location;
+
+      if (urlString.includes('://github.com')) {
```
eugenepeh: following the previous conversation, let's change it to check if it has a valid organization.

PR #708, opened 2019-05-27
[#531,#707] Analysis: use all formats by default

### LINKED ISSUE #531: Identify file types automatically
Current: If file types are not specified in the config files, RepoSense analyzes a set of default file types
Suggested: If file types are not specified, analyze all file types modified during the analysis period

### LINKED ISSUE #707: Full directory captured as file format when file type is not of standard extension
image

The file `makefile` is just a file without any extension `XXX.ext`. It should be captured as the file type `makefile`, but RepoSense captured the full directory where the file is located at instead.

Found this issue when I was working on #708 , but I don't believe it's due to regression.

jylee-git:
```
By default, a set of common programming file formats is used for
analysis. To override the default set, one would need the knowledge of
the file formats used in their repository.

Let's,
* simplify this process by using all file formats present in the
repository by default.
* fix the wrong detection of full directory path instead of filename for
files without file extensions.
```

resolves #531 
resolves #707


jylee-git:
> Please disallow binary files.
> 
> I believe #711 should be solved before this PR. Otherwise, there will be a lot of unnecessary files that are captured by this default file type.

Should I address the binary file issue together in this PR?

fzdy1914:
> Should I address the binary file issue together in this PR?

My suggestion is to put this on hold and solve #711 in a separate pr. Because I believe solving #711 is not trivial and the implementation may need to be discussed.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -34,7 +34,7 @@
     private transient AuthorConfiguration authorConfig;
     private transient boolean isStandaloneConfigIgnored;
     private transient List<CommitHash> ignoreCommitList;
-    private transient boolean isFormatsOverriding;
+    private transient boolean isCsvConfigOverridingFormats;
```
fzdy1914: This naming is weird. Did you consider the situation where we pass in formats through Command Line?

src/main/java/reposense/model/Format.java
```diff
@@ -1,17 +1,18 @@
...
  * Represents a file format in {@code RepoConfiguration}.
  */
 public class Format {
-    public static final List<String> DEFAULT_FORMAT_STRINGS = Arrays.asList(
+    public static final List<Format> DEFAULT_TEST_FORMATS = convertStringsToFormats(Arrays.asList(
             "adoc", "cs", "css", "fxml", "gradle", "html", "java", "js",
-            "json", "jsp", "md", "py", "tag", "txt", "xml");
-    public static final List<Format> DEFAULT_FORMATS = convertStringsToFormats(DEFAULT_FORMAT_STRINGS);
+            "json", "jsp", "md", "py", "tag", "txt", "xml"));
+    public static final List<Format> NO_SPECIFIED_FORMATS = convertStringsToFormats(Collections.emptyList());
```
fzdy1914: Same for this.

frontend/src/index.pug
```diff
@@ -136,7 +136,7 @@ html
             .summary-charts__fileformat--breakdown(v-if="filterBreakdown")
               template(v-if="filterGroupSelection === 'groupByRepos'")
                 .summary-charts__fileformat--breakdown__legend(
-                    v-for="fileFormat in Object.keys(repo[0].fileFormatContribution)")
+                    v-for="fileFormat in getFileFormatsTouchedInRepo(repo)")
```
fzdy1914: Why we now have a situation where different users may get different file types?

frontend/src/static/js/v_authorship.js
```diff
@@ -199,7 +199,7 @@ window.vAuthorship = {
     },
 
     addBlankLineCountToFileFormat(path, lineCount, filesInfoObj) {
-      let fileFormat = path.split('.').pop();
+      let fileFormat = path.split(/[./\\]/).pop();
```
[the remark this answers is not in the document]
jylee-git: Yes

src/main/java/reposense/model/RepoConfiguration.java
jylee-git: > Did you consider the situation where we pass in formats through Command Line?

Yes, but they don't set the variable `isFormatsOverriding` to true. The only case where `isFormatsOverriding` is set to true is when you are using CSV config files and that you set the column in `File Formats` in `repo-config.csv` to `override:java;md;...`

Meanwhile I'll change the variable name back to `isFormatsOverriding` and see how we will proceed with this

frontend/src/index.pug
jylee-git: Because some users may have contributed to certain file formats, and some other may not. So right now, if all file formats are captured, this is how the `commits.json` file will look like:
image

Where file formats that are not touched by authors will no longer be assigned to the authors under the `authorFileFormatContributionMap` object.

Should we stick to this form? Or should we take an approach where:- 
1. Detect all file formats touched by all authors, store them in a list first
2. Then set all these detected formats to each authors, and default to 0 for those formats not touched by the authors
?

fzdy1914: OK，I think your approach is better. Let's just use this one.

src/main/java/reposense/model/RepoConfiguration.java
fzdy1914: OK.

frontend/src/static/js/v_authorship.js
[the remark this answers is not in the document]
fzdy1914: I think this will result in a lot of single file format? I personally recommend adding a `no-format` format. Can ask senior's opinion.

jylee-git:
> Can you add a unit test, in a new testrepo alpha branch, to ensure that all the non-binary files' format are captured?
> Do include some rare formats that was not in our previous default formats.

This has already been done in #728 though, where one of the test case being to filter out files that are binary (regardless of what the format is; even files can be binary when the format is `txt`).

eugenepeh:
>
This has already been done in #728 though, where one of the test case being to filter out files that are binary (regardless of what the format is; even files can be binary when the format is txt).

Yes but it doesnt test whether the formats extraction (if not specify) works, does it?

jylee-git:
> Yes but it doesnt test whether the formats extraction (if not specify) works, does it?

Oh right. WIll add a test case for formats that are not specified

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -121,7 +122,7 @@ private static ArgumentParser getArgumentParser() {
                 .nargs("*")
                 .metavar("FORMAT")
                 .type(new AlphanumericArgumentType())
-                .setDefault(Format.DEFAULT_FORMAT_STRINGS)
+                .setDefault(new ArrayList<String>())
```
yong24s: If we are not adding elements into the `ArrayList`, use `Collections.emptyList()`

frontend/src/static/js/v_summary.js
```diff
@@ -236,7 +236,17 @@ window.vSummary = {
 
       return totalBars;
     },
-
+    getFileFormatsTouchedInRepo(repo) {
```
yong24s: Is it implied or necessary to name it with `Touched`?

Usually less is better, no need to keep a mental note or define its meaning.

PR #717, opened 2019-05-29
[#231] modify local repository URL to the correct remote repository 

### LINKED ISSUE #231: Clicking on ramps in ramp chart generated using local repos redirects to non-existent link
Following #210, the clicking on ramps in ramp chart generated using local repos will redirect user to non-existence link. This is because local repository lacks organization info.

Let's fix this by making the links redirect to the correct remote repository if the remote url can be found in the local repository, otherwise make it non-clickable.

**Note: an alternative solution is to show the commits info locally in the dashboard instead of redirecting to GitHub repo.

jamessspanggg:
Fixes #231 

```
When analysing local repositories, clicking on their links from the 
ramp slices, octocat icon and repository title leads users to 
non-existent links.

This will cause users to not be able to locate to the correct remote 
repository that the local repository was cloned from.

Let's update the URL of these links to redirect to the correct remote
repository by updating location to the remote location that is 
obtained from running the git command: git remote get-url origin
```

### DIFF
--- a/src/main/java/reposense/git/GitRemote.java
+++ b/src/main/java/reposense/git/GitRemote.java
@@ -20,6 +20,10 @@ public class GitRemote {
         Path rootPath = Paths.get(config.getLocation().toString());
         String command = "git remote get-url origin";
 
-        return runCommand(rootPath, command).replace("\n", "");
+        try {
+            return runCommand(rootPath, command).replace("\n", "");
+        } catch (RuntimeException re) {
+            return null;
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


fzdy1914:
What if the local repo does not have a remote at all? These links will still be broken. The user chooses to use local maybe simply because it does not have a remote. 

Let's ask seniors if the PR worth going on.

damithc:
We can do our best to find the remote and use that. If we cannot find a suitable remote, we can omit the links from the report

eugenepeh:
>We can do our best to find the remote and use that. If we cannot find a suitable remote, we can omit the links from the report

Actually my idea would be to open them as a link to the local file on browser.
As highlighted by @fzdy1914 , a remote url may not be available. Also, even if they are, the link on the code views may still be correspondingly broken because the files may not be up on the remote repository. Worse still, the one who use it on local repo may not even be a github user.

jamessspanggg:
Closing this PR as opening local repository links as links to the local files on browser is a better option.

PR #728, opened 2019-06-04
[#711] Analysis: ignore binary files

### LINKED ISSUE #711: Code view: Binary files visible
Expected: binary files should be hidden from code view (even though authors have modified/added these files in).

Current: binary files visible in code view and displays the gibberish codes.

image

jylee-git:
```
Binary files are included in the commits analysis and are shown in the 
report view.

Users will not be able to read the binary codes shown to 
them in the report view.

Let's exclude binary files from the analysis.

The method implemented to determine whether a file is binary or not is 
by utilizing "git diff" functionality return a list of committed files 
with the number of lines added and deleted. A file is deemed to be 
binary when the number of lines added and deleted are both "-".
```

Resolves #711


src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -223,4 +234,15 @@ private static int getStartingLineNumber(String linesChanged) {
...
+
+    /**
+     * Returns true if {@code filePath} is a binary file.
+     */
+    private static boolean isBinaryFile(Path filePath, Path repoRoot) {
+        if ((GitDiff.getNumLinesModified(repoRoot, filePath, Optional.empty(), Optional.empty())).isPresent()) {
+            return false;
+        } else {
+            return true;
+        }
```
damithc: `return !GitDiff.getNumLinesModified(repoRoot, filePath, Optional.empty(), Optional.empty()).isPresent()`?

damithc:
Mark as a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) if still work in progress.

jylee-git:
> Mark as a [draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) if still work in progress.

Ok noted. I can't convert to a draft PR anymore, but will take note next time.

jylee-git:
I would like to get an approval to the implementation of filtering binary files before working on the test cases, in which I would like to create a new branch in `testrepo-Alpha` for the binary files test cases.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -176,6 +192,20 @@ private static void getAllFileInfo(RepoConfiguration config, Path directory, Lis
                     continue;
                 }
 
+                if (!FileUtil.isValidPath(filePath.toString())) {
```
fzdy1914: Seems to be duplicate here.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -176,6 +192,20 @@ private static void getAllFileInfo(RepoConfiguration config, Path directory, Lis
...
+                    logger.warning(String.format(INVALID_FILE_PATH_MESSAGE_FORMAT, filePath));
+                    continue;
+                }
+
+                if (!fileExistsInRepo(relativePath, Paths.get(config.getRepoRoot()))) {
+                    continue;
+                }
+
+                if (isBinaryFile(relativePath, Paths.get(config.getRepoRoot()))) {
+                    logger.log(Level.FINE, relativePath + " is a binary file and will be ignored.");
```
fzdy1914: This warning will spam the console output. For now, maybe a better solution is to check the file is inside format list first, then, we check if it is binary.

src/main/java/reposense/git/GitDiff.java
```diff
@@ -18,4 +21,18 @@ public static String diffCommit(String root, String lastCommitHash) {
...
         return runCommand(rootPath, "git diff -U0 " + lastCommitHash);
     }
+
+    /**
+     * Returns the raw message of the number of lines modified of a given file {@code relativePath} form the commit
+     * {@code priorCommitHash} to the commit {@code lastCommitHash}.
+     */
+    public static String diffNumLinesModified(Path rootPath, String relativePath,
+            Optional<String> fromCommitHash, Optional<String> toCommitHash) {
+        String priorCommitHash = (fromCommitHash.isPresent()) ? fromCommitHash.get() : EMPTY_COMMIT_HASH;
```
fzdy1914: If `fromCommitHash` and `toCommitHash` are always `Optional.empty` when you call it, then it is unnecessary to have these two arguments.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -223,4 +253,26 @@ private static int getStartingLineNumber(String linesChanged) {
 
         return Integer.parseInt(chunkHeaderMatcher.group(STARTING_LINE_NUMBER_GROUP_NAME));
     }
+
+    /**
+     * Returns true if {@code relativePath} exists in the HEAD of the currently checked out commit.
+     * If {@code returnMessage} is empty, it means that the file does not exist in the remote repo.
+     */
+    private static boolean fileExistsInRepo(String relativePath, Path repoRoot) {
+        String returnMessage = GitDiff.diffNumLinesModified(repoRoot, relativePath, Optional.empty(), Optional.empty());
```
fzdy1914: This method and the following method calls the same git command. In your logic now, this git command will be run twice, which is a waste of resource. Please combine them into one method.

src/main/java/reposense/authorship/FileInfoExtractor.java
jylee-git: Will remove the logging part. 




> For now, maybe a better solution is to check the file is inside format list first, then, we check if it is binary.

Because we are going to implement "detect all file formats" in #708 , we might need to check for all files. Also, even if file formats are specified by users, some files might still be binary,  regardless of the file format (for rare cases). So I think it might be better if all files that come in gets checked for binary files, regardless of the white-listed file format list.

jylee-git:
Current implementation now is to grab the list of non-binary files using `git diff` command beforehand, then filter out files that are not inside this list (astheese files are deemed to be binary files).

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -93,6 +97,7 @@
         }
 
         String[] fileDiffResultList = fullDiffResult.split(DIFF_FILE_CHUNK_SEPARATOR);
+        HashSet<Path> nonBinaryFilesList = getListOfNonBinaryFiles(config);
```
fzdy1914: Can use `Set` instead of `HashSet`

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -123,6 +132,23 @@
...
+    public static HashSet<Path> getListOfNonBinaryFiles(RepoConfiguration repoConfig) {
+        String receivedMsg = GitDiff.gitGetModifiedFiles(Paths.get(repoConfig.getRepoRoot()));
+        String[] listOfFiles = receivedMsg.split("\n");
+
+        HashSet<String> nonBinaryFiles =  Arrays.stream(listOfFiles)
+                .filter(file -> !file.startsWith(BINARY_FILE_LINE_DIFF_RESULT))
+                .map(rawNonBinaryFile -> rawNonBinaryFile.split("\t")[2])
+                .collect(Collectors.toCollection(HashSet::new));
+
+        return nonBinaryFiles.stream().filter(fileName -> FileUtil.isValidPath(fileName))
```
fzdy1914: Why you do an unnecessary `collect` and `stream`? Can concatenate them together. If you are afraid of clearness, you can add comments.

jylee-git:
> Do add the relevant UG, too.

Will it be a good idea to add it in the Q&A?

fzdy1914:
> > Do add the relevant UG, too.
> 
> Will it be a good idea to add it in the Q&A section of the UG?

I think it should be added not only in the Q&A, because it is a notable behavior.

docs/UserGuide.md
```diff
@@ -345,6 +346,10 @@ Alternatively, you can use csv config files to customize the analysis as before
 
 Follow this [guide](PublishingGuide.md) to automate publishing of your report to GitHub Pages.
 
+## Note on Binary Files
+
+RepoSense currently does not support analyzing of [Binary files](https://www.webopedia.com/TERM/B/binary_file.html). Common binary files include images (`.jpg`, `.png`), applications (`.exe`), zip files (`.zip`, `.rar`) and certain document types (`.docx`, `.pptx`).
+
```
fzdy1914: Add: `These files will be ignored during analysis.`

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -123,6 +133,23 @@
...
 
+    /**
+     * Returns the list of non-binary files for the repo {@code repoConfig}.
+     */
+    public static Set<Path> getListOfNonBinaryFiles(RepoConfiguration repoConfig) {
+        String receivedMsg = GitDiff.gitGetModifiedFiles(Paths.get(repoConfig.getRepoRoot()));
+        String[] listOfFiles = receivedMsg.split("\n");
+
+        // Gets rid of binary files and files with invalid directory name, then convert into type path
+        // and return as a Set.
```
fzdy1914: I think `then convert into type path and return as a Set.` is not necessary as the message is easily implied from the code.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -223,4 +255,11 @@ private static int getStartingLineNumber(String linesChanged) {
 
         return Integer.parseInt(chunkHeaderMatcher.group(STARTING_LINE_NUMBER_GROUP_NAME));
     }
+
+    /**
+     * Returns true if {@code filePath} is valid and non-binary file.
```
fzdy1914: ~file~

docs/UserGuide.md
```diff
@@ -16,6 +16,7 @@
     * [`author-config.csv`](#author-configcsv)
 * [Analyzing Multiple Repos](#analyzing-multiple-repos)
 * [Using Travis-CI to automate publishing of the report to GitHub Pages](#using-travis-ci-to-automate-publishing-of-the-report-to-github-pages)
+* [Note on Binary Files](#note-on-binary-files)
```
fzdy1914: Seems weirder if the part is here. Let the seniors decide where to place them.

eugenepeh: I think can remove the entire topic on `Note on Binary Files` for now and just leave it in FAQ

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -164,6 +191,7 @@ private static void setLinesToTrack(FileInfo fileInfo, String fileDiffResult) {
      * based on {@code config} and inserts it into {@code fileInfos}.
      */
     private static void getAllFileInfo(RepoConfiguration config, Path directory, List<FileInfo> fileInfos) {
+        Set<Path> nonBinaryFilesList = getListOfNonBinaryFiles(config);
         try (Stream<Path> pathStream = Files.list(directory)) {
             for (Path filePath : pathStream.collect(Collectors.toList())) {
```
eugenepeh: the new implementation should be able to replace the existing pathStream implementation.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -168,6 +169,18 @@ public static void createDirectory(Path dest) throws IOException {
         Files.createDirectories(dest);
     }
 
+    /**
+     * Returns true if {@code path} is a valid path.
+     */
+    public static boolean isValidPath(String path) {
```
eugenepeh: There is a more exhaustive `isValidPath` done by one of your seniors in the past.
Please check if its still relevant https://github.com/reposense/RepoSense/commit/3157e1b204ebbd1f2e9693d1d8af24423bcd6624

Technically if this new implementation can prevent that issue, then we should remove all the deprecated codes.

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -163,6 +166,18 @@ public void generateFileInfo_fileWithoutSpecialCharacters_correctFileInfoGenerat
...
+    @Test
+    public void getListOfNonBinaryFiles_directoryWithBinaryAndNonBinaryFiles_success() {
+        GitCheckout.checkoutBranch(config.getRepoRoot(), BINARY_AND_NON_BINARY_FILES_BRANCH);
+        Set<Path> files = FileInfoExtractor.getListOfNonBinaryFiles(config);
+
+        Assert.assertEquals(1, files.size());
+        // Non binary files should be captured
+        Assert.assertTrue(files.contains(Paths.get("binaryFileTest/nonBinaryFile.txt")));
+        // Binary files should be ignored
+        Assert.assertFalse(files.contains(Paths.get("binaryFileTest/binaryFile.txt")));
```
eugenepeh: try to make the test a little more exhaustive, adding in images, pdf n stuffs.
Also a few more directories and sub folders will be good as well, we need to make sure the implementation is recursive as well.

For files test, we should at least aim for min 10 files.

src/main/java/reposense/util/FileUtil.java
jylee-git: I believe the one done up in this commit https://github.com/reposense/RepoSense/commit/3157e1b204ebbd1f2e9693d1d8af24423bcd6624 serves a different purpose than as compared to this implementation.

The implementation of `isValidPath` in this PR only deals with files in the repo that contain illegal characters (typically in Windows OS), ensuring that the file name is compatible with whatever the user's OS is, whereas the one in https://github.com/reposense/RepoSense/commit/3157e1b204ebbd1f2e9693d1d8af24423bcd6624 deals with globs.

eugenepeh: I see. Can you please rename that to isValidIgnoreGlob, with the `path` rename to `IgnoreGlob` as well, to avoid any possible conflict of apis/misleadings in future?

E.g. `private static boolean isValidIgnoreGlob(File repoRoot, String ignoreGlob) {`

jylee-git: Ok, made the changes.

docs/UserGuide.md
```diff
@@ -374,3 +374,6 @@ It is also possible that another user has overriden the authorship of your lines
 
 #### Q: I am able to run RepoSense on my repository on a Linux/Mac OS, but it fails on a Windows OS?
 **A:** It is possible you may have some file names with [special characters](https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#naming-conventions) in them, which is disallowed in Windows OS. As such, *RepoSense* is unable to fully clone your repository, thus failing the analysis.
+
+#### Q: Some files are not captured by RepoSense!
+**A:** The files may be [binary files](https://www.webopedia.com/TERM/B/binary_file.html). RepoSense currently does not support analyzing binary files. Common binary files include images (`.jpg`, `.png`), applications (`.exe`), zip files (`.zip`, `.rar`) and certain document types (`.docx`, `.pptx`). Therefore, these files will be ignored during analysis.
```
yong24s: Can remove `currently` since we have no intention of supporting binary files.

RepoSense ~currently~ does not ~support~ analyzing binary files.

Reason: We only analyze code/doc contributions.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -109,20 +110,36 @@
...
             }
         }
 
         return fileInfos;
     }
 
+    /**
+     * Returns the list of non-binary files for the repo {@code repoConfig}.
+     */
+    public static Set<Path> getListOfNonBinaryFiles(RepoConfiguration repoConfig) {
```
yong24s: Is it necessary to use a `Set`, will there be duplicates?

Update the JavaDocs according. Returns a {@code Set<Path>} of non-binary files

src/main/java/reposense/git/GitDiff.java
```diff
@@ -18,4 +20,13 @@ public static String diffCommit(String root, String lastCommitHash) {
         Path rootPath = Paths.get(root);
         return runCommand(rootPath, "git diff -U0 " + lastCommitHash);
     }
+
+    /**
+     * Returns the raw message of a list of committed files with the corresponding number of lines added and deleted
+     *  in the repo {@code repoRoot}.
```
yong24s: Extra space between `*  in`.

src/main/java/reposense/authorship/FileInfoExtractor.java
jylee-git: No there won't be duplicates. But set is used here for performance considerations, when repeatedly calling `isValidAndNonBinaryFile` to see if this set contains a specific file.

Will update JavaDocs accordingly

docs/UserGuide.md
```diff
@@ -374,3 +374,6 @@ It is also possible that another user has overriden the authorship of your lines
 
 #### Q: I am able to run RepoSense on my repository on a Linux/Mac OS, but it fails on a Windows OS?
 **A:** It is possible you may have some file names with [special characters](https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#naming-conventions) in them, which is disallowed in Windows OS. As such, *RepoSense* is unable to fully clone your repository, thus failing the analysis.
+
+#### Q: Some files are not captured by RepoSense!
+**A:** The files may be [binary files](https://www.webopedia.com/TERM/B/binary_file.html). RepoSense does not analyze binary files. Common binary files include images (`.jpg`, `.png`), applications (`.exe`), zip files (`.zip`, `.rar`) and certain document types (`.docx`, `.pptx`).
```
yong24s: Let's use Wikipedia instead? https://en.wikipedia.org/wiki/Binary_file
The link contains ads that we have no control on.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -109,20 +110,36 @@
...
 
         return fileInfos;
     }
 
+    /**
+     * Returns a {@code Set} of non-binary files for the repo {@code repoConfig}.
+     */
+    public static Set<Path> getListOfNonBinaryFiles(RepoConfiguration repoConfig) {
+        String receivedMsg = GitDiff.gitGetModifiedFiles(Paths.get(repoConfig.getRepoRoot()));
+        String[] listOfFiles = receivedMsg.split("\n");
```
yong24s: Same as the comment on `message`, for `receivedMsg`, use a more appropriate name.

docs/UserGuide.md
jylee-git: Ok noted. Wasn't aware of this as I'm running ad blockers

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -223,4 +223,11 @@ private static int getStartingLineNumber(String linesChanged) {
 
         return Integer.parseInt(chunkHeaderMatcher.group(STARTING_LINE_NUMBER_GROUP_NAME));
     }
+
+    /**
+     * Returns true if {@code filePath} is valid and non-binary.
+     */
+    private static boolean isValidAndNonBinaryFile(String filePath, Set<Path> nonBinaryFilesSet) {
+        return (FileUtil.isValidPath(filePath) && nonBinaryFilesSet.contains(Paths.get(filePath)));
```
yong24s: Extra parenthesis `()`

Small optimization: It may be better to test this `nonBinaryFilesSet.contains(Paths.get(filePath)` first before testing `I/O`

src/main/java/reposense/util/FileUtil.java
```diff
@@ -168,6 +171,20 @@ public static void createDirectory(Path dest) throws IOException {
...
     }
 
+    /**
+     * Returns true if {@code path} is a valid path.
+     * Produces log messages when the invalid file path is skipped.
+     */
+    public static boolean isValidPath(String path) {
+        try {
+            Paths.get(path);
+        } catch (InvalidPathException | NullPointerException ex) {
```
yong24s: When will `NullPointerException` be thrown? `path` is null?

jylee-git: My bad, `NullPointerException` won't be thrown under any circumstances, and we won't need to deal with a null input `path`.

src/main/java/reposense/authorship/FileInfoExtractor.java
jylee-git: > Small optimization: It may be better to test this `nonBinaryFilesSet.contains(Paths.get(filePath)` first before testing `I/O`

Not possible in this current implementation, as attempting to parse paths with illegal windows characters into `nonBinaryFilesSet.contains(Paths.get(filePath)` will throw exceptions if we do not check for I/O first.

yong24s: Following https://github.com/reposense/RepoSense/commit/fed37581bbf17ae61a587256633be47924284b70, I think you can assume all file path will be valid?

jylee-git: These two test cases do not call `GitLsTree.validateFilePaths` to check for the branch "windows illegal filename branch" as they directly checkout from master after validating that the master branch does not contain any illegal filenames:

https://github.com/reposense/RepoSense/blob/e1fbec2e4615302fabc345b5a0187358d72d8e29/src/test/java/reposense/authorship/FileInfoExtractorTest.java#L146-L157

https://github.com/reposense/RepoSense/blob/e1fbec2e4615302fabc345b5a0187358d72d8e29/src/test/java/reposense/authorship/FileInfoExtractorTest.java#L73-L85


Hence the need to check for valid file path

yong24s: This discussion is beyond this PR. If require, let's open a new issue for this.

My view is the unit test should be removed or changed to use `GitLsTree.validateFilePaths`. Now we are coding to accommodate this unit test. On runtime, we can assume that all filepaths are valid.

What do you think @eugenepeh?

eugenepeh: its good to have fail safe during run time to conform that assumption.
If the test is not thorough enough, then we better raise an issue for rework.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -109,20 +110,36 @@
...
         }
 
         return fileInfos;
     }
 
+    /**
+     * Returns a {@code Set} of non-binary files for the repo {@code repoConfig}.
+     */
+    public static Set<Path> getListOfNonBinaryFiles(RepoConfiguration repoConfig) {
+        String rawGitDiffResult = GitDiff.gitGetModifiedFiles(Paths.get(repoConfig.getRepoRoot()));
```
eugenepeh: I think raw result is not very useful, it may be better to add another api in GitDiff e.g. gitGetModifiedFileList that provides the result as an ArrayList to reduce some clutter in FileInfoExtractor.

jylee-git: Processed and returned the modified files as an ArrayList in the existing GitDiff method

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -128,11 +127,10 @@
      * Returns a {@code Set} of non-binary files for the repo {@code repoConfig}.
      */
     public static Set<Path> getListOfNonBinaryFiles(RepoConfiguration repoConfig) {
```
eugenepeh: just to standardize the naming
`getListOfNonBinaryFiles` -> `getNonBinaryFileList`

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -128,11 +127,10 @@
...
-        String rawGitDiffResult = GitDiff.gitGetModifiedFiles(Paths.get(repoConfig.getRepoRoot()));
-        String[] listOfFiles = rawGitDiffResult.split("\n");
+        List<String> listOfFiles = GitDiff.gitGetModifiedFiles(Paths.get(repoConfig.getRepoRoot()));
```
eugenepeh: `listOfFiles` -> `modifiedFileList`

src/main/java/reposense/git/GitDiff.java
```diff
@@ -18,4 +23,14 @@ public static String diffCommit(String root, String lastCommitHash) {
         Path rootPath = Paths.get(root);
         return runCommand(rootPath, "git diff -U0 " + lastCommitHash);
     }
+
+    /**
+     * Returns a list of committed files with the corresponding number of lines added and deleted in the repo
+     * {@code repoRoot}.
+     */
+    public static List<String> gitGetModifiedFiles(Path repoRoot) {
```
eugenepeh: Don't need the git here

`gitGetModifiedFiles` -> `getModifiedFileList`

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -163,6 +168,29 @@ public void generateFileInfo_fileWithoutSpecialCharacters_correctFileInfoGenerat
         Assert.assertEquals(5, fileInfo.getLines().size());
     }
 
+    @Test
+    public void getListOfNonBinaryFiles_directoryWithBinaryAndNonBinaryFiles_capturesNonBinaryFiles() {
```
eugenepeh: `getListOfNonBinaryFiles` -> `getNonBinaryFiles`
`directoryWithBinaryAndNonBinaryFiles` -> `directoryWithBinaryFiles` - the presence of non binary files is implied.
`capturesNonBinaryFiles` -> `success` - no need explain result if its just a success or failure procedure

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -28,6 +31,8 @@
     private static final String DIRECTORY_WITH_VALID_WHITELISTED_NAME_BRANCH = "directory-with-valid-whitelisted-name";
     private static final String BRANCH_WITH_VALID_WHITELISTED_FILE_NAME_BRANCH =
             "535-FileInfoExtractorTest-branchWithValidWhitelistedFileName.txt";
+    private static final String BINARY_AND_NON_BINARY_FILES_BRANCH =
```
eugenepeh: `BINARY_AND_NON_BINARY_FILES_BRANCH` -> `BRANCH_WITH_BINARY_FILES`

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -28,6 +31,8 @@
...
+            "728-FileInfoExtractorTest-generateFileInfo_directoryWithBinaryAndNonBinaryFiles_success";
```
eugenepeh: do rename the branch accordingly

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -223,4 +221,11 @@ private static int getStartingLineNumber(String linesChanged) {
 
         return Integer.parseInt(chunkHeaderMatcher.group(STARTING_LINE_NUMBER_GROUP_NAME));
     }
+
+    /**
+     * Returns true if {@code filePath} is valid and non-binary.
```
eugenepeh: and the file is not in binary.

docs/UserGuide.md
```diff
@@ -374,3 +374,6 @@ It is also possible that another user has overriden the authorship of your lines
 
 #### Q: I am able to run RepoSense on my repository on a Linux/Mac OS, but it fails on a Windows OS?
 **A:** It is possible you may have some file names with [special characters](https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file#naming-conventions) in them, which is disallowed in Windows OS. As such, *RepoSense* is unable to fully clone your repository, thus failing the analysis.
+
+#### Q: Some files are not captured by RepoSense!
```
eugenepeh: keep it to an appropriate tone
`Why are some files not captured by RepoSense?`

src/test/java/reposense/authorship/FileInfoExtractorTest.java
yong24s: > capturesNonBinaryFiles -> success - no need explain result if its just a success or failure procedure

That was my suggestion. Sorry didn't see clearly. 👀

src/main/java/reposense/git/GitDiff.java
```diff
@@ -18,4 +23,14 @@ public static String diffCommit(String root, String lastCommitHash) {
...
     }
+
+    /**
+     * Returns a list of committed files with the corresponding number of lines added and deleted in the repo
+     * {@code repoRoot}.
+     */
+    public static List<String> getModifiedFiles(Path repoRoot) {
+        String diffCommand = String.format("git diff --numstat %s %s", EMPTY_TREE_HASH, CHECKED_OUT_COMMIT_REFERENCE);
+        String diffResult = runCommand(repoRoot.toAbsolutePath(), diffCommand);
+        return new ArrayList<>(Arrays.asList(diffResult.split("\n")));
```
yong24s: Are we modifying the returned list?

If not, can just `return Arrays.asList(diffResult.split("\n"));`?

src/main/java/reposense/git/GitDiff.java
```diff
@@ -18,4 +23,14 @@ public static String diffCommit(String root, String lastCommitHash) {
         Path rootPath = Paths.get(root);
         return runCommand(rootPath, "git diff -U0 " + lastCommitHash);
     }
+
+    /**
+     * Returns a list of committed files with the corresponding number of lines added and deleted in the repo
+     * {@code repoRoot}.
+     */
+    public static List<String> getModifiedFiles(Path repoRoot) {
```
yong24s: Should also follow Eugene's suggestion to standardize to `FilesList` -> `getModifiedFilesList`

eugenepeh:
Nits to the message, note that try to make your reason stronger, create a bigger impact to encourage changes

>Binary files are included in the commits analysis and are shown in the 
report view. 

Binary files are included in the commits analysis and are shown in the 
report view `as interpreted plain text.`

>Users will not be able to read the binary codes shown to 
them in the report view.

The produced results are gibberish; junks that do not make sense and
not useful to the readers.

PR #729, opened 2019-06-04
[#724] Modify zoom-in feature based on selected granularity

### LINKED ISSUE #724: Unable to activate zoom-in feature when granularity is set to `week`
When granularity is set to `commit` or `day`, the zoom-in feature works fine. But however when it is set to `week`, I am unable to activate the zoom-in feature.

I am not sure if this is intended since the ramp slices may be more far apart when granularity is set to `week`,  but the User Guide did not state that the zoom-in feature is only limited when granularity is set to `commit` or `day`.

jamessspanggg:
Fixes #724 and a continuation of #573

- [x] Implement zoom-in feature for 'week' granularity
- [x] Change ramp slices of zoom-in panel to be in terms of the granularity set

```
The zoom-in feature opens the zoom panel to show the selected
sub-range of a ramp chart in terms of 'commit'. This feature, however, 
only works when the granularity is set to commit or day.

Users will not be able to use this feature with granularity set to week,
as the method getUserCommits() gets all user commits by interating 
through an undefined key object 'dayCommits', which stops the method
execution.

Moreover, long sub-range of ramp charts will contain dense information 
with ramp slices overlapping each other. This happens even when 
granularity is set to day or week, which makes it difficult for users to 
notice substantial info on these ramp charts.

Let's correctly implement the zoom-in feature when granularity is set to
week. Furthermore, let's modify the feature for the ramp chart within 
zoom panel to be in terms of the granularity selected, which allow users 
to observe selected granularity of commits more clearly within the 
zoomed-in range.
```


frontend/src/tabs/zoom.pug
```diff
@@ -10,7 +10,14 @@
       v-bind:udate="info.untilDate",
       v-bind:avgsize="info.avgCommitSize")
 
-  .zoom__day(v-for="day in info.user.commits")
+  .zoom__week(v-if="filterTimeFrame === 'week'", v-for="week in info.user.commits")
```
fzdy1914: The `,` here should be unnecessary. Also, Vue strongly discourages using `v-if` and `v-for` together. If it is needed, please create another div for `v-if`

frontend/src/tabs/zoom.pug
```diff
@@ -10,7 +10,14 @@
...
+    template(v-if="week.insertions > 0")
+      h3 Week of {{ week.date }}
+      ul
```
fzdy1914: These three lines are quite similar with the below one, is it possible to write these two in one template? So you can also avoid the problem I mentioned before.

frontend/src/static/js/v_zoom.js
```diff
@@ -10,8 +14,19 @@ window.vZoom = {
     getSliceLink(slice) {
       return `${window.getBaseLink(this.info.user.repoId)}/commit/${slice.hash}`;
     },
+
+    getFilterTimeFrame() {
+      window.decodeHash();
+      const hash = window.hashParams;
+      if (hash.timeframe) {
+        this.filterTimeFrame = hash.timeframe;
```
fzdy1914: This is not enough. Previously, `commit` and `day` granularity have the same zoom-in panel. Now, the zoom-in pannel for 'week' is different. When the user firstly opens a zoom-in pannel with `day`/`commit` granularity and then switch to 'week' granularity. The zoom-in pannel should react to this action. (Also change to `week` granularity)

frontend/src/static/js/v_summary.js
```diff
@@ -420,6 +421,7 @@ window.vSummary = {
           const commit = commits.shift();
           week.insertions += commit.insertions;
           week.deletions += commit.deletions;
+          commit.commitResults.forEach((commitResult) => week.commitResults.push(commitResult));
```
fzdy1914: I am not so sure if there will be only one commit results in `commit.commitResults`, please check. If so, we do not need `forEach`.

jamessspanggg: There might be cases with more than one commit results.

Screen Shot 2019-06-10 at 11 10 49 AM

frontend/src/static/js/v_zoom.js
jamessspanggg: So for `commit` and `day` granularity, they will have the same ramp (ramp slices are in terms of `commit`) in the zoom-in panel. But for `week` granularity, the ramp in the zoom-in panel should have ramp slices that are in terms of commits for the entire `week`?

jamessspanggg: Or should it be: 
- `commit` granularity will have ramp slices in terms of `commit` in the zoom-in panel
- `day` granularity will have ramp slices in terms of `day` in the zoom-in panel
- `week` granularity will have ramp slices in terms of `week` in the zoom-in panel

jamessspanggg:
> A bug found when playing around with the feature.
> a

Yes I can duplicate this bug as well. But this bug occurs when you select `week` as granularity, reload the page and select the view code icon. (or when u select `week`, select view code icon and reload the page) Will address it in another PR as it is not due to this current PR.

frontend/src/static/js/v_zoom.js
jamessspanggg: @fzdy1914

fzdy1914: Sorry for not seeing the reply. Maybe I am not clear enough. See the following pics.
Now:
a
Expect:
image

fzdy1914: After changing the granularity to `week`, the right panel should also be the one for the week.

jamessspanggg: Oh you mean for the zoom panel to change accordingly when switching the `granularity`. May I address this as a separate PR as it might not be suitable to address this issue in this PR?

jamessspanggg: > Or should it be:
> 
> * `commit` granularity will have ramp slices in terms of `commit` in the zoom-in panel
> * `day` granularity will have ramp slices in terms of `day` in the zoom-in panel
> * `week` granularity will have ramp slices in terms of `week` in the zoom-in panel

@fzdy1914 
Also, shouldn't the zoom-in panel perform as described above? Because the current implementation of zoom-in feature is basically showing the zoomed-in ramp slices in terms of `commit` no matter granularity is set to `commit`, `day` or `week`, which might be redundant as they show the same ramp.

fzdy1914: >May I address this as a separate PR as it might not be suitable to address this issue in this PR?

I think it is better to do it in this PR. Firstly, it is considered as a whole feature. Meanwhile, it is not so difficult to implement.

fzdy1914: > > Or should it be:
> > 
> > * `commit` granularity will have ramp slices in terms of `commit` in the zoom-in panel
> > * `day` granularity will have ramp slices in terms of `day` in the zoom-in panel
> > * `week` granularity will have ramp slices in terms of `week` in the zoom-in panel
> 
> @fzdy1914
> Also, shouldn't the zoom-in panel perform as described above? Because the current implementation of zoom-in feature is basically showing the zoomed-in ramp slices in terms of `commit` no matter granularity is set to `commit`, `day` or `week`, which might be redundant as they all show the same ramp.

I cannot decide which one is better. We can ask @damithc's opinion of this zoom-in panel.

damithc: I think the zoom feature was only meant for commit granularity. For other granularities, there is no risk of having too many ramps on top of each other.

jamessspanggg: I think there are still cases where many ramps on top of each other when granularity is set to `day` and `week`. For example:
Screen Shot 2019-06-13 at 12 29 47 PM
Screen Shot 2019-06-13 at 12 29 54 PM

damithc: > I think there are still cases where many ramps are on top of each other when granularity is set to `day` and `week`.

Yes, if the analysis is over a long period. We can support zoom for other granularity too if that is not too much extra work, but it is not that urgent as our immediate use case is analysis of student work over a semester.

jamessspanggg: I have modified it to support zoom for other granularities.

frontend/src/static/js/v_zoom.js
```diff
@@ -10,6 +14,22 @@ window.vZoom = {
     getSliceLink(slice) {
       return `${window.getBaseLink(this.info.user.repoId)}/commit/${slice.hash}`;
     },
+
+    updateUser(repos) {
+      const userOrig = repos[this.info.repoIndex][this.info.userIndex];
+      const { commits } = userOrig;
+      const commitsWithinDuration = commits.filter(
+          (commit) => commit.date >= this.info.sinceDate && commit.date <= this.info.untilDate,
```
fzdy1914: The date of commits has already been filtered, see [here](https://github.com/reposense/RepoSense/blob/master/frontend/src/static/js/v_summary.js#L562). So maybe one is enough.

frontend/src/static/js/v_zoom.js
```diff
@@ -10,6 +14,22 @@ window.vZoom = {
     getSliceLink(slice) {
       return `${window.getBaseLink(this.info.user.repoId)}/commit/${slice.hash}`;
     },
+
+    updateUser(repos) {
+      const userOrig = repos[this.info.repoIndex][this.info.userIndex];
```
fzdy1914: The way you implement it indeed works, however, it is not ideal. Instead of this line to get `userOrig`, can you find another way to get this and update this? When updating the data, the original user does not need an index to do so.  You should investigate the whole logic of how v_zoom works before make it also works for `week`. There should be a more elegant method instead of `updateUser`. This time, I will not give you the hint directly. I hope you can find it out by yourself. This is a good opportunity for you to improve your programming skills and debug ability.

fzdy1914: Feel free to ask me for a hint if you cannot find a better way.

fzdy1914: > Feel free to ask me for a hint if you cannot find a better way.

Maybe you can hold on to refactor your code. I believe I need to reconsider my way.  Sorry for the disturbance and I will let you know if there are updates.

frontend/src/static/js/v_zoom.js
jamessspanggg: The openTabZoom method is not called when user switches between the granularity, it only updates the `info.user` and `filterTimeFrame` within the zoom component. 

Since I am passing the list of commits that has not been filtered yet to the zoom component, I will need to filter commits that within the selected zoom range.

frontend/src/static/js/v_zoom.js
jamessspanggg: I will try to find a more elegant solution to this as well.

jamessspanggg: @fzdy1914 I am unable to find a more elegant way of updating the user as the splitting of commits based on granularity is done in getFiltered() in v_summary.js, so the only way that I can think of to update the user is to pass the entire filtered repositories and use (repo index, user index) to locate the specific user.

fzdy1914: I am considering to put these info in the url(which we already have), so when updating, we can use the info in the url directly to update zoom-in tab. This will also benefit remembering the state of zoom-in in url. Now we only need `zoom_start_time` and `zoom_end_time` in url more to remember the zoom. But it is only an idea, you may consider whether it is good or not. Your original implement is also acceptable. I will give a review if you think your original implement is better.

jamessspanggg: For now I guess we can just stick with this current implementation? Only when we are targeting the issue of remembering zoom-in state, we just put the extra info required `(repoIndex, userIndex, zoomStart, zoomEnd)` in the URL instead of passing them as props.

frontend/src/index.pug
```diff
@@ -148,26 +148,26 @@ html
                       v-bind:style="{ 'color' : contributionBarColors[fileFormat] }"
                   )
                   span &nbsp; {{ fileFormat }} &nbsp;
-            .summary-chart(v-for="(user, i) in repo")
+            .summary-chart(v-for="(user, userIndex) in repo")
               .summary-chart__title
-                .summary-chart__title--index {{ i+1 }}
+                .summary-chart__title--index {{ userIndex+1 }}
```
fzdy1914: Better use `userIndex + 1`.

frontend/src/static/js/v_zoom.js
fzdy1914: If it is the case, we should only process the commits in `v_zoom`, even when using `openTabZoom`. You should modify the existing logic to make the whole logic more fluent.

fzdy1914:
Is this PR ready for review?

jamessspanggg:
Yes it is.

frontend/src/static/js/v_zoom.js
```diff
@@ -1,15 +1,42 @@
...
+          (commit) => commit.date >= this.info.sinceDate && commit.date <= this.info.untilDate,
+      );
+      user.commits = filtered;
+    },
+
     getSliceLink(slice) {
       return `${window.getBaseLink(this.info.user.repoId)}/commit/${slice.hash}`;
     },
+
+    updateUser(repos) {
```
fzdy1914: Do we need to pass the `repos` in as an argument? Can't it be directly read?

frontend/src/static/js/v_summary.js
```diff
@@ -131,6 +131,7 @@ window.vSummary = {
     },
     filterTimeFrame() {
       this.getFiltered();
+      this.$root.$emit('updateFilterTimeFrame', this.filtered, this.filterTimeFrame);
```
fzdy1914: I am wondering if we can modify the `filterTimeFrame` property of `v_zoom` directly and watch on the `filterTimeFrame` of `v_zoom` instead.
The reason I have this question is that we do not use `$emit` and `$on` before in our code. Not sure this method is ideal or not.

@ongspxm Can you also give your suggestion?

jamessspanggg: From what I learnt about Vue, you can pass data from parent to child component using `props` while the other way round is through `$emit`. But since `v_zoom` is not the child of `v_summary` nor the other way round, I think the only way of passing data is through the listening of a global `$root` event. But yeah I am not sure so it is good to hear from someone who is more experienced in Vue.

frontend/src/static/js/v_zoom.js
jamessspanggg: U mean directly read from `this.filtered` in `v_summary`? I don't think so as they are separate components that have no parent/child relation.

frontend/src/static/js/v_summary.js
fzdy1914: Since @ongspxm  is not replying, let's go to next review rounds.

jamessspanggg:
Put this as ongoing as I need to reconsider my solution.

jamessspanggg:
@eugenepeh I actually implemented it to change according to the `granularity`. But according to @damithc in #793 
> Yes, I think so. It's cleaner that way. Making it reactive fully may not be worth the effort, and hence, best not to make it reactive at all. Zoom is a complicated (to implement) feature. We should not mix it up with other features if it can be helped.

jamessspanggg:
> > Hence I removed the zoom tab's reactivity to granularity
> 
> If thats the case, I think it will be better if we state the granularity used in the zoom tab to avoid any misconception

@eugenepeh ok noted. Would this way of stating the granularity be fine?

Screen Shot 2019-07-13 at 10 39 14 PM

eugenepeh:
Perhaps make it next to the repo title?
image

@damithc ur opinion on this?

damithc:
> @damithc ur opinion on this?

Yes, I'm fine with having it near repo title (but maybe in smaller unbold text?) e.g.,

**Org_repo_branch** <sub>(granularity: one ramp per commit)</sub>

frontend/src/static/css/style.scss
```diff
@@ -559,3 +559,20 @@ header{
...
+        &--repo{
+            display:inline;
+            font-size:1.5rem;
+            font-weight:400;
+        }
+
+        &--granularity{
+            display:inline;
+            font-size:1rem;
+        }
```
eugenepeh: We already have a custom style for smaller font isn't it? can we just reuse that?

jamessspanggg: You mean the `mini-font` class in PR #762 ? The PR is currently still ongoing. 
> actually, instead of `contribution-text-font`, we can just give it a representative name e.g. `mini-font` and make it a global standard in our style sheet.

_Originally posted by @eugenepeh in https://github.com/reposense/RepoSense/pull/762_

Currently there is no global custom style for smaller font. But yeah it would be better if we create a global custom style for `large-font`, `medium-font`, `mini-font`.

eugenepeh: I see, this can be merged after #762, adopting the same solution then

jamessspanggg: Ok noted.

eugenepeh:
Some changes needed on the proposed message:

>When activating the zoom-in feature with granularity set to 'commit'
or 'day', the zoom-in panel shows a ramp chart with ramp slices in
terms of 'commit'. When granularity is set to 'week', the zoom-in 
panel is unable to be activated as the implementation has not been 
completed yet.

Can just mention that the zoom-in feature only works when the granularity is set to commit or day, current paragraph is like going 1 big round just to tell the reader that. 

The zoom-in feature opens a panel to show the selected sub-range of a ramp chart in terms of 'commit'.  This feature, however, only works when the granularity is set to commit or day.

>If the RepoSense analysis is over a long period of time, the ramp chart
will still contain dense information with ramp slices overlapping each 
other even when granularity is set to 'day' or 'week', which is 
difficult for users to notice the substantial info on daily/weekly 
commits.

Add in your reason why week is not implemented here. The reason why we want to **ADDITIONALLY** have different granularities should be further explained with a `moreover...` or etc. Note that this is because it is considered a separate work from just fixing it to work in `week` granularity as well.

>Let's modify the zoom-in feature such that the ramp slices in the zoom-
in panel is in terms of the granularity selected, which allow users to
observe selected granularity of commits more clearly within the 
zoomed-in range.

you forgot to mention that this fixes the problem of not displaying for week granularity here

jamessspanggg:
> Some changes needed on the proposed message:

@eugenepeh I have modified the commit message.

PR #733, opened 2019-06-07
[#689] v_summary.js: remove unused remnants code

### LINKED ISSUE #689: Unused variable filterSortReverse
I observed that there is a variable that is no longer used in the current RepoSense version.

It is the variable filterSortReverse within window.v_summary. I think it is previously used when there is a check box to reverse the sorting order for the groups.
Screen Shot 2019-05-22 at 10 42 33 AM

When sorting the groups, there is an extra conditional statement that checks for the value of this variable. This statement is redundant as filterSortReverse will always be false.
Screen Shot 2019-05-22 at 10 45 41 AM

Suggestion: remove filterSortReverse and all the conditional statements that checks for it to slightly improve the performance.

jamessspanggg:
Fixes #689 

```
Unused code was found in v_summary.js; specifically code associated with
the variable, filterSortReverse. They are remnants left behind by the 
improvement of sort controls introduced in #618.

The unused code was originally bind to the reverse checkbox in the older
version of RepoSense. As the checkbox was replaced by the introduction
of ascending and descending options in #618, the code is now unused
and redundant. This is also partially due to the complexity brought upon 
by new sorting group, the reversing has to be done within the code of 
descending's logic, rendering those code not reusable.

Let's remove the unused code associated to the variable, 
filterSortReverse and it's implementations.
```


eugenepeh:
>RepoSense allow sorting of groups in ascending and descending order.

The actual reason for it is because now the reversing logic is now done within the descending order's code itself to simplify the complexity brought upon by the new sorting option.

eugenepeh:
>In v_summary.js, there is a filterSortReverse variable that was used to
reverse the summary charts in the previous versions of RepoSense.

Unused codes was found in v_summary.js; specifically codes associated with the variable, filterSortReverse. They are remnants left behind by the improvement of sort controls introduced in #618. 

>The variable is now unused and redundant as the reversing logic is
now done by the sorting of groups in descending order.

The codes was originally bind to the reverse checkbox in the older version of RepoSense. As the checkbox were replaced by the introduction of ascending and descending options in #618, the codes are now unused and redundant. This is also partially due to the complexity brought upon by new sorting group, the reversing has to be done within the code of descending's logic, rendering those codes not reusable.

PR #737, opened 2019-06-10
[#185] Author: Use practical display names

### LINKED ISSUE #185: Use realistic person names for test repos
Current:
image

Problem: It's unlikely that real use case will be for such short names. Using this kind of _too simple_ test cases can cause bugs to go unnoticed (e.g., a bug in handling spaces or special chars in a name) and also can cause the actual UI to look different from the UI we see in test cases (e.g., text overflows due to long names).

Suggestion: Use a variety of names that cover various equivalence partitions of the test input. e.g., names with a slash (`Ravi s/o Ramesh`)

jylee-git:
```
Truncated names are set for author display name in author-config.csv 
file.

This may cause some bugs to go unnoticed when different variations of 
names are used, such as long names or names with special characters.

Let's modify "Author's Display Name" column to use practical display 
names instead of using truncated names to fulfill the equivalence 
partition of the test input.
```

Resolves #185


jylee-git:
On hold while waiting for solutions to #738

fzdy1914:
It seems that we can close this for now？

jylee-git:
Yes i think we can close it.

PR #739, opened 2019-06-11
[#738] CsvParser: fix incorrect splitting of literals commas

### LINKED ISSUE #738: Issues when Display name contains commas in author-config.csv
Commas in name causes issue in Displaying the name in report.

In excel, when user keys in names with commas (e.g.: `Tang Tjun Chii, Zachary`): -
image

When viewed as a raw file, `"` gets added in automatically when user did not do so: -
image

When running RepoSense using the above `author-config.csv` and displaying it in the report: -
image

Found this issue when dealing with #185

jylee-git:
```
We parse CSV by using comma as a delimiter.

When commas are used in a column as value, by enclosing double quotes,
they are mistakenly treated as a delimiter. This results in incorrect
parsing of values.

Let's use third-party CSV parser library, Apache Commons CSV to
handle the intricacies of CSV parsing in a correct and clean
manner.
```

Resolves #738


author-config.csv
```diff
@@ -1,9 +1,9 @@
 Repository's Location,Branch,Author's GitHub ID,Author's Email,Author's Display Name,Author's Git Author Name,Ignore Glob List
-https://github.com/reposense/testrepo-Beta.git,master,nbriannl,,Nbr,,
-https://github.com/reposense/testrepo-Beta.git,master,zacharytang,zacharytang.tc@gmail.com,Zac,,src/main**
-https://github.com/reposense/testrepo-Beta.git,master,April0616,,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting,
-https://github.com/reposense/testrepo-Beta.git,master,CindyTsai1,,Cin,YuHsuan,src/test**
-https://github.com/reposense/testrepo-Beta.git,add-config-json,nbriannl,,Nbr,,
-https://github.com/reposense/testrepo-Beta.git,add-config-json,zacharytang,zacharytang.tc@gmail.com,Zac,,src/main**
-https://github.com/reposense/testrepo-Beta.git,add-config-json,April0616,,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting,
-https://github.com/reposense/testrepo-Beta.git,add-config-json,CindyTsai1,,Cin,YuHsuan,src/test**
+https://github.com/reposense/testrepo-Beta.git,master,nbriannl,,Labayna Neil Brian Narido,,
```
fzdy1914: I don't think it is a good idea to use one's full name who is not in our team. Please create a new test branch and add related test cases (CsvParser).

jylee-git: My bad for this. But for the default `author-config.csv` file, will it be good if we still modify the display names to be more practical (but using random names instead)?

fzdy1914: I am not sure, my idea is just to leave it there. Can let the seniors decide.

jylee-git: Ok. I'll revert changes to `author-config.csv` to the original state.

fzdy1914:
Why close this PR? We do a force push to the branch if the edit is huge but not creating a separate one.

jylee-git:
Closed by mistake. Meant to close #737 and not this

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -71,6 +74,16 @@
     private static final List<Author> AUTHOR_CONFIG_SPECIAL_CHARACTER_AUTHORS = Arrays.asList(
             FIRST_SPECIAL_CHARACTER_AUTHOR, SECOND_SPECIAL_CHARACTER_AUTHOR, THIRD_SPECIAL_CHARACTER_AUTHOR);
 
+    private static final String FIRST_SPECIAL_CHARACTER_DISPLAY_NAME = "\"Tay Fan Gao, Douya\"";
```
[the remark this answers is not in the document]
fzdy1914: If so, we should remove this `"`.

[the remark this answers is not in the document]
jylee-git: Yes, `"` will be displayed for names with commas, which is necessary, Unless if we explicitly filter out `"` in the front end report view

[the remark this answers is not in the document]
fzdy1914: I think we should and we are able to remove this `"`. This `"` is not added by the user so it is weird to be displayed. You may detect if the field contains `,`. If so, you trim the `"`.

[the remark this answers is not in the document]
jylee-git: Ok, made the changes.

src/main/java/reposense/parser/AuthorConfigCsvParser.java
```diff
@@ -110,6 +110,7 @@ private static void setEmails(AuthorConfiguration config, Author author, List<St
      * Otherwise, use github id from {@code author}.
      */
     private static void setDisplayName(AuthorConfiguration config, Author author, String displayName) {
+        displayName = displayName.replace("\"", "");
```
fzdy1914: The scope of this is too big. What if a user has a display name of `A"C` ?

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -58,7 +58,7 @@ public CsvParser(Path csvFilePath) throws IOException {
             String line;
 
             for (int lineNumber = 2; (line = br.readLine()) != null; lineNumber++) {
-                String[] elements = line.split(ELEMENT_SEPARATOR);
+                String[] elements = line.split(ELEMENT_SEPARATOR_REGEX);
```
fzdy1914: The trim and dealing of this should be placed in CSV parser, directly after this line. You should make sure all the elements have the same context as the user's intended one.

src/main/java/reposense/parser/AuthorConfigCsvParser.java
jylee-git: Made the changes to remove `"` only if detected at the beginning and end of an element.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -59,6 +59,11 @@ public CsvParser(Path csvFilePath) throws IOException {
 
             for (int lineNumber = 2; (line = br.readLine()) != null; lineNumber++) {
                 String[] elements = line.split(ELEMENT_SEPARATOR_REGEX);
+                for (int colNum = 0; colNum < elements.length; colNum++) {
```
fzdy1914: Can use `for(String element: elements)`

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -59,6 +59,11 @@ public CsvParser(Path csvFilePath) throws IOException {
...
+                    if (elements[colNum].startsWith("\"") && elements[colNum].endsWith("\"")) {
```
fzdy1914: Unfortunately, this detection is also incorrect. What if the origin user input is `"input"`, this `"` is ignored in our case. To solve the problem, you should study how CSV deal with `,`. A proper one should be `element.contains(",")`.  If the split text contains `,`, it is guaranteed that there are start and end `"`.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -59,6 +59,11 @@ public CsvParser(Path csvFilePath) throws IOException {
...
+                        elements[colNum] = elements[colNum].replaceAll("^\"|\"$", "");
```
fzdy1914: Using `String.substring` method should be easier in this case instead of regex.

src/main/java/reposense/parser/CsvParser.java
jylee-git: Here I want to edit the contents in `elements[]` under certain conditions. So may not be appropriate to use range based loop as `element` is just an iterative copy of the value in `elements[]`?

https://stackoverflow.com/questions/15844443/why-does-the-foreach-statement-not-change-the-element-value

src/main/java/reposense/parser/CsvParser.java
jylee-git: Noted. Made the changes.

src/main/java/reposense/parser/CsvParser.java
fzdy1914: Right，forgot this.

src/test/resources/CsvParserTest/authorconfig_specialCharacter_test.csv
```diff
@@ -1,4 +1,4 @@
 Repository's Location,Branch,Author's GitHub ID,Author's Email,Author's Display Name,Author's Git Author Name,Ignore Glob List
-https://github.com/reposense/testrepo-Beta.git,master,‘Processed�‘Cooked�,,,,
-https://github.com/reposense/testrepo-Beta.git,master,(codeeong),,,,
+https://github.com/reposense/testrepo-Beta.git,master,‘Processed�‘Cooked�,,"Tay Fan Gao, Douya",,
+https://github.com/reposense/testrepo-Beta.git,master,(codeeong),abc@xyz.eu,"""""Tora, S/O,"" Doyua, T.""",,
```
eugenepeh: create a separate test file for this as in the event of a fail arise, it may be hard to troubleshoot the causation.

jylee-git: You mean to create a seperate file just to test commas and double-quotes?

eugenepeh: yes, also just to add on... since your this algorithm will affect all the columns, you should include tests for multiple columns too.

jylee-git: Made a new test file to cater to commas and double quotes.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -58,7 +61,14 @@ public CsvParser(Path csvFilePath) throws IOException {
             String line;
 
             for (int lineNumber = 2; (line = br.readLine()) != null; lineNumber++) {
-                String[] elements = line.split(ELEMENT_SEPARATOR);
+                String[] elements = line.split(ELEMENT_SEPARATOR_REGEX);
+                for (int colNum = 0; colNum < elements.length; colNum++) {
+                    Matcher matcher = DOUBLEQUOTE_DETECTOR_REGEX.matcher(elements[colNum]);
+                    if (matcher.find()) {
+                        elements[colNum] = elements[colNum].substring(1, elements[colNum].length() - 1);
```
yong24s: You are removing `"` only? Use `StringsUtil#removeQuote`?

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -58,7 +61,14 @@ public CsvParser(Path csvFilePath) throws IOException {
             String line;
 
             for (int lineNumber = 2; (line = br.readLine()) != null; lineNumber++) {
-                String[] elements = line.split(ELEMENT_SEPARATOR);
+                String[] elements = line.split(ELEMENT_SEPARATOR_REGEX);
+                for (int colNum = 0; colNum < elements.length; colNum++) {
+                    Matcher matcher = DOUBLEQUOTE_DETECTOR_REGEX.matcher(elements[colNum]);
```
yong24s: Can explain what is `DOUBLEQUOTE_DETECTOR_REGEX` matching?

jylee-git: Matching fields that contain either double quotes or commas. Fields containing `"` or `,` will end up having double-quotes added to the beginning and the end of field.

So  `Ben " S`  --> `"Ben "" S"` and also `Ben , S` --> `"Ben , S"`

yong24s: Then do you really need regex for this?

i.e. Can directly use `StringsUtil#removeQuote`, if required test the output.

jylee-git: Nope no need for regex. Thought too much about filtering off those characters.

yong24s:
Suggested changes for commit message.

- Second paragraph: Generalize a bit, as this PR is fixing a problem in CsvParser.
- Third paragraph: Talk about how is regex being used.

```
We parse CSV by using comma as a delimiter.

When commas are used in a column as value, by enclosing double quotes,
they are mistakenly treated as a delimiter. This results in incorrect
parsing of values.

Let's use regex to properly split each column by matching commas that
are not within double quotes.
```

jylee-git:
> Suggested changes for commit message.
> 
> * Second paragraph: Generalize a bit, as this PR is fixing a problem in CsvParser.
> * Third paragraph: Talk about how is regex being used.

Noted and thanks for the feedback

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -71,6 +77,27 @@
     private static final List<Author> AUTHOR_CONFIG_SPECIAL_CHARACTER_AUTHORS = Arrays.asList(
             FIRST_SPECIAL_CHARACTER_AUTHOR, SECOND_SPECIAL_CHARACTER_AUTHOR, THIRD_SPECIAL_CHARACTER_AUTHOR);
 
+    private static final String FIRST_SPECIAL_CHARACTER_DISPLAY_NAME = "Tay Fan Gao, Douya";
```
eugenepeh: The result will still contain the double quote?

src/test/resources/CsvParserTest/authorconfig_commasAndDoubleQuotes_test.csv
```diff
@@ -0,0 +1,4 @@
+Repository's Location,Branch,Author's GitHub ID,Author's Email,Author's Display Name,Author's Git Author Name,Ignore Glob List
+https://github.com/reposense/testrepo-Beta.git,master,‘Processed�‘Cooked�,,"Tay Fan Gao, Douya","Tay Fan Gao, Douya ""SOC, Y2S1""",
```
eugenepeh: the special symbols in the other test should not be included here: `‘Processed�‘Cooked�`

src/test/java/reposense/parser/CsvParserTest.java
jylee-git: Should not contain double quotes. Resulting display name should be `Tay Fan Gao, Douya`

eugenepeh: Sorry, actually meant to tag the following line. Let's continue the discussion over there

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -71,6 +77,30 @@
     private static final List<Author> AUTHOR_CONFIG_SPECIAL_CHARACTER_AUTHORS = Arrays.asList(
             FIRST_SPECIAL_CHARACTER_AUTHOR, SECOND_SPECIAL_CHARACTER_AUTHOR, THIRD_SPECIAL_CHARACTER_AUTHOR);
 
+    private static final Author FIRST_COMMAS_AND_DOUBLEQUOTES_AUTHOR = new Author("ProcessedCooked");
+    private static final Author SECOND_COMMAS_AND_DOUBLEQUOTES_AUTHOR = new Author("codeeong");
+    private static final Author THIRD_COMMAS_AND_DOUBLEQUOTES_AUTHOR = new Author("jordancjq");
+    private static final String FIRST_COMMAS_AND_DOUBLEQUOTES_DISPLAY_NAME = "Tay Fan Gao, Douya";
+    private static final String SECOND_COMMAS_AND_DOUBLEQUOTES_DISPLAY_NAME = "\"\"Tora, S/O,\" Doyua, T.\"";
```
eugenepeh: I am a little confuse on how it came to the conclusion that the result, at the front for example, only few of all double quotes were captured. Whats the rationale for it?

jylee-git: In MS Excel, every double quote you entered will be "translated" to **two** double quotes in raw format.
image

So `""Tora, S/O," Douya, T."` in excel means `"""""Tora, S/O,"" Douya, T."""` in RAW form.

- Because certain characters are added, such as `,` and `"`, the entire string will be enclosed with a double-quote.
- For each double-quote entered in Excel, it will be **two** double quotes in RAW form.

damithc:
Why don't we use a third-party library for handling CSV files?

eugenepeh:
>Why don't we use a third-party library for handling CSV files?

@yong24s was it due to some limitation last time?

yong24s:
Mainly because its overkill and doesn't simplify our code.

damithc:
> Mainly because its overkill and doesn't simplify our code.

We can reconsider, especially if the library handles issues like _comma-within-cell_ transparently.

yong24s:
@jylee-git 

[Apache CSVParser](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVParser.html
) looks good. You know of any alternatives?

We can use it directly in our `CsvParser` class.
Also, need to find out which formats `i.e.` `RFC4180` or `EXCEL` suits us best.

jylee-git:
> @jylee-git
> 
> [Apache CSVParser](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVParser.html) looks good. You know of any alternatives?
> 
> We can use it directly in our `CsvParser` class.
> Also, need to find out which formats `i.e.` `RFC4180` or `EXCEL` suits us best.

I think apache looks good. Will play around and implement it.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -51,105 +51,106 @@ public CsvParser(Path csvFilePath) throws IOException {
...
         return false;
     }
 
     /**
-     * Checks that {@code position} in within the range of {@code element} array and
-     * value in {@code position} is not empty.
+     * Returns the value of {@code record} at {@code colNum}.
      */
-    private boolean containsValueAtPosition(final String[] elements, int position) {
-        return elements.length > position && !elements[position].isEmpty();
```
yong24s: Is it safe to access a non-existing column? i.e. `record.get(-1)`

Maybe we can use [`CsvRecord#isConsistent`](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVRecord.html#isConsistent--) to replicate `elements.length > position` behavior?

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -51,105 +51,106 @@ public CsvParser(Path csvFilePath) throws IOException {
...
-     * Returns the value of {@code position} if it is in {@code element} and not empty.
-     * Otherwise returns the {@code defaultValue}.
+     * Returns the values in {@code record} as a list with the {@link CsvParser#OVERRIDE_KEYWORD} prefix removed.
+     * Returns an empty list if {@code record} at {@code colNum} is empty.
      */
-    protected String getValueInElement(final String[] elements, int position, String defaultValue) {
-        return (containsValueAtPosition(elements, position)) ? elements[position] : defaultValue;
-    }
+    protected List<String> getAsListWithoutOverridePrefix(final CSVRecord record, int colNum) {
+        if (colNum >= record.size()) {
```
yong24s: Same comment as `containsValueAtPosition`.

Should have a method for checking.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -51,105 +51,106 @@ public CsvParser(Path csvFilePath) throws IOException {
...
     }
 
     /**
-     * Checks if {@code position} in {@code element} is prefixed with the override keyword.
+     * Returns true if the {@code record} at {@code colNum} is prefixed with the override keyword.
      */
-    protected boolean isElementOverridingStandaloneConfig(final String[] elements, int position) {
-        return (containsValueAtPosition(elements, position)) && elements[position].startsWith(OVERRIDE_KEYWORD);
+    protected boolean isElementOverridingStandaloneConfig(final CSVRecord record, int colNum) {
+        return colNum < record.size() && record.get(colNum).startsWith(OVERRIDE_KEYWORD);
```
yong24s: Same comment for error checking.

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -46,31 +48,23 @@ public RepoConfigCsvParser(Path csvFilePath) throws IOException {
...
+        boolean isFormatsOverriding = isElementOverridingStandaloneConfig(record, FILE_FORMATS_POSITION);
+        List<Format> formats = Format.convertStringsToFormats(
+                getAsListWithoutOverridePrefix(record, FILE_FORMATS_POSITION));
 
-        boolean isIgnoreGlobListOverriding = isElementOverridingStandaloneConfig(elements, IGNORE_GLOB_LIST_POSITION);
-        if (isIgnoreGlobListOverriding) {
-            removeOverrideKeywordFromElement(elements, IGNORE_GLOB_LIST_POSITION);
-        }
-        List<String> ignoreGlobList = getManyValueInElement(elements, IGNORE_GLOB_LIST_POSITION);
+        boolean isIgnoreGlobListOverriding = isElementOverridingStandaloneConfig(record, IGNORE_GLOB_LIST_POSITION);
```
yong24s: booleans not in use anymore?

jylee-git: It's still being used in an input parameter upon initializing `RepoConfiguration` 

https://github.com/reposense/RepoSense/blob/4c915c422e30b20f8c1674c036dc2c7a984af162/src/main/java/reposense/parser/RepoConfigCsvParser.java#L74-L77

src/main/java/reposense/parser/CsvParser.java
jylee-git: I considered using [`CsvRecord#isConsistent`](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVRecord.html#isConsistent--) but decided not to due to the test case [CsvParserTest#merge_emptyLocation_success](https://github.com/reposense/RepoSense/blob/4c915c422e30b20f8c1674c036dc2c7a984af162/src/test/java/reposense/parser/CsvParserTest.java#L256) .

Apparently the test file `repo-config.csv` is corrupted, where it's supposed to be 6 columns normally but this only has 5 columns (missing ignore commit hash column).
https://github.com/reposense/RepoSense/blob/4c915c422e30b20f8c1674c036dc2c7a984af162/src/test/resources/CsvParserTest/repoconfig_merge_empty_location_test/repo-config.csv .

Should I fix the CSV file in question?

yong24s: For such case, just update the csv. We may have missed it 😨.

jylee-git: Ok. Will use `isConsistent` method and fix the CSV file.

jylee-git: > Is it safe to access a non-existing column? i.e. `record.get(-1)`

In any case, is there a reason why we need to take `get(-1)`  into account?

yong24s: Its just an illustration, for cases where columns are missing. 
i.e. `record.get(5)` but there's only `4` columns.

`CsvRecord#isConsistent` should be an elegant approach to handle this.

Just make sure not to change the behavior of the parser, we skip the erroneous lines without terminating the program.

jylee-git: > Just make sure not to change the behavior of the parser, we skip the erroneous lines without terminating the program.

Yes that's the current behavior

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -51,105 +51,107 @@ public CsvParser(Path csvFilePath) throws IOException {
...
 
-        try (BufferedReader br = new BufferedReader(new FileReader(csvFilePath.toFile()))) {
-            // Skip first line, which is the header row
-            br.readLine();
-            String line;
-
-            for (int lineNumber = 2; (line = br.readLine()) != null; lineNumber++) {
-                String[] elements = line.split(ELEMENT_SEPARATOR);
+        try {
+            Reader csvReader = new FileReader(csvFilePath.toFile());
```
yong24s: `FileReader` implements `AutoCloseable`.

Should use try-with-resources:
 `try (Reader csvReader = new FileReader(csvFilePath.toFile()))`

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -51,105 +51,107 @@ public CsvParser(Path csvFilePath) throws IOException {
...
-     * Returns the value of {@code element} at {@code position} as a {@code List},
-     * delimited by {@code COLUMN_VALUES_SEPARATOR} if it is in {@code element} and not empty.
-     * Otherwise returns an empty {@code List}.
-     */
-    protected List<String> getManyValueInElement(final String[] elements, int position) {
-        if (!containsValueAtPosition(elements, position)) {
+        if (rawValue.isEmpty()) {
             return Collections.emptyList();
         }
+        return Arrays.stream(rawValue.split((COLUMN_VALUES_SEPARATOR))).map(String::trim).collect(Collectors.toList());
```
yong24s: Can use `getAsList`.

jylee-git: I don't really know how to do that here. the `rawValue` is `String` and not `CSVRecord`, so I can't just call `getAsList` because of the override keyword.

yong24s: 🤔 Hmm, seems like its not so easy and direct.

There are methods that rely on `get` and its method signature `(final CSVRecord record, int colNum)`.

Let me think about it first. **Don't do anything drastic yet**.


```
protected List<String> getAsList(final CSVRecord record, int colNum) {
>    if (get(record, colNum).isEmpty()) {
>        return Collections.emptyList();
>    }
>    return Arrays.stream(get(record, colNum).split(COLUMN_VALUES_SEPARATOR))
>            .map(String::trim)
>            .collect(Collectors.toList());
}

protected List<String> getAsListWithoutOverridePrefix(final CSVRecord record, int colNum) {
    String rawValue = isElementOverridingStandaloneConfig(record, colNum)
            ? get(record, colNum).replaceFirst(OVERRIDE_KEYWORD, "")
            : get(record, colNum);

>    if (rawValue.isEmpty()) {
>        return Collections.emptyList();
>    }
>    return Arrays.stream(rawValue.split((COLUMN_VALUES_SEPARATOR)))
>		    .map(String::trim)
>		    collect(Collectors.toList());
}
```

yong24s: @jylee-git 

You can extract the similar code chunks/SLAP first.

jylee-git:
```java
protected List<String> getAsListWithoutOverridePrefix(final CSVRecord record, int colNum) {
    List<String> data = getAsList(record, colNum);
    if (isElementOverridingStandaloneConfig(record, colNum)) {
        data.set(0, data.get(0).replaceFirst(OVERRIDE_KEYWORD, ""));
    }
    return data;
}
```

?

yong24s: Nice! 👍

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -22,14 +25,11 @@
     protected static final String COLUMN_VALUES_SEPARATOR = ";";
     protected static final Logger logger = LogsManager.getLogger(CsvParser.class);
 
-    private static final String ELEMENT_SEPARATOR = ",";
     private static final String OVERRIDE_KEYWORD = "override:";
     private static final String MESSAGE_UNABLE_TO_READ_CSV_FILE = "Unable to read the supplied CSV file.";
-    private static final String MESSAGE_MALFORMED_LINE_FORMAT = "Warning! line %d in CSV file, %s, is malformed.\n"
-            + "Content: %s";
+    private static final String MESSAGE_MALFORMED_LINE_FORMAT = "Warning! line %d in CSV file, %s, is malformed.";
```
eugenepeh: Is this the downside of using the api? We can't provide the content that is malformed?

src/test/resources/repoconfig_merge_test/repo-config.csv
```diff
@@ -1,2 +1,2 @@
-Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config
-https://github.com/reposense/testrepo-Beta.git,master,,collated**,
+Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config,Ignore Commits List
+https://github.com/reposense/testrepo-Beta.git,master,,collated**,,
```
eugenepeh: also, the csv wouldn't be parsed if the column numbers do not match?

src/main/java/reposense/parser/CsvParser.java
jylee-git: Not that it's not possible, but printing the content will also print lots of irrelevant details. Printing it results in this : -

`CSVRecord [comment=null, mapping={Repository's Location=0, Branch=1, File formats=2, Ignore Glob List=3, Ignore standalone config=4, Ignore Commits List=5}, recordNumber=1, values=[https://github.com/reposense/testrepo-Alpha.git, master, , , , 2fb6b9b2dd9fa40bf0f9815da2cb0ae8731436c7;c5a6dc774e22099cd9ddeb0faff1e75f9cf4f151;cd7f610e0becbdf331d5231887d8010a689f87c7;768015345e70f06add2a8b7d1f901dc07bf70582]]`

eugenepeh: Hmm... it would be best to if we can, because line number alone can cause some discrepancy in understanding 

E.g. some may think the line number begins from the first line of entry (excluding the header)

src/test/resources/repoconfig_merge_test/repo-config.csv
jylee-git: Yes

src/main/java/reposense/parser/CsvParser.java
jylee-git: Made the changes

build.gradle
```diff
@@ -32,6 +32,7 @@ dependencies {
     implementation  group: 'org.apache.ant', name: 'ant', version: '1.10.3'
```
eugenepeh: Just a minor thing, do help me arrange this implementation together with the others below.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -51,105 +52,113 @@ public CsvParser(Path csvFilePath) throws IOException {
...
+     * Returns an empty list if {@code record} at {@code colNum} is empty.
      */
-    protected String getValueInElement(final String[] elements, int position, String defaultValue) {
-        return (containsValueAtPosition(elements, position)) ? elements[position] : defaultValue;
+    protected List<String> getAsListWithoutOverridePrefix(final CSVRecord record, int colNum) {
+        List<String> data = getAsList(record, colNum);
+        if (isElementOverridingStandaloneConfig(record, colNum)) {
+            data.set(0, data.get(0).replaceFirst(OVERRIDE_KEYWORD, ""));
+        }
+        data.removeIf(String::isEmpty);
```
yong24s: The only way to have an empty string is caused by the replacement in line 135?

If so, its better to move this line to the if block,

jylee-git: Made the changes

Issue #680, opened 2019-05-17
Error message of Vue


fzdy1914:
I found a error message when using developer tool of Vue for our report.
![image](https://user-images.githubusercontent.com/35621726/57942729-d3b20480-7904-11e9-8ce1-6e58ed0d8a0e.png)

PR #740, opened 2019-06-11
[#666] CodeView: load un-authored code only when toggled

### LINKED ISSUE #666: Code view fails to load
The code view of the second row [here](https://nus-cs3281.github.io/2019-dashboard/#search=jacob&sort=searchPath&sortWithin=name&since=2018-12-01&until=2019-05-14&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=false) fails to load for some reason

image

jamessspanggg:
Fixes #666 

```
Viewing the changes in code of an author causes the app to 
render all authored and un-authored segments within modified files.
A small change in large file will cause the entire file to be preloaded.

When an author has small changes done to a lot of large files, it will
lead to an unresponsive page as loading the large un-authored segments
will affect the rendering speed. Furthermore, the preloading of all 
un-authored code is not necessary as users might only view the 
un-authored code under certain circumstances. (e.g. when the user 
does not understand why the author added the authored code)

Let's modify the app to only load the un-authored code when the 
user toggles the 'Show untouched code' icon.
```


jamessspanggg:
Not fixing the codacy camelCase error to standardise the naming of vue components (v_xxx).

fzdy1914:
image
One error message shown in the console.

jamessspanggg:
> One error message shown in the console.

It is due to the following line where it tries to bind a key on a template.
Screen Shot 2019-06-11 at 7 43 13 PM

This error is not related to my PR. Should I address this in another PR?

fzdy1914:
> This error is not related to my PR. Should I address this in another PR?

Never mind, I found I have already raised an issue about this #680. Just fix it in another PR.

frontend/src/index.pug
```diff
@@ -206,6 +206,9 @@ html
     vuetemplate#v_zoom
       include tabs/zoom.pug
 
+    vuetemplate#v_segment
+      include segment.pug
```
fzdy1914: Maybe `tabs/segment` will be better? I think @ongspxm is planning to extract more template into its own file. And this `segment` belongs to ` tab`.

frontend/src/index.pug
```diff
@@ -272,21 +275,15 @@ html
...
-                  template(v-for="segment in file.segments")
-                    .segment(v-bind:class="{ untouched: !segment.authored, active: segment.lines.length<5 }")
-                      .closer(v-if="!segment.authored && segment.lines.length>4", onclick="toggleNext(this)")
-                        i.fas.fa-ellipsis-h.icon
-                        span.tooltiptext.tooltiptext--close Show untouched code
-                        span.tooltiptext.tooltiptext--open Hide untouched code
-                      .code(v-for="(line, index) in segment.lines")
-                        .line-number {{ segment.lineNumbers[index] + "\n" }}
-                        .line-content {{ line + "\n" }}
+                template(v-for="segment in file.segments")
```
fzdy1914: I think we don't need a `template` here, as `v_segment' itself is a template. A div is enough.

frontend/src/static/js/v_segment.js
```diff
@@ -0,0 +1,16 @@
+window.vSegment = {
+  props: ['segment', 'path'],
+  template: window.$('v_segment').innerHTML,
+  data() {
+    return {
+      loaded: this.segment.authored || this.segment.lines.length < 5,
```
fzdy1914: I don't think we need to preload these segment with lines less than 5. Since the segment is small, loading a 5-line segment will not take much time that the user will even notice. However, as a whole, loading many of these small segments takes great time.

frontend/src/segment.pug
```diff
@@ -0,0 +1,9 @@
+.segment(v-bind:class="{ untouched: !segment.authored, active: segment.lines.length<5 }")
+  .closer(v-if="!segment.authored && segment.lines.length>4", onclick="toggleNext(this)", v-on:click="loadCode()")
```
fzdy1914: This may be abstracted to `onclick="toggleNext(this); loaded = true"` so we do not need two listener for `on click` method. Or put `toggleNext(this)` inside `loadCode()`.

frontend/src/static/js/v_segment.js
jamessspanggg: The reason behind preloading segments with lines less than 5 is to display it together with the authored code. Since there is less than 5 lines of code in the un-authored segment, it is better to display together with the authored segment instead of requiring to toggle it just to see those few lines of un-authored segment. 

An example would be the following:
Screen Shot 2019-06-11 at 11 38 14 PM
It will be inconvenient for the user to toggle the icon just to see line 416.

fzdy1914: I see. I didn't notice that. Just ignore this comment.

frontend/src/static/js/v_authorship.js
```diff
@@ -316,4 +316,7 @@ window.vAuthorship = {
     this.initiate();
     this.setInfoHash();
   },
+  components: {
+    v_segment: window.vSegment,
```
fzdy1914: Not sure if this is necessary.

jamessspanggg: Yes I do think it is necessary as the segment is a part of the authorship page. Also, it opens a path to pass data (via props) from the authorship component to the segment component.

jamessspanggg: Correct me if I am wrong because I am still new to Vue.

fzdy1914: I think you are correct, never mind. The way we use Vue in RepoSense is a bit different so I am not sure also.

frontend/src/static/js/v_segment.js
```diff
@@ -0,0 +1,17 @@
...
+    };
+  },
+  methods: {
+    loadCode() {
+      if (!this.loaded) {
```
eugenepeh: is the condition necessary if its going to be true either way?

frontend/src/static/js/v_authorship.js
eugenepeh: @ongspxm for your input

frontend/src/static/css/style.scss
```diff
@@ -425,6 +425,7 @@ header{
                     .hljs{
                         // overwrite hljs library
                         display:inherit;
+                        padding:0;
```
eugenepeh: why is this change necessary?

jamessspanggg: It was because `v-hljs` was moved within `segment`, and the `hljs` library has an in-built padding value which causes paddings in between each `segment` as such:
Screen Shot 2019-06-18 at 11 42 35 PM

So the reason for overwriting the padding is to eliminate the spacing between each segment.

frontend/src/static/js/v_authorship.js
ongspxm: yeah, the `window.vSegment` is an JS object, not an actual vue object. Legacy problems, cos last time we not big enough to put everything as Vue component. You can ask for permission to rewrite the frontend, and start shifting everything into their own component. cos it is getting quite big now.

PR #743, opened 2019-06-12
[#598] Report: show error message summary

### LINKED ISSUE #598: Use better error display in frontend report
```
The generated report uses a generic placeholder for repositories that
failed analysis.

The placeholder looks similar to a blank rampchart, which may be
overlooked as it blends in with other rampcharts.

Also, the placeholder does not show useful errors to `Repo Owners` [1],
which may be useful for troubleshooting, as they may not necessary have
the CLI logs.

[1] CS2103/CS2113/CS3281 students
```

jylee-git:
```
Repos that fail to be cloned or branched will be blended with the 
other successfully analyzed repos with a failure message.

Users may fail to differentiate between the repos that failed the 
analysis and the ones that did not.

Let's collate the list of repos that experienced issues during the 
analysis and display them in a message box on top of the report 
view.
```

Fixes #598


frontend/src/index.pug
```diff
@@ -127,6 +127,14 @@ html
             label
               input(type="checkbox", v-model="filterBreakdown").mui-checkbox.summary-picker__checkbox
               span breakdown by file format
+        #error-message-box(v-if="errorMessages.length")
+          .error-message-box
+            #error-message-box__close-button(onclick="this.parentElement.style.display='none'") &times
```
yong24s: `&` -> `&amp;`

src/main/java/reposense/report/ErrorSummary.java
```diff
@@ -0,0 +1,38 @@
...
+public class ErrorSummary {
+    private static ErrorSummary instance = null;
+    private static List<Map<String, String>> errorList = new ArrayList<>();
+
+    public static ErrorSummary getInstance() {
+        if (instance == null) {
+            instance = new ErrorSummary();
+        }
+        return instance;
+    }
```
yong24s: Why Singleton?

It's only used in `ReportGenerator` right?

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -107,23 +115,35 @@ public static void generateReposReport(List<RepoConfiguration> configs, String o
...
-            analyzeRepos(outputPath, repoLocationMap.get(clonedRepoLocation), repoCloner.getCurrentRepoDefaultBranch());
+            analyzeRepos(outputPath, repoLocationMap.get(clonedRepoLocation),
+                    repoCloner.getCurrentRepoDefaultBranch());
+            successfullyClonedRepos.add(clonedRepoLocation);
         }
+
+        // handles configs that failed to clone and clean up the cloned repo directory.
+        handleCloningFailed(configs.stream()
+                .filter(config -> !successfullyClonedRepos.contains(config.getLocation()))
+                .collect(Collectors.toList()));
```
yong24s: Is it simpler to store a list of `RepoConfiguration` that failed analysis then pass it to `handleCloningFailed()`

src/main/java/reposense/report/ErrorSummary.java
jylee-git: I was thinking that this feature could be extended to report other issues, such as invalid file name that couldn't be analyzed, etc

frontend/src/index.pug
jylee-git: I intended to display an `X` and not ampersand.

But do you mean to add the semicolon `;` right after the entity?

src/main/java/reposense/report/ReportGenerator.java
jylee-git: I think doing so may be complicated as well due to the nature of asynchronous cloning process.

Assuming that I maintain a list of uncloned `RepoConfiguration` by initializing it with all `configs`. Every time a `RepoLocation` has been successfully cloned, I'll need to get a list  of `RepoConfiguration` that's successfully cloned and remove from the list of uncloned `RepoConfiguration`. Then only at the end I have a final list of `RepoConfiguration` that failed to clone and pass it to `handleCloningFailed()`.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -154,7 +174,7 @@ private static void analyzeRepos(String outputPath, List<RepoConfiguration> conf
             } catch (RuntimeException e) {
                 logger.log(Level.SEVERE, String.format(MESSAGE_BRANCH_DOES_NOT_EXIST,
                         config.getBranch(), config.getLocation()), e);
-                generateEmptyRepoReport(repoReportDirectory.toString());
+                handleBranchingFailed(config);
```
yong24s: We may need to create more specific exceptions.

So you can just catch and add to `ErrorSummary`. This way you can handle them generally.

Also, we can then store `LOG_ERROR_CLONING`, `LOG_BRANCH_DOES_NOT_EXIST` and each specific error messages in its exception class.

Start by creating two exceptions? 

In the future, we can group them as print only, clean up, etc.

yong24s: if this is difficult then let's do it in another PR.

Its because `ProcessRunner` throws general `RuntimeException`. If you were to create some specific git exception, we need tests too.

frontend/src/static/css/style.scss
```diff
@@ -104,6 +104,47 @@ header{
...
+        .error-message-box {
+            padding: 15px;
+            border-radius: 4px;
+            background-color: #F8D7DA;
+            border-color: #F5C6CB;
+            color: #721C24;
+            margin-bottom: 15px;
+            text-align: left;
+        }
+        #error-message-box__close-button {
```
fzdy1914: We are using `scss`, can you look into how other CSS sections deal with parent and child labels? Like [this](https://github.com/reposense/RepoSense/blob/master/frontend/src/static/css/style.scss#L151-L188)

src/main/java/reposense/report/CommitReportJson.java
```diff
@@ -21,28 +19,6 @@
     private final Map<Author, Float> authorContributionVariance;
     private final Map<Author, String> authorDisplayNameMap;
 
-    /**
```
fzdy1914: This will conflict with #777,  can we have a discussion on how to deal with no author error?

jylee-git: I think repos (with no preset author in config files) that has no commits on the specified dates (--> no authors found) should not be an error, but rather have a placeholder stating that `no commits were found during this time period`.

src/main/java/reposense/report/ReportGenerator.java
jylee-git: I think it may be hard to do it at this point, as PR #765 is revamping the entire cloning process. 
Should we wait for #765 to merge in before working on individual exception classes in a future PR?

frontend/src/static/js/api.js
```diff
@@ -37,6 +37,12 @@ window.api = {
           window.app.untilDate = data.untilDate;
           window.app.repoSenseVersion = data.repoSenseVersion;
 
+          data.errorList.forEach((errorBlock) => {
+            const repoName = `${errorBlock.repoName}`;
+            const errorMessage = `${errorBlock.errorMessage}`;
+            window.app.errorMessages.push([repoName, errorMessage]);
```
fzdy1914: Use a dict or an object instead of a list for `errorBlock`

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -229,22 +254,51 @@ private static void updateAuthorList(RepoConfiguration config) throws NoAuthorsW
...
-    */
+     * Generates a report at the {@code repoReportDirectory}.
+     */
     public static void generateEmptyRepoReport(String repoReportDirectory, String displayName) {
         CommitReportJson emptyCommitReportJson = new CommitReportJson(displayName);
         FileUtil.writeJsonFile(emptyCommitReportJson, getIndividualCommitsPath(repoReportDirectory));
         FileUtil.writeJsonFile(Collections.emptyList(), getIndividualAuthorshipPath(repoReportDirectory));
     }
 
+    private static void generateIndividualRepoReport(
```
fzdy1914: Why reorder this part?

jylee-git: I reordered it so that the methods that generate the reports are all clustered below

fzdy1914: OK

jylee-git:
@fzdy1914 I have failing codacy tests being `'cy' is not defined (no-undef)`. You have any idea what's going on?

yong24s:
@jylee-git 

Don't worry about Codacy. Its using a different check as our eslint.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -239,22 +264,51 @@ private static void updateAuthorList(RepoConfiguration config) throws NoAuthorsW
...
+     */
+    private static void handleBranchingFailed(RepoConfiguration failedRepoConfig) {
+        ErrorSummary.getInstance().addErrorMessage(failedRepoConfig.getDisplayName(),
+                String.format(LOG_BRANCH_DOES_NOT_EXIST, failedRepoConfig.getBranch()));
+        failedRepoConfigsList.add(failedRepoConfig);
+    }
+
+    /**
+     * Adds all {@code failedRepoConfigs} into the list of failed repos and logs into the list of errors in the
+     *  summary.
```
yong24s: Extra space between `*  summary.`.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -239,22 +264,51 @@ private static void updateAuthorList(RepoConfiguration config) throws NoAuthorsW
...
     }
 
-    private static void generateIndividualRepoReport(
-            CommitContributionSummary commitSummary, AuthorshipSummary authorshipSummary, String repoReportDirectory) {
-        CommitReportJson commitReportJson = new CommitReportJson(commitSummary, authorshipSummary);
-        FileUtil.writeJsonFile(commitReportJson, getIndividualCommitsPath(repoReportDirectory));
-        FileUtil.writeJsonFile(authorshipSummary.getFileResults(), getIndividualAuthorshipPath(repoReportDirectory));
+    /**
+     * Adds {@code failedRepoConfig} into the collated list of failed repos and logs into the list of errors in the
+     *  summary.
```
yong24s: Extra space between `*  summary.`.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -239,22 +264,51 @@ private static void updateAuthorList(RepoConfiguration config) throws NoAuthorsW
...
+    private static void handleCloningFailed(List<RepoConfiguration> failedRepoConfigs) {
+        for (RepoConfiguration failedConfig : failedRepoConfigs) {
+            ErrorSummary.getInstance().addErrorMessage(failedConfig.getDisplayName(),
+                    String.format(LOG_ERROR_CLONING, failedConfig.getLocation()));
+            failedRepoConfigsList.add(failedConfig);
+        }
+    }
+
+    /**
+     * Remove configs from {@code repoConfigs} that failed to clone or branch out previously.
```
yong24s: `Removes configs`

src/main/java/reposense/report/ReportGenerator.java
yong24s: Yup, the PR is quite large already.

Please create an issue on this.

src/main/java/reposense/report/ReportGenerator.java
yong24s: Hmm, I was thinking that we can append failed repos to a list in the `catch` part of the code.
But, devs may forget to append when create new git exceptions.

So, your approach may be better in this case. 😃

frontend/src/static/js/v_summary.js
```diff
@@ -117,6 +117,7 @@ window.vSummary = {
       minDate: '',
       maxDate: '',
       contributionBarColors: {},
+      errorMessages: window.app.errorMessages,
```
bluein-green: Perhaps it might be better to pass `errorMessages` as a prop instead from `window.app` vue object to `vSummary`

jylee-git: This solution somehow dosen't work

eugenepeh:
Seems like this PR supercedes #790 
@jamessspanggg why is there a collision of work?

jylee-git:
> Seems like this PR supercedes #790
> @jamessspanggg why is there a collision of work?

There's no collision. #777 deals with cases where no authors are specified, and that no commits ( --> no authors) are detected during analysis; #790 changes the color of the `no authors` message to red to make it conspicuous; whereas this PR deals with error when analyzing (e.g.: cloning/branching failed)

jamessspanggg:
> Seems like this PR supercedes #790
> @jamessspanggg why is there a collision of work?

> IMO, I don't think we should treat repos that have no commits made during the specified time period be considered as error, but rather show a placeholder instead (or if need be, we can put the text `NO COMMITS FOUND WITHIN THIS PERIOD OF TIME` to be a different, conspicuous color).

_Originally posted by @jylee-git in https://github.com/reposense/RepoSense/pull/777#issuecomment-507684789_

> Yes, it shouldn't be an error.

_Originally posted by @yong24s in https://github.com/reposense/RepoSense/pull/777#issuecomment-507684789_

Author with no commits found is not considered as an error. Hence we are dealing it separately.

eugenepeh:
I see, thanks for the clarification.

frontend/cypress/tests/errorSummary_messageBox.js
```diff
@@ -0,0 +1,32 @@
+describe('error summary', () => {
```
eugenepeh: please rename this file to `chartView_errorSummary_messageBox.js`, so that all files begin with the major component's name

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -239,22 +264,51 @@ private static void updateAuthorList(RepoConfiguration config) throws NoAuthorsW
...
+                    String.format(LOG_ERROR_CLONING, failedConfig.getLocation()));
+            failedRepoConfigsList.add(failedConfig);
+        }
+    }
+
+    /**
+     * Removes configs from {@code repoConfigs} that failed to clone or branch out previously.
+     */
+    private static void removeFailedRepoConfigs(List<RepoConfiguration> repoConfigs) {
+        repoConfigs.removeAll(failedRepoConfigsList);
```
eugenepeh: instead of this, perhaps we can remove it on the go, using an iterator?
https://stackoverflow.com/questions/1196586/calling-remove-in-foreach-loop-in-java

jylee-git: Considering this approach. Meanwhile, is there a reason to use an iterator over the current approach?

eugenepeh: it is unsafe to remove elements within a for each loop, hence iterator

jylee-git: Let's use `Collection.removeIf` instead, as the underlying function implement iterator to remove the elements as well.

eugenepeh: The rationale is to remove the method, removeFailedRepoConfigs as a whole by removing elements within the for each loop you're using itself. Reason -> we try to avoid global variables whenever possible.

jylee-git: Noted. I've removed the method `removeFailedRepoConfigs` and remove the failed configs on the go.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -119,59 +125,78 @@
...
-     * Analyzes all repos in {@code configs} and generates their report.
+     * Analyzes all repos in {@code configsToAnalyze} and generates their report.
+     * Also removes {@code configsToAnalyze} that failed to analyze from {@code configs}.
      */
-    private static void analyzeRepos(String outputPath, List<RepoConfiguration> configs, String defaultBranch) {
-        for (RepoConfiguration config : configs) {
-            config.updateBranch(defaultBranch);
+    private static void analyzeRepos(String outputPath, List<RepoConfiguration> configs,
+            List<RepoConfiguration> configsToAnalyze, String defaultBranch) {
+        for (RepoConfiguration configToAnalyze : configsToAnalyze) {
```
eugenepeh: Here is not using an iterator. Using removeIf with for each loop is unsafe as mentioned.

frontend/src/static/css/style.scss
```diff
@@ -115,6 +115,54 @@ header {
...
+      &__message {
+        border-bottom: 2px solid;
+        font-size: .75rem;
+        font-weight: bolder;
+        padding-bottom: 10px;
+      }
+
+      &__failed-repo {
+        &--name {
+          font-size: .75rem;
```
eugenepeh: why don't put directly under the parent?

src/main/java/reposense/report/ReportGenerator.java
jylee-git: My bad. Forgot that I'm still getting the copy of the config by reference.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -119,59 +125,80 @@
...
         RepoCloner repoCloner = new RepoCloner();
         RepoLocation clonedRepoLocation = null;
 
-        for (RepoLocation location : repoLocationMap.keySet()) {
-            repoCloner.clone(outputPath, repoLocationMap.get(location).get(0));
+        List<RepoLocation> repoLocationList = new ArrayList<>(repoLocationMap.keySet());
+        Set<RepoLocation> successfullyClonedRepos = new HashSet<>();
+
+        // clones another repo in parallel while analyzing a cloned repo (if any)
+        for (RepoLocation location : repoLocationList) {
```
jylee-git: @eugenepeh do we need to worry about unsafe removal of `RepoConfiguration` elements here even as `RepoLocation` won't be removed?

eugenepeh: Nope, that is not necessary. As long as you can perform the removal without utilizing global variable, its fine.

frontend/src/static/css/style.scss
```diff
@@ -115,6 +115,52 @@ header {
...
+        line-height: 10px;
+
+        &:hover {
+          color: white;
+        }
+      }
+
+      &__message {
+        border-bottom: 2px solid;
+        font-size: .75rem;
```
eugenepeh: I meant why don't you put the font-size under the parent, so that all child component will inherit by default

jylee-git: Oh I thought you meant to eliminate the unnecessary nesting of `&--name`. Will make the changes

frontend/src/index.pug
```diff
@@ -126,6 +126,14 @@ html
               label
                 input(type="checkbox", v-model="filterBreakdown").mui-checkbox.summary-picker__checkbox
                 span breakdown by file format
+        #error-message-box(v-if="Object.entries(errorMessages).length")
+          .error-message-box
+            .error-message-box__close-button(onclick="this.parentElement.style.display='none'") &times;
+            .error-message-box__message The following issues occurred when analyzing the following repositories:
+            .error-message-box__failed-repo(v-for="errorBlock in errorMessages")
+              i.fas.fa-exclamation
+              span.error-message-box__failed-repo--name {{errorBlock.repoName}}
```
jamessspanggg: spaces after and before double curly braces

frontend/src/static/css/style.scss
```diff
@@ -115,6 +115,50 @@ header {
       margin-bottom: 1em;
       margin-left: 1em;
     }
+
+    .error-message-box {
+      background-color: #F8D7DA;
```
jamessspanggg: Please use custom class `mui-color` instead of specifying the color on your own.

frontend/src/static/css/style.scss
```diff
@@ -115,6 +115,50 @@ header {
...
+      border-color: #F5C6CB;
+      border-radius: 4px;
+      color: #721C24;
+      font-size: .75rem;
+      margin-bottom: 15px;
```
jamessspanggg: Please use `rem` instead of `px`

frontend/src/static/css/style.scss
```diff
@@ -115,6 +115,50 @@ header {
...
+        padding-left: .2rem;
+      }
+
+      &__failed-repo--reason {
+        padding-left: 3rem;
+      }
+    }
+
+    .fa-exclamation {
+      font-size: smaller;
```
jamessspanggg: maybe can use the global `mini-font` class?

frontend/src/index.pug
```diff
@@ -126,6 +126,14 @@ html
               label
                 input(type="checkbox", v-model="filterBreakdown").mui-checkbox.summary-picker__checkbox
                 span breakdown by file format
+        #error-message-box(v-if="Object.entries(errorMessages).length")
```
jamessspanggg: why the need of two divs?

src/main/java/reposense/report/ReportGenerator.java
eugenepeh: A little change here, lets turn this into a `for` loop instead of using a `for each` loop and let the counter start from 1 and ends at n + 1. 

Main idea here is to not have an additional `if (clonedRepoLocation != null) {` outside of the loop, so that we can save some duplicate codes as well as remove the failed configs on the go instead of making use of a `successfullyClonedRepos`.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -239,22 +266,39 @@ private static void updateAuthorList(RepoConfiguration config) throws NoAuthorsW
...
+     * Generates a report at the {@code repoReportDirectory}.
+     */
     public static void generateEmptyRepoReport(String repoReportDirectory, String displayName) {
         CommitReportJson emptyCommitReportJson = new CommitReportJson(displayName);
         FileUtil.writeJsonFile(emptyCommitReportJson, getIndividualCommitsPath(repoReportDirectory));
         FileUtil.writeJsonFile(Collections.emptyList(), getIndividualAuthorshipPath(repoReportDirectory));
     }
 
+    private static void generateIndividualRepoReport(
+            CommitContributionSummary commitSummary, AuthorshipSummary authorshipSummary, String repoReportDirectory) {
```
eugenepeh: Try to follow the same parameter ordering as existing apis. Let `String repoReportDirectory` be the first parameter

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -73,6 +75,11 @@ public void testDateRange() throws IOException, URISyntaxException, ParseExcepti
         verifyAllJson(actualFiles, FT_TEMP_DIR);
     }
 
+    @After
+    public void afterEach() {
+        ErrorSummary.getInstance().clearErrorList();
```
eugenepeh: maybe make it a before each and place it directly after `setUp()` to make it neater.

yong24s:
@jylee-git 

Heads up, there are merge conflicts.

jylee-git:
> @jylee-git
> 
> Heads up, there are merge conflicts.

Give me a while to deal with this.. The conflicts are complicated this time round

jylee-git:
Ready for review @yong24s @bluein-green

frontend/src/static/js/v_summary.js
yong24s: @jamessspanggg 

Can take a look?

jamessspanggg: Yes, you can pass the `errorMessages` as a prop using `v-bind`.

frontend/src/static/css/style.scss
```diff
@@ -115,6 +115,50 @@ header {
...
+      }
+
+      &__failed-repo--name {
+        font-weight: bolder;
+        padding-left: .2rem;
+      }
+
+      &__failed-repo--reason {
+        padding-left: 3rem;
+      }
```
bluein-green: Might be better to nest the classes, e.g.
```
&__failed-repo {
  &--name {
    ...
  }

  &--reason {
    ...
  }
}

File: frontend/src/index.pug [hunk 27]
```diff
@@ -132,6 +132,13 @@ html
               label
                 input(type="checkbox", v-model="filterBreakdown").mui-checkbox.summary-picker__checkbox
                 span breakdown by file format
+        .error-message-box(v-if="Object.entries(errorMessages).length")
+          .error-message-box__close-button(onclick="this.parentElement.style.display='none'") &times;
```
bluein-green: Might be better to create a method in main.js to hide the box rather than reference the element in html

eugenepeh:
@jylee-git can you resolve the conflict?

PR #744, opened 2019-06-12
[#741] Fix view code icon not opening the CodeView

### LINKED ISSUE #741: CodeView tab closes when viewing author's code
As pointed out by @fzdy1914, there is a bug where clicking on the code icon closes the CodeView tab.

To replicate:
1. Set granularity as `week`
2. Reload the page
3. Click to view author's code

week-granularity-codeview-error

jamessspanggg:
Fixes #741 

```
The splitCommitsWeek() method executes a dateRounding() method that 
rounds up the filterSinceDate to the next nearest Monday. When the 
report page is reloaded, the filterSinceDate is set to the value of 
the 'since' hash by the getDates() method. Since within the 
getFiltered() method, splitCommitsWeek() is executed before
getDates(), the filterSinceDate stores an empty string which is 
passed to the dateRounding() method and returns a NaN value. 

Since splitCommitsWeek() method is only called when granularity is 
'week', if users were to set granularity as 'week' and reload the page,
clicking on the view code icon does not open the CodeView panel. 

Let's shift the getDates() method to be executed before 
splitCommitsWeek() to make sure that splitCommitWeek() method
is called with a valid filterSinceDate, which fixes the described bug.
```


fzdy1914:
One potential bug found when playing around with the feature.
a
The tab open state is not stored in the url params correctly. But it should not be related to this. Maybe you can raise a PR about this issue.

jamessspanggg:
> The tab open state is not stored in the url params correctly. But it should not be related to this. Maybe you can raise a PR about this issue.

Noted. How about for the zoom tab? Do we wanna achieve whereby re-loading the page with the zoom panel activated will result into the same page with the exact same zoom panel activated? That might be a little tricky.

Issue #662, opened 2019-05-16
Add a 'merge group' option


damithc:
Current: a group in ramp chart can have multiple rows
![image](https://user-images.githubusercontent.com/1673303/57852395-a63a5d80-7815-11e9-9b5e-2f0156a669a6.png)

Suggestion: allow the user to tick a `[ ] merge group` which should result in all rows of a group (and the contribution bars) being merged into one

Justification: the user can see all contribution of a author as one chart. Similarly, can see all activity in a repo as one chart.

PR #749, opened 2019-06-14
[#748] Restore authorship tab

### LINKED ISSUE #748: Reloading the page does not show authorship tab
As mentioned by @fzdy1914, after reloading the report page with the authorship tab opened, the reloaded page does not show the authorship tab.

reload-tab-authorship

jamessspanggg:
Fixes #748 

```
Reloading the report page with the authorship tab opened will result
to a closed authorship tab.

The info.name property is only created when clicking on the view code
icon and calling openTabAuthorship() method of v_summary. The
initiate() method of v_authorship sets the isTabActive property to be 
false if info.name property does not exist. Reloading the report page 
does not call openTabAuthorship(), which means that the info.name 
property will not exist and the isTabActive is false regardless of 
url value.

Let's modify the initiate() method of v_authorship.js to check for
info.author property instead of info.name as the property exists 
when reloading the page as well as clicking on the view code icon.
This modification properly restores the last viewed authorship tab, 
if any.
```


jamessspanggg:
@fzdy1914 as per our conversation [here](https://github.com/reposense/RepoSense/pull/744#issuecomment-501750208), I was wondering when reloading the page with the zoom panel opened, would it be okay to only show an opened blank tab?

reload-page

fzdy1914:
> @fzdy1914 as per our conversation [here](https://github.com/reposense/RepoSense/pull/744#issuecomment-501750208), I was wondering when reloading the page with the zoom panel opened, would it be okay to only show an opened blank tab?
> 
> reload-page

For now, let's just solve the code view tab bug in this PR. It is acceptable to open the corresponding code view tab when the original link is zoom view. For the remembering of zoom view, it can be a separate feature to implement. You can raise an issue about this and implement it after Prof's permission.

fzdy1914:
> LGTM

Sorry, I misunderstood your meaning. As described in my comment, when opening the zoom view, you should also update the author in the hash to be the new one.

jamessspanggg:
> > LGTM
> 
> Sorry, I misunderstood your meaning. As described in my comment, when opening the zoom view, you should also update the author in the hash to be the new one.

I have modified the code such that when opening the zoom view, the 'tabRepo' and 'tabAuthor' hash will be removed. Will decide on what zoom view hashes to put in the URL in another PR that addresses the remembering of zoom view.

frontend/src/static/js/main.js
```diff
@@ -245,8 +254,8 @@ window.app = new window.Vue({
     if (!this.isTabActive) {
       window.removeHash('tabAuthor');
       window.removeHash('tabRepo');
-      window.addHash('tabOpen', this.isTabActive);
-      window.encodeHash();
     }
+    window.addHash('tabOpen', this.isTabActive);
```
fzdy1914: Why this change?

jamessspanggg: To always have a `tabOpen` hash in the URL, whether be it `true` or `false`. 
Previously there will only be a `tabOpen` hash when it is `false`.

fzdy1914: OK, reasonable.

frontend/src/static/js/main.js
```diff
@@ -181,6 +181,10 @@ window.app = new window.Vue({
       this.activateTab('authorship');
     },
     updateTabZoom(obj) {
+      // remove hashes from authorship tab if available
+      window.removeHash('tabAuthor');
+      window.removeHash('tabRepo');
```
eugenepeh: what is the tab zoom's role in this if its not being restored?

jamessspanggg: This basically removes the `tabAuthor` and `tabRepo` hash that belongs to the authorship tab. Since we are switching to the zoom tab, those hashes are unnecessary.

eugenepeh: Aren't the details in zoom tab still tied to the author and repo?
If we are going to make zoom tab restore-able in the future, don't we need these details?

jamessspanggg: Yes we do require those details, but if we are not removing the `tabAuthor` and `tabRepo` hash, reloading the page with zoom tab open will result to the authorship tab opened with files related to the unremoved `tabAuthor ` and `tabRepo` hash. 

zoomtab

Since this PR is not about restoring the zoom tab, I can make another PR that handles the `tabAuthor` and `tabRepo` correctly and removing these two lines.
https://github.com/reposense/RepoSense/blob/2347cec8c87149eaa6389e6835456c1c9e52e91a/frontend/src/static/js/main.js#L183-L186

eugenepeh: >Yes we do require those details, but if we are not removing the tabAuthor and tabRepo hash, reloading the page with zoom tab open will result to the authorship tab opened with files related to the unremoved tabAuthor and tabRepo hash.

In that case, would adding the previous `tabType` (while retaining the `tabAuthor` and `tabRepo`), helps?

Reason is because it is never good to remove things that are just going to be added back in another PR. In normal circumstances, reverting such PRs wouldn't be possible without manual intervention. Hence, as a practice, we should try to avoid doing such things. This also serves as the justification for your addition of the `tabType` component.

jamessspanggg: > In that case, would adding the previous `tabType` (while retaining the `tabAuthor` and `tabRepo`), helps?

Yes I think it'll help. Reloading the page with zoom tab will probably lead to an empty tab.

> Reason is because it is never good to remove things that are just going to be added back in another PR. 

Noted with thanks. Will try to implement it.

jamessspanggg: @eugenepeh @fzdy1914 

I am changing the restoring of tabs to be within `v_summary` component instead of `app` component. My reasoning behind this is because:

- `v_summary` component is originally responsible of opening tabs, so it makes sense for `v_summary` to handle the restoring of tabs as well, instead of adding code in `app` component to handle restoring of tabs.
- restoring zoom tab will require `filtered` within `v_summary`, so if we handle the restoring of zoom tab within `app` component, we will not be able to access `filtered` as `v_summary` will not be mounted yet.

Additionally, I am changing the url hash from using `tabRepo, tabAuthor` to `repoIndex, userIndex`. This is because to restore zoom tab, we would require to access `filtered` within `v_summary`. The `repoIndex` and `userIndex` is used to keep track of which specific element within `filtered`.

eugenepeh: >Additionally, I am changing the url hash from using tabRepo, tabAuthor to repoIndex, userIndex. This is because to restore zoom tab, we would require to access the filtered array within v_summary as it consists of the repositories that are filtered based on the selected granularity. The repoIndex and userIndex is used to keep track of which specific element within filtered.

Wouldn't that requires the filtered to be loaded first, e.g. search value takes effect beforehand, for this to work?

jamessspanggg: Yes. The filtered will be loaded first, then the tabs will be restored.

eugenepeh: >Yes. The filtered will be loaded first, then the tabs will be restored.

You will need to rethink about this. As your index works on the filtered, modifying the search bar will mess the filtered repos, resulting the restoring not working as expected.

eugenepeh:
>@eugenepeh Your inputs?

I don't mind either as there is common objective between the 2 which is to restore the tab functionality.

frontend/src/static/js/main.js
jamessspanggg: Hmmm that is true. 

But for zoom tab we will definitely require to use `filtered` as `filtered` consists of the commits being grouped in terms of the granularity. If we really need to use `repos`, then we would have to specifically group the commits in `window.REPOS` in terms of the granularity then restore the zoom tab (as opposed to just using `filtered`).

Another reason why I used `filtered` is to take into account of restoring zoom tab from merged groups in PR #662. When the merge group option is checked, `filtered` consist of the merged groups.

eugenepeh: You might need to figure a way out to get this working. The reason I brought it up is because the problem exist in your current work.

jamessspanggg: I have reverted the commits and will deal with restoring of zoom tab in another PR.

jamessspanggg: @eugenepeh I have figured out a new way of restoring zoom tab that uses `window.REPOS` instead of `filtered`. Will proceed with that after this PR.

frontend/src/static/js/main.js
```diff
@@ -207,6 +209,24 @@ window.app = new window.Vue({
       }
     },
 
+    renderTabHash() {
+      window.decodeHash();
+      const hash = window.hashParams;
+      if (!hash.tabOpen) {
+        return;
+      }
+      const tabOpen = hash.tabOpen === 'true';
```
eugenepeh: this variable seems a bit redundant

frontend/src/static/js/v_zoom.js
```diff
@@ -10,8 +10,20 @@ window.vZoom = {
...
+      addHash('tabRepo', user.repoId);
+    },
+
   },
   components: {
     v_ramp: window.vRamp,
   },
+
+  created() {
+    this.setInfoHash();
```
eugenepeh: changes to zoom tab should be together with the other PR

eugenepeh:
>The info.name property is only created when clicking on the view code
icon and calling openTabAuthorship() method of v_summary. The
initiate() method of v_authorship sets the isTabActive property to be 
false if info.name property does not exist. Reloading the report page 
does not call openTabAuthorship(), which means that the info.name 
property will not exist and the isTabActive will always be false after 
reloading the page.

This should be the second paragraph, your first paragraph should set the scene on why are you even talking about all this.

Also minor tweak: 
isTabActive remains false regardless of url value

>Let's modify the initiate() method of v_authorship.js to check for
info.author property instead of info.name as the property exists 
when reloading the page as well as clicking on the view code icon. 

... to properly restore the last viewed authorship tab, if any.

Issue #131, opened 2018-05-28
Redirect functional test to use testrepo


eugenepeh:
Functional is performing the test on repositories that are beyond our control.

This means that if the repositories' owners make any changes to their repository, 
it will affect our test results.

Let's use the state-preserved, forked repositories under our organization instead.

PR #752, opened 2019-06-18
[#662] Add merge group option

### LINKED ISSUE #662: Add a 'merge group' option

jamessspanggg:
Fixes #662 

- [x] Add merge group checkbox and functionality
- [x] Disable merge group option when `groupByNone` selected
- [x] Disable links of ramp slices when merge group checked, `groupByAuthor` is selected and granularity is set to `day` or `week` as an author might have contributed to different repositories within the same ramp slice
- [x] Remove chart title row that includes the title, code icon and octocat icon when merge group option is checked.

```
Within ChartView, a group has multiple rows of ramp charts that belongs
to the group itself.

It may be difficult for users to analyse the contribution of the entire
group or to compare the contribution between groups.

Let's add a merge group checkbox such that when the checkbox is
checked, all ramp charts of the same group will be merged together
to form a single ramp chart. This allows users to easily analyse 
the total contribution on the group level.
```


jamessspanggg:
> The design is OK, but I found that the zoom feature is not available now. Maybe you should disable zoom when `merge group`.

The zoom feature does work when `merge group` when granularity is set to `commit` or `day`. It only does not work when granularity is set to `week` as mentioned in #724

frontend/src/ramp.pug
```diff
@@ -26,9 +26,14 @@
...
-          borderLeftWidth: getWidth(slice) + 'em',\
-          right: (getSlicePos(tframe === 'day' ? slice.date : slice.endDate) * 100) + '%'\
-      }"
+      v-bind:style="[\
+          {\
+            zIndex: user.commits.length - j,\
+            borderLeftWidth: getWidth(slice) + 'em',\
+            right: (getSlicePos(tframe === 'day' ? slice.date : slice.endDate) * 100) + '%' \
+          },\
+          mergegroup && groupby === 'groupByAuthors' ? \
```
fzdy1914: Can combine with this, looks similar? The dynamic style is harder to maintain.
image

fzdy1914:
> A bug found when playing around:

To reproduce: 
1. open your preview, 
2. click `merge group`
3. refresh (buggy: merge group not clicked, ramp not merged)
4. click `merge group`(buggy too)

frontend/src/static/js/v_summary.js
```diff
@@ -348,7 +357,10 @@ window.vSummary = {
       // array of array, sorted by repo
       const full = [];
 
-      this.repos.forEach((repo) => {
+      // create deep clone of this.repos if merging groups
+      const groups = this.isMergeGroup ? JSON.parse(JSON.stringify(this.repos)) : this.repos;
```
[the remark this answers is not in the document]
jamessspanggg: Without deep clone, merging groups will modify the original `repos`, which will trigger The Vue watcher `repos()`.https://github.com/reposense/RepoSense/blob/9b0f4f6600cf49543817d8803cf73dadd600856f/frontend/src/static/js/v_summary.js#L123-L125
Within the Vue watcher `repos()`, it will execute `this.getFiltered()` which will cause an infinite loop.

frontend/src/ramp.pug
jamessspanggg: ok noted.

jamessspanggg:
> Wondering if this is necessary. Attribute `disabled` should already disable the event. And you can add the CSS of `disabled` in scss file. Have a test if it is true.

Yes the `disabled` attribute works. However, setting `disabled` attribute will only blur out the drop down and the checkbox but not their labels. It might not look obvious to the user that the `merge group` checkbox is disabled. Should I blur out the labels as well or this is fine?

Image below shows disabled `sort within groups by` and `merge group`:
Screen Shot 2019-06-26 at 12 31 21 PM

fzdy1914:
> Image below shows disabled `sort within groups by` and `merge group`:
> Screen Shot 2019-06-26 at 12 31 21 PM

The `sort within groups by' looks good for me, maybe only blur the `merge group` label.

frontend/src/ramp.pug
```diff
@@ -26,9 +26,14 @@
...
-          borderLeftWidth: getWidth(slice) + 'em',\
-          right: (getSlicePos(tframe === 'day' ? slice.date : slice.endDate) * 100) + '%'\
-      }"
+      v-bind:style="[\
+          {\
+            zIndex: user.commits.length - j,\
+            borderLeftWidth: getWidth(slice) + 'em',\
+            right: (getSlicePos(tframe === 'day' ? slice.date : slice.endDate) * 100) + '%' \
+          },\
+          mergegroup && groupby === 'groupByAuthors' ?\
```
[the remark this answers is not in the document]
jamessspanggg: Yes only needed for `day` or `week` granularity as there might be commits from different repos in a day or a week.

jamessspanggg:
> Can you have a look of the approach I mentioned?

Noted. Will try to do so.

jamessspanggg:
>  This `isMergeGroup` is not so different from our existing one. I believe the handle of info may be added [here]

The reason that I place the merging of info [here](https://github.com/reposense/RepoSense/blob/369186b9b299ba6a41735b6d76d91391c0c4e92a/frontend/src/static/js/v_summary.js#L396-L398) is because it is after the `this.sortFiltered()` method, as the method sorts the repos in terms of groups based on the `sortGroupSelection`. After sorting in terms of groups, it is easier to merge the groups.

frontend/src/static/js/v_summary.js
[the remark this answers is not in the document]
fzdy1914: > it is easier to merge the groups.

OK, makes sense.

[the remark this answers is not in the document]
fzdy1914: > Within the Vue watcher `repos()`, it will execute `this.getFiltered()` which will cause an infinite loop.

Do we really need the watcher for `repos()` now?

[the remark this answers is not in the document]
jamessspanggg: Actually I'm also not sure. Depends on the author's purpose on placing a watcher for `repos()`.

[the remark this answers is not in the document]
fzdy1914: A watcher of large data like `repos()` will have performance drawbacks. Can you help investigate whether this is truly necessary? From my suspicion, we are already watching every flag which may affect repo(). And we do not have other methods which may directly modify `repos()`?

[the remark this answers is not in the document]
jamessspanggg: Okay will look deeper into it.

But I have tried running the report without the `repos()` watcher. One thing to note is that opening the CodeView without the watcher will lead to `tabAuthor` and `tabRepo` hashes to not be in the url.
Screen Shot 2019-06-28 at 4 54 47 PM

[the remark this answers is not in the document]
jamessspanggg: > Can you help investigate whether this is truly necessary?

**Current:**
The watcher `repos()` will be called when opening/closing tabs (authorship and zoom). This is because the opening and closing of tabs will lead to calling `getUsers()` method and set the `repos` as shown below. 
https://github.com/reposense/RepoSense/blob/369186b9b299ba6a41735b6d76d91391c0c4e92a/frontend/src/index.pug#L29-L35

The calling of `getUsers()` when opening/closing tabs is redundant as they will lead to the same contents of `repos` being set.

**Suggestion:** 
To remove the `repos()` watcher without affecting the current implementation, we can have a `users` data within the `app` component and calls `getUsers()` to set the `users` when the `app` component is initially created. Then we pass `users` instead of `getUsers()` [here](https://github.com/reposense/RepoSense/blob/369186b9b299ba6a41735b6d76d91391c0c4e92a/frontend/src/index.pug#L31) (to prevent `getUsers` to be executed repeatedly when opening/closing tabs).

Furthermore, I realised that there is some redundant code here in line [#117](https://github.com/reposense/RepoSense/blob/369186b9b299ba6a41735b6d76d91391c0c4e92a/frontend/src/static/js/main.js#L117) and [#131](https://github.com/reposense/RepoSense/blob/369186b9b299ba6a41735b6d76d91391c0c4e92a/frontend/src/static/js/main.js#L131) in main.js. It is setting `this.users` to be an empty array. But then `this.users` is not used and it does not exist in the data of the `app` component. Is it because previously there is a `users` data in the `app` component?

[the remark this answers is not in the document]
fzdy1914: I think the suggestion is acceptable.  You may start on #780 first before this PR.

@ongspxm You input?

frontend/src/static/js/v_summary.js
```diff
@@ -358,7 +370,10 @@ window.vSummary = {
...
```
fzdy1914: If we do not watch `repos` for now, do we still need this deep clone?

frontend/src/static/js/v_summary.js
```diff
@@ -385,6 +400,72 @@ window.vSummary = {
...
+
+        group[0].commits = mergedCommits;
+        group[0].fileFormatContribution = mergedFileFormatContribution;
+        group[0].totalCommits = totalMergedCommits;
+        group[0].variance = mergedVariance;
+
+        // clear all users and add merged group in filtered group
+        this.filtered[groupIndex] = [];
+        const mergedGroup = group[0];
+        this.filtered[groupIndex].push(mergedGroup);
```
fzdy1914: Can directly use `this.filtered[groupIndex].push(group[0]);`

frontend/src/static/js/v_summary.js
```diff
@@ -385,6 +400,72 @@ window.vSummary = {
...
+      });
+    },
+    mergeCommits(user, merged, dateToIndexMap) {
+      // merge commits with the same date
+      user.commits.forEach((commit) => {
+        const {
+          commitResults, date, insertions, deletions,
+        } = commit;
```
fzdy1914: Can this be put in one line?

jamessspanggg: Putting this in one line will result to eslint error.

frontend/src/static/js/v_summary.js
jamessspanggg: Yes we do as we do not want to affect the original `repos` when we are merging here:
https://github.com/reposense/RepoSense/blob/65c1766c8c5459984c0437f3b1f754af46bef3d3/frontend/src/static/js/v_summary.js#L432-L434

fzdy1914: OK

frontend/src/static/js/v_summary.js
fzdy1914: Is this because of the extra `,` ?

jamessspanggg: No, I think eslint only allows a max of 3 variables in order for it to be in a single line.

Screen Shot 2019-07-29 at 3 12 48 PM

fzdy1914: OK, that's fine.

frontend/src/ramp.pug
```diff
@@ -32,6 +32,7 @@
             borderLeftWidth: getWidth(slice) + 'em',\
             right: (getSlicePos(tframe === 'day' ? slice.date : slice.endDate) * 100) + '%' \
           },\
+          /* disallow clickable ramp slices when merging groups that are grouped by authors */\
```
eugenepeh: Also. add in the rationale that its due to different repository, hence unable to form a url for clickable link

frontend/src/static/js/v_summary.js
```diff
@@ -373,9 +373,8 @@ window.vSummary = {
       // array of array, sorted by repo
       const full = [];
 
-      // create deep clone of this.repos if merging groups
+      // create deep clone of this.repos to not modify the content of this.repos when merging groups
```
eugenepeh: `the original content`, would be better as it place stronger emphasis

jamessspanggg:
Putting this PR on hold until #830 merges.

damithc:
Hope we can merge this soon as we can use it in CS2103 and CS2113 to detect teams moving too slow.

PR #757, opened 2019-06-20
[#754] Url: include date hashes only when modified

### LINKED ISSUE #754: URL should not capture end date unless it is specifically set
Scenario: I set up a report by specifying a start date but no end date. Furthermore, I set up Travis to auto-generate the report periodically. Then, I share the URL of the report with others.

Problem: The URL captures the end date of that specific report. when others visit that URL in future, they only see ramps up to the original end date although the report has been updated to include more recent commits.

Suggestion: If the end date of the analysis period is unspecified, the URL should not capture a specific end date either.

jamessspanggg:
Fixes #754 

```
The initial URL of the report page captures a start and until 
hash parameter even when the users did not specify the start date
and end date fields in the CLI.

The URL is then shared to others. If there are further updates to the 
report, the URL will still navigate to the report page with ramps up 
until the original start date and until date instead of the most updated 
dates which may include more recent/older commits.

Let’s modify the report’s initial URL to not capture the start and end 
date hashes when users did not specify the respective date field in the 
CLI. Without the start date or end date hash parameter in the URL, the
URL will navigate to the report with the most updated start date and 
end date.
```


jamessspanggg:
I'm not sure whether it is appropriate to add the new test case that tests for no until date provided.

jamessspanggg:
> This PR is greatly conflicting an existing one, maybe it is better to wait for the other one to be merged first?

ok noted.

jamessspanggg:
> We add the hash only when the `until date` is modified.

Sounds like a better idea. Am on it.

jylee-git:
Looks good here. 

But should we take into account where a user changes the `until date`, resets the date range (by clicking the `reset date range` or clears the until date) then shares the link to the others, without knowing that the until date hash is still there? @damithc

damithc:
> But should we take into account where a user changes the `until date`, resets the date range (by clicking the `reset date range` or clears the until date) then shares the link to the others, without knowing that the until date hash is still there? @damithc

As discussed. You may want to post the decision here for the record.

jamessspanggg:
As per our discussion, our decision is to:
- Only capture the since hash or until hash when the respective dates fields are modified
- Clearing since date or until date using the ✖️ button will remove the respective hashes
- Clicking on `reset date range` will clear both since hash and until hash

jamessspanggg:
> * Only capture the since hash or until hash when the respective dates fields are modified
> * Clearing since date or until date using the ✖️ button will remove the respective hashes
> * Clicking on `reset date range` will clear both since hash and until hash

@damithc I have modified the code based on the decision above.

I have generated a report site with `-s 18/05/2019 -u 18/06/2019`, copied the [link](https://cranky-mcclintock-a400a3.netlify.com/#tabOpen=true&search=&sort=groupTitle&sortWithin=title&timeframe=commit&groupSelect=groupByRepos&breakdown=false), and regenerate the site again with `-s 18/04/2019 -u 18/07/2019`.

Using the [link](https://cranky-mcclintock-a400a3.netlify.com/#tabOpen=true&search=&sort=groupTitle&sortWithin=title&timeframe=commit&groupSelect=groupByRepos&breakdown=false) copied, the url navigates to the most updated since date of `18/04/2019` and until date of `18/07/2019` instead of the previous `18/05/2019 and 18/06/2019`.

damithc:
Looks good.
On a separate issue, the date controls shouldn't be allowed to move beyond the original report duration, right? That can be a separate PR though.

jamessspanggg:
> On a separate issue, the date controls shouldn't be allowed to move beyond the original report duration, right? That can be a separate PR though.

Yes, but the date controls that exceed the original duration will not be recorded in the hash.

damithc:
> Yes, but the date controls that exceed the original duration will not be recorded in the hash.

Nevertheless, it's misleading to show a report that doesn't match the values shown in the toolbar, which is the situation when a user pushes the start/end date beyond the applicable range (i.e., the report shows a different duration from the one indicated in the toolbar).

frontend/src/static/js/v_summary.js
```diff
@@ -518,8 +536,10 @@ window.vSummary = {
 
     // updating filters programically //
     resetDateRange() {
-      this.tmpFilterSinceDate = this.minDate;
-      this.tmpFilterUntilDate = this.maxDate;
+      // set both dates to empty string which will call their respective
```
fzdy1914: A suggestion from me:
The main purpose of this should be `let their watcher remove the hash` right? You only mentioned `what to do` but not mention `why do so`. Usually, we are able to read `what to do` from the code directly, but we may not understand `why do so`.

jamessspanggg: Yes the main purpose of this is to `let watcher remove the hash` and `let watcher reset both tmpFilterSinceDate and tmpFilterUntilDate to minDate and maxDate respectively`. OK will amend the comment. Thanks for the feedback.

frontend/src/static/js/v_summary.js
```diff
@@ -141,19 +143,29 @@ window.vSummary = {
     },
     tmpFilterSinceDate() {
       if (this.tmpFilterSinceDate && this.tmpFilterSinceDate >= this.minDate) {
+        if (this.tmpFilterSinceDate !== this.minDate) { // If user modifies the since date field
```
eugenepeh: Should remove this guard clause. 
Reason: what if my intention was to share the link that was meant to use the min or max date?

jamessspanggg: With the current implementation, if the user wants to share the link with min or max date, the user would have to modify date controls respectively and followed by setting the date controls back to the min or max date.

@damithc from our last discussion, we did not take into account when the user wants to share the link that was meant to use the min or max date. Your input?

damithc: If the report was generated by specifying a start date, it should be captured in the URL when the report is opened the first time. Similarly, if there was an end date to begin with, it should be in the initial URL. i.e., the initial URL should match the dates specifically provided by the user via CLI.  Is that possible?

jamessspanggg: @damithc yes it is possible.

jamessspanggg: @damithc by implementing based on your comment, do we still want to remove the since date and until date hashes when the user resets the date range?

jamessspanggg: @eugenepeh I have updated the code such that if the user did specify `sinceDate` and `untilDate` in the CLI, then `sinceDate` and `untilDate` hashes will be captured in the initial URL. On the other hand, if user did not specify `sinceDate` and `untilDate` in the URL, the hashes will not be captured in the initial URL.

So now if the user's intention was to share the link that uses min or max date, the user should specify the `sinceDate` and `untilDate` in the CLI arguments.

damithc: Perhaps we can do this:
* The initial URL will contain dates matching CLI parameters e.g., if end date is not provided via CLI, the URL will not have an end date either.
* If a date is changed subsequently, the URL will take the new date e.g., if the start date is changed, the URL will change the start date too
* If a date is reset, the URL will take the reset date.

That means the user should not change dates if she wants to retain the behavior of the CLI parameters in the URL. If she resets dates, she will be sharing the report for the current period shown (i.e., not dynamic). I think this approach caters for all possibilities. What do you think?

jamessspanggg: > What do you think?

Sounds good to me.

jamessspanggg: Updated according to the requirements.

eugenepeh: Hmm, I still think that any modification to the date (even if going back to min or max date) should be considered modified. Because thinking at it logically, no one will go back to that specific date (without using reset), unless they noticed something is wrong, right?

jamessspanggg: @eugenepeh yes, currently any modification to the date will be considered as modified. (Even if going back to min or max date)

eugenepeh: Hmm I see, because you're using watcher.

I kind of dislike this double "loading" style, because any error in coding will result in infinite loop. 
Can you try removing line 152 and modify 364 to ignore `!this.tmpFilterUntilDate` if it helps?

jamessspanggg: Line 152 `this.tmpFilterSinceDate = this.filterSinceDate;` is dedicated for when the user clears the since date field using the HTML5 date picker `clear` button:

Screen Shot 2019-07-28 at 6 04 28 PM

So if we remove line 152, the since date won't reset to the min date.

eugenepeh: I guess that is due to datepicker is doing `v-model="tmpFilterUntilDate"`.
What if we split the `v-model` 2 way binding to something like this?
```
   v-bind:value="something"
   v-on:input="something = $event.target.value"
```
So that the input variable and display variable are now different. Will we able to get rid of the double load this way?

jamessspanggg: > I guess that is due to datepicker is doing `v-model="tmpFilterUntilDate"`.
> What if we split the `v-model` 2 way binding to something like this?

I have tried this, the dates still will not reset to their values when clicking the `clear` button.

eugenepeh: display wise or value wise? Note that the quote I gave above is just an example.
The actual value for  `v-bind:value` should be `filterUntilDate` itself and 
`v-on:input` should probably be `v-on:input="tmpFilterUntilDate = $event.target.value"`

jamessspanggg: Actually my two-way binding implementation was something like:
`v-bind:value=filterSinceDate`, `v-on:input="updateTmpFilterSinceDate` 
where `updateTmpFilterSinceDate` sets `tmpFilterSinceDate` to `minDate` if `$event.target.value` is empty (which means user has cleared the since date), else sets `tmpFilterSinceDate` as `$event.target.value`.

Actually both `tmpFilterSinceDate` and `filterSinceDate` will reset to the `minDate` value when user has clicked the `clear` button, but I'm not sure why the UI still displays `dd/MM/yyyy` (cleared input) instead of the `minDate` value, .

jamessspanggg: Realised what was the issue. I forgot to set `$event.target.value` back to `filterSinceDate` after clearing the since date field.

Removed the double loading style of tmpFilterSinceDate and tmpFilterUntilDate watcher.

frontend/src/static/js/v_summary.js
```diff
@@ -537,10 +533,8 @@ window.vSummary = {
 
     // updating filters programically //
     resetDateRange() {
-      // setting both dates to empty string will trigger their respective
-      // watchers to reset their values and remove their hashes in the url
-      this.tmpFilterSinceDate = '';
-      this.tmpFilterUntilDate = '';
+      this.tmpFilterSinceDate = this.minDate;
+      this.tmpFilterUntilDate = this.maxDate;
```
eugenepeh: Let this be the original code, that way you should be able to remove the guard clause, which i highlighted before. Since the double loading is gone.

frontend/src/static/js/v_summary.js
eugenepeh: >where updateTmpFilterSinceDate sets tmpFilterSinceDate to minDate if $event.target.value is empty

Why do we need to set the `tmpFilterSinceDate` to minDate when its just a receiving variable though? The `tmpFilterSinceDate` watcher should naturally update the `filterSinceDate` isn't it?

frontend/src/static/js/v_summary.js
jamessspanggg: Sorry, I wasn't really clear with the reasoning behind separating the input variable and display variable.

Now I understand the reason. Have updated the code.

frontend/src/static/js/v_summary.js
```diff
@@ -554,6 +567,14 @@ window.vSummary = {
 
     updateTmpFilterUntilDate(event) {
       const until = event.target.value;
+      this.hasModifiedUntilDate = true;
+
+      if (!this.isSafariBrowser) {
+        this.tmpFilterUntilDate = until;
+        event.target.value = this.filterUntilDate;
```
eugenepeh: why is it that we need to set the `event.target.value`?

jamessspanggg: Because clearing the date fields using the HTML5 date picker button will clear the `event.target.value`. Without setting back `event.target.value`, clearing the date field using the button will result to `dd/MM/yyyy` instead of the filterSinceDate.

PR #758, opened 2019-06-21
[#46] Show total time after batch processing

### LINKED ISSUE #46: Show total time after batch processing
It would be useful to see the total time taken after running the CLI command.

jamessspanggg:
Fixes #46 

```
After running the CLI command, the total time taken to process and 
generate the report is not shown in the console output.

Users might find it useful to keep track of the time taken to generate
the report.

Let's use the logger to output the total time taken from running the 
CLI command until the report is generated.
```


fzdy1914:
@damithc for #46, do you prefer the info to be displayed in the console or the report?

damithc:
> @damithc for #46, do you prefer the info to be displayed in the console or the report?

Would be nice to have both, but can be done as two separate PRs.

src/main/java/reposense/RepoSense.java
```diff
@@ -34,12 +34,15 @@
     private static final int SERVER_PORT_NUMBER = 9000;
     private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("E MMM d HH:mm:ss yyyy z");
     private static final String VERSION_UNSPECIFIED = "unspecified";
+    private static final String MESSAGE_ELAPSED_TIME = "Elapsed processing time: %f seconds";
```
fzdy1914: Instead of seconds, I prefer a `**hour**minute**second` way, which can be more intuitive.

src/main/java/reposense/RepoSense.java
```diff
@@ -34,12 +34,15 @@
...
 
     /**
      * The entry point of the program.
      */
     public static void main(String[] args) {
         try {
+            long startTime = System.nanoTime();
```
fzdy1914: Not sure if it is a good place to leave it here.

jamessspanggg: Maybe before the `generateReposReport` method? Since cloning and analysing occurs in that method.
https://github.com/reposense/RepoSense/blob/0848879f9e89f1fac3dea51285a9c503d2554f33/src/main/java/reposense/RepoSense.java#L63-L66

src/main/java/reposense/RepoSense.java
jamessspanggg: ok noted.

src/main/java/reposense/RepoSense.java
fzdy1914: What I mean is to have a seperate class for the responsibility of time recording.And call it in main method. But not sure if it going to be too complicated.

yong24s: @fzdy1914 
Interesting suggestion.

If we make it an `interface` then we can track time elapsed for each step in the analysis. 🤔

fzdy1914: That's true, we can even know the time taken for every repo. However, the priority of that is low. This is more for our developer to know more info about the analysis.

fzdy1914:
@jamessspanggg How is the process of this PR? If the interface one is too complicated, it is OK to make it functioning by extracting the related code to a single util class first. Like `timeutil`. Then we can extend the behaviour of this `timeutil` in the future.

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -0,0 +1,29 @@
+package reposense.util;
+
+/**
+ * Contains time related functionalities.
+ */
+public class TimeUtil {
+    private static final String MESSAGE_ELAPSED_TIME =
+            "Elapsed processing time: %d hour(s) %d minute(s) %.2f second(s)";
```
fzdy1914: Not sure if I am too mean or not. I suggest not to have `hour` / `minute` displayed if the build does not take an `hour` / `minute` to finish. 

However, this is just left as a suggestion. May go ahead after seniors also agree with this.

jylee-git: I personally agree with this also

jamessspanggg: @fzdy1914 can I continue with making this change?

fzdy1914: Sure, if you also agree with this suggestion.

PR #762, opened 2019-06-21
[#753] Display contribution per chart

### LINKED ISSUE #753: Show line count for each chart
Current: we show the contribution for the group.
Suggestion: we can also show the LoC count for each chart in the group
image

bluein-green:
Resolves #753.

```
Contribution is shown only for each group as a whole.

While it is possible to view the contribution based on each chart by 
hovering over the contribution bar, this requires extra work by the
user. It will be more user-friendly if the contribution can be viewed
on the page itself.

Let's display the contribution for each chart in the chart title.
This information is shown using the same style as the group's total 
contribution.
```


frontend/src/static/css/style.scss
```diff
@@ -244,6 +244,10 @@ header{
             color:mui-color("grey");
             text-decoration:none;
         }
+
+        &--contribution{
+            font-size: 0.65rem;
```
jamessspanggg: No space after colon `:`

frontend/src/static/css/style.scss
```diff
@@ -244,6 +244,10 @@ header{
             color:mui-color("grey");
             text-decoration:none;
         }
+
+        &--contribution{
+            font-size:0.65rem;
```
eugenepeh: since they use the same font-size, would prefer to use a separate class that applies the font size to all relevant elements.

frontend/src/static/css/style.scss
```diff
@@ -152,6 +152,11 @@ header{
     padding:2rem 1rem;
 }
 
+.contribution-text-font {
```
jamessspanggg: `.contribution-text-font` should be nested within `.summary-charts`

eugenepeh: actually, instead of `contribution-text-font`, we can just give it a representative name e.g. `mini-font` and make it a global standard in our style sheet.

frontend/src/static/css/style.scss
```diff
@@ -45,6 +45,11 @@ a{
     }
 }
 
+.mini-font{
+  font-size:.65rem;
```
jamessspanggg: I think it is better for you to indent 4 spaces instead of 2 because the code style of PR #803 is not confirmed yet. If PR #803 merges before this, then you can resolve the conflicts.

eugenepeh:
@bluein-green please resolve the conflict and seek a review from prof after you're done.

PR #765, opened 2019-06-25
[#759] Cloning: prevent double cloning of the same repo from remote

### LINKED ISSUE #759: Increase performance by not cloning a repo twice
(Problem applies to users of Windows OS)
### Current
Currently, RepoSense does a `git clone --bare` of each individual repos and check if any of the files/folders contain illegal characters in the repo. Then when no illegal characters are detected, RepoSense performs a proper clone of the repository (again).

This incurs heavy performance penalty, especially if the remo repo is huge in size (e.g.: TEAMMATES) as repos will be cloned twice (note `clone bare` and `clone` clones the repo with the same total size).

### Suggested
Do a clone bare of the repo, check for illegal file names in windows. Once done, do a clone **from the local directory where the bare repo is located** to a new location. Doing so will just copy the files over, and do a chekout to the master branch. See [this](https://stackoverflow.com/questions/6403982/accessing-source-files-from-git-baremaster).
image

### LINKED ISSUE #761: Analyzing multiple branches (one or more with illegal filename) from same repo causes various issues
The branch `windows-illegal-filename` of [testrepo-alpha](https://github.com/reposense/testrepo-Alpha.git) contains illegal windows filenames.

When analyzing multiple branches of testrepo-alpha in this order: -

1. master
2. windows-illegal-filename

RepoSense only validates the files in the master branch, which contains no files with illegal branch name. It then analyzes the contents in master branch, then followed by the `windows-illegal-filename` branch. The filename `windows:Illegal?Characters!File(Name).java` gets truncated to `windows` in Windows OS when checking out to this branch. `windows` ends up getting analyzed (though it's not shown in the report as `windows` is not in the whitelisted file name).

Then, when analyzing multiple branches of testrepo-alpha in the reverse order: -

1. windows-illegal-filename
2. master

Since the branch `windows-illegal-filename` contains file names with illegal characters, validating file paths will throw exception, **and ignore the whole repo, including the master branch (which is supposed to be clean)**.

### LINKED ISSUE #805: Error message of command line
Running command `gradlew run` on windows gives the following error:
image

jylee-git:
```
In Windows OS, each repo will be bare cloned from the remote, checked 
for any illegal characters in the file path, then cloned from the 
remote again to be analyzed.

The double cloning of the same repo from the remote increases analysis 
time significantly, especially when the user has a sub-optimal internet 
bandwidth.

Also when setting repo-config file, if the user specifies a repo branch 
that contains file path with illegal characters in the repo to be 
analyzed first before the other branches of the same repo, an exception
will be thrown for the first branch with files containing illegal 
characters in the repo, and the other branches of the same repo will be
ignored.

Let's: -
- Locally clone from the previously cloned bare repo instead to reduce 
  the overall time of cloning and analysis.
- Check the other branches of the same repo for files containing 
  illegal characters as well and analyze them as well.

Each repo will be bare cloned. Then for each branch to be analyzed, 
the bare repo will be used to check for any illegal characters in the 
file path. If no illegal characters are found, perform a full clone 
from the bare repo into a new location and directly check out into the 
targeted branch to be analyzed.
```

Resolves #759 
Resolves #761
Resolves #805


src/main/java/reposense/git/GitClone.java
```diff
@@ -22,28 +21,30 @@
...
-     * Clones a bare repo specified in the {@code repoConfig}.
+     * Returns the command to clone a bare repo specified in the {@code repoConfig}
+     * into the folder {@code outputFolderName}.
      */
-    public static void cloneBare(RepoConfiguration repoConfig) throws IOException {
-        FileUtil.deleteDirectory(repoConfig.getRepoRoot());
-        clone(
-            repoConfig.getLocation(), repoConfig.getRepoFolderName(), repoConfig.getRepoName(), GIT_CLONE_BARE_OPTION);
+    public static String getCloneBareCommand(RepoConfiguration repoConfig, String outputFolderName) {
+        return ("git clone --bare " + repoConfig.getLocation() + " " + outputFolderName);
```
yong24s: Extra parenthesis.

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -72,10 +75,9 @@ public static void validateFilePaths(RepoConfiguration config) throws InvalidFil
     /**
      * Returns an Array of {@code String} containing file paths of all tracked files.
      */
-    private static String[] getFilePaths(RepoConfiguration config) {
-        Path rootPath = Paths.get(config.getRepoRoot());
+    private static String[] getFilePaths(Path clonedRepoLocation, RepoConfiguration config) {
```
yong24s: `clonedRepoLocation` may be confusing we have a `RepoLocation` class.

src/main/java/reposense/git/GitClone.java
```diff
@@ -68,15 +69,29 @@ public static void clone(RepoConfiguration repoConfig)
...
-    private static void clone(RepoLocation location, String repoFolderName, String repoName, String additionalCommand)
-            throws IOException {
-        Path rootPath = Paths.get(FileUtil.REPOS_ADDRESS, repoFolderName);
+    public static void cloneBare(RepoConfiguration repoConfig, String outputFolderName) throws IOException {
+        Path rootPath = Paths.get(FileUtil.REPOS_ADDRESS, repoConfig.getRepoFolderName());
+        FileUtil.deleteDirectory(Paths.get(rootPath.toString(), outputFolderName).toString());
         Files.createDirectories(rootPath);
-        String command =
-                String.format("git clone %s %s %s", additionalCommand, addQuote(location.toString()), repoName);
+        String command = String.format("git clone --bare %s %s", addQuote(repoConfig.getLocation().toString()),
```
yong24s: Use getCloneBareCommand?

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -31,28 +30,32 @@
...
     private static final Pattern ILLEGAL_WINDOWS_CHARACTER_PATTERN = Pattern.compile("[:\\\\*?|<>:\"]");
 
+    /**
+     * Clones a bare repo in {@code repoConfig} and verifies that the repo contains only file paths that are
+     * compatible in Windows.
+     * @throws IOException if it fails to create/delete the folders.
+     * @throws InvalidFilePathException if the repository contains invalid file paths that are not compatible with
+     * Windows.
+     */
+    public static void validateFilePaths(RepoConfiguration repoConfig) throws IOException, InvalidFilePathException {
```
yong24s: Standardize `repoConfig` -> `config`.

It shouldn't be different from the overloaded method. Also, it should be the commonly/usually used convention.

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -31,28 +30,32 @@
...
 
     /**
      * Verifies that the repository in {@code config} contains only file paths that are compatible with Windows.
      * Skips check if the operating system is not Windows.
-     *
      * @throws InvalidFilePathException if the repository contains invalid file paths that are not compatible with
      * Windows.
      */
-    public static void validateFilePaths(RepoConfiguration config) throws InvalidFilePathException, GitCloneException {
+    public static void validateFilePaths(RepoConfiguration config, Path clonedBareRepoLocation)
```
yong24s: Same comment for `clonedBareRepoLocation`

jylee-git:
> If I understand correctly, the approach is to clone a bare repo and the rest branch analysis will clone this bare repo?

Yes

> Will this causes the analyze time to be greatly increased? Because it is still cloning for every branch. Can you have a test for analyzing multiple branches of a large repo?

No, because you are essentially cloning from your own local hard disk. 

The current approach is such that a bare repo will be cloned from the remote first, check (only the first branch that pops up in the config list for any invalid filenames), then do a clone from the remote again. This takes up lots of time, especially if someone has a bad internet connection speed and needs to download huge repos like TEAMMATES.

The proposed approach is such that clone from your own local hard disk, which should be fast even for dated HDDs. Cloning from a bare TEAMMATES in my PC using a dated HDD is instantaneous.

fzdy1914:
> I have also tested using the standard repo and author config files to run the current and proposed approach. The current approach takes an estimate 8 minutes to analyze all specified repos whereas the proposed approach takes up to 4 minutes to analyze all specified repos. So I believe this should provide a major performance boost to Windows users.

Please resolve the conflicts and request a review again.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -94,29 +92,25 @@ public void cleanup() {
...
         try {
-            GitLsTree.validateFilePaths(config);
-
-            FileUtil.deleteDirectory(config.getRepoRoot());
-            logger.info(String.format(MESSAGE_START_CLONING, config.getLocation()));
             Path rootPath = Paths.get(FileUtil.REPOS_ADDRESS, config.getRepoFolderName());
             Files.createDirectories(rootPath);
-            crp = runCommandAsync(rootPath, "git clone " + addQuote(config.getLocation().toString()));
+            FileUtil.deleteDirectory(FileUtil.getBareRepoPath(config).toString());
+            Files.createDirectories(Paths.get(FileUtil.REPOS_ADDRESS, config.getRepoFolderName()));
```
fzdy1914: image
Seems to create the same directory for two times?

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -152,16 +157,29 @@ private static void analyzeRepos(String outputPath, List<RepoConfiguration> conf
...
-                GitCheckout.checkout(config.getRepoRoot(), config.getBranch());
+                GitRevParse.assertBranchExists(config, FileUtil.getBareRepoPath(config));
+            } catch (GitBranchException gbe) {
+                logger.log(Level.SEVERE, String.format(MESSAGE_BRANCH_DOES_NOT_EXIST,
+                        config.getBranch(), config.getLocation()), gbe);
+                generateEmptyRepoReport(repoReportDirectory.toString(), Author.NAME_FAILED_TO_CLONE_OR_CHECKOUT);
+                continue;
+            }
+
+            try {
```
fzdy1914: Can combine the try catch block if the exception has no overlapping.

src/main/java/reposense/git/GitRevParse.java
```diff
@@ -0,0 +1,26 @@
+package reposense.git;
```
fzdy1914: Add a test for this class.

src/main/java/reposense/git/GitClone.java
```diff
@@ -22,28 +21,30 @@
...
         try {
-            FileUtil.deleteDirectory(repoConfig.getRepoRoot());
-            logger.info("Cloning from " + repoConfig.getLocation() + "...");
-            clone(repoConfig.getLocation(), repoConfig.getRepoFolderName(), repoConfig.getRepoName(), "");
+            FileUtil.deleteDirectory(config.getRepoRoot());
+            logger.info("Cloning from " + config.getLocation() + "...");
+
+            Path rootPath = Paths.get(FileUtil.REPOS_ADDRESS, config.getRepoFolderName());
+            Files.createDirectories(rootPath);
+            String command = String.format("git clone %s %s", addQuote(config.getLocation().toString()),
```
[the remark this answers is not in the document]
jylee-git: If you mean the `clone` method in here
image

the additional command parameter isn't necessary anymore, so I just directly run the command in this method.

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -31,28 +30,32 @@
...
 
+    /**
+     * Clones a bare repo in {@code config} and verifies that the repo contains only file paths that are
+     * compatible in Windows.
+     * @throws IOException if it fails to create/delete the folders.
+     * @throws InvalidFilePathException if the repository contains invalid file paths that are not compatible with
+     * Windows.
+     */
+    public static void validateFilePaths(RepoConfiguration config) throws IOException, InvalidFilePathException {
+        GitClone.cloneBare(config, FileUtil.getBareRepoFolderName(config));
```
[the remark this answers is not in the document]
jylee-git: This particular method is only used in `GitLsTreeTest` unit test, where the bare repo isn't cloned previously.

src/main/java/reposense/report/RepoCloner.java
jylee-git: My bad, overlooked this

src/main/java/reposense/git/GitClone.java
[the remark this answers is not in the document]
fzdy1914: OK, but having a separate method may provide better reusability.

src/main/java/reposense/git/GitLsTree.java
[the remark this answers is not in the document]
fzdy1914: If so, we may need to move this method to the test? Can ask the senior's opinion.

jylee-git:
@reposense/stage2-reviewers do help to review this. Thanks

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -162,16 +167,26 @@ private static void analyzeRepos(String outputPath, List<RepoConfiguration> conf
...
+                GitLsTree.validateFilePaths(config, FileUtil.getBareRepoPath(config));
+            } catch (GitBranchException gbe) {
+                logger.log(Level.SEVERE, String.format(MESSAGE_BRANCH_DOES_NOT_EXIST,
+                        config.getBranch(), config.getLocation()), gbe);
+                generateEmptyRepoReport(repoReportDirectory.toString(), Author.NAME_FAILED_TO_CLONE_OR_CHECKOUT);
+                continue;
+            } catch (InvalidFilePathException ipe) {
+                generateEmptyRepoReport(repoReportDirectory.toString(), Author.NAME_FAILED_TO_CLONE_OR_CHECKOUT);
+                continue;
+            }
```
yong24s: The exception is getting messy. We may need to improve it in another PR.

eugenepeh: Yes, there shouldn't be 2 try catch within a method. It is an indication of insufficient SLAP here.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -94,29 +92,24 @@ public void cleanup() {
...
-
-            FileUtil.deleteDirectory(config.getRepoRoot());
-            logger.info(String.format(MESSAGE_START_CLONING, config.getLocation()));
+            FileUtil.deleteDirectory(FileUtil.getBareRepoPath(config).toString());
             Path rootPath = Paths.get(FileUtil.REPOS_ADDRESS, config.getRepoFolderName());
             Files.createDirectories(rootPath);
-            crp = runCommandAsync(rootPath, "git clone " + addQuote(config.getLocation().toString()));
+
+            logger.info(String.format(MESSAGE_START_CLONING, config.getLocation()));
+            crp = runCommandAsync(rootPath, GitClone.getCloneBareCommand(config,
```
eugenepeh: can I have this command be executed from GitClone itself?

src/main/java/reposense/report/ReportGenerator.java
jylee-git: I don't think it would be a good idea to just combine all try catch blocks into one as well as it may be harder to debug if things fail.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -104,8 +102,8 @@ private boolean spawnCloneProcess(String outputPath, RepoConfiguration config) t
             Files.createDirectories(rootPath);
 
             logger.info(String.format(MESSAGE_START_CLONING, config.getLocation()));
-            crp = runCommandAsync(rootPath, GitClone.getCloneBareCommand(config,
-                    FileUtil.getBareRepoFolderName(config)));
+            crp = GitClone.cloneBareAsync(config, rootPath, FileUtil.getBareRepoFolderName(config));
+
         } catch (RuntimeException | IOException e) {
```
eugenepeh: The RuntimeException can only be produced by the `runCommandAsync` previously right?
In that case, make `cloneBareAsync` handle that exception and throw it back out as a clone exception.

src/main/java/reposense/report/ReportGenerator.java
eugenepeh: >I don't think it would be a good idea to just combine all try catch blocks into one as well as it may be harder to debug if things fail.

Which is why you will need to increase the level of abstraction, combining some of the codes and handle some of the raw exceptions at lower level.
Ideally if the exceptions are unique to each individual line of code, there wouldn't be problem combining all the try catch blocks into it.

jylee-git: Noted

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -142,48 +144,54 @@ private static void cloneAndAnalyzeRepos(
     /**
      * Analyzes all repos in {@code configs} and generates their report.
      */
-    private static void analyzeRepos(String outputPath, List<RepoConfiguration> configs, String defaultBranch) {
+    private static void analyzeRepos(String outputPath, List<RepoConfiguration> configs, String defaultBranch)
+            throws IOException {
```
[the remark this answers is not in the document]
jylee-git: Err I think you are not on the updated changes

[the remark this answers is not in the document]
eugenepeh: Yeah, I've clicked the submit review in the afternoon but didn't went through. Can ignore this one.

src/main/java/reposense/git/GitClone.java
```diff
@@ -68,15 +79,36 @@ public static void clone(RepoConfiguration repoConfig)
...
+        String command = getCloneBareCommand(config, outputFolderName);
+        runCommand(rootPath, command);
+    }
+
+    /**
+     * Performs a full clone from {@code clonedBareRepoLocation} into the folder {@code outputFolderName} and
+     * directly branches out to {@code targetBranch}.
+     * @throws IOException if it fails to delete a directory.
+     */
+    public static void cloneFromBareAndUpdateBranch(Path rootPath, Path clonedBareRepoLocation,
```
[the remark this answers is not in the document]
jylee-git: This method is only call only
- when repo is cloned from the remote successfully
- the branch(es) exist
- no illegal file names were found (in windows)

Then only this method is called. Also, the cloning is done locally.

If a runtime exception is caught it can only be due to I/O issue. Should we still throw it as I/O exception?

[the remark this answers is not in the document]
eugenepeh: IOException is not a RuntimeException. So any exception that comes from commandrunner is currently not handled. Please catch and handle them.

As for the IOException, do catch it as well, log it and throw it out as GitCloneException as all this contributes to cloning error.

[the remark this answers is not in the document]
jylee-git: What I meant was that if running `runCommand(...)` ends up throwing a RuntimeException, it can only be due to related to file handling issues (like user were to go and delete those files when RepoSense is still analyzing), as all possible scenarios (failure to clone due to invalid URL, invalid branch name, illegal file names) have already been handled beforehand. 

But nevertheless, I'll catch the RTE and throw it as GitCloneException so as to avoid an abrupt program termination when the other repos are being analyzed.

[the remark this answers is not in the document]
eugenepeh: >What I meant was that if running runCommand(...) ends up throwing a RuntimeException, it can only be due to related to file handling issues (like user were to go and delete those files when RepoSense is still analyzing), as all possible scenarios (failure to clone due to invalid URL, invalid branch name, illegal file names) have already been handled beforehand.

Yes, there could also be bunch of other various reasons such as corrupted files etc. In the end it will still be the git clone command throwing the error, hence we should handle it as such.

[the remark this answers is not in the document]
jylee-git: Ok noted

src/main/java/reposense/git/GitClone.java
```diff
@@ -22,28 +23,38 @@
  * Git clone is responsible for cloning a local/remote repository into a new directory.
  */
 public class GitClone {
-
-    public static final String GIT_CLONE_BARE_OPTION = "--bare";
     private static final Logger logger = LogsManager.getLogger(GitClone.class);
 
     /**
-     * Clones a bare repo specified in the {@code repoConfig}.
+     * Runs {@code git clone --bare} command asynchronously to clone a bare repo specified in the {@code config}
```
eugenepeh: `@code` is to highlight code so shouldn't actually be used on string; we normally use it to highlight variables. Use a normal quotation mark or html `<q>` instead.

src/main/java/reposense/git/GitClone.java
```diff
@@ -22,28 +23,38 @@
...
+     * into the folder {@code outputFolderName}.
+     *
+     * @return an instance of {@link CommandRunnerProcess} to allow tracking the status of the cloning process.
```
eugenepeh: `@link` is for url so use `@code` instead

src/main/java/reposense/git/GitClone.java
```diff
@@ -22,28 +23,38 @@
...
+     * @throws GitCloneException when exception has occurred while running the command.
```
eugenepeh: when an error occurs during command execution would be better. The difference between error and exception is like the difference between describing programs and codes.

src/main/java/reposense/git/GitClone.java
```diff
@@ -68,15 +79,46 @@ public static void clone(RepoConfiguration repoConfig)
     }
 
     /**
-     * Clones a repo given the repo location into a directory.
-     * @throws IOException if it fails to create a directory.
+     * Clones a bare repo specified in {@code config} into the folder {@code outputFolderName}.
+     *
```
eugenepeh: no need for extra new line

src/main/java/reposense/git/GitClone.java
```diff
@@ -68,15 +79,46 @@ public static void clone(RepoConfiguration repoConfig)
...
+        String command = getCloneBareCommand(config, outputFolderName);
         runCommand(rootPath, command);
     }
+
+    /**
+     * Performs a full clone from {@code clonedBareRepoLocation} into the folder {@code outputFolderName} and
+     * directly branches out to {@code targetBranch}.
+     *
+     * @throws IOException if it fails to delete a directory.
+     * @throws GitCloneException if an exception has occurred while running the command.
```
eugenepeh: similarly here, the new line and `@throw` message

src/main/java/reposense/git/GitClone.java
```diff
@@ -68,15 +79,46 @@ public static void clone(RepoConfiguration repoConfig)
...
+            runCommand(rootPath, command);
+        } catch (RuntimeException rte) {
+            logger.severe("Exception met while cloning or checking out " + config.getDisplayName() + "."
+                    + "Analysis terminated.");
+            throw new GitCloneException(rte);
+        }
+    }
+
+    /**
+     * Returns the command to clone a bare repo specified in the {@code config}
```
eugenepeh: Constructs the command would probably sound better

src/main/java/reposense/util/FileUtil.java
```diff
@@ -170,6 +173,21 @@ public static void createDirectory(Path dest) throws IOException {
         Files.createDirectories(dest);
     }
 
+    /**
+     * Returns the path to the bare repo version of {@code config} that is relative to the root path.
```
eugenepeh: Returns the relative path... will do

PR #766, opened 2019-06-26
[#721] Report: remove usage of spread operator

### LINKED ISSUE #721: Compatibility issue with Microsoft Edge
There seems to be compatibility issue when viewing the report using Microsoft Edge.

Before PR #573 , summary view works fine, but the code panel can't load any author's code (and just a blank code panel).
image

After PR #573 , summary view doesn't load any of the commits data.
image

jylee-git:
```
Microsoft Edge browser does not support spread operator in JavaScript 
ES6.

Users using Microsoft Edge will not be able to load the report.

Let's remove the usage of spread operator so as to allow the report
to be run from Microsoft Edge browser.
```

Resolves #721


frontend/src/static/js/v_summary.js
```diff
@@ -552,17 +552,10 @@ window.vSummary = {
             (commit) => commit.date >= tsince && commit.date <= tuntil,
         );
 
-        const commits = [];
```
fzdy1914: Why delete this? This should not be a part of this PR.

jylee-git: My bad, we had a misunderstanding about the purpose of these few lines of code.

frontend/src/static/js/v_summary.js
```diff
@@ -562,7 +562,9 @@ window.vSummary = {
         });
 
         const { avgCommitSize } = this;
-        const user = { ...userOrig, commits };
+
```
eugenepeh: no need for the additional new lines

yong24s:
We use `report` to refer to the `frontend`.

PR #768, opened 2019-06-26
[#745] Toolbar: edit flexbox properties

### LINKED ISSUE #745: Make tool bar responsive
The `search` field is not visible under certain screen widths:
image

bluein-green:
Fixes #745.

```
The toolbar does not adjust to changes in window or pane width.

Under certain widths, the `search` field is not visible.

Let's edit the flexbox properties of the toolbar to allow its elements
to shrink and wrap to occupy more rows when the window or pane width 
is decreased.
```


fzdy1914:
@bluein-green We normally do not include `result` part in the commit message. We can check it from the preview. If you indeed want to give a preview, add a separate comment for it.

bluein-green:
> It seems that the first five are in one line and behaves differently with the last four.

@fzdy1914 Yes that is the case. The toolbar is originally split into 3 sections: 

1. the first 5 elements you mentioned
2. dates & `reset date range`
3. `breakdown by file format`

Do I put all of the elements in the same section then?

fzdy1914:
> 1. the first 5 elements you mentioned
> 2. dates & `reset date range`
> 3. `breakdown by file format`
> 
> Do I put all of the elements in the same section then?

I believe it is better to put all the elements in the same section.  The ordering can be:
`top`:
  `first`
  `second`
  ...
  `fifth`
  `dates and reset date`
  `breakdown`

fzdy1914:
@bluein-green 
Hi, for our PR workflow. 

After you have finished one PR, please request a review from `stage1-reviewer`. After you have edited the PR based on one's requirement, re-request a review from the reviewer.

Issue #770, opened 2019-06-27
Two issues when no pre-set authors in the config files and no commits were made during the specified time period


jylee-git:
Two issues occur when a repo with no pre-specified authors in any config files and no commits made during a pre-specified time period. When running with the following arguments `-r https://github.com/reposense/testrepo-Alpha.git https://github.com/reposense/testrepo-Delta.git -s 1/2/2019 -u 25/6/2019 -v` , the following issues occur : -

### Bug 1
Since `testrepo-Alpha` has no pre-specified authors, and no commits were made during that time period, such repos will not be shown in the report page. Also, an exception will be thrown shown below: -
![image](https://user-images.githubusercontent.com/35756209/60243600-28c13d00-98eb-11e9-9e20-cd08f8c871b0.png)

Users will not be able to determine if anyone has made any commits into that repo during that time period. Hence, we should still show the repo (even when no authors are specified) but display a blanket message stating `no commits were made during that time period`.

### Bug 2 (related)
Colors will not be assigned to any of the file formats. This could be due to the exceptions thrown form the bug above.

PR #771, opened 2019-06-27
[#770] Fix bug of file format contribution icons with no colours

jamessspanggg:
Fixes Bug 2 of #770 

```
The 'users' array within 'this.repos' may be empty in situations 
such as analysing repositories with no author commits within
the specified date.

The empty 'users' array will lead to bugs such as not assigning colours
to the file format icons.

Let's use 'this.filtered' instead of 'this.repos' to prevent such bugs
from occurring as 'this.filtered' filters out repositories with empty 
'users' array.
```


eugenepeh:
From the bug description and the commit message

```
When analysing repositories with standalone config, if there is a 
repository with no authors that match the pre-specified authors 
within standalone config file, the 'users' array within 'this.repos' 
will be empty.
```

I don't think this truly solves the problem. 
E.g. reposense_testrepo-Alpha just disappear from the report, is not a good solution. 
It should follow the same style as the other repositories, when encounter an error, a place holder should still be there along with the error message.
image

fzdy1914:
> I don't think this truly solves the problem.

Actually, he has separate the bug into two part. This part is stand alone from the other one. This just fixes an improper use of `user`. But the commit message needed to be edited anyway.

jamessspanggg:
Am closing this PR as a more general solution #777 is found.

fzdy1914:
> Am closing this PR as I have found a more general solution #777 that adds an error place holder and prevents file format icons with no colours.

But I believe the changes made in this PR is still needed. The original design does have drawbacks.

jamessspanggg:
> > Am closing this PR as I have found a more general solution #777 that adds an error place holder and prevents file format icons with no colours.
> 
> But I believe the changes made in this PR is still needed. The original design does have drawbacks.

ok noted.

fzdy1914:
@jamessspanggg 
>  But I believe the changes made in this PR is still needed. The original design does have drawbacks.

You just need to edit the commit message to show why the previous implement may lead to bugs.

jamessspanggg:
> @jamessspanggg
> 
> > But I believe the changes made in this PR is still needed. The original design does have drawbacks.
> 
> You just need to edit the commit message to show why the previous implement may lead to bugs.

noted.

eugenepeh:
Hmm.. with the first bug fix, will this problem still actually appears?

jamessspanggg:
> Hmm.. with the first bug fix, will this problem still actually appears?

No it will no longer appear, but myself and @fzdy1914 thinks that it is more appropriate to use `this.filtered` as opposed to `this.repos` because 
- `this.filtered` will definitely not contain empty `users`, which ensures that this bug will no longer appear
- as we are displaying the ramp chart in terms of `this.filtered` instead of `this.repos`, it makes sense to set the contribution bar colours in terms of `this.filtered`

eugenepeh:
Good points but with certain downside to this

>this.filtered will definitely not contain empty users, which ensures that this bug will no longer appear

This means that we wouldn't realize if there is such a problem in future

>as we are displaying the ramp chart in terms of this.filtered instead of this.repos, it makes sense to set the contribution bar colours in terms of this.filtered

The colour would be different upon changing of search terms and other variables

PR #775, opened 2019-06-28
[#767] Clear zoom region when settings change

### LINKED ISSUE #767: Zoom region is not cleared when settings are changed
The zoom region is sometimes not cleared when a different grouping / order is selected or when search is used. It is also not cleared when the granularity and date range are changed.

This could be misleading for users as the current zoom region's commits do not correspond to those on the code panel.

zoom_region_not_clearing

**Suggestion**
Call `deactivateAllOverlays()` when settings in toolbar are changed.

bluein-green:
Resolves #767.

```
The zoom region on the ramp chart is not cleared when settings in the
toolbar are changed. 

This could be misleading for users as the commits currently in the 
zoom region on the ramp chart do not correspond to those in the
commits panel.

Let's clear the ramp chart's zoom region by calling 
`deactivateAllOverlays()` when the settings are changed.
```


frontend/src/static/js/v_summary.js
```diff
@@ -125,16 +125,20 @@ window.vSummary = {
...
       this.getFiltered();
+      deactivateAllOverlays();
     },
     sortWithinGroupSelection() {
       this.getFiltered();
+      deactivateAllOverlays();
     },
     filterTimeFrame() {
       this.getFiltered();
+      deactivateAllOverlays();
```
jamessspanggg: This might not be necessary as we are planning to make the zoom tab reactive towards `filterTimeFrame`

fzdy1914: Correct, this is not necessary if we are going to make it responsive to zoom.

frontend/src/static/js/v_summary.js
```diff
@@ -123,16 +123,20 @@ window.vSummary = {
   watch: {
     sortGroupSelection() {
       this.getFiltered();
+      deactivateAllOverlays();
```
jamessspanggg: instead of placing `deactivateAllOverlays()` in all watchers, how about place it within `getFiltered()`?

bluein-green: There is a watcher that calls `getFiltered()` but does not require the overlays to be removed, e.g. `filterBreakdown()`. What do you suggest in that case?

jamessspanggg: I guess for simplicity's sake, can just remove the overlays as well? @fzdy1914 your inputs?

fzdy1914: Hmmm, I believe it is better to put it inside `getFiltered()`. 

For control like `filterBreakdown`, after looking at the source code of how this flag works. This watcher seems unnecessary. You may have an investigation about it. @bluein-green

bluein-green: I tried removing the watcher for `filterBreakdown`. In doing so, the breakdown functionality still works. However, as `setSummaryHash()` is only called from `getFiltered()`, the hash would not immediately be set to `breakdown=true` if `filterBreakdown = false` initially. (the hash _would_ contain `breakdown=true` when changing other settings that result in `getFiltered()` being called)

What do you think about putting `deactivateAllOverlays()` in `getFiltered()`, retaining the watcher for `filterBreakdown` and calling `setSummaryHash()` in the watcher? It seems to work when I tried it.

fzdy1914: I suggest, putting `deactivateAllOverlays()` in `getFiltered()`, remove the watcher for `filterBreakdown` and add the hash directly in the `filterBreakdown`.

fzdy1914: I believe we can still have the watcher for `filterBreakdown`, but to change the `getFiltered()` to `setSummaryHash()`.

frontend/src/static/js/v_summary.js
```diff
@@ -137,7 +137,7 @@ window.vSummary = {
       this.getFiltered();
     },
     filterBreakdown() {
-      this.getFiltered();
+      this.setSummaryHash();
```
eugenepeh: Hmm, is it necessary to have a watcher? How about just make it activate on click or something?

https://github.com/reposense/RepoSense/blob/bf2fa03a34da18fc89112efc1c2ef544b716c699/frontend/src/index.pug#L133-L134

fzdy1914: It is not so necessary I believe.

frontend/src/static/js/v_summary.js
```diff
@@ -388,6 +388,7 @@ window.vSummary = {
       this.filtered = full;
 
       this.sortFiltered();
+      deactivateAllOverlays();
```
fzdy1914: Why we put this line here but not in `getFilterd()`?

bluein-green: This is inside `getFiltered()`:
https://github.com/reposense/RepoSense/blob/bf2fa03a34da18fc89112efc1c2ef544b716c699/frontend/src/static/js/v_summary.js#L356-L392

fzdy1914: OK, got it.

PR #782, opened 2019-07-02
[#781] Fix broken program when reserved character is used without escaping

### LINKED ISSUE #781: Unable to use `--since *` when running jar file
I ran `./gradlew clean build` to build the jar file after #690. Trying to run the jar file from command line resulted in an error:

image

Editing the program arguments in intellij configuration to use the same arguments (`-r https://github.com/reposense/reposense.git --since *`) and running from intellij does result in reposense running successfully though.

jylee-git:
```
Following #690, asterisk is used in since date argument as a shorthand
to the date of the first commit.

As asterisk is also a reserved character in shell, using it without
escaping cause the program to be unusable.

Let's change the shorthand to d1 so it can be used directly without the
need to escape.
```

Fixes #781

### DIFF
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -259,7 +259,7 @@ In addition, there are some _optional_ extra parameters you can use to customize
   Example:`--since 21/10/2017` or `-s 21/10/2017` <br>
   > Note: -
   > - If the start date is not specified, only commits made one month before the end date (if specified) or the date of generating the report, will be captured and analyzed.
-  > - If `^` is specified as the start date (`--since d1` or `-s d1`), then the earliest commit date of all repositories will be taken as the since date.
+  > - If `d1` is specified as the start date (`--since d1` or `-s d1`), then the earliest commit date of all repositories will be taken as the since date.
 * **`--until, -u END_DATE`**: The end date of analysis (`-u` as alias). The analysis excludes the end date. Format: `DD/MM/YYYY`<br>
   Example:`--until 21/10/2017` or `-u 21/10/2017` <br>
   > Note: If the end date is not specified, the date of generating the report will be taken as the end date.
--- a/docs/UserGuide.md
+++ b/docs/UserGuide.md
@@ -259,7 +259,7 @@ In addition, there are some _optional_ extra parameters you can use to customize
   Example:`--since 21/10/2017` or `-s 21/10/2017` <br>
   > Note: -
   > - If the start date is not specified, only commits made one month before the end date (if specified) or the date of generating the report, will be captured and analyzed.
-  > - If `*` is specified as the start date (`--since *` or `-s *`), then the earliest commit date of all repositories will be taken as the since date.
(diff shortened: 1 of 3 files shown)


jylee-git:
@jamessspanggg @bluein-green Please help me to verify that `^` works and is not a reserved character in your respective OSes.

bluein-green:
It works on my system

jamessspanggg:
Works on mine too.

yong24s:
@damithc

You okay with `^`?

damithc:
> You okay with `^`?

I guess we can use it.

yong24s:
`--since \*` works right?

So the intention of this PR has be to fix incorrect program behavior when it is used without escaping.

```
Following #690, asterisk is used in since date argument as a shorthand
to the date of the first commit.

As asterisk is also a reserved character in shell, using it without
escaping cause the program to be unusable.

Let's change the character to caret ^ so it can be used directly
without the need to escape.
```

jylee-git:
> `--since \*` works right?

Dosen't work as well

yong24s:
Seems like it doesn't work on Windows shells even after escaping 🤔.

I have updated the commit message accordingly.

yong24s:
@jylee-git 

Is `^` working for you on Windows? It seems to be reserved as well?

```
>^
More?
```

jylee-git:
> @jylee-git
> 
> Is `^` working for you on Windows? It seems to be reserved as well?
> 
> ```
> >^
> More?
> ```

Works in powershell and git bash, but not in CMD

yong24s:
We have to support common shells.

🤔 How about `-`?

damithc:
> We discussed previously that it is better to use a special character, but should we just use an alphanumeric character instead? Alphanumeric character shouldn't be a reserved character in any OSes.
> 

yes, let's use a letter/word. How about `1` or `day1` or  `d1` (to indicate that we mean _from the start of the project_)?

jylee-git:
> 🤔 How about `-`?

`-` works for me on both Powershell and CMD, though I'll need those with Mac OS and other OSes to try out as well.

> yes, let's use a letter/word. How about `1` or `day1` or `d1` (to indicate that we mean _from the start of the project_)?

`day1` or `d1` should be fine.


I would prefer `day1` or `d1` over `-`, since we don't know yet whether `-` is a reserved character in the shells of other OSes.

jylee-git:
`--since d1` confirmed working in all common shells in Windows for my side.

docs/UserGuide.md
```diff
@@ -259,7 +259,7 @@ In addition, there are some _optional_ extra parameters you can use to customize
   Example:`--since 21/10/2017` or `-s 21/10/2017` <br>
   > Note: -
   > - If the start date is not specified, only commits made one month before the end date (if specified) or the date of generating the report, will be captured and analyzed.
-  > - If `*` is specified as the start date (`--since *` or `-s *`), then the earliest commit date of all repositories will be taken as the since date.
+  > - If `^` is specified as the start date (`--since d1` or `-s d1`), then the earliest commit date of all repositories will be taken as the since date.
```
yong24s: > If `d1`

PR #783, opened 2019-07-02
[#780] v_summary.js: remove Vue watcher for repos

### LINKED ISSUE #780: Redundant Vue watcher 'repos()'
As per my conversation with @fzdy1914 [here](https://github.com/reposense/RepoSense/pull/752#discussion_r298509683),

In v_summary.js, there is a watcher method that triggers when `repos` changes (shown below). A watcher for large data such as `repos` will affect the report's performance.
https://github.com/reposense/RepoSense/blob/d2a0cceeff6cce4b584cdc8fdaedb18dd7d7a7c6/frontend/src/static/js/v_summary.js#L122-L125

Currently, only `getUsers()` method directly modify `repos`. The `getUsers()` method is called when:
- report page is initially loaded 
- opening and closing of  `tabAuthorship/tabZoom`

The `getUsers()` method sets the same value to `repos`, so it is only required to be executed once page is firstly loaded.

Let's modify `getUsers()` to only be called once when the page is firstly loaded and remove the `repos()` watcher as there are no longer methods that directly modify `repos`.

jamessspanggg:
Fixes #780 

```
A redundant Vue watcher was found in v_summary.js, specifically the Vue
watcher for repos, as repos should remain constant and not be modified.

This is primarily used to call the getUsers() method, however this 
itself is redundant as getUsers() shouldn't be called more than once 
as it tends to overwrite repos with the same set of values.

Additionally, using Vue watcher for large data, such as repos, affects 
the performance of the report.

Let's remove the Vue watcher for repos and modify getUsers() to be 
called only once when page is firstly loaded.
```


frontend/src/static/js/v_authorship.js
```diff
@@ -92,10 +92,11 @@ window.vAuthorship = {
     },
 
     setInfoHash() {
-      const { addHash, removeHash } = window;
+      const { addHash, removeHash, encodeHash } = window;
       addHash('tabAuthor', this.info.author);
       addHash('tabRepo', this.info.repo);
       removeHash('tabOpen');
+      encodeHash();
```
eugenepeh: what does this has to do with encodeHash though?

jamessspanggg: Previously when you open the authorship tab, the watcher `repos` will be triggered and the `getFiltered()` method will be called. The `getFiltered()` method calls `setSummaryHash` which calls the `encodeHash()` method.

Now since we removed the watcher for `repos`, `getFiltered` won't be called when you open the authorship tab, the `tabAuthor` and `tabRepo` hash will not be reflected in the URL. That is why I added the encodeHash here that adds `tabRepo` and `tabAuthor` when the `tabAuthorship` component is created.

eugenepeh:
@ongspxm @chelseyong can both of you help to take a look at this PR in case we overlooked anything?

eugenepeh:
Nits to propose message:

>A redundant Vue watcher was found in v_summary.js, specifically the Vue
watcher for repos, as repos should remain constant and not be modified.
Currently, only the getUsers() method modifies repos but the 
modification is also redundant as it sets the same value to repos
every time it is called.

Move this part to second paragraph:
... This is primary used to call the getUsers() method, however this itself is redundant as getUsers()  shouldn't be called more than once as it tends to overwrite repos with the same set of values.

**Additionally, using** Vue watcher for large data, such as repos, affects the performance of the report.

jamessspanggg:
> Nits to propose message:

Have modified the commit message.

PR #636, opened 2019-04-06
Monitoring of progress for feature-grouping branch

eugenepeh:
This PR is just to monitor the progress for the grouping feature that is meant to fix #520 
Related PR:
#547 

If anyone wish to continue the feature, do feel free to do so.

PR #787, opened 2019-07-03
[#520] Grouping: allow custom grouping of files

### LINKED ISSUE #520: Custom grouping of files

jylee-git:
- [x] Introduce `FileType` and `FileTypeManager`
- [x] Remove `Format` class
- [x] Introduce custom grouping option

```
"Group" is a new type of classification introduced in this PR to 
classify files/directories that match the given globs into different 
categories. This is similar to how we use the file formats to perform 
breakdown in the ChartView as well as search and filter in CodeView.

This Group functionality provides the ability to customize each 
category, giving users more control over the information reflected in 
the report. To utilize this feature, users simply have to specify the 
categories and the corresponding globs for the classification. 
E.g.: the user can categorize the entire "src/" folder as code, which 
is especially useful when the codebase consists of more than 1 file 
format.

Let's allow users to specify their own custom groups for files in a 
repository, and view a breakdown of contribution by these groups in 
the report.

The user can specify a config file named "groups-config.csv", which 
contains the repository's location, group names and globs to include 
for the group. 

If no group is assigned to a file, it is assigned the default 
group: `other`.

If groups are specified by the user, the Code View and the Contribution
bar will show a breakdown by groups for the checkboxes. Else, breakdown
by file formats are shown instead.

As the code for Format and Grouping are largely similar, let's,
* replace the existing Format class with a new class "FileType", which 
  uses file globs to determine if a file matches the formats or groups
  glob.
* introduce a new class "FileTypeManager" that holds two list of 
  FileType: "Format" and "Group".
```

Fixes #520 
Related PR #547 , #636


jylee-git:
@eugenepeh can I confirm with you the implementation of custom grouping feature.

What we are trying to achieve is:

Whether the user has supplied a custom group or not, it will still be dealt with as `fileType` with the underlying type as `Group`.

- If no custom grouping is given, the formats will be converted to `fileType` where each group will be named with the respective file format and the `filePaths` will be any files that end with the format name. For example: format of java will be converted into `fileType` with the name being `java` and `fileTypes` being `**java`.
- If custom grouping is given, it will attempt to be parsed into `GroupConfigCsvParser` initially, validate the inputs , then transfer over to `fileType`.

That being said, we still need the list of supplied formats in `RepoConfiguration` as we need to validate if the user wants a certain file format to be analyzed, regardless of whether the user is using custom groups or not. 

Your opinions on this?

eugenepeh:
>Whether the user has supplied a custom group or not, it will still be dealt with as fileType with the underlying type as Group.

For now, there isn't a need to have underlying type. Basically, to integrate formats and groups together, we should use a new structure. A 2 fields structure, `Label` and `Pattern`, instead of the original `Format`.

So imagine the old format was simply,
`java`

With the new structure, all the formats e.g. `java` should be converted from:
* Format -> `java`

into something like this:
* Label -> `java`, Pattern -> `**.java` 

which allows cohesive integration with the group.

>If custom grouping is given, it will attempt to be parsed into GroupConfiguration initially, validate the inputs , then transfer over to fileType. Then GroupConfiguration is no longer used.

GroupConfiguration is suppose to be the model for the new configuration file, refer to this conversation
https://github.com/reposense/RepoSense/pull/547#issuecomment-470312239

>That being said, we still need the list of supplied formats in RepoConfiguration as we need to validate if the user wants a certain file format to be analyzed in the later part of the program, regardless of whether the user is using custom groups or not.

Yes that should be the case. Only formats stated will be analyzed, and if groups are supplied, it should absorb and replace the formats. 
When groups are indicated, you can do something like this to perform the integration with supplied formats
`images/**/*.{png,svg}`

jylee-git:
> For now, there isn't a need to have underlying type. Basically, to integrate formats and groups together, we should use a new structure. A 2 fields structure, `Label` and `Pattern`, instead of the original `Format`.
> 
> So imagine the old format was simply,
> `java`
> 
> With the new structure, all the formats e.g. `java` should be converted from:
> 
> * Format -> `java`
> 
> into something like this:
> 
> * Label -> `java`, Pattern -> `**.java`
> 
> which allows cohesive integration with the group.

Ok that's what I have done so far.


> When groups are indicated, you can do something like this to perform the integration with formats supplied
> `images/**/*.{png,svg}`

I don't think this is trivial. What if the user were to set formats in `repo-config.csv` without the intention to override the ones in the standalone config and that there's formats set in the standalone config?

What I can think of is to append these formats to the group globs **only after reading the standalone config file (if any)**.


What do you think?

eugenepeh:
>I'm not really sure how to handle cases where a format is not supposed to be analyzed but the user specifies that format in the group file paths.
E.g.: formats java and adoc are the only ones analyzed (cpp isn't), but the user specifies a group called code and with filepath **.cpp.

Good point, I had not thought of that before.
To avoid duplicate copies, perhaps here's what we can do.

Let's make FileType hold 2 lists
1 -> format
2 -> group

and it will be responsible for providing the white list, and also the list to export.
e.g. If group is not provided, format will be exported. Else group will be exported.

>Also, to determine whether a file format is in the whitelist, should we still retain the current method of looking through the Formats to determine if a file format in inside that list? Or is your preferred approach one that looks through the fileTypes to determine if the glob matches the file (format)?

It will be better if we use glob, to ensure that it works for filtering.

jylee-git:
Just to check, do you want to deprecate `Format` and `Group` class?

eugenepeh:
>Just to check, is your intention to deprecate Format and Group class and integrate everything into FileType?

The new structure that suppose to replace the integrated Format and Group `Label -> java, Pattern -> **.javaclass` can be called FileType.

The one that manages the 2 lists, can prob be named FileTypeManager.

However, if you do have any ideas/suggestions/doubts, do feel free to state them out. Reviewers may not be always right.

jylee-git:
> > Just to check, is your intention to deprecate Format and Group class and integrate everything into FileType?
> 
> The new structure that suppose to replace the integrated Format and Group `Label -> java, Pattern -> **.javaclass` can be called FileType.
> 
> The one that manages the 2 lists, can prob be named FileTypeManager.
> 
> However, if you do have any ideas/suggestions/doubts, do feel free to state them out. Reviewers may not be always right.

Noted. I've replaced and combined `Format` and `Group` into one class called `FileType`, and used `FileTypeManager` to handle both formats and custom groups.

fzdy1914:
@eugenepeh For the user case, I believe it may be better to retain the original formats and add a button to decide whether to group files according to the config given. 

Benefit: 
1. User can choose to see the formats and groupings.
2. Can be done mainly at frontend. The backend only needs to parse the config and send to the frontend.

Drawbacks:
1. **May** reduce frontend performances.

eugenepeh:
>@eugenepeh For the user case, I believe it may be better to retain the original formats and add a button to decide whether to group files according to the config given.

There is a few reason why we're doing so, mainly is due to reduce contradiction as well as performance:
1. If we have both group and formats, there will be double count in the contributed lines in code view
e.g. Java: 1000, Code: 1000 but overlaps each other.
2. The search bar in code view is already using glob, making everything use the glob filter saves code and efficiency. 

Of course there are also other problems such as how to display the groups and file formats in the chart view etc. 

Your suggestions are good which was also part of my plan, but we should always try to divide the work and keep them into manageable pieces.

src/main/java/reposense/model/FileType.java
```diff
@@ -0,0 +1,75 @@
...
+public class FileType {
+    private static final String FILE_TYPE_VALIDATION_REGEX = "[A-Za-z0-9]+";
+    private static final String MESSAGE_ILLEGAL_FILE_TYPE = "The provided file type, %s, contains illegal characters.";
+
+    private String label;
+    private PathMatcher pathsGlob;
+
+    public FileType(String label, List<String> paths) {
+        validateFileType(label);
+        this.label = label;
```
fzdy1914: The name of `label` can be decided, maybe `name` or `type`? Can ask seniors' opinion.

src/main/java/reposense/model/FileType.java
```diff
@@ -0,0 +1,75 @@
...
+        setPathsGlobMatcher(paths);
+    }
+
+    /**
+     * Ensures that the string {@code label} is a valid file type.
+     * @throws IllegalArgumentException if {@code label} does not meet the criteria.
+     */
+    private void validateFileType(String label) {
```
fzdy1914: Make it static?

src/main/java/reposense/model/FileType.java
```diff
@@ -0,0 +1,75 @@
...
+    public static List<FileType> convertStringFormatsToFileTypes(List<String> formats) {
+        return formats.stream().map(FileType::convertStringFormatToFileType).collect(Collectors.toList());
+    }
+
+    public static FileType convertStringFormatToFileType(String format) {
+        return new FileType(format, Collections.singletonList("**" + format));
+    }
+
+    public boolean isFileGlobMatching(String fileName) {
+        return getPathsGlobMatcher().matches(Paths.get(fileName));
```
fzdy1914: Usually，we use `this.pathsGlob.matches()` directly.

src/main/java/reposense/model/LocationsCliArguments.java
```diff
@@ -13,7 +13,7 @@
     private boolean isStandaloneConfigIgnored;
 
     public LocationsCliArguments(List<String> locations, Path outputFilePath, Date sinceDate,
-            Date untilDate, List<Format> formats, boolean isAutomaticallyLaunching,
+            Date untilDate, List<FileType> formats, boolean isAutomaticallyLaunching,
```
fzdy1914: Arg name here.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -48,20 +49,22 @@ public RepoConfiguration(RepoLocation location, String branch) {
                 false, false, false);
     }
 
-    public RepoConfiguration(RepoLocation location, String branch, List<Format> formats, List<String> ignoreGlobList,
+    public RepoConfiguration(RepoLocation location, String branch, List<FileType> formats, List<String> ignoreGlobList,
```
fzdy1914: Arg name here

src/test/java/reposense/model/FileTypeTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.junit.Test;
+
+public class FileTypeTest {
+    public static final List<String> DEFAULT_TEST_FORMAT_STRINGS = Arrays.asList(
+            "adoc", "cs", "css", "fxml", "gradle", "html", "java", "js",
+            "json", "jsp", "md", "py", "tag", "txt", "xml");
+    public static final List<FileType> DEFAULT_TEST_FORMATS = FileType.convertStringFormatsToFileTypes(
+            DEFAULT_TEST_FORMAT_STRINGS);
+    public static final List<FileType> NO_SPECIFIED_FORMATS = Collections.emptyList();
+
```
fzdy1914: Can consider adding a test with multiple formats?

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -200,26 +200,22 @@ public static void updateRepoConfig(RepoConfiguration config) {
 
         if (!Files.exists(configJsonPath)) {
             logger.info(String.format(MESSAGE_NO_STANDALONE_CONFIG, config.getLocation(), config.getBranch()));
-            return;
-        }
-
-        if (config.isStandaloneConfigIgnored()) {
+        } else if (config.isStandaloneConfigIgnored()) {
```
fzdy1914: No need to change the block format to `if ... elseif ... else ...`. Instead, we use original style to check if there are exceptional cases to deal with using `if` one by one.

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -105,8 +107,11 @@ private void generateReport(String inputDates)
                 new RepoConfigCsvParser(((ConfigCliArguments) cliArguments).getRepoConfigFilePath()).parse();
         List<AuthorConfiguration> authorConfigs =
                 new AuthorConfigCsvParser(((ConfigCliArguments) cliArguments).getAuthorConfigFilePath()).parse();
+        List<GroupConfiguration> groupConfigs =
+                new GroupConfigCsvParser(((ConfigCliArguments) cliArguments).getGroupConfigFilePath()).parse();
```
fzdy1914: The way you use the parser does not seems to be optional? (Not sure)

src/main/java/reposense/authorship/AuthorshipReporter.java
```diff
@@ -25,6 +25,7 @@ public static AuthorshipSummary generateAuthorshipSummary(RepoConfiguration conf
                 .filter(Objects::nonNull)
                 .collect(Collectors.toList());
 
-        return FileResultAggregator.aggregateFileResult(fileResults, config.getAuthorList(), config.getFormats());
+        return FileResultAggregator.aggregateFileResult(fileResults, config.getAuthorList(),
+                config.getFileTypeLabels());
```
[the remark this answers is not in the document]
jylee-git: `fileTypeLabels` refer to the the "name" of the group/format, (e.g.: code, test,)

src/main/java/reposense/model/FileType.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Represents a file type for use in {@link FileTypeManager}.
+ */
+public class FileType {
+    private static final String FILE_TYPE_VALIDATION_REGEX = "[A-Za-z0-9]+";
+    private static final String MESSAGE_ILLEGAL_FILE_TYPE = "The provided file type, %s, contains illegal characters.";
```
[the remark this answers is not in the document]
jylee-git: Do you mean the regex? If so, the regex is only used to validate the `label`

src/main/java/reposense/RepoSense.java
```diff
@@ -94,6 +97,14 @@ public static void main(String[] args) {
             // Ignore exception as the file is optional.
         }
 
+        try {
```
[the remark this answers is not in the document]
jylee-git: Can't do that. A user may supply only `author-configs.csv` and not `group-config.csv`; or vice-versa.

Doing so may cause a supplied optional csv file to be ignored instead.

frontend/src/static/js/v_summary.js
```diff
@@ -678,8 +678,8 @@ window.vSummary = {
   created() {
     this.renderFilterHash();
     this.getFiltered();
+    this.processFileTypes();
```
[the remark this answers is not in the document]
jylee-git: No need to. This swap occurred after merging those conflicts.

src/main/java/reposense/authorship/FileResultAggregator.java
```diff
@@ -17,15 +16,15 @@
      * Returns the {@code AuthorshipSummary} generated from aggregating the {@code fileResults}.
      */
     public static AuthorshipSummary aggregateFileResult(List<FileResult> fileResults, List<Author> authors,
-        List<Format> formats) {
-        AuthorshipSummary authorContributionSummary = new AuthorshipSummary(fileResults, authors, formats);
+            List<String> fileTypes) {
```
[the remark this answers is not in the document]
jylee-git: Isn't it supposed to be 8 spacing when writing on a new line inside a parenthesis for method signature?

src/systemtest/java/reposense/ConfigSystemTest.java
jylee-git: Erm.. We just parse in the `groupConfigs` here in the systems test. The groupConfigs in `RepoSense#getRepoConfigurations` is optional if the user does not want to supply custom groupings.

src/main/java/reposense/git/GitUtil.java
```diff
@@ -77,11 +77,11 @@ static String convertToFilterAuthorArgs(Author author) {
     /**
      * Returns the {@code String} command to specify the file formats to analyze for `git` commands.
      */
-    static String convertToGitFormatsArgs(List<Format> formats) {
+    static String convertToGitFormatsArgs(List<FileType> formats) {
```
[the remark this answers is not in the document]
jylee-git: I personally advise against renaming the arg name to `fileTypes` as there's `groups` in `FileTypeManager` to prevent any confusion that this method may include groups as well.

jylee-git:
@fzdy1914 I don't recommend renaming `formats` to `fileTypes` for those that you have commented, as groups uses the class `FileType` as well. This is to prevent any confusion whereby (future) developers may use `groups` in `FileType` in replacement for `formats` when the method requires formats.

src/main/java/reposense/RepoSense.java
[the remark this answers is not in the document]
fzdy1914: OK, understand.

src/main/java/reposense/authorship/FileResultAggregator.java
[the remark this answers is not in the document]
fzdy1914: Oops, didn't look carefully. You are correct.

src/systemtest/java/reposense/ConfigSystemTest.java
fzdy1914: OK, I saw your previous comment of why can not combine the try together.

src/main/java/reposense/model/FileType.java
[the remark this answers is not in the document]
fzdy1914: You can ignore this comment.

src/main/java/reposense/authorship/AuthorshipReporter.java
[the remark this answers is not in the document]
fzdy1914: OK.

fzdy1914:
> @fzdy1914 I don't recommend renaming `formats` to `fileTypes` for those that you have commented, as groups uses the class `FileType` as well. This is to prevent any confusion whereby (future) developers may use `groups` in `FileType` in replacement for `formats` when the method requires formats.

If it is the case, shouldn't we have two subclasses which inherit from `Filetype` to ensure there is not misusage of `FileType`?

jylee-git:
> > @fzdy1914 I don't recommend renaming `formats` to `fileTypes` for those that you have commented, as groups uses the class `FileType` as well. This is to prevent any confusion whereby (future) developers may use `groups` in `FileType` in replacement for `formats` when the method requires formats.
> 
> If it is the case, shouldn't we have two subclasses which inherit from `Filetype` to ensure there is not misusage of `FileType`?

In a way, there is `FileTypeManager` that handles `Format` and `Group` though the way of handling format and group is different from your suggestion.

https://github.com/reposense/RepoSense/pull/787#issuecomment-511124580
https://github.com/reposense/RepoSense/pull/787#issuecomment-511418553

fzdy1914:
> In a way, there is `FileTypeManager` that handles `Format` and `Group` though the way of handling format and group is different from your suggestion.
> 
> [#787 (comment)](https://github.com/reposense/RepoSense/pull/787#issuecomment-511124580)
> [#787 (comment)](https://github.com/reposense/RepoSense/pull/787#issuecomment-511418553)

This is a reasonable approach. Then, we should always consider passing a `FileTypeManager` object instead of `Group`s or `Format`s directly. We may call `getGroup` or `getFormat` within the method to ensure there is no misuse of File type.

@eugenepeh Your opinion about this issue?

eugenepeh:
>This is a reasonable approach. Then, we should always consider passing a FileTypeManager object instead of Groups or Formats directly. We may call getGroup or getFormat within the method to ensure there is no misuse of File type.
>
>@eugenepeh Your opinion about this issue?

To me, I find that there is no value in creating / keeping the `Format` or `Group` class because they can be treated the same, hence you can simply name them as such within the `FileTypeManager` itself. 

Yes, we should use `getGroup`/`getFormat` or rather name it as `getWhitelist` if you're going to use it to filter the files according to filetype. Preferably as an immutable list to avoid misuse of filetype as pointed out by @fzdy1914

jylee-git:
> > This is a reasonable approach. Then, we should always consider passing a FileTypeManager object instead of Groups or Formats directly. We may call getGroup or getFormat within the method to ensure there is no misuse of File type.
> > @eugenepeh Your opinion about this issue?
> 
> To me, I find that there is no value in creating / keeping the `Format` or `Group` class because they can be treated the same, hence you can simply name them as such within the `FileTypeManager` itself.
> 
> Yes, we should use `getGroup`/`getFormat` or rather name it as `getWhitelist` if you're going to use it to filter the files according to filetype. Preferably as an immutable list to avoid misuse of filetype as pointed out by @fzdy1914

For the approach to parse in `FileTypeManager` for cases like this (to prevent misuse):

https://github.com/reposense/RepoSense/blob/a5bc45a7942ae98ce826ca38ce18fee25e8de8fa/src/main/java/reposense/git/GitUtil.java#L77-L88

I understand and have made the changes.

But for the comments that are targeting the CLI arguments classes, I think that the name of the variable should still be `formats` and not involving `FileTypeManager` because only file formats are involved when parsing CLI arguments.

eugenepeh:
>But for the comments that are targeting the CLI arguments classes, I think that the name of the variable should still be formats and not involving FileTypeManager because only file formats are involved when parsing CLI arguments.

You can implement the function in `FileTypeManager` to return the formatted string so that there won't be any misuse.

jylee-git:
> You can implement the function in `FileTypeManager` to return the formatted string so that there won't be any misuse.

Sorry I don't quite understand..

Take this as example:

https://github.com/reposense/RepoSense/blob/3717fde7f3319018d4bd4bbac6aa5de565ab6660/src/main/java/reposense/parser/ArgsParser.java#L156-L176

https://github.com/reposense/RepoSense/blob/a5bc45a7942ae98ce826ca38ce18fee25e8de8fa/src/main/java/reposense/model/LocationsCliArguments.java#L11-L26

When parsing the CLI arguments input by the user, are you trying to mean that instead of using `FileType` in line 176 of `ArgsParser`:
```java
List<FileType> formats = FileType.convertStringFormatsToFileTypes(results.get(FORMAT_FLAGS[0]));
```

Inititlize a `FileTypeManager` right here instead and set the formats in directly?

eugenepeh:
What has 

`List<FileType> formats = FileType.convertStringFormatsToFileTypes(results.get(FORMAT_FLAGS[0]));`

got to do with

`static String convertToGitFormatsArgs(List<FileType> formats)`?

jylee-git:
> What has
> 
> `List<FileType> formats = FileType.convertStringFormatsToFileTypes(results.get(FORMAT_FLAGS[0]));`
> 
> got to do with
> 
> `static String convertToGitFormatsArgs(List<FileType> formats)`?

Nothing to do with it and totally unrelated. I'm talking about a different issue here and gave an example here hopefully to give a clearer idea of what I'm trying to ask here.

`static String convertToGitFormatsArgs(List<FileType> formats)` is used in the git log command. I have already changed the method to take in `FileTypeManager` so as to prevent misuse by other developers who might be confused.

But now I'm referring to the comments made by @fzdy1914 above that are not tagged as resolved yet . All those deal with the parsing in CLI arguments by the user (classes involved: `CliArguments`, `ConfigCliArguments` and `LocationsCliArguments`). The issue here is that should we name the variables as `formats`? `fileTypes`? or change the type from `FileType` to `FileTypeManager` and set the file formats provided by the user into `FileTypeManager` directly? Nothing to do with `static String convertToGitFormatsArgs(List<FileType> formats)`.

eugenepeh:
>But now I'm referring to the comments made by @fzdy1914 above that are not tagged as resolved yet . All those deal with the parsing in CLI arguments by the user (classes involved: CliArguments, ConfigCliArguments and LocationsCliArguments). The issue here is that should we name the variables as formats? fileTypes? or change the type from FileType to FileTypeManager and set the file formats provided by the user into FileTypeManager directly? Nothing to do with static String convertToGitFormatsArgs(List<FileType> formats).

I see, in that case, let's create FileTypeManager directly.

eugenepeh:
Sorry @jylee-git , on second thought, it is actually better to retain the `List<FileType> formats = FileType.convertStringFormatsToFileTypes(results.get(FORMAT_FLAGS[0]));` and return `List<FileType>` directly.

Reason: it is not good for CLIArguments to return "non-raw" type.
The creation of FileTypeManager should only be available after that.

To minimise possible misuse, can make FileTypeManager only constructable with `List<FileType> formats`.

jylee-git:
> Sorry @jylee-git , on second thought, it is actually better to retain the `List<FileType> formats = FileType.convertStringFormatsToFileTypes(results.get(FORMAT_FLAGS[0]));` and return `List<FileType>` directly.
> 
> Reason: it is not good for CLIArguments to return "non-raw" type.
> The creation of FileTypeManager should only be available after that.
> 
> To minimise possible misuse, can make FileTypeManager only constructable with `List<FileType> formats`.

Ok noted

damithc:
Guys, let's push this forward as it can be quite useful for CS2103 and CS2113.

fzdy1914:
@damithc OK, I will finish the review asap.

frontend/src/static/js/v_summary.js
```diff
@@ -389,24 +389,24 @@ window.vSummary = {
...
+          Object.keys(user.fileTypeContribution).forEach((fileType) => {
+            if (!Object.prototype.hasOwnProperty.call(fileTypeColors, fileType)) {
+              fileTypeColors[fileType] = selectedColors[i];
               i = (i + 1) % selectedColors.length;
             }
           });
         });
-        this.contributionBarColors = colors;
       });
+      this.contributionBarFileTypeColors = fileTypeColors;
```
fzdy1914: The reason moving this out the loop?

src/test/java/reposense/git/GitLogTest.java
```diff
@@ -22,14 +22,14 @@ public void gitLog_existingFormats_hasContent() {
...
-        config.setFormats(Collections.singletonList(new Format("py")));
+        config.setFormats(FileType.convertStringFormatsToFileTypes(Collections.singletonList("py")));
         String content = GitLog.get(config, getAlphaAllAliasAuthor());
         Assert.assertTrue(content.isEmpty());
     }
 
     @Test
     public void gitLog_includeAllJavaFiles_success() {
-        config.setFormats(Collections.singletonList(new Format("java")));
+        config.setFormats(FileType.convertStringFormatsToFileTypes(Collections.singletonList("java")));
```
fzdy1914: See [here](https://github.com/reposense/RepoSense/pull/787/files#diff-37415d1fc78bc1091efc3c5eec01d822R11), I believe better to rename the method to `convertFormatStringsToFileTypes`

src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -62,8 +65,12 @@
     private static final List<String> THIRD_AUTHOR_GLOB_LIST = Arrays.asList("**[!(.md)]", "collated**");
     private static final List<String> FOURTH_AUTHOR_GLOB_LIST = Collections.singletonList("collated**");
 
-    private static final List<Format> CONFIG_FORMATS = Format.convertStringsToFormats(Arrays.asList(
+    private static final List<FileType> CONFIG_FORMATS = FileType.convertStringFormatsToFileTypes(Arrays.asList(
             "java", "adoc", "md"));
```
fzdy1914: Consider making it singletonList?

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -57,11 +57,11 @@
 
     private static final String TEST_REPO_DELTA_LOCATION = "https://github.com/reposense/testrepo-Delta.git";
     private static final String TEST_REPO_DELTA_BRANCH = "HEAD";
-    private static final List<Format> TEST_REPO_DELTA_FORMATS =
-            Format.convertStringsToFormats(Arrays.asList("java", "fxml"));
+    private static final List<FileType> TEST_REPO_DELTA_FORMATS =
+            FileType.convertStringFormatsToFileTypes(Arrays.asList("java", "fxml"));
```
fzdy1914: Singleton list as well?

src/test/java/reposense/util/TestUtil.java
```diff
@@ -133,6 +135,27 @@ public static void compareAuthor(Author expectedAuthor, Author actualAuthor) {
...
 
+    /**
+     * Compares attributes of {@code expectedFormats} and the white-listed formats in {@code actualFileTypeManager}.
+     *
+     * @throws AssertionError if the formats in {@code actualFileTypeManager} is different from
+     * {@code expectedFormats}.
+     */
+    public static void compareWhitelistedFormats(
+            List<FileType> expectedFormats, FileTypeManager actualFileTypeManager) {
+        Assert.assertEquals(expectedFormats, actualFileTypeManager.getFormats());
```
fzdy1914: A bit unnecessary from my point for these two methods. 
Can directly use `Assert.assertEquals(TEST_REPO_BETA_CONFIG_FORMATS, config.getFileTypeManager().getFormats)` 

Can ask seniors opinion.

src/test/java/reposense/model/RepoConfigurationTest.java
jylee-git: Not possible? As the resulting CONFIG_FORMATS will have more than one elements inside

src/test/java/reposense/parser/CsvParserTest.java
jylee-git: Same as the above

frontend/src/static/js/v_summary.js
jylee-git: This one wasn't intended; probably a result of the heavy merge conflicts

src/main/java/reposense/git/GitLog.java
```diff
@@ -45,7 +45,7 @@ public static String getWithFiles(RepoConfiguration config, Author author) {
         command += GitUtil.convertToGitDateRangeArgs(config.getSinceDate(), config.getUntilDate());
         command += " --pretty=format:" + PRETTY_FORMAT_STRING + " --stat";
         command += GitUtil.convertToFilterAuthorArgs(author);
-        command += GitUtil.convertToGitFormatsArgs(config.getFormats());
+        command += GitUtil.convertToGitFormatsArgs(config.getFileTypeManager());
```
eugenepeh: should use `config.getFileTypeManager().getFormats()` instead, principle of least knowledge

src/main/java/reposense/model/FileType.java
eugenepeh: label itself or name are fine with me, can consult others' opinion too.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -247,8 +247,8 @@ private static void generateIndividualRepoReport(
     }
 
     /**
-    * Generates a report at the {@code repoReportDirectory}.
-    */
+     * Generates a report at the {@code repoReportDirectory}.
```
eugenepeh: this change doesn't seems to be the concern of this PR

src/test/java/reposense/util/TestUtil.java
eugenepeh: yes, indeed quite redundant

jylee-git: Have removed those

src/main/java/reposense/parser/GroupConfigCsvParser.java
```diff
@@ -0,0 +1,86 @@
...
+        String location = get(record, LOCATION_POSITION);
+        String groupName = get(record, GROUP_NAME_POSITION);
+        List<String> globList = getAsList(record, FILES_GLOB_POSITION);
+
+        GroupConfiguration groupConfig = null;
+        try {
+            groupConfig = findMatchingGroupConfiguration(results, location);
+        } catch (InvalidLocationException e) {
+            e.printStackTrace();
+        }
```
yong24s: What happens when `InvalidLocationException` is thrown?

`groupConfig` is `null` and the following lines will fail?

src/main/java/reposense/parser/GroupConfigCsvParser.java
```diff
@@ -0,0 +1,86 @@
...
+
+        for (GroupConfiguration groupConfig : results) {
+            if (groupConfig.getLocation().equals(config.getLocation())) {
+                return groupConfig;
+            }
+        }
+
+        results.add(config);
+        return config;
+    }
```
yong24s: Not necessary in this PR.

Can make this a generic method but require creation of a new parent class.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -113,7 +112,8 @@
                 continue;
             }
 
-            if (config.getFormats().isEmpty() || Format.isInsideWhiteList(filePath, config.getFormats())) {
+            if (!config.getFileTypeManager().hasSpecifiedFormats()
+                    || config.getFileTypeManager().isInsideWhitelistedFormats(filePath)) {
```
eugenepeh: can simply make `isInsideWhitelistedFormats` return true if there is no specified formats.

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -0,0 +1,103 @@
...
+
+    private List<FileType> formats;
+    private List<FileType> groups;
+
+    public FileTypeManager(List<FileType> formats) {
+        this.formats = formats;
+        groups = new ArrayList<>();
+    }
+
+    public String getFileType(String fileName) {
```
eugenepeh: I think name this `getFileTypeLabel` will be less confusing because you have a model name `FileType` plus, you have another function named `getFileTypeLabels`

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -326,6 +362,29 @@ public void setAuthorEmailsAndAliasesMap(Map<String, Author> authorEmailsAndAlia
...
+    }
+
+    /**
+     * Returns the labels used for the file types depending on whether the user has specified a custom grouping.
+     */
+    public List<String> getFileTypeLabels() {
+        return fileTypeManager.getFileTypeLabels();
+    }
+
+    public String getFileType(String fileName) {
```
eugenepeh: same here

src/main/java/reposense/parser/GroupConfigCsvParser.java
jylee-git: That line should fail and not be parsed in.

Have corrected this part of the code.

src/main/java/reposense/parser/GroupConfigCsvParser.java
jylee-git: I don't quite get it why is it not necessary..

We need to find out if an existing `GroupConfiguration` (tagged to `RepoLocation`) exists in order to append the new groups in. If no existing `GroupConfiguration` exists we need to add a new `GroupConfiguration` into the existing list of `results` and return it.

And also a generic method to replace `findMatchingGroupConfiguration` may not be a good idea as the one in `AuthorConfigCsvParser#findMatchingAuthorConfiguration` has additional checks if the branch is equal (matching group doesn't need to check if the branch is equal or not).

src/test/java/reposense/model/RepoConfigurationTest.java
fzdy1914: I mean instead of `Arrays.asList `, use `Collections.singletonList`

jylee-git: image

Only possible to use singletonList if you are parsing in one item inside to become a list. But I'm parsing in 3 items

fzdy1914: OK, sorry for the inconvenience caused.

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -19,7 +19,10 @@ public FileTypeManager(List<FileType> formats) {
         groups = new ArrayList<>();
     }
 
-    public String getFileType(String fileName) {
+    /**
+     * Returns the assigned file type {@code label} of the file {@code fileName}.
```
eugenepeh: avoid using `@code` on variable that doesn't exist

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -19,7 +19,10 @@ public FileTypeManager(List<FileType> formats) {
...
+     */
+    public String getFileTypeLabel(String fileName) {
```
eugenepeh: Actually on a second thought, why don't you just return `FileType` directly? When used on a `String` parameter input, it will automatically calls `.toString()` to convert it to a `String` type variable.

src/main/java/reposense/parser/GroupConfigCsvParser.java
yong24s: > And also a generic method to replace findMatchingGroupConfiguration may not be a good idea as the one in AuthorConfigCsvParser#findMatchingAuthorConfiguration has additional checks if the branch is equal (matching group doesn't need to check if the branch is equal or not).

Sorry I meant, not necessary to make a generic method now. Its doable, can always get your base classes to implement a boolean method for evaluation.

jylee-git: Ohh I think I get what you mean now..

Maybe we can deal with this over another PR specifically to clean up codes?

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -175,7 +175,7 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
             Date sinceDate = cliSinceDate.orElse(getDateMinusAMonth(cliUntilDate));
             Date untilDate = cliUntilDate.orElse(getCurrentDate());
             List<String> locations = results.get(REPO_FLAGS[0]);
-            List<Format> formats = Format.convertStringsToFormats(results.get(FORMAT_FLAGS[0]));
+            List<FileType> formats = FileType.convertFormatStringsToFileTypes(results.get(FORMAT_FLAGS[0]));
```
yong24s: Should rename `formats` too?

src/main/java/reposense/authorship/model/FileResult.java
```diff
@@ -28,6 +31,10 @@ public String getPath() {
         return path;
     }
 
+    public String getFileType() {
+        return fileType;
+    }
```
yong24s: Same as Eugene's [comment](https://github.com/reposense/RepoSense/pull/787/files#r309656953).

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.stream.Collectors;
+
+/**
+ * {@code FileTypeManager} is responsible for holding a list of whitelisted formats and user-specified custom
+ * groupings.
+ */
+public class FileTypeManager {
+    private static final String DEFAULT_GROUP = "other";
+
+    private List<FileType> formats;
```
yong24s: Internally we are referring it as `fileTypes`?

src/main/java/reposense/parser/ArgsParser.java
jylee-git: Would prefer to retain `formats`. `fileTypes` may confuse others

src/main/java/reposense/model/FileTypeManager.java
jylee-git: Data type wise, yes

src/main/java/reposense/model/FileTypeManager.java
jylee-git: This can be done also. Will get the method to return FileType directly

src/main/java/reposense/authorship/model/FileResult.java
jylee-git: The `fileType` in `FileResult` has to be String in this case (line 15). Because `FileResult` will be directly serialized into JSON based on the stricture of `FileResult`.

image

This will be the resulting json (on the left side) if we were to change the data type to `FileType`, which will contain information that is unnecessary for the front end.

eugenepeh: @jylee-git can you try this?

https://stackoverflow.com/questions/53318844/is-it-possible-to-serialize-object-to-json-using-tostring-method-without-custom

jylee-git: That link provides solution for jackson serializer, not gson (the one we are using now).

But I'll check if there's any of sorts for gson

jylee-git: There's a similar solution for Gson. Will use customized serializer instead

eugenepeh:
oops sorry @jylee-git can you also add a sample group-config.csv (like the other configs) in the userguide? add a few examples so that the user knows how to configure for multiple repo.

damithc:
Guys, how's this coming along?

docs/UserGuide.md
```diff
@@ -354,6 +355,22 @@ If `author-config.csv` is not given and the repo has not provide author details
...
+#### `groups-config.csv`
+
+Optionally, you can provide a `groups-config.csv`(which should be in the same directory as `repo-config.csv` file) to provide details on any custom groupings for files in specified repositories ([example](groups-config.csv)). It should contain the following columns:
+
+Column Name | Explanation
+----------- | -----------
+Repository's Location | Same as `repo-config.csv`.
+Group Name | Name of the group e.g.,`test`.
+Globs * | The list of file path globs to include for specified group. e.g.,`**/test/*;**.java`.
+<sup>* **Multi-value column**: multiple values can be entered in this column using a semicolon `;` as the separator.</sup>
```
damithc: Is it supposed to be like that?
image

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -0,0 +1,105 @@
...
+            for (FileType group : groups) {
+                if (group.isFileGlobMatching(fileName)) {
+                    result = group;
+                }
+            }
+            return result;
+        } else {
+            return getFileFormat(fileName);
+        }
+    }
```
damithc: Use a guard clause to reduce nesting?
```java
if (!hasCustomGroups()) {
    return getFileFormat(fileName);
}
```

docs/UserGuide.md
jylee-git: Hmm didn't notice this as IntelliJ formatted this nicely and not over a new row.. Made the changes

docs/UserGuide.md
```diff
@@ -14,6 +14,7 @@
   * [Customize Using csv Config Files](#customize-using-csv-config-files)
     * [`repo-config.csv`](#repo-configcsv)
     * [`author-config.csv`](#author-configcsv)
+    * [`groups-config.csv`](#groups-configcsv)
```
yong24s: Sorry to bring this up now, not sure if plurality is preferred.

@damithc 
Is group~s~-config.csv better?

damithc: Good catch. In hindsight, all three should have used the plural, to be more accurate of what they contain. 
For the sake of consistency, we can make this `group-config.csv`

eugenepeh:
>Having a breakdown by groups allows users to customize the 
breakdown they want to view in the report, providing users more 
control over the information reflected in the report.

Can you provide a short explanation on what group is and use formats in your explanation?
Need to provide some context to the reader.

jylee-git:
> > Having a breakdown by groups allows users to customize the
> > breakdown they want to view in the report, providing users more
> > control over the information reflected in the report.
> 
> Can you provide a short explanation on what group is and use formats in your explanation?
> Need to provide some context to the reader.

Have updated the commit message

eugenepeh:
Further nits on the commit message:

>Grouping is a new feature that classifies files/directories that 
match the given globs into a file type.

`Group` is a new type of classification introduced in this PR to classify files/directories that 
match the given globs into different categories. This is similar to how we use the file formats to perform breakdown in the ChartView as well as search and filter in CodeView.

>Having a breakdown by groups allows users to customize the 
breakdown they want to view in the report, providing users more 
control over the information reflected in the report.

This `Group` functionality provides the ability to customize each category, giving users more control over the information reflected in the report. To utilize this feature, users simply have to specific the categories and the corresponding globs for the classification. E.g. user can categories the entire `src/` folder as `code`, which is especially useful when the codebase consist of more than 1 file format.

jylee-git:
> To utilize this feature, users simply have to specific the categories and the corresponding globs for the classification. E.g. user can categories the entire `src/` folder as `code`, which is especially useful when the codebase consist of more than 1 file format.

I provided an example in the 5th paragraph of the message body

PR #789, opened 2019-07-04
[#785] ChartView: fix bug of author with contribution bar but without ramps

### LINKED ISSUE #785: Weird analysis result for recent one month
In our [netlify preview](http://reposense.netlify.com/), there are two authors in reposense who have contribution bar but no ramp. We need to investigate why it happens. (May related to co-authorship)
image

jamessspanggg:
Fixes #785 

```
The current git blame command used in RepoSense contains the option -w, 
which ignores whitespace when comparing the parent’s version and the 
child’s to find where the lines came from.

When an author did not change the content but only the position of 
specific lines of code, the previous author will be recognised as the 
author that changed these lines. If the previous author did not 
contribute during the analysis period, his/her ramp chart will be empty 
while the contribution bar will be present.

Let's add an extra check in aggregateBlameInfo for lines of code 
with the author committed date that is not within the duration of 
specified since date and until date. Those lines of code's author will
be set to unknown, which will resolve the issue above.
```


src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -75,20 +80,32 @@ private static void aggregateBlameAuthorInfo(RepoConfiguration config, FileInfo
         String blameResults = getGitBlameResult(config, fileInfo.getPath());
         String[] blameResultLines = blameResults.split("\n");
         Path filePath = Paths.get(fileInfo.getPath());
+        Long sinceDateInMs = config.getSinceDate().getTime();
+        Long untilDateInMs = config.getUntilDate().getTime();
+        int systemRawOffset = TimeZone.getTimeZone(ZoneId.systemDefault()).getRawOffset();
```
fzdy1914: The user may pass in a `timezone` via `-t`, if it is the case, should we still use `systemDefault`?

jamessspanggg: The `timezone` via `-t` is only used for the generation time of the report, but not affecting to the `sinceDate` and `untilDate` of the report. The `sinceDate` and `untilDate` is generated based on the system's default.

e.g. the purpose of `timezone` is only used here
Screen Shot 2019-07-17 at 4 43 51 PM

damithc: > The `timezone` via `-t` is only used for the generation time of the report, but not affecting to the `sinceDate` and `untilDate` of the report. The `sinceDate` and `untilDate` is generated based on the system's default.

It would be nice to have the timezone considered for all time-related things such as the start time. That can be a separate PR though.

src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -25,11 +27,18 @@
...
 
 /**
  * Contains templates for git testing.
  */
 public class GitTestTemplate {
+    protected static final Date BLAME_TEST_SINCE_DATE = TestUtil.getDate(2018, Calendar.FEBRUARY, 6);
+    protected static final Date BLAME_TEST_UNTIL_DATE = TestUtil.getEndOfDate(2018, Calendar.FEBRUARY, 8);
+    protected static final Date MOVED_FILE_SINCE_DATE = TestUtil.getDate(2018, Calendar.FEBRUARY, 7);
+    protected static final Date MOVED_FILE_UNTIL_DATE = TestUtil.getEndOfDate(2018, Calendar.FEBRUARY, 9);
+    protected static final Date PR_617_SINCE_DATE = TestUtil.getDate(2019, Calendar.MARCH, 28);
```
fzdy1914: Should rename it to its relevant name. Otherwise, we need to go to PR 617 to see what it is about.

src/test/java/reposense/util/TestUtil.java
```diff
@@ -104,6 +104,19 @@ public static Date getDate(int year, int month, int date) {
                 .getTime();
     }
 
+    /**
+     * Creates and returns a {@code Date} object with the specified {@code year}, {@code month}, {@code day} that is
+     * at the end of the day.
+     */
+    public static Date getEndOfDate(int year, int month, int date) {
```
fzdy1914: I somehow remember we already got a smililar method in our [code base](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/parser/ArgsParser.java#L232)? Can consider extracting all date related method to a `DateUtil` or `TimeUtil`. But can done in a separate PR though.

src/test/java/reposense/template/GitTestTemplate.java
jamessspanggg: PR 617 is a one-line file used to test the author with emails that have the `+` operator within it, for e.g. `35282962+githubid@users.noreply.github.com`

I am unsure how to rename it, would EMAIL_WITH_ADDITION_TEST be alright?

fzdy1914: Acceptable.

src/systemtest/resources/30daysFromUntilDate/expected/reposense_testrepo-Charlie_master/authorship.json
```diff
@@ -9926,7 +9926,7 @@
       {
         "lineNumber": 39,
         "author": {
-          "gitId": "charlesgoh"
+          "gitId": "-"
```
eugenepeh: this one seems to be wrong, please check that your solution is working; no discrepancy between results and fact before requesting for review.

jamessspanggg: @eugenepeh 
I don't think this is wrong as this line was committed during `1st Nov 2017 10.35am GMT+8`.
Screen Shot 2019-07-24 at 11 47 02 AM

Whereas the test analysis period is from `1st Oct 2017` to `1st Nov 2017`. The current implementation of setting sinceDate and untilDate is getting the date at `12.00am`. Since `1st Nov 2017 10.35am` is after `1st Nov 2017 12.00 am`, the line should be set as an unknown author `-`.
Screen Shot 2019-07-24 at 11 48 48 AM

src/test/java/reposense/authorship/FileAnalyzerTest.java
```diff
@@ -83,6 +79,8 @@ public void analyzeFile_blameTestFileIgnoreFakeAuthorCommitFullHash_success() {
 
     @Test
     public void analyzeFile_blameTestFileIgnoreAllCommit_success() {
+        config.setSinceDate(BLAME_TEST_SINCE_DATE);
+        config.setUntilDate(BLAME_TEST_UNTIL_DATE);
```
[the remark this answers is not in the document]
jamessspanggg: As the current implementation checks if each line's commit date is within `sinceDate` and `untilDate`, config must be set with a `sinceDate` and `untilDate` before running `FileInfoAnalyzer.analyzeFile()`

src/systemtest/resources/30daysFromUntilDate/expected/reposense_testrepo-Charlie_master/authorship.json
eugenepeh: yes but the `"content": "image::Ui.png[width\u003d\"800\"]"` below is the one updated by charlesgoh, isn't it? So why is his gitId not being tagged to the content?

src/test/java/reposense/authorship/FileAnalyzerTest.java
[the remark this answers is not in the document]
eugenepeh: I see, but if it affects so many unit tests, should it be applied to all the unit test, under the GitTestTemplate.java using before or beforeClass?

src/systemtest/resources/30daysFromUntilDate/expected/reposense_testrepo-Charlie_master/authorship.json
jamessspanggg: Yes it is updated by charlesgoh. 

This PR basically sets any line's author to an unknown `-` when the commit date of that line is not within the since date and until date. 

The reason being is to prevent the situation as follow:
Since we are using the option `-w` in our git blame, users that moved a specific line without modifying the content will not be recognised as the author of that line, the previous author will be recognised as the author instead. But if the previous author's commit date for that particular line is before the since date, the code view will still show this line. As said by Prof Damith,

> The code should not be in the chart because Jun An didn't contribute the code during the analysis period. 

_Originally posted by @damithc in https://github.com/reposense/RepoSense/issues/785#issuecomment-507974771_

Hence setting these lines to an unknown `-` will prevent the line to be shown in the code view.

eugenepeh: Yes, but the content should match the date and the author as well. Because when you define an `untilDate`, RepoSense is supposed to check out to the latest commit before the date as well. So if charlesgoh is not suppose to be the author, the `"content": "image::Ui.png[width\u003d\"800\"]"` should not be the content as well.

jylee-git: I'm still getting a grasp of this issue, but meanwhile in `GitRevList#getCommitHashUntilDate`:

image

I want the resulting commit hash made **before 1 Nov 2017, 00:00hrs** but the resulting `revListCommand` to be used reflects a date-time to be **before 1 Nov 2017, 25:59:59**.

This results in the commit has `13762e5ed965557cd8bd90ad789ba280d095a91e` which falls on **1 Nov 2019, 21:59hrs SGT**.

In the UG:
image

It states that the analysis **excludes the end date** (i.e.: The until date should be **before** 1 Nov 2017, 00:00 hrs).


Now, tracing the source code: 
image
image


The docs states `Checks out to the latest commit before {@code untilDate} in {@code branchName} branch if {@code untilDate} is not null.` But going into the method `getCommitHashUntilDate`, the docs states that `Returns the latest commit hash inclusive and until the end of the day of {@code date}.`.

Why is there a contradiction here?

eugenepeh: This was due to the result of several different opinions about the until date in the past. 
We ended with a conclusion that the until date should be included in the analysis, however I guess the docs were not updated to match that conclusion. Do open an issue to fix up on that. 

Anyway, regarding the result, it is still wrong in the sense where the content, author and date doesn't seems to fit together. Hence, @jamessspanggg you will still need to research on that.

jamessspanggg: @eugenepeh for the example of charlesgoh that you gave above, the commit date was actually `11/7/2018 10.35a.m.` whereas our until date is `11/7/2018 12.00a.m.` If we set the until date to `11/7/2019 11.59p.m.`, the content will match the date and the author.

jylee-git: Hmm ok. Will open up an issue on that.

But in any case, I think this until date issue also plays a part in whether the authorship should be credited to `charlesgoh` because a different hash was checked out as a result of the confusion as to whether untilDate is to be included or not.

As we thought that the until date should have been excluded in the analysis, let's say that I modify the `GIT_LOG_UNTIL_DATE_FORMAT` to take the date at 00:00hrs instead of 23:59hrs, and get the latest commit hash before 1 Nov 2017 00:00hrs. The commit hash returned should be `bde21cfaf0bfe306931b2670c56dadecf223512a`, which falls on 31/10/2017, 22:32hrs.

Here at [line 39 of their UG](https://github.com/reposense/testrepo-Charlie/blame/bde21cfaf0bfe306931b2670c56dadecf223512a/docs/UserGuide.adoc#L39), line 39 should be credited to `limmlingg` or `-` (depending on the author and date configs).


image
Here is the `30daysfromutildate` system test, where the left side shows the result when I were to check out to the hash that is before the until date **(exclusive)**, where the right side shows the current implementation result where RepoSense checks out to the hash that is the untilDate **inclusive**.

eugenepeh: @jamessspanggg 

Hmm, revlist and checkout should be using the same until date format, hence I still suspecting that somewhere is wrong.

Currently, going by what you mentioned:
Yes, the date matches the author, however the content still doesn't match. If it is supposed to be the content that is not authored by `charlesgoh` the width should be 790 and not 800.
Hence, there is still a mismatch between content and author even though, assuming that that's an issue with the dates.

jamessspanggg: @eugenepeh revlist uses until date that is `11.59p.m.` even though the until date is `12.00a.m.`.


The mismatch of content and author is due to the until date in the config not matching with the until used in realist.

eugenepeh: >The until date in the config was used to compare whether the commit date is within range of since date and until date.

I see, in that case, do put up a fix to make both synchronize to 23:59 of the until date.

jamessspanggg: Ok. Will do so in a separate pr.

src/test/java/reposense/authorship/FileAnalyzerTest.java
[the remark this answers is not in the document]
jamessspanggg: The `sinceDate` and `untilDate` of different tests are different tho. So I'm not sure if its suitable to place under `before` or `beforeClass`.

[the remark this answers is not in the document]
eugenepeh: >The sinceDate and untilDate of tests are different tho. So I'm not sure if its suitable to place under before or beforeClass.

Actually on second thought, why aren't we using full range?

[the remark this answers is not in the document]
jamessspanggg: I’m not sure though I just followed the dates set from last time. But some test cases are to test when only since date is provided, some for when only until date is provided etc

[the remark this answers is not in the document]
eugenepeh: Only those that requires to be set, should be set though.
This change seems added that requirement to many other unit tests.

src/systemtest/resources/30daysFromUntilDate/expected/reposense_testrepo-Charlie_master/authorship.json
jamessspanggg: @eugenepeh the current issue no longer exists and this PR is ready for review.

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -10,13 +10,17 @@
     @Test
     public void noAnnotationTest() {
         config.setAnnotationOverwrite(false);
+        config.setSinceDate(BLAME_TEST_SINCE_DATE);
+        config.setUntilDate(BLAME_TEST_UNTIL_DATE);
```
eugenepeh: can re-evaluate if theres still any values in this 2 tests?

src/test/java/reposense/authorship/FileAnalyzerTest.java
```diff
@@ -15,52 +13,51 @@
...
 import reposense.template.GitTestTemplate;
-import reposense.util.TestUtil;
 
 
 public class FileAnalyzerTest extends GitTestTemplate {
 
     @Test
     public void blameTest() {
+        config.setSinceDate(BLAME_TEST_SINCE_DATE);
+        config.setUntilDate(BLAME_TEST_UNTIL_DATE);
```
eugenepeh: if this applies to all blame test, prob should add it to before

src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -25,11 +27,20 @@
...
 
 /**
  * Contains templates for git testing.
  */
 public class GitTestTemplate {
+    protected static final Date BLAME_TEST_SINCE_DATE = TestUtil.getSinceDate(2018, Calendar.FEBRUARY, 6);
+    protected static final Date BLAME_TEST_UNTIL_DATE = TestUtil.getUntilDate(2018, Calendar.FEBRUARY, 8);
+    protected static final Date MOVED_FILE_SINCE_DATE = TestUtil.getSinceDate(2018, Calendar.FEBRUARY, 7);
+    protected static final Date MOVED_FILE_UNTIL_DATE = TestUtil.getUntilDate(2018, Calendar.FEBRUARY, 9);
+    protected static final Date EMAIL_WITH_ADDITION_TEST_SINCE_DATE =
```
eugenepeh: think this is specifically meant for only that unit test, this constant can be moved to the test class itself.

jamessspanggg: noted.

src/test/java/reposense/authorship/FileAnalyzerTest.java
jamessspanggg: This does not apply to all blame tests, but majority of them. Should I still add it to `@before` and overwrite accordingly for those tests that do not use the above since and until dates?

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
jamessspanggg: The `blameTest.java` has no author tag `@@author`. Hence these two tests yields the same output and they are not directly testing on whether the `annotationOverwrite` functionality works correctly.

eugenepeh: If thats the case, can
* put up an issue
* remove this entire test file
* move `BLAME_TEST_SINCE_DATE` and `BLAME_TEST_UNTIL_DATE` to `FileAnalyzerTest` class

src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -25,11 +27,16 @@
...
+import reposense.util.TestUtil;
 
 /**
  * Contains templates for git testing.
  */
 public class GitTestTemplate {
+    protected static final Date BLAME_TEST_SINCE_DATE = TestUtil.getSinceDate(2018, Calendar.FEBRUARY, 6);
+    protected static final Date BLAME_TEST_UNTIL_DATE = TestUtil.getUntilDate(2018, Calendar.FEBRUARY, 8);
+    protected static final Date MOVED_FILE_SINCE_DATE = TestUtil.getSinceDate(2018, Calendar.FEBRUARY, 7);
+    protected static final Date MOVED_FILE_UNTIL_DATE = TestUtil.getUntilDate(2018, Calendar.FEBRUARY, 9);
```
eugenepeh: same for all these, if they are only used in the test class itself, move all to the test class to avoid bloating the template

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
jamessspanggg: I was wondering if the `annotationTest.java` file is initially meant to be used for the `AnnotatorAnalyzerTest` instead, as it contains the author tag and the file name matches up with the test name.

Screen Shot 2019-12-24 at 10 29 27 PM

src/test/java/reposense/template/GitTestTemplate.java
jamessspanggg: Noted. Have done so.

PR #790, opened 2019-07-04
ChartView: modify UI of placeholder for repositories with no author commits

jamessspanggg:
```
Within ChartView, ramp charts with error messages are blended together 
with the other ramp charts without error messages as they both have 
similar UI.

Let's modify the UI of ramp charts with error messages to use red color 
for their info message font to distinguish itself from other ramp 
charts without error messages.
```

**Updated UI**

group by repo:
Screen Shot 2019-07-05 at 11 34 54 AM

group by author:
Screen Shot 2019-07-05 at 11 35 05 AM

group by none:
Screen Shot 2019-07-05 at 11 35 15 AM

### DIFF
--- a/frontend/src/index.pug
+++ b/frontend/src/index.pug
@@ -150,10 +150,11 @@ html
                   span &nbsp; {{ fileFormat }} &nbsp;
             .summary-chart(v-for="(user, i) in repo")
               .summary-chart__title
-                .summary-chart__title--index {{ i+1 }}
-                .summary-chart__title--repo(v-if="filterGroupSelection === 'groupByNone'") {{ user.repoName }}
-                .summary-chart__title--author-repo(v-if="filterGroupSelection === 'groupByAuthors'") {{ user.repoName }}
-                .summary-chart__title--name(v-if="filterGroupSelection !== 'groupByAuthors'") {{ user.displayName }} ({{ user.name }})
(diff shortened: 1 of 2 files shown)


fzdy1914:
@damithc Your opinion of the UI change?

damithc:
> @damithc Your opinion of the UI change?

I'm OK with it 👍

frontend/src/index.pug
```diff
@@ -131,7 +131,10 @@ html
           .summary-charts(v-for="repo of filteredRepos")
             .summary-charts__title(v-if="filterGroupSelection === 'groupByRepos'") {{ repo[0].repoName }}&nbsp;&nbsp;
               .summary-charts__title--contribution(title="Total contribution of group") [{{ getGroupTotalContribution(repo) }} lines]
-            .summary-charts__title(v-if="filterGroupSelection === 'groupByAuthors'") {{ repo[0].displayName }} ({{ repo[0].name }})&nbsp;&nbsp;
+            .summary-charts__title(
+                v-if="filterGroupSelection === 'groupByAuthors'",
+                v-bind:class="{ red: repo[0].name === '-' }"
```
fzdy1914: The class name `red` is a bit weird. But I cannot find a better one myself.

jamessspanggg: how about `warning`?

fzdy1914: Can you check on the `.warn` class of CSS? `.warning` is also better than `red`.

PR #792, opened 2019-07-05
[#791] CodeView: bugfix for unexpected colors in line numbers

### LINKED ISSUE #791: Unexpected colors assigned to line numbers in code view
image

image

So far this only happens to `.pug` files for my case.

jamessspanggg:
Fixes #791 

```
The css property of line number only overwrites the color property of 
classes 'hljs-number' and 'hljs-comment'. Others classes such as
'hljs-quote', 'hljs-name' and more are not overwritten.

This will cause the color property of certain line numbers to inherit 
from the hljs classes that are not overwritten.

Let's overwrite all hljs classes to ensure the color property of the 
line number to always be grey.
```

### DIFF
--- a/frontend/src/static/css/style.scss
+++ b/frontend/src/static/css/style.scss
@@ -483,12 +483,9 @@ header{
                 -ms-user-select:none; /* Internet Explorer/Edge */
                 user-select:none; /* Non-prefixed version, currently supported by Chrome and Opera */
 
-                .hljs-number{
-                    color:mui-color('grey');
-                }
-
-                .hljs-comment{
-                    color:mui-color('grey');
+                // overwrite all hljs colours
+                [class^='hljs']{
+                    color: mui-color('grey');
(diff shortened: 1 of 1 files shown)


frontend/src/static/css/style.scss
```diff
@@ -483,12 +483,9 @@ header{
...
 
-                .hljs-number{
-                    color:mui-color('grey');
-                }
-
-                .hljs-comment{
-                    color:mui-color('grey');
+                // overwrite all hljs colours
+                [class^='hljs']{
+                    color: mui-color('grey');
```
jylee-git: From existing formatting convention: -

`color: mui` --> `color:mui`

jamessspanggg:
@fzdy1914 but it is nested within ‘line-number’ though so it won’t clash with other divs. There are around 20+ hljs-xxx classes.

Issue #705, opened 2019-05-24
Restricting command line output to one line


eugenepeh:
The console output too much messages; making it difficult to differentiate important messages. 

Hence, we should restrict the status message from occupying too many lines on the terminal screen, making it self contained in a single line by constantly overwriting.

jylee-git:
I don't know if there's a way to overwrite the same line using logging.. AFAIK, there's no way to overwrite to same line using logging

eugenepeh:
The logging can be reimplemented if such feature does not exist; can even write our own API too if we have to as it also provides us with greater control.

damithc:
Alternatively, we can try to format the output such that important messages are hard to miss.
Overwriting has the downside that we lose the ability to see the sequence of events that happened before.

eugenepeh:
>Alternatively, we can try to format the output such that important messages are hard to miss.

The intention is to make it less important message to take up less space though.
E.g. if we use RepoSense to analyze 50+ repos, its going to flood the CLI with bunch of messages, which can be tedious to scroll through all of them for the error messages.

So my suggestion is to have the info messages to always overwrite itself, leaving the error messages right on top of it.

E.g. 
```
[ERROR] ...
[ERROR] ...
[INFO] ... this should always takes up only one line which inform user purely about the progress ...
```

If the following line is going to be another error message, delete the last [INFO] line, output the new [ERROR] message, then prints the next [INFO] line again.

>Overwriting has the downside that we lose the ability to see the sequence of events that happened before.

If the user wants to keep all the messages, we can add in a verbose flag afterward.
Otherwise, the log file would've contain all the necessary info in chronological order as well, in the event that error occurs.

bluein-green:
I came up with something like this:

![terminal_v3](https://user-images.githubusercontent.com/35626067/60500217-863efa80-9cec-11e9-8afb-db8783ce9019.gif)

Specifically, I did this in `CustomLogFormatter#format`, that is used by the console handler:

```
if (record.getLevel().equals(Level.INFO)) {
    builder.append("\r");
    builder.append("\u001b[1000D");
    builder.append("\u001b[K");
    builder.append(dateFormat.format(new Date(record.getMillis()))).append(" - ");
    builder.append(formatMessage(record))
} else {    // Level.WARNING and higher
    builder.append("\r");
    builder.append(dateFormat.format(new Date(record.getMillis()))).append(" - ");
    builder.append("\u001b[37;43m");
    builder.append("[ERROR]");
[...]

PR #794, opened 2019-07-05
[#705] Log: distinct warning and error messages

bluein-green:
Part of #705: making error messages more obvious in console output.

```
It is difficult to distinguish `Level.WARNING` and `Level.SEVERE`
log messages from the rest of the log messages in the console.

Users may miss out these important messages.

Let's make these messages more distinct by inserting a highlighted
`[ERROR]` status before the message.
```

### DIFF
--- a/src/main/java/reposense/system/CustomLogFormatter.java
+++ b/src/main/java/reposense/system/CustomLogFormatter.java
@@ -5,6 +5,8 @@ import static org.fusesource.jansi.Ansi.ansi;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
 import java.util.logging.SimpleFormatter;
@@ -22,6 +24,10 @@ public class CustomLogFormatter extends SimpleFormatter {
             .reset().toString();
     private static final String WARNING_HIGHLIGHT = ansi().bg(Ansi.Color.YELLOW).fg(Ansi.Color.BLACK).a("[WARNING]")
             .reset().toString();
+    private static final Map<Level, String> formatMap = new HashMap<Level, String>() {{
+            put(Level.WARNING, WARNING_HIGHLIGHT);
+            put(Level.SEVERE, ERROR_HIGHLIGHT);
+        }};
 
     public CustomLogFormatter() {
         AnsiConsole.systemInstall();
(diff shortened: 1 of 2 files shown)


bluein-green:
The `ERROR` highlight looks like this:

error_highlight

src/main/java/reposense/system/CustomLogFormatter.java
```diff
@@ -12,11 +13,17 @@
...
 
     private static final DateFormat dateFormat = new SimpleDateFormat("hh:mm:ss");
+    private static final String ERROR_HIGHLIGHT = "\u001b[37;41m[ERROR]\u001b[0m";
 
     @Override
     public synchronized String format(LogRecord record) {
         StringBuilder builder = new StringBuilder();
         builder.append(dateFormat.format(new Date(record.getMillis()))).append(" - ");
+
+        if (record.getLevel().intValue() >= Level.WARNING.intValue()) {
```
yong24s: You are appending `[ERROR]` to severity above `warning` right?


Your PR message:

>It is difficult to distinguish `Level.WARNING` and `Level.SEVERE`
>...
>Let's make these messages more distinct by inserting a highlighted
>`[ERROR]` status before the message.

Not sure it would help in readability/distinguishability.

yong24s:
@damithc @eugenepeh 

Your opinion on the [interface](https://github.com/reposense/RepoSense/pull/794#issuecomment-508701863)?

Should we make the entire error line `red`?

damithc:
I have no objections to the format shown in the screenshot. On a separate matter, the screenshot shows 'no authors found' as an error; shouldn't it be a warning instead?

src/main/java/reposense/system/CustomLogFormatter.java
bluein-green: Would specifying `[WARNING]` (in another colour perhaps) instead for warnings be more desirable?

jamessspanggg:
> shouldn't it be a warning instead?

Yes I think warning might be more suitable. My mistake there.

eugenepeh:
I am okay with either, but this is actually hard to test and may not work in certain OS.
Can explore using [jansi](https://github.com/fusesource/jansi)

bluein-green:
Updated `[WARNING]` and `[ERROR]` highlights:

image

src/main/java/reposense/system/CustomLogFormatter.java
```diff
@@ -1,22 +1,43 @@
...
     @Override
     public synchronized String format(LogRecord record) {
         StringBuilder builder = new StringBuilder();
         builder.append(dateFormat.format(new Date(record.getMillis()))).append(" - ");
+
+        if (record.getLevel().equals(Level.SEVERE)) {
+            builder.append(ERROR_HIGHLIGHT).append(" ");
+        } else if (record.getLevel().equals(Level.WARNING)) {
+            builder.append(WARNING_HIGHLIGHT).append(" ");
+        }
```
yong24s: Let's practice `open-closed principle` here.

```
Map<Level, String> formatMap  = new HashMap<Level, String>() {{
    put(Level.SEVERE, ERROR_HIGHLIGHT);
    put(Level.WARNING, WARNING_HIGHLIGHT);
}};

@Override
public synchronized String format(LogRecord record) {
    ...
    if (formatMap.containsKey(record.getLevel())) {
        builder.append(formatMap.get(record.getLevel())).append(" ");
    }
```

yong24s:
@bluein-green 

What colours are you using? I believe its `red` and `yellow` from CI logs. 

Seems like your terminal is using customized color profile 😆.

bluein-green:
Yes, I used red background with white text for `[ERROR]` and yellow background with black text for `[WARNING]`

eugenepeh:
@damithc for your review,

You should be able to see the results on the travis https://travis-ci.org/reposense/RepoSense/jobs/556686558

PR #797, opened 2019-07-09
[#764] Toolbar: fix bug of Safari browser unable to activate search functionality

### LINKED ISSUE #764: Search functionality does not work with Safari browser
As described in title:

mac-search

jamessspanggg:
Fixes #764 

```
The 'enter' key submits html forms in browsers such as Chrome and 
Firefox, but not in Safari.

Since the getFiltered() method is executed on submit of the html form,
Safari browser users are unable to activate the search functionality 
when search value is keyed in and 'enter' key is pressed.

Let's move the execution of getFiltered() method to be within 
updateFilterSearch(), to ensure that Safari users are also able to 
activate the search functionality.
```


frontend/src/index.pug
```diff
@@ -77,7 +77,7 @@ html
 
     vuetemplate#v_summary
       #summary.wrapper
-        form.summary-picker.mui-form--inline(v-on:submit.prevent="getFiltered();")
+        form.summary-picker.mui-form--inline(onsubmit="return false;")
```
fzdy1914: If it does nothing, why we still need this `"return false;"`?

frontend/src/static/js/v_summary.js
```diff
@@ -268,6 +268,7 @@ window.vSummary = {
     // model functions //
     updateFilterSearch(evt) {
       this.filterSearch = evt.target.value;
+      this.getFiltered();
```
fzdy1914: Maybe we can watch on `filterSearch` just like other flags?

frontend/src/index.pug
jamessspanggg: `return false;` is to prevent the page from reloading after submitting.

In browsers such as Chrome and Firefox, pressing enter after keying in search value will submit the form. Without `return false;`, the page will reload.

frontend/src/static/js/v_summary.js
jamessspanggg: if we create a watcher for `filterSearch`, the watcher will be triggered every time u type/modify a character, which is not ideal for our case. We want to trigger `getFiltered()` only when the user presses the key `enter` after keying the search value.

frontend/src/index.pug
jamessspanggg: The implementation is similar to the filter by glob functionality.

https://github.com/reposense/RepoSense/blob/37af0b469682d4eb73a0e9f2311ee51bc12754db/frontend/src/index.pug#L238-L240

frontend/src/static/js/v_summary.js
fzdy1914: Reasonable。

eugenepeh:
unable to review this, since i don't have safari browser.
any other mac users among the interns?

yong24s:
@damithc uses a macbook 😆.

damithc:
> @damithc uses a macbook 😆.

Yes, but not Safari. I'm running Windows in the Macbook :-p

yong24s:
Tested and works on Safari 12.0.3.

PR #799, opened 2019-07-09
[#705] Restrict command line output to 1 line

bluein-green:


bluein-green:
Closing because solution doesn't work when terminal has to scroll.

PR #800, opened 2019-07-10
[#763] v_summary.js: ensure getFiltered() is only called once when reloading page

### LINKED ISSUE #763: Redundant getFiltered() method calls when reloading page
When reloading the page, the data in v_summary such as `filterGroupSelection, sortGroupSelection, sortWithinGroupSelection, filterTimeFrame...` will change based on the URL hash values. When the data changes according to the URL hash values, the watch methods that watches for changes in those data will be triggered. The watch methods below will execute `getFiltered()` whenever it's watched data changes.
Screen Shot 2019-06-24 at 11 24 31 AM
 
This means that when reloading the page, **there will be a number of calls to `getFiltered()` method but only one call is needed to filter the repositories**. This will affect the performance of the large reports.

To replicate, change some of the values in the summary controls. Then open the chrome console tab, place a breakpoint on the `getFiltered()` method in v_summary.js and observe the call stack when reloading the page.

getFiltered

jamessspanggg:
Fixes #763 

```
When reloading the report page, the data in v_summary such as
filterSinceDate, filterUntilDate, filterTimeFrame etc will change based 
on the URL hash values. The changes in data will trigger the watchers 
for those data, and each of them will call getFiltered() method.

Only one call for getFiltered() is required after rendering the filter 
hash, the other getFiltered() calls by the watchers are redundant.
When the report is large in size, repeatedly executing getFiltered()
when reloading the report will also significantly affect the loading 
performance of our report.

Let's only allow watchers to execute getFiltered() after the first 
cycle of watcher updates to prevent redundant calls to getFiltered().
```


jamessspanggg:
@fzdy1914 The number of modified toolbar controls. (Modified as in not the same as the default value)

By default, the watcher for tmpFilterSinceDate and tmpFilterUntilDate will be called as their default values are empty strings ‘ ‘.

So if the user changes group by to ‘groupByAuthors’ and granularity to ‘week’ (both are different compared to their default values), getFiltered will be called 5 times. (1 time for created() hook in v_summary, 2 times for tmpFilterSinceDate and tmpFilterUntilDate, and another 2 times for group by and granularity)

frontend/src/static/js/v_summary.js
```diff
@@ -117,27 +117,40 @@ window.vSummary = {
...
       contributionBarColors: {},
+      isWatcherUpdated: false, // keep track of whether first cycle of watcher updates is finished
     };
   },
   watch: {
     repos() {
-      this.getFiltered();
+      if (this.isWatcherUpdated) {
+        this.getFiltered();
+      }
```
yong24s: Should this be method with a descriptive name?

jamessspanggg: Yes I think it would be better to make this as a method.

frontend/src/static/js/v_summary.js
```diff
@@ -407,6 +408,11 @@ window.vSummary = {
         this.contributionBarColors = colors;
       });
     },
+    getFilteredAfterWatcherUpdated() {
+      if (this.isWatcherUpdated) {
+        this.getFiltered();
```
eugenepeh: instead of doing this, why not just return in `getFiltered` if `this.isWatcherUpdated` is false instead?

PR #801, opened 2019-07-10
[#723] Fix bug of Safari browser not able to modify sinceDate and untilDate

### LINKED ISSUE #723: Unable to modify sinceDate and untilDate for Safari browser
When using Safari browser to launch the report, I am unable to modify the two dates as the controls does not appear when I hover my cursor over the dates.
Screen Shot 2019-06-03 at 11 45 24 AM

jamessspanggg:
Fixes #723 

```
Safari browser users are unable to modify sinceDate and untilDate in
the report.

This is due to the datepickers used are HTML5 elements; they are not
supported by Safari. Furthermore, the alternative to modify the date
fields, through the keydown events, are disabled.

Let's allow Safari browser users to use a text-type input for their
sinceDate and untilDate inputs. They can modify the dates by
keying the numeric date values into the date fields followed by hitting
the key, `enter`.
```


jamessspanggg:
Yep that is correct. But users should know not to input characters for the date field. Even if the date field consists of characters, the ramp chart won’t be changed until a correct date format is inputted.

eugenepeh:
I am okay with the solution, would be best if we can limit the input type to only digits using e.g. `type="number"`

jamessspanggg:
> would be best if we can limit the input type to only digits using e.g. `type="number"`

May not be possible as the date's format includes the `-` character as well. eg: `2018-08-31`

frontend/src/index.pug
```diff
@@ -113,12 +113,12 @@ html
                 option(value="week") Week
               label granularity
             .mui-textfield.summary-picker__date
-              input(type="date", v-model="tmpFilterSinceDate", onkeydown="return false",
-                v-bind:min="minDate", v-bind:max="filterUntilDate")
+              input(type="date", v-model="tmpFilterSinceDate", v-bind:min="minDate", v-bind:max="filterUntilDate",
+                onkeypress="return (event.charCode >= 48 && event.charCode <= 57) || event.charCode === 45")
```
fzdy1914: Can consider making it a method.

jamessspanggg:
Updated implementation:

- If Safari browser is detected, a text-type input will be used for the date fields. The date fields can only be numerical or the `-` character. Date filters will only take place after hitting enter.
- For other browsers, it is the same implementation as previous. (modifying the date fields by typing is not allowed)

yong24s:
@eugenepeh 

The implementation works exactly as described.

As `0-9` and `-` are permitted, user may enter faulty dates like `20188-12-12` and `2018--12-12`.
The report managed to parse these two faulty dates and there are changes made to the ramp chart.

Shouldn't we block faulty dates and show red line under the date picker?

jamessspanggg:
> Shouldn't we block faulty dates and show red line under the date picker?

Dates with invalid format is now blocked and a red line will be shown under until correct format is typed.

eugenepeh:
>Dates with invalid format is now blocked and a red line will be shown under until correct format is typed.

Instead of doing this, how about we try to make it easier for the users? 
Similar to how those CD key entering field work, we can help them by inserting the dash for them when they arrive at specific position, and disable them from entering dashes at all.

jamessspanggg:
I have added the inserting of dash when the user first arrives with 4 digits such as `2018` and after the  6th digit such as `2018-08`. However, I kept the blocking of invalid dates as well as the red line because users might still hit `enter` before completing the date format.

Illustration of dash inserting:

date

eugenepeh:
@yong24s @fzdy1914 can both of you review the change?

frontend/src/static/js/v_summary.js
```diff
@@ -542,6 +544,49 @@ window.vSummary = {
...
+        // invalid until date detected
+        event.currentTarget.style.borderBottomColor = 'red';
+      }
+    },
+
+    validateInput(event) {
+      const key = event.keyCode;
+      const isIntegerKey = (key >= 48 && key <= 57);
+      const isArrowKey = (key >= 37 && key <= 40);
+      const isOtherKey = (key === 8 || key === 13 || key === 46); // backspace, enter, delete keys
```
yong24s: Should also include the following keys for convenience on a 104-key keyboard:
Integers: `96-105`
Dash: `109`

frontend/src/static/js/v_summary.js
```diff
@@ -542,6 +544,49 @@ window.vSummary = {
...
+      if (!(isIntegerKey || isOtherKey || isArrowKey)) {
+        event.preventDefault();
+      }
+    },
+
+    formatInputDate(event) {
+      const date = event.target.value;
```
yong24s: Possible to intercept `delete`/`backspace`, then check if last char is `-` and remove last two chars.

This will make the control more coherent when modifying.

jamessspanggg: Has updated based on the comments:

date

jamessspanggg:
> just wondering, is it possible to export the safari apis to separate js?

Yes I think it is possible. Would you prefer that?

eugenepeh:
>Yes I think it is possible by making the safari js a component of v_summary.js. Would you prefer that?

Yeah, I think it will be better that way to separately maintain the set of safari apis while not cluttering the main apis.

jamessspanggg:
> just wondering, is it possible to export the safari apis to separate js?

I have separated the safari methods to a separate `safari_date.js`.

I have tried a few ways of separating the methods into a new js file but only this way of doing it works. I am unsure whether this is an ideal way.

frontend/src/index.pug
```diff
@@ -115,14 +115,14 @@ html
             .mui-textfield.summary-picker__date
               input(v-if="isSafariBrowser", type="text", placeholder="yyyy-mm-dd", v-bind:min="minDate",
                 v-bind:max="filterUntilDate", v-bind:value="tmpFilterSinceDate", v-on:keyup.enter="updateTmpFilterSinceDate",
-                v-on:keydown="validateInputDate($event); deleteDashInputDate($event);", v-on:input="appendDashInputDate", maxlength=10)
+                onkeydown="formatInputDateOnKeyDown(event)", oninput="appendDashInputDate(event)", maxlength=10)
```
eugenepeh: double indent for line wrap

PR #802, opened 2019-07-10
[#661] Change chart title format

### LINKED ISSUE #228: Use a counter to ensure created folders are unique
From @yamidark comment in #210 

> Hmm, this may give rise to the same problem mentioned in #206, whereby we give 2 different locations but the folder name is same, and use the same branch, end up overriding each other's report.
Perhaps we can temporarily set the repoName to be the absolute path, just replace the file separators with _?

In #210, location accepts both `github url` and `disk location`.

`Display name`is used for the name of the cloned file, which are generated by normalizing a path or a git url.

For example, `https://github.com/organization/main.git` -> `organization_main_branch` and
`C:/path/to/repository` -> `repository_branch`.

Thus, it is possible for a repository with a similar name and branch to interfere with RepoSense's analysis.

### LINKED ISSUE #661: Tweak chart title format

bluein-green:
Resolves #661.

```
The chart title uses underscores to separate `org`, `repo`
and `branch`. Furthermore, `displayName` is used for both 
the chart title and repo analysis output folder name.

`org`/`repo`/`branch` might have underscores in them, making it
difficult to discern each term. Using a single parameter `displayName`
to represent both the output folder name and the chart title would
constrain us to not use `/`. Thus we cannot format the
chart title as `org/repo[..]` in the backend.

Let's use a different format `org/repo[branch]` as the chart title.
```


src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -65,11 +66,13 @@ public RepoConfiguration(RepoLocation location, String branch, List<Format> form
...
         if (organization != null) {
-            displayName = organization + "_" + repoName + "_" + branch;
+            displayName = organization + "/" + repoName + "[" + branch + "]";
             repoFolderName = organization + "_" + repoName;
+            outputFolderName = organization + "_" + repoName + "_" + branch;
         } else {
-            displayName = repoName + "_" + branch;
+            displayName = repoName + "[" + branch + "]";
             repoFolderName = repoName;
+            outputFolderName = repoName + "_" + branch;
```
yong24s: Can help to restructure this `if-else` block for readability.

```
displayName = repoName + "[" + branch + "]";
...

if (organization != null) {
    displayName = organization + "/" + displayName;
    ...
}

```

bluein-green:
@yong24s 

I meant that if we were to use only 1 variable `displayName` for 2 uses (as the name of the output folder and the chart title), we would be constrained to not use `/`. Thus we cannot have `org/repo[..]` which is what we want for the chart title.

src/main/java/reposense/model/RepoConfiguration.java
yong24s: @bluein-green 

Can do the same for the rest in the `else` block too?

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -63,12 +64,14 @@ public RepoConfiguration(RepoLocation location, String branch, List<Format> form
...
         String organization = location.getOrganization();
         String repoName = location.getRepoName();
+        displayName = repoName + "[" + branch + "]";
+        outputFolderName = repoName + "_" + branch;
 
         if (organization != null) {
-            displayName = organization + "_" + repoName + "_" + branch;
             repoFolderName = organization + "_" + repoName;
+            displayName = organization + "/" + displayName;
+            outputFolderName = organization + outputFolderName;
```
yong24s: What I want is:

```
repoFolderName = repoName;
displayName = repoName + "[" + branch + "]";
outputFolderName = repoName + "_" + branch;

if (organization != null) {
    repoFolderName = organization + "_" + repoFolderName ;
    displayName = organization + "/" + displayName;
    outputFolderName = organization + outputFolderName;
}
```

Is it more readable? 🤔

bluein-green: It would be. But `repoFolderName` is final, so we can't prepend `organization` to it.

yong24s: Hmm, I don't think its that critical to use `final` for `repoFolderName` since we only use it through its getter. 🤔 

@eugenepeh, Should we remove the `final` keyword for better code readability?

eugenepeh: looks like its resolved

bluein-green:
I meant that if we used a single parameter `displayName` for 2 uses (as the output folder name and chart title), we would be constrained to not use `/`.

Would it be better if the rationale is reworded as:

```
Using a single parameter `displayName` to represent both
the output folder name and the chart title would
constrain us to not use `/`. Thus we cannot format the
chart title as `org/repo[..]` in the backend.
```

> Can just be shorten to , `Let's use a different format `org/repo[branch]` for the chart title.
> The folder name didn't change so there isn't really a need to emphasize on that.

Ok, I see.

eugenepeh:
>Using a single parameter `displayName` to represent both
the output folder name and the chart title would
constrain us to not use `/`. Thus we cannot format the
chart title as `org/repo[..]` in the backen

Sounds much better. 👍

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -63,12 +64,14 @@ public RepoConfiguration(RepoLocation location, String branch, List<Format> form
...
+        outputFolderName = repoName + "_" + branch;
 
         if (organization != null) {
-            displayName = organization + "_" + repoName + "_" + branch;
             repoFolderName = organization + "_" + repoName;
+            displayName = organization + "/" + displayName;
+            outputFolderName = organization + "_" + outputFolderName;
         } else {
-            displayName = repoName + "_" + branch;
             repoFolderName = repoName;
```
yong24s: Move out of the `if-else` block, as suggested previously.

It's okay to remove the `final` keyword on `repoFolderName`.

bluein-green: Ok.

eugenepeh:
A little more nit for the proposed message:

>`org`/`repo`/`branch` might have underscores in them, making it
difficult to discern each term. Using a single parameter `displayName`
to represent both the output folder name and the chart title would
constrain us to not use `/`. Thus we cannot format the
chart title as `org/repo[..]` in the backend.

`org`, `repo` and `branch`  ...

This has shown that the output folder name and the chart title have little compatibility, and difficulty in coming up with a coherent format.

>Let's use a different format `org/repo[branch]` as the chart title.

Let's use a separate variable, `outputFolderName`, for the out folder name so that a different format, `org/repo[branch]`, can be used for the chart title.

PR #812, opened 2019-07-12
[#807] RepoConfiguration: Make `repoFolderName` unique

### LINKED ISSUE #807: Ensure `repoFolderName`s are unique
If we have 2 (local) repos with paths:
- .... /dir1/name_of_repo
- .... /dir2/name_of_repo

`repoFolderName` for the 2 `RepoConfiguration`s will be "name_of_repo". Thus for both repos, the folder that contains the cloned data will be named `/repos/name_of_repo`.

Consequently, when repo 1 has been cloned, and RepoSense attempts to clone repo 2 and analyse repo 1 in parallel, the analysis of repo 1 would try to refer to `/repos/name_of_repo`, which could have been deleted (as part of cloning repo 2) or could contain cloned data of repo 2.

On my system, running Reposense with 2 such local repos together results in analysis failure for both repos, while doing so 1 repo at a time generates the reports successfully.

I believe this is somewhat related to #228.

**Suggestion**
Append `location.hashCode()` to the `repoFolderName` when constructing each `RepoConfiguration`. Using the hashCode provides a deterministic and (mostly) unique way of identifying each folder as compared to appending a count value or such.

Or use a name-based UUID derived from the full `location` of the repo for the `repoFolderName`

bluein-green:
Resolves #807.

```
Cloning 2 repos with similar repo/organisation names might result in
2 `RepoConfiguration`s with identical `repoFolderName`.

The folder containing cloned data will both be named
`/repos/repoFolderName`. Hence, when one repo is being analysed while
the other is being cloned in parallel, the folder contents will be
modified, resulting in interference in the analysis and thus analysis
failure.

Let's append a unique identifier to the `repoFolderName`. This
identifier is a UUID generated deterministically from the full 
location name of a repo, which should be unique for different repos.
```


bluein-green:
@fzdy1914 

What is the value of the `uniqueIdentifier` in `RepoConfiguration` for each repo when you ran with those arguments?

From my tests using repo location strings `E:/reposense` and `D:/reposense`, the `uniqueIdentifier`s are `2ed2e2ab-f2ba-3257-9ce8-7bdfa52a5308` and `1c456fb5-69f2-3286-b6a8-1394275a7c90` respectively.

fzdy1914:
> @fzdy1914
> 
> What is the value of the `uniqueIdentifier` in `RepoConfiguration` for each repo when you ran with those arguments?
> 
> From my tests using repo location strings `E:/reposense` and `D:/reposense`, the `uniqueIdentifier`s are `2ed2e2ab-f2ba-3257-9ce8-7bdfa52a5308` and `1c456fb5-69f2-3286-b6a8-1394275a7c90` respectively.

The unique identifier is the same as yours. But the generated report only contains one repo(as shown in my image before).

bluein-green:
I see. I believe that is the expected behaviour as per #228 since the `displayName`s of both repos are identical, hence the analysis output folder names are identical and one analysis output might override the other. That will be fixed in another PR after #661 is merged because I will be ensuring the `outputFolderName` parameter is unique. 

For this PR, for your inputs, before `RepoCloner#cleanup` is executed, there would be 2 folders in the `repos` folder. On master, there should only be 1 folder because of overwriting of the folder when the second repo is cloned.

bluein-green:
Closing this as the work done in here is superseded in #829.

PR #814, opened 2019-07-13
Security patch: update cypress to 3.4.1

yong24s:
```
A potential security vulnerability exists in one of our indirect
dependencies, lodash, which requires immediate update to version 4.17.13
or later. As cypress is our only dependency that utilizes the vulnerable
version of lodash, updating it will hence resolve the potential security
issue.

Let's update cypress to version 3.4.1, which is the earliest version
that uses lodash version 4.17.13 or later.
```


eugenepeh:
@yong24s why the need to update cypress?

yong24s:
Its using an outdated insecure dependency.

eugenepeh:
will need a propose message for that too though

E.g.

A potential security vulnerability exists in one of our indirect dependencies, lodash, which requires immediate update to version 4.17.13 or later. As cypress is our only dependency that utilizes the vulnerable version of lodash, updating it will hence resolve the potential security issue.

Let's update cypress to version 3.4.0.

eugenepeh:
btw @yong24s the rationale for 3.4.0, is it cause its latest version or is it the earliest version that uses lodash version 4.17.13 or later? do include that in the message too

yong24s:
🤔 The changes on their github is still on the develop branch, have to wait until it get released.

PR #824, opened 2019-07-19
[#817] Cypress: add code view loading time benchmark test

### LINKED ISSUE #817: Add benchmark test for opening of code view
To ensure that speed of loading code view remain unchanged

bluein-green:
Resolves #817.

```
We do not have a benchmark for the time taken to load code view. 

This makes it difficult to ascertain whether new code added will slow
down its opening, which will result in a less positive user experience.

Let's add a benchmark test for the opening of the code view. The
threshold for the load time is 8s, with the buffer time being
the larger of 3s or 10% of the threshold. If out of 5 trials, the
load times of all trials exceed the maximum time, the test is deemed to
have failed.
```


frontend/cypress/tests/codeView_load_benchmark.js
```diff
@@ -0,0 +1,34 @@
+describe('load code view benchmark', function() {
+  const MAXIMUM_LOADING_TIME = 8000;
```
jamessspanggg: How did you determine the threshold of 8 seconds?

The current arguments used for Cypress test is `--repo https://github.com/reposense/RepoSense.git --view` which will analyse the repository from one month from today's date until today. In the future if the author Eugene makes a lower contribution compared to this month, even if the code view loading speed has decreased, this benchmark test will not be able to detect it as the author Eugene has lower contribution. 

Unless we set the arguments used for Cypress test to be within a constant duration, using a constant threshold may not be a good idea for this benchmark test.

yong24s:
Depending on Travis-CI load, the test timing can fluctuate greatly. 

For tests that involve resolving of elements with `cy.get`, an usual 1 second test may take more than 30 seconds to complete. Hence, we have increased the default timeout of 30 seconds but sometimes it is still not long enough. 

@eugenepeh 
We may need to do some baseline test. Like have a timing test for a test that normally complete in 1 second and use it as a factor? Like ten_seconds = 10 * factor, where factor >=1?

eugenepeh:
>We may need to do some baseline test. Like have a timing test for a test that normally complete in 1 second and use it as a factor? Like ten_seconds = 10 * factor, where factor >=1?

I think we can set number of retries, if all the retries have hit more than the threshold and the buffer, then we can deem that the test has failed the expectation. For now, I guess we can just the buffer to be 3 seconds or 10% (whichever is greater) more than the usual time of completion? 

Given that the resources are already allocated to us by travis, I don't think a little fluctuation will be sufficient enough to jeopardize the test. We can always review the timing afterward if it happens.

frontend/cypress/tests/codeView_load_benchmark.js
```diff
@@ -0,0 +1,64 @@
...
+
+          cy.log(`trial ${i+1} loading time: ${loadingTimeSeconds.toFixed(3)}s`);
+
+          loadingTimes.push(loadingTime);
+        });
+  };
+
+
+  for (let i = 0; i < NUM_TRIALS; i++) {
+    it(`time taken to load code view (trial ${i+1})`, function() {
```
eugenepeh: just to check, does this restart the test from scratch every time?

bluein-green: Yes. In each test, the base url will be visited before the actual test.

bluein-green:
@eugenepeh 

bluein-green:
> Where's the 8s threshold?

Line 3:
https://github.com/reposense/RepoSense/blob/e0708b7d7d4d0c319685196faa14d38281fbe8bf/frontend/cypress/tests/codeView_load_benchmark.js#L3

eugenepeh:
>Line 3:

Thanks, sorry I must be a little too tired that I overlooked it.

yong24s:
@eugenepeh 

It failed due to high load.

eugenepeh:
>@eugenepeh
It failed due to high load.

No, the thing is, it ran for 5 trials, only one has failed. 
Theoretically, it should still be a passed. Hence, that's an error with the code.
Anyway, @bluein-green can you code it such that 
* it takes care of that scenario
* also, don't have to make it run 5 full times if it pass on either of previous runs.

frontend/cypress/support.js
```diff
@@ -6,7 +6,7 @@ Cypress.Screenshot.defaults({
 });
 
 beforeEach(() => {
-  cy.visit('/');
+  cy.visit('/', { timeout: 120000 });
```
yong24s: Instead of increasing this, it is possible to to use `try-catch` to handle `cy.get` timeouts?

eugenepeh: yes, use a try-catch to bypass the errors.

bluein-green: As far as I know, we can't bypass the errors from `cy.visit()` in this case. See [here](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress.html#You-cannot-add-a-catch-error-handler-to-a-failed-command)

PR #829, opened 2019-07-22
[#228, #807] Ensure `repoFolderName`, `outputFolderName` and `displayName` are unique

### LINKED ISSUE #228: Use a counter to ensure created folders are unique

### LINKED ISSUE #807: Ensure `repoFolderName`s are unique

bluein-green:
Resolves #228.
Resolves #807. 

```
Local repos with the same folder name but different locations will give
rise to the same `repoName` in `RepoLocation`. As a result, the
`repoFolderName`, `outputFolderName` and `displayName` will be 
identical.

Implications:
- Cloning 2 such local repos will result in 2 `RepoConfiguration`s
  with identical `repoFolderName`. The folder containing cloned data
  will both be named `/repos/repoFolderName`. Hence, when one repo is
  being analysed while the other is being cloned in parallel, the
  folder contents will be modified, resulting in interference in the
  analysis and thus analysis failure.
- Assuming that the repos are cloned successfully, the 2
  `RepoConfiguration`s will have idential `outputFolderName`, and one
  repo's analysis output will be overwritten by the other.
- Assuming the 2 output folders are created successfully, since the
  `displayName`s are identical and the frontend uses the 
  `displayName` as the `window.REPOS` object's keys, only one repo's
  output will be displayed in chart view.
  
Let's check for uniqueness of the `repoName` in the `RepoLocation`
constructor, and append the number of duplicated names so far to the
duplicate `repoName` to ensure it is unique.
```


bluein-green:
This PR supersedes #812

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -143,6 +148,22 @@ public static void setStandaloneConfigIgnoredToRepoConfigs(
         configs.stream().forEach(config -> config.setStandaloneConfigIgnored(isStandaloneConfigIgnored));
     }
 
+    /**
+     * Ensures that the {@code displayName} of each {@code RepoConfiguration} is unique.
+     */
+    public static void makeDisplayNamesUnique(List<RepoConfiguration> configs) {
```
jylee-git: The doc comments should be more comprehensive; something like 

`Appends a unique identifier to the display name if more than one {@link RepoConfiguration} share the same display name`.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -255,12 +276,27 @@ public void setBranch(String branch) {
         authorConfig.setBranch(branch);
     }
 
+    /**
+     * Replaces the branch parameter in {@code displayName} with {@code branch}.
+     */
     public void updateDisplayName(String branch) {
```
jylee-git: Can be private method also

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -143,6 +148,22 @@ public static void setStandaloneConfigIgnoredToRepoConfigs(
...
+     * Ensures that the {@code displayName} of each {@code RepoConfiguration} is unique.
+     */
+    public static void makeDisplayNamesUnique(List<RepoConfiguration> configs) {
+        Map<String, Integer> uniqueNames = new HashMap<>();
+        for (RepoConfiguration config : configs) {
+            if (!uniqueNames.containsKey(config.getDisplayName())) {
+                uniqueNames.put(config.getDisplayName(), 1);
+            } else {
+                String currName = config.getDisplayName();
+                config.appendUniqueIdToDisplayName("(" + uniqueNames.get(currName).toString() + ")");
```
jylee-git: I think the parentheses should be handled in `appendUniqueIdToDisplayName` instead

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -255,12 +276,27 @@ public void setBranch(String branch) {
...
     }
 
+    public void appendUniqueIdToDisplayName(String toAppend) {
+        this.displayName = displayName + " " + toAppend;
+    }
+
+    /**
+     * Replaces the branch parameter in {@code outputFolderName} with {@code branch}.
+     */
     public void updateOutputFolderName(String branch) {
```
jylee-git: This can be changed to private as well

src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -299,4 +310,33 @@ public void repoConfig_overrideStandaloneConfig_success()
...
+        Assert.assertNotEquals(config1.getOutputFolderName(), config2.getOutputFolderName());
+    }
+
+    @Test
+    public void repoConfig_makeDisplayNamesUnique_success() throws InvalidLocationException {
+        RepoConfiguration config1 = new RepoConfiguration(new RepoLocation(localTestRepo1));
+        RepoConfiguration config2 = new RepoConfiguration(new RepoLocation(localTestRepo2));
+        List<RepoConfiguration> configs = new ArrayList<>();
+        configs.add(config1);
+        configs.add(config2);
```
jylee-git:
```java
List<RepoConfiguration> configs = Arrays.asList(config1, config2);
```

yong24s:
As mentioned by @jylee-git, the problem of using `UUID` is its overkill for a not so common problem but now everyone is paying the same price (the cost of limited path length).

We only need to guarantee that no `outputFolderName` is the same, then we don't need cryptographic hashes.

Is it possible for `RepoCloner` to check if `outputFolderName` already exists, if so append a digit to its name?

fzdy1914:
I agree with the point of @yong24s.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -143,6 +145,86 @@ public static void setStandaloneConfigIgnoredToRepoConfigs(
...
+            if (!displayNames.get(i).equals(uniqueDisplayNames.get(i))) {
+                configs.get(i).setDisplayName(uniqueDisplayNames.get(i));
+            }
+        }
+    }
+
+    /**
+     *
+     * @param configs
+     */
```
yong24s: Incomplete JavaDocs.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -143,6 +145,86 @@ public static void setStandaloneConfigIgnoredToRepoConfigs(
...
+            if (!outputFolderNames.get(i).equalsIgnoreCase(uniqueOutputFolderNames.get(i))) {
+                configs.get(i).setOutputFolderName(uniqueOutputFolderNames.get(i));
+            }
+        }
+    }
+
+    /**
+     *
+     * @param configs
+     */
```
yong24s: Incomplete JavaDocs.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -143,6 +145,86 @@ public static void setStandaloneConfigIgnoredToRepoConfigs(
...
+                configs.get(i).setRepoFolderName(uniqueRepoFolderNames.get(i));
+            }
+        }
+    }
+
+    /**
+     *
+     * @param names
+     * @return
+     */
```
yong24s: Incomplete JavaDocs.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -255,12 +341,25 @@ public void setBranch(String branch) {
...
     }
 
-    public void updateDisplayName(String branch) {
-        this.displayName = displayName.substring(0, displayName.lastIndexOf('[') + 1) + branch + "]";
+    /**
+     * Replaces the branch parameter in {@code displayName} with {@code branch}.
+     */
+    private void updateDisplayName(String branch) {
+        int uniqueIdIndex = displayName.lastIndexOf(' ');
+        String uniqueId = (uniqueIdIndex == -1) ? "" : displayName.substring(uniqueIdIndex);
```
yong24s: The same code is used in `RepoConfiguration#updateOutputFolderName`

Can refactor this into another common method.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -143,6 +145,86 @@ public static void setStandaloneConfigIgnoredToRepoConfigs(
         configs.stream().forEach(config -> config.setStandaloneConfigIgnored(isStandaloneConfigIgnored));
     }
 
+    /**
+     * Appends a unique identifier to the {@code displayName} of a {@code RepoConfiguration} when
```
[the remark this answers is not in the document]
bluein-green: It is supposed to be ['a' for "unique"](https://dictionary.cambridge.org/dictionary/english/unique)

bluein-green:
> It may be simpler and transparent to make names unique in `RepoConfiguration` constructor using `RepoConfiguration#makeNamesUnique`.

@yong24s 
I tried implementing that with a static `Map` to keep track of unique names, but it did not work.

I've decided to go about this in a different way. For local repos with duplicate names, their `repoName` in `RepoLocation` will be the same. Hence, in the constructor of `RepoLocation`, the `repoName` will be made unique, and as a result, the `repoName` used in the constructor of `RepoConfiguration` will be unique.

src/main/java/reposense/model/RepoConfiguration.java
bluein-green: Oops, forgot to remove those

src/test/resources/RepoConfigurationTest/repoconfig_uniqueIdentifier_test2/local_alpha/test1.txt
```diff
@@ -0,0 +1 @@
+ hello
```
eugenepeh: What are this 2 files for? 
If its meant to allow the addition of empty folders, please do the gitignore solution in https://stackoverflow.com/questions/115983/how-can-i-add-an-empty-directory-to-a-git-repository

Do state the reason in the comments.

bluein-green: Yes, it was to allow for the addition of empty folders. I have used the gitignore solution.

fzdy1914:

PR #830, opened 2019-07-22
[#779] Restore zoom tab

### LINKED ISSUE #779: Restore zoom tab
zoom tab is currently not restore-able with url and we should aim to fix this soon

jamessspanggg:
Fixes #779

```
The zoom tab hashes are not encoded in the URL when zoom tab is opened.

Zoom tab will not be restorable with the URL after reloading the page.

Let's add the encoding of zoom tab hashes to the URL make the zoom tab 
restorable after reloading the page.
```


frontend/src/static/js/main.js
```diff
@@ -211,6 +215,23 @@ window.app = new window.Vue({
...
     },
 
+    renderZoomTabHash() {
+      const hash = window.hashParams;
+      const info = {
+        author: hash.tabAuthor,
+        repo: hash.tabRepo,
+        avgCommitSize: hash.avgCommitSize,
+        sinceDate: hash.zoomSince,
+        untilDate: hash.zoomUntil,
```
eugenepeh: would be better to name the key to be the same as the hash to avoid some misinterpretation.

jamessspanggg: the current work does not support merge ramp yet.

eugenepeh: then i guess either hold this one or the other. Otherwise we will end up reviewing things that might not get merged.

jamessspanggg: holding the merge group option will be a better choice.

jamessspanggg: Holding on to the merge group option PR will be a better choice.

jamessspanggg: Actually, since the merge group option PR has a higher priority and restoring the merge ramps would be quite complicated, putting this PR on hold first might be a better choice.

frontend/src/static/js/main.js
```diff
@@ -182,6 +182,10 @@ window.app = new window.Vue({
     },
 
     updateTabAuthorship(obj) {
+      // remove hashes from zoom tab if available
+      window.removeHash('zoomSince');
+      window.removeHash('zoomUntil');
+      window.removeHash('avgCommitSize');
```
fzdy1914: Can I know why we have this `avgCommitSize` stored in the hash?  Shouldn't it be calculated directly from the data?

frontend/src/static/js/v_summary.js
```diff
@@ -691,6 +691,17 @@ window.vSummary = {
...
+  beforeMount() {
+    this.$root.$on('restoreCommits', (user) => {
+      this.getUserCommits(user);
+      if (this.filterTimeFrame === 'week') {
+        this.splitCommitsWeek(user);
+      }
+    });
+  },
+  beforeUpdate() {
+    this.canGetFiltered = true; // enable getFiltered() after first cycle of watcher updates
```
fzdy1914: Seems unrelated with this PR.

frontend/src/static/js/main.js
jamessspanggg: Yes it can be calculated from the data but that will require another method that calculates that. Whereas this only require to pass the already computed `avgCommitSize` to the hash.

fzdy1914:
Hi @jamessspanggg , should we restart this PR as it is quite important? And since `openZoomTab` function has already be updated, this issue should be easier to be implemented right?

fzdy1914:
Reminder to continue on this PR. :)

jamessspanggg:
@fzdy1914 noted. Will be working on this.

jamessspanggg:
@fzdy1914 just to clarify, when the user opens the zoom view for a specific period, then he/she flips the settings for `merge group` or `granularity`. When reloading the page, should the same zoom view appear?

fzdy1914:
> @fzdy1914 just to clarify, when the user opens the zoom view for a specific period, then he/she flips the settings for `merge group` or `granularity`. When reloading the page, should the same zoom view appear?

I am unable to determine the use case either. To follow the current implementation of  v_authorship, I personally think it is necessary to show the same zoom view as then we can perfectly reproduce the page. 

Even if we want to disable the zoom in the future, it can be easily achieved from the previous implementation by just removing the hash when changing the setting.

jamessspanggg:
> To follow the current implementation of v_authorship, I personally think it is necessary to show the same zoom view as then we can perfectly reproduce the page.

Ok. In that case, then the solution to this problem may not be as simple, since the settings (such as `merge group`, `filterGroupBy`, `sortGroupsBy`, `sortWithinGroupsBy`) of the v_summary can be different than v_zoom.

Also, currently a computed property is used to represent the filtered user, which governs the behaviour of #1026, is this behaviour intended?

fzdy1914:
> Ok. In that case, then the solution to this problem may not be as simple, since the settings (such as `merge group`, `filterGroupBy`, `sortGroupsBy`, `sortWithinGroupsBy`) of the v_summary can be different than v_zoom.

Can't we just add `start, end, repo, i` these four extra attribute?
> 
> Also, currently a computed property is used to represent the filtered user, which governs the behaviour of #1026, is this behaviour intended?

It is not intended and should be fixed.

fzdy1914:
> It is not intended and should be fixed.

Let me carefully reconsider the logic.

jamessspanggg:
> Can't we just add `start, end, repo, i` these four extra attribute?

I don't think that's sufficient. The order of filtered repositories in the v_summary can be different than the filtered repository in v_zoom, before reloading (in terms of `filterGroupSelection`, `merge group`, `sort groups by`, `sort within groups by` etc). To get back the filtered repository in v_zoom after reloading, we cannot depend on the filtered repositories in v_summary. Instead, we must specifically filter the repositories based on the control settings in v_summary that is right before we activate the zoom feature (rather than the control settings before reloading).

Should this PR be halt until the implementation regarding #1026 is confirmed?

eugenepeh:
>Should this PR be halt until the implementation regarding #1026 is confirmed?

Can proceed based on the conclusion of our discussion in #1026 

>For now, I would think that having the correct date range and the last opened (the author) zoom tab would be sufficient. It is normal to have some non-important controls get reset back to default upon refresh etc, especially for a non-critical component like zoom tab

jamessspanggg:
@fzdy1914 I'm currently stuck on how to approach this PR. But I came up with an idea, where the zoom view is independent of the summary view's settings:

v_summary passes all the settings such as `filterTimeFrame, filterGroupSelection, i, merge group, sorting groups by, sort within groups by... etc` to v_zoom when opening tab zoom. When the page is reloaded, v_zoom first gets the settings from the hash. It then calls a method in v_summary that filters the repositories based on those settings (`similar to getFiltered() in v_summary`), to get back the specific filtered repository at the moment where zoom tab was previously opened. The filtered repository will then be passed back to v_zoom for display.

There are some downsides to this approach:
- a lot of variables being passed from v_summary to v_zoom ( `filterTimeFrame, filterGroupSelection...`)
- overhead to filter all repositories once again, with the previous settings at the moment where zoom tab is opened

May I get your opinion on this?

fzdy1914:
> May I get your opinion on this?

I am OK to it, however, the overhead may be too high to care a `not very important feature`.  

Since you have already implemented it, @eugenepeh Can I get your input for this?

frontend/src/static/js/main.js
```diff
@@ -188,17 +188,38 @@ window.app = new window.Vue({
...
+    },
+
+    removeZoomHashes() {
+      window.removeHash('zoomSince');
+      window.removeHash('zoomUntil');
+      window.removeHash('repoIndex');
+      window.removeHash('userIndex');
+      window.removeHash('avgCommitSize');
+      window.removeHash('zoomFilterGroupSelection');
+      window.removeHash('zoomFilterTimeFrame');
```
eugenepeh: Sorry haven't been following for a while, whats the diff between `zoomFilterTimeFrame` and `zoomSince` + `zoomUntil`?

jamessspanggg: `zoomFilterTimeFrame` refers to the `granularity` of each ramp slice, `zoomSince` and `zoomUntil` refers to the since and until date of the commits in zoom view.

All these settings are recorded at the moment when user opens the zoom tab, to make sure that changes (in v_summary) after the opening the zoom tab won't affect the zoom tab.

eugenepeh: I see, just realized that the same convention is being used on summary. Maybe we should rename it to match the fields that they're intended for to avoid confusion.

fzdy1914:
@eugenepeh , If you are OK with the overhead caused, I can continue with the reviewing process.

eugenepeh:
>@eugenepeh , If you are OK with the overhead caused, I can continue with the reviewing process.

I am ok to have all the minor params to be added to the hash (even though there isn't a need to), but we should minimally shorten the params name, and also use a prefix (e.g. all zoom tabs hashes should begins with z e.g `zoomSince -> zs`) to categorized all the corresponding hashes. Otherwise with so many minor params lying around, its going to get more and more confusing in the long run.

fzdy1914:
> > @eugenepeh , If you are OK with the overhead caused, I can continue with the reviewing process.
> 
> I am ok to have all the minor params to be added to the hash (even though there isn't a need to), but we should minimally shorten the params name, and also use a prefix (e.g. all zoom tabs hashes should begins with z e.g `zoomSince -> zs`) to categorized all the corresponding hashes. Otherwise with so many minor params lying around, its going to get more and more confusing in the long run.

The `overhead` I mean is that, our repo needs to be filtered twice to produce the extra setting for v_zoom, which will be a lot of time when there are many repos.

frontend/src/static/js/v_summary.js
```diff
@@ -792,12 +808,51 @@ window.vSummary = {
...
     },
+
+    restoreZoomFiltered(info) {
+      const {
+        repoIndex, userIndex, filterGroupSelection, isMergeGroup, filterTimeFrame, zoomSince,
+        zoomUntil, sortingOption, sortingWithinOption, isSortingDsc, isSortingWithinDsc,
+      } = info;
+      let filtered = [];
+
+      const groups = JSON.parse(JSON.stringify(this.repos));
```
eugenepeh: >The overhead I mean is that, our repo needs to be filtered twice to produce the extra setting for v_zoom, which will be a lot of time when there are many repos.

I see, do you mean this part? Then yes, its not quite okay because of the additional parsing.
Also, the merge group should be more targeted and not perform on all the repos.

In short, would prefer as little processing as possible to be performed. In fact, can even consider storing indexes of the repos in merged group instead so that no filtering is needed.

jamessspanggg: @fzdy1914 @eugenepeh 

I have modified such that only the repos that are involved in the zoom tab will be processed. 

The zoom tab hashes and params are also shortened.

fzdy1914:
I may need some time to review this, but I will do it asap.

jamessspanggg:
> Can we put all zoom related properties in one data structure so we do not need to specify all the property in the function declaration?

Have done so for a method `matchZoomUser`. 

Should I do it for `sortFiltered`, `groupByRepos`, `groupByAuthors` and `groupByNone` as well? As they are also called with/without the zoom related properties.

frontend/src/static/js/main.js
```diff
@@ -225,6 +245,30 @@ window.app = new window.Vue({
       }
     },
 
+    renderZoomTabHash() {
+      const hash = window.hashParams;
+      const info = {
```
fzdy1914: `zoomInfo` instead?

frontend/src/static/js/v_summary.js
```diff
@@ -433,14 +433,23 @@ window.vSummary = {
       });
       this.filtered = full;
 
-      this.sortFiltered();
+      this.getOptionWithOrder();
+
+      const filterInfo = {
```
fzdy1914: `filterControl` instead?

frontend/src/static/js/v_summary.js
```diff
@@ -610,19 +616,27 @@ window.vSummary = {
...
+    sortFiltered(filtered, filterInfo) {
+      const {
+        filterGroupSelection, sortingOption, sortingWithinOption, isSortingDsc, isSortingWithinDsc,
+      } = filterInfo;
       this.getOptionWithOrder();
       let full = [];
-      if (this.filterGroupSelection === 'groupByNone') {
+
+      const sortingInfo = {
+        sortingOption, sortingWithinOption, isSortingDsc, isSortingWithinDsc,
```
fzdy1914: See if it can combine with line 611?

frontend/src/static/js/v_summary.js
```diff
@@ -690,59 +704,74 @@ window.vSummary = {
...
 
       this.$emit('view-zoom', {
-        filterGroupSelection: this.filterGroupSelection,
-        avgCommitSize,
-        user,
-        location: this.getRepoLink(repo[index]),
-        sinceDate: since,
-        untilDate: until,
-        isMergeGroup: this.isMergeGroup,
+        zRepo: user.repoName,
```
fzdy1914: Can this function also be using a data structure? It's fine if not using one.

jamessspanggg: You mean encompass all the properties into a zoomInfo object and only pass the zoomInfo object? Can ah

frontend/src/static/js/v_summary.js
jamessspanggg: Line 621 is it

frontend/src/static/js/v_summary.js
jamessspanggg: Don't think its necessary to do this tho. There's only 3 parameters for this function.

fzdy1914: My bad, not looking clearly.

frontend/src/static/js/main.js
```diff
@@ -189,17 +189,37 @@ window.app = new window.Vue({
...
       window.encodeHash();
     },
 
+    removeAuthorshipHashes() {
+      window.removeHash('tabAuthor');
+      window.removeHash('tabRepo');
+    },
+
+    removeZoomHashes() {
+      window.removeHash('zS');
```
eugenepeh: After this PR, can you put up an issue to combine all the zoom hashes into a single hashes using base64 encoding?
This is to centralize the use of aliases, to limit the appearance of e.g. `zU` in our code

jamessspanggg: Yes will do.

jamessspanggg:
Putting this on hold until we reach a consensus for #1145

jamessspanggg:
Putting on hold as there are some regressions.

frontend/src/static/js/v_zoom.js
```diff
@@ -51,18 +51,25 @@ window.vZoom = {
   watch: {
     info() {
       this.updateExpandedCommitMessagesCount();
+      this.setInfoHash();
```
jamessspanggg: Added this as previously, opening new zoom tabs will not update the existing zoom hashes

jamessspanggg:
Ready for review once again, solved the regressions

jamessspanggg:
@eugenepeh @fzdy1914 may i get ur reviews here once again? This is because there were regressions due to recent PRs merged.

0blivious:
Do you intend to keep the sorting criteria in zoom after reloading? Now every time I reload the page the sorting goes back to `time des`. It seems in #910 we are trying to restore the authorship tab control.

jamessspanggg:
> Do you intend to keep the sorting criteria in zoom after reloading? Now every time I reload the page the sorting goes back to `time des`. It seems in #910 we are trying to restore the authorship tab control.

nope only the ramps and their `granularity`. That is being managed in #1057

jamessspanggg:
@FD

jamessspanggg:
@

jamessspanggg:
@fzdy1914

jamessspanggg:
@fzdy1914 is this ready to merge?

fzdy1914:
I will merge this now. And please keeping eye on the potential regression。

jamessspanggg:
> I will merge this now. And please keeping eye on the potential regression。

Will do.

PR #837, opened 2019-07-25
[#703] Log: show analysis progress

### LINKED ISSUE #703: Show the progress of analysis in command line
It will be good to show user the queue number of the repo in current analysis. 

E.g. [2/52] analysing ...

bluein-green:
Resolves #703.

```
The progress in analyzing the repos is not shown.

Users who are waiting for the analysis to complete may want
to know how many repos RepoSense has analyzed so far.

Let's show the queue number of the repo that is currently
being analyzed. Repos that failed to clone or checkout will
not cause the queue number to increase.
```


src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -59,10 +59,13 @@
             "No authors found with commits for %s (%s).";
     private static final String MESSAGE_START_ANALYSIS = "Analyzing %s (%s)...";
     private static final String MESSAGE_COMPLETE_ANALYSIS = "Analysis of %s (%s) completed!";
+    private static final String FRAGMENT_NUM_REPOS_ANALYZED = "[%d/%d]";
     private static final String MESSAGE_REPORT_GENERATED = "The report is generated at %s";
     private static final String MESSAGE_BRANCH_DOES_NOT_EXIST = "Branch %s does not exist in %s! Analysis terminated.";
 
     private static Date earliestSinceDate = null;
+    private static int totalNumReposToAnalyze = 0;
+    private static int numReposAnalyzed = 0;
```
yong24s: Any reason why `static` variables are preferred here?

As the same can be done using in `for (RepoConfiguration config : configs)` of `ReportGenerator#analyzeRepos`.

bluein-green: `ReportGenerator`'s methods in general are used from a static context.

`ReportGenerator#analyzeRepos` only analyses the configs that have the same location. Hence, when we want to analyse configs from multiple locations, `analyzeRepos` will be invoked multiple times and a class-wide variable to keep track of the total number of "repos" (configs) analyzed, `numReposAnalyzed`, is needed.

As for `totalNumReposToAnalyze`, this value is determined from the list of `RepoConfiguration`s, which is found only in `ReportGenerator#generateReposReport`. To access this value in `analyzeRepos`, I thought it is necessary to store it as a class-wide variable.

yong24s: Not sure if its important, but can the implementation play well with tests?

https://travis-ci.org/reposense/RepoSense/jobs/563838015#L1215

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -131,6 +135,7 @@ private static void cloneAndAnalyzeRepos(
                 analyzeRepos(outputPath, repoLocationMap.get(clonedRepoLocation),
                         repoCloner.getCurrentRepoDefaultBranch());
             }
+
```
jylee-git: Unnecessary blank line here

src/main/java/reposense/report/ReportGenerator.java
jylee-git: You should set this to 0 in `generateReposReport` since you have declared this as static variable

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -62,10 +62,13 @@
             "No authors found with commits for %s (%s).";
     private static final String MESSAGE_START_ANALYSIS = "Analyzing %s (%s)...";
     private static final String MESSAGE_COMPLETE_ANALYSIS = "Analysis of %s (%s) completed!";
+    private static final String FRAGMENT_NUM_REPOS_ANALYZED = "[%d/%d]";
     private static final String MESSAGE_REPORT_GENERATED = "The report is generated at %s";
     private static final String MESSAGE_BRANCH_DOES_NOT_EXIST = "Branch %s does not exist in %s! Analysis terminated.";
 
     private static Date earliestSinceDate = null;
+    private static int totalNumReposToAnalyze = 0;
+    private static int numReposAnalyzed = 0;
```
eugenepeh: Don't really like this solution because the values can be easily modified and creates clutters around ReportGenerator.
Can we create a class for this, that is responsible for keeping the count as well as produce the progress message?
Can call it `ProgressTracker` or any better names that you can come up with.

PR #840, opened 2019-07-26
[#839] Set config to be inclusive of until date

### LINKED ISSUE #839: Until date in config does not match with until date in git rev-list
The current until date used in config excludes the until date while the until date used in git rev-list includes the until date.

Screen Shot 2019-07-25 at 11 07 04 PM

jamessspanggg:
Fixes #839 

```
The until date in the config does not match with the until date used in
git rev-list. Until date in the config is in 00:00:00 time whereas
until date used in git rev-list uses the time 23:59:59 of the day
before.

This discrepancy can cause the content and author of each line detected
by git rev-list to be different, and have caused a disparity in the
output results in #789.

Let's standardise the until date in the config to use time of 23:59:59 
to match with the until date used in git rev-list.

Reference:
https://github.com/reposense/RepoSense/pull/789#discussion_r307350220
```


src/main/java/reposense/parser/DateArgumentType.java
```diff
@@ -15,12 +15,12 @@
  */
 public class DateArgumentType implements ArgumentType<Optional<Date>> {
     private static final String PARSE_EXCEPTION_MESSAGE_INVALID_DATE_STRING_FORMAT = "Invalid Date: %s";
-    private static final DateFormat CLI_ARGS_DATE_FORMAT = new SimpleDateFormat("dd/MM/yyyy");
+    private static final DateFormat CLI_ARGS_DATE_FORMAT = new SimpleDateFormat("dd/MM/yyyy HH:mm:ss");
```
yong24s: May need to enforce user only enters `dd/MM/yyyy` and not `dd/MM/yyyy+`. 

Try the following:
```
java -jar RepoSense.jar --since "10/10/2018 10:10:10"
java -jar RepoSense.jar --since "10/10/2018 10:"
```

It seems to get parsed properly 🤔.

jamessspanggg: Hmm, this problem exists even before this PR. Should I fix it in this PR or open an issue for this?

jamessspanggg: I have read online stating that `setLenient(false)` does not achieve strict date formatting. Hence examples such as `--since 1/1/201` will still be parsed successfully.

jamessspanggg: Would implementing a regex to check if the argument is exactly `dd/MM/yyyy` format be an ideal solution?

yong24s: > I have read online stating that setLenient(false) does not achieve strict date formatting.

I think the dates like `1/1/201` are okay.

However, my concern is the behavior of `--since "10/10/18 11:11:11 22:22:22 ... 00:00:00"` and `--since "10/10/10 11/11/11 12/12/12"` as they are considered valid 😂.

I think we should do enforce `dd/MM/yyyy` and fast fail.
@eugenepeh your opinion?

eugenepeh: >I think the dates like 1/1/201 are okay.

Its okay with me, so no need to open an issue for it.

Why the need to include time in cli argument though?

jamessspanggg: The time is not included in the cli argument, but the cli argument `sinceDate` will be appended with time `00:00:00` while `untilDate` will be appended with `23:59:59` so that they are parsed into Date objects such that the `untilDate` is included in the analysis.

eugenepeh: I see, sounds good to me. However, just to be wary of users' accidental input, can you add a few tests with time to ensure that it is stable?

jamessspanggg: You mean tests for accidental input with time such as `--since 11/11/2019 00:00:00`?

yong24s: @jamessspanggg 
Ya something to test our assumption. Take a look at `ArgsParserTest`.

`--since "10/10/18 11:11:11 22:22:22"` + `00:00:00` = `10/10/18 00:00:00`
`--since "10/10/18 11:11:11"` + `00:00:00` = `10/10/18 00:00:00`
`--since "10/10/18 11:"` + `00:00:00` = `10/10/18 00:00:00`
`--since "10/10/18 11"` + `00:00:00` = `10/10/18 00:00:00`

eugenepeh: >You mean tests for accidental input with time such as --since 11/11/2019 00:00:00?

Yes, maybe something like what @yong24s has illustrated in the first comment. 
E.g. `--since "10/10/2018 10:"`

jamessspanggg: > E.g. `--since "10/10/2018 10:"`

Currently arguments
- `--since "10/10/2018 10:` = `10/10/2018 10:00:00`
- `--since "10/10/2018 11:11:11 12:12:12` = `10/10/2018 11:11:11`

Should these arguments be considered as invalid? or should we only get the date from the argument and append the time accordingly (eg. `00:00:00` for sinceDate, `23:59:59` for untilDate)

jylee-git: Using --since with invalid dates without the double quotes should fail (and intended).

Using --since with the double quotes is not an intended option (double quotes isn't highlighted in the UG). So was wondering if this should still be a concern

jamessspanggg: I have updated the parsing of `sinceDate` and `untilDate` such that only the first substring that matches the date regex of `dd/mm/yyyy` will be used.

For e.g. 
- `--since "10/10/2018 10:"` will be parsed with `sinceDate` of 10/10/2018, and will be appended with time of `00:00:00`
- `--until "10/10/2018 11/11/11 12/12/12"` will be parsed with `untilDate` of 10/10/2018, and will be appended with time of `23:59:59`

eugenepeh: 👍 do add a unit test too.

jamessspanggg: Added unit tests.

src/main/java/reposense/parser/SinceDateArgumentType.java
```diff
@@ -25,8 +27,12 @@
...
     @Override
     public Optional<Date> convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
-        return (FIRST_COMMIT_DATE_SHORTHAND.equals(value))
-                ? Optional.of(ARBITRARY_FIRST_COMMIT_DATE)
-                : super.convert(parser, arg, value);
+        if (FIRST_COMMIT_DATE_SHORTHAND.equals(value)) {
+            return Optional.of(ARBITRARY_FIRST_COMMIT_DATE);
+        } else {
+            Matcher matcher = Pattern.compile(DATE_FORMAT_REGEX).matcher(value);
+            String sinceDate = matcher.find() ? matcher.group(1) : value;
```
yong24s: Can move these two lines to parent class to avoid duplication.

Also `Match#group` may throw `IndexOutOfBoundsException`.

Finally, can remove the else block. See [happy path](https://nus-cs2103-ay1819s1.github.io/cs2103-website/se-book-adapted/chapters/codeQuality.html#make-the-happy-path-prominent).

src/main/java/reposense/parser/UntilDateArgumentType.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Verifies and parses a string-formatted until date to a {@code Date} object.
+ */
+
+public class UntilDateArgumentType extends DateArgumentType {
+    @Override
+    public Optional<Date> convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
+        Matcher matcher = Pattern.compile(DATE_FORMAT_REGEX).matcher(value);
+        String untilDate = matcher.find() ? matcher.group(1) : value;
```
yong24s: Related to the above comment.

src/main/java/reposense/parser/SinceDateArgumentType.java
jamessspanggg: Might not be able to move the two lines to parent class as the matching of the first substring that matches the date format must happen before appending the time (`00:00:00` for since dates and `23:59:59` for until dates)

yong24s: Don't have be in `DateArgumentType#convert`, just make a new protected method?

jamessspanggg: Oh ya, okay got it thanks so much.

jamessspanggg: @yong24s 
Do we need to try catch the `IndexOutOfBoundsException` from `matcher.group`? Since we already have a `matcher.find()` before `matcher.group(1)`, which ensures that the matcher will have at least a match.

src/main/java/reposense/parser/DateArgumentType.java
yong24s: @jylee-git 
> So was wondering if this should still be a concern

Valid point.

We are also considering the possibility of typos that may increase the time of analysis. i.e. `1/1/2019` -> `1/1/201`.

Usually, this will be a low priority but since its a easy fix for added robustness 🤭.

src/main/java/reposense/parser/SinceDateArgumentType.java
```diff
@@ -25,8 +25,11 @@
...
     public Optional<Date> convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
-        return (FIRST_COMMIT_DATE_SHORTHAND.equals(value))
-                ? Optional.of(ARBITRARY_FIRST_COMMIT_DATE)
-                : super.convert(parser, arg, value);
+        if (FIRST_COMMIT_DATE_SHORTHAND.equals(value)) {
+            return Optional.of(ARBITRARY_FIRST_COMMIT_DATE);
+        } else {
+            String sinceDate = extractDate(value);
+            return super.convert(parser, arg, sinceDate + " 00:00:00");
+        }
```
yong24s:
```
if (FIRST_COMMIT_DATE_SHORTHAND.equals(value)) {
    return Optional.of(ARBITRARY_FIRST_COMMIT_DATE);
}

String sinceDate = extractDate(value);
return super.convert(parser, arg, sinceDate + " 00:00:00");
```

eugenepeh:
>The current until date in the config does not match with the until date 
used in git rev-list. Until date in the config is in 00:00:00 time
whereas until date used in git rev-list is in 23:59:59 time. The until
date is previously decided to be included in the analysis.

This line is not necessary.
~~The until date is previously decided to be included in the analysis.~~

Also missing a problem statement, why this constitute to a problem?

jamessspanggg:
> This line is not necessary.
> ~The until date is previously decided to be included in the analysis.~
> 
> Also missing a problem statement, why this constitute to a problem?

Updated the commit message.

eugenepeh:
Nits to proposed message:

```
The until date in the config does not match with the until date used in
git rev-list. Until date in the config is in 00:00:00 time whereas
until date used in git rev-list uses the time 23:59:59 of the day
before.

This discrepancy can cause the content and author of each line detected
by git rev-list to be different, and have caused a disparity in the
output results in #789.

Let's standardise the until date in the config to use time of 23:59:59 
to match with the until date used in git rev-list.

Reference:
https://github.com/reposense/RepoSense/pull/789#discussion_r307350220
```

jamessspanggg:
> Nits to proposed message:

Updated.

Issue #798, opened 2019-07-09
Repeated commits detected in report


jamessspanggg:
To replicate, run `--repo https://github.com/reposense/RepoSense.git -v -s d1 -i`

<img width="1643" alt="Screen Shot 2019-07-09 at 4 22 28 PM" src="https://user-images.githubusercontent.com/32864116/60871943-6072a200-a266-11e9-8fc8-47383fa6fb1d.png">

PR #841, opened 2019-07-26
[#798] Fix duplicate commit results for author using multiple aliases

jylee-git:
```
When analyzing repos without a specified author configuration, authors 
with multiple aliases will result in duplicate commits being recorded 
and displayed in the report view.

Let's remove the duplicate commit messages from being displayed in the 
zoom panel.

This is performed by changing the data type of commitResults for each 
author from List to LinedHashSet to remove the duplicates.
```


damithc:
Use imperative mood in commit naming `Changed` -> `Change`

jamessspanggg:
For the commit message, can you explain more on how you are removing the duplicate commit messages? e.g. using LinkedHashSet() instead of List for `commitResults`

fzdy1914:
> When analyzing repos without a specified author configuration, authors 
> with multiple aliases will result in duplicate commits being recorded 
> and displayed in the report view.

Can you explain why this will happens? One commit can only have one author right?

jylee-git:
Closing this PR as this isn't the ideal solution, and hence may be more complicated and needs further investigation.

PR #843, opened 2019-07-29
[#319] Add unit test for minimal standalone config

### LINKED ISSUE #319: Add a testrepo with minimal standalone config
This is to test all the default values.

bluein-green:
Resolves #319.

```
We do not have a repo that contains a standalone
config with only the mandatory fields filled in.

Having such a repo and a unit test is needed to check
whether the resulting `RepoConfiguration` based on the
minimal standalone config has its fields assigned
default values.

Let's add a repo containing a minimal standalone config.
```


src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -299,4 +301,27 @@ public void repoConfig_overrideStandaloneConfig_success()
...
     }
+
+    @Test
+    public void repoConfig_minimalStandaloneConfig_fieldsAssignedDefaultValues() throws GitCloneException,
+            InvalidLocationException {
+        RepoConfiguration expectedConfig = new RepoConfiguration(new RepoLocation(TEST_REPO_MINIMAL_STANDALONE_CONFIG),
+                "master");
+
+        Author firstAuthor = new Author("bluein-green");
+        Author secondAuthor = new Author("jylee-git");
```
eugenepeh: can you test the author email, author name and display name as well?

eugenepeh: and its ignore glob list too.

bluein-green: What do you mean?

eugenepeh: Hmm, actually I wanted to mention its because when we use standalone configs, the author are not important as `Author`, but rather `StandaloneAuthor`, so we should verify its value.

But on second thought, I think its fine just to simply check that it is equal to the initialized `Author`. 

Can ignore my above comments.

PR #844, opened 2019-07-30
[#842] Fix out of bound ramp slices

### LINKED ISSUE #842: Chart view: ramp appears outside chart when granularity is week
When the `since` and `until` dates are set such that they are quite close (say 3 days apart or less - it varies for different ramp slices), and when the `granularity` is set to week, the charts will have horizontal scrollbars and the ramp will appear at the very right.

weekGranularity

jamessspanggg:
Fixes #842 

```
Ramp slices appear outside of the ramp charts in week granularity.

This is due to the implementation of splitCommitsWeek() does not take
into account when the duration between since date and until date is
less than one week. When the duration is less than a week, the
algorithm calculates commits up to the following Monday, even if they
are beyond the until date. Hence, it causes a distortion in the ui.

Let's revamp the logic within splitCommitsWeek to check, and trim any
commits beyond the until date.
```

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -423,7 +423,7 @@ window.vSummary = {
       if (nextMondayDate <= untilDate) {
         this.pushCommitsWeek(sinceMs, nextMondayMs - 1, res, commits);
         this.pushCommitsWeek(nextMondayMs, untilMs, res, commits);
-      } else { // duration between since date and until date is less than one week
+      } else {
         this.pushCommitsWeek(sinceMs, untilMs, res, commits);
       }
       user.commits = res;
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -417,8 +417,8 @@ window.vSummary = {
       const untilDate = this.filterUntilDate;
 
       const nextMondayMs = (new Date(nextMondayDate)).getTime();
-      const untilMs = (new Date(untilDate)).getTime();
       const sinceMs = new Date(this.filterSinceDate).getTime();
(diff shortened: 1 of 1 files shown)


jamessspanggg:
@bluein-green can u help me check if you can reproduce the bug now?

frontend/src/static/js/v_summary.js
```diff
@@ -413,21 +413,19 @@ window.vSummary = {
...
       const res = [];
 
-      const sinceDate = dateRounding(this.filterSinceDate, 0); // round up for the next monday
+      const nextMondayDate = dateRounding(this.filterSinceDate, 0); // round up for the next monday
       const untilDate = this.filterUntilDate;
 
-      const sinceMs = (new Date(sinceDate)).getTime();
+      const nextMondayMs = (new Date(nextMondayDate)).getTime();
       const untilMs = (new Date(untilDate)).getTime();
+      const sinceMs = new Date(this.filterSinceDate).getTime();
```
yong24s: Move `sinceMs` before `untilMs` to minimize changes.

jamessspanggg: Made the changes.

eugenepeh:
Nits to proposed message:

```
Ramp slices appear outside of the ramp charts in week granularity.

This is due to the implementation of splitCommitsWeek() does not take
into account when the duration between since date and until date is
less than one week. When the duration is less than a week, the
algorithm calculates commits up to the following Monday, even if they
are beyond the until date. Hence, it causes a distortion in the ui.

Let's revamp the logic within splitCommitsWeek to check, and trim any
commits beyond the until date.
```

jamessspanggg:
> Nits to proposed message:

Updated.

PR #845, opened 2019-07-30
[#382] Authorship: add tagged author that did not commit into list of authors

### LINKED ISSUE #382: Annotation tagged author may not appear in report
By default, if no configurations were found, RepoSense will perform the analysis for all the authors found in commit history.

However if the author named in the annotation tag, `@@author`, is not found in the commit history, her contribution won't show up in the report.

This can be a very rare case, hence putting this is up for discussion on whether should we fix it in the near future.

bluein-green:
Resolves #382.

```
If no configurations are found, RepoSense will use all authors found
in the commit history by default. However, if an author named in the
annotation tag `@@author` is not found in the commit history, that
person's contribution will not show up in the report.

Let's add such an author into the list of authors, and attribute the
lines to said author instead of to `Author.UNKNOWN_AUTHOR`.
```


jamessspanggg:
Closing this due to inactivity.

PR #847, opened 2019-07-31
[#846] Fix bug of show ramps for this period button only works once

### LINKED ISSUE #846: Show ramp chart for this period button in code view only works once
* select a sub range to display in code view
* press the `Show ramp chart for this period` 
* modify the date range in summary view
* press `Show ramp chart for this period` again -> doesn't work

Here's the url I have when encountering this error
https://deploy-preview-844--reposense.netlify.com/#tabOpen=true&search=&sort=groupTitle&sortWithin=title&since=2017-10-09&until=2019-07-30&timeframe=commit&groupSelect=groupByRepos&breakdown=false&tabType=zoom

jamessspanggg:
Fixes #846 

```
Clicking on show ramp chart for this period button in the commits
panel calls updateDateRange(), which recomputes the duration of the
ramp chart according to the drags by the user. The method proceeds 
to call deactivateAllOverlays(), which removes the drags by the user.

Since no more drags are left after the first execution of 
updateDateRange(), clicking on the show ramp chart for this period
button will not be able to activate updateDateRange().

Let's use the zoom since date and until date to recompute the duration
of the ramp chart, instead of using the drags. This will ensure that
clicking on the button will always show the ramp chart for the 
designated period.
```

### DIFF
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -525,19 +525,10 @@ window.vSummary = {
       this.tmpFilterUntilDate = this.maxDate;
     },
 
-    updateDateRange() {
-      if (drags.length > 0) {
-        const since = new Date(this.filterSinceDate).getTime();
-        const until = new Date(this.filterUntilDate).getTime();
-        const range = until - since;
-
-        const getStr = (time) => getDateStr(new Date(time));
-        this.tmpFilterSinceDate = getStr(since + range * drags[0] / 100);
-        this.tmpFilterUntilDate = getStr(since + range * drags[1] / 100);
-
-        drags = [];
-        deactivateAllOverlays();
-      }
+    updateDateRange(zoomSince, zoomUntil) {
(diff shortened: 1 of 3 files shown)


frontend/src/static/js/v_summary.js
```diff
@@ -525,19 +525,10 @@ window.vSummary = {
...
-        const getStr = (time) => getDateStr(new Date(time));
-        this.tmpFilterSinceDate = getStr(since + range * drags[0] / 100);
-        this.tmpFilterUntilDate = getStr(since + range * drags[1] / 100);
-
-        drags = [];
-        deactivateAllOverlays();
-      }
+    updateDateRange(zoomSince, zoomUntil) {
+      this.tmpFilterSinceDate = zoomSince;
+      this.tmpFilterUntilDate = zoomUntil;
```
bluein-green: Not sure if it is better to rename `zoomSince` to `since` and `zoomUntil` to `until` for generality

PR #848, opened 2019-07-31
Modify UI of zoom tab title

jamessspanggg:
```
The current zoom tab title displays in the same format when charts
are group by authors, repos or none. The period of the zoom tab
is also not shown in the zoom tab.

Let's modify the UI of the zoom tab title with the following sequence:
group title -> chart title -> period of zoom tab -> granularity.
Furthermore, the chart title will not be shown if the merge group
option is checked, this is because a merged group may consists of 
more than one chart.

Let's also modify the authorship's title format to match with the zoom
tab's title format as well.
```


jamessspanggg:
The updated zoom tab UI according to @damithc's requirements:

When `group by none or repos`:
Screen Shot 2019-07-31 at 12 37 51 PM

When `group by authors`:
Screen Shot 2019-07-31 at 12 38 04 PM

frontend/src/tabs/zoom.pug
```diff
@@ -1,10 +1,17 @@
...
+    .zoom_title--group
+      h1(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.displayName }} ({{ info.user.name }})
+      h1(v-else) {{ info.user.repoName }}
+    .zoom__title--chart
+      span &#8627; &nbsp;
+      h3(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.repoName }}
+      h3(v-else) {{ info.user.displayName }} ({{ info.user.name }})
+    .zoom__title--period
+      span &#8627; &nbsp;
+      span {{ info.sinceDate }} ~ {{ info.untilDate }} &nbsp;
```
bluein-green: Would it better to use "to" instead of '~'?

frontend/src/tabs/zoom.pug
```diff
@@ -1,10 +1,17 @@
...
+      a(v-on:click="openSummary") [Show ramp chart for this period]
+    .zoom__title--granularity.mini-font *granularity: one ramp per {{ filterTimeFrame }}
```
bluein-green: I think there isn't a need for the '*' if the note is in its own line now

frontend/src/tabs/zoom.pug
```diff
@@ -1,10 +1,17 @@
 #zoom
-  h1 {{ info.user.displayName }} ({{ info.user.name }})
-  a(v-on:click="openSummary") Show ramp chart for this period
-
   .zoom__title
-    .zoom__title--repo {{ info.user.repoName }} &nbsp;
-    .zoom__title--granularity.mini-font (granularity: one ramp per {{ filterTimeFrame }})
+    .zoom_title--group
+      h1(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.displayName }} ({{ info.user.name }})
+      h1(v-else) {{ info.user.repoName }}
```
bluein-green: The font size for the header doesn't seem to be the same as that in the authorship view. I think it would be good if they both had the same font size to give a more uniform look.

jamessspanggg:
Your inputs on the UI? @damithc

damithc:
> Your input on the UI? @damithc

Looks fine to me 👍

frontend/src/tabs/zoom.pug
jylee-git: Same thoughts here

frontend/src/tabs/zoom.pug
```diff
@@ -1,10 +1,17 @@
...
+    .zoom__title--chart
+      span &#8627; &nbsp;
+      h3(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.repoName }}
+      h3(v-else) {{ info.user.displayName }} ({{ info.user.name }})
+    .zoom__title--period
+      span &#8627; &nbsp;
+      span {{ info.sinceDate }} ~ {{ info.untilDate }} &nbsp;
+      a(v-on:click="openSummary") [Show ramp chart for this period]
```
yong24s: Is it possible to make changes to the sinceDate and untilDate instead?

Reason: The dotted selection get reset or not in the correct range of dates.

jamessspanggg: Sorry but I don't really get what you mean. Do u mean making changes to the sinceDate and untilDate when clicking the `show ramp chart for this period` button?

yong24s: Yes. As using the dotted selection is not reliable.

jamessspanggg: Oh yes that issue is being solved in #847. `sinceDate` and `untilDate` of zoom will be used instead of the dotted selections.

frontend/src/tabs/zoom.pug
eugenepeh: Yes, it would be better for the authorship and zoom tab to be the same, that way a common tab panel style sheet can be used.

frontend/src/tabs/zoom.pug
eugenepeh: Yes, if the contribution bar's hover is using `to` to bridge the range, lets stick to the same style.

frontend/src/tabs/zoom.pug
eugenepeh: I am fine with the * but if we going to highlight it in the style like a footnote, it would probably be better to place it under the ramp

frontend/src/tabs/zoom.pug
eugenepeh: @jamessspanggg u have missed out the comment here. 
Not necessary to invent a common tab panel style sheet yet but do try to make the headers similar to each other.

jamessspanggg: Sorry for missing out this. I have standardised the fonts used and created two global fonts `large-font` and `medium-font`.

eugenepeh:
Perhaps the display of author's name and date range should also be standardized.
@damithc what do you think? which is better?

Zoom Tab
image
Authorship Tab
image

damithc:
> Perhaps the display of author's name and date range should also be standardized.
> @damithc what do you think? which is better?

Yes, I think it should be standardized.
I like the first one because it matches how the data is narrowed down from the group to the duration.

jamessspanggg:
Okay. Shall I modify the authorship to match with the zoom tab as well?

eugenepeh:
>Okay. Shall I modify the authorship to match with the zoom tab as well?

Yes please, thanks @jamessspanggg

jamessspanggg:
Updated authorship tab to match with zoom tab.

eugenepeh:
Nits to the proposed message:
```

The zoom tab header remains the same irregardless of the grouping
option (authors, repos or none). The period of the zoom tab is also not
shown in the zoom tab.

This creates a problem when the merge group option is used; a merged
group may consists of more than one chart, and the title used in the
header might not apply to all the ramps within the merged chart.

Let's modify the UI of the zoom tab title to use a new sequence:
`group title` -> `chart title` -> `period of zoom tab` ->
`granularity`, and hide the chart title when the merge group option is
checked.

Let's also modify the authorship's title format to match with the zoom
tab's title format as well.
```

PR #853, opened 2019-08-02
[#851] Disable ramp url if merged group and group by authors is selected

### LINKED ISSUE #851: Zoom tab url directing to wrong repository when merged group
image

image

jamessspanggg:
Fixes #851 

```
When the merge group option is checked and charts are grouped by
authors, the url of some ramps may point to the wrong repository.

This is due to our current structure of the merged group, where each
merged group stores only one repository name, which then causes 
all url of ramps to point to the same repository.

Since changing the structure of the merged group that allows each ramp
to store their respective repositories may take quite an effort, let's 
disable the url of the ramps when merge group option is checked and 
charts are grouped by authors.
```

### DIFF
--- a/frontend/src/ramp.pug
+++ b/frontend/src/ramp.pug
@@ -1,4 +1,4 @@
-.ramp
+.ramp(v-bind:class="{ 'merged-author' : mergegroup && groupby === 'groupByAuthors' }")
   template(v-if="tframe === 'commit'")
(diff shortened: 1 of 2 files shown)


jamessspanggg:
Putting this on hold to reconsider my solution.

jamessspanggg:
Closing as found a better solution in PR #853

PR #857, opened 2019-08-14
[#185] Use realistic person names for test repos

Tejas2805:


jylee-git:
https://github.blog/2019-02-14-introducing-draft-pull-requests/

When making a PR next time, you may want to draft pull request if it isn't ready for review at the time of making the PR.

fzdy1914:
Regarding this issue,  we have a discussion before. We believe it is not good to use the student's real name in this case because they are not a part of our developer team.

jylee-git:
> Regarding this issue, we have a discussion before. We believe it is not good to use the student's real name in this case because they are not a part of our developer team.

We should use realistic, but dummy names instead of the real names of those students.

Tejas2805:
> > Regarding this issue, we have a discussion before. We believe it is not good to use the student's real name in this case because they are not a part of our developer team.
> 
> We should use realistic, but dummy names instead of the real names of those students.

Alright I will make the changes accordingly. 

(I thought since using proper repositories, names need to e proper too)

Tejas2805:
Can anyone guide me as to how to fix this error in Travis? I am not being able to understand how to remove /r/n and just use /n?

PR #874, opened 2019-08-24
[#858] Restore authorship tab

### LINKED ISSUE #858: Authorship tab not restorable after PR #757
**Current:** 
Authorship tab is currently not restorable after refreshing the page.

**Reason:** 
Restoring authorship tab requires the since date and until date hashes whereas PR #757 has made changes where the date hashes will not be captured in the url if not specified by the user.

jamessspanggg:
Fixes #858 

```
Authorship tab is not restorable after refreshing the page.

Restoring authorship tab requires the since and until date hashes 
whereas PR #757 has made changes where the date hashes will not 
be captured in the url if not specified by the user.

Let's 
- use the window's since and until date to restore the authorship tab 
  if the date hashes are unavailable. 
- move the restoring of tab hashes to be executed after loadSummary() 
  to ensure that the window's since and until date is available.
- extract the deactivation of tab pane to a method instead of using the
  updated hook as the updated hook will be called repeatedly whenever
  there are data changes that affect the DOM.
```


frontend/src/static/js/main.js
```diff
@@ -126,15 +126,15 @@ window.app = new window.Vue({
...
-    updateReportDir() {
+    initialize() {
       window.REPORT_ZIP = null;
 
       this.users = [];
-      this.updateReportView();
+      this.updateReportViewAndRenderTabHash();
     },
-    updateReportView() {
+    updateReportViewAndRenderTabHash() {
```
fzdy1914: It seems that it is not necessary to combine `renderTabHash` with `updateReportView`? I think it may be better to separate it.

jamessspanggg: The reason why I combined these two methods is because `renderTabHash` must be executed after the async method `updateReportView`.

fzdy1914: Can it be:
```
.then(() => this.updateReportView())
.then(() => this.renderTabHash());
```
? 
In the `window.api.loadSummary()`

fzdy1914: It should also work, right?

jamessspanggg: Yes this also works but then `this.initialize` will also call `updateReportView` and I'm unsure how to make sure that `renderTabHash` is executed after `updateReportView`. That is why I placed `renderTabHash` to be within `updateReportView`.

jamessspanggg: @fzdy1914 I have modified the code to separate `updateReportView` and `renderTabHash`. But I am unsure if this is the best way of doing this.

frontend/src/static/js/main.js
```diff
@@ -260,21 +280,13 @@ window.app = new window.Vue({
     CircleSpinner: window.VueLoadingSpinner.Circle,
   },
   created() {
-    this.updateReportDir();
-    this.renderTabHash();
+    this.initialize();
```
fzdy1914: Then, this change is not necessary right? There is no need to put them in the same method (initialize)? Why the original one does not work?

jamessspanggg: Removed the `initialise` method

frontend/src/static/js/main.js
```diff
@@ -223,7 +234,16 @@ window.app = new window.Vue({
 
       if (this.isTabActive) {
         if (hash.tabType === 'authorship') {
-          this.renderAuthorShipTabHash(hash.since, hash.until);
+          let { since, until } = hash;
+
+          // get since and until dates from window.app if not found in hash
+          if (!since) {
+            since = window.app.sinceDate;
```
eugenepeh: perhaps can shorten to something like
`since = hash.since ? hash.since : window.app.sinceDate;`?

jamessspanggg: Updated accordingly.

PR #875, opened 2019-08-24
[#185] Use realistic person names for test repos

### LINKED ISSUE #185: Use realistic person names for test repos

Tejas2805:
Fixes #185 

```
The names being used for the repositories are short.

Using this kind of too simple test cases can cause bugs to 
go unnoticed and also can cause the actual UI to look 
different from the UI we see in test cases.

Let's change the names to real use case names which 
we are more likely to see. We do this by adding long 
names, names with special characters and numbers, and a 
blank instead of a name.

This has been done this way to cover various equivalence
 partitions of the test input.
```

### DIFF
--- a/author-config.csv
+++ b/author-config.csv
@@ -1,9 +1,9 @@
 Repository's Location,Branch,Author's GitHub ID,Author's Email,Author's Display Name,Author's Git Author Name,Ignore Glob List
-https://github.com/reposense/testrepo-Beta.git,master,nbriannl,,Nbr,,
-https://github.com/reposense/testrepo-Beta.git,master,zacharytang,zacharytang.tc@gmail.com,Zac,,src/main**
-https://github.com/reposense/testrepo-Beta.git,master,April0616,,Fan,LAPTOP-7KFM2KSP\User;Fan Yuting,
-https://github.com/reposense/testrepo-Beta.git,master,CindyTsai1,,Cin,YuHsuan,src/test**
-https://github.com/reposense/testrepo-Beta.git,add-config-json,nbriannl,,Nbr,,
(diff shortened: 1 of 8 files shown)


jamessspanggg:
@Tejas2805 please maintain the commit message body at 72 characters per line.

damithc:
@Tejas2805 in the commit message, instead of saying `has been changed`, say `let's change`

Tejas2805:
Is any update required ?

jylee-git:
Looks good here

Tejas2805:
@eugenepeh thanks for your guidance. Will keep it in mind. Just one question. 

> we can do without this information

^ I thought I had to write why have the changes been done in the way they have been done and that's why I added it. So is it not necessary to have such a line ?

eugenepeh:
>^ I thought I had to write why have the changes been done in the way they have been done and that's why I added it. So is it not necessary to have such a line ?

The subsequent paragraphs 4th & below are generally meant to be the technical details. 
`{why it is done that way}` in terms of technical aspects, e.g. qualitative comparison between 2 solutions, pros and cons etc. 

It is generally not needed for clear cut solutions.

Tejas2805:
Okay got it. Thanks for your help.

PR #877, opened 2019-08-24
[#876] FileType: default file type to "other" if file format is not of standard convention

### LINKED ISSUE #876: RepoSense crashes for files with unorthodox naming convention
When:
* Format is not specified, 
* File name does not contain formats (`.java`, `.py`, ...), and
* File name contains special characters and spaces (`pull request`, `$gradlew_backup$`, ...)

RepoSense crashes with `The provided file type, <<some weird file name>>, contains illegal characters.`.

jylee-git:
```
The label of a FileType only allows alphanumeric characters.

When format is not specified, files without a standard file format
and with special characters (e.g.: "$gradle_backup$", "Backup File") 
will fail the FileType validation test and RepoSense terminates.

Let's:
* Allow FileType label to accept any non-empty string
* Enforce file format validation to be alphanumeric
* Label the FileType to "other" where file format cannot be deduced 
  for files with unorthodox naming convention
* Ensure that each files exists in the host computer before attempting 
  to analyze them.

A file format will be classified as "other" if the string of the 
format portion of the file name contains non-alphanumeric characters, 
e.g.: "main.j@va".
```

Resolves #876


jylee-git:
> Can we have a real file test for this fix? Like having a branch in test repo to test?

I feel that it's not necessary to do so as the file name will still be extracted from each individual file as a String to test.

On a side note, tt can warrant a system test, but it would require an overhaul of `ConfigSystemTest` as the system test is restrictive with only the flexibility of playing with date ranges. This will be out of the scope of this PR and will need a new PR to deal with this.

yong24s:
For files without extension,  `i.e.` `some file; whoami`, we should wrap the filename with double quotes when executing in `CommandRunner` and also verify its existence before executing.

jylee-git:
> For files without extension, `i.e.` `some file; whoami`, we should wrap the filename with double quotes when executing in `CommandRunner` and also verify its existence before executing.

`GitBlame` is the only one in `CommandRunner` that touches file names, but the existing code base already has double quotes wrapped when executing the command.

Will do something about asserting that the file exists.

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -15,6 +18,8 @@
     private List<FileType> formats;
     private List<FileType> groups;
 
+    private final Logger logger = LogsManager.getLogger(this.getClass());
```
yong24s: Interesting approach here, but lets follow the convention we are using, for slightly better performance:
`private static final Logger logger = LogsManager.getLogger(FileTypeManager.class);`

Using your approach, we can remove the need to declare `private final Logger logger...` in every class, if we make it an interface? 🤔

src/main/java/reposense/model/FileType.java
```diff
@@ -17,37 +17,53 @@
...
      * Ensures that the string {@code label} is a valid file type.
-     * @throws IllegalArgumentException if {@code label} does not meet the criteria.
+     * @throws IllegalArgumentException if {@code label} is an empty string.
      */
-    public static void validateFileType(String label) {
-        if (!label.matches(FILE_TYPE_VALIDATION_REGEX)) {
-            throw new IllegalArgumentException(String.format(MESSAGE_ILLEGAL_FILE_TYPE, label));
+    public static void validateFileTypeLabel(String label) {
+        if (label.isEmpty()) {
+            throw new IllegalArgumentException();
```
yong24s: Move the exception message here.

See next comment.

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -48,7 +53,14 @@ private FileType getFileFormat(String fileName) {
                     "This exception should not happen as we have performed the whitelisted formats check.");
         } else {
             String[] tok = fileName.split("[./\\\\]+");
-            return FileType.convertStringFormatToFileType(tok[tok.length - 1]);
+            String deducedFormat = tok[tok.length - 1];
+            try {
+                return new FileType(deducedFormat, Collections.singletonList("**" + deducedFormat));
+            } catch (IllegalArgumentException iae) {
+                logger.warning(String.format("Unable to determine the file format for the file \"%s\". "
+                        + "This file will be treated as the file type \"other\".", fileName));
```
yong24s: As commented above.

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -48,7 +53,14 @@ private FileType getFileFormat(String fileName) {
                     "This exception should not happen as we have performed the whitelisted formats check.");
         } else {
             String[] tok = fileName.split("[./\\\\]+");
-            return FileType.convertStringFormatToFileType(tok[tok.length - 1]);
+            String deducedFormat = tok[tok.length - 1];
```
yong24s: `deducedFormat` could be better described as `extension` or `label` or `format`.

src/main/java/reposense/model/FileTypeManager.java
jylee-git: Will change back to use the existing convention.

>Using your approach, we can remove the need to declare private final Logger logger... in every class, if we make it an interface? 

I'm not really sure if that's possible, but for now let's stick to the existing convention

src/main/java/reposense/model/FileType.java
jylee-git: Doing so will require the modification of `validateFileTypeLabel` and `FileType` constructor to also take in the full file name just to use the exception message to inform the user that we can't deduce the format **of this particular file**.

Unless if we don't inform the user in the first place and just set the fileType of the file as `other`?

fzdy1914:
Can I know the state of this PR?

jylee-git:
Need @yong24s 's input on the unresolved comments above

src/main/java/reposense/model/FileType.java
```diff
@@ -17,37 +17,53 @@
...
     public static FileType convertStringFormatToFileType(String format) {
+        validateFileFormat(format);
         return new FileType(format, Collections.singletonList("**" + format));
     }
 
+    /**
+     * Ensures that {@code format} is a valid file format.
+     * @throws IllegalArgumentException if {@code format} is not alphanumeric.
+     */
+    public static void validateFileFormat(String format) throws IllegalArgumentException {
```
eugenepeh: lets group the validation methods together.

jylee-git:
> @jylee-git can you please help me to check if the filetype / fileformat strings are used in any of the commandrunner (which includes git) execution?

Only one usage of file type/format in command runner

File format is used in git log command `GitLog#get`, line 22, where file formats are converted into git format args using `GitUtil#convertToGitFormatsArgs`.

eugenepeh:
>Only one usage of file type/format in command runner

As this introduces potential security issues, let's take this off the proposed items:
* Loosen the validation of FileType label to accept any non-empty 
  string

>When format is not specified, files without a standard file format
and with special characters (e.g.: "$gradle_backup$", "Backup File") 

For files without a standard file format and with special characters, let's treat them as files without format and group them under others by default

jylee-git:
> > When format is not specified, files without a standard file format
> > and with special characters (e.g.: "$gradle_backup$", "Backup File")
> 
> For files without a standard file format and with special characters, let's treat them as files without format and group them under others by default

| File name | FileType (format) to be assigned |
|------------|------------------------------------|
| Backup File | other |
| makefile | makefile |
| Backup File.java | java |
| $duke_backup$.java | java |
| duke.j@va | other |

i.e.: After tokenizing the file name, if the result contains any non alpha-numeric characters, we consider them as `other`?

eugenepeh:
>i.e.: After tokenizing the file name, if the result contains any non alpha-numeric characters, we consider them as other?

Yes, that should be good for now since its not an industrial format hence we can treat them as others. 
For the format field, we should continuing to prevent people from using non-alphanumeric characters.

PR #878, opened 2019-08-28
[#865] AuthorConfig: further improve the branch field

### LINKED ISSUE #865: AuthorConfig: further improve the branch field
In the author config, branch is an optional field that will automatically uses the default branch in the repo field.

However, this can be improved such that if the branch field is not filled, we can bind that AuthorConfig to all the RepoConfig that has the same repo's location irregardless of branch.

0blivious:
Fixes #865
```
In the author-config.csv, branch is an optional field that will
automatically uses the default branch in the repo field.

However, this can be improved such that if the branch field is not
filled, we can bind that AuthorConfig to all the RepoConfig that has the
same repo's location irregardless of branch. This can help users to
easily setup config files to analyse multiple branches of a repo.

Let's bind the AuthorConfig to all the RepoConfig that has the same
repo's location irregardless of branch if the branch field is not
filled.
```


jylee-git:
Can we have a test case to have this scenario covered?

0blivious:
I am not very sure about the test case you are referring to. I have updated the csv test file but currently there seems to be no junit test targeting at the merge function btw RepoConfig and Author Config. Do we need to have that?

0blivious:
ongoing

jylee-git:
> ongoing

Ping here once done to continue with reviews

src/main/java/reposense/model/Author.java
```diff
@@ -140,6 +140,7 @@ public PathMatcher getIgnoreGlobMatcher() {
      */
     public void appendIgnoreGlobList(List<String> ignoreGlobList) {
         validateIgnoreGlobs(ignoreGlobList);
+        this.ignoreGlobList.removeAll(ignoreGlobList);
```
fzdy1914: This line added does not match the name of the method. Make it a separate method like `setIgnoredGlobList` if it is truly needed.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -105,14 +106,22 @@ public static void merge(List<RepoConfiguration> repoConfigs, List<AuthorConfigu
...
+                matchingRepoConfig.setAuthorConfiguration(authorConfig);
+            }
         }
 
         for (AuthorConfiguration authorConfig : authorConfigs) {
             if (authorConfig.getLocation().isEmpty()) {
                 for (RepoConfiguration repoConfig : repoConfigs) {
                     repoConfig.addAuthors(authorConfig.getAuthorList());
                 }
+            } else if (authorConfig.isDefaultBranch()) {
```
fzdy1914: The logic here is acceptable. But we may have a helper functions like `getLocationMatchingRepoConfigs` to reduce the code complexity.

src/main/java/reposense/model/Author.java
0blivious: Using the current approach it is needed to remove the duplicates.

0blivious:
> A quick review:
> Have you considered the situation where there is no `author-config.csv` specified? What will the branch be used?

Not very sure about the question. According to the current UG if author-config.csv is not given and the repo has not provide author details in a standalone config file, all the authors of the repositories within the date range specified (if any) will be analyzed. This should not be affected by this PR.

src/main/java/reposense/model/Author.java
```diff
@@ -139,6 +139,16 @@ public PathMatcher getIgnoreGlobMatcher() {
...
+        validateIgnoreGlobs(ignoreGlobList);
+        this.ignoreGlobList.addAll(ignoreGlobList);
+        updateIgnoreGlobMatcher();
+    }
+
+    /**
+     * Validates and adds a list of ignore glob into the {@code Author} class instance variable without duplicates
+     * and updates the ignore glob matcher.
+     */
+    public void appendIgnoreGlobListWithoutDuplicates(List<String> ignoreGlobList) {
```
fzdy1914: A better name is `setIgnoreGlobList' I believe.

fzdy1914: If it is truly `appendIgnoreGlobListWithoutDuplicates`, we should use a set instead of removing them all.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -105,14 +106,22 @@ public static void merge(List<RepoConfiguration> repoConfigs, List<AuthorConfigu
...
+                List<RepoConfiguration> matchingRepoConfigs = getMatchingRepoConfigsByRepoLocation(repoConfigs,
```
fzdy1914: Suggest to be `getMatchingRepoConfigsByLocation` for simplisity.

src/main/java/reposense/model/Author.java
0blivious: Now if we have a AuthorConfig with empty branch field, it will map to multiple repos. If the repos have common strings in the IgnoreGlobList, addAll for list will result in duplicates. Can use linkedhashset to replace arraylist without error. Do you wish to change the name from `IgnoreGlobList` to `IgnoreGlobSet` also?

fzdy1914: Sorry for the late reply.
For me, a pseudo code like below will be enough:
```
if(!list.contains(item)){
    list.add(item)
}
```

fzdy1914: Sorry for the incovenient caused, can you switch back to the list version?

0blivious: Updated

src/main/java/reposense/model/Author.java
```diff
@@ -144,6 +144,20 @@ public void appendIgnoreGlobList(List<String> ignoreGlobList) {
         updateIgnoreGlobMatcher();
     }
 
+    /**
+     * Validates and adds a list of ignore glob into the {@code Author} class instance variable without duplicates
+     * and updates the ignore glob matcher.
+     */
+    public void appendIgnoreGlobListWithoutDuplicates(List<String> ignoreGlobList) {
```
yong24s: Can this method replace `Author#appendIgnoreGlobList`?

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -105,14 +106,22 @@ public static void merge(List<RepoConfiguration> repoConfigs, List<AuthorConfigu
...
         }
 
         for (AuthorConfiguration authorConfig : authorConfigs) {
             if (authorConfig.getLocation().isEmpty()) {
                 for (RepoConfiguration repoConfig : repoConfigs) {
                     repoConfig.addAuthors(authorConfig.getAuthorList());
                 }
+            } else if (authorConfig.isDefaultBranch()) {
+                List<RepoConfiguration> matchingRepoConfigs = getMatchingRepoConfigsByLocation(repoConfigs,
+                        authorConfig.getLocation());
```
yong24s: Can use a break at a higher level for readability.

```
List<RepoConfiguration> matchingRepoConfigs = 
    getMatchingRepoConfigsByLocation(repoConfigs, authorConfig.getLocation());
```

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -148,7 +157,7 @@ private static RepoConfiguration getMatchingRepoConfig(
             List<RepoConfiguration> repoConfigs, AuthorConfiguration authorConfig) {
         for (RepoConfiguration repoConfig : repoConfigs) {
             if (repoConfig.getLocation().equals(authorConfig.getLocation())
-                    && repoConfig.getBranch().equals(authorConfig.getBranch())) {
+                    && (authorConfig.isDefaultBranch() || repoConfig.getBranch().equals(authorConfig.getBranch()))) {
```
yong24s: Can use a descriptive boolean for this condition, to improve readability?
Also, you can swap `authorConfig` and `repoConfig` around in the second part of the condition.

i.e. `boolean isDefaultBranchOrRepoConfigBranch = (authorConfig.isDefaultBranch() || authorConfig.getBranch().equals(repoConfig.getBranch()))`

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -158,7 +167,7 @@ private static RepoConfiguration getMatchingRepoConfig(
     /**
      * Returns a list of {@link RepoConfiguration} where the {@link RepoLocation} matches {@code targetRepoLocation}.
      */
-    private static List<RepoConfiguration> getMatchingRepoConfigsByRepoLocation(
+    private static List<RepoConfiguration> getMatchingRepoConfigsByLocation(
```
yong24s: Is this change necessary?

src/main/java/reposense/model/Author.java
0blivious: Can replace but I am not very sure if you want to keep the original one. 
 Compared to `appendWithoutDuplicates`, `Author#appendIgnoreGlobList` saves the time looping and searching for duplicates before adding, so it has better performance if we can make sure no duplicates exist, like the case in `AuthorConfiguration#update` which uses
`Author#appendIgnoreGlobList`.

src/main/java/reposense/model/RepoConfiguration.java
0blivious: This was changed according to fzdy1914's review.
> Suggest to be `getMatchingRepoConfigsByLocation` for simplisity.

fzdy1914: I thought this is a newly added method previously so I made that comment. But anyway I believe the current one is more clear.

src/main/java/reposense/model/Author.java
eugenepeh: If there is no clear advantage in keeping the `appendIgnoreGlobList`, it will be better to replace it since the processing speed of this is marginal. 

The new method can be named `importIgnoreGlobList` instead

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -105,14 +106,22 @@ public static void merge(List<RepoConfiguration> repoConfigs, List<AuthorConfigu
                 continue;
             }
 
-            matchingRepoConfig.setAuthorConfiguration(authorConfig);
+            if (!isDefaultBranch) {
```
eugenepeh: This check should be combined into `getMatchingRepoConfig` itself.
Reason: the entire purpose of this foreach block is to validate and perform the following line.
The logger warning message above can then be updated to match the new logic

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -147,8 +156,10 @@ public static void setGroupConfigsToRepos(List<RepoConfiguration> repoConfigs,
     private static RepoConfiguration getMatchingRepoConfig(
             List<RepoConfiguration> repoConfigs, AuthorConfiguration authorConfig) {
         for (RepoConfiguration repoConfig : repoConfigs) {
+            boolean isDefaultBranchOrMatchRepoConfigBranch = (authorConfig.isDefaultBranch()
```
eugenepeh: can be named `isMatchingBranch` instead

src/main/java/reposense/model/RepoConfiguration.java
0blivious: Sorry I don't really get this comment. So you want something like, if the `AuthorConfig` use default branch, `getMatchingRepoConfig` should not return any matching repo? The check here is actually used to prevent overwriting of `AuthorConfig` if one of them uses default brance.

eugenepeh: >So you want something like, if the AuthorConfig use default branch, getMatchingRepoConfig should not return any matching repo?

If this is intended to be part of the check, then I would say yes.
The thing is, now you have 2 separate checks that leads to the final conclusion which is 
```
                matchingRepoConfig.setAuthorConfiguration(authorConfig);
```
They should be ideally together, otherwise the `matchingRepoConfig` doesn't really serve much purpose if you're not going to use it in the end `if (!isDefaultBranch)`.


Coming back to the main topic (which I realized that I didn't asked earlier), given that you have only have an author config that is using default branch value, does that means that it won't be set/bind to any `RepoConfiguration` in this case?

0blivious: No, it will bind to all `RepoConfig` that has the same location

eugenepeh: It doesn't seems that you're performing `setAuthorConfiguration` if the author config is using default branch. In that case the ignore glob list is not being propagate to the authors.

Can you revise your PR? Also put up the test case for the scenario regarding the issue with ignore glob list.

0blivious: When using default branch, the `ignoreGlobList` is propagated via the method `matchingRepoConfig.addAuthors(authorConfig.getAuthorList())` in the code below. The `addAuthors` calls `addAuthor` in `AuthorConfig` which calls `propagateIgnoreGlobList`. The test case now already check for this issue.

eugenepeh: >When using default branch, the ignoreGlobList is propagated via the method matchingRepoConfig.addAuthors(authorConfig.getAuthorList()) in the code block below.

I see. With this new implementation, it doesn't quite make sense to split the process into 2 groups anymore; it become quite confusing. 

Can you combine and restructure the logic such that,

if location empty -> add to all repos, continue
find all matched location repos
if matched location repos is empty -> print repo not found, continue
if branch is default -> add to all matched location repos, continue
find all matched branch repos
if branch matches -> add to the repo, continue
print repo branch not found

>The test case now already check for this issue.

The test case did not include any ignore globs, can you add some to ensure that the functionality is as of expected?

0blivious: The reason for separating the loops is because now for matched repo we use `setAuthorConfiguration` which saves the time adding the authors into authorList in `AuthorConfig` one by one. If we don't complete all the matched repo first, the `setAuthorConfiguration` will overwrite the authorlist data of empty location/default branch ones.

0blivious: > The test case did not include any ignore globs, can you add some to ensure that the functionality is as of expected?

The `merge_twoRepoConfigs_success` in `CSVParserTest` uses a authorconfig.csv in which the second author using default branch without any ignoreGlobList, while the repoconfig.csv contains 2 repo that will be mapped to the default branch author, and the repo all have ignoreGlobList `Collated**`. The expected repo contains expected second author having ignoreGlobList Collated**. And the expected repo is compared to actual repo at the end using `TestUtil.compareRepoConfig`, which checks if the author having same ignoreGlobList (if the colated** is propagated to authorlist)

eugenepeh: >The reason for separating the loops is because now for matched repo we use setAuthorConfiguration which saves the time adding the authors into authorList in AuthorConfig one by one. If we don't complete all the matched repo first, the setAuthorConfiguration will overwrite the authorlist data of empty location/default branch ones.

Yes, but with the additional logics, it is no longer ideal to separate the logic anymore; it hinders the readability.

 >The merge_twoRepoConfigs_success in CSVParserTest uses a authorconfig.csv in which the second author using default branch without any ignoreGlobList, while the repoconfig.csv contains 2 repo that will be mapped to the default branch author, and the repo all have ignoreGlobList Collated**. The expected repo contains expected second author having ignoreGlobList Collated**. And the expected repo is compared to actual repo at the end using TestUtil.compareRepoConfig, which checks if the author having same ignoreGlobList (if the colated** is propagated to authorlist)

Yes, but without any ignore globs in the author configuration, can you really be sure that it is propagating and not overriding?

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -345,7 +355,7 @@ public void merge_emptyLocation_success() throws ParseException, IOException, He
         RepoConfiguration.merge(actualConfigs, authorConfigs);
 
         Assert.assertEquals(2, actualConfigs.size());
-        Assert.assertEquals(expectedConfigs, actualConfigs);
+        //Assert.assertEquals(expectedConfigs, actualConfigs);
```
eugenepeh: why is there a commented line?

0blivious: Sorry I forgot to remove it. Code updated

fzdy1914:
@0blivious The CI is failing, can you help look into it?

0blivious:
CI fixed

PR #880, opened 2019-08-31
Modify sort by author name to be according to authors' display names

jamessspanggg:
According to @damithc in the slack chat group:
> Yes, it should be display name, which we can control e.g., in this case, we have prefixed it with tutorial ID. The github username is a random string we cannot control. Sorting by it doesn't have much value.

```
The sort by author name implementation is according to the authors' 
GitHub usernames.

Github usernames are strings that users have no control over, 
an unconfigurable field.

Let's modify the sorting of author names to be according to the authors' 
display names.
```


jamessspanggg:
> Minor nit to message:

@eugenepeh done.

PR #883, opened 2019-09-03
[#868] Report: link footer to main site

### LINKED ISSUE #868: Footer: link to main site
Current: footer links to specific release
image

Suggested: keep the version number but link to the main site, as this is simply a way to market the product.

Tejas2805:
Fixes #868 

```
The footer provides a link to a specific release.

However, the footer link could also be used to navigate to the main
site, which would help to market our product.

Let's break the link into two parts to market RepoSense better.
'RepoSense' will link to the main site whereas the 'version number'
will link to a specific release.
```


jamessspanggg:
@Tejas2805 please wrap your commit message body at 72 characters.

jylee-git:
Should we put that commit number to a new line instead?

```
Generated by Reposense on ...
Version/commit abcdefg
```

Tejas2805:
> @Tejas2805 please wrap your commit message body at 72 characters.

@jamessspanggg Done.

yong24s:
> Should we put that commit number to a new line instead?
> 
> ```
> Generated by Reposense on ...
> Version/commit abcdefg
> ```

Current way looks good to me. Can ask Prof. for comment.

damithc:
Can we style it like this?
>[**RepoSense**]() ([v1.2.3]())

Tejas2805:
> Can we style it like this?
> 
> > **RepoSense** (v1.2.3)

Can do.

frontend/src/index.pug
```diff
@@ -37,10 +37,15 @@ html
           .timestamp-footer
             span Generated by &nbsp;
             a(
-              v-bind:href="getRepoSenseLink()", target="_blank",
+              v-bind:href="getRepoSenseMainSiteLink()", target="_blank",
             )
-              strong RepoSense {{ repoSenseVersion }}
-            span &nbsp; on {{ creationDate }}
+              strong RepoSense
+            span  (
```
fzdy1914: Two space here, one space is enough right?

Tejas2805: Umm, actually so the way this works is if there is a single space then using "span" just prints it. So I had to give to two space so that an actual space comes before the bracket.

Tejas2805: But I just realised there is an extra space after the last bracket, so will correct that and push again. 

Screen Shot 2019-09-16 at 11 51 04 PM

Tejas2805:
@fzdy1914 Could you request review from stage2-reviewers? I am not allowed to do it.

frontend/src/index.pug
eugenepeh: We should standardize to use `&nbsp;` for such intended spaces
E.g. double spaces, spaces at end of the line should be replaced with `&nbsp;` to highlight its presence

frontend/src/index.pug
```diff
@@ -37,10 +37,15 @@ html
           .timestamp-footer
             span Generated by &nbsp;
```
eugenepeh: if we not using additional buffer space at the side of "RepoSense" hyperlink word, can remove the additional space.

frontend/src/index.pug
```diff
@@ -37,10 +37,15 @@ html
...
             )
-              strong RepoSense {{ repoSenseVersion }}
-            span &nbsp; on {{ creationDate }}
+              strong RepoSense
+            span  (
+            a(
+              v-bind:href="getRepoSenseVersionSpecificLink()", target ="_blank",
+            )
+              strong {{ repoSenseVersion }}
+            span )&nbsp;on {{ creationDate }}
```
eugenepeh: the `&nbsp;` can be changed to normal space here since its rather obvious

frontend/src/static/js/main.js
```diff
@@ -250,7 +250,11 @@ window.app = new window.Vue({
       return JSON.stringify(dataObj);
     },
 
-    getRepoSenseLink() {
+    getRepoSenseMainSiteLink() {
```
eugenepeh: `getRepoSenseMainSiteLink` -> `getRepoSenseHomeLink` will be more appropriate.
MainSite is like suggesting you have other mirror sites

frontend/src/static/js/main.js
```diff
@@ -250,7 +250,11 @@ window.app = new window.Vue({
...
+      return 'https://github.com/reposense/RepoSense';
+    },
+
+    getRepoSenseVersionSpecificLink() {
```
eugenepeh: `Specific` can be removed cause its implied with singularity

Tejas2805: Okay.

frontend/src/static/js/main.js
Tejas2805: Didn't think of it this way. Thank you.

frontend/src/index.pug
Tejas2805: > We should standardize to use `&nbsp;` for such intended spaces
> E.g. double spaces, spaces at end of the line should be replaced with `&nbsp;` to highlight its presence

Oh okay. Didn't know we were using `&nbsp;` for this purpose. Thank you for guiding.

frontend/src/index.pug
Tejas2805: > Can we style it like this?
> 
> > **RepoSense** (v1.2.3)

I think the Professor wants this additional buffer based on his last reply. So, I think can keep it.

eugenepeh: If that's the case, the buffer on the other end should be retained as well [on line 43 removed].

Tejas2805: I think have retained it. Check my latest commit?

frontend/src/index.pug
```diff
@@ -37,10 +37,15 @@ html
...
             )
-              strong RepoSense {{ repoSenseVersion }}
-            span &nbsp; on {{ creationDate }}
+              strong RepoSense
+            span &nbsp;(
+            a(
+              v-bind:href="getRepoSenseVersionLink()", target ="_blank",
+            )
+              strong {{ repoSenseVersion }}
+            span ) on {{ creationDate }}
```
eugenepeh: you didn't, hence my comment earlier.

image

Double check your PR;that's the only area with 2 spaces

Tejas2805: Oh okay. Got it now. My bad. The  ** &nbsp;** was causing two spaces. Have fixed it. Thanks @eugenepeh for guiding.

damithc:
I think we should link to `https://reposense.org` to make it future proof?

frontend/src/static/js/main.js
```diff
@@ -250,7 +250,11 @@ window.app = new window.Vue({
       return JSON.stringify(dataObj);
     },
 
-    getRepoSenseLink() {
+    getRepoSenseHomeLink() {
+      return 'https://github.com/reposense/RepoSense';
```
damithc: Should be `reposense.org`?

Tejas2805: Have updated the link, but wished to understand the difference since both the links open the same page for me. Is it that the former link could have changed with time but the present one will continue to remain the same?

damithc: `reposense.org` is the product website while the github link is the dev website. At the moment former points to the latter, but eventually it will be a separate website, such as https://markbind.org/

frontend/src/static/js/main.js
```diff
@@ -250,7 +250,11 @@ window.app = new window.Vue({
       return JSON.stringify(dataObj);
     },
 
-    getRepoSenseLink() {
+    getRepoSenseHomeLink() {
+      return 'https://reposense.org';
```
damithc: It seems Chrome `https://reposense.org` causes browsers to give a security warning but `reposense.org` does not. Is it the same for you guys?

Tejas2805: Just checked. This error didn't come for me yesterday. But it is coming right now.

damithc: Should we omit `https://` part?

yong24s: Seems to be fine now.

@damithc 
If its convenient for you, adding a `CNAME` record to `https://reposense.org` will be the best course of action as the url shorter service doesn't seem to be reliable.

Tejas2805: Yup. Seems to be okay now.

damithc: I'm still getting the error. 
GoDaddy (our domain vendor) is supposed to redirect to GitHub.
Is there any downside to omitting https:// part? The browser seems to go to https:// anyway.

eugenepeh: i believe the problem is that the redirection from `reposense.org` to GitHub is just a service provided by the domain provider. Using `https://` in this case is incorrect as `reposense.org` is not tied to any physical server with ssl cert. 

Alternative, if we really want to add the `https://`, we can consider creating a gh page for reposense. GitHub provides the service where you can use your own domain name for your gh page, as well as the ssl certificate when you do so.

frontend/src/static/js/main.js
```diff
@@ -250,7 +250,11 @@ window.app = new window.Vue({
       return JSON.stringify(dataObj);
     },
 
-    getRepoSenseLink() {
+    getRepoSenseHomeLink() {
+      return 'reposense.org';
```
yong24s: It has to contain the protocol i.e. `http://reposense.org` else its being appended as a path.

yong24s: You can try it on [Netlify](https://deploy-preview-883--reposense.netlify.com/).

PR #888, opened 2019-09-09
[#886] Fix test causing Appveyor to fail

### LINKED ISSUE #886: Appveyor CI is failing
image

myteo:
Fixes #886

```
Appveyor CI is failing due to GitRevParseTest which checks for
existence of master branch.

This is because the git rev-parse command in the failing test method
has been using the wrong directory; the reference should be using the
target repo's root. The test, however, uses the repo folder name (which
is parent of the target repo's root), causing the test to fail.

Lets fix the test to use the correct directory.

When not in the target repo's root, git tends to use the resources of
higher order repositories (if it is found). On AppVeyor, RepoSense has
always been cloned as a repository, which subsequently clones the
target repositories into one of its subfolders for analysis. Hence, the
issue has been kept hidden as the verification has been performed
against the RepoSense's resources and not the target's resources.

This issue came to the light in the release commit when AppVeyor clones
the RepoSense as a release, which does not contain the master branch,
causing the git rev-parse to fail.
```

### DIFF
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -231,7 +231,7 @@ public class RepoConfiguration {
     }
 
     /**
-     * Gets the path to the root folder of the repository
+     * Gets the path to the root folder of the repository.
      */
     public String getRepoRoot() {
         String path = FileUtil.REPOS_ADDRESS + File.separator + getRepoFolderName() + File.separator;
@@ -244,7 +244,7 @@ public class RepoConfiguration {
     }
 
     /**
-     * Gets the name of the folder containing the cloned repository; the parent directory of the repo's root folder
+     * Gets the name of the folder containing the cloned repository; the parent directory of the repo's root folder.
      */
     public String getRepoFolderName() {
         return repoFolderName;
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -230,6 +230,9 @@ public class RepoConfiguration {
         }
     }
 
+    /**
+     * Gets the path to the root folder of the repository
+     */
     public String getRepoRoot() {
         String path = FileUtil.REPOS_ADDRESS + File.separator + getRepoFolderName() + File.separator;
 
@@ -240,6 +243,9 @@ public class RepoConfiguration {
         return path;
     }
 
+    /**
+     * Gets the name of the folder containing the cloned repository; the parent directory of the repo's root folder
+     */
     public String getRepoFolderName() {
         return repoFolderName;
     }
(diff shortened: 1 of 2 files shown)


myteo:
> Hmm, I don't quite get your commit message because

> Paths.get(FileUtil.REPOS_ADDRESS, config.getRepoFolderName()) and Paths.get(config.getRepoRoot()) are essentially the same except that getRepoRoot() has this additional chunk of code

        if (!getRepoName().isEmpty()) {
            path += getRepoName() + File.separator;
        }
> which wasn't described in your solution.

Sorry I didn't explain fully in the commit message. When RepoSense clones a repo, the following directory structure is created:

Screenshot 2019-09-09 at 8 37 05 PM

`getRepoFolderName()` will get the parent directory while `getRepoRoot()` will get the actual git repo.

> Also, another question we need to address is, why this issue did not surface for 43dadf1 but appears in 816bb1a?

I looked at the Appveyor logs for [43dadf1](https://ci.appveyor.com/project/eugenepeh/reposense/builds/27106233) and [816bb1a](https://ci.appveyor.com/project/eugenepeh/reposense/builds/27246107?fullLog=true).

For 43dadf1, RepoSense is cloned this way:
```
> git clone -q --branch=master https://github.com/reposense/RepoSense.git C:\projects\reposense
> git checkout -qf 43dadf14f5814b99e508a2ddfbde59112b21e134
```

For 816bb1a, it is cloned this way:
```
> git clone -q -n --branch=v1.10 https://github.com/reposense/RepoSense.git C:\projects\reposense
> git checkout -qf v1.10
```

For that particular test case which is failing, it is verifying that master branch exists. Notice that in 816bb1a, `v1.10` is specified in the branch field so master branch will not exist in the cloned repo, causing the test to fail.

The explanation in my original commit message is wrong, detached head is not the reason for failing, I have updated it.

eugenepeh:
@myteo thanks for the investigation, and sorry for the late reply, can you also help to add header comments to the 2 methods (`getRepoRoot` and `getRepoFolderName`) in RepoConfiguration to avoid any confusion in the future?

Specifically, for the `getRepoFolderName` we should state that it is the `folder containing the cloned repository; the parent directory of the repo's root`

and `getRepoRoot` can just state that it is the `root folder of the repository`

myteo:
@eugenepeh Added comments and updated message. Thanks for refining commit message.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -230,6 +230,9 @@ public void updateBranch() throws GitBranchException {
         }
     }
 
+    /**
+     * Gets the path to the root folder of the repository
```
eugenepeh: minor nits, just need to add full stop at the end of the sentence to keep it standardized.

myteo:
Done.

PR #895, opened 2019-09-17
[#885] Allow reset search filter

### LINKED ISSUE #885: Ramp chart: give a way to reset search filter
Current: the user has to manually delete the keywords and press enter to go back to the full view
Suggestion: give a way to clear the search keywords
image

0blivious:
Fixes #885
```
Currently, the user has to manually delete the keywords and press enter
to go back to the full view.

However, this can be improved by providing a reset button to clear the
search bar. This can help the user save time for deleting. 

Let's provide a button to reset the search bar.
```


yong24s:
@0blivious 

I think Prof. wants a clear button beside the search box.

damithc:
> I think Prof. wants a clear button beside the search box.

Yes, something similar to the `x` we have for start/end date controls, if possible.

0blivious:
Screenshot (10)
Like this can? The `x` will only be displayed when there is some input in the search bar.

damithc:
> Like this can? The `x` will only be displayed when there is some input in the search bar.

Yes, like that. Is there a reusable text box control that comes with that kind of functionality, just like the date picker?

0blivious:
Implemented use html input type search, but now the after pressing the clear button only the search bar will be cleared, the filtered summary below will not be reset. Is this ok?

damithc:
> Implemented use html input type search, but now the after pressing the clear button only the search bar will be cleared, the filtered result below will not be reset. Is this ok?

Is it hard to reset results as well? It feels more intuitive if both the text and the results are reset at once.

Tejas2805:
> > Implemented use html input type search, but now the after pressing the clear button only the search bar will be cleared, the filtered result below will not be reset. Is this ok?
> 
> Is it hard to reset results as well? It feels more intuitive if both the text and the results are reset at once.

@0blivious Hey, I was working on this issue, didn't see you had a PR already. My bad. For Prof's Recommendation, you can try replacing **v-on:change** with **v-on:input** should be okay.

0blivious:
@Tejas2805 Thx for the suggestion. Code updated

yong24s:
@0blivious 
The `X` doesn't show on Firefox.

0blivious:
@yong24s According to my search only Webkit derived browsers have the x button for input type search. Firefox does not support this style. I have changed the code to use pure css and js for the clear button.

frontend/src/static/css/style.scss
```diff
@@ -251,6 +251,38 @@ header {
...
+    right: -1rem;
+    vertical-align: middle;
+  }
+
+  #close-icon:after {
+    background-color: mui-color('grey', '600');
+    border-radius: 50%;
+    bottom: 5%;
+    color: white;
+    content: 'X';
```
yong24s: Use `×` :-).

frontend/src/static/css/style.scss
```diff
@@ -251,6 +251,38 @@ header {
...
+    font-weight: bold;
+    height: .5rem;
+    margin: auto;
+    padding: .1rem;
+    position: absolute;
+    right: 2%;
+    text-align: center;
+    width: .5rem;
+  }
+
```
yong24s:
```
/* Prevents edge from adding a clear button in search input box */
.summary-picker__search::-ms-clear {
    display: none;
}
```

You'll need this style to prevent having two `X` on Edge.

0blivious:
> Seems to have used a little too many code for a commonly found UI element.
> Mui library should have the component ready, consider using it since its already in our assets.

I am not very familiar with the Mui library and it seems that its `Textfield for search` is built on top of input type search, so the cross button will not work on Firefox. That's why I switched to pure css implementation......Can give me some hint about what MUI component can be used for this search bar / how to make the textfield work? Really appreciated.

eugenepeh:
>I am not very familiar with the Mui library and it seems that its Textfield for search is built on top of input type search, so the cross button will not work on Firefox. That's why I switched to pure css implementation......Can give me some hint about what MUI component can be used for this search bar / how to make the textfield work? Really appreciated.

Let me put it another way, here's where the problem comes in:
Software engineering is **NOT** simply working getting the end results. There are many other considerations and factors such as cost (in this case working hours etc), maintainability, value etc.

In your case, you have concluded that by simply changing the `input type=search` doesn't work for firefox, you implemented nearly 50 lines of change. But what's the value that it brings us? Is the reset search filter an important feature? If the change doesn't apply to firefox, are there significant number of users affected? Who is going to maintain it? And most importantly, if you spend 1 month implemented a feature (just stating an example here) that your boss doesn't want, it's all going to be wasted effort. 

I am not saying that your work is bad; it is decent and it works. But if firefox comes up with an update that fix this issue the next day, it is all wasted effort. At the same time, we can consider firefox is a small community and not implement the changes for them.

In this case, I am sorry to say that I would rather not implement this change for the firefox users because of the maintainability, as well as the controls doesn't fit well with the date reset control. An alternative we can go for (which is what I initially meant when I mentioned `commonly found UI element`) is to use libraries to implement this for all browsers. Because the only changes we have to make upon remove or modifying this feature is simply remove the script and that itself reduce the cost/value/maintainability problem.

damithc:
> Now the reset is implemented using mui button which will always appear besides the search bar. Is this one better now? Or I just use the search one and ignore the performance on firefox.

I'm OK with both options.

frontend/src/index.pug
```diff
@@ -87,8 +87,9 @@ html
         form.summary-picker.mui-form--inline(onsubmit="return false;")
           .summary-picker__section
             .mui-textfield
-              input.summary-picker__search(type="text", v-on:change="updateFilterSearch", v-model="filterSearch")
+              input.summary-picker__search(type="text", v-on:input="updateFilterSearch", v-model="filterSearch")
```
eugenepeh: this change doesnt seems to be part of what the PR is addressing.

0blivious:
Move the button into the textbox control and make the button appear inside search bar.
FireShot Capture 033 - RepoSense Report_ - http___localhost_9000_#search=adit

frontend/src/static/css/style.scss
```diff
@@ -253,14 +253,35 @@ header {
...
+    bottom: -8px;
+    box-shadow: none;
+    color: mui-color('grey');
+    font-size: .75rem;
+    font-weight: bold;
+    height: 15px;
+    left: -8px;
+    line-height: 5px;
+    margin: 0;
+    padding: 0;
```
eugenepeh: Perhaps the 
* bottom: -8px;
* height: 15px;
* line-height: 5px; 

Can be just changed to 
* vertical-align: middle;
* background: transparent;

eugenepeh:
@0blivious  why are the changes are different from requested?

Tejas2805:
Just a question - Will it not be better to have the `X` appear only when something is typed - As of now it is permanent and is there even if nothing is typed.

0blivious:
> @0blivious why are the changes are different from requested?

The requested code will make the cross button appear below the text so I adjusted a bit to make them align.

The image below is the result of the requested code.
FireShot Capture 058 - RepoSense Report_ - http___localhost_9000_#search=dddd

And this is the PR version
FireShot Capture 059 - RepoSense Report_ - http___localhost_9000_#search=cddd

0blivious:
> The `x` feels somehow horizontally compressed? The other part looks fine to me.
> 
> Do remember to reply to [#895 (comment)](https://github.com/reposense/RepoSense/pull/895#issuecomment-560430316)

It appears ok on my computer? It might be a bit troublesome to adjust the shape as the button now is implemented by simply using the letter `X` as the button. 


> Just a question - Will it not be better to have the `X` appear only when something is typed - As of now it is permanent and is there even if nothing is typed.

Might need some extra lines of code to implement this function since the current css package does not support this. Is it really necessary? I actually feel make the grey cross button always appear looks better since all the other field besides it have the little grey triangle at the right side.

fzdy1914:
> Not sure if the others want to take this into account.
Don't think it is very confusing, it is acceptable.

fzdy1914:
The user has to manually delete the keywords and press another enter 
to go back to the full view.

However, this can be improved by providing a reset button to clear 
the search bar, which can help the user save time for deleting. 

Let's provide a button to reset the search bar.

Please mind not to use `currently`, `now` and 72-word line break.

PR #896, opened 2019-09-18
[#893] Report#ContributionBar: fix incorrect width

### LINKED ISSUE #893: Improper width computed for contribution bar when breakdown by file type
When breakdown by file type, the width of one of the contribution bar is computed incorrectly.

image
An example is the blue `.md` contribution bar. 

Reproduce: https://nus-cs2103-ay1920s1.github.io/ip-dashboard/#sort=totalCommits%20dsc&groupSelect=groupByAuthors&search=whneo97&sortWithin=title&since=2019-08-16&timeframe=commit&mergegroup=false&breakdown=true&tabOpen=true&tabType=authorship&tabAuthor=whneo97&tabRepo=whneo97%2Fduke%5Bmaster%5D

myteo:
Fixes #893.

```
When breakdown by file format is selected, the width of the
contribution bars are computed incorrectly for some file types.

This occurs because the current implementation for calculating the
contribution bars does not keep track of the bar width of the current
line properly. In addition, the current implementation has rather
convoluted logic.

Lets fix and improve the logic for calculating contribution bar width
for file types.
```


frontend/src/static/js/v_summary.js
```diff
@@ -205,46 +205,36 @@ window.vSummary = {
...
-            const negativeWidth = res.pop();
-            res[res.length - 1] += negativeWidth;
+        if (currentBarWidth + barWidth < fullBarWidth) {
+          contributionBars.push(barWidth);
+          currentBarWidth += barWidth;
+        } else {
+          contributionBars.push(fullBarWidth - currentBarWidth);
+          barWidth -= fullBarWidth - currentBarWidth;
+          const numOfFullBars = Math.floor(barWidth / fullBarWidth);
+          for (let i = 0; i < numOfFullBars; i += 1) {
```
fzdy1914: `i+= 1` can be `i++` ?

myteo: Oh it causes lint error

fzdy1914: Then it is fine.

Issue #872, opened 2019-08-23
ConfigCsvFiles: Improve exception handling when the csv file(s) is/are malformed


jylee-git:
For example, `author-config.csv` should contain 7 columns (especially the headers). However, the user provided 6 columns (fewer columns than requred).

When program is run, exception is caught but the program is terminated abruptly, without giving the user enough info as to where went wrong:
![image](https://user-images.githubusercontent.com/35756209/63599911-216d9580-c5f5-11e9-861c-c7930104eab5.png)

###To-do
Improve on the exception handling, by informing the user that the csv file does not contain the required number of columns

anubh-v:
I think this issue can be broken into 2 related issues:

A) The user provides the correct number of columns in the header, but every record is malformed. This could happen because every record either has fewer / more columns than the header, or lacks data in mandatory columns.

In this case, RepoSense will not register any of the CSV records in the generated List of RepoConfigurations. It would be good to alert the user when this happens.

Note that when this issue occurs in the `repo-config.csv` file, then RepoSense will end abruptly just as mentioned in this issue, because RepoSense currently expects at least 1 repository to be specified in `repo-config.csv`.

B) The user provides fewer columns than required in the header, and the records are technically not malformed. In this case, RepoSense ends abruptly as shown by @jylee-git . As mentioned by @jylee-git, it would be good to provide more helpful exception messages.

anubh-v:
Here are my proposed solutions for the 2 issues identified above:

To solve, issue A, we could use the logger to explicitly warn the user that **all** the records in the CSV have been considered malformed. In the case of `repo-config.csv`, we should throw an exception since the file is not optional.

To solve issue B, we could perform validation on the size of the header in the CSV to ensure the number of columns in the header are as expected.

jylee-git:
The reason for this issue is as follows (in technical terms): 
Take an example of `repo-config.csv`, which has 6 header columns. Only the first column (repo location) is required. The rest of the fields are optional.

The apache CSV parser used will parse these CSV files by detecting the number of header columns into `CSVRecord` with a fixed size of columns. So if the number of header columns are lesser than the original 6 columns (let's say the `repo-config.csv` file provided only has 4 columns), each CSVRecord generated will contain the row of data of size 4 (indexes 0-3).

[...]

PR #903, opened 2019-09-24
[#871, #952] CSV files: Ignore blank lines before the header

### LINKED ISSUE #871: csv files: ignore blank lines
In csv config files, we can ignore blank lines rather than abort upon encountering a blank line.

### LINKED ISSUE #952: CsvParser: Error logs display incorrect line numbers when CSV file has empty lines
When we warn users about malformed lines in the CSV files, we provide the line numbers of the malformed lines in the log messages.

However, if the CSV file has empty lines, then the displayed line numbers are inaccurate.
Note that *empty* line refers to a line with only the newline character.

For example, suppose we have the following `repo-config.csv` file.
```
Repository Location,Branch,File formats,Ignore Glob List,Ignore standalone config, Ignore commit list
https://github.com/reposense/testrepo-Alpha.git, ,md,,yes


https://github.com/reposense/testrepo-Beta.git, , , ,,,,,,,
https://github.com/reposense/testrepo-Delta.git,master, , ,yes,
```
Line 1 has the CSV header.
To be well formed, each line should have 6 comma separated values.
**Line 2** is malformed because it only has 5 values and **line 5** is also malformed.
**Lines 3 and 4** are empty lines.

The observed error logs:
image

The logs correctly show that **lines 2 and 5** are malformed, but these lines are labelled as **line 2 and line 3** respectively.

The bug occurs because our CsvParser ignores empty lines and does not account for them when calculating the line numbers.

anubh-v:
Fixes #871, fixes #952 
```
CsvParser: add logic to ignore blank lines before the header

When CSV config files are parsed, the CsvParser currently considers
the first non-empty line as the header.

Hence, CSV files with blank lines before the header are parsed
incorrectly. The blank line is considered the header, which
causes all rows in the CSV to be flagged as malformed.

Let's modify CsvParser such that it skips all the blank lines
preceding the header, and picks the first non-blank line as
the header.

The Apache Commons CSV library currently does not
provide any methods to skip blank lines preceding the
header. This limitation is documented here:
https://issues.apache.org/jira/browse/CSV-127
Hence, let's implement the line-skipping behavior
ourselves using a BufferedReader.
```


src/main/java/reposense/parser/CsvParser.java
```diff
@@ -54,8 +54,9 @@ public CsvParser(Path csvFilePath) throws IOException {
         List<T> results = new ArrayList<>();
         Iterable<CSVRecord> records;
 
-        try (Reader csvReader = new FileReader(csvFilePath.toFile())) {
-            records = CSVFormat.DEFAULT.withFirstRecordAsHeader().parse(csvReader);
+        try (BufferedReader csvReader = new BufferedReader(new FileReader(csvFilePath.toFile()))) {
```
yong24s: Prefer `try (Reader csvReader ...`

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -76,6 +77,23 @@ public CsvParser(Path csvFilePath) throws IOException {
         return results;
     }
 
+    /**
+     * Returns an array of Strings representing the header of a CSV file.
+     *
+     * @param reader A BufferedReader opened on the CSV file.
+     * @return an array of Strings where each Strings is a column name from the CSV header.
+     * @throws IOException
+     */
```
yong24s: JavaDocs are optional for `private` methods but good to have.

Generally, avoid trivial or empty `@annotation`. 
You don't have to write everything but try to capture the precise behavior of the method.

```
/**
 * Returns the first non-empty line as an array of Strings representing
 * the CSV header row.
 *
 * @throws IOException if there are error accessing the given csv file.
 */
```

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -76,6 +77,23 @@ public CsvParser(Path csvFilePath) throws IOException {
...
+    private String[] getHeader(BufferedReader reader) throws IOException {
+        String currentLine = "";
+
+        while (currentLine.length() == 0) {
```
yong24s: May need to test this on an empty file.

src/main/java/reposense/parser/CsvParser.java
anubh-v: Thanks for the review @yong24s.
I think we'll need to use `BufferedReader` here, since the `getHeader` method relies on the `BufferedReader`'s ability to read lines (as documented [here](https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html#readLine--))

anubh-v:
@yong24s, thanks for the review.

I have made the following changes:
- Updated the javadocs as suggested.
- Added a test case that parses a CSV config with empty and blank lines before header row.
- Added a test case that parses a CSV config with only empty and blank lines. Note that, in my implementation, the `CsvParser` throws an `IOException` if the CSV config has only empty or blank lines.

The Travis CI build is failing because trailing whitespaces are detected in the **test CSV files** created for the test cases. What would be an appropriate method of resolving this?
Should I edit `config/travis/check-trailing-whitespace.sh` so that it ignores these test CSV files?

src/main/java/reposense/parser/CsvParser.java
anubh-v: Thanks. I have added a test case for this in `CsvParserTest.java`

anubh-v:
@yong24s - may I get your thoughts on the changes?
How should we proceed with the test CSV files? They contain trailing whitespaces by design, and these are failing the Travis build.

yong24s:
@eugenepeh 

Should we exclude test files in `src/systemtest/resources` and `src/test/resources` from being checked by `config/travis/check-trailing-whitespace.sh`?

eugenepeh:
>Should we exclude test files in src/systemtest/resources and src/test/resources from being checked by config/travis/check-trailing-whitespace.sh?

In that case, can create a subfolder each (named e.g. whitespace) under the respective resources to contain tests that require trailing whitespace, and white list that folder.

anubh-v:
Thanks @yong24s @eugenepeh , I've added a separate folder for the test cases that have trailing whitespaces and I've excluded this folder in the `check-trailing-whitespace.sh` script.

anubh-v:
@eugenepeh, thanks for your suggestion.

I looked at the `CSVFormat` [documentation](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html) again. There are 3 methods that may be relevant to us: `withTrim()`, `withIgnoreEmptyLines()`, and `withIgnoreSurroundingSpaces()`.

However, I was unable to get the expected behaviour from these 3 methods.

I have now applied these methods in the `CsvParser`, and the Travis / Appveyor build show that the test cases associated with this issue are failing. (see the `repoConfig_redundantLines_success` and `authorConfig_emptyConfig_fail` test cases in `CsvParserTest`).

It may be that I am not applying the `CSVFormat` methods correctly. Please let me know if this is the case.

### Minimal example

To illustrate the problem with these methods, here is a minimal program that uses `CSVFormat` to parse a CSV file.

```
import java.io.FileReader;                                                                                                                            
import java.io.Reader;                                                          
import java.io.IOException;                                                     
                                                                                
import org.apache.commons.csv.CSVFormat;                                        
import org.apache.commons.csv.CSVParser;                                        
import org.apache.commons.csv.CSVRecord;                                        
                                                                                
public class CsvTest {                                                          
                                                                                
    Iterable<CSVRecord> records;                                                
                                                                                
    String csvAddress = System.getProperty("user.dir") + "/src/main/java/sample.csv";
                                                                                
    try (Reader reader = new FileReader(csvAddress)) {                          
        records = CSVFormat.DEFAULT.withFirstRecordAsHeader().withIgnoreEmptyLines().withTrim().parse(reader);
                                                                                
        for (CSVRecord record : records) {                                      
            System.out.println(record);                                                
        }                                                                       
                                                                                
    } catch (IOException ioe) {                                                 
      System.out.println(ioe);                                                  
    }                                                                           
  }                                                                             
}                                                                               
  
```

Input CSV (3 blank lines before the header)

```
    
    
        
Buyer, Seller, Licenser
        Tom,Bob,Smith
Sam, Bob, Tim
```

Expected Output:
```
CSVRecord [comment=null, mapping={=0}, recordNumber=1, values=[Tom, Bob, Smith]]
CSVRecord [comment=null, mapping={=0}, recordNumber=2, values=[Sam, Bob, Tim]]
```

Observed output: 
**The first blank line is still parsed as the header.**
Remaining blank lines are converted to empty strings.

```
CSVRecord [comment=null, mapping={=0}, recordNumber=1, values=[]]
CSVRecord [comment=null, mapping={=0}, recordNumber=2, values=[]]
CSVRecord [comment=null, mapping={=0}, recordNumber=3, values=[Buyer, Seller, Licenser]]
CSVRecord [comment=null, mapping={=0}, recordNumber=4, values=[Tom, Bob, Smith]]
CSVRecord [comment=null, mapping={=0}, recordNumber=5, values=[Sam, Bob, Tim]]
```

Hence, it seems that using `withTrim` and `withIgnoreEmptyLines` does not prevent the blank lines from being parsed as headers and records. I faced the same problem with `withIgnoreSurroundingSpaces`

Is there something I am doing wrong here?

### Side note

We are using the `DEFAULT` `CSVFormat`,  which is already set to ignore empty lines.
Hence, I think we do not need to apply `withIgnoreEmptyLines` again.

In the examples above, I still applied `withIgnoreEmptyLines` for clarity.

anubh-v:
@eugenepeh 

I have used Gradle to setup a simple project for my minimal example above.
You can download the zip below and run `./gradlew run` to observe the program output.

[testCsv.zip](https://github.com/reposense/RepoSense/files/3774411/testCsv.zip)

anubh-v:
Thanks for the response @eugenepeh 

> When the document mention `allowing empty lines.` doesn't necessary mean ignore empty lines, it may just mean that it accepts empty line input as a legit input. 

Perhaps I can explain my understanding of the documentation.
From the documentation of the DEFAULT CSVFormat, I see that this format has a number of settings. One of the settings turns on the flag to ignore empty lines.

image

Based on this (and some manual testing), I concluded that the `DEFAULT` format will not parse empty lines in the CSV file into records.

This matches the documentation's explanation of the `withIgnoreEmptyLines` flag. [Link to docs](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#withIgnoreEmptyLines-boolean-).

`ignoreEmptyLines - true to ignore the empty lines between the records, false to translate empty lines to empty records.`

> Otherwise, why do they even come up with withIgnoreEmptyLines option?

It could be because users who define their own CSV formats (rather than using `DEFAULT`) would like to switch between ignoring/not ignoring empty lines.

anubh-v:
> Try instead of using withFirstRecordAsHeader(), use withHeader().
Also do try out switching the order of the .withFirstRecordAsHeader().withTrim().withIgnoreEmptyLines()
to something like .withTrim().withIgnoreEmptyLines().withHeader() because some libraries may work differently depending on how you build it.

Thanks for the suggestions @eugenepeh. I will try them out.

I also have another solution to propose. Could I get your thoughts on it?

Currently, we parse the first line of the CSV file as the header.
We get erroneous behavior if header is malformed.
For example, the header may have only whitespace (i.e. this issue) or the header may not have the correct number of columns (as seen in #872)

**What if we defined the CSV header in our code, rather than parsing the header from the CSV?**

This will eliminate bugs arising from malformed headers (i.e. leading blank lines, wrong number of columns in header).

For example, given a CSV of this structure:
```
ID, Country, Name
1, Brazil, Sam
2, USA, Bob
```

we could parse it as follows:
```
Iterable<CSVRecord> records = CSVFormat.DEFAULT.withHeader("ID", "Country", "Name").withTrim().parse(reader);
```

We can define a header for each type of config file.

eugenepeh:
>What if we defined the CSV header in our code, rather than parsing the header from the CSV?

The purpose of the CSV header is to guide the user in filling the fields

>This will eliminate bugs arising from malformed headers (i.e. leading blank lines, wrong number of columns in header).

Its actually good for the errors to occur so that the user knows that they are filling in the fields wrongly or they have used an outdated config file.

The error message can be modified to be more helpful though, in assisting the user in getting an updated one, e.g. link to the csv file.

anubh-v:
Thanks, @eugenepeh. Let me address your comments.

>Try instead of using withFirstRecordAsHeader(), use withHeader().
Also do try out switching the order of the .withFirstRecordAsHeader().withTrim().withIgnoreEmptyLines()
to something like .withTrim().withIgnoreEmptyLines().withHeader() because some libraries may work differently depending on how you build it.

(1) I tried the different orderings, and it does not achieve our goal of ignoring the blank lines that occur before the header of the CSV config. The parser still considers the first blank line as the header.

I also looked into the source code of `CSVFormat` and `CSVParser`. To the best of my understanding, the order in which we call `withHeader()`, `withTrim()`, `withIgnoreEmptyLines()` does not matter. In fact, we should not need to call `withIgnoreEmptyLines()` as the documentation states this flag is already enabled in the `DEFAULT` CSVFormat that we are using.

**I have reverted this PR to my original approach in which I use a `BufferedReader` to skip past the blank lines in the CSV.**

(2) I have also attempted an alternative solution for this issue.
**In this PR**, we are inferring the CSV's header from the config file.
**In my alternative solution**, we define the header in our code and parse the CSV according to this header.

This means that all blank lines are considered malformed and they hence get ignored.

Please look at PR #933 to see this alternative solution. 

Here are the pro's and con's of the two approaches (i.e. **this PR** vs PR #933)

|PR number | Description of approach | Pros | Cons |
|---| ------------------------------| -------| --------|
|PR 903 | Use a `BufferedReader` to skip past the blank lines | Silently ignores the blank lines before the header | Implemented using 10 lines of code |
|PR 933 | Define the header in our code instead of inferring it from the CSV file | 1 line implementation that uses existing methods in Apache Commons CSV library | Does not silently ignore blank lines. The logger prints a warning for each blank line before the header. |

What do you think of the alternative solution ?

### Other replies
>The purpose of the CSV header is to guide the user in filling the fields

I am not proposing that we no longer require the user to write a header in the CSV.
I am simply suggesting that RepoSense should not need to read the CSV in order to determine the size of the header. The correct header size could be pre-defined in the codebase.

src/test/resources/CsvParserTest/require_trailing_whitespaces/repoconfig_redundantLines_test.csv
```diff
@@ -0,0 +1,6 @@
+  
+
+
+                   
+Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config,Ignore Commits List
+https://github.com/reposense/testrepo-Beta.git,master,,,,
```
eugenepeh: please add several more lines with spaces in between as well

anubh-v: I've added 2 more repositories and several blank / empty lines.

anubh-v:
Thanks very much for the review @eugenepeh 
I've modified the test CSV according to your comments. Ready for further review.

>Incorrect header size may imply that they are not using the correct header as well, the fields would probably be filled wrongly consequently, hence it is appropriate to signal an error early to the user.
Anyway, that is a whole separate issue and that conversation should be continued (if required to) in a separate issue/PR.

This issue has already been documented in #872.
I can fix it once this PR is resolved.

anubh-v:
@eugenepeh, may I get your feedback on the changes? Thanks for your time.

eugenepeh:
Do include this in your proposed message:
https://issues.apache.org/jira/browse/CSV-127

anubh-v:
Thanks @eugenepeh  and @yong24s 

anubh-v:
Is there anything else I need to do so that this PR can be merged?

anubh-v:
@jylee-git , may I seek your help to provide the 3rd approval for this PR? Thanks very much!

anubh-v:
@eugenepeh, thanks for the response.

I've done a partial fix for this bug, by adding logic to track the number of lines in the file before the **first** CSV record.

My fix will resolve the bug if the CSV file has **no empty lines after the header of the CSV**.
Empty lines are lines which only have a newline character. (They are different from *blank* lines, which consist of whitespace and a newline character).

If there are empty lines AFTER the CSV header, the bug will still be observed.
This is an existing bug in RepoSense, which I have documented in issue #952.

The bug occurs because our CsvParser ignores empty lines and does not account for them when calculating the line numbers.

Should we tackle this bug separately?

anubh-v:
> If your solution is meant to target the blank lines, fixing this bug should be part of the solution. 

@eugenepeh, this PR tackles **blank** lines. The bug is caused by **empty** lines.

Blank lines have only whitespace and a newline character.
Empty lines only have a newline character.

This bug is also **not introduced in this PR**. It is present in the existing RepoSense master branch.

> However, before we discuss on whether should this bug be tackled separately, could you please do an investigation? We will need to re-evaluate whether apache's csv library fits our use case.

Are you asking me to investigate whether this bug can be resolved using the Apache CSV library ?

eugenepeh:
>@eugenepeh, this PR tackles blank lines. The bug is caused by empty lines.

In most context, blank and empty lines belong to the same category, both does not provide any information. I know you're trying to highlight the difference that causes the bug, but to most people, it wouldn't mean a difference.

>This bug is also not introduced in this PR. It is present in the existing RepoSense master branch.

Yes, but when working on an issue, do look at it holistically. Prior to this PR, we can simply warn the users that bug may arise due to blank lines (so if any things happen, it can be said that its users' problem). However, if you're going to give the assurance that RepoSense will work as normal with blank lines, then the corresponding bugs should also be fixed. 

>Are you asking me to investigate whether this bug can be resolved using the Apache CSV library ?

Yes.

anubh-v:
@eugenepeh, I have fixed the line numbering bug. 
I did this by turning off the parser's default policy of ignoring empty lines.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -76,6 +79,25 @@ public CsvParser(Path csvFilePath) throws IOException {
...
+     * @throws IOException if the file has only empty or blank lines, or if there is an error accessing the file.
+     */
+    private String[] getHeader(BufferedReader reader) throws IOException {
+        String currentLine = "";
+
+        // read from file until we encounter a line that is neither blank nor empty
+        while (currentLine.length() == 0) {
+            currentLine = Optional.ofNullable(reader.readLine())
+                                  .map(String::trim)
+                                  .orElseThrow(IOException::new);
```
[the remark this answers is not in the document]
anubh-v: I have fixed the line wrapping.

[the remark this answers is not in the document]
anubh-v: @eugenepeh, I have made a change my approach for the exception handling.

I created a new exception, `InvalidCsvException`, which signifies that there is a major error in the contents of a CSV file, such that the CSV is totally unusable.
For example, the CSV file may be totally empty, or all of its records may be malformed.

Reason for creating new exception, rather than using `IOException`: The code becomes more readable as the exception is more specific.

There is a minor problem, though. Ideally, we would like the logger to inform the user if the `author-config.csv` and `group-config.csv` have major problems.

However, currently, RepoSense will silently ignore all `IOExceptions` that occur for `author-config.csv` and `group-config.csv`.

To fix this, we may need to throw the `InvalidCsvException` from `parse` in `CsvParser` to the main code in `RepoSense.java.`, where it can be handled for `author-config.csv` and `group-config.csv`.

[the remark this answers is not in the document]
eugenepeh: Sounds good to me

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -173,6 +182,12 @@ public void authorConfig_emptyLocation_success() throws ParseException, IOExcept
         Assert.assertEquals(AUTHOR_CONFIG_NO_SPECIAL_CHARACTER_AUTHORS, authorConfig.getAuthorList());
     }
 
+    @Test (expected = IOException.class)
+    public void authorConfig_emptyConfig_fail() throws IOException {
```
jylee-git: For method name, instead of just stating `fail`, better to state a better descriptive result (`..._throwsIoException()`)

eugenepeh:
Can be merged after fixing the method name that @jylee-git has pointed out

src/test/java/reposense/parser/CsvParserTest.java
anubh-v: Done.

anubh-v:
> Can be merged after fixing the method name that @jylee-git has pointed out

@eugenepeh, what do you think about the exception handling issue?
Should it be fixed in this PR?

Problem with current situation: If the `author-config.csv` or `group-config.csv` files are present but totally empty, then they are silently ignored without any warning to the user.

Suggested fix: Throw the `InvalidCsvException` out of `parse` in `CsvParser` to the main code in `reposense.java`, and then use this exception to warn the user about the empty `author-config` / `group-config` files.

eugenepeh:
>Should I make this fix in this PR?

@anubh-v i think that is closer to what you're trying to accomplish in #916, this PR should purely focus on dealing with blank lines.

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -183,7 +183,7 @@ public void authorConfig_emptyLocation_success() throws ParseException, IOExcept
     }
 
     @Test (expected = IOException.class)
-    public void authorConfig_emptyConfig_fail() throws IOException {
+    public void authorConfig_emptyConfig_throwsIoException() throws IOException {
```
eugenepeh: for this, can just use the same case letters as the `IOException` to avoid complicating the standard.

anubh-v: But our checkstyle rules prohibit having more than 2 consecutive capital letters right?
I get a checkstyle error when I set the test case name to `authorConfig_emptyConfig_throwsIOException`

anubh-v: I just update the test case to use the name `authorConfig_emptyConfig_throwsIOException`.

Below is the checkstyle error seen on the Travis logs:
image

It looks like we cannot use more than 2 consecutive capital letters. Am I doing something wrong here?

eugenepeh: I see.  My concern was to retain original name to avoid double standard. But since its prohibited by checkstyle then I guess we can stick to `throwsIoException`.

anubh-v: Done. I have changed the name back to `authorConfig_emptyConfig_throwsIoException`

anubh-v:
I have changed the test case name back to `authorConfig_emptyConfig_throwsIOException`.
There are no other changes pending for this PR.

Issue #576, opened 2019-02-26
refactor frontend to move v_summary and v_authorship into sub component files


ongspxm:
In #573, we have done some major refactoring to the frontend, moving out files into their own components. This is done as part of the effort to make reposense's dashboard much more extendable in the future.

As such, we should move `v_summary` and `v_authorship` into their own `.pug` templates as well. This should be done asap, but after the current frontend PRs are closed.

Tejas2805:
@eugenepeh @damithc @yong24s for refactoring v_summary which of the following options do you all think I should go ahead with?

**Option 1:**
Refactoring the whole `v_summary` into one sub-component file.

**Option 2:**
Refactoring the `v_summary` and `summary-charts` into different sub-component files.

fzdy1914:
@Tejas2805 Latter is preferred. For the same issue, you may also explore to separate a `summary-charts.js` from `v_summary.js`.

PR #908, opened 2019-10-05
[#576]Refactor v_authorship into sub component file

Tejas2805:
Part of #576 

```
Currently, the entire v_authorship template resides in the index pug file.

Let's refactor v_authorship to move it into its own component so that 
reposense's dashboard is much more extendable in the future. 
```


Tejas2805:
@eugenepeh @yong24s If you both okay with it, I think this would be a suitable time to merge the refactored code as there is no ongoing PR related to authorship and hence we can avoid conflicts.

PR #914, opened 2019-10-07
[#900] group-config.csv: make location optional

### LINKED ISSUE #900: group-config.csv: provide a way to specify groups for all repos
Current: groups need to be specified for each repo, one row per repo per group. For example, specifying 3 groups for 330 repos result in nearly 1000 rows.

Suggestion: allow a way to specify groups for all repos at once

For example, we can allow using `*` as the repo location.

0blivious:
Fixes #900
```
In the group-config.csv, location is a compulsory field that only has one
value.  Specifying groups for all repos requires the user to manually key
in groups for each repo.

However, this can be improved such that if the location field is not
filled, we can bind that GroupConfig to all the RepoConfig. This can
help the user save the time listing all the possible combinations of
repos and groups when they want to specify groups for all repos.

Let's bind the GroupConfig to all the RepoConfig if the location
field is not filled.
```


src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -3,13 +3,17 @@
...
+
+import reposense.system.LogsManager;
 
 /**
  * {@code FileTypeManager} is responsible for holding a list of whitelisted formats and user-specified custom
  * groupings.
  */
 public class FileTypeManager {
     private static final String DEFAULT_GROUP = "other";
+    private static final Logger logger = LogsManager.getLogger(FileTypeManager.class);
```
anubh-v: Do you plan to use this logger somewhere?

src/main/java/reposense/model/FileTypeManager.java
```diff
@@ -79,6 +83,23 @@ public boolean hasSpecifiedFormats() {
         return groups;
     }
 
+    /**
+     * Adds new groups from {@code groupList}. Skips groups that have already been added previously.
+     */
+    public void addGroups(List<FileType> groupList) {
+        for (FileType group : groupList) {
```
anubh-v: You could use stream operations to simplify this loop into a single line
for example, `groupList.stream().filter(this::containsGroup).forEach(groups::add)`

anubh-v:
Perhaps the user guide should be updated in accordance with this change?

src/main/java/reposense/model/FileTypeManager.java
0blivious: I forgot to remove the logger. I added coz I wanted to give a warning message if the group already exists in the list, sth like `Skipping group as %s already in repository %s %s`. But then I notice fileTypeManager does not store the repo location and branch name. So if this warning message is not that important I feel I should not let fileTypeManager access additional information?

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -124,6 +124,9 @@ public static void setGroupConfigsToRepos(List<RepoConfiguration> repoConfigs,
             List<GroupConfiguration> groupConfigs) {
         for (GroupConfiguration groupConfig : groupConfigs) {
             if (groupConfig.getLocation().isEmpty()) {
+                for (RepoConfiguration repoConfig : repoConfigs) {
```
fzdy1914: For these lines, it may be better to just set matchingRepoConfig to all the repo, so we reduce code duplicates.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -123,19 +123,20 @@ public static void merge(List<RepoConfiguration> repoConfigs, List<AuthorConfigu
     public static void setGroupConfigsToRepos(List<RepoConfiguration> repoConfigs,
             List<GroupConfiguration> groupConfigs) {
         for (GroupConfiguration groupConfig : groupConfigs) {
+            List<RepoConfiguration> matchingRepoConfigs;
             if (groupConfig.getLocation().isEmpty()) {
-                continue;
+                matchingRepoConfigs = repoConfigs;
```
fzdy1914: Seems that it can be feature simplified with `? :`, but it is OK to leave it here.

eugenepeh:
@0blivious theres a conflict that needs to be resolved

0blivious:
@eugenepeh conflict resolved

PR #916, opened 2019-10-09
[#942] CsvParser: Improve exception handling when csv config has no valid configs

### LINKED ISSUE #942: RepoConfigCsvParser: Improve exception handling when repo-config csv has no valid records
RepoSense crashes with an unhelpful error message when all the repo configs in `repo-config.csv.` are malformed.

To recreate this error: create a `repo-config.csv` with the following contents:

```
Repository Location,Branch,File formats,Ignore Glob List,Ignore standalone config, Ignore commits
https://github.com/reposense/RepoSense.git,master,,,
https://github.com/reposense/testrepo-Alpha.git,master,,,
https://github.com/reposense/testrepo-Beta.git,master,,,
```

Note that **all 3 repo configs in this CSV are malformed.** They all have only **5 columns**, although the CSV header has **6 columns.**

Now, run  `java -jar RepoSense.jar --config <path to directory containing repo-config.csv>`

RepoSense will crash as shown in the screenshot below:

image

Reason for the crash:
The `cloneAndAnalyzeRepos` method in `ReportGenerator` class is given a List of RepoConfigurations and the method assumes there will be at least 1 RepoConfiguration in the List.

Please see line 144 in `ReportGenerator.java`:
image

To fix this error, let's check that the `repo-config.csv` has at least 1 **valid** RepoConfiguration **during the parsing process**. We can show a helpful error message to the user if all records in `repo-config.csv` are malformed.

anubh-v:
Fixes #942
This PR also partially implements the changes requested in #937 
```
CsvParser: Add better exception handling for the case where
the csv config has zero valid records

RepoSense crashes with an unhelpful error message if the repo-config
csv has zero valid records.

If the repo-config csv has no valid records, we should provide a
clear warning to the user and then terminate execution. In the
case of optional csv files, we should provide a warning and
then continue execution.

Let's implement this behaviour by explicity checking the number of
valid records found by CsvParser. 
Let's throw an InvalidCsvException if there are no valid records
in a csv file. 
```


anubh-v:
@jylee-git, any thoughts on this PR?

jylee-git:
From the proposed commit message, I think it's a good idea, but I suspect a blank header can be considered as another column under this circumstance (where expected number of columns = 1 column):

| Name | |
|------|----|
|Alice||
|Bob||
|Charlie|some random words that the user left there unexpectedly|
|Deb||

image


Can you help to investigate if such cases can happen? Where such case is treated as two columns instead of one column.

anubh-v:
Thanks for the response @jylee-git.

The answer would be that it depends on how the file is encoded in plaintext.

i.e. if the text file looks the one below, the `CSVParser` in the Apache Commons CSV library considers the CSV as having 2 columns
```
Name, 
Alice
Bob
Charlie
Delta
```

Note that the header row has a comma after `Name` --> `Name,` --> hence it is considered to have 2 columns. All the records will be considered malformed since they all have only 1 value while the header has 2 columns.

If the header did not have this comma (as below), it would be considered to have 1 column. All the records would be considered well formed.

```
Name
Alice
Bob
Charlie
Delta
```

In the example, you shared, the number of columns again depends on how the header was encoded in plaintext.

In the following CSV, we would have 1 column and the `Charlie` record would be considered malformed since it has 2 values.

```
Name
Alice
Bob
Charlie, some random words that user left here
Delta
```

In the following CSV, we would have 2 columns and **only** the `Charlie` record would be considered well formed.

```
Name,
Alice
Bob
Charlie, some random words that user left here
Delta
```

However, I am not quite very clear how these examples are related to this PR.
Do you see a problem in the implementation?

jylee-git:
> I am not very clear how these examples are related to this PR.
> Do you see a problem / gap in the implementation?



Actually, totally not related.. My bad for not reading the PR contents carefully..

WIll review the code

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -42,6 +44,19 @@ public RepoConfigCsvParser(Path csvFilePath) throws IOException {
         };
     }
 
+    /**
+     * @throws IOException if there are 0 valid records in the given csv, or if there are errors accessing the file.
+     */
+    @Override
+    public List<RepoConfiguration> parse() throws IOException {
+        List<RepoConfiguration> repoConfigs = super.parse();
+        if (repoConfigs.size() == 0) {
```
jylee-git:
```java
if (repoConfigs.isEmpty()) {
    //...
}
```

Would be more appropriate here

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -42,6 +44,19 @@ public RepoConfigCsvParser(Path csvFilePath) throws IOException {
...
+            throw new IOException(MESSAGE_ZERO_REPO_CONFIGS);
+        }
+
```
jylee-git: Eliminate unnecessary blank lines

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -42,6 +44,19 @@ public RepoConfigCsvParser(Path csvFilePath) throws IOException {
         };
     }
 
+    /**
+     * @throws IOException if there are 0 valid records in the given csv, or if there are errors accessing the file.
+     */
+    @Override
+    public List<RepoConfiguration> parse() throws IOException {
```
jylee-git: Need to follow the proper Javadoc comment conventions for all public methods, even if this method overrides the parent class, by stating the purpose of this method as well.

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -388,4 +390,10 @@ public void repoConfig_overrideKeyword_success() throws ParseException, IOExcept
         Assert.assertTrue(config.isIgnoreGlobListOverriding());
         Assert.assertTrue(config.isIgnoreCommitListOverriding());
     }
+
+    @Test (expected = IOException.class)
+    public void repoConfig_zeroValidRecords_fail() throws IOException {
```
jylee-git: Let's follow the existing naming convention: `repoConfig_zeroValidRecords_throwsIOException()`

src/main/java/reposense/parser/RepoConfigCsvParser.java
anubh-v: Good point. I have changed this.

src/main/java/reposense/parser/RepoConfigCsvParser.java
anubh-v: My apologies - changed.

src/test/java/reposense/parser/CsvParserTest.java
anubh-v: Thanks, I will update the name.
However, in our project Checkstyle configuration (in `config/checkstyle/checkstyle.xml`), we do not allow variable names to have more than 2 consecutive capital letters.

I will use the name `repoConfig_zeroValidRecords_throwsIoException`

src/main/java/reposense/parser/RepoConfigCsvParser.java
anubh-v: Thanks, I have updated the Javadocs for the `parse` method in `RepoConfigCsvParser`
I have also updated the Javadocs for `parse` in the parent class (`CsvParser`), as that method also did not have a proper comment.

anubh-v:
Thanks very much for the feedback @jylee-git 
I've made the changes. Ready for further review.

anubh-v:
@jylee-git, may I get your feedback on the changes?
Thanks for your time.

jylee-git:
Thanks for reminding; overlooked this by accident

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -42,6 +44,21 @@ public RepoConfigCsvParser(Path csvFilePath) throws IOException {
         };
     }
 
+    /**
+     * Parses the csv file associated with this instance of the {@code CsvParser} and returns a {@code List}
+     * containing the RepoConfigurations in this file.
+     *
+     * @throws IOException if there are 0 valid records in the csv, or if there are errors accessing the file.
```
jylee-git: @throws ... if there are **no** valid records ...

anubh-v: Thanks. I will update this comment.

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -42,6 +44,21 @@ public RepoConfigCsvParser(Path csvFilePath) throws IOException {
...
+     */
+    @Override
+    public List<RepoConfiguration> parse() throws IOException {
+        List<RepoConfiguration> repoConfigs = super.parse();
+        if (repoConfigs.isEmpty()) {
+            throw new IOException(MESSAGE_ZERO_REPO_CONFIGS);
+        }
+        return repoConfigs;
```
eugenepeh: can't such a solution be introduced in the parent method so that it will apply to author config as well?

anubh-v: This overriding version of `parse` will terminate RepoSense execution if there are no valid records in the CSV.

Is this appropriate for the author config? I felt that we **should not** terminate RepoSense if there are no valid records in the author config, since author config is an optional file.

src/main/java/reposense/parser/RepoConfigCsvParser.java
anubh-v: Done.

src/main/java/reposense/parser/RepoConfigCsvParser.java
eugenepeh: Yes but the work on this PR doesn't seems to be addressing the original problem stated on #872 
So perhaps you should first address the problem and this PR can come after that.

anubh-v: @eugenepeh, sorry for the confusion.

The original problem described in #872 is that RepoSense terminates with an unhelpful error when the user does not provide the correct number of columns in a CSV.

This PR solves the problem that RepoSense terminates with an unhelpful error when there are zero valid configs in the `repo-config.csv`.

I should have created a separate issue for this PR. I will do so now.
My apologies for the error.

anubh-v: For the original problem given in #872, I plan to tackle it after PR #903 is merged, as my envisioned solution relies on the `getHeader()` method introduced in PR #903.

anubh-v:
I originally filed this PR under issue #872, as I felt the bug being tackled by this PR was closely related.

However, I should have created a separate issue for this PR to prevent confusion and facilitate smooth communication.

I have now created a separate issue for this PR. This PR fixes issue #942.

anubh-v:
@eugenepeh, I've created a separate issue for this PR. Sorry for the confusion earlier. Could I seek your comments on the changes?

eugenepeh:
@anubh-v looks good to me, but as mentioned, it will be more appropriate to implement this change after the original issue is addressed, so will put this on hold for now.

anubh-v:
@eugenepeh, thanks very much for spending time to review this PR.

I don't fully understand the rationale for placing this PR on hold.
May I clarify why issue #872 needs to be solved first?

I do plan to work on issue #872, but this work depends on another of my PR to be merged (#903)

This PR addresses a bug that is distinct from the one mentioned in #872.

eugenepeh:
>I don't fully understand the rationale for placing this PR on hold.
May I clarify why issue #872 needs to be solved first?
This PR addresses a bug that is distinct from the one mentioned in #872.

For PRs that are dependent on others, it is preferred to have them on hold (even if its ready to merge) to preempt any additional changes that has to be made upon integration. 

For your this PR, it is dependent on at which level do you intend to handle the exception from #872. It may also ended up that we don't need this PR at all, if its neater to handle it at the lower levels.

anubh-v:
@eugenepeh, thanks, I appreciate the clarification.

anubh-v:
@eugenepeh, @jylee-git, for this PR, after some thought, I think it is better appropriate to handle this exception in the `CsvParser.parse()` method, instead of creating an overriding `parse` method in `RepoConfigCsvParser`. This keeps the code simpler. What do you think?

eugenepeh:
>@eugenepeh, @jylee-git, for this PR, after some thought, I think it is more appropriate to throw this exception from the CsvParser.parse() method, instead of creating an overriding parse method in RepoConfigCsvParser. This keeps the code simpler. What do you think?

I am ok but would that be a problem given that the csv files can be optional?

anubh-v:
> I am ok but would that be a problem given that the csv files can be optional?

It should not be an issue. 
RepoSense will only terminate if there are no valid configs in `repo-config.csv`.  In the case of `author-config.csv` and `group-config.csv`, RepoSense will emit a warning and continue executing (as if these files were never provided).

Below will be the output when `repo-config.csv` has no valid configs:
*RepoSense **terminates** after emitting the warning.*

image

Below will be the output when `author-config.csv` has no valid configs:
*RepoSense continues execution after emitting the warning. It acts as if the `author-config.csv` was never provided.*

image

Below will be the output when `group-config.csv` has no valid configs:
*RepoSense continues execution after emitting the warning. It acts as if the `group-config.csv` was never provided.*

image


Is this alright?

anubh-v:
@eugenepeh, may I get your comments on this?

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -87,6 +91,10 @@ public CsvParser(Path csvFilePath, int expectedHeaderSize) throws IOException {
         } catch (InvalidCsvException ice) {
             throw new IOException(ice.getMessage(), ice);
         }
+
+        if (results.isEmpty()) {
+            throw new IOException(String.format(MESSAGE_ZERO_VALID_CONFIGS, csvFilePath.getFileName()));
```
fzdy1914: Should we define a new error type for this kind of error?

anubh-v: Should we use the `InvalidCsvException` instead?
The only issue is that we would need to change the `throws` clause of this `parse` method, which means several test cases need to be updated.

I'll go ahead and make this change.

fzdy1914: Can I know which test cases should be updated? Do remember that in one pinned issue, we are decided to use generic exception handling.

anubh-v: This will affect:
-  8 test cases in `AuthorConfigParserTest`
- 4 test cases in `GroupConfigParserTest`
- 8 test cases in `RepoConfigParserTest`
- 1 test case in `ArgsParserTest`

fzdy1914: Yes, I think we can take this chance to update the relevant test to generic exception handling.

anubh-v: @fzdy1914, is it ok if I update the style of **all test cases within the affected files** , even if some tests are not affected by this change?

For example, only 1 test case in `ArgsParserTest` is affected by this change.
Should I also update the other tests in `ArgsParserTest` to the new style?

anubh-v: @fzdy1914, I have updated the `parse` method in `CsvParser` to throw `InvalidCsvException` when the csv files are malformed. 

Because of this change in the `throws` clause of `parse`, I have had to update `RepoSense.java` as well as the following test files:
- `ArgsParserTest`
- `ConfigSystemTest`
- `RepoConfigParserTest`
- `GroupConfigParserTest`
- `AuthorConfigParserTest`
- `RepoConfigurationTest`

src/main/java/reposense/RepoSense.java
```diff
@@ -82,6 +83,8 @@ public static void main(String[] args) {
             logger.log(Level.WARNING, ioe.getMessage(), ioe);
         } catch (ParseException pe) {
             logger.log(Level.WARNING, pe.getMessage(), pe);
+        } catch (InvalidCsvException ice) {
```
anubh-v: @fzdy1914, should we combine the catch clauses for 
`IOException`, `ParseException`, `InvalidCsvException` ? 
The logic to handle these exceptions is the same.

fzdy1914: Yes, please.

https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/parser/InvalidCsvException.java#L6

I am also wondering if we should change the `InvalidCsvException` to extend `IOException` instead of generic `Exception`

anubh-v: > I am also wondering if we should change the InvalidCsvException to extend IOException instead of generic Exception


I am not sure if this is suitable. `IOExceptions` are thrown for errors that prevent us from accessing / writing a file. The `InvalidCsvException` is thrown when we can access the csv file, but the csv's contents are not in line with RepoSense's expectations.

anubh-v: @fzdy1914, I have collapsed the catch blocks

fzdy1914:
No need to include #937 as it is not a direct fix.

src/main/java/reposense/RepoSense.java
```diff
@@ -78,10 +79,8 @@ public static void main(String[] args) {
             if (cliArguments.isAutomaticallyLaunching()) {
                 ReportServer.startServer(SERVER_PORT_NUMBER, cliArguments.getOutputFilePath().toAbsolutePath());
             }
-        } catch (IOException ioe) {
+        } catch (IOException | ParseException | InvalidCsvException ioe) {
```
fzdy1914: The name of the exception should `e` as it can stand for multiple exceptions now.

src/main/java/reposense/RepoSense.java
```diff
@@ -103,9 +104,9 @@ public static void main(String[] args) {
         } catch (FileNotFoundException fnfe) {
             // FileNotFoundException thrown as author-config.csv is not found.
             // Ignore exception as the file is optional.
-        } catch (IOException ioe) {
-            // for all other IO exceptions, log the error and continue
-            logger.log(Level.WARNING, ioe.getMessage(), ioe);
+        } catch (IOException | InvalidCsvException excp) {
```
fzdy1914: same `e` here.

anubh-v:
> No need to include #937 as it is not a direct fix.

Ok. I have removed mention of [issue 937](https://github.com/reposense/RepoSense/issues/937) from the PR title.

src/main/java/reposense/RepoSense.java
anubh-v: Done.

src/main/java/reposense/RepoSense.java
anubh-v: I missed this. Done.

anubh-v:
@jylee-git, @eugenepeh, thanks very much for your time on this pull request.

Based on discussion with @fzdy1914, I have changed the exception type used for this error.
Previously, we were using `IOException`. Now, we are using `InvalidCsvException`.

The `throws` clause of the `parse` method in `CsvParser has been updated accordingly.
As a result, I had to update several test cases as well.

I would appreciate your review on these changes. Thanks very much.

anubh-v:
@jylee-git thanks for your time, may I get your feedback on these changes?

anubh-v:
@eugenepeh, may I get your comments on this?

anubh-v:
@fzdy1914 is this PR ok for being merged?

PR #922, opened 2019-10-16
[#920] Ramp chart: show percentile for each item

### LINKED ISSUE #920: Ramp chart: show percentile for each item
Current: we show the index number for each items in the ramp chart
image

Suggestion: also show the percentile of each item so that it is easy to see whether an item is in the 20%th percentile and so on. Perhaps as a grey color number at this end?
image

0blivious:
Fixes #920 
```
We have index number for each group in the ramp chart but not its
percentile. It is hard to tell the ranking position of the group.

This can be improved to show the percentile of each item so that it
is easy to see whether an item is in the 20%th percentile and so on.

Let's add a percentile number at the right side of each group.
```


jamessspanggg:
Perhaps limiting to 2 decimal places will look better? 
@damithc what's your opinion?

Screen Shot 2019-10-19 at 11 14 19 PM

damithc:
Yes, should limit. One decimal place is enough for a list of 1000 items?

0blivious:
I have updated the code to keep the decimal place to be 1 decimal point. Now even 0 will be mapped to 0.0. Is this ok or we prefer integer value if the original percentile number is an integer?

damithc:
> I have updated the code to keep the decimal place to be 1 decimal point. Now even 0 will be mapped to 0.0. Is this ok or we prefer integer value if the original percentile number is an integer?

I don't mind it. I think we should use the same number of decimal places for all percentile values in the chart.

frontend/src/static/js/v_summary.js
```diff
@@ -761,6 +761,10 @@ window.vSummary = {
       return filtered;
     },
 
+    getPercentile(index) {
+      return (Math.round(index * 1000 / (this.filtered.length - 1)) / 10).toFixed(1);
```
jamessspanggg: Is there a particular reason for using this way of computing instead of 
`((index * 100) / (this.filtered.length - 1)).toFixed(1)`?

0blivious: According to my search `ToFixed` might get different result on different browsers, depending on how they implemented it. The `math.round` was used to ensure the consistency of the result calculated across browsers.

yong24s:
@damithc 
Prof. can view the report to check if the requirements are met?

Should the percentile be shown when the ramps are grouped by group title?
Also, should the percentile order also invert when grouped by order is inverted?

damithc:
> Should the percentile be shown when the ramps are grouped by group title?

Yes.

> Also, should the percentile order also invert when grouped by order is inverted?

I guess no need.

I have one concern though. Does the percentage number is for all items _above_ the number? In that case the last item should not be 100% as there is one item below it. If the number is for the item it's attached to AND all the items above it? In that case the number at the top should not be 0% as it represents one item already.  The latter is preferred.  Let's say there are 5 items in the list. The percentage values should be as follows, right?
20% 
40%
60%
80%
100%
Currently it is:
0%
25%
50%
75%
100%

0blivious:
Percentile formula adjusted

damithc:
> @damithc
> Prof. can view the report to check if the requirements are met?

I'm OK with the current functionality.

eugenepeh:
Hmm, for this percentile feature, its a bit weird if we use "fractional values".
Also, it kinda gives the false impression that you're "20%" ahead of the following team,
especially when the sorting order is going by the title and not contribution. 

If the intended idea is to compare how each group is performing against their peers, shouldn't it be the percentage of the lines contribution contributed by the group to total number of lines?

damithc:
> Hmm, for this percentile feature, its a bit weird if we use "fractional values".
> Also, it kinda gives the false impression that you're "20%" ahead of the following team,
> especially when the sorting order is going by the title and not contribution.

The objective is to find who is in the bottom x% by whatever the search criteria is. For example, when sorted by contribution, we can find which students are in the bottom 15% in terms of contribution. Without this feature, the user has to do a mental calculation.

eugenepeh:
>The objective is to find who is in the bottom x% by whatever the sort criteria is. For example, when sorted by contribution, we can find which students are in the bottom 15% in terms of contribution. Without this feature, the user has to do a mental calculation.

Alright.
@0blivious please update the docs too

Tejas2805:
Can we do something like - We show the percentile only when sorting by `contribution` or `variance`. If sorted by `title` we don't show the percentile - or will this be a bit extreme ?

damithc:
> Can we do something like - We show the percentile only when sorting by `contribution` or `variance`. If sorted by `title` we don't show the percentile - or will this be a bit extreme ?

I think no need. If the percentage is meaningless, the user can ignore it. On the other hand, the user might find a use for it in ways we didn't anticipate.

eugenepeh:
>@eugenepeh May I know which part of the doc you wish to edit. The current UG doesnt have a separate section to describe the grouping feature. I have mentioned the percentile in the section Tool Bar, but I feel adding a separate section in Chart Panel under Interpreting the Report is more appropriate. Also we might need an update on all the doc image since the UI changed a lot now.

Just the image update on ui and short memo on the feature, I see that you have done both. LGTM.

0blivious:
> Percentile doesn't come for items when `group by none` is selected. Don't we want for each of the items for such a case?

That means the items are separated by "Repo and author". Since the percentile is used to display ranking, I feel it is only useful when we group them by either repo or author.

@damithc Do you think we need to have percentile even when  `group by none`

damithc:
> @damithc Do you think we need to have percentile even when `group by none`

there might be cases where the percentile value can still be useful when `group by none`, especially if the view is sorted by some criterion. I don't mind having it for that view as well.

0blivious:
@jylee-git I have updated the UG to add the example. Could you help check if the new one is ok?

0blivious:
> Some difference in the alignment with the `group by none`, could it be fixed?

`group by repo` has larger space as the intention here is to align the group percentile with group title. Would you like me to align it with the 1st summary-chart's title in each group ?
image

eugenepeh:
>group by repo has larger space as the intention here is to align the group percentile with group title. Would you like me to align it with the 1st summary-chart's title in each group ?

Given that both were allocated equal amount of space horizontally, the `%` should be flushed to the right in both case for standardization.

image

image

damithc:
How's this going?

0blivious:
> How's this going?

Code completed, waiting for final review then can be merged.

Tejas2805:
@fzdy1914 Can merge?

PR #928, opened 2019-10-26
[#736] CLI args: allow "--ignore-standalone-config" flag to override repo-config CSV

### LINKED ISSUE #736: Deal with "--ignore-standalone-config" when parsing config files
Current: when parsing CLI with CSV configs, `--ignore-standalone-config` is ignored (no errors/warnings shown to the user).

Suggested: Terminate the program (with error message) or warn the user if user sets `--ignore-standalone-config` when using CSV configs. 

Reason: Users might not be aware that the ignore flag has to be set in the csv file when parsing in .csv files.

anubh-v:
Fixes #736 
```
CLI Arguments: Allow --ignore-standalone-config flag to override CSV

RepoSense ignores the --ignore-standalone-config command line flag
when parsing a repo-config csv file.

We can allow the --ignore-standalone-config command line flag to
override all entries in the "Ignore Standalone Config" column of
repo-config.csv. This provides users with an easy method of setting
this flag for all repos, without having to edit each line of the csv.

To achieve this, let's move the isStandaloneConfigIgnored attribute
from the LocationsCliArguments class, to its superclass: CliArguments.
This will allow us to access the --ignore-standalone-config
flag's value when parsing RepoConfigurations from both
LocationsCliArguments and ConfigCliArguments.
```


anubh-v:
@jylee-git,  for this issue, we have the choice of either warning the user that the `--ignore-standalone-config` flag is ignored OR terminating the execution of RepoSense.

In the current implementation, I've chosen to issue a warning.

However, perhaps it is better to terminate the program? If we only provide a warning, the user may not read it and might continue using RepoSense with the impression that the `--ignore-standalone-config` flag works.

Terminating the program would clearly highlight the incompatibility of `--ignore-standalone-config` with `--config`. 

Do you think program termination is more appropriate?

jylee-git:
Need to ask @damithc for his opinion.

@damithc , as discussed sometime back, when a user chooses to use his/her config files and `--ignore-standalone-config` flag is used, what do you think will be an appropriate response?

anubh-v:
@eugenepeh thanks very much for working with me over the various PRs.
May I get your thoughts on this PR as well?

jylee-git:
From the proposed commit message title and the body, I can't really tell what's the intended approach here.

Is the intended approach (when the user uses `--ignore-standalone-config` flag) to ignore standalone configs for **all** repos in the repo-config.csv (regardless of whether the ignore standalone config flag in their individual repo location is set or not)? Or...

anubh-v:
@jylee-git, I have edited the commit message and PR title to make the approach clearer.

> Is the intended approach (when the user uses --ignore-standalone-config flag) to ignore standalone configs for all repos in the repo-config.csv (regardless of whether the ignore standalone config flag in their individual repo location is set or not)?

Yes.

Conversely, if the user **does not specify --ignore-standalone-config** on the command line, then we use the entries in the "Ignore Standalone Config" column of `repo-config.csv` to decide whether the standalone config is ignored for any given repo. 
This is the same as the current usage of `repo-config.csv`.

src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -185,6 +187,39 @@ public void repoConfig_ignoresStandaloneConfigInCli_success()
         TestUtil.compareRepoConfig(expectedConfig, actualConfig);
     }
 
+    @Test
```
fzdy1914: The logic of this test can be improved.  If you just use assert true and assert false, what is the meaning of having `inputWithoutIgnoreStandaloneFlag`? You may refer to other test methods to see how to we test our parameters.
We set up an `expectedConfig` and compare with the `actualConfig`.

anubh-v: @fzdy1914, thanks for your response.
I don't really understand the concern about the logic of this test.
The purpose of having `inputWithoutIgnoreStandaloneFlag` is to ensure that the CSV is parsed correctly when the user does not specify `--ignore-standalone-config` on the command line.

I can still change this test case to set up some expected configs and compare against actual configs.

anubh-v: @fzdy1914, I have changed the test case to use expected and actual configs.
Could I get your thoughts on it?

src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -185,6 +188,48 @@ public void repoConfig_ignoresStandaloneConfigInCli_success()
...
+        repoCharlieExpectedConfigWithIgnoreStandalone.setStandaloneConfigIgnored(true);
+
+        String inputWithoutIgnoreStandaloneFlag = new InputBuilder().addConfig(IGNORE_STANDALONE_FLAG_OVERRIDE_CSV_TEST)
+            .build();
+        String inputWithIgnoreStandaloneFlag = new InputBuilder().addConfig(IGNORE_STANDALONE_FLAG_OVERRIDE_CSV_TEST)
+            .addIgnoreStandaloneConfig()
+            .build();
+
+        CliArguments argsWithoutIgnoreStandaloneFlag = ArgsParser.parse(
+            translateCommandline(inputWithoutIgnoreStandaloneFlag));
```
jylee-git: Wrapped lines should have 8 spaces

src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -185,6 +188,48 @@ public void repoConfig_ignoresStandaloneConfigInCli_success()
...
+
+        RepoConfiguration repoCharlieActualConfig = repoConfigsWithSomeIgnoringStandaloneConfig.get(0);
+        RepoConfiguration repoDeltaActualConfig = repoConfigsWithSomeIgnoringStandaloneConfig.get(1);
+        TestUtil.compareRepoConfig(repoCharlieActualConfig, repoCharlieExpectedConfigWithoutIgnoreStandalone);
+        TestUtil.compareRepoConfig(repoDeltaActualConfig, repoDeltaExpectedConfig);
+
+        repoCharlieActualConfig = repoConfigsWithAllIgnoringStandaloneConfig.get(0);
+        repoDeltaActualConfig = repoConfigsWithAllIgnoringStandaloneConfig.get(1);
+        TestUtil.compareRepoConfig(repoCharlieActualConfig, repoCharlieExpectedConfigWithIgnoreStandalone);
+        TestUtil.compareRepoConfig(repoDeltaActualConfig, repoDeltaExpectedConfig);
```
jylee-git: For comparison tests, expected should come before actual as per the current convention. So let's standardize it

```java
TestUtil.compareRepoConfig(EXPECTED, ACTUAL);
```

src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -185,6 +188,48 @@ public void repoConfig_ignoresStandaloneConfigInCli_success()
...
+
+        CliArguments argsWithoutIgnoreStandaloneFlag = ArgsParser.parse(
+            translateCommandline(inputWithoutIgnoreStandaloneFlag));
+        CliArguments argsWithIgnoreStandaloneFlag = ArgsParser.parse(
+                translateCommandline(inputWithIgnoreStandaloneFlag));
+
+        List<RepoConfiguration> repoConfigsWithSomeIgnoringStandaloneConfig =
+                RepoSense.getRepoConfigurations((ConfigCliArguments) argsWithoutIgnoreStandaloneFlag);
+        List<RepoConfiguration> repoConfigsWithAllIgnoringStandaloneConfig =
+                RepoSense.getRepoConfigurations((ConfigCliArguments) argsWithIgnoreStandaloneFlag);
```
jylee-git: This test case as a whole is confusing as it seems to me that you are combining multiple two sets of test cases into one.

Can we split this TC be split into two? i.e.:
1. RepoConfigs without the CLI ignore standalone config flag --> the column in the CSV file should not be "overriden"
2. RepoConfigs with the CLI ignore standalone config flag --> the flag should take priority and "override" the column in CSV file

anubh-v: This makes sense. Thanks for the suggestion @jylee-git 
There already exists a test case for parsing a CSV **without the CLI ignore standalone config flag**. 

However, the CSV used for this test case only has a single repo config.
I think this test CSV should have 2 repo configs - with one repo config ignoring the standalone, and the other not ignoring.

src/test/java/reposense/model/RepoConfigurationTest.java
anubh-v: Done

anubh-v:
@jylee-git, @fzdy1914, I've made the following changes

- Update user guide
- Split my test case into 2 test cases
  (1) `repoConfig_ignoreStandaloneConfigInCli_overrideCsv` tests the case where we parse `repo-config.csv` with the `--ignore-standalone-config` command line flag
  (2) `repoConfig_withoutIgnoreStandaloneConfigInCli_useCsv` tests the case where we parse `repo-config.csv` without the `--ignore-standalone-config` command line flag

- Modified the implementation so that the main logic for this PR is executed in `RepoConfiguration#setStandaloneConfigIgnoredToRepoConfigs` rather than in `RepoSense.java`

anubh-v:
@jamessspanggg, could I get your thoughts on these changes? Thanks very much for you time.

anubh-v:
Thanks very much for your time @jamessspanggg, @fzdy1914, @jylee-git

anubh-v:
@yong24s, may I get your review for this PR? Thanks very much

anubh-v:
@eugenepeh, may I get your review on this PR? Are there any further changes I should make?

eugenepeh:
>@eugenepeh, may I get your review on this PR? Are there any further changes I should make?

In the event there are enough approvals from reviewers, @reposense/admins feel free to make the decision to do a last call and/or merge.

PR #932, opened 2019-11-01
[#931]Code view: Default sorting in descending order

### LINKED ISSUE #931: Code view: default sorting should be LoC in descending order
Current: the default order is ascending
Suggestion: showing bigger changes on top make more sense as they are the more interesting ones

image

Tejas2805:
Fixes #931 

```
Currently, in Code View, the default order is ascending.

Let's make the default order descending as it is more meaningful to have
bigger changes on top as they are the more interesting changes.
```


damithc:
Let's try to merge this soon as it can be useful for CS2103 PE

damithc:
If the files are sorted in descending order, the `reverse` tick box must be ticked by default?

jylee-git:
image

The checkbox should be checked by default if we are implementing this PR; else it won't make sense here.

Tejas2805:
> If the files are sorted in descending order, the `reverse` tick box must be ticked by default?

I was thinking the same but then I thought its better to not have it ticked by default as it will be a bit weird. 

We are thinking from this point of view cause we know that ascending was the default before and descending is reverse. But for someone who is using it for the first time, descending will be the default and ascending will become reverse. So I think it's okay to not have it ticked from before.

@damithc Let me know if you still think that the box should be ticked by default is how we should proceed ahead. I will do the change.

damithc:
> We are thinking from this point of view cause we know that ascending was the default before and descending is reverse. But for someone who is using it for the first time, descending will be the default and ascending will become reverse. So I think it's okay to not have it ticked from before.

That is a good point.  I think the better solution is to change it to a drop-down with options `Ascending` and `Descending`?

Tejas2805:
> > We are thinking from this point of view cause we know that ascending was the default before and descending is reverse. But for someone who is using it for the first time, descending will be the default and ascending will become reverse. So I think it's okay to not have it ticked from before.
> 
> That is a good point. I think the better solution is to change it to a drop-down with options `Ascending` and `Descending`?

Can do this, with the default being `Descending`.

Tejas2805:
@damithc Done.

damithc:
> @damithc Done.

I'm OK with the UI. Can proceed to dev reviews.

frontend/src/static/js/v_authorship.js
```diff
@@ -39,7 +39,7 @@ window.vAuthorship = {
       totalLineCount: '',
       totalBlankLineCount: '',
       filesSortType: 'lineOfCode',
-      toReverseSortFiles: false,
+      toReverseSortFiles: 'descending',
```
jylee-git: Maybe rename this variable name to `sortByOrder`.

Tejas2805: Yeah makes sense. Will do the needful.

Tejas2805:
> Let's try to merge this soon as it can be useful for CS2103 PE

@eugenepeh Could you please review the PR? The Professor thinks this might be useful for the CS2103 PE which is on Friday, so if all looks okay, we can also merge.

frontend/src/static/js/v_authorship.js
```diff
@@ -210,7 +211,7 @@ window.vAuthorship = {
     },
 
     sortFiles() {
-      this.sortingFunction = (a, b) => (this.toReverseSortFiles ? -1 : 1)
+      this.sortingFunction = (a, b) => (this.sortByOrder === 'descending' ? -1 : 1)
```
eugenepeh: all these seems like unnecessary change when u can just bind the value true or false to `toReverseSortFiles`;
changing a binary value option to string compare is opening up spaces for bugs like typos/errors

Tejas2805: The `value` property of an HTML `option` element can only be a string. 

So if we want to do something like this - I will have to change the dropdown back to a checkbox. Let me know how you want me to proceed.

eugenepeh: Do look up `v-bind:`, take note of the syntax that we can utilize.

Tejas2805: Just checked. I have accordingly made the changes. Thanks for guiding

frontend/src/static/js/v_authorship.js
```diff
@@ -199,6 +199,7 @@ window.vAuthorship = {
       this.isLoaded = true;
 
       this.activeFilesCount = this.selectedFiles.length;
+      this.sortFiles();
```
eugenepeh: this should be able to be removed since the value of toReverseSortFiles is watched

Tejas2805: We cannot remove it because even if `toReverseSortFiles` is watched, there is nothing to activate it. So, the default sorting is still ascending order, which defeats the purpose of the PR. Please refer to the screenshot below.

Screen Shot 2019-11-20 at 8 12 55 PM

eugenepeh: @Tejas2805 I am going to let you figure this out yourself so that maybe you will understand how vue works better this way.

Think about why there isn't a need to add the function there in the first place, yet the files will came out in an arranged order.

Tejas2805: I know that cause `toReverseSortFiles` is watched, whenever we change the sorting order, it will automatically cause the files to sort themselves, based on the value in `toReverseSortFiles` since its being `watched` but won't we have to activate it the first time to sort the files in required order? I tried running the program by removing `sortFiles()`  and the files came in ascending order with `order` as Descending.
So I am confused as to whether you are saying that we shouldn't need `sortFiles()` at all, or should I call the function somewhere else.

eugenepeh: >So I am confused as to whether you are saying that we shouldn't need sortFiles() at all, or should I call the function somewhere else.

Yes, you dont need sortFiles at all, you just need to change some code somewhere to make the default descending.

Tejas2805: @eugenepeh I have made the change. Wished to understand something- Is this way preferred cause this reduces the call stack?

fzdy1914: The original implementation will increase one additional unnecessary event loop of Vue, which is quite costly if there are multiple files.

PR #933, opened 2019-11-01
[#871]  CSV files: Ignore blank lines before the header (alternative implementation)

### LINKED ISSUE #871: csv files: ignore blank lines

anubh-v:
Fixes #871 

This PR contains an alternative solution to PR #903.

In PR #903, a `BufferedReader` is used to skip all the blank lines before the header.
In this PR, the header is defined within the code and is not inferred from the CSV file.
This means that all blank lines are considered malformed and they hence get ignored.
 

```
CsvParser: add logic to ignore blank lines before the header

When CSV config files are parsed, the CsvParser currently considers
the first non-empty line as the header.

Hence, CSV files with blank lines before the header are parsed
incorrectly. The blank line is considered the header, which
causes all rows in the CSV to be flagged as malformed.

Let's modify the parsing process so that the headers for the various
config files are defined within our code, rather than inferred from
the csv config.

This will ensure that leading blank lines in the csv configs
are not considered as headers.

```


Tejas2805:
@anubh-v In case, you ever have two alternatives in mind, please try to do either of the following: 

1) Suggest both the alternatives first and let the senior developers decide which method will be the best to proceed with.
2) Do a PR with one of the alternatives, stating the reason why you chose to go ahead with this alternative and not the other one. 

Creating two PR's for the same issue makes unnecessary clutter which can be avoided.

anubh-v:
Thanks for the feedback @Tejas2805 
The idea for this alternative PR came about during the discussion of my original solution.
I opened this PR as i felt it may be helpful for the team to compare the implementation of the 2 approaches, before deciding.

I'll keep your advice in mind in future.

damithc:
Two PRs for the same issue can be wasteful, but it is OK to do that when there is a need for comparison and it's easier the create two PRs than try to describe the difference in words.

anubh-v:
Closing this as we decide to adopt the alternative implementation where CSV header is inferred from the user supplied CSV, rather than defining the CSV header in the code.

This approach makes it simpler to detect specific problems with the user provided CSV.

Issue #912, opened 2019-10-07
Developer guide: Fix broken links in table of contents, and add useful links


anubh-v:
The links outlined in the image below are broken:

![toc](https://user-images.githubusercontent.com/35621759/66294545-cdf0a600-e91b-11e9-9ba3-bd326b589bc1.png)

These links can be replaced by a single link to the section shown below:

![jss](https://user-images.githubusercontent.com/35621759/66294435-8f5aeb80-e91b-11e9-9c2c-998fa77bb19e.png)

In addition, we could add links to the "Before writing code" and "Building and running RepoSense from code" sections, to make the content more obvious to new readers.

![newlinks](https://user-images.githubusercontent.com/35621759/66294805-70108e00-e91c-11e9-8c59-d5481b8978f7.png)

PR #935, opened 2019-11-01
[#934] CsvParser: Remove extra comma in warning logs

### LINKED ISSUE #934: CsvParser: Remove extra comma in warning logs
To recreate this issue:

1. Save the following into a `repo-config.csv` file in a new directory.

```
Repository Location,Branch,File formats,Ignore Glob List,Ignore standalone config, Ignore commits list
https://github.com/reposense/RepoSense.git, ,md, ,,
                 
            
```

Note that **there are 2 blank rows after the first record**.

2. Run `java -jar RepoSense.jar --config <path to directory containing repo-config.csv>`

Within the output logs, we will see

```
11:43:28 - [WARNING] Line 3 in CSV file, repo-config.csv, is malformed.
Content: ,
11:43:28 - [WARNING] Line 4 in CSV file, repo-config.csv, is malformed.
Content: ,

... < rest of logs > ...
```

There is an extra comma in `Content: , `
This can be fixed by modifying the `getRowContentAsRawString`method in the `CsvParser` class.

We should fix this, as the extra comma may make debugging more confusing.

anubh-v:
Fixes #934 

```
CsvParser: Remove extra comma from warning logs

When a line in a CSV config is found to be malformed, the logger
prints the contents of the line in a warning message.
The current method causes an extra comma to be printed.

The extra comma may make debugging more confusing.

Let's modify the method used to extract the contents of a line,
such that there are no extra commas.
```


anubh-v:
@Tejas2805 (or any other available team member), could I seek your comments on this PR?
Thanks for your time!

anubh-v:
@eugenepeh, for this PR, I have modified `getRowContentAsRawString` method in the `CsvParser` class?

This method is only used to convert a `CSVRecord` into a String, so that it can be shown in the logs.

Would it be beneficial to test `getRowContentAsRawString` using reflection? 
I think it might be too trivial to warrant a test.

Tejas2805:
> @Tejas2805 (or any other available team member), could I seek your comments on this PR?
> Thanks for your time!

@anubh-v I am a developer who is learning too. So I can't review but I have requested a review for you.

anubh-v:
The `toString` method for `CSVRecord` does not give the same String as our implementation of `getRowContentAsRawString`

For example, if our CSV file is as follows:

```
Buyer,Seller,Licenser
Tom,Bob,Smith                                                
Sam,Bob,Tim
```

then applying `toString` on the 1st `CSVRecord` gives:
```
CSVRecord [comment=null, mapping={Buyer=0, Seller=1, Licenser=2}, recordNumber=1, values=[Tom, Bob, Smith]]
```

Applying `getRowContentAsRawString` gives:
`Tom,Bob,Smith`

From the `CSVRecord` [documentation](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVRecord.html), I don't see a way of obtaining similar behaviour to `getRowContentAsRawString`.

We can obtain an `Iterator<String>` from the `CSVRecord` using `iterator()`, but I think we cannot assume anything about how `toString()` behaves for a `Iterator<String>` ?

anubh-v:
@eugenepeh, thanks very much for your time and feedback.

> Yes, but at the same time, my original question is actually asking whether have you explored this possibility in the first place.

Ok, I will explore whether using `toString()` on the `Iterator<String>` obtained from the `CSVRecord` can give us a good string representation of the record.

Let me clarify why I did **not** look into this earlier. I felt that since the `Iterator` type does not give any guarantees for its `toString` method, we should not write code that assumes `toString()` will always work a certain way for the `Iterator<String>` obtained from `CSVRecord`.

> Also, just to clarify regarding the issue that you have posted #912, are empty lines suppose to appear as a record? This question is related to your work on #903

Sorry, but issue #912 is not related to this PR right? It is related to the Developer Guide.
Are you referring to issue #934 ?

eugenepeh:
@anubh-v 

>Just to clarify, I did try this suggestion. It does not give us a good string representation of the record, as shown in my earlier comment.

Good to know that you've tried it out. However, try to keep it a practice to document things that you've explored as well. It will help future developers to understand the problem, constraints etc.

>Sorry, but issue #912 is not related to this PR right? It is related to the Developer Guide.
Are you referring to issue #934 ?

Yes, I have accidentally tagged the wrong issue. Sorry about that.

anubh-v:
> Good to know that you've tried it out. However, try to keep it a practice to document things that you've explored as well. It will help future developers to understand the problem, constraints etc.

Sorry @eugenepeh, but there appears to be some confusion. I *did* document my findings in my very first reply. Here is an excerpt from my reply ([direct link](https://github.com/reposense/RepoSense/pull/935#issuecomment-552048538)).

> The `toString` method for `CSVRecord` does not give the same String as our implementation of `getRowContentAsRawString`
> 
> For example, if our CSV file is as follows:
> 
> ```
> Buyer,Seller,Licenser
> Tom,Bob,Smith                                                
> Sam,Bob,Tim
> ```
> 
> then applying `toString` on the 1st `CSVRecord` gives:
> 
> ```
> CSVRecord [comment=null, mapping={Buyer=0, Seller=1, Licenser=2}, recordNumber=1, values=[Tom, Bob, Smith]]
> ```
> 
> Applying `getRowContentAsRawString` gives:
> `Tom,Bob,Smith`

Perhaps I did not communicate my finding clearly.
Let me clarify things by quickly summarising the discussion so far.

#### Discussion point 1: Can the CSVRecord's `toString` method be used instead of the existing `getRowContentAsRawString` method?

You asked me to check whether we could simply use the `toString` method of `CSVRecord` to get a String representation of the record. This could allow us to entirely remove the `getRowContentAsRawString` method.

I tested the `toString` method for `CSVRecord` and found that it does **not** give a suitable String representataion of the record. i.e. it is not equivalent to `getRowContentAsRawString`.
I documented this finding in my first reply (also included in **this** comment).

#### Discussion point 2: Can the CSVRecord's `iterator` method be used to get a `String` representation of the record?

While reading the CSVRecord documentation (available [here](https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVRecord.html)), I found that the `CSVRecord` has an `iterator()` method that returns a `Iterator<String>` containing the values of the record. 

This *appeared promising*, but I cannot find a method of easily concatenating all the Strings in the `Iterator<String>`, **without using something like the** `getRowContentAsRawString` **method**.

I have tried applying `toString` on the `Iterator<String>`, as shown in the code snippet below.

```
Reader reader = new FileReader(...path to file...);
records = CSVFormat.DEFAULT.withHeader().parse(reader);
for (CSVRecord record : records) {
    System.out.println(record.iterator());
    // Also tried the equivalent: System.out.println(record.iterator().toString());
}
```

The output simply shows some information about the underlying type.
```
java.util.Arrays$ArrayItr@3e3abc88
java.util.Arrays$ArrayItr@7229724f
java.util.Arrays$ArrayItr@4c873330
```

My conclusion is that there is no simpler alternative to our `getRowContentAsRawString` method, and hence this method should be kept.

anubh-v:
Let me also respond to the question on whether empty lines can appear as a record.
> Also, just to clarify regarding the issue that you have posted #934 , are empty lines suppose to appear as a record? This question is related to your work on #903

Let me use the following definitions:

1. Empty line - Line that has only the newline character.
2. Blank line - Line that has some non-zero number of whitespace characters, followed by a newline character.

In both our original `CsvParser`  and the updated `CsvParser` in PR #903, empty lines do not show up as records regardless of their position in the CSV file. They are ignored by the parser.

In our original `CsvParser`, 
  - if there is a blank line before the CSV header, then RepoSense will crash.
  - if there are no blank lines before the CSV header, then any blank lines after the header will still be parsed as CSVRecords, but they will be filtered out as malformed records because they do not have as many columns as the header.

In the updated `CsvParser` in PR #903, 
  - if there are blank lines before the CSV header, then RepoSense will ignore them. They do not get parsed as records.
  - if there are blank lines after the CSV header, they will still be parsed as CSVRecords, but they will be filtered out as malformed records because they do not have as many columns as the header.

When working on PR #903, I tried various combinations of the methods in the `CSVFormat` API. I found that there is no "out of the box" method to ignore all the blank lines in the CSV, regardless of their position.

anubh-v:
@eugenepeh, I've addressed your questions in the comments above.
My main conclusion is that the `getRowContentAsRawString` method in `CsvParser` cannot be replaced by `toString` or any "out of the box" functionality in the Apache Commons CSV library.

Please let me know if you have further clarifications.

anubh-v:
> Do help me to put up an issue regarding the problem I highlighted earlier:
> 
> > this is not exact representation of the original entry in the csv file

Sorry, @eugenepeh, I am a little confused. Isn't this exactly the issue we're solving in *this* PR?

eugenepeh:
>Sorry, @eugenepeh, I am a little confused. Isn't this exactly the issue we're solving in this PR?

iirc, what we are getting back now is a "reconstructed" version of the entry in the csv file, it is not the exact string at that was read from the file. E.g. if a column is surrounded by double quotes, the quotes wont be printed out.

anubh-v:
> iirc, what we are getting back now is a "reconstructed" version of the entry in the csv file, it is not the exact string at that was read from the file. E.g. if a column is surrounded by double quotes, the quotes wont be printed out.

Thanks for clarifying @eugenepeh. I've raised a separate issue for this.

anubh-v:
@jylee-git may I seek your help to provide the second approval for this PR? Thanks very much!

anubh-v:
Hi everyone, would it be possible for me to get a further review on this PR?
@yong24s, @jamessspanggg, @jylee-git , I would really appreciate your time on this. Thanks very much for your work.

anubh-v:
Thanks very much @fzdy1914 

@eugenepeh, does this PR need a third approval?

eugenepeh:
>@eugenepeh, does this PR need a third review?

@anubh-v not necessary for small PR. 
Just one last thing to highlight: can you make use of the `setEmptyValue(...)` method of `StringJoiner` to set an empty value so that we can better illustrate an empty record? E.g. `[EMPTY LINE IN CSV]` etc
If possible, verify that it works with white spaces as well.

anubh-v:
@eugenepeh, I tried the `setEmptyValue` method of `StringJoiner`.
It does not fit out requirements because the empty value is discarded even when we add an empty string to the `StringJoiner`.

I've added a simple piece of logic to represent all empty / blank lines as `[EMPTY LINE]`.
It will look like this:
image


Is this ok?

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -154,11 +155,16 @@ protected boolean isElementOverridingStandaloneConfig(final CSVRecord record, in
      * Returns the contents of {@code record} as a raw string.
      */
     private String getRowContentAsRawString(final CSVRecord record) {
-        StringBuilder inputRowString = new StringBuilder();
-        for (int colNum = 0; colNum < record.size(); colNum++) {
-            inputRowString.append(get(record, colNum)).append(",");
+        String emptyLineMessage = "[EMPTY LINE]";
```
eugenepeh: sounds good to me, this can be a constant though

anubh-v: Done. I didn't do this initially as I felt it may clutter the class level constants.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -27,6 +27,7 @@
     protected static final Logger logger = LogsManager.getLogger(CsvParser.class);
 
     private static final String OVERRIDE_KEYWORD = "override:";
+    private static final String EMPTY_LINE_STRING_REPRESENTATION = "[EMPTY LINE]";
```
eugenepeh: since its going to be displayed as a message, can just name it `MESSAGE_EMPTY_LINE`

anubh-v: Done.

anubh-v:
@eugenepeh, I've resolved the merge conflict for this PR. No other pending changes.

anubh-v:
@eugenepeh, I've addressed all the outstanding issues for this PR.
Is there any other improvement this PR needs?

Issue #947, opened 2019-11-20
Inconsistency color of repo names in authorship and zoom view


Tejas2805:
Both the zoom view and authorship view highlight the repo name in different colours.

Let's make them the same for consistency. 

![Screen Shot 2019-11-20 at 2 08 51 PM](https://user-images.githubusercontent.com/35946746/69213652-adb13980-0b9f-11ea-8840-45c60c6f6cac.png)

![Screen Shot 2019-11-20 at 2 09 02 PM](https://user-images.githubusercontent.com/35946746/69213658-b0139380-0b9f-11ea-9953-4c70e65e3b61.png)

Tejas2805:
Or have we kept it different to show the difference between the two views? In that case, maybe we can give some kind of header to make it easy to differentiate between the two views.

damithc:
> Or have we kept it different to show the difference between the two views?

I don't think so.

>In that case, maybe we can give some kind of header to make it easy to differentiate between the two views.

Yes, we can do that.

jamessspanggg:
Closing this as fixed by #948

PR #948, opened 2019-11-20
[#947]Synchronising different tab views

Tejas2805:
Fixes part of #947 
* [x] Add view header to differentiate between two views #947 
* [x] Add link to zoom view title to make it consistent with authorship view #947 

```
It's difficult to distinguish between the authorship and zoom 
views as there is no easy way to identify which is which. 

Let's add a header to both the views for easy identification. 
Let's add a link to the zoom title to make it more consistent 
with authorship
```


jamessspanggg:
@damithc need your input for the UI.

Authorship View:
Screen Shot 2019-11-22 at 9 58 18 PM

Zoom View:
Screen Shot 2019-11-22 at 9 58 24 PM

frontend/src/tabs/zoom.pug
```diff
@@ -1,6 +1,7 @@
 #zoom
+  span.large-font Zoom View
   .zoom__title
-    .zoom_title--group.large-font
+    a.zoom_title--group.large-font
```
jamessspanggg: The link is missing from the title.

damithc:
`Code` and `Commits`? `View` is redundant.

frontend/src/tabs/zoom.pug
Tejas2805: This was something I wanted to confirm too - The original zoom title doesn't have a link, so I left it like that. Should I add the link?

Tejas2805:
> `Code` and `Commits`? `View` is redundant.

Maybe `Code Panel` and `Commits Panel` to make it consistent with UG.

frontend/src/tabs/zoom.pug
jamessspanggg: Yes I think you should. There's no point leaving a link that directs to nowhere. It also standardises between the two views.

Tejas2805:
@jamessspanggg I have added the link but I feel maybe it's better to avoid it (lot of changes to make the link work) and we anyway didn't have it originally, so I guess we can avoid it. But for consistency, in the two views, we might need to keep it. Let me know, what you think about it.

jamessspanggg:
> @jamessspanggg I have added the link but I feel maybe it's better to avoid it (lot of changes to make the link work) and we anyway didn't have it originally, so I guess we can avoid it.

Yes to think about it, it may be better to avoid it as the title of `Commits Panel` is slightly different compared to the `Code Panel` when switching between `Group By`. Thanks for the reminder.

frontend/src/tabs/zoom.pug
```diff
@@ -1,8 +1,13 @@
...
   .zoom__title
     .zoom_title--group.large-font
-      span(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.displayName }} ({{ info.user.name }})
-      span(v-else) {{ info.user.repoName }}
+       span(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.displayName }} ({{ info.user.name }})
+       a(
+       v-bind:href="info.location", target="_blank",
+       v-bind:title="'Click to open the repo'"
+       )
+         span(v-if="info.filterGroupSelection !== 'groupByAuthors' ") {{ info.user.repoName }}
```
fzdy1914: `v-if` should be put inside `<a>`, otherwise, there may be a link inside HTML with nothing inside, which is a bad practice. Do note the indentation as well.
image
Suggestion: use `v-else` for simplesity:
```
span(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.displayName }} ({{ info.user.name }})
a(
  v-else,
  v-bind:href="info.location", target="_blank",
  v-bind:title="'Click to open the repo'"
)
  span {{ info.user.repoName }}
```

Tejas2805: Would you prefer if I keep the link or remove it?

fzdy1914: You may just use my edited version.

fzdy1914: > **EDIT 1:** Is it that the link will be created irrespective of the selected choice and that is a bad practice? 

Correct. It is not a good practice because it is leaving extra information that is not used in the HTML. Although it is not reflected in the website, others can find the link easily in the raw HTML.
 
> Though it kind of makes sense as an empty link if found can be used for attacks

Correct. Of course, this link you leave here will no cause any security issue. But what if you are building a secure website next time? As a high-standard SE repo, it is nice if we can always keep the high standard in mind.

frontend/src/tabs/zoom.pug
```diff
@@ -1,8 +1,14 @@
 #zoom
+  span.large-font Commits Panel
   .zoom__title
     .zoom_title--group.large-font
-      span(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.displayName }} ({{ info.user.name }})
-      span(v-else) {{ info.user.repoName }}
+       span(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.displayName }} ({{ info.user.name }})
```
fzdy1914: It seems that you forgot my comment on the indentation. We use 2 space indentation for `pug` files.

frontend/src/tabs/zoom.pug
```diff
@@ -1,8 +1,14 @@
...
+       a(
+       v-else,
```
fzdy1914: Meanwhile, it is always nice if you can reference how the indentation goes in other part of this file.
For example:
the line after `a(` should be extra indented with 2 space because it is the inside code of this `<a>`.

frontend/src/tabs/zoom.pug
Tejas2805: Ah sorry about that. Seemed to have missed it

frontend/src/tabs/zoom.pug
Tejas2805: Got it. Thanks for explaining.

Tejas2805:
@fzdy1914 for your review

fzdy1914:
> @fzdy1914 for your review

Since you are a member of RepoSense now, you may just re-request a review me in the side panel and set the label of PR to the correct state.

frontend/src/static/js/v_summary.js
```diff
@@ -681,26 +681,27 @@ window.vSummary = {
       });
     },
 
-    openTabZoomSubrange(userOrig) {
+    openTabZoomSubrange(userOrig, repo, index) {
```
eugenepeh: Since these lines are modified, can you help me to rename `userOrig` to `user` so that it is consistent with how authorship parameters were named?

frontend/src/static/css/style.scss
```diff
@@ -29,7 +29,7 @@ a {
 
 .loader {
   align-items: center;
-  background: #FFF;
+  background: mui-color('white');
```
eugenepeh: irrelevant to this PR, and should be removed.

frontend/src/static/js/v_summary.js
Tejas2805: Okay will.

frontend/src/static/css/style.scss
Tejas2805: I agree that it is not specific to this PR but this was such a minute change that I didn't wish to create a new PR for this. So are you suggesting: 
1) This isn't an issue and doesn't need to be fixed
2) Do a different PR for fixing this 
3) Include this in a PR which affects css styles

eugenepeh: Yes, `This isn't an issue and doesn't need to be fixed`, or at least not necessary unless we make direct changes to the file.

frontend/src/static/js/v_summary.js
eugenepeh: >If I rename this, there is a const user in the function which I will have to rename too and correspondingly change it in zoom.pug too. Let me know how you want me to go about it.

Can go ahead, if it doesn't affect any functionality or readability

Tejas2805: @eugenepeh @yong24s @fzdy1914 Could you all recommend the new name for `const user`? As of now, I have interchanged `user` and `userOrig`

fzdy1914: Just a question, does `userOrig` stands for `original user`? I am unsure about this `Orig`.

damithc: > Just a question, does `userOrig` stands for `original user`? I am unsure about this `Orig`.

This question is a hint that the variable name can be improved. Avoid abbreviations as much as possible.

eugenepeh: >@eugenepeh @yong24s @fzdy1914 Could you all recommend the new name for const user? As of now, I have interchanged user and userOrig

I don't think `original user` has any meaning behind it, probably just the way the developer named it, hence should be renamed to `user`.

jamessspanggg: I think the previous developer named it as `userOrig` because he creates a shallow clone of the original user using `Object.assign({}, userOrig)`. So this is to differentiate between the original user (which should not be modified) and the cloned one (which we will be modified for the zoom tab functionality).

eugenepeh: >I think the previous developer named it as userOrig because he creates a shallow clone of the original user using Object.assign({}, userOrig). So this is to differentiate between the original user (which should not be modified) and the cloned one (which will be modified for the zoom tab functionality).

I see, in this case, probably should named the `userOrig` to `sourceUser` or named the cloned user to `clonedUser`. The latter is preferred because the variable names shouldn't be brought across functions.

Tejas2805: Okay! Moving forward by naming it as `clonedUser`

frontend/src/static/js/v_summary.js
```diff
@@ -681,26 +681,27 @@ window.vSummary = {
...
-        this.openTabZoom(userOrig, tsince, tuntil);
+        this.openTabZoom(user, tsince, tuntil, repo, index);
       }
     },
 
-    openTabZoom(userOrig, since, until) {
+    openTabZoom(user, since, until, repo, index) {
       const { avgCommitSize } = this;
-      const user = Object.assign({}, userOrig);
+      const clonedUser = Object.assign({}, user);
```
fzdy1914: On a second thought, I think the implementation can be further improved.

The idea of this cloned user is to modify the commits within the user without affecting the original users. However, as pointed out by #971, the implementation is not ideal as it is still directly modifying the `property` field of the child component.  To fix it, we can make the `clonedUser` as a computed field inside `v_zoom` as how it is done in #971.

https://github.com/reposense/RepoSense/pull/971/files#diff-64d5324a6164c3a10fbca7a7508041d7L18-L26

So, there are two choices now. fix this issue inside this PR as we are already touching it or fix it inside #971 first as it is more related.

fzdy1914: @eugenepeh Your thought?

Tejas2805: I think its better if we let #971 handle it, as it actually works on this aspect of the issue. We can merge this one, after it.

eugenepeh: Fixing in #971 would be better in that case, thanks @fzdy1914

fzdy1914:
Can you update your proposed commit message to get it merged?

Tejas2805:
@fzdy1914 Done!

PR #949, opened 2019-11-21
[#944] Implement authorship analysis

### LINKED ISSUE #944: Improve code authorship attribution
Currently in the code panel, an author is credited for lines of code that was last modified them. However, another author might have written the line initially and the current author (author who last modified the line) might have only made a slight modification to the line. In such a case, the current author gets credited for work that is not entirely done by them.

Proposed solution:
Assign partial or full credit of a line to the current author based on whether they introduced the line or someone else introduced the line and they only made a slight modification to it. This is done by checking the ancestor line (previous version of the line) to see how similar it is to the line. Users add a command line flag to indicate if they want this feature to be turned on - it is turned off by default.

For a given line **l** and author **a** who last modified the line, we can assign credit to author **a** using the following steps:
1) Check the commit in which author **a** modified line **l**
2) Get the diffs of the commit and find the hunk which line **l** belongs to
3) Go through all the deleted lines in the hunk and compare them with line **l** to see how similar they are (using Levenshtein string distance metric)
4) If we do not find a deleted line with high similarity with line **l**, we give author **a** full credit. If we find such a deleted line, we find out the author of the deleted line.
5) If the author of the deleted line is not author **a**, we give partial credit to author **a**. If the author of the deleted line is author **a**, we set line **l** as the deleted line and repeat from step 1.

After we assign partial or full credit to the author, we can display the results in the code panel by using a darker shade of green to highlight code with full credit.

Picture1

myteo:
Fixes #944.

```
A line is credited to the author who last modified it.

Another author might have written the line initially and the current
author only modified it slightly. In such a case, the current author
gets credited for work that is not entirely done by them.

Lets analyze how similar a line is to its ancestor line (previous
version of the line) and give partial or full credit to the author
based on the analysis. This feature is turned off by default as
it takes quite long to run.
```


myteo:
Please let me know if you need any clarifications as this is quite a big PR 😬

eugenepeh:
>Please let me know if you need any clarifications as this is quite a big PR 😬

Indeed a big PR that is few hundred lines long. Can you give a brief technical implementation summary on the concept? If possible, would be preferable if it can break down to few smaller PRs.

myteo:
> Indeed a big PR that is few hundred lines long. Can you give a brief technical implementation summary on the concept?

I have added some descriptions in the issue (#944). Please let me know if any further clarification is needed.

> If possible, would be preferable if it can break down to few smaller PRs.

I will create a feature branch and create a smaller PRs towards the feature branch. Is that ok?

eugenepeh:
>I will create a feature branch and create a smaller PRs towards the feature branch. Is that ok?

Sorry for late reply, yes that would be preferred.

fzdy1914:
Just a question, shouldn't the PR targets to a feature branch? not master?

myteo:
> Just a question, shouldn't the PR targets to a feature branch? not master?

Yes it should be feature branch. I am still working on splitting up the PR, will try to get that out when I have more time.

fzdy1914:

PR #955, opened 2019-12-03
[#872] CsvParser: improve exception handling when a csv file has incorrect number of columns in header

### LINKED ISSUE #872: ConfigCsvFiles: Improve exception handling when the csv file(s) is/are malformed

anubh-v:
Fixes #872 

```
CsvParser: Improve exception handling when CSV has invalid header size

RepoSense crashes abruptly when any of the CSV config files
do not have the expected number of columns in their headers.

We should catch this error early and provide the user with a
clearer error message.

Let's add a header validation step to the CSV parsing process.
This step will verify if the CSV file's header has the right
number of columns.
```


anubh-v:
@eugenepeh, @jylee-git, may I get your thoughts on this PR?

There is one limitation in the current solution.
The `parse` method in `CsvParser` will throw an `IOException` when the CSV file's header has the wrong number of columns.

However, the main code in `RepoSense.java` will silently ignore the `IOException` for the `author-config.csv` and `group-config.csv` files.

Ideally, the users should be alerted that the `author-config.csv` and `group-config.csv` are malformed, so they can fix the files and re-run the analysis.

Proposed fix: Throw `InvalidCsvException` out from `parse` in `CsvParser`.

I did not make this fix yet because it will affect many test files and significantly increase the size of the PR. 

I can make this change once the core logic of this PR is approved, or perhaps in another PR.

anubh-v:
Hi everyone, could I get any team member's help to set the "toReview" label for this PR?
I don't have the permissions for that. Thanks!

eugenepeh:
@anubh-v I have sent you the invite to become repository member, let me know if you didn't receive it.

anubh-v:
@eugenepeh, thanks for the invite.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -195,6 +200,20 @@ private String getRowContentAsRawString(final CSVRecord record) {
         return contentAsString;
     }
 
+    /**
+     * Checks if a given header has the right number of columns.
+     * @throws InvalidCsvException if the header does not have as many columns as expected.
+     */
+    private void validateHeader(String[] possibleHeader) throws InvalidCsvException {
```
jylee-git: For the javadocs, let's include the input argument in the docs wherever possible.

`Checks if {@code possibleHeader} contains the expected number of columns. ...`

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -207,4 +226,9 @@ private String getRowContentAsRawString(final CSVRecord record) {
      * {@code record} and add created objects into {@code results}.
      */
     protected abstract void processLine(List<T> results, final CSVRecord record) throws ParseException;
+
+    /**
+     * Gets the number of columns expected in the header.
```
jylee-git: Better starting word to use  here (same for those below) is `Returns the number of expected cols ...`

src/test/java/reposense/parser/CsvParserTest.java
```diff
@@ -229,6 +231,12 @@ public void authorConfig_invalidLocation_success() throws IOException {
         Assert.assertEquals(3, config.getAuthorList().size());
     }
 
+    @Test (expected = IOException.class)
+    public void authorConfig_invalidHeaderSize_throwsIoException() throws IOException {
+        AuthorConfigCsvParser authorConfigCsvParser = new AuthorConfigCsvParser(AUTHOR_CONFIG_INVALID_HEADER_SIZE);
+        authorConfigCsvParser.parse();
+    }
```
jylee-git: Do the same set of test cases for group and repo config  as well

src/main/java/reposense/parser/GroupConfigCsvParser.java
```diff
@@ -79,4 +84,12 @@ private static GroupConfiguration findMatchingGroupConfiguration(
         results.add(config);
         return config;
     }
+
+    /**
+     * Gets the number of columns expected in the header.
```
jylee-git: Be more specific in which header are we referring to here (in this case the group config header). Same for the other two

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -207,4 +226,9 @@ private String getRowContentAsRawString(final CSVRecord record) {
...
+     */
+    protected abstract int getHeaderSize();
```
fzdy1914: Since `HEADER_SIZE` is a feature that all its child have, instead of an abstract method, is it better to make it a protected attribute and the child only need to overwrite this attribute and no need to overwite the method again and again?

fzdy1914:
> Hi everyone, could I get any team member's help to set the "toReview" label for this PR?
> I don't have the permissions for that. Thanks!

For PR reviewing process, please only ask `stage1-reviewer` to review the PR first instead of both `stage1-reviewer` and `stage2-reviewer`.

anubh-v:
Thanks very much for your time @jylee-git, @fzdy1914. 
I will respond to the feedback by tonight.

> For PR reviewing process, please only ask stage1-reviewer to review the PR first instead of both stage1-reviewer and stage2-reviewer.

My apologies for this. I have corrected this for my other open PR.

src/main/java/reposense/parser/CsvParser.java
anubh-v: Done

src/test/java/reposense/parser/CsvParserTest.java
anubh-v: Done

src/main/java/reposense/parser/CsvParser.java
anubh-v: Thanks for the suggestion.

What about keeping the expected header size as a private variable in `CsvParser` ? 
This variable can be set in the constructor of `CsvParser`.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -36,20 +36,25 @@
...
 
     private Path csvFilePath;
+    private int expectedHeaderSize;
     private int numOfLinesBeforeFirstRecord = 0;
 
     /**
      * @throws IOException if {@code csvFilePath} is an invalid path.
      */
-    public CsvParser(Path csvFilePath) throws IOException {
+    public CsvParser(Path csvFilePath, int expectedHeaderSize) throws IOException {
```
anubh-v: @fzdy1914, @jylee-git, do you think we can could keep the expected header size as a private variable in `CsvParser` like this?

With this approach, each child just needs to pass in the correct value of `expectedHeaderSize` when calling the constructor of the superclass.

src/main/java/reposense/parser/CsvParser.java
fzdy1914: I agree with your point. Go ahead.

src/main/java/reposense/parser/CsvParser.java
fzdy1914: I think it is a good approach. Do request a review from us again after you have finished everything.

src/main/java/reposense/parser/CsvParser.java
anubh-v: I removed the `getHeader` method entirely, since we are passing the expected header size to the constructor of `CsvParser`

src/main/java/reposense/parser/GroupConfigCsvParser.java
anubh-v: Thanks for the feedback. I removed the getHeader method entirely, since we are passing the expected header size to the constructor of CsvParser

anubh-v:
@jylee-git, @fzdy1914, I have made changes to your feedback.

I removed the `getHeader` method entirely, and chose to pass the expected header size via the constructor of `CsvParser` instead.

fzdy1914:
@anubh-v Thanks for the effort. You may choose to resolve the comment by yourself if you think the comment is trivial or you have solve the request.

anubh-v:
@jylee-git, @fzdy1914,
There is one limitation in the current solution.
The `parse` method in `CsvParser` will throw an `IOException` when the CSV file's header has the wrong number of columns.

However, the driver code in RepoSense.java will silently ignore the `IOException` for the `author-config.csv` and `group-config.csv` files.

Ideally, the users should be alerted that the `author-config.csv` and `group-config.csv` are malformed, so they can fix the files and re-run the analysis.

Proposed fix: Throw `InvalidCsvException` out from `parse` in `CsvParser`, and use it to log a warning for `author-config.csv` and `group-config.csv`, in the `RepoSense.java` code.

Should this fix be done within this PR, or separately? *It will significantly increase the PR size as changing the `throws` clause of the `parse` method will affect many existing test cases*.

src/main/java/reposense/RepoSense.java
```diff
@@ -93,17 +94,23 @@ public static void main(String[] args) {
         try {
             authorConfigs = new AuthorConfigCsvParser(cliArguments.getAuthorConfigFilePath()).parse();
             RepoConfiguration.merge(repoConfigs, authorConfigs);
-        } catch (IOException ioe) {
-            // IOException thrown as author-config.csv is not found.
+        } catch (FileNotFoundException fnfe) {
```
anubh-v: @jylee-git, @fzdy1914, I added a `catch` block here so we can warn the user if the author / group config files have an invalid header size (rather than silently ignoring the problem).

Should this be done within this PR, or separately?

fzdy1914: > Throw InvalidCsvException out from parse in CsvParser, and use it to log a warning for author-config.csv and group-config.csv, in the driver RepoSense.java code.

Hi, can you consider my Logic? Since we are dealing with CSV files, how about we directly set the `InvalidCsvException` a child of `IOException`. Then, instead of catching `FileNotFoundException`, we catch `InvalidCsvException` here and do the warning. We still leave the other `IOException` unhandled. In the meantime, we do not need to tough the test case a lot as it is a child of `IOException`.

anubh-v: But then we will be silently ignoring the IOExceptions besides the InvalidCsvException, right?
Would it be better to warn user about all IOExceptions that occur while parsing author and group configs, so the user can re-run the analysis ?

fzdy1914: OK, let's ask the senior's opinion. @yong24s @eugenepeh Your inputs, please?

eugenepeh: why is `FileNotFoundException` used for `if the author / group config files have an invalid header size`?

fzdy1914: His implementation is only ignoring `FileNotFoundException` instead of ignoring all `IOException` as we previously did.

anubh-v: @eugenepeh,  `FileNotFoundException` **will not** be thrown if the config files have an invalid header size. The `parse` method in `CsvParser` will throw an `IOException` if the config files have an invalid header size.

In our current implementation, we are silently ignoring all `IOExceptions` which are thrown while parsing author / group config files. However, it would be better to alert the user if the author / group configs have an invalid header size (or any other problem with the file).

Hence, I decided to distinguish between the case where the author / group config **is not present** and the case where the author / group config **is present but cannot be parsed because they are malformed / other IO error.**

The `FileNotFound` catch-block covers the first case (author / group config **is not present**).

eugenepeh: I see, sounds good to me

src/main/java/reposense/parser/GroupConfigCsvParser.java
```diff
@@ -23,8 +23,13 @@
     private static final int GROUP_NAME_POSITION = 1;
     private static final int FILES_GLOB_POSITION = 2;
 
+    /**
+     * Number of columns expected in the header of repo-config.csv file.
+     */
+    private static final int HEADER_SIZE = 3;
```
eugenepeh: can we do something like `EXPECTED_HEADER_SIZE = FILES_GLOB_POSITION + 1 // last position + 1`?

I would prefer it to be together with all the `POSITION` constants so that no one will forget to update it.

anubh-v: Done.

eugenepeh: Either is fine with me, but all the comments e.g.
```
    // Number of columns expected in the header of repo-config.csv file.
```
is rather redundant to me as this comment should be in the header comments of `CsvParser` instead.

Do add the inline comment `// last position + 1` so that subsequent developers would be reminded to update it

anubh-v: Done. I've added the inline comment `// last position + 1` 
On second thought, perhaps we do not need a header comment in `CsvParser` as the attribute name (`expectedHeaderSize`) is self explanatory.

anubh-v:
@fzdy1914, @jylee-git, may I get your review on the changes? Thanks very much.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -36,20 +37,25 @@
...
         if (csvFilePath == null || !Files.exists(csvFilePath)) {
-            throw new IOException("Csv file does not exists in given path.\n"
+            throw new FileNotFoundException("Csv file does not exists in given path.\n"
```
jylee-git: Csv file does not exist   ~s~

anubh-v: Done

yong24s:
@anubh-v 

Can you fix the tests? Thanks :-)

anubh-v:
@jamessspanggg, @yong24s, I've resolved the conflict and fixed the tests

Issue #436, opened 2018-12-17
Different author with same alias will not trigger any warning


fzdy1914:
As stated in the title. The alias will belong to the last author who claims it.

Tejas2805:
Removed `FirstTimers` label as it wasn't as easy as it looked.

PR #957, opened 2019-12-08
[#436] Proposed fix when there are duplicate aliases.

dezhanglee:
In this case, when there are duplicate alias, using emails may be more accurate as it is more reasonable to assume everyone has a unique email.

Proposed fix for issue #436


fzdy1914:
I think it may be an overkill for such a small issue. My initial thought is just to output a warning to the user.

Tejas2805:
@dezhanglee Will, you be working on this PR in the future? If not, can you close it? If you intend to complete it, you can keep it open.

Tejas2805:
Taking over this PR as it has become stale and no response from the dev.

PR #959, opened 2019-12-09
[#946] Improve visualization on sorting parameters

### LINKED ISSUE #946: CodeView: improve visualization on different sort parameters
We should place more emphasis on the values of interest using graphical effects.

E.g. if the sort by is on file name, we should highlight the file name in one way or another

image

jinyingtan:
Fixes #946 

```
No emphasis on the values being sorted on in the title of each file 
under authorship view 

Emphasizing what is being sorted would help to locate while scrolling
and know what are the values being sort on. 

Let's place what is being sorted at the front of the title.  
```


Tejas2805:
I feel the bracket is a bit of an overkill and the bold is not visible enough? Also for `LoC` it gets repeated in the end, so it looks redundant. 

Screen Shot 2019-12-09 at 11 25 15 PM

@damithc What do you think about the UI?

damithc:
Highlighting might be too strong too, if using yellow. After all, sorting is not the most important thing in that page. But a more subtle background color might work. We might want to wait for @eugenepeh's input too, as he is the one who raised this issue.

eugenepeh:
>Also for LoC it gets repeated in the end, so it looks redundant.

Yes, it looks pretty redundant. I think one way we can resolve this problem, instead of highlighting, would be to follow how Google presents their search result

image

We can use a slightly lighter tone of color for the entire title, and use `bold black for the value` under the sort category. 

Another way is to `bring the value under sort to the front of title`, so that it is more conspicuous. 
E.g. 
Given that the title has the default order `a b c d`
if b is the value under sort, then it should be displayed as `b a c d`

I am okay with either solution, can seek other developers' input too.

As for the "FileType" that has been missing from our file title (thanks for highlighting that out in this PR), I think we should probably insert it into the title permanently in a separate PR.

jinyingtan:
I would go for bringing the sort value `in front` and using `bold black` for the value as the current title is working with 2 different colors to differentiate the path & the LOC, which I feel that it is still good to keep

damithc:
> I would go for bringing the sort value `in front` and using `bold black` for the value as the current title is working with 2 different colors to differentiate the path & the LOC, which I feel that it is still good to keep

Go ahead @jinyingtan . Let us know when ready.

jinyingtan:
I have made the layout changes. I added brackets `()` for the sort of `LOC` and `file type` to aid in readability. Let me know if the changes are alright.

jinyingtan:
@damithc @eugenepeh @Tejas2805 let me know if the changes are alright

damithc:
Hmm... Netlify preview for this PR comes up empty 🤔

jinyingtan:
I think the netlify preview is viewable now

damithc:
I'm OK with the UI 👍

Tejas2805:
@damithc The font color for the number of lines is different in different ways of sorting. Maybe we can make that consistent ? 

Screen Shot 2020-01-13 at 4 38 06 PM

Screen Shot 2020-01-13 at 4 38 23 PM

damithc:
> @damithc The font color for the number of lines is different in different ways of sorting. Maybe we can make that consistent ?

No objections from me.

frontend/src/static/js/v_authorship.js
```diff
@@ -5,6 +5,20 @@ const filesSortDict = {
...
 
+const activeSortSectionDict = {
+  lineOfCode: (file) => `(${file.lineCount} lines)`,
+  path: (file) => file.path,
+  fileName: (file) => file.path.split(/[/]+/).pop(),
+  fileType: (file) => `(${file.fileType})`,
+};
+
+const pathSectionDict = {
+  lineOfCode: (file) => file.path,
```
eugenepeh: it doesnt quite make sense to set `lineOfCode` to `file.path`
perhaps we should do some change to the underlying structure?

Tejas2805:
@jinyingtan Will you be completing this PR? If not, do close it, so that other developers can try it out. 

You can keep it open, if you plan to complete it eventually.

jinyingtan:
Apologies for not completing the PR, as I have quite a bit of school work on hand. I'll close this PR

PR #966, opened 2019-12-13
[#904] Fix additional upward movement on expanding code

### LINKED ISSUE #904: Additional movement when expanding code
In some cases, when clicking the `+` button, the code seems to move up a bit before expanding.

expand-code

ang-zeyu:
Fixes #904 

```
Clicking the expand button may cause the code to move up a bit 
before expanding.

This occurs more frequently when the number of code lines to
load is large, or the client is slow. This is due to asynchronous
data loading by Vue, causing observable css updates to the
buttons before the code has been loaded into the DOM.

Let's execute the css change for the buttons only after the
the segment's code lines have been loaded.
```


frontend/src/static/js/v_segment.js
```diff
@@ -9,7 +9,10 @@ window.vSegment = {
   methods: {
     loadCode() {
       this.loaded = true;
-      window.toggleNext(this.$el.childNodes[0]);
+      // For updating buttons only once code has loaded into DOM
```
fzdy1914: Can be `Displaying code block only after the code has loaded into DOM`

ang-zeyu: Indeed, I overlooked clarifying between dom / css updates. Thank you!

> Can be `Displaying code block only after the code has loaded into DOM`

Shouldn't this still be `For updating buttons only once code has loaded into DOM` though?
Since as you mentioned, the css update (buttons) follows the secondary dom update (v-for code)

fzdy1914: It is just a suggestion. Actually, the CSS update not only updates the button but also controls the code block to be displayed, which is the major functionality of this CSS update. That's the reason I give the above recommendation.

ang-zeyu:
Updated comment

Issue #852, opened 2019-08-01
Zoom tab displays commits not within filter time


jamessspanggg:
Current: 
Instead of zooming in terms of the position of the dotted lines, the current zoom feature gets the nearest since and until dates from the dotted lines, and zooms into commits that are within the duration.

Issue: 
Ramps that are not included in the dotted lines are also included in the zoom tab.

<img width="1657" alt="Screen Shot 2019-08-01 at 11 48 03 PM" src="https://user-images.githubusercontent.com/32864116/62308056-56377280-b4b7-11e9-99b9-c59aab8128a0.png">

damithc:
Ideally, the duration should match the position of the dotted lines. if that is hard to do, we can make it match the nearest ramp _inside_ the box.

BTW, the triangles are supposed to be called _ramps_, not _ramp slices_.

fzdy1914:
I found another bug similar to this is that: if you send two same data range for zoom, the commits displayed in second time are not correct.

fzdy1914:
Meanwhile, I don't think your investigation for the bug is correct as the commit displayed is out of range.

![image](https://user-images.githubusercontent.com/35621726/70912685-71aeae80-204f-11ea-9aca-c832b56cd18b.png)

fzdy1914:
Supporting gif
![a](https://user-images.githubusercontent.com/35621726/70916177-32379080-2056-11ea-871a-97efd2ef0c72.gif)

PR #967, opened 2019-12-14
[#1145] SideBar: revamp display logic

### LINKED ISSUE #1145: Reactivity of zoom tab with changes in summary
The zoom tab is currently reactive to changes in summary (such as granularity changes). However, when merge group is checked, the zoom tab is no longer reactive to the changes in summary.

To replicate, any PR preview will do.
Kapture 2020-03-14 at 16 50 24

Should we make the zoom tab totally independent of the summary? As it will affect the implementations of PRs such as #779, #1101 and #967.

fzdy1914:
Fixes #1145 

```
When sidebar needs to be refreshed, such as re-rendering the 
authorship tab and zoom tab, we set `isTabActive` to false first to 
trigger the redrawing process. To ensure that the child component 
gets the updated changes, we bind a unique key to the child 
component to trigger re-rendering.

The previous implementation is done in such a way because the 
child component fails to get the changes in `info` prop due to the 
way we update the `info`[1]. This causes the SideBar component 
to be completely re-rendered, instead of only updating the 
component based on the changes in `info`. If we are able to let the 
SideBar component catch the change of `info`, we can directly 
watch the property and update the data upon the `info` 
change.

Let's declare the `info` prop explicitly and watch the changes in 
`info` prop so Vue is so Vue is able to catch the property change [2] 
to utilize components re-usage. Let's also pass a shallow copied 
`user` to the zoom tab, as we are intending to make zoom tab
and summary tab independent of one another.
```

[1] https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats
[2] https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties

### DIFF
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -117,10 +117,12 @@ window.app = new window.Vue({
         minDate: '',
         name: '',
         repo: '',
+        totalCommits: '',
       },
       tabZoom: {
         avgCommitSize: 0,
         filterGroupSelection: '',
+        filterTimeFrame: '',
         location: '',
         isMergeGroup: false,
         sinceDate: '',
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -117,7 +117,6 @@ window.app = new window.Vue({
         minDate: '',
         name: '',
         repo: '',
-        totalCommits: '',
       },
       tabZoom: {
         avgCommitSize: 0,
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -106,11 +106,28 @@ window.app = new window.Vue({
     userUpdated: false,
 
     isLoading: false,
-    isCollapsed: false,
     isTabActive: true, // to force tab wrapper to load
 
     tabType: 'empty',
-    tabInfo: {},
+    tabInfo: {
+      tabAuthorship: {
+        author: '',
+        location: '',
+        maxDate: '',
+        minDate: '',
+        name: '',
+        repo: '',
+      },
+      tabZoom: {
+        avgCommitSize: 0,
+        filterGroupSelection: '',
+        location: '',
+        isMergeGroup: false,
+        sinceDate: '',
+        untilDate: '',
+        user: null,
+      },
+    },
     creationDate: '',
 
     errorMessages: {},
@@ -171,14 +188,11 @@ window.app = new window.Vue({
 
     // handle opening of sidebar //
     activateTab(tabName) {
-      // changing isTabActive to trigger redrawing of component
-      this.isTabActive = false;
       if (document.getElementById('tabs-wrapper')) {
         document.getElementById('tabs-wrapper').scrollTop = 0;
       }
 
       this.isTabActive = true;
-      this.isCollapsed = false;
       this.tabType = tabName;
 
       window.addHash('tabOpen', this.isTabActive);
@@ -196,11 +210,11 @@ window.app = new window.Vue({
     },
 
     updateTabAuthorship(obj) {
-      this.tabInfo.tabAuthorship = Object.assign({}, obj);
+      this.tabInfo.tabAuthorship = Object.assign({}, this.tabInfo.tabAuthorship, obj);
       this.activateTab('authorship');
     },
     updateTabZoom(obj) {
-      this.tabInfo.tabZoom = Object.assign({}, obj);
+      this.tabInfo.tabZoom = Object.assign({}, this.tabInfo.tabZoom, obj);
       this.activateTab('zoom');
     },
 
@@ -245,10 +259,6 @@ window.app = new window.Vue({
           // handle zoom tab if needed
         }
       }
-    },
-
-    generateKey(dataObj) {
-      return JSON.stringify(dataObj);
     },
 
     getRepoSenseHomeLink() {
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -61,11 +61,6 @@ window.vAuthorship = {
         this.selectedFileTypes = this.fileTypes.slice();
       }
     },
-
-    info() {
-      this.initiate();
-      this.setInfoHash();
-    },
   },
 
   methods: {
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -26,7 +26,11 @@ const repoCache = [];
 const minimatch = require('minimatch');
 
 window.vAuthorship = {
-  props: ['info'],
+  props: {
+    info: {
+      type: Object,
+    },
+  },
   template: window.$('v_authorship').innerHTML,
   data() {
     return {
@@ -58,6 +62,11 @@ window.vAuthorship = {
       } else {
         this.selectedFileTypes = this.fileTypes.slice();
       }
+    },
+
+    info() {
+      this.initiate();
+      this.setInfoHash();
     },
   },
 
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -61,6 +61,11 @@ window.vAuthorship = {
         this.selectedFileTypes = this.fileTypes.slice();
       }
     },
+
+    info() {
+      this.initiate();
+      this.setInfoHash();
+    },
   },
 
   methods: {
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -27,9 +27,7 @@ const minimatch = require('minimatch');
 
 window.vAuthorship = {
   props: {
-    info: {
-      type: Object,
-    },
+    info: Object,
   },
   template: window.$('v_authorship').innerHTML,
   data() {
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -705,12 +705,12 @@ window.vSummary = {
 
     openTabZoom(user, since, until, repo, index) {
       const { avgCommitSize } = this;
-
+      const clonedUser = Object.assign({}, user); // so that changes in summary won't affect zoom
       this.$emit('view-zoom', {
         filterGroupSelection: this.filterGroupSelection,
         filterTimeFrame: this.filterTimeFrame,
         avgCommitSize,
-        user,
+        user: clonedUser,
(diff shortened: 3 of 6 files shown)


frontend/src/static/js/v_authorship.js
```diff
@@ -315,10 +323,11 @@ window.vAuthorship = {
     },
   },
 
-  created() {
+  mounted() {
     this.initiate();
     this.setInfoHash();
   },
+
```
fzdy1914: This line is added to follow the new line standard. Actually, we do not have a convention on when to have a new line in between the methods.

frontend/src/static/js/main.js
```diff
@@ -105,11 +105,28 @@ window.app = new window.Vue({
...
-    tabInfo: {},
+    tabInfo: {
+      tabAuthorship: {
+        author: '',
+        location: '',
+        maxDate: '',
+        minDate: '',
+        name: '',
+        repo: '',
+        totalCommits: ''
```
jamessspanggg: since `repo` and `location` are objects, should the initial values be null instead?

frontend/src/static/js/v_zoom.js
```diff
@@ -60,12 +64,16 @@ window.vZoom = {
...
     },
   },
-  created() {
-    this.filterCommits();
-  },
   mounted() {
+    this.filterCommits();
     this.updateExpandedCommitMessagesCount();
   },
+  watch: {
```
jamessspanggg: Can move the `watch` methods to be below the `data()`? To standardise the position of Vue hooks.

jamessspanggg:
@fzdy1914 travis is failing with lint errors. Please fix it as well.

frontend/src/static/js/main.js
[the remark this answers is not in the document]
fzdy1914: Not Sure though, let me check tmr if it is truely unused so I can clean it up in this PR as well.

fzdy1914: Don't think so, see https://github.com/reposense/RepoSense/blob/master/frontend/src/static/js/v_summary.js#L673-L680, they are actually strings.

jamessspanggg: my bad, made a mistake there.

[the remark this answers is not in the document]
fzdy1914: Checked, it is indeed unused in authorship tab.

frontend/src/static/js/v_authorship.js
```diff
@@ -315,10 +323,11 @@ window.vAuthorship = {
     },
   },
 
-  created() {
```
jamessspanggg: One more question, can you justify the change from `created` to `mounted`?

fzdy1914: My bad, changing it back.

fzdy1914:
Actually, this PR also fixes #852. However, even it fixes the issue, the logic there is not correct, I will open another PR all that issue.

jamessspanggg:
Perhaps can change the PR message slightly?
> When sidebar needed to be refreshed, like rerendering the authorship tab and zoom tab,
> we set `isTabActive` to false first to trigger the redrawing process. To ensure the child
> component gets the change updated, we bind a unique key to the child component to
> trigger rerendering.

...needs to be refreshed
.. ensure that the child component gets the updated changes

> The previous implementation is done in such a way because the child component fails
> to get the change of `info` property bound which due to the way we update the info[1].
> And it will cause SideBar Component to have to go through the whole circle every time
> when we trigger a rerendering. If we are able to let the SideBar component catch the
> change of `info`, we can directly watch the property and update the data upon the `info`
> change.

... get the change in `info` props due to the way we update the `info`[1]
... This causes the sidebar component to be completely re-rendered, instead of only updating the component based on the changes in `info`.

> Let's declare the `info` data explicitly and edit its corresponding updating method so
> Vue is able to catch the property change[2] to utilize components re-usage. Meanwhile,
> some related unused method are removed.
> 
> [1] https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats
> [2] https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties

Let's declare the `info` prop explicitly and watch the changes in `info` prop so Vue is....

frontend/src/static/js/v_authorship.js
```diff
@@ -26,7 +26,9 @@ const repoCache = [];
 const minimatch = require('minimatch');
 
 window.vAuthorship = {
-  props: ['info'],
+  props: {
+    info: Object,
```
jamessspanggg: One question: won't declaring it in the previous way using `props: ['info']` will achieve the same purpose? Since you are watching the `info` prop.

fzdy1914: It will, but the way declaring it clearly is what Vue suggested though. 
https://vuejs.org/v2/guide/components-props.html#Prop-Types

fzdy1914: I believe declaring it clearly has better code quality.

jamessspanggg: ok noted.

jamessspanggg:
>  If we can't make everything reactive, might as well not make anything reactive so that there is no confusion to the user. It is fine the right-side panel to not be reactive to the left-side panel, across the entire application.

Perhaps we can disable the reactivity of zoom tab in this PR?

fzdy1914:
> > If we can't make everything reactive, might as well not make anything reactive so that there is no confusion to the user. It is fine the right-side panel to not be reactive to the left-side panel, across the entire application.
> 
> Perhaps we can disable the reactivity of zoom tab in this PR?

I think so. What you like to take over this PR and do the change for me?

jamessspanggg:
> I think so. Whould you like to take over this PR and do the change for me?

Okay sure.

fzdy1914:
Then I will close this PR for now.

jamessspanggg:
@fzdy1914 i can push directly to this PR right? so that I can continue ur work

jamessspanggg:
@fzdy1914 the watchers are removed, the user passed to zoom tab is shallow copied, so currently changes in summary won't affect zoom tab.

fzdy1914:
> @fzdy1914 the watchers are removed, the user passed to zoom tab is shallow copied, so currently changes in summary won't affect zoom tab.

No need to remove the watchers, we only need to change the input user to a cloned user.

fzdy1914:
> > @fzdy1914 the watchers are removed, the user passed to zoom tab is shallow copied, so currently changes in summary won't affect zoom tab.
> 
> No need to remove the watchers, we only need to change the input user to a cloned user.

Watcher is used when opening another zoom tab.

frontend/src/static/js/main.js
```diff
@@ -247,10 +263,6 @@ window.app = new window.Vue({
       }
     },
 
-    generateKey(dataObj) {
-      return JSON.stringify(dataObj);
```
jamessspanggg: this deletion is due to the method no longer being used

fzdy1914: Correct.

frontend/src/static/js/main.js
```diff
@@ -106,11 +106,28 @@ window.app = new window.Vue({
...
     isTabActive: true, // to force tab wrapper to load
 
     tabType: 'empty',
-    tabInfo: {},
+    tabInfo: {
+      tabAuthorship: {
+        author: '',
+        location: '',
+        maxDate: '',
+        minDate: '',
```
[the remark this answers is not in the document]
fzdy1914: I previously thought `totalCommits` is not used anywhere? Can you recheck?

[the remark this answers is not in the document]
jamessspanggg: yep you are right

frontend/src/static/js/v_authorship.js
```diff
@@ -59,6 +61,11 @@ window.vAuthorship = {
         this.selectedFileTypes = this.fileTypes.slice();
       }
     },
+
+    info() {
+      this.initiate();
+      this.setInfoHash();
```
eugenepeh: this seems like a duplication of `created()`, is it necessary to have both?

jamessspanggg: Yes I think so.

Previously, when authorship tab is already open, clicking to another author's code will close the tab and open the tab to trigger the redrawing of the entire authorship tab, which will start over and call `created()` of the authorship tab.

This PR removes that unnecessary redrawing logic, so `created()` won't be called again when user clicks on another author's code. That is why we need a watcher that listens to the changes in `info`, and updates the authorship tab accordingly.

eugenepeh:
Note that #830 would also need to implement the same changes; depending on whichever is merged first

jamessspanggg:
> Note that #830 would also need to implement the same changes; depending on whichever is merged first

I think this should be merged first, as it affects both #830 and #1101

frontend/src/static/js/v_summary.js
```diff
@@ -706,11 +705,12 @@ window.vSummary = {
 
     openTabZoom(user, since, until, repo, index) {
       const { avgCommitSize } = this;
-
+      const clonedUser = Object.assign({}, user); // so that changes in summary won't affect zoom
```
ang-zeyu: Changes to `v_authorship/zoom` looks good.
I'm not too familiar with `v_summary` to know if there are any other issues here though =X

Just one question: this is what fixes #1145 right?

jamessspanggg: Yep, changes in summary won't affect zoom anymore.

jamessspanggg:
@fzdy1914 can merge this?

fzdy1914:
@jamessspanggg Seems that I can't change the author of this PR to you. So I will set you to be the co-author of this PR. For the below two info, which one would you choose to use as the co-author?

image

jamessspanggg:
> @jamessspanggg Seems that I can't change the author of this PR to you. So I will set you to be the co-author of this PR. For the below two info, which one would you choose to use as the co-author?
> 
> image

The jamessspanggg one. Thank you!

frontend/src/static/js/v_zoom.js
```diff
@@ -45,6 +48,11 @@ window.vZoom = {
       return nonEmptyCommitMessageCount;
     },
   },
+  watch: {
+    info() {
+      this.updateExpandedCommitMessagesCount();
```
jamessspanggg: just realised that `setInfoHash` is missing here, will fix in #779

PR #969, opened 2019-12-15
[#879] Display commit tags in commits panel

### LINKED ISSUE #879: Commits panel: Capture and display tags beside each commits
With CS21x3 making use of git tags actively, we can consider enhancing the commits panel such that tags (if any) are captured and displayed beside the commit as well.

jamessspanggg:
Fixes #879 

```
The tags of each commit are not shown in the commits panel.

With the active use of git tags in CS21x3 repositories, it is 
useful to provide tutors on the usage and information of tags
in each commit.

Let's display the tags of each commit (if there's any) in the 
commits panel.
```


jamessspanggg:
@damithc prof, can I get your opinion on the UI of the tags? 

To view an example in netlify, click on `view breakdown of commits` of author Eugene Peh. You should be able to click on the tags on top of the ramp chart, which will scroll to the corresponding commit.

damithc:
> To view an example in netlify, click on `view breakdown of commits` of author Eugene Peh. You should be able to click on the tags on top of the ramp chart, which will scroll to the corresponding commit.

Not bad. Can also try this style:
image

`class="badge badge-dark text-white font-weight-light"`

jamessspanggg:
@damithc is the following UI ok?

Screen Shot 2019-12-17 at 12 11 20 AM

damithc:
> @damithc is the following UI ok?

Yes. Add some margin so that the tag doesn't touch the box below? Also, no need for repo/branch?

damithc:
UI-wise, can we try dark grey? Or light grey with black text? As tags are not the most important thing in that view, it should not be the thing that attracts the viewer's attention the most.

jamessspanggg:
> UI-wise, can we try dark grey? Or light grey with black text?

Dark grey with white text:
Screen Shot 2019-12-17 at 12 54 27 PM

Light grey with black text:
Screen Shot 2019-12-17 at 12 56 12 PM

damithc:
> Light grey with black text:
> Screen Shot 2019-12-17 at 12 54 27 PM
> 
> Dark grey with white text:
> Screen Shot 2019-12-17 at 12 56 12 PM

I'm OK with either. Choose the one that goes better with the rest of the page.

jamessspanggg:
> I'm OK with either. Choose the one that goes better with the rest of the page.

Ok. I'm going with the dark grey with white font.

jamessspanggg:
> Just a question, can we add the tag in the JSON only when there is a tag?

Yes I think so, by setting the tag field in Java to null before converting to JSON. Thanks for question, will modify the implementation.

frontend/src/tabs/zoom.pug
```diff
@@ -12,6 +12,13 @@
       span {{ info.sinceDate }} to {{ info.untilDate }} &nbsp;
       a(v-on:click="openSummary") [Show ramp chart for this period]
     .zoom__title--granularity.mini-font granularity: one ramp per {{ filterTimeFrame }}
+    .zoom__title--tags
+      template(v-for="commit in info.user.commits")
```
fzdy1914: Why not just use div?

frontend/src/tabs/zoom.pug
```diff
@@ -31,15 +38,19 @@
...
-            v-if="slice.messageBody !== ''",
-            v-on:click="updateExpandedCommitMessagesCount",
-            onclick="toggleNext(this)",
-            title="Click to show/hide the commit message body"
-          )
-            i.commit-message--button.fas.fa-ellipsis-h
+          .title
+            a(v-bind:href="getSliceLink(slice)" target="_blank") {{ slice.messageTitle }}
+            span &nbsp; ({{ slice.insertions }} lines) &nbsp;
+            .tag(v-if="slice.tag", v-bind:class="[`${slice.hash}`]")
```
fzdy1914: Why do you need to bind the hash to this tag?

frontend/src/tabs/zoom.pug
jamessspanggg: Reason is because using `template` does not introduce extra nodes in the DOM, whereas using `div` will. Since the functionality of these two nested `template` is to loop through the commits, it is unnecessary to create divs to complicate the DOM.

frontend/src/tabs/zoom.pug
jamessspanggg: Because I have a method allows users to scroll to the specific commit (based on the html class name) when the tag is clicked.  I used the hash as the classname because the hash is unique, don't need to worry about duplicates.

https://github.com/reposense/RepoSense/blob/4a63761c17967ef5e14722130b0da3fccf048410/frontend/src/static/js/v_zoom.js#L45-L50

src/main/java/reposense/commits/model/CommitResult.java
```diff
@@ -11,19 +11,21 @@
...
+    private final String tag;
     private final int insertions;
     private final int deletions;
 
     private final transient Author author;
     private final transient Date time;
 
     public CommitResult(Author author, String hash, Date time, String messageTitle,
-            String messageBody, int insertions, int deletions) {
+            String messageBody, String tag, int insertions, int deletions) {
```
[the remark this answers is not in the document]
jamessspanggg: I have tried using `Optional<String>` for the `tag` field.

When converting to JSON, it will convert the tag fields as JSON objects instead of a key-value pair.

So with empty tag fields, the tag will also appear as empty JSON objects:
Screen Shot 2019-12-17 at 9 38 57 PM
whereas if we use the current implementation (set string to null), the tag field will only appear in the JSON when it is not null.

frontend/src/tabs/zoom.pug
fzdy1914: OK

src/main/java/reposense/commits/model/CommitResult.java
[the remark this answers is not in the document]
fzdy1914: OK, that's fine.

frontend/src/tabs/zoom.pug
fzdy1914: OK

frontend/src/tabs/zoom.pug
```diff
@@ -32,14 +39,17 @@
...
+            i.fas.fa-tags
+            span &nbsp; {{ slice.tag }}
           a(
             v-if="slice.messageBody !== ''",
             v-on:click="updateExpandedCommitMessagesCount",
             onclick="toggleNext(this)",
             title="Click to show/hide the commit message body"
           )
-            i.commit-message--button.fas.fa-ellipsis-h
+              i.commit-message--button.fas.fa-ellipsis-h
```
fzdy1914: Seem unnecessary.

jylee-git:
Just checking, is the PR supposed to capture multiple tags tagged to each commit also?

image

jamessspanggg:
> Just checking, is the PR supposed to capture multiple tags tagged to each commit also?

Yes it should. My bad, will implement it.

jamessspanggg:
> Can we have a test case with multiple tags in a commit?

Test case added. Thanks for reviewing @jylee-git

damithc:
@jamessspanggg I'm OK with the UI. Ask for my review again (for the code) when ready to merge (i.e., after others have approved).

jamessspanggg:
> Is this intended?

It is only intended for the tags on top of the ramp chart as clicking on them will direct to the corresponding commit. I have fixed the mistake.

eugenepeh:
Something seems to broke the preview, not sure if its related to this PR. 
image

I will trigger a rebuild and review again afterward

src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
```diff
@@ -170,6 +170,29 @@ public void analyzeCommits_commitsWithEmptyCommitMessageTitleOrBody_success() th
...
+        List<CommitResult> expectedCommitResults = new ArrayList<>();
+
+        expectedCommitResults.add(new CommitResult(author, "62c3a50ef9b3580b2070deac1eed2b3e2d701e04",
+                parseGitStrictIsoDate("2019-12-20T22:45:18+08:00"), "Single Tag Commit",
+                "", new String[] {"1st"}, 2, 1));
+        expectedCommitResults.add(new CommitResult(author, "c5e36ec059390233ac036db61a84fa6b55952506",
+                parseGitStrictIsoDate("2019-12-20T22:47:21+08:00"), "Double Tag Commit",
+                "", new String[] {"2nd-tag", "1st-tag"}, 1, 0));
+
+        config.setBranch("879-CommitInfoAnalyzerTest-analyzeCommits_commitsWithMultipleTags_success");
```
eugenepeh: I think past branches use the PR number instead, but I am okay with either

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -78,9 +81,27 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
-        return new CommitResult(author, hash, date, messageTitle, messageBody, insertion, deletion);
+        return new CommitResult(author, hash, date, messageTitle, messageBody, tags, insertion, deletion);
+    }
+
+    /**
+     * Extracts the tag names in {@code tags}.
+     */
+    private static void extractTagNames(String[] tags) {
+        for (int i = 0; i < tags.length; i++) {
+            tags[i] = tags[i].substring(tags[i].lastIndexOf("tag: ") + 5);
```
eugenepeh: probably make `"tag: "` a constant

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -31,6 +31,7 @@
 
     private static final String LOG_SPLITTER = "\\|\\n\\|";
     private static final String REF_SPLITTER = ",\\s";
+    private static final String TAG_TITLE = "tag:";
```
eugenepeh: would `prefix` be more appropriate than `title`?

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -100,7 +101,7 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
      */
     private static void extractTagNames(String[] tags) {
         for (int i = 0; i < tags.length; i++) {
-            tags[i] = tags[i].substring(tags[i].lastIndexOf("tag: ") + 5);
+            tags[i] = tags[i].substring(tags[i].lastIndexOf(TAG_TITLE) + 5);
```
eugenepeh: On this line, you have purposely added a space after `tag:` isn't it? hence the constant `5` over there.
To make it "safer" usage, probably use `TAG_TITLE.length()` instead of 5, and `.trim()` the result or just use `tag: ` in the first place

jamessspanggg: yep noted. thanks for the advice.

frontend/src/static/css/style.scss
```diff
@@ -715,3 +723,17 @@ header {
     margin: 0;
   }
 }
+
+/* Tags in commits */
+.tag {
```
fzdy1914: One last question, shouldn't this tag be inside v_zoom only? Can change the place of this tag?

jamessspanggg: Yes can do. I placed it outside as the `commit-message` class is placed out of v_zoom as well.

jamessspanggg: Tried to move the tag to be within v_zoom. Since the tag is in two different places (on top of the ramp and on each commit), it is better to place the `.tag` css class outside as if we put inside then might need to duplicate in two places.

fzdy1914: Maybe we should put commit-message inside v_zoom as well? I don't think we need to duplicate the code.

jamessspanggg: Moved the `.tag` class into `.zoom`. Should I move `commit-message` in this PR as well? As this PR maybe unrelated to moving `commit-message`

fzdy1914: No need to move commit-message.

PR #971, opened 2019-12-16
[#852] Zoom tab displays commits not within filter time rage

### LINKED ISSUE #852: Zoom tab displays commits not within filter time

fzdy1914:
Fixes #852

When we try to open the zoom tab for the same user with the same 
date range for the second time, commits displayed the second time 
contain commit that does not fall within the filter time range.

This is mainly due to we edit the `user` object of `info` property 
directly. However, `info` is the property of the child component so
 we are not supposed to change it directly. The situation mentioned 
above occurs because this property is overwritten again by the 
parent, which related to the way we re-render the sidebar.

Let's use another computed value to represent the user with filtered
 commits so we do not need to modify the property directly. Minor 
changes are done together to improve code quality.


frontend/src/tabs/zoom.pug
```diff
@@ -1,31 +1,32 @@
...
-      span(v-else) {{ info.user.repoName }}
+      span(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ filteredUser.displayName }} ({{ filteredUser.name }})
+      span(v-else) {{ filteredUser.repoName }}
     .zoom__title--chart.medium-font(v-if="!info.isMergeGroup")
       span &#8627; &nbsp;
-      span(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ info.user.repoName }}
-      span(v-else) {{ info.user.displayName }} ({{ info.user.name }})
+      span(v-if="info.filterGroupSelection === 'groupByAuthors'") {{ filteredUser.repoName }}
+      span(v-else) {{ filteredUser.displayName }} ({{ filteredUser.name }})
+      span(v-else) {{ filteredUser.displayName }} ({{ filteredUser.name }})
```
jamessspanggg: Is this repeated?

fzdy1914: My bad.

fzdy1914:
Fixed the issue mentioned in #948

Tejas2805:
One nit. 

The `info.user` here should be `filteredUser`

Screen Shot 2020-01-02 at 4 44 05 PM

Tejas2805:
Also, can you update the commit message?

fzdy1914:
> Also, can you update the commit message?

I don't feel that the commit message needs to be updated as the changes are quite small. It is more like improving our code quality.

frontend/src/static/js/v_zoom.js
```diff
@@ -5,41 +5,44 @@ window.vZoom = {
...
     },
-
-    getCommitMessageBodyCount() {
+    totalCommitMessageBodyCount() {
       let nonEmptyCommitMessageCount = 0;
-      this.info.user.commits.forEach((commit) => {
+      this.filteredUser.commits.forEach((commit) => {
         commit.commitResults.forEach((commitResult) => {
-          if (commitResult.messageBody !== '') {
+          if (commitResult.messageBody !== '' && commitResult.insertions > 0) {
```
fzdy1914: This line is added as we only display commits with insertions > 0

Tejas2805:
@eugenepeh @yong24s For your final review.

PR #972, opened 2019-12-17
[#941] GitLsTree: modify regex to capture all windows reserved characters

### LINKED ISSUE #941: RepoSense for Windows fails to catch improper file name
For this repo

https://github.com/JacobToresson/duke/tree/master/text-ui-test%20/.idea

The culprit folder name is `text-ui-test ` (note the space at the end of the folder name). This folder name is not allowed in Windows but RepoSense fails to display a proper error message for the failure to analyze the repo.

Suspected issue: at the validation stage (where directory names are validated to ensure that they work for windows), such directory name (above) failed to be caught.

image

jylee-git:
Resolves #941

```
The regex for illegal file names in Windows does not take into account
directories with trailing whitespaces
(e.g.: "folderName /fileName.txt").

Such invalid directory names are wrongly validated as correct, causing
the analysis to be terminated with GitCloneException but the root
cause of the problem is repo with illegal directory names.

Let's modify the regex to capture trailing whitespaces in directory
names, and all other naming conventions illegal in Windows that were
not captured previously.
```


src/main/java/reposense/git/GitLsTree.java
```diff
@@ -19,14 +19,15 @@
  * Git ls tree is responsible to obtain the list of staged files of a branch.
  */
 public class GitLsTree {
+    // Although forward-slash (/) is an invalid character in Windows file path, it is not included in the regex as
+    // for the output of git-ls-tree, files in directories are separated by forward-slash (e.g.: folder/name/file.txt).
+    // Also, it is not possible to create and commit files with forward-slash characters in UNIX OSes.
+    private static final Pattern ILLEGAL_WINDOWS_FILE_PATTERN = Pattern.compile(
+            "^((CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(?=\\..*|$))|(^\\s)|[<>:\"\\\\|?*\\x00-\\x1F]|([\\s.]$)",
+            Pattern.CASE_INSENSITIVE);
```
jylee-git: Sources:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN

The regex used is extracted from the one given [here (src 1)](https://stackoverflow.com/a/6804755/11570538), but modified such that: -
- The logic is reversed to capture illegal/reserved characters/names in the directories and to output to the user the offending character/name
- the regex adapts to the file names output by git-ls-tree command.

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -19,14 +19,18 @@
...
 public class GitLsTree {
+    // Although forward-slash (/) is an invalid character in Windows file path, it is not included in the regex as
+    // for the output of git-ls-tree, files in directories are separated by forward-slash (e.g.: folder/name/file.txt).
+    // Also, it is not possible to create and commit files with forward-slash characters in UNIX OSes.
+    private static final Pattern ILLEGAL_WINDOWS_FILE_PATTERN = Pattern.compile(
+            "^((CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(?=\\..*|$))" // contains reserved values (e.g.: com1)
+                    + "|[<>:\"\\\\|?*\\x00-\\x1F]" // contains any reserved characters in directory name
+                    + "|(^\\s)|(/\\s+)" // file or folder names with leading whitespaces
+                    + "|([\\s.]$)|([\\s.]+/)", // folder or file names ending with period or whitespaces
+            Pattern.CASE_INSENSITIVE);
```
jylee-git: Sources:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN

The regex used is extracted from the one given [here (src 1)](https://stackoverflow.com/a/6804755/11570538), but modified such that: -
- The logic is reversed to capture illegal/reserved characters/names in the directories and to output to the user the offending character/name
- the regex adapts to the file names output by git-ls-tree command.

On the side note, I decided to put comments beside every lines here due to the complexity of this regex, though not sure if it may be excessive.

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -19,14 +19,18 @@
...
 public class GitLsTree {
+    // Although forward-slash (/) is an invalid character in Windows file path, it is not included in the regex as
+    // for the output of git-ls-tree, files in directories are separated by forward-slash (e.g.: folder/name/file.txt).
+    // Also, it is not possible to create and commit files with forward-slash characters in UNIX OSes.
+    private static final Pattern ILLEGAL_WINDOWS_FILE_PATTERN = Pattern.compile(
+            "^((CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(?=\\..*|$))" // contains reserved values (e.g.: com1)
+                    + "|[<>:\"\\\\|?*\\x00-\\x1F]"  // contains any reserved characters in directory name
+                    + "|(^\\s)|(/\\s+)"             // file or folder names with leading whitespaces
+                    + "|([\\s.]$)|([\\s.]+/)",      // folder or file names ending with period or whitespaces
+            Pattern.CASE_INSENSITIVE);
```
jylee-git: MSDN Documentation on windows file/directory naming:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN

The regex used is extracted from the one given [here](https://stackoverflow.com/a/6804755/11570538), but modified such that: -
- The logic is reversed to capture illegal/reserved characters/names in the directories and to output to the user the offending character/name
- the regex adapts to the file names output by git-ls-tree command.

On the side note, I decided to put comments beside every lines here due to the complexity of this regex, though not sure if it may be excessive or following the convention.

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -19,14 +19,18 @@
...
 public class GitLsTree {
+    // Although forward-slash (/) is an invalid character in Windows file path, it is not included in the regex as
+    // for the output of git-ls-tree, files in directories are separated by forward-slash (e.g.: folder/name/file.txt).
+    // Also, it is not possible to create and commit files with forward-slash characters in UNIX OSes.
+    private static final Pattern ILLEGAL_WINDOWS_FILE_PATTERN = Pattern.compile(
+            "((?<=^|/)(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(?=\\..*|/|$))" // contains reserved values (e.g.: com1)
+            + "|[<>:\"\\\\|?*\\x00-\\x1F]"      // contains any reserved characters in directory name
+            + "|(^\\s)|((?<=/)\\s+)"            // file or folder names with leading whitespaces
+            + "|([\\s.]$)|([\\s.]+(?=/))",      // folder or file names ending with period or whitespaces
+            Pattern.CASE_INSENSITIVE);
```
jylee-git: MSDN Documentation on windows file/directory naming:
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN

The regex used is extracted from the one given [here](https://stackoverflow.com/a/6804755/11570538), but modified such that: -
- The logic is reversed to capture illegal/reserved characters/names in the directories and to output to the user the offending character/name
- the regex adapts to the file names output by git-ls-tree command.

On the side note, I decided to put comments beside every lines here due to the complexity of this regex, though not sure if it may be excessive or following the convention.

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -19,14 +19,18 @@
  * Git ls tree is responsible to obtain the list of staged files of a branch.
  */
 public class GitLsTree {
+    // Although forward-slash (/) is an invalid character in Windows file path, it is not included in the regex as
```
fzdy1914: One nit is to follow the existing class attribute order.

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -45,10 +49,8 @@ public static void validateFilePaths(RepoConfiguration config, Path clonedBareRe
 
         for (String path : paths) {
             path = StringsUtil.removeQuote(path);
-            Matcher matcher = ILLEGAL_WINDOWS_CHARACTER_PATTERN.matcher(path);
 
-            if (matcher.find()) {
-                logger.log(Level.SEVERE, String.format(MESSAGE_INVALID_PATH, path, matcher.group()));
+            if (!isValidWindowsFilename(path)) {
```
fzdy1914: Shouldn't we go through this test only when RepoSense is runned on windows?

src/test/java/reposense/git/GitLsTreeTest.java
```diff
@@ -13,12 +19,44 @@
 import reposense.util.SystemUtil;
 
 public class GitLsTreeTest extends GitTestTemplate {
+    private static Method isValidWindowsFilenameMethod;
+
+    @BeforeClass
+    public static void beforeSubClass() throws NoSuchMethodException {
```
fzdy1914: Method name can be just `beforeClass`

src/main/java/reposense/git/GitLsTree.java
jylee-git: There’s already a check at the start of the method to see if the host is on Windows OS.

src/test/java/reposense/git/GitLsTreeTest.java
jylee-git: This is on purpose as the `@BeforeClass` of the parent class and this sub class are meant to be called without overriding it, so that the `@BeforeClass` of both the parent and sub class will be called.

Also if I override it by changing the method name, I'll need to call the `@BeforeClass` of the parent class as well, 


image

image

I'll commit these changes first, then see if this new change is fine

src/main/java/reposense/git/GitLsTree.java
fzdy1914: OK

jamessspanggg:
can't really review on this, not using windows :/

src/test/java/reposense/git/GitLsTreeTest.java
```diff
@@ -1,24 +1,66 @@
...
+        Assume.assumeTrue(SystemUtil.isWindows());
+
+        List<String> validDirectoryNames = Arrays.asList("com10.txt", "folder name/file name.txt", ".txt");
+        for (String fileName : validDirectoryNames) {
+            Assert.assertTrue((boolean) isValidWindowsFilenameMethod.invoke(null, fileName));
+        }
+    }
+
+    @Test
+    public void isValidWindowsFilename_illegalFileNames_success()
```
0blivious: Maybe the name can be `isValidWindowsFilename_illegalFileNames_fail`?

src/main/java/reposense/git/GitLsTree.java
```diff
@@ -19,14 +19,18 @@
...
-    // Although forward-slash (/) is an invalid character in Windows file path, it is not used as the forward-slash in
-    // the output of git-ls-tree is ambiguous. i.e. /path/to/test\file or /path/contains:a/file.
-    // Also, it is not possible to create and commit such a file on Unix systems.
-    private static final Pattern ILLEGAL_WINDOWS_CHARACTER_PATTERN = Pattern.compile("[:\\\\*?|<>:\"]");
+    // Although forward-slash (/) is an invalid character in Windows file path, it is not included in the regex as
+    // for the output of git-ls-tree, files in directories are separated by forward-slash (e.g.: folder/name/file.txt).
+    // Also, it is not possible to create and commit files with forward-slash characters in UNIX OSes.
+    private static final Pattern ILLEGAL_WINDOWS_FILE_PATTERN = Pattern.compile(
+            "((?<=^|/)(CON|PRN|AUX|NUL|COM[1-9]|LPT[1-9])(?=\\..*|/|$))" // contains reserved values (e.g.: com1)
+            + "|[<>:\"\\\\|?*\\x00-\\x1F]"      // contains any reserved characters in directory name
```
eugenepeh: perhaps group all the escaped symbols together for readability

jylee-git:
> @jylee-git
> I made some changes to your commit message, do check to see if it capture your work correctly.
> 
> For the second paragraph, I used GitCloneException instead of cryptic exception message as I believe the exception message is just GitCloneException#message. Otherwise, it is a good word for describing a messy stack trace :-).

Yes the commit message correctly reflects the work done. Thanks for helping with the commit message @yong24s

PR #976, opened 2019-12-17
[#181] Inconsistent thickness of different contribution bars

### LINKED ISSUE #181: Inconsistent thickness (height) of different contribution lines
The height of the contribution appears to be different for different contibutors.

Let's fix this to make it look consistent for every contributor.

image

fzdy1914:
fixes #181

The height of the contribution bar appears to be different for different
 contributors.

It is because we use rem to represent the height, which may result 
in a non-integer px value. And the browsers may display the same 
floating-point px differently.

Let's make it look consistent for every contributor by explicitly 
setting the height of the pixel to be in px.


fzdy1914:
> Could you standardise the thickness for the `fileType` bars as well?
> 
> https://github.com/reposense/RepoSense/blob/8c5592e2fbaaff5d0e8d05ee9e89db13c3a6ef74/frontend/src/static/css/style.scss#L371-L377

Done

fzdy1914:
> LGTM. But is it possible to use a single CSS class for the `height` and the `margin-top` for the contribution bars? This is so that we don't need to modify separately each time.

Made the change.

jylee-git:
Travis failed on 

image

frontend/src/static/css/style.scss
```diff
@@ -364,16 +364,9 @@ header {
...
-      margin-top: .1rem;
-    }
-
-    &--bar--fileType {
-      display: inline-block;
+      height: 4px;
+      margin-top: 2px;
       float: left;
-      height: .3rem;
-      margin-top: .1rem;
```
jylee-git: the attribute `float` should come before height and margin

fzdy1914: Updated just now.

fzdy1914:
> Could you update the frontend test also?

Which frontend test do you mean?

PR #977, opened 2019-12-17
[#940] ReportGenerator#generateReposReport: SLAP copying of template file

### LINKED ISSUE #940: ReportGenerator#generateReposReport: Copying of template files can be abstracted out, in adherence to SLAP
Before cloning and analyzing the repositories in `generateReposReport`, we copy some template files to the `reposense-report` directory. 

**This is done in lines 90-91 in the code snippet below.**

https://github.com/reposense/RepoSense/blob/a90ec6bda53ecfd7fa934a0c1b56bfc8ee6c18d2/src/main/java/reposense/report/ReportGenerator.java#L87-L114

This operation is specified at quite a granular level, and it clashes with the higher level of abstraction used for the other operations in `generateReposReport`. We can abstract out the copying of template files, into a separate method and then call this method from `generateReposReport`.

This would make all operations in `generateReposReport` use a more consistent level of abstraction.

ccyccyccy:
Fixes #940 
```
The operation to copy template file before cloning and analyzing the
repositories in ReportGenerator#generateReposReport is specified at a 
granular level.

It clashes with the higher level of abstraction used for the other
operations in ReportGenerator#generateReposReport.

Let's abstract out the copying of template files into a separate method
called ReportGenerator#prepareTemplateFile and then call this method
from ReportGenerator#generateReposReport.

```


src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -112,6 +111,11 @@
         return reportFoldersAndFiles;
     }
 
+    private static void prepareTemplateFile(String outputPath) throws IOException {
```
anubh-v: @ccyccyccy, perhaps this method can be documented using Javadoc?

ccyccyccy: Thanks for the reminder. I forgot about it.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -112,6 +111,15 @@
         return reportFoldersAndFiles;
     }
 
+    /**
+     * Copies the template file to the specified {@code outputPath} for the repo report to be generated
+     * @throws IOException if {@code is} refers to an invalid path.
```
eugenepeh: For header comments, especially static methods, we do not quote internal variables in javadoc because its meant to be hidden/abstract. So in this case, you can be more explicit about it, e.g. if the template resource is not found.

ccyccyccy: Thanks for the feedback. Just changed it.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -112,6 +111,15 @@
         return reportFoldersAndFiles;
     }
 
+    /**
+     * Copies the template file to the specified {@code outputPath} for the repo report to be generated
```
anubh-v: The sentence is missing a period.

fzdy1914:
@ccyccyccy Can you help resolve @anubh-v 's comment?

src/main/java/reposense/report/ReportGenerator.java
ccyccyccy: Fixed.

ccyccyccy:
> 
> 
> @ccyccyccy Can you help resolve @anubh-v 's comment?

Oops okay done.

PR #983, opened 2020-01-02
[#964] Add UG link to report page

### LINKED ISSUE #964: Consider adding link to UG in the report page
In `README.md`, the link to the UG always points to the release version of the UG.

For some who run RepoSense using master branch, or any other version aside from the release branch, there may be feature and/or functional discrepancies between master (or any specific commit) and release branch.

It would be good if there's a link from the report page that directly links to the UG corresponding to the RepoSense version that was ran on.

Tejas2805:
Fixes #964 

```
The link to the UG in the documents points to the released
version of the UG. As a result, there may be some discrepancies 
between the master and the release branch. 

Let's add a link to the UG on the report page corresponding to the 
version that RepoSense was ran on.
```


Tejas2805:
@damithc @fzdy1914 We could also do something like this?

Screen Shot 2020-01-02 at 1 25 39 PM 1

fzdy1914:
> @damithc @fzdy1914 We could also do something like this?
> 
> Screen Shot 2020-01-02 at 1 25 39 PM 1

If it is someone else using RepoSense, they may not want extra info to be displayed. But, it is still acceptable.

damithc:
> @damithc @fzdy1914 We could also do something like this?
> 
> Screen Shot 2020-01-02 at 1 25 39 PM 1

Limit it to one line? Also, there is no need for three links. We can link the version/commit number to the UG instead?

fzdy1914:
One nit, we do not use `currently` or `now` in the commit message as it is directly implied.

frontend/src/static/js/main.js
```diff
@@ -254,15 +254,12 @@ window.app = new window.Vue({
       return 'http://reposense.org';
     },
 
-    getRepoSenseVersionLink() {
+    getRepoSenseUserGuideVersionLink() {
       const version = window.app.repoSenseVersion;
       if (!version) {
         return 'https://github.com/reposense/RepoSense';
```
anubh-v: Should this link also be changed?

frontend/src/static/js/main.js
```diff
@@ -254,15 +254,12 @@ window.app = new window.Vue({
       return 'http://reposense.org';
     },
 
-    getRepoSenseVersionLink() {
+    getRepoSenseUserGuideVersionLink() {
```
anubh-v: I think it might be sufficient to use a shorter name - `getUserGuideLink` ?
From the code itself, it is relatively clear that the version is being used.

Tejas2805: I think it is better to be a bit explicit with the function name. Maybe, I can make it - `getUserGuideVersionLink`?

frontend/src/static/js/main.js
Tejas2805: If the version doesn't exist, the URL will not work properly. That's why I think we should keep this as it is.

anubh-v: Ok, understood.

frontend/src/static/js/main.js
anubh-v: >Maybe, I can make it - getUserGuideVersionLink?

I would personally still learn towards `getUserGuideLink`, but this is also ok.

Tejas2805: @fzdy1914 your views on this?

fzdy1914: `getUserGuideVersionLink` is better, since we may implement a direct link to the user guide in the future.

Tejas2805:
@anubh-v @jylee-git For your review.

Tejas2805:
> Looks good, but if link on landing page is misleading, maybe we should fix it as well?
Sorry, still a bit confused. Didn't exactly get what you were trying to say.

eugenepeh:
>Sorry, still a bit confused. Didn't exactly get what you were trying to say. @eugenepeh
Is it that we should point to the current build and not to the latest release ?

That was meant to address the original issue #964 as well as your proposed message:
>The link to the UG in the documents points to the released
version of the UG. As a result, there may be some discrepancies 
between the master and the release branch. 

Hence, if the link on the landing page (README.md) is confusing, maybe we should explicitly state both as shown in the above example.

jylee-git:
i.e.: update the `readme.md` as well to link to both the master and release branch of the UG

Tejas2805:
> Hence, if the link on the landing page (README.md) is confusing, maybe we should explicitly state both as shown in the above example.

Thinking of doing it for Developer Guide as well to maintain consistency.

Tejas2805:
> Though I'd personally like to have a visual text on the dashboard that states something along the line of `user guide here` or `guide on interpreting the report` as well, but I also understand the concern on cluttering the UI too much.

I have a suggestion for this:

Screen Shot 2020-01-29 at 12 35 17 AM

anubh-v:
If the commit hash is not visible on the dashboard, it may potentially be harder to document bugs.
We could embed the link to the user guide in the "RepoSense" text instead?

Tejas2805:
> If the commit hash is not visible on the dashboard, it may potentially be harder to document bugs.
> We could embed the link to the user guide in the "Generated by RepoSense" text instead?

`RepoSense` is already linked to the home site.

anubh-v:
> RepoSense is already linked to the home site.
Yes, I meant linking to the user guide instead of the home site.

jylee-git:
`Generated by RepoSense (User Guide - COMMITVER) on DATETIME` ?

Tejas2805:
> `Generated by RepoSense (User Guide - COMMITVER) on DATETIME` ?

Screen Shot 2020-01-29 at 4 06 52 PM

Tejas2805:
@damithc Which one do you think should we go ahead with?

damithc:
> Screen Shot 2020-01-29 at 4 06 52 PM

I'm OK with this version too.

Tejas2805:
@jylee-git For your review.

PR #991, opened 2020-01-15
[#988] Ramp chart: inconsistency in the length of contribution line

### LINKED ISSUE #988: Ramp chart: inconsistency in the length of contribution line
When toggling between the `breakdown by filetype` checkbox, the length of the contribution line is inconsistent even though the LoC remains constant.

Kapture 2020-01-14 at 23 24 12
Link to reproduce:  https://jamessspanggg.github.io/OpenSourceContribution/

0blivious:
Fixes #988
```
Toggling the `breakdown by filetype` checkbox changes the length of the
contribution bar even though the LoC remains constant.

This occurs due to the inaccurate data in json file. The
authorFileTypeContributionMap in commit.json have multiple entry
of the same key. In the frontend, the last entry would be taken
instead of all of the entries provided. This results in the wrong
calculation of total contribution of each file type, and can be
corrected by overridding FileType#hashCode

Let's override FileType#hashCode to correctly update the LoC
number instead of creating a new entry.

FYI, in the backend, we use AuthorshipSummary#addAuthorContributionCount
to increment line number and we consider FileType (the key) and
LoC (the value) as an entry in the HashMap. The HashMap is then output
to the commit.json file as authorFileTypeContributionMap. However, since
Java's native implementation of HashCode digest all data stored in
FileType, other data like path and pathsGlob is also taken into
consideration when comparing the key to update the entry.Thus in
AuthorshipSummary#authorFileTypeContributionMap, each author can have
multiple entry with same FileType#label. And since FileType#toString only
print FileType#label, the resulted authorFileTypeContributionMap in json
file will have multiple entry of the same key.
```


fzdy1914:
From your proposed commit message, the underlying issue is JS or Java?

yong24s:
@0blivious

Good fix. Let's take this chance to work on your commit message.

For the **first paragraph**:
```
Toggling the `breakdown by filetype` checkbox changes the length of the
contribution bar even though the LoC remains constant.
```
> Make it concise and capture the problem in the issue.

For the **second paragraph**:
```
This is caused by the inaccurate data in commit.json file.
```
> Good statement 👍.

Continued on **second paragraph**:
```
In the backend, we use AuthorshipSummary#addAuthorContributionCount
to increment line number and we consider filetype and path as an entry in
the HashMap.
Thus, in AuthorshipSummary#authorFileTypeContributionMap
each author can have multiple entry of the same key.
In the frontend, the last entry would be used instead of all of the entries
provided in ... .
This results in the wrong calculation of total contribution of each filetype,
and can be corrected by overridding FileType#hashCode.
```
> Preferably, reference **Class#Method** to help reviewers understand the changes.
> I'm not sure if this example fully capture the problem your described.
> Also note that my example is also not complete.
> Please try to add on and modified this paragraph on your own :)

Let's override FileType#hashCode to correctly update the LoC
number instead of creating a new entry.
> Take note of word choices.
There is a mixed usage of `filetype` and `file type`, `LoC` and `LOC`. Generally, use terms we frequently use in code or project.

It takes time to get good at writing commit messages. You are at a good start :-). Let me know if you need further clarifications.

0blivious:
@yong24s Thanks very much for helping me improve my commit message. Can help me check the new one? Is this one better now?

0blivious:
@eugenepeh Thx for the suggestion. I have shortened the 2nd graph and put technical details into the 4th graph. Could you help me check the new commit message?

PR #998, opened 2020-01-16
[#984] Fix miscalculation of expanded commit message count

### LINKED ISSUE #984: v_zoom: expandedCommitMessagesCount does not reflect the true expanded commit message.
We count the number of div with `commit-message active` class name to get the `expandedCommitMessagesCount`. However, reflected in the image below, block without a message body got assigned with a  `commit-message active` class name which counted into `expandedCommitMessagesCount`, resulting in an error.

image

jamessspanggg:
Fixes #984 

```
We get the number of expanded commit message count based on
the number of divs with class name 'commit-message active'.

Commit messages without commit message bodies also have 
default class name of 'commit-message active', which should 
not be the case. This leads to the expanded commit message 
count to not reflect on the true number of commit messages
that are expanded.

Let's append the class name 'message-body' to commit messages 
that have commit message bodies, to distinguish between 
messages that can be expanded and vice versa.
```


fzdy1914:
We get the number of expanded commit message count in zoom tab based on 
the number of divs with class name 'commit-message active'.

Commit messages without commit message bodies also have default class 
name of 'commit-message active', which leads to the expanded commit
message count to not reflect on the true number of commit messages that 
are expanded.

Let's append the class name 'message-body' to commit messages that have 
commit message bodies, to distinguish between messages that can be 
expanded and vice versa.

PR #1000, opened 2020-01-16
[#999] Remove empty zoom__day divs

### LINKED ISSUE #999: Commits Panel: redundant empty divs when granularity is set to week
When granularity is set to week, the commits panel shows empty `zoom__day` divs, which is unnecessary and complicates the HTML DOM.

Screen Shot 2020-01-17 at 12 19 11 AM

jamessspanggg:
Fixes #999 

```
When granularity is set to week, the 'pushCommitsWeek' method 
gets all possible weeks (week 1, 2, 3, etc...) from start date 
until the end date and pushes it to 'user.commits', no matter 
the week has any commits at all. Additionally, commits with 
zero insertions but with deletions are also not taken into 
consideration in the commits panel.

Weeks that have no commits will also be pushed into 
'user.commits', which is not useful and redundant. Commits with
only deletions are maybe useful as it represents
refactoring/removing redundant code.

Let's add a check to add a week only if it has commits. This 
will ensure that the 'user.commits' will have no weeks with zero 
commits. Let's also remove the check for commits with zero
insertions, so that commits with only deletions are included.
```


fzdy1914:
> Let's also add a check to only add a week's commits if the 
> commits are not empty. This will ensure that the
> 'filteredUser.commits' will have no commits with zero
> insertions.

Actually, zero insertion commit may got other usages. But I am OK to remove them.

Tejas2805:
> Actually, zero insertion commit may got other usages. But I am OK to remove them.

Even, I was thinking the same, maybe we can ask the Prof for his view on it and then accordingly decide to remove them or not?

fzdy1914:
> Actually, zero insertion commit may got other usages. But I am OK to remove them.
> 
> Even, I was thinking the same, maybe we can ask the Prof for his view on it and then accordingly decide to remove them or not?

@damithc @eugenepeh Can I get your thoughts?

damithc:
> @damithc @eugenepeh Can I get your thoughts?

I can give my inputs if this affects the user experience. Does it?

fzdy1914:
> I can give my inputs if this affects the user experience. Does it?

Currently, we will not display all commits with 0 insertions. Do you think we may have a situation in the future to display those commits with 0 insertions?

damithc:
> Currently, we will not display all commits with 0 insertions. Do you think we may have a situation in the future to display those commits with 0 insertions?

Does that mean we only consider insertions i.e., ignore deletions?

fzdy1914:
> Does that mean we only consider insertions i.e., ignore deletions?

Yes, for now, we only consider commits with at least 1 insertion. However, some commits like merging commits(resolve conflicts) will be ignored.

damithc:
Another question: in the current version, the size of the ramp is calculated based on insertions only?

frontend/src/static/js/v_summary.js
```diff
@@ -554,7 +554,9 @@ window.vSummary = {
         };
 
         this.addLineContributionWeek(endOfWeekMsWithinUntilMs, week, commits);
-        res.push(week);
+        if (week.commitResults.length > 0) {
+          res.push(week);
```
eugenepeh: is this change still necessary?

jamessspanggg:
> Actually, zero insertion commit may got other usages. But I am OK to remove them.

@fzdy1914 zero insertion commits will not be removed. I realised that I phrased my commit message wrongly. The current PR removes the weeks with zero commits, not commits with zero insertions.

frontend/src/static/js/v_summary.js
jamessspanggg: The reason why I added this is because currently, weeks' with zero commits are still added to the result (`res`). This change removes the weeks with zero commits as they are not useful.

eugenepeh: It is preferrable to have only 1 logic governs 1 behavior, either `if (week.commitResults.length > 0) {`  or  `.zoom__day(v-if="day.insertions > 0")`, not both. 
Reason: 
* Its duplicated / unnecessary logic
* The next dev (if it's not you) who is going to implement a change on this, going to waste some precious amount of time wondering about the weird behavior before realizing there's a second logic preventing the change

jamessspanggg: Yes I understand, but the logic is not really duplicated though.

My reasoning for implementing `if (week.commitResults.length > 0) {` is because there maybe weeks with no commits at all, so by having this condition we will not have empty weeks. By not having this condition will lead to unnecessary `<!---->` in the DOM:
Screen Shot 2020-01-23 at 9 41 31 PM

The `.zoom__day(v-if="day.insertions > 0")` is to not include commits with only deletions, as we only calculate the ramp size based on number of insertions. 

But I understand that it may confuse other devs on this. Should I remove the `if (week.commitResults.length > 0) {` logic?

eugenepeh: >The .zoom__day(v-if="day.insertions > 0") is to not include commits with only deletions, as we only calculate the ramp size based on number of insertions.

Hmm, ideally it should include all form of work done (including deletion).

@damithc should we put up an issue to add in the deletion commits? We can subsequently add a toggle that allows user to hide the deletion commits if its not relevant to them.

damithc: > @damithc should we put up an issue to add in the deletion commits? We can subsequently add a toggle that allows user to hide the deletion commits if its not relevant to them.

It is a good idea but I'm not sure how deep the ramification go. For example, translating lines deleted into a ramp size will be tricky. Should the ramp size reflect linesAdded + linesDeleted?

jamessspanggg: @eugenepeh in this case, should I remove the check for `.zoom__day(v-if="day.insertions > 0")` and put up an issue to modify the ramp size to reflect linesAdded + linesDeleted?

eugenepeh: Yes, please do remove the check. The commit view should show all commits including deletion.

For the ramp size, it would be tricky to include the lines deleted as highlighted by prof. I will put up an issue to discuss the modification.

jamessspanggg: Yep done so.

frontend/src/tabs/zoom.pug
```diff
@@ -39,8 +39,8 @@
     button(v-if="expandedCommitMessagesCount === 0" v-on:click="toggleAllCommitMessagesBody(true)") show all commit messages body
     button(v-else v-on:click="toggleAllCommitMessagesBody(false)") hide all commit messages body
 
-  .zoom__day(v-for="day in filteredUser.commits")
-    template(v-if="day.insertions > 0")
```
fzdy1914: https://github.com/reposense/RepoSense/blob/master/frontend/src/static/js/v_zoom.js#L27

This line should also be modified right?

fzdy1914:
@jamessspanggg Can you try to merge the commit message?

jamessspanggg:
> @jamessspanggg Can you try to merge the commit message?

done

jamessspanggg:
@fzdy1914 is it possible to merge this by this week?

Issue #960, opened 2019-12-10
Split CSS file into its relevant part


fzdy1914:
Part of #795.

As it is quite a huge ramp to directly projectify the whole frontend. We can take the process step by step. 

The final goal of #795 is to have all component related code into one file. So it is nice if we can split out the CSS/JS/HTML of one component into its own file first. Then, in the final combination part, we can reduce much effort. 

For now, the JS/HTML code of each component has already rested in its own file, we can explore to split the CSS file into its own part.

fzdy1914:
@eugenepeh @yong24s Your input, please?

yong24s:
Sounds good.

eugenepeh:
>@eugenepeh @yong24s Your input, please?

Yes, sounds great

fzdy1914:
Putting the priority to high to encourage developers to solve this issue.

PR #1003, opened 2020-01-18
[#960]Refactor Summary CSS

Tejas2805:
Fixes part of #960 

```
The JS/HTML code of each component has already rested in its 
own file, but all the CSS are in one file. 

Let's refactor the Summary CSS into a new file for better 
scalability and readability.
```


Tejas2805:
@fzdy1914 @jamessspanggg There are a lot of PRs regarding style which are currently in progress, maybe we can merge this PR as soon as possible as there is no PR affecting Summary CSS as soon as possible?

Tejas2805:
@reposense/stage1-reviewers @jamessspanggg @fzdy1914 For review.

jamessspanggg:
I'm wondering if we should delay the progress of these CSS refactoring to the near end of the semester, since there are a lot of ongoing PRs that touches on the front end.

frontend/src/static/css/v_summary.scss
```diff
@@ -0,0 +1,185 @@
...
+  &__contrib {
+    text-align: left;
+
+    &--bar {
+      background-color: mui-color('red');
+      height: .3rem;
+      margin-top: .1rem;
+    }
+
+    &--bar--fileType {
```
fzdy1914: Can you delete this block, please? It is deleted in a previous PR. Do CHECK that the block you delete and added back are exactly the same because there may be other merged PR touches the same file.

Tejas2805: Yup, deleted.

PR #1011, opened 2020-01-19
[#1001] CSS: Increase z-index for center divider div

### LINKED ISSUE #1001: Code segments and toggle buttons overlap on top of resizable tab
Observe that some code segments and toggle icons overlap on top of the resizable tab when scrolled across:

Kapture 2020-01-17 at 1 17 44

Probably just need to set the z-index for the resizable tab.

anubh-v:
Fixes #1001 

```
CSS: Increase z-index to 100, for all divs of the center divider

When we scroll horizontally in the Code Panel, the code snippets
and code toggle buttons overlap with the center divider.

This does not look visually appealing.

Let's fix this bug by increasing the
the z-index of the #tab-resize div, and the div that closes the right
tab. We can use a value of 100, since the z-index of #tab-resize-guide
div is already set to 100. For better scalability, let's use a SASS map
to store these z-index values.
```


anubh-v:
@jamessspanggg, thanks for the tip! I will investigate.

anubh-v:
We can use CSS variables, SASS variables or SASS maps to manage the `z-index` values (as opposed to hardcoding the `z-index` for each style rule).

I chose a SASS map as I think it is more compact than a list of variables.
However, I am not sure if we should keep the map in a separate file, or within `style.scss`.

@fzdy1914, @jamessspanggg, let me know if this method is appropriate? Thanks very much for reviewing.

Note: I also consolidated all the other `z-index` values being used in `style.scss` into 
`_z-indices.scss`, so that we can get a better idea of the benefit of this approach.

Tejas2805:
For my understanding, just want to know something. Why do we need a separate file for this?

jamessspanggg:
> For my understanding, just want to know something. Why do we need a separate file for this?

It's for easier management's sake as it may clutter and increase the size of `style.css` file if we do not separate it.

anubh-v:
@jamessspanggg , I've use SASS functions, as suggested.

I noticed that we need to validate the SASS map's key at a few points, so I abstracted out this validation logic into its own function in `_utils.scss` (to prevent repetition).

frontend/src/static/css/style.scss
```diff
@@ -115,7 +116,7 @@ header {
     position: fixed;
     right: $tabs-width;
     width: 2px;
-    z-index: 100;
+    z-index: map-get($z-indices, 'center-divider');
```
jamessspanggg: I think u forgot to use `z-index('centre-divider')` here

jamessspanggg:
> I noticed that we need to validate the SASS map's key at a few points, so I abstracted out this validation logic into its own function in `_utils.scss` (to prevent repetition).

Wondering if it's worth it to create a new `.scss` file to only store that validation logic. @fzdy1914 what do u think?

frontend/src/static/css/style.scss
anubh-v: Sorry, yes I missed this. Fixed.

anubh-v:
The Travis build has passed, but the icon isn't being updated on this PR.

I will update the most recent commit's message (in order to re-trigger the build).

image

fzdy1914:
> > I noticed that we need to validate the SASS map's key at a few points, so I abstracted out this validation logic into its own function in `_utils.scss` (to prevent repetition).
> 
> Wondering if it's worth it to create a new `.scss` file to only store that validation logic. @fzdy1914 what do u think?

I think it is fine because we may get other utils functions in the future.

anubh-v:
Thanks everyone for the advice on this PR!

fzdy1914:
When we scroll horizontally in the Code Panel, the code snippets and 
code toggle buttons overlap with the center divider.

Let's fix this bug by increasing the z-index of the #tab-resize div, and 
the div that closes the right tab. We can use a value of 100, since the 
z-index of #tab-resize-guide div is already set to 100. For better 
scalability, let's also use a SASS map to store these z-index values.

PR #987, opened 2020-01-06
[#522] Implement style checker on .pug files

Tejas2805:
Fixes #522 

```
RepoSense currently has CheckStyle, ESLint and Stylelint to check our java, 
javascript and css code styles automatically. However, .pug files are not 
covered by any of these tools.

Let's implement pug-lint to check the code styles of .pug files.

TODO: Add a line wrap rule in the future
```

PR #1025, opened 2020-01-29
[#892] Ramp chart: allow filtering by file type/group

### LINKED ISSUE #892: Ramp chart: allow filtering by file type/group
Current: code view can be filtered by file type/group but ramp chart cannot be filtered like that

image

Suggested: allow filtering the ramp chart by file type/group

Rationale: for example, to find who wrote most java code, one can filter out other file types and sort by contribution

0blivious:
Fix #892 

```
We have code view can be filtered by file type/group but ramp chart
in Chart Panel cannot be filtered and sorted like that.

This can be improved by only display selected portion of contribution
bar corresponding to selected file type. Also, if the user choose to sort
by contribution, only the file types in selected ones will be considered.

Let's add a set of file type check boxes in Chart panel and only display
and sort according to selected file types.
```


0blivious:
@fzdy1914 @damithc can I ask for your inputs on UI before further progress? I am trying some different types of UI on the page.
1. `group by Repo like`: the file selection check boxes will be displayed on the top, just below tool bar. Selected and relevant file type colored circle will be displayed under the group title.
2. `group by Author like`: the file selection check boxes and file type colored circle will be displayed on the top, just below tool bar.
3. We can make the file selection check boxes appear below each group title and the selection will affect contribution bar in that group only.

damithc:
> @fzdy1914 @damithc can I ask for your inputs on UI before further progress? I am trying some different types of UI on the page.

Can you some visuals/screenshots?

0blivious:
> Can you some visuals/screenshots?
Sure, updated. The first 2 can be tried on netlify preview by using corresponding `group by` option

damithc:
1. The filtering is for the entire chart, not for each group. That means it should appear only once at the top of the chart. 
2. In that one place we show the tick boxes, we can combine tickboxes and colors e.g., use the matching color as the background color for the file type. 
image

Would that work?

fzdy1914:
Or we can change the color beside the file type into a radio button with the corresponding color?

I believe the color is repeated in all group because it is inconvenient to check each time on top to determine the file type for the color .

0blivious:
> Or we can change the color beside the file type into a radio button with the corresponding color?

I tried but the tick may not become so visible if we fill the radio button with color.

> I believe the color is repeated in all group because it is inconvenient to check each time on top to determine the file type for the color .

Yes I agree with that. However, if we group by none there will be no more groups. And I feel it is not very good to display the file type color on top above every author. A better solution might be that we fix the toolbar (involve the search, date picker, file type selection, etc) all on the top of the page. How do you feel about this?

0blivious:
> Would that work?

We might need to change all text color to black as the current file type color has some dark ones.
See the `gradle` and `md` type below, which one do you feel better?
image
Or we can change the text color instead of the background color.
image




> Or we can change the color beside the file type into a radio button with the corresponding color?

I tried this but does not looks very good. The tick inside becomes not so visible.

damithc:
Did you try this? 
image
The text color needs to be black or white based on how dark the bg color is (similar to how GitHub labels do it)

0blivious:
image
This one ok?

damithc:
> image
> This one ok?

OK for me.

0blivious:
>[#892] may also be solved in this PR. If we select certain file types, the sorting option will only apply to these file types. e.g., if we select only java, sort within groups by contribution will display the person write most java code on top instead of the person with most commits.

For #892, is it preferred I do it in this PR, or open another one after this is merged?

frontend/src/index.pug
```diff
@@ -170,32 +170,48 @@ html
...
             .summary-charts__fileType--breakdown(v-if="filterBreakdown")
-              template(v-if="filterGroupSelection === 'groupByRepos'")
-                .summary-charts__fileType--breakdown__legend(
-                    v-for="fileType in getFileTypes(repo)")
-                  i.fas.fa-circle(
-                      v-bind:style="{ 'color' : contributionBarFileTypeColors[fileType] }"
-                  )
-                  span &nbsp; {{ fileType }} &nbsp;
-              template(v-else)
+              template
```
jamessspanggg: is the empty template necessary

frontend/src/index.pug
```diff
@@ -109,8 +109,8 @@ html
               label sort groups by
             .mui-select.sort-within-group
               select(
-                  v-model="sortWithinGroupSelection",
-                  v-bind:disabled="filterGroupSelection === 'groupByNone' || isMergeGroup"
+                v-model="sortWithinGroupSelection",
```
jamessspanggg: The spacing changes aren't necessary.

frontend/src/index.pug
```diff
@@ -170,32 +170,48 @@ html
...
-                    v-for="fileType in getFileTypes(repo)")
-                  i.fas.fa-circle(
-                      v-bind:style="{ 'color' : contributionBarFileTypeColors[fileType] }"
-                  )
-                  span &nbsp; {{ fileType }} &nbsp;
-              template(v-else)
+              template
                 .summary-charts__fileType--breakdown__legend(
-                    v-for="fileType in Object.keys(contributionBarFileTypeColors)")
+                  v-if="filterGroupSelection !== 'groupByNone'"
```
jamessspanggg: Vue does not recommend putting v-for and v-if together in the same div, check: https://vuejs.org/v2/guide/conditional.html#v-if-with-v-for

frontend/src/index.pug
fzdy1914: Actually, as the standard discussed in #987, the change here is fine.

0blivious: Can we have a PR that fix all the spacing in pug? Otherwise very troublesome to use auto-indent when writing code.

fzdy1914: Yes, #987 is targeting the issue. Just be patient for it to be merged.

frontend/src/index.pug
```diff
@@ -170,6 +170,27 @@ html
               )
                 span {{ getReportIssueEmailAddress() }}
             div.error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}
+        .fileTypes(v-if="filterBreakdown")
+          div.checkboxes.mui-form--inline(v-if="Object.keys(contributionBarFileTypeColors).length > 0")
+            label(
+              v-bind:style="{ 'background-color' : '#000000', 'color' : '#ffffff' }"
+            )
+              input.mui-checkbox--fileType(type="checkbox" v-model="checkAllFileTypes")
+              span
```
jamessspanggg: Is this redundant?

frontend/src/index.pug
```diff
@@ -170,6 +170,27 @@ html
...
+                span.select.medium-font All:&nbsp;
+            template(v-for="fileType in Object.keys(checkBoxFileTypeFontColors)" )
+              label(
+                v-bind:key="fileType"
+                v-bind:style="{ 'background-color' : contributionBarFileTypeColors[fileType]," +
+                "'color' : checkBoxFileTypeFontColors[fileType]}"
```
jamessspanggg: line wrap indentation

frontend/src/index.pug
```diff
@@ -184,18 +205,12 @@ html
...
-                    v-for="fileType in getFileTypes(repo)")
-                  i.fas.fa-circle(
-                      v-bind:style="{ 'color' : contributionBarFileTypeColors[fileType] }"
-                  )
-                  span &nbsp; {{ fileType }} &nbsp;
-              template(v-else)
+              template(v-if="filterGroupSelection !== 'groupByNone'")
                 .summary-charts__fileType--breakdown__legend(
-                    v-for="fileType in Object.keys(contributionBarFileTypeColors)")
+                  v-for="fileType in getFileTypes(repo)" :key="fileType"
```
jamessspanggg: is it possible to standardise the `v-bind` usage? Rather than using the shorthand

frontend/src/static/js/v_summary.js
```diff
@@ -565,8 +603,7 @@ window.vSummary = {
       // commits are not contiguous, meaning there are gaps of days without
       // commits, so we are going to check each commit's date and make sure
       // it is within the duration of a week
-      while (commits.length > 0
-          && (new Date(commits[0].date)).getTime() <= endOfWeekMs) {
+      while (commits.length > 0 && (new Date(commits[0].date)).getTime() <= endOfWeekMs) {
```
jamessspanggg: is this change needed?

frontend/src/index.pug
0blivious: Need both

frontend/src/static/js/v_summary.js
```diff
@@ -516,10 +534,30 @@ window.vSummary = {
...
         });
         this.contributionBarFileTypeColors = fileTypeColors;
       });
+
+      Object.keys(this.contributionBarFileTypeColors).forEach((fileType) => {
+        const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(this.contributionBarFileTypeColors[fileType]);
+        const red = parseInt(result[1], 16);
+        const green = parseInt(result[2], 16);
+        const blue = parseInt(result[2], 16);
+
```
Tejas2805: Just wish to understand. Why are both `green` and `blue` using `result[2]`?

0blivious: Fixed

frontend/src/index.pug
```diff
@@ -171,6 +171,27 @@ html
...
             div.error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}
+        .fileTypes(v-if="filterBreakdown")
+          div.checkboxes.mui-form--inline(v-if="Object.keys(contributionBarFileTypeColors).length > 0")
+            label(
+              v-bind:style="{ 'background-color' : '#000000', 'color' : '#ffffff' }"
+            )
+              input.mui-checkbox--fileType(type="checkbox" v-model="checkAllFileTypes")
+              span
+                span.select.medium-font All:&nbsp;
+            template(v-for="fileType in Object.keys(checkBoxFileTypeFontColors)" )
```
jamessspanggg: Perhaps you can use `Object.entries` here?

frontend/src/static/js/v_summary.js
```diff
@@ -520,10 +538,30 @@ window.vSummary = {
...
+            if (!this.fileTypes.includes(fileType)) {
+              this.fileTypes.push(fileType);
+            }
           });
         });
         this.contributionBarFileTypeColors = fileTypeColors;
       });
+
+      Object.keys(this.contributionBarFileTypeColors).forEach((fileType) => {
+        const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(this.contributionBarFileTypeColors[fileType]);
```
jamessspanggg: can leave a comment on what this line does?

frontend/src/static/js/v_summary.js
```diff
@@ -520,10 +538,30 @@ window.vSummary = {
...
+        const green = parseInt(result[2], 16);
+        const blue = parseInt(result[3], 16);
+
+        const luminosity = 0.2126 * red + 0.7152 * green + 0.0722 * blue; // per ITU-R BT.709
+
+        if (luminosity < 120) {
+          this.checkBoxFileTypeFontColors[fileType] = '#ffffff';
+        } else {
+          this.checkBoxFileTypeFontColors[fileType] = '#000000';
+        }
```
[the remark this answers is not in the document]
jamessspanggg: `this.checkBoxFileTypeFontColors[fileType] = luminosity < 120 ? '#ffffff' : '#000000'`

frontend/src/static/js/v_summary.js
```diff
@@ -570,7 +608,7 @@ window.vSummary = {
       // commits, so we are going to check each commit's date and make sure
       // it is within the duration of a week
       while (commits.length > 0
-          && (new Date(commits[0].date)).getTime() <= endOfWeekMs) {
+      && (new Date(commits[0].date)).getTime() <= endOfWeekMs) {
         const commit = commits.shift();
         week.insertions += commit.insertions;
         week.deletions += commit.deletions;
```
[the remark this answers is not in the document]
0blivious: error due to auto indent

jylee-git:
Fix the checkstyle errors in the pug files first. I think a PR that enables pug file checkstyle was merged recently.

0blivious:
> Fix the checkstyle errors in the pug files first. I think a PR that enables pug file checkstyle was merged recently.

Fixed.

jylee-git:
There seems to be a bug here

* Before un-checking `backend`
image

* After unchecking `backend` : notice that backend icon changes from pink to green color
image

damithc:
image

Something is not right here. These two should match. We either use file extensions or custom groups, not both together.

0blivious:
> Something is not right here. These two should match. We either use file extensions or custom groups, not both together.

The reason is that at the top we display all file type labels of all group and for `group.csv` some repos dont have this customised group. As seen below, `testrepo-Alpha` doesn't have any customised group assigned. That's the reason why we have labels like 'json' on top.

image

damithc:
> The reason is that at the top we display all file type labels of all group and for `group.csv` some repos dont have this customised group. As seen below, `testrepo-Alpha` doesn't have any customised group assigned. That's the reason why we have labels like 'json' on top.

Got it.

0blivious:
@jylee-git bug fixed

fzdy1914:
Seems that the review is done?

The report showed in the netlify review does not reflect your change?

0blivious:
> Seems that the review is done?
> 
> The report showed in the netlify review does not reflect your change?

What change is you referring to?

frontend/src/static/css/v_summary.scss
```diff
@@ -2,6 +2,26 @@
 
 /* summary chart */
 #summary,
+.fileTypes {
```
fzdy1914: The indentation doesn't seems to be right here.

0blivious: If we follow the indentation in other css then means I probably need to shift everything below #summary to right by 2 space though (all other style inside #summary)......Should I do it?

Tejas2805:
> For this feature to have more added value, perhaps the sort function should be based on the new summed value of the contribution bar?
> 
> image

Makes sense. How about we do this in another PR?

0blivious:
> For this feature to have more added value, perhaps the sort function should be based on the new summed value of the contribution bar?

The draft for sort is already opened in #1091, just need to resolve conflicts. Would you like it to be done in this PR?

eugenepeh:
>The draft for sort is already opened in #1091, just need to resolve conflicts. Would you like it to be done in this PR?

Hmm, then it would be better for them to be merged together into master at the same time. Reason of it being that releasing incomplete features are never an acceptable practice in the industry; it tarnishes the image of the product. Even though we have version release, it doesn't mean that it would allow us to merge incomplete features because that would also mean that we can't ever release a new version for as long as the feature remains incomplete.

If there's the consideration where it might introduce too much changes to be reviewed, can create a temporary branch e.g. 892 and merge them into it one by one. The 892 can then subsequently be merged into master with less nit picking as all of the work are considered reviewed.

0blivious:
Sorting by file type is implemented in this PR.

jylee-git:
Something's wrong with the netlify build. Can you re-trigger it?

Meanwhile, are you going to include the contents in #1091 into this PR as well?

0blivious:
> Meanwhile, are you going to include the contents in #1091 into this PR as well?

Have add the content in alr, will close #1091

jylee-git:
For the sorting part, will it be **internally** sorted as well?

I see that sorting by group works, but not internally

0blivious:
> For the sorting part, will it be **internally** sorted as well?

Fixed

frontend/src/static/css/v_summary.scss
```diff
@@ -1,189 +1,209 @@
...
-  &__section {
-    align-items: inherit;
-    display: flex;
-    flex: 0 1 auto;
-    flex-flow: inherit;
-    justify-content: inherit;
-  }
+/* Summary */
+#summary {
+  .fileTypes {
```
jamessspanggg: Is it possible to place your new css selectors at the bottom such that the other components won't be shown with changes.

0blivious: I actually did shift everything 2 space to right cause the original css is not really correct.
```
#summary,
.sth{}
.sth{}
```
So everything below is not under `#summary`

jamessspanggg: Ahh I see. My bad for not spotting that.

jamessspanggg:
@fzdy1914 @eugenepeh may I get your reviews on this PR? We intend to include this change for the upcoming release.

Tejas2805:
Also, I think the fact that it will get sorted based on the files actively selected should be mentioned in the UG too.

jamessspanggg:
> Just a question, why the CSS is changed in such a large way?

The following is the reason:

I actually did shift everything 2 space to right cause the original css is not really correct.
```
#summary,
.sth{}
.sth{}
```
So everything below is not under `#summary`

_Originally posted by @0blivious in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjQwNDA2NjQzOnYy/pull_request_review_threads/discussion_

jamessspanggg-dsc:
just a question: changes in #1148 won't really affect this PR right, as they all obtain the colours from fileTypeColors?

damithc:
Another thing: file type selection can be encoded in the URL so that a specific file selection can be shared through the URL. That too can be a separate PR.

0blivious:
> just a question: changes in #1148 won't really affect this PR right, as they all obtain the colours from fileTypeColors?

wont. I just obtain color from fileTypeColors while you are changing the generation scheme

0blivious:
> Another thing: file type selection can be encoded in the URL so that a specific file selection can be shared through the URL. That too can be a separate PR.

Noted, will open new PRs for these two.

fzdy1914:
The codacy check fails here is not important right?

0blivious:
> The codacy check fails here is not important right?

It is due to the new cypress testing. I follow the previous naming convention. Dont think there is really violation.

fzdy1914:
OK, got it.

fzdy1914:
Will merge this PR soon.

fzdy1914:
Can you help retrigger the build?

PR #1028, opened 2020-01-30
[#1027] Modify UI for tags

### LINKED ISSUE #1027: Weird arrangement with overlonged tag list in Zoom tab
image

jamessspanggg:
Fixes #1027 

```
Some tags split to the next line when the list of tags are 
too long. Tags also appear vertically close together when 
the list occupies more than one line.

Let's change the css property 'display' to be inline-block
and add some top and bottom margins to each tag.
```


0blivious:
LGTM, but I feel a bit more spacing will look nicer. Also, I feel it is better to align the left side of the left most tag with the blue rectangle side. The tags also appear to be closer to the rectangle than to the sentence above?
image

jamessspanggg:
@0blivious have added more margins to the tags. 

> The tags also appear to be closer to the rectangle than to the sentence above?

Actually the top and bottom margins are the same, but the tags appear closer to the sentence is because the sentence does not fully occupy the bottom of it's own div.

Screen Shot 2020-01-30 at 10 44 19 PM

PR #1033, opened 2020-01-31
[#1021]: Use BASE_URL constant instead of string 

### LINKED ISSUE #1021: Make base URL a variable
As discussed in PR #1017 , `https://gtihub.com` is used at several points in our js files.
Let's keep the base URL as a variable.

yash-chowdhary:
Fixes #1021 

```
There are multiple occurrences of the string 'https://github.com'.

Let's replace the string with a constant 'BASE_URL' in the files.
```


yash-chowdhary:
My PR fails the cypress frontend tests because of the BASE_URL `const` variable being declared in multiple locations. Using `let` or `var` causes a lint error as it expects `const` to be used instead. There's a workaround where the `const` can be declared in the function blocks, but this defeats the purpose of the PR. 
I've even tried creating a `constants.js` file but cypress complains about `import` statements not being allowed outside modules. Does anyone have any suggestions or workarounds?

jamessspanggg:
Hi @yash-chowdhary, thanks for making the PR! Can you follow the commit message conventions in [link](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message)? You can also look at the existing PRs as a reference.

> My PR fails the cypress frontend tests because of the BASE_URL `const` variable being declared in multiple locations. 

Would declaring the BASE_URL in main.js once as `window.BASE_URL` work? In this way, all other locations are able to access the url.

jamessspanggg:
hi @yash-chowdhary can you update ur commit message?

yash-chowdhary:
Hi @jamessspanggg, thanks for the tip on hot to fix the issue with the test! Regarding [the comment above](https://github.com/reposense/RepoSense/pull/1033#issuecomment-581047169), would you like me to change the individual commits or did you mean the PR title and description?

Tejas2805:
> Did you mean the PR title and description?
Yes.  Also, @yash-chowdhary You can refer to [page](https://github.com/reposense/RepoSense/blob/master/docs/Process.md#how-to-contribute-to-the-reposense-repository) to understand how to write the commit message for the PR.

frontend/src/static/js/v_summary.js
```diff
@@ -100,6 +100,7 @@ function dateRounding(datestr, roundDown) {
   return getDateStr(datems);
 }
 
+
```
Tejas2805: Can you remove this extra line too

yash-chowdhary:
> There are other un-replaced BASE_URL in the files, can you please also help replace them? In `v_authorship` and `v_ramp`.

I noticed that these two files make use of `http` instead of `https`. Should we standardize it to `https`?

Tejas2805:
> I noticed that these two files make use of `http` instead of `https`. Should we standardize it to `https`?
Can.

PR #1034, opened 2020-01-31
[#1031] Commits panel: ramp appears outside row 

### LINKED ISSUE #1031: Commits panel: ramp appears outside row
In Commits panel, the left part of the ramp goes out of the blue rectangle as shown in the image below. It should be cropped like what we did for the Chart panel.

image

0blivious:
Fixes #1031
```
The ramp slice in Commits panel appear outside of the blue rectangle.

We should crop the ramp slice if it appears outside of the row, like
what we did for the Chart panel.

Let's add set the overflow property to be hidden for the ramp div.
```


jamessspanggg:
> We should crop the ramp slice if it appears outside of the row, like
> what we did for the Code panel.

Do you mean Chart panel?

0blivious:
> Do you mean Chart panel?

No, I mean the Commits/Zoom Panel. See the red triangle on the right.
image

jamessspanggg:
As in, your commit message, should it be Chart panel?

> what we did for the Code panel.

jamessspanggg:
This indeed solves the problem. 

However, ramps in chart panel do not require the `overflow:hidden` property for it to not overflow. Can you look at why the ramps in chart panel won't overflow whereas the ramps in zoom panel will?

0blivious:
> This indeed solves the problem.
> 
> However, ramps in chart panel do not require the `overflow:hidden` property for it to not overflow. Can you look at why the ramps in chart panel won't overflow whereas the ramps in zoom panel will?

Cause the div `summary-chart` has overflow: auto

jamessspanggg:
> Cause the div `summary-chart` has overflow: auto

In that case, you may remove the overflow: auto property in summary-chart, since it's redundant if you put it in ramp as well.

0blivious:
> In that case, you may remove the `overflow: auto` property in summary-chart, since it's redundant 

It is needed for the chart title, for eg an author with super long name.

jamessspanggg:
> It is needed for the chart title, for eg an author with super long name.

Tried reducing the size of the window such that the author name occupies multiple lines and removing the property:
Kapture 2020-01-31 at 19 30 23

0blivious:
It does produce some difference in scroll bars on my computer though. I use windows and chrome.

with overflow:auto
image


without overflow:auto
image

jylee-git:
Request from me again after resolving the conflict

0blivious:
@jylee-git conflict resolved

PR #1035, opened 2020-01-31
[#823] AnnotatorAnalyzer: Fix code disowning mechanism

### LINKED ISSUE #823: AnnotatorAnalyzer: Bug in code disowning mechanism
### Refer to the report here: https://reposense-issue-823.netlify.com

image

 git-id `some-randomuser` did not make any commits to that repo. Should the following 3 lines (7-9) still be tagged to the person who wrote the code?

image

Should the `@@author` tag (line 13) be tagged to an author? Even so, the tag is written to `rjcai-tee` who did not write the `@@author` tag.

anubh-v:
Fixes #823 

```
AnnotatorAnalyzer: Fix code disowning mechanism

Suppose that "random-user" is not the git author name of any author
in a git repository (i.e. it is just a placeholder name).

RepoSense does not correctly mark code as disowned when a user marks
a code block with the "@@author random-user" tag.
Only the line containing the "@@author random-user" tag is treated
as disowned, while the remaining code block is still attributed to
the author deduced from the git blame analysis.

This bug occurs because the code in AnnotatorAnalyzer uses
a constant UNKNOWN_AUTHOR to represent both the case where
there is no currently active annotation override AND the case
where the user has used a "@@author random-user" tag.

Let's use Optionals to wrap the author names obtain from the
@@author tags, so we can identify the case where user has used
a "@@author random-user" tag.
```


anubh-v:
The system tests' expected output is in line with the existing behaviour of AnnotatorAnalyzer, but we are modifying this behaviour in this PR. This is causing the system tests to fail.

I will document the problem so we can move forward.

anubh-v:
@fzdy1914, for this PR, I would need to edit the expected output from the system tests.

Should I make the changes first and then ask for a review?

jylee-git:
> @jylee-git, @fzdy1914, @eugenepeh , for this PR, I would need to edit the expected output from the system tests.
> 
> Should I make the changes first and then ask for a review?
> 
> Alternatively, I could fix this bug on a separate branch and incrementally fix the system tests' output over several smaller PRs to that branch. This may be better as we can review one system test at a time.

Yes you can, for PRs that deal with bug fixing.

anubh-v:
> Yes you can, for PRs that deal with bug fixing.

@jylee-git, thanks for your response. To clarify, do you mean I should make a new branch for this bug fix and then make PRs against that branch? Or would you prefer I make all changes (bugfix + systemtest updates) in this PR?

anubh-v:
Perhaps we should discuss the implementation first, before I update all the expected output for the system tests?

jylee-git:
> > Yes you can, for PRs that deal with bug fixing.
> 
> @jylee-git, thanks for your response.
> 
> To clarify, do you mean I should make a new branch for this bug fix and then make PRs against that branch? Or would you prefer I make all changes (bugfix + systemtest updates) in this PR?

My bad about not clarifying this properly.

What i meant was that you should update the expected results from the systemtest in the same PR as the bugfix as well. This applies to others (like features, etc).

You should also update all the expected results from test cases/system test before asking for a review

anubh-v:
Thanks for the response @jylee-git 
I'm closing this PR, because I'll be working on this bug fix on a separate branch, as discussed with @fzdy1914.

We'll need to make non trivial changes to the system tests' expected output. Hence, it'll be easier to review the changes over multiple PRs.

PR #1008, opened 2020-01-19
[#1005,#1010] Code Panel: match font with GitHub and break lines when overflow

jamessspanggg:
Fixes #1005 and #1010 

```
When the file path or lines of code overflows, it takes up a 
large amount of empty space in order to accomodate to 
it's length. The font size of the code and file path is also 
rather big and takes up much space.

Let's set the file path and lines of code to break into next 
lines when it overflows. Also, let's set the font-family 
and font-size of the code to match with GitHub.
```

- [x] break file path or lines of code to next line when overflow
- [x] match font family and font size with GitHub's
- [x] decrease the size of the file path
- [x] ~~match the font sizes of group title in summary panel and file title in code panel~~
- [x] ~~match the font styles and sizes between code panel and commit panel~~

PR #1038, opened 2020-01-31
[#1037] Add word-wrap break-word to error box

### LINKED ISSUE #1037: Summary Page: Error message extends out of red box if box is too thin
image

The issue happens when the right panel is shortened too much.

ccyccyccy:
Fixes #1037 
```
The error message appears outside of the red error box when the word is
too long, and the red box is too thin.

We should fit the entire message into the red box, while still allowing
users to view the full message.

Let's use overflow: auto for the error message box CSS.
```


ccyccyccy:
@eugenepeh Thanks for the suggestion. I changed it to use `overflow: auto` so that users can scroll to view the entire message instead of having it all mangled up.

eugenepeh:
>@eugenepeh Thanks for the suggestion. I changed it to use overflow: auto so that users can scroll to view the entire message instead of having it all mangled 

Looks good but I think it might look better if we let the overflow happens at panel level (e.g. the summary view), where the error box will have a minimum width to fit the content, while the summary view will overflow to keep the full error box in view.

Something like that instead
image

But you might need to put in more work to ensure that a gap (between error box and end of the ramp chart) like this doesn't happen
image

eugenepeh:
Current solution is good but it somehow creates a chunk of underutilized space, so if you can address this, I am fine with the current solution too.

image

Tejas2805:
@ccyccyccy Any update on this?

Tejas2805:
@ccyccyccy What's the update on this. Maybe you can look into #1008 regarding overflow and try to solve this issue.

ccyccyccy:
I'm trying to make the overflow happen at a panel level, but facing troubles getting the CSS to work.

> Maybe you can look into #1008 regarding overflow and try to solve this issue.

Will take a look at that.

fzdy1914:

PR #1039, opened 2020-01-31
[#1022] FileInfoAnalyzer: Deprecate "//@reused" tag

### LINKED ISSUE #1022: FileInfoAnalyzer: Deprecate the "//@reused" tag
Currently, `FileInfoAnalyzer` will not analyse a file if the first line of the file has the `//@reused` string.
The file will not be included in the final Code Panel view.

May I check what is the meaning of this `//@reused` mechanism?
Is it meant as a way to omit files?

Relevant code snippets:
https://github.com/reposense/RepoSense/blob/6fe782466dc827ab5c37c095000dfaba6e2ee3db/src/main/java/reposense/authorship/FileInfoAnalyzer.java#L27-L33

https://github.com/reposense/RepoSense/blob/6fe782466dc827ab5c37c095000dfaba6e2ee3db/src/main/java/reposense/authorship/FileInfoAnalyzer.java#L49-L53

https://github.com/reposense/RepoSense/blob/6fe782466dc827ab5c37c095000dfaba6e2ee3db/src/main/java/reposense/authorship/FileInfoAnalyzer.java#L129-L143

**Some investigation**
After some searching, I found that this mechanism was introduced in [this commit](https://github.com/reposense/RepoSense/commit/55787167ef17fc4c0979e2e27c41c416d9299093).
Back then, the file analysis logic was in `FileAnalyzer`.
`FileAnalyzer` was deleted and `FileInfoAnalyzer` took its place in [this commit](https://github.com/reposense/RepoSense/commit/87571e26e9792c62b602b2760906db234b68b3c7).

Please correct me if I'm wrong about the history here.

anubh-v:
Fixes #1022 

```
FileInfoAnalyzer: Deprecate "//@reused" tag

RepoSense will not include a file in the authorship analysis
if the first line of the file contains the "//@reused" string.

This feature is not being used.

Let's remove this feature to simplify the codebase.
```


src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -43,12 +42,13 @@
...
      * by {@code config}.
-     * Returns null if the file contains the reused tag, the file is missing from the local system, or none of the
+     * Returns null if the file is missing from the local system, or none of the
      * {@code Author} specified in {@code config} contributed to the file in {@code fileInfo}.
      */
     public static FileResult analyzeFile(RepoConfiguration config, FileInfo fileInfo) {
         String relativePath = fileInfo.getPath();
-        if (isReused(config.getRepoRoot(), relativePath)) {
+
+        if (isEmptyFile(config.getRepoRoot(), relativePath)) {
```
anubh-v: This `isReused` method was originally checking 2 things:
(1) if the file is empty
(2) if the first line of the file contains the `//@reused` string

We would still want to keep the first check. However, shouldn't we check **whether the file exists** before checking if the file is empty? Currently, the `analyzeFile` method checks if the file is empty **before** confirming that the file exists.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -127,13 +127,13 @@ private static String getGitBlameResult(RepoConfiguration config, String filePat
...
     /**
-     * Returns true if the first line in the file at {@code repoRoot}'s {@code relativePath} contains the reused tag.
+     * Returns true if file at {@code repoRoot}'s {@code relativePath} has 0 lines.
      */
-    private static boolean isReused(String repoRoot, String relativePath) {
+    private static boolean isEmptyFile(String repoRoot, String relativePath) {
         Path path = Paths.get(repoRoot, relativePath);
         try (BufferedReader br = new BufferedReader(new FileReader(path.toFile()))) {
             String firstLine = br.readLine();
-            if (firstLine == null || firstLine.contains(REUSED_TAG)) {
```
anubh-v: Instead of reading the first line of the file and checking if the `BufferedReader` returns null, should we use other simpler methods of verifying if the file is empty?

For example, we could use the `length` method of `java.io.File`.
Please see [the java docs here](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#length--)

anubh-v:
@fzdy1914, @eugenepeh, I've removed the `@reused` tag feature.
However, I have some questions about possible code improvements, which I have listed in my comments above.

anubh-v:
@fzdy1914, thanks very much for reviewing our work.
May I get your thoughts on this PR as well?

src/main/java/reposense/authorship/FileInfoAnalyzer.java
fzdy1914: Yes, I agree with that. We should add a method in FileUtil to check.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
fzdy1914: I believe it should be checked somewhere in the previous step, but you can recheck.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
anubh-v: Done.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
anubh-v: Done - I've placed the check for file existence before checking that the file is non-empty.

anubh-v:
@fzdy1914, I've made the changes discussed. Ready for further review.

fzdy1914:
@reposense/members Final call to review this PR.

PR #1040, opened 2020-02-01
[#1023] Authorship Panel: expand all button always appear when first open the page

### LINKED ISSUE #1023: Authorship Panel: expand all button always appear when first open the page
Currently, the `expand all` button will always appear on the authorship panel when we first open the page or when refresh the page, until the loading is complete. The button to display is supposed to be `collapse all` as we are to display all the file details initially.

Screenshot (14)

After the loading is completed, the button will change to `collapse all` all of a sudden. We could modify the js file to make the `collapse all` appears from the very beginning, even before the loading is completed.

0blivious:
Fixes #1023
```
The `expand all` button will always appear on the authorship
panel when we first open the page or when refresh the page, until the
loading is complete. The button to display is supposed to be
`collapse all` as we are to display all the file details initially.

This can be `collapse all` appears from the very beginning.

Let's display the `collapse all` button at the very beginning, and update
all relevant checkboxes, radio buttons using proper Vue model.
```


Tejas2805:
@0blivious We don't use `Currently` or `Now` in the commit messages. Can you please update it accordingly.

0blivious:
While doing this PR, I realize some other issue in the button. 

1. The selectAll checkbox for file type will always be ticked even when we untick all other file type boxes. I feel we should untick the `all` once we untick any other file type check box.
image

2. The logic in this page is, when there is at least one active class, the button should be `Collapse all`. However, if we select a single file type, collapse the file details, then we tick another file type, the button will be `Expand all`.
image

The current PR here resolves both issue I mentioned above. I am not very sure if I need to open another PR for these issues.

fzdy1914:
> 1. The selectAll checkbox for file type will always be ticked even when we untick all other file type boxes. I feel we should untick the `all` once we untick any other file type check box.

It is truly a bug, however, the new logic should be untick the `all` if we untick *ALL* other file type check box.

fzdy1914:
For the second point, it is a good fix and the logic seems to be reasonable.

frontend/src/static/js/v_authorship.js
```diff
@@ -40,21 +41,40 @@ window.vAuthorship = {
...
       sortingFunction: (a, b) => -1 * window.comparator(filesSortDict.lineOfCode)(a, b),
-      isSearchBar: false,
-      isCheckBoxes: true,
     };
   },
 
   watch: {
+    selectedFileTypes() {
+      if (this.selectedFileTypes.length === this.fileTypes.length) {
+        this.isSelectAllChecked = true;
```
fzdy1914: Can make it a computed value?

0blivious: Here the new logic is: 1. untick the all if we untick ALL other file type check box. 2. tick `all` if we manually tick all other file type. For this reason, it is impossible to make it a computed value.

fzdy1914: OK, got your point.

frontend/src/static/js/v_authorship.js
```diff
@@ -40,21 +41,40 @@ window.vAuthorship = {
...
+      }
+    },
     filesSortType() {
       this.sortFiles();
     },
     toReverseSortFiles() {
       this.sortFiles();
     },
+    selectedFiles() {
+      setTimeout(this.updateCount, 0);
```
fzdy1914: Last question, what does this do?

0blivious: To drag the `updateCount` function to the end of the event queue so we render the file messages first. I kept it from the previous code, not very sure if it is still relevant now.

fzdy1914: OK

0blivious:
> Can you also include the logic that if one is not ticked, the `All` checkbox doesn't remain ticked anymore?
> Screen Shot 2020-02-12 at 4 01 15 AM

As requested by @fzdy1914, the new logic should be untick the all if we untick ALL other file type check box.

Tejas2805:
> As requested by @fzdy1914, the new logic should be untick the all if we untick ALL other file type check box.

I agree with that. But I also think that having `All` ticked when all are not files are not display is a misnomer and can be thought of as a bug

0blivious:
@damithc Prof may I ask what is your view on this tick all logic?

damithc:
> I agree with that. But I also think that having `All` ticked when all are not files are not display is a misnomer and can be thought of as a bug

Not clear. Can you elaborate?

fzdy1914:
> > I agree with that. But I also think that having `All` ticked when all are not files are not display is a misnomer and can be thought of as a bug
> 
> Not clear. Can you elaborate?

The issue is that, as our previous implementation, the `All` button will be unticked only when all other buttons are unticked. 

However, @Tejas2805 suggested that, the `All` button will be unticked once an other buttons are unticked.

I am actually OK with both, but i feel the previous logic is more convenient during the usage.

damithc:
> I am actually OK with both, but i feel the previous logic is more convenient during the usage.

I think it's common to untick the `all` box if at least one item is not selected.

Tejas2805:
Step to reproduce:
1) Click on `hide all file details`
2) Untick `frontend`
3) Tick `frontend`
4) Scroll all the way to the bottom

Expected result: Files should still be hidden due to Step 1
Actual result: Files affected by Step 3 are not hidden

Screen Shot 2020-02-15 at 12 49 15 AM

0blivious:
> Step to reproduce:
> 
> 1. Click on `hide all file details`
> 2. Untick `frontend`
> 3. Tick `frontend`
> 4. Scroll all the way to the bottom
> 
> Expected result: Files should still be hidden due to Step 1
> Actual result: Files affected by Step 3 are not hidden

It is indeed a bug but not relevant to this PR. The bug exist in the original authorship implementation. I will open another PR for this.

0blivious:
> One question:
> 
> Playing around with the "select all file types" button throws this vue-warn log:
> image
> 
> If we are using `v-model`, is it a requirement to have a setter?

If we are using a computed property for v-model and try to modify it, we need. I have changed the on-click function to use vue setter and getter to handle this.

frontend/src/static/js/v_authorship.js
```diff
@@ -107,11 +117,11 @@ window.vAuthorship = {
         file.className = renameValue;
       });
 
-      this.activeFilesCount = isActive ? this.selectedFiles.length : 0;
+      this.isActive = isActive;
     },
 
     updateCount() {
-      this.activeFilesCount = document.getElementsByClassName('file active').length;
+      this.isActive = document.getElementsByClassName('file active').length;
```
ccyccyccy: Shouldn't this line be `this.isActive = document.getElementsByClassName('file active').length > 0;`?

0blivious: don't need. .length is enough alr

ccyccyccy: I know it works, but it would be confusing for future developers if `this.isActive` is sometimes a `Boolean` and other times a `Number`, and the name implies that it is a boolean value also.

0blivious: done

jamessspanggg:
@0blivious can retrigger CI?

PR #1048, opened 2020-02-03
Add commit message length configuration border

jamessspanggg:
```
GitHub's maximum length of commit message's subject line
and body is defaulted at 50 and 72 characters respectively.

Commits panel does not show the indication of the maximum
lengths.

Let's add a dotted lines for commit message bodies at 72
characters and subject lines at 50 characters, these
lines indicate whether a commit message's subject
line or body exceed GitHub's maximum length.
```


damithc:
> Let's add a dashed border for commit message bodies at 72
> characters as well as scrollable boxes for subject line that
> is fixed at 50 characters.

For this, I meant just draw a box behind the text so that we can see if the subject is longer than 50 chars. A scrollable box will require users to scroll, which is unnecessary work.

jamessspanggg:
Ok noted prof.

jamessspanggg:
@damithc since the commit message subject line is limited to 50 characters, should we change the font to `monospace` (currently it is `"Titillium Web",sans-serif;`) ? So that each font takes equal amount of space and it is easier to get the length of 50 characters.

Changing the font however, might introduce inconsistencies in font between commits panel and authorship panel.

damithc:
> @damithc since the commit message subject line is limited to 50 characters, should we change the font to `monospace` (currently it is `"Titillium Web",sans-serif;`) ? So that each font takes equal amount of space and it is easier to get the length of 50 characters.
> 
> Changing the font however, might introduce inconsistencies in font between commits panel and authorship panel.

I think we should use the same monospace font for code, github commit message (subject and body) as all are code-like. Anyway, try it and post some screenshots to see how it looks.

jamessspanggg:
> I think we should use the same monospace font for code, github commit message (subject and body) as all are code-like. Anyway, try it and post some screenshots to see how it looks.

I have tried two ways of showing the 50 character limitation:

1. With the box that wraps the text at 50 characters.
Cons: looks slightly odd.
Screen Shot 2020-02-03 at 9 04 27 PM

2. Using the ellipsis icon from the 50th character onwards:
Cons: require user to hover over the link to see full commit subject line
Screen Shot 2020-02-03 at 9 14 37 PM

damithc:
We don't want to hide anything; just want to shame the ones that violates the limit. What if you show a vertical white dotted line after 50 chars. Subtle but makes the point.

jamessspanggg:
@damithc switched both limits lines to be dotted. Is the dotted lines in the commit message title too subtle?

Screen Shot 2020-02-04 at 12 11 36 AM

damithc:
> @damithc switched both limit lines to be dotted. Are the dotted lines in the commit message title too subtle?

Yes. Use black, which is also consistent with what we do in the message body?

jamessspanggg:
> Yes. Use black, which is also consistent with what we do in the message body?

Screen Shot 2020-02-04 at 10 30 52 AM

frontend/src/static/css/style.scss
```diff
@@ -321,7 +333,7 @@ header {
   }
 
   &--button {
-    color: mui-color('grey', '700');
+    color: mui-color('grey');
```
jamessspanggg: Modified to standardise with the icon colors in `v_authorship`

damithc:
> See the message below, the dotted line appear in the middle of the PR number. For the commit title length, I feel we should not count the PR number as it is not input by the user.

PR number is part of the subject line, added by the user. It's just a project convention, not automatic.

jamessspanggg:
Yep placing the PR number is a choice, not automatic.

frontend/src/static/css/style.scss
```diff
@@ -313,6 +313,18 @@ header {
       padding: .4rem;
       resize: none;
       width: 100%;
+
+      pre {
+        position: relative;
+        .dotted-border {
```
0blivious: shouldnt we have a blank line here?

frontend/src/static/css/style.scss
```diff
@@ -332,4 +344,21 @@ header {
...
+    font-family: monospace;
+    .within-border {
+      display: inline;
+    }
+    .not-within-border {
+      border-left: 1px dotted mui-color('black'); // 50th character line
+      display: inline;
+      pointer-events: none;
+    }
+  }
```
0blivious: Need blank line btw class style

frontend/src/static/css/style.scss
0blivious: Need a blank line here

frontend/src/static/css/style.scss
```diff
@@ -332,4 +344,21 @@ header {
...
+    display: inline;
+    font-family: monospace;
+    .within-border {
+      display: inline;
+    }
+    .not-within-border {
+      border-left: 1px dotted mui-color('black'); // 50th character line
+      display: inline;
+      pointer-events: none;
+    }
```
0blivious: should have blank line btw class styles

frontend/src/static/css/style.scss
```diff
@@ -332,4 +344,21 @@ header {
...
+  .message-title {
+    display: inline;
+    font-family: monospace;
+    .within-border {
+      display: inline;
+    }
+    .not-within-border {
+      border-left: 1px dotted mui-color('black'); // 50th character line
+      display: inline;
+      pointer-events: none;
```
0blivious: Why do we need the pointer-events none here? It doesnt seem to produce any difference on my computer.

jamessspanggg: because users might want to select and copy&paste the chunks of code, without pointer-events none, we are unable to select the code.

Tejas2805: Just one nit. I am okay with the style name `not-within-border`. But I think `outside-border` might be better. Can go ahead with whatever other reviewers agree too with.

jamessspanggg: > I understand that we need this for the pre part to select the code body, but this one is for the title? I can still select the part of title I want without this property.

Oops my bad, I thought u were referring to the message body. Yes in this case, the `pointer-events:none` is unnecessary. Thanks for pointing it out!

jamessspanggg:
@jylee-git have updated the UG. Could you help me check if the description is ok?

docs/UserGuide.md
```diff
@@ -107,6 +107,7 @@ The `Commits Panel` allows users to see the commits attributed to a specific aut
 * The date range for the `Chart Panel` can be updated by clicking on the "Show ramp chart for this period" below the name of the author.
 * The ramp slices displayed in the ramp chart for the `Commits Panel` represents individual commits.
 * The commit messages body can be expanded or collapsed by clicking on the `...` icon beside each commit message title.
+* To promote Git's 50/72 commit message rule, the commit message title shows a dotted border at 50 characters while the commit message body shows a dotted border at 72 characters.
```
jylee-git: I think something like this may be better

> To promote and encourage the 50/72 rule for commit messages, a dotted border will be shown for commit message subject that exceeds 50 characters and shown for commit message body after the 72nd character mark.

or

> * To promote and encourage the 50/72 rule for commit messages, a dotted border will be shown for:
>   * commit message subject that exceeds 50 characters 
>   * commit message body after the 72nd character mark.

jamessspanggg: thanks for the advice, have went with the second option

Tejas2805: Better to make them consistent?

jamessspanggg:
> * The line at the subject title is rather unsightly; it looks more like a visual bug than a feature, would underline be better for this job?

> * Dotted line looks almost like a solid line, perhaps a grey dashed line would be better to not obscure the view of users?

@damithc @eugenepeh what do y'all think of the UI?

Screen Shot 2020-02-23 at 9 25 56 AM

eugenepeh:
>@damithc @eugenepeh what do y'all think of the UI?

Perhaps the underline can be solid and the dashed line for the 72 characters can use lighter shade of grey? Do wait for @damithc 's input too.

jamessspanggg:
> Perhaps the underline can be solid and the dashed line for the 72 characters can use lighter shade of grey? Do wait for @damithc 's input too.

Ok noted. Updated the previous comment with the UI.

damithc:
> Perhaps the underline can be solid and the dashed line for the 72 characters can use lighter shade of grey? Do wait for @damithc 's input too.

hmm... After seeing both, I kind of prefer vertical lines; while it may not be nice to look at, a vertical line is a good visual cue to indicate 'it should end here'. Also for consistency, we should make make the two as similar as possible, given both have the same purpose. If both look the same, that should also alleviate the worry that users will mistake it as a glitch. What do you think?

damithc:
BTW, I believe there is an another ongoing PR that is trying to tweak the page background color, which has a bearing on the colors we choose for this feature.

jamessspanggg:
> BTW, I believe there is an another ongoing PR that is trying to tweak the page background color, which has a bearing on the colors we choose for this feature.

I believe it's from this PR [#1080](https://github.com/reposense/RepoSense/pull/1080), which modifies the code and commits panel to be white.

 > hmm... After seeing both, I kind of prefer vertical lines; while it may not be nice to look at, a vertical line is a good visual cue to indicate 'it should end here'. Also for consistency, we should make make the two as similar as possible, given both have the same purpose.

I agree that we should use vertical lines for both. Would grey vertical dashed lines be alright? Grey would look ok on white backgrounds, as shown in the commit message body.

Screen Shot 2020-02-23 at 4 31 55 PM

eugenepeh:
>hmm... After seeing both, I kind of prefer vertical lines; while it may not be nice to look at, a vertical line is a good visual cue to indicate 'it should end here'. Also for consistency, we should make make the two as similar as possible, given both have the same purpose. If both look the same, that should also alleviate the worry that users will mistake it as a glitch. What do you think?

My concern on vertical line for the title is more on the inconsistency of presence (when the title is shorter than 50 chars, it won't appear), which is why it kind of appear to looks like a glitch to me. 

Though I agree that it may be easier to identify the 50 chars cut off point, but I personally feel that underline would probably look more professional because it looks similar to how people are writing beyond the allocated space on forms and etc.

damithc:
> Though I agree that it may be easier to identify the 50 chars cut off point, but I personally feel that underline would probably look more professional because it looks similar to how people are writing beyond the allocated space on forms and etc.

I see what you mean. The form field analogy didn't occur to me because that behavior never happens in an online form, and in Web pages, underlines usually mean other things (such as links or tooltips). Let's stick with vertical lines for now. It will look a bit odd, but probably we can make it subtle enough to not to irritate the viewer. Or it may be OK for it to be irritating to the viewer in this instance because we are exposing an incorrect behavior of the commit author. The line will not even appear if the authors stuck to the standard. @jamessspanggg let's make both lines same thickness and same style, and if possible, same shade of grey.

jylee-git:
What about this alternative instead of the dotted line:

Any commit subject that exceeds 50 char will have a ⚠️ sign or similar that states commit message exceeds 50 char?

damithc:
> Any commit subject that exceeds 50 char will have a ⚠️ sign or similar that states commit message exceeds 50 char upon being hovered?

I feel that's beyond RepoSense's role. We should only provide a basis for the user to make their own judgement, without providing our own judgement.

Let's go with the current approach, dogfood for a while, and refine later if it doesn't work.

Tejas2805:
@fzdy1914 Can merge this first? Another PR by @jamessspanggg on hold due to this.

fzdy1914:
@jamessspanggg CI is failing, can you kindly recheck?

PR #1050, opened 2020-02-03
[#927] Hotfix: remove duplicate commits in contributions

### LINKED ISSUE #927: Ramp chart: Duplicate ramps
This only occurs when no `author-config.csv` is provided and one author made commits using two different GitHub Author names. For example, **Alice Tan** and **alice99**. Authors are retrieved manually and they are treated as 2 separate authors.

If a commit was made using the following author configuration:
Author Name: Alice Tan
Author Email: alice99@users.noreply.github.com

This commit will be retrieved 2 times when retrieving commits made by **Alice Tan** and **alice99**. When assigning the commit ownership, **Alice Tan** will get for credit for both the commits, leading to duplicate ramps.

To see this bug, run the following command:
```
./gradlew run -Dargs="--repos http://github.com/AY1920S1-CS2103T-T12-2/main.git --view --since 25/09/2019 --until 25/09/2019"
```

One of the authors (Ryan Tay) will have duplicate ramps:

Screenshot 2019-10-23 at 1 09 02 AM

ang-zeyu:
Hotfix #927 
Hotfix #798 
Hotfix #1206

```
Remove duplicate commits in daily contributions

Two or more authors under with the same extracted commit may result in
the commit being added more than once to an author’s daily contribution.

This occurs when these authors use some identical aliases, which causes
the same author to be retrieved from the commit result for adding to
the daily contribution.
This can be reproduced by configuring said authors, or using no author
config under the above conditions.

Let's add a duplicate commit check to the commit analyzing process to
prevent this.
```


ang-zeyu:
Result from the added test case without the fix:

Screenshot from 2020-02-04 22-32-23

anubh-v:
@ang-zeyu, thanks for contributing!

I'm wondering whether we should check for duplicate commits at an earlier stage, in `CommitInfoAnalyzer` rather than `CommitResultAggregator`.

The check for duplicates might be placed in the `analyzeCommits` method:
https://github.com/reposense/RepoSense/blob/a9fedc6689c5bea63706bde2c01101b42b058cb2/src/main/java/reposense/commits/CommitInfoAnalyzer.java#L53-L62

What do you think?

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -111,6 +112,15 @@ private static float getContributionVariance(List<AuthorDailyContribution> contr
                 addDailyContributionForNewDate(authorDailyContributions, commitStartDate);
             }
 
+
+            List<CommitResult> currentDayCommitResults =
+                    authorDailyContributions.get(authorDailyContributions.size() - 1).getCommitResults();
+            Predicate<CommitResult> commitResultDuplicatePredicate = currentDayCommitResult ->
+                    currentDayCommitResult.getHash().equals(commitResult.getHash());
```
anubh-v: Could we use the `equals` method of `CommitResult` for this duplicate check?

ang-zeyu:
Hi @anubh-v, thanks for looking through it!

> The check for duplicates might be placed in the `analyzeCommits` method:
> https://github.com/reposense/RepoSense/blob/a9fedc6689c5bea63706bde2c01101b42b058cb2/src/main/java/reposense/commits/CommitInfoAnalyzer.java#L53-L62
> 
> What do you think?

Should be good :+1: 
I could try placing it at extraction instead as well, if that's preferred. 
Though its wouldn't be very in line with `extraction` then.

src/main/java/reposense/commits/CommitResultAggregator.java
ang-zeyu: Hmm, issue is that two commits can have the same hash but different commit titles ( possibly other fields as well ).
The example command given in #927 demonstrates this
`./gradlew run -Dargs="--repos http://github.com/AY1920S1-CS2103T-T12-2/main.git --view --since 25/09/2019 --until 25/09/2019"`

anubh-v:
> After trying it out, it might not be a good idea to put it at the analyze stage, as distinct() dosen't work due to sorted() being present.

Thanks for the response @ang-zeyu. I'm not really clear what you meant here - could you clarify?
What's the issue between `distinct` and `sort`?

I looked into the `distinct` operation and found that we can remove duplicate commits by applying the `distinct` to the `CommitResult` stream in `analyzeCommits` - **given** that we also override the `hashCode` method of `CommitResult`.

Here's what I tried:

Overriding `hashCode`:
```
@Override
public int hashCode() {
    return hash.hashCode();
}
```

Applying `distinct` in `CommitInfoAnalyzer#analyzeCommits`

```
return commitInfos.stream()
                .map(commitInfo -> analyzeCommit(commitInfo, config))
                .filter(commitResult -> !commitResult.getAuthor().equals(Author.UNKNOWN_AUTHOR)
                        && !CommitHash.isInsideCommitList(commitResult.getHash(), config.getIgnoreCommitList()))
                .sorted(Comparator.comparing(CommitResult::getTime))
                .distinct()
```

Do you see any issues with this approach?
There's a [discussion](https://stackoverflow.com/questions/21333646/stream-and-the-distinct-operation) here that is relevant.

@fzdy1914 may we get your input on this PR? We are discussing different methods of filtering out duplicate commits.

ang-zeyu:
I mistakenly assumed `distinct` did not use the `hashCode`, and that the issue was with how java processes the pipeline for `distinct` + `sorted`, thanks!

A rather minor thing:

I've implemented the hashcode, but put `distinct()` before `sorted()` in the pipeline.

The issue is that we are ordering the stream first by the commit result's time, but then calling `distinct()` after it. The equality check for distinct checks all fields of `CommitResult`, which may not be in line with the ordering given after `sorted()`

https://stackoverflow.com/questions/46177093/java-8-stream-distinct-is-not-working
( the post where I got the mistaken idea `distinct` + `sorted` causes issues all the time )

The reason this seems to pass is because for the test case I added, it so happens the time stamp for the duplicate commits are unique ( in the sense of all commits for the test repo ). Hence, the two commits end up after one another, and `distinct()` correctly removes the duplicate commit. ( Of course, it would be very unlikely two commits with different hashes have the same timestamp )

anubh-v:
@ang-zeyu, so the question is whether the order of `distinct` and `sorted` matters in the stream operations?

From the [documentation](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#distinct--), there doesn't seem to be any indication that the order of operations matters.

I looked at the post you linked, but I couldn't replicate the "buggy" behaviour with using `distinct` after `sorted`.  I've attached the program I used to test this (had to use `txt` file extension instead of `.java` as GitHub doesn't allow uploading `.java` files)
[sort.txt](https://github.com/reposense/RepoSense/files/4196429/sort.txt)


Anyway, I think the code looks ok now.
With regards to testing, I am not sure if we should test this behavior in the system test or in the integration tests. @fzdy1914 what do you think?

ang-zeyu:
> I looked at the post you linked, but I couldn't replicate the "buggy" behaviour with using `distinct` after `sorted`. I've attached the program I used to test this (had to use `txt` file extension instead of `.java` as GitHub doesn't allow uploading `.java` files)
> [sort.txt](https://github.com/reposense/RepoSense/files/4196429/sort.txt)

Thanks for the experiment @anubh-v!
You can produce it by replacing the use of `comparator` with implementing `comparable`. ( didn't realise this as well until your experiment )
The likely cause is that using a `comparator` "instructs" java explicitly that we are using a separate form of equality for sorting the stream, which avoids `distinct()` making the assumption that equal elements follow each other.
Strange quirks of java 😐 

> Anyway, I think the code looks OK
> With regards to testing, I am not sure if we should test this behavior in the system test or in the integration tests. @fzdy1914 what do you think?

I'll add in an integration test as well later, which I think would be in line with the rest of the `analyze` integration tests. ( I used a system test before as there weren't any integration tests for `aggregation` )

Let me know if either should be removed!

ang-zeyu:
Updated with integration tests as well

src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
```diff
@@ -65,6 +66,17 @@ public void analyzeCommits_eugeneAuthorNoIgnoredCommitsNoDateRange_success() {
         Assert.assertEquals(NUMBER_EUGENE_COMMIT, commitResults.size());
     }
 
+    @Test
+    public void analyzeCommits_duplicateAuthorsDuplicateCommits_success() {
+        config.setAuthorList(Arrays.asList(getAlphaAllAliasAuthor(), getAlphaAllAliasAuthor()));
```
anubh-v: In this test, we are creating a list that has duplicate authors, so that we can test the `CommitInfoAnalyzer`'s ability to remove duplicate commits.

I would suggest that instead of creating an author list with duplicate authors, we configure the `SINCE_DATE` and `END_DATE` of the `RepoConfiguration` instance to include a time period where such duplicate commits exist.

For example, try running RepoSense with 
```
./gradlew run -Dargs="--repo https://github.com/reposense/testrepo-Alpha.git --since 6/5/2018 --until 9/11/2018 --view --ignore-standalone-config"
```
`testrepo-Alpha` is our test repo in `CommitInfoAnalyzerTest` and this time period has some duplicate commits:


image


We can use this configuration for this test and check if the duplicate commit is removed by `CommitInfoAnalyzer` ?

ang-zeyu: Updated

jamessspanggg:
@jylee-git may I get your opinion on this solution? since you have tried fixing this problem last time

jylee-git:
Sry for the late response, took me a while to consolidate this.

@ang-zeyu   I'm thinking of a different solution, but you might want to test it out to see if it works or not. 

Consider this:
**`repo-config.csv`**
| repo location | branch | ... |
|------|--------|-------|
| https://github.com/reposense/RepoSense.git | master |  |

**`author-config.csv`**
| repo location | branch | author's gitID | ... |
|------|--------|-------|----|
| https://github.com/reposense/RepoSense.git | master | eugenepeh  | ... |
| https://github.com/reposense/RepoSense.git | master | Eugene Peh  | ... |

(leave the other fields blank).
Run with the args `-c ./ -s d1 -i -v`

In `gitLog#get`, notice the line `GitUtil.convertToFilterAuthorArgs(author)` where the method `convertToFilterAuthorArgs` returns the email addresses appended with the author's git id as well, if the config file doesn't provide the email address that the author committed using.

`Author` instances will be generated, with the email address of:
* the author `eugenepeh` being `eugenepeh@users.noreply.github.com`, and 
* the author `Eugene Peh` being `Eugene Peh@users.noreply.github.com`.

 `gitLog` will be run twice,
* once with `... --author="^.* <\(.*+\)\?eugenepeh.users.noreply.github.com>$\|^eugenepeh <.*>$"`
* once with `... --author="^.* <\(.*+\)\?Eugene Peh.users.noreply.github.com>$\|^Eugene Peh <.*>$"

Consider one of the many commits:
```
>>>COMMIT INFO<<<
0637f4b826428e6565deeca59131f1c9a3637037|
|Eugene Peh|
|eugenepeh@users.noreply.github.com|
|2019-08-18T21:25:45+08:00|
|AboutUs: update team members (#867)|
|    As more members join and leave us, it is about time for us to distinct
    the active members so that users can easily seek help from us.

    Let's move our team list to a new page to separate our active and past
    members into different section. Also, let's include and update the
    roles to accredit our senior members who have been helping out with
    many of the work, including mentoring the newcomers.|
|tag: v1.10rc2|
 2 files changed, 94 insertions(+), 12 deletions(-)
```

This particular commit appeared for both git log commands that were run.

Then in the later part of the code, this particular commit result will be attributed to the author `Eugene Peh` twice because the email `eugenepeh@users.noreply.github.com` and the author ID `Eugene Peh` satisfies the individual `gitlog` commands that were run.

So one solution that you can consider is to consider removing the use of email to run `GitLog` when the email isn't provided (i.e.: when running gitlog, only search for commits by the authorID instead of both authorID and email).

Let me know if that works or not, or which solution might be a better fit for this problem.

ang-zeyu:
> Sry for the late response, took me a while to consolidate this.

No worries; Thanks for the in depth explanation! 😄 

I noticed this as well while fixing this. The issue is tied deeply to the current architecture.
I don't think we should fix this during extraction ( explained later ), but I realise my fix is also not ideal.

First of all, if we have two authors, with some identical aliases ( by configuration or the result of no configuration ),
I think we could agree that the ideal behaviour is to attribute _all commits that can be possibly attributed_ to each of these authors, without causing any missing or duplicate commits for each author while doing so.

Currently, the analyzing process looks at the result of a gitLog and extracts the author first by name if possible, falling back to the email, then to `UNKNOWN_AUTHOR`.

This is the root of the issue. 
With reference to your earlier example,
this causes the author of `Eugene Peh` to be incorrectly tagged to the `CommitInfo` object.

Thus the fix I originally had in mind:
- When resolving the author of a result of a `gitLog`, we resolve this **author by author**. That is,
we do the extraction and analyzing process for a author in one go, and assign the author of the resulting `CommitInfo`s to the author irregardless.
- The problem: huge architectural restructure needed as you might imagine, which will likely result in a lot of changes to existing test outputs, etc.

If we remove the use of email when running `gitLog` as you suggested, it would result in some missing commits for the `eugenepeh` author, and would also be masking the issue.

A simple interim fix hence, that is easily reversible, is to add a one line `distinct()` call in the commit analyzing process. This would result in no duplicate commits, though it would still result in missing commits for `eugenepeh` ( as was before ). 

I'm open to doing the architectural restructure instead if preferred in one go, but as mentioned,
this interim fix is _'current-behaviour-agnostic'_ in that only removes duplicate commits from the current state of code of RepoSense, while preserving the 'missing' commits from the current state as well. ( and is easily reversible )

Let me know if the 'real fix' is preferred!

jylee-git:
If that's the case, should we make a code overhaul instead to solve this issue?

My opinion is that this isn't a pressing issue that has to be dealt with asap; so if we can deal with the underlying issue without pushing interim fixes, that would be the best.

@fzdy1914 @eugenepeh need your inputs: from this comment https://github.com/reposense/RepoSense/pull/1050#issuecomment-593073797 , how do you think we should proceed from here?

eugenepeh:
>@fzdy1914 @eugenepeh need your inputs: from this comment #1050 (comment) , how do you think we should proceed from here?

why not use the commit's hash to remove duplicates?

ang-zeyu:
> why not use the commit's hash to remove duplicates?

It's what this PR already does currently ( the 'interim' fix mentioned ), although it uses equality of `CommitResult` instead.

fzdy1914:
Does it take a lot of effort to fix the real logic behind after this PR is merged? Maybe this PR can be merged first and fix the inner cause later.

ang-zeyu:
> Does it take a lot of effort to fix the real logic behind after this PR is merged? Maybe this PR can be merged first and fix the inner cause later.

Quite a bit, this is the current architecture:
```
Extract all commits of all authors
Analyze all such commits
Aggregate the analyzed results
```

To fix the root issue, it would have to be changed to roughly:
```
for all authors
  extract author's commits
  analyze author's commits
  aggregate author's analyzed results ( being non-conservative here though, not sure if this is necessary )
aggregate all results
```

ang-zeyu:
> My opinion is that this isn't a pressing issue that has to be dealt with asap; so if we can deal with the underlying issue without pushing interim fixes, that would be the best.

Another potential impetus for this fix besides removing duplicate commits from the user's view would be vue complaining of duplicate keys in the zoom view, which could cause view update errors.

What do you think? :o @jylee-git @fzdy1914

fzdy1914:
@eugenepeh Need your input here. Should we just remove the dupicated commits first?

eugenepeh:
>Does it take a lot of effort to fix the real logic behind after this PR is merged? Maybe this PR can be merged first and fix the inner cause later.

I think what causes this might be because (correct me if my speculation is wrong) current logic sieve through all the commits for each author/aliases which results in "double-count". 

Can you run a quick check, such that if 2 authors have added the same aliases (e.g. alice99 in https://github.com/reposense/RepoSense/issues/927), would it result in both author having the same contributions?

ang-zeyu:
> I think what causes this might be because (correct me if my speculation is wrong) current logic sieve through all the commits for each author/aliases which results in "double-count".
> 
> Can you run a quick check, such that if 2 authors have added the same aliases (e.g. alice99 in #927), would it result in both author having the same contributions?

Hmm not sure what you meant; There's multiple points where all commits are sieved through ( `extraction`, `analyzing`, `aggregation` ). The cause is as mentioned here https://github.com/reposense/RepoSense/pull/1050#issuecomment-593073797.

Between the analyzing and extraction stage, the analyzed `CommitResult`'s author is not tied correctly ( and not possible )  back to the author that was responsible for extracting the commit (`commitInfo`).

eg suppose we have specified in the configuration or as a result of no configuration
```
commit - user: Ryan Tay email: ryantay232@gmail.com
author1 - user: Ryan Tay email: irrelevant
author2 - user: ryantay232 email: ryantay232@gmail.com
```
After analyzing, there will be two `CommitInfos` containing the commit with the exact same information.
During analyzing however, the process of 'tying' the author back to the author which retrieved the commit follows this ( you can trace it from the `getAuthor` call in the `analyzeCommit` method ):

> Currently, the analyzing process looks at the result of a gitLog and extracts the author first by name if possible, falling back to the email, then to UNKNOWN_AUTHOR.

This causes both commits to be tied to author 1 hence, and its also not possible to absolutely know which author the `CommitInfo` was extracted from.

The 'proper' fix would be an architectural restructure as illustrated here https://github.com/reposense/RepoSense/pull/1050#issuecomment-593416800

The issue would be the same with aliases essentially, if both authors extract the _same commit_ but then end up having both commits assigned to one of them because we don't know which author actually extracted that commit.

eugenepeh:
```
commit - user: Ryan Tay email: ryantay232@gmail.com
author1 - user: Ryan Tay email: irrelevant
author2 - user: ryantay232 email: ryantay232@gmail.com
```

>This causes both commits to be tied to author 1 hence, and its also not possible to absolutely know which author the CommitInfo was extracted from.

I see, so in this example, the author2 would have 0 commit while author1 would have 2?

As you suggested, it seems that the current code has some fundamental errors. Ideally the CommitInfo should only be allocated once or allocated equally to both authors, do check with Prof on his preference. Let's proceed with this PR being a hotfix and implement a patch afterward then.

ang-zeyu:
> I see, so in this example, the author2 would have 0 commit while author1 would have 2?

yup

> do check with Prof on his preference. Let's proceed with this PR being a hotfix and implement a patch afterward then.

Will do 👍
Thanks for reviewing this! @eugenepeh @fzdy1914 @jylee-git @jamessspanggg @anubh-v 

I've updated it on the latest again, no changes / conflicting code

ang-zeyu:
Resolved minor test file conflict with #1067

src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
```diff
@@ -125,6 +125,26 @@ public void analyzeCommits_emailWithAdditionOperator_success() {
...
+    @Test
+    public void analyzeCommits_duplicateAuthorsDuplicateCommits_success() throws ParseException {
+        Author author = new Author(EUGENE_AUTHOR_NAME);
+        List<CommitResult> expectedCommitResults = new ArrayList<>();
+        expectedCommitResults.add(new CommitResult(author, LATEST_COMMIT_HASH,
+                parseGitStrictIsoDate("2018-11-08T13:50:40+08:00"),
+                "Add config.json with invalid fields (#2)",
+                "", null, 17, 0));
+
+        config.setAuthorList(Arrays.asList(author, author));
```
anubh-v: The test does achieve the goal of verifying that duplicate commits are removed.
However, when we create the author list, would it be better if one of the author instances has the author name `eugenepeh` and the other instance has the author name `Eugene Peh` ?

I understand that the bug can be replicated with the current test setup too, but perhaps using `eugenepeh` and `Eugene Peh` allows us to better simulate the bug that appears in practice?

ang-zeyu: Hmm, there's the system test for that already though, I feel that it may be better to keep the duplicate authors here so the purpose of the test `duplicate commits are removed` is clearer.
Otherwise we are also testing the commit tagging behaviour here ( which should be fixed )

anubh-v: Ok - yes the system test does cover it already

ang-zeyu:
@fzdy1914 could this be merged? There were no changes between @eugenepeh's approval and now 😁 Will raise the architectural restructure up as an issue as well after!

eugenepeh:
@ang-zeyu can you change one of ur commit messages slightly to trigger a rebuild?
Seems like the build is stuck for this PR

ang-zeyu:
> @ang-zeyu can you change one of ur commit messages slightly to trigger a rebuild?
> Seems like the build is stuck for this PR

Still stuck =\
Will continue trying to rebuild periodically

PR #1053, opened 2020-02-04
[#674] List binary files touched

### LINKED ISSUE #674: List binary files touched
Suggested: In the code view, use another tab to list binary files touched by the author.

ccyccyccy:
Fixes #674 
```
Binary files changes are currently not reflected in the authorship
panel

Let's add a line in the authorship panel for each of the binary files
modified by the author, as well as a checkbox in the authorship panel
to filter binary files.

A new fields are added to FileResult: boolean isBinary. The isBinary
field is to help identify the files, and the binaryFileAuthors is
needed since the file could have been modified by multiple authors. The
analysis of binary files follows the same process of Extractor ->
Analyzer -> Aggregator, with some steps skipped as they are not
relevant to binary files.
```
Since the netlify preview doesn't include binary files, I hosted a preview [here](https://ccyyccyy.surge.sh/#search=&sort=groupTitle&sortWithin=title&timeframe=commit&mergegroup=false&groupSelect=groupByRepos&breakdown=false&tabOpen=true&tabType=authorship&tabAuthor=Tejas%20Bhuwania&tabRepo=reposense%2FRepoSense%5Bmaster%5D)


Tejas2805:
>Should the modifying the binary file contribute to the total line diff? Currently it contributes to the total line count

I think this is fine.

jylee-git:
> 1. I added an additional `isBinaryFile` parameter to various functions to identify if the file is binary or not. Should I instead create new method instead of reusing the exiting ones? But I feel that creating new methods would result in quite a lot of code duplication.

Best to avoid code duplication. You can think of refactoring the existing code base if that helps.

> 2. The result currently looks like this:
>    image
>    Is this fine or should I make it distinctively different from the usual non-binary files?

I don't think we should blend the binary file in in this way. People may think you are writing `Binary diff not shown` as a one-line text file, which can happen. You can just name a text file as any format (like `.png`)

 My opinion on this is to either:
1. Have a separate section (possibly below all the non-binary files?) that lists the binary files that are touched by this author
2. Something like point 1, but a collapsible box (like the error summary red box on the summary panel), that states something like `list of binary files touched`, 
image
3. (See image - assume that first file is binary)
image

@damithc your opinions?


> 3. Should the modifying the binary file contribute to the total line diff? Currently it contributes to the total line count:

My take: let's see how to proceed with this after point 2 is addressed.

> 4. SystemTest is failing due to the additional field in `FileResult`. How do I generate the new expected outputs?

You can overwrite the expected files by running RepoSense with the various configs set (see the 3 methods in SystemTest class to understand the various parameters used to run it).

damithc:
> 1. Should the modifying the binary file contribute to the total line diff? Currently it contributes to the total line count:

I think it should not.

jylee-git:
@damithc sorry I meant to tag you on the part of how to show the users the list of binary files touched.

>> 2. The result currently looks like this:
>    image
>    Is this fine or should I make it distinctively different from the usual non-binary files?

damithc:
> @damithc sorry I meant to tag you on the part of how to show the users the list of binary files touched.

I don't mind listing binary files together with other files as long as they are clearly distinguishable as binary files (e.g., use a different color); if not, they could be mistaken for one-line contributions (as @jylee-git mentioned). No objections to listing them in a different section either. If we do the latter, perhaps show them at the bottom of the page rather than at the top (as binary files are not that important)?

Tejas2805:
Let's try to follow Github's way of showing Binary files. Since our goal is to try and modify our Ui based on how Github displays.

Screen Shot 2020-02-05 at 1 23 26 PM

1. It shows binary files at the bottom
2. Doesn't add it to line count
3. Displays `Binary file not shown` in white
4. We don't show the size of the file 

@damithc What do you think?

fzdy1914:
We can display BIN at normal line number place

ccyccyccy:
>     1. It shows binary files at the bottom

I feel that we should display the file sorted along with the rest of the files if we follow the Github way of showing binary files.

ccyccyccy:
I updated the display to this.

image

Does this look fine?

fzdy1914:
> I updated the display to this.
> 
> image
> 
> Does this look fine?

The `1 line` still shows in the heading.

fzdy1914:
The margin looks weird, else looks good to me.

ccyccyccy:
> > 1. Should the modifying the binary file contribute to the total line diff? Currently it contributes to the total line count:
> 
> I think it should not.


I think that modifying the binary file should count as a single line change to show how many binary files have been touched, as well as allowing users to filter them away

image

Or we can show how many binary files are touched without adding them to the line count? Like this

image

Your thoughts? @damithc

damithc:
> I think that modifying the binary file should count as a single line change to show how many binary files have been touched, as well as allowing users to filter them away

It's good to show the file count and treat it as a separate group of files. Not sure it is appropriate to treat them as single line changes as there is no good justification for it. Either we have to not count them for LoC or treat them as 0 LoC.

ccyccyccy:
> It's good to show the file count and treat it as a separate group of files. Not sure it is appropriate to treat them as single line changes as there is no good justification for it. Either we have to not count them for LoC or treat them as 0 LoC.

How about this? The file count will not contribute to the **All** line count, and we just show the number of binary files touched.

image

damithc:
> How about this? The file count will not contribute to the **All** line count, and we just show the number of binary files touched.
> 
> image

Sounds good to me.

fzdy1914:
I think it is still possible to add the file number count to the `all`. Otherwise, it is weird when unclicking `all`.

Tejas2805:
> How about this? The file count will not contribute to the **All** line count, and we just show the number of binary files touched.
> 
> image

We need to update the UG accordingly.

ccyccyccy:
> I think it is still possible to add the file number count to the `all`. Otherwise, it is weird when unclicking `all`.

Maybe we can change its colour and show a tooltip when hovered over it, and update in the UG as well.

image

ccyccyccy:
## Summary
### Header

image

### FIle

image

1. It shows binary files along with the rest of the files
2. Doesn't add it to line count
3. We don't show the size of the file



@damithc What do you think of this UI?

ccyccyccy:
> ## Summary
> ### Header
> 
> image
> ### FIle
> 
> image
> 
>     1. It shows binary files along with the rest of the files
> 
>     2. Doesn't add it to line count
> 
>     3. We don't show the size of the file
> 
> 
> @damithc What do you think of this UI?

@damithc Really need your opinion on the UI here.

damithc:
> @damithc Really need your opinion on the UI here.

Looks good to me. Sorry, I missed this earlier. Thanks for the reminder.

frontend/src/tabs/authorship.pug
```diff
@@ -68,8 +74,11 @@
             title="click to view the blame view of file"
           )
             i.button.fas.fa-user-edit
-        pre.hljs
+        pre.hljs(v-if="!file.isBinary")
           template(v-for="segment in file.segments")
             v_segment(v-bind:segment="segment", v-bind:path="file.path")
+        .binary-segment(v-if="file.isBinary")
```
jamessspanggg: Can use `v-else` here

src/systemtest/resources/30daysFromUntilDate/expected/reposense_testrepo-Beta_master/authorship.json
```diff
@@ -456,11 +457,13 @@
...
       "-": 52,
       "nbr": 11
-    }
+    },
+    "binaryFileAuthors": []
   },
   {
     "path": "src/main/resources/view/PersonInfoOverview.fxml",
     "fileType": "fxml",
+    "isBinary": false,
```
jamessspanggg: Instead of creating a new field `isBinary`, is it possible to just set the `fileType` to be binary? Then can just check if `fileType == 'Binary'` to see if the file is binary

In this way the json files won't have to add another extra field to each file object

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -77,7 +99,29 @@ private static FileResult generateFileResult(FileInfo fileInfo) {
...
+            fileInfo.getPath(), fileInfo.getFileType(), fileInfo.getLines(), authorContributionMap);
+    }
+
+    /**
+     * Generates and returns a {@code FileResult} with the authorship results from binary {@code fileInfo} consolidated.
+     * Returns null if none of the {@code Author} specified in {@code config} contributed to the file in
+     * {@code fileInfo}.
+     */
+    private static FileResult generateBinaryFileResult(RepoConfiguration config, FileInfo fileInfo) {
+        String authorsString = GitLog.getBinaryFileAuthors(config, fileInfo.getPath());
```
jamessspanggg: It seems like this code gets all the authors that have modified the binary file. 

Are we crediting all the users that have modified the binary file within since date and until date? or should we only get the author that last modified the binary file?

src/main/java/reposense/git/GitLog.java
```diff
@@ -50,4 +50,19 @@ public static String getWithFiles(RepoConfiguration config, Author author) {
...
     }
+
+    /**
+     * Returns the authors who modified the binary file at {@code filePath}, in the repository specified in
+     * {@code config}.
+     */
+    public static String getBinaryFileAuthors(RepoConfiguration config, String filePath) {
+        Path rootPath = Paths.get(config.getRepoRoot());
+
+        String command = "git log --pretty=format:\"%an\t%ae\" ";
```
jamessspanggg: I am not sure whether this is necessary - rerunning `git log` to get the list of authors that modified the file, won't that info be within the first `git log` already?

src/systemtest/resources/30daysFromUntilDate/expected/reposense_testrepo-Beta_master/authorship.json
ccyccyccy: I believe that we should separate it.
Say we have something like this.

image

What should happen if I uncheck `txt`? Should `picture.txt` disappear as well?
Similarly, what should happen if I uncheck `Binary`?

But no matter how we want it to behave currently, separating `FileType` and `isBinary` would be helpful if we were to change its behaviour in the future.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
ccyccyccy: Yes it is getting all the authors that have modified the binary file.

I think that we should get all the authors so that it is kind of similar to what we have right now. Currently multiple authors who modified the same non-binary file are all credited.

src/main/java/reposense/git/GitLog.java
ccyccyccy: Hmm not sure which first `git log` you mean. The only place where `git log` is used is within `generateCommitSummary`, where it returns a `CommitContributionSummary`. But `CommitContributionSummary` doesn't contain any information on authors of specific files, unless I pack the additional information required into `CommitContributionSummary`?

frontend/src/tabs/authorship.pug
ccyccyccy: Fixed.

src/main/java/reposense/git/GitLog.java
jamessspanggg: The `git log` used in `CommitInfoExtractor#extractCommitInfos`

ccyccyccy: Yeah so that method is only called in `CommitsReporter#generateCommitSummary` and returns `CommitContributionSummary` which do not contain the information about the authors of each file.

ccyccyccy:
@damithc Need your opinion on the following behaviour.

Say we have something like this.

image

What should happen if I uncheck `txt`? Should `picture.txt` disappear as well?
Similarly, what should happen if I uncheck `Binary`?

damithc:
> What should happen if I uncheck `txt`? Should `picture.txt` disappear as well?
> Similarly, what should happen if I uncheck `Binary`?

When we say `txt`, we implicitly assume non-binary which means `picture.txt` should not be counted under `.txt` for any purpose. What do you guys think?

0blivious:
> When we say `txt`, we implicitly assume non-binary which means `picture.txt` should not be counted under `.txt` for any purpose. What do you guys think?

I agree with the `txt` logic. However, what if we provide customised file type like `doc` for all `png` and `txt` files? Should the png file be displayed when unticking `pic`? Personally I think we should not display the png file if we untick either `binary` or `doc`.

frontend/src/static/js/v_authorship.js
```diff
@@ -170,21 +171,34 @@ window.vAuthorship = {
...
-        if (lineCnt) {
-          totalLineCount += lineCnt;
-          const out = {};
-          out.path = file.path;
-          out.lineCount = lineCnt;
-          out.fileType = file.fileType;
 
+        // skip if binary file not touched by author
+        if (file.isBinary
+          && !file.binaryFileAuthors.map((author) => author.gitId).includes(this.info.author)) {
```
0blivious: Do we really need this `binaryFIleAuthors`? Can we just use the `authorContributionMap` to find all the authors made change to this binary file?

ccyccyccy: Thanks for the suggestion! I never thought of that. Will change to use `authorContributionMap` instead.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -91,7 +105,7 @@
         }
 
         String[] fileDiffResultList = fullDiffResult.split(DIFF_FILE_CHUNK_SEPARATOR);
-        Set<Path> nonBinaryFilesSet = getNonBinaryFilesList(config);
+        Set<Path> nonBinaryFilesSet = getFilesList(config, false);
```
jamessspanggg: Perhaps can have a `getNonBinaryFilesSet` and `getBinaryFilesSet` that calls `getFilesList`? it will be more clearer

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -142,7 +142,7 @@ public void generateFileInfo_fileWithoutSpecialCharacters_correctFileInfoGenerat
     }
 
     @Test
-    public void getNonBinaryFilesList_directoryWithBinaryFiles_success() {
+    public void getFilesList_directoryWithBinaryFiles_success() {
```
jamessspanggg: need to add another test for non-binary files

frontend/src/static/css/v_authorship.scss
```diff
@@ -131,6 +135,20 @@
       }
     }
 
+    .binary-segment {
+      background-color: mui-color('white');
+      border-left: .25rem solid mui-color('white');
```
jamessspanggg: The `border-left` seems redundant as it has the same colour as the background

src/main/java/reposense/authorship/FileInfoExtractor.java
ccyccyccy: I don't think it is worth it to have such a thin wrapper over the function, considering that we should change `getAllFileInfo` to `getAllNonBinaryFileInfo` and `getAllBinaryFileInfo` as well. 
That is 4 new methods and possibly even more methods for the same concept of indicating if binary or non-binary files is requested, which would add more complexity to the code.

However, I can see that the variable name is somewhat confusing. How about changing `isBinaryFiles` to `binaryFilesWanted` or `binaryFilesRequested`?

src/test/java/reposense/authorship/FileInfoExtractorTest.java
ccyccyccy: Actually this test already contains tests for both binary and non-binary files already. I will change the name of the test.

ccyccyccy:
> Should the png file be displayed when unticking `pic`?

Do you mean unticking `Binary`?

> However, what if we provide customised file type like `doc` for all `png` and `txt` files? Should the png file be displayed when unticking `pic`? Personally I think we should not display the png file if we untick either `binary` or `doc`.

For consistency, I feel that customised file type should follow the same behaviour as the normal ones, i.e. binary files are considered as `Binary` only, without regard for the original file type.

0blivious:
I feel I should be able to collapse the white box below when I click hide all file details and when clicking the title.
image

frontend/src/static/js/v_authorship.js
```diff
@@ -42,6 +42,7 @@ window.vAuthorship = {
       toReverseSortFiles: true,
       hasActiveFile: true,
       filterSearch: '*',
+      isBinaryFilesChecked: true,
```
0blivious: Can I check whats the general line number for a normal binary file (a picture)? If the line number if generally large then better set this to false.

ccyccyccy: Not sure what you mean by line number? Currently I count their linecount as 0 and only show a single line of `Binary file not shown` no matter how large the binary file.

ccyccyccy:
> I feel I should be able to collapse the white box below when I click hide all file details and when clicking the title.
> 
> Also the margin looks different compared to normal files.
> image
> 
> image

Thanks for that! I fixed both issues.

ccyccyccy:
> There is a UI bug:
> 
> Screen Shot 2020-03-09 at 1 40 02 PM
Thanks for noticing! Just fixed it.
> 
> Also, should we show the binary file type in the summary?
> Screen Shot 2020-03-09 at 1 42 26 PM

I don't think so.
The breakdown is by number of lines changed, and I don't see how it can be meaningfully compared with number of binary files changed.

frontend/src/static/js/v_authorship.js
0blivious: I was wondering how the binary file will be sort according to LoC. So for your implementation they will all be placed at the bottom since all their line count is 0 in the json file?

0blivious:
> > However, what if we provide customised file type like `doc` for all `png` and `txt` files? Should the png file be displayed when unticking `doc`? Personally I think we should not display the png file if we untick either `binary` or `doc`.
> 
> For consistency, I feel that customised file type should follow the same behaviour as the normal ones, i.e. binary files are considered as `Binary` only, without regard for the original file type.

Not very sure about this. The user might feel confused if they provide customised group for binary files and the filter seems not functioning. Also if binary files are considered as `Binary` only, why not put it as a file type instead of using attribute `isBinary` if we will stick to this design?

@damithc can we get your opinion on this?

frontend/src/static/js/v_authorship.js
ccyccyccy: Yes sorting by line count will put them all at the bottom.

damithc:
> @damithc can we get your opinion on this?

What are the options?

0blivious:
> What are the options?
Assuming  the user provide customised file type `doc` for all png and txt files.
1. unticking either `doc` or `binary` will hide png files
2. unticking `binary` will hide png files (even when `doc` is still ticked`).
3. unticking both `binary` and `doc` will hide png files.

ccyccyccy:
> Not very sure about this. The user might feel confused if they provide customised group for binary files and the filter seems not functioning.

Definitely agree with this.

> Also if binary files are considered as `Binary` only, why not put it as a file type instead of using attribute `isBinary` if we will stick to this design?

I don't want the frontend to be too coupled with the generated JSON file. Assigning binary files as a `Binary` filetype would lose information about their original filetype. If we decide in the future to change the way we present binary files, it would be much easier to do so if we preserve all the information about the file.

damithc:
> Assuming the user provide customised file type `doc` for all png and txt files.
> 
> 1. unticking either `doc` or `binary` will hide png files
> 2. unticking `binary` will hide png files (even when `doc` is still ticked`).
> 3. unticking both `binary` and `doc` will hide png files.

One policy we wanted to follow for custom groups is, a file should not count for more than one group. If png files are counted for the `doc` group, they should not count for the `binary` group.
I think it is easier to treat binary files as a separate group even if their extension is included in a custom group?
Alternatively, we can allow including binary files into custom groups in which case `binary` group actually means 'other binary files'.

ccyccyccy:
> I think it is easier to treat binary files as a separate group even if their extension is included in a custom group?

I feel that this should be the most intuitive design.
Since custom groups are supposed to override their original filetype, it should override the binary file types as well.

ccyccyccy:
After some thought, I realise that the fundamental issue is that the `Binary` checkbox represents the number of files modified, while the other checkboxes represent number of lines added, and therefore we shouldn't merge them together.

So I feel that we should treat binary files as a separate group even if their extension is included in a custom group

jamessspanggg:
Please resolve the conflicts @ccyccyccy

ccyccyccy:
> Please resolve the conflicts @ccyccyccy

Fixed.

jamessspanggg:
Was thinking if we can add a link to GitHub that displays the binary file at its corresponding commit revision. This way users can view the binary file as well. Can be an addition in a separate PR tho since this PR is quite large.

frontend/src/static/js/v_authorship.js
```diff
@@ -169,22 +170,29 @@ window.vAuthorship = {
...
       files.forEach((file) => {
-        const lineCnt = file.authorContributionMap[this.info.author];
-        if (lineCnt) {
-          totalLineCount += lineCnt;
-          const out = {};
-          out.path = file.path;
-          out.lineCount = lineCnt;
-          out.fileType = file.fileType;
+        // skip file not touched by author
+        if (!(this.info.author in file.authorContributionMap)) {
```
jamessspanggg: Question: Why is this logic added for this PR?

0blivious:
Noticed that the icon to blame view is there for binary files. Is this really needed?

Also, clicking on any check box for file type will clear the `filter by glob` and make the radio button focus back to the one in front of file type. However, this is not working for binary.

frontend/src/static/js/v_authorship.js
ccyccyccy: The original intention of `if(lineCnt)` is to check if the author has contributed to the file.
However, it will fail for binary files as I assigned line count of binary files to be 0.
So instead I check directly for the existence of the author instead.

ccyccyccy:
> Noticed that the icon to blame view is there for binary files. Is this really needed?

Don't think it's needed. The blame view is just gibberish. I will remove the icon.

> Also, clicking on any check box for file type will clear the `filter by glob` and make the radio button focus back to the one in front of file type. However, this is not working for binary.

Thanks for noticing! I did not know about that interaction. Fixed it.

frontend/src/static/css/v_authorship.scss
```diff
@@ -148,6 +152,19 @@
       }
     }
 
+    .binary-segment {
+      background-color: mui-color('white');
+      .indicator {
```
Tejas2805: add an extra line here

frontend/src/static/css/v_authorship.scss
```diff
@@ -148,6 +152,19 @@
...
+        float: left;
+        font-weight: bold;
+        padding-left: 1rem;
+      }
+      .bin-text {
```
Tejas2805: add an extra line here

ccyccyccy:
> There is a section in the UG which says binary files are not tracked. Can update that too in this PR.

Thanks for the catch. Updated.

ccyccyccy:
@fzdy1914 Need your help for reviewing here~

ccyccyccy:
> 
> 
> A bit long to review. Meanwhile, in which test repo and which author can I see your change that include binary file?

No problem. The netlify preview doesn't include any binary files, so I hosted one [here](ccyyccyy.surge.sh). There are binary files under Jame's authorship. 

I hosted an earlier version one because the current master branch have some bug regarding the authorship panel. (See [here](https://github.com/reposense/RepoSense/issues/1167))

src/main/java/reposense/authorship/model/FileResult.java
```diff
@@ -13,15 +13,35 @@
...
+
+    public static FileResult createNonBinaryFileResult(String path, FileType fileType, ArrayList<LineInfo> lines,
+            HashMap<Author, Integer> authorContributionMap) {
+        return new FileResult(path, fileType, lines, authorContributionMap, false);
+    }
+
+    public static FileResult createBinaryFileResult(String path, FileType fileType, HashMap<Author,
+            Integer> authorContributionMap) {
+        return new FileResult(path, fileType, new ArrayList<>(), authorContributionMap, true);
+    }
```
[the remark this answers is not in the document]
ccyccyccy: I don't think we should combine the two together.

The caller would need to insert the empty `ArrayList<LineInfo> lines` manually if the methods were combined, which is not ideal.

The method names are also more obvious, as compared to a `true` or `false` passed in as argument into a combined method.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -69,13 +71,25 @@
...
+     * Extracts a list of relevant binary files given in {@code config}.
+     */
+    public static List<FileInfo> extractBinaryFileInfos(RepoConfiguration config) {
+        logger.info(String.format(MESSAGE_START_EXTRACTING_BINARY_FILE_INFO, config.getLocation(), config.getBranch()));
+
+        List<FileInfo> binaryFileInfos = new ArrayList<>();
+        getAllFileInfo(config, binaryFileInfos, true);
+        binaryFileInfos.sort(Comparator.comparing(FileInfo::getPath));
+        return binaryFileInfos;
+    }
```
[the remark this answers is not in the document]
ccyccyccy: Hmm don't think there's a duplicate here either. `extractFileInfos` is doing a lot of checks to ensure that the line by line analysis is done correctly, but `extractBinaryFileInfos` doesn't care about that.

I refactored the code a little to make the difference more obvious.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -123,14 +137,15 @@
     }
 
     /**
-     * Returns a {@code Set} of non-binary files for the repo {@code repoConfig}.
+     * Returns a {@code Set} of files for the repo {@code repoConfig}, filtering the files depending on whether
+     * binary files {@code isBinaryFiles} are requested.
```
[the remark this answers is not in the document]
ccyccyccy: Thanks! I was having trouble on how to properly describe the function.

ccyccyccy:
@jylee-git Thanks for the review! 

> Also, is it possible if we modify the configs for one the repo(s) (e.g.: adding formats `pptx`, `png`, `jpg` for RepoSense itself) such that we are able to preview the list of binary files changed?

I think we will need to whitelist them on the master branch at `_reposense/config.json`, since we are currently targeting the master branch of Reposense in `repo-config.csv`.

Should I open another PR to do this?

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -69,13 +71,24 @@
...
 
         fileInfos.sort(Comparator.comparing(FileInfo::getPath));
         return fileInfos;
     }
 
+    /**
+     * Extracts a list of relevant binary files given in {@code config}.
+     */
+    public static List<FileInfo> extractBinaryFileInfos(RepoConfiguration config) {
+        logger.info(String.format(MESSAGE_START_EXTRACTING_BINARY_FILE_INFO, config.getLocation(), config.getBranch()));
```
jylee-git: No need for the extra log, it can give too much clutter to the terminal output

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -69,13 +71,24 @@
...
+
+        List<FileInfo> binaryFileInfos = getAllFileInfo(config, true);
+        binaryFileInfos.sort(Comparator.comparing(FileInfo::getPath));
+        return binaryFileInfos;
+    }
```
jylee-git: If the user only wants to analyze between any two periods of time, and that binary files were not modified between the two specified times (note that these binary files could exist before the analysis start date), will this method capture that as well?

If so, shouldn't we exclude such binary files since it wasn't modified before?

ccyccyccy: The current method would capture all binary files, regardless of whether it was modified.

However, the actual date range filtering is done under `GitLog#getBinaryFileAuthors`, which would return an empty string if it was not modified within the specified time.

src/main/java/reposense/authorship/FileInfoExtractor.java
ccyccyccy: Removed.

ccyccyccy:
@fzdy1914 @jylee-git Are there any further reviews for this?

jylee-git:
> @jylee-git Thanks for the review!
> 
> > Also, is it possible if we modify the configs for one the repo(s) (e.g.: adding formats `pptx`, `png`, `jpg` for RepoSense itself) such that we are able to preview the list of binary files changed?
> 
> I think we will need to whitelist them on the master branch at `_reposense/config.json`, since we are currently targeting the master branch of Reposense in `repo-config.csv`.
> 
> Should I open another PR to do this?

@fzdy1914 your inputs for this.

jylee-git:
@eugenepeh can I get your help to see if the implementation for the backend part is ok? Thanks for your time.

frontend/src/tabs/authorship.pug
```diff
@@ -50,26 +50,37 @@
               span(v-bind:title="getFileTypeBlankLineInfo(fileType)")
                 span {{ fileType }}: {{ getFileTypeExistingLinesObj[fileType] }}&nbsp;
                 span.bloc ({{ fileTypeBlankLinesObj[fileType] }})
+          label(v-if="numberOfBinaryFiles > 0")
+            input.mui-checkbox--fileType(type="checkbox", v-bind:checked="isBinaryFilesChecked",
+              v-on:change="selectBinaryFile")
```
eugenepeh: line wrap indent

frontend/src/tabs/authorship.pug
```diff
@@ -50,26 +50,37 @@
...
+            span.binary-fileType(v-bind:title="this.numberOfBinaryFiles + \
+            ' binary files (not included in total line count)'")
```
eugenepeh: same for here

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -123,14 +132,16 @@
     }
 
     /**
-     * Returns a {@code Set} of non-binary files for the repo {@code repoConfig}.
+     * Returns a {@code Set} of non-binary files for the repo {@code repoConfig}
+     * if {@code isBinaryFiles} is set to `false`.
+     * Otherwise, returns a {@code Set} of binary files for the repo {@code repoConfig}
      */
-    public static Set<Path> getNonBinaryFilesList(RepoConfiguration repoConfig) {
+    public static Set<Path> getFilesList(RepoConfiguration repoConfig, boolean isBinaryFiles) {
```
eugenepeh: `isBinaryFile`

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -123,14 +132,16 @@
...
         List<String> modifiedFileList = GitDiff.getModifiedFilesList(Paths.get(repoConfig.getRepoRoot()));
 
-        // Gets rid of binary files and files with invalid directory name.
+        // Gets rid of files with invalid directory name and filters by whether file is binary.
         return modifiedFileList.stream()
-                .filter(file -> !file.startsWith(BINARY_FILE_LINE_DIFF_RESULT))
+                .filter(file -> isBinaryFiles == file.startsWith(BINARY_FILE_LINE_DIFF_RESULT))
                 .map(rawNonBinaryFile -> rawNonBinaryFile.split("\t")[2])
```
eugenepeh: how would this affect binary file?

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -41,7 +44,7 @@
      * Returns null if the file is missing from the local system, or none of the
      * {@code Author} specified in {@code config} contributed to the file in {@code fileInfo}.
      */
-    public static FileResult analyzeFile(RepoConfiguration config, FileInfo fileInfo) {
+    public static FileResult analyzeNonBinaryFile(RepoConfiguration config, FileInfo fileInfo) {
```
eugenepeh: instead of `NonBinaryFile`, can we name it `TextFile` for simplicity?

src/systemtest/resources/sinceBeginningDateRange/expected/reposense_testrepo-Beta_master/authorship.json
```diff
@@ -2,6 +2,7 @@
   {
     "path": "src/main/resources/view/InfoPanel.fxml",
     "fileType": "fxml",
+    "isBinary": false,
```
eugenepeh: is it possible to output this field only if its true?

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -151,13 +151,20 @@ public void getNonBinaryFilesList_directoryWithBinaryFiles_success() {
...
         // Binary files should be ignored
-        binaryFilesList.forEach(binFile -> Assert.assertFalse(files.contains(Paths.get(binFile))));
+        binaryFilesList.forEach(binFile -> Assert.assertFalse(nonBinaryFiles.contains(Paths.get(binFile))));
+
+        Set<Path> binaryFiles = FileInfoExtractor.getFilesList(config, true);
+        Assert.assertEquals(5, binaryFiles.size());
+        // Binary files should be captured
+        binaryFilesList.forEach(binFile -> Assert.assertTrue(binaryFiles.contains(Paths.get(binFile))));
+        // Non binary files should be ignored
+        nonBinaryFilesList.forEach(nonBinFile -> Assert.assertFalse(binaryFiles.contains(Paths.get(nonBinFile))));
```
eugenepeh: Can we have this test separated? It is not very informative on which part causes the failure should this test fails.

src/main/java/reposense/authorship/AuthorshipReporter.java
```diff
@@ -18,13 +18,22 @@
      * Generates and returns the authorship summary for each repo in {@code config}.
      */
     public static AuthorshipSummary generateAuthorshipSummary(RepoConfiguration config) {
-        List<FileInfo> fileInfos = FileInfoExtractor.extractFileInfos(config);
+        List<FileInfo> fileInfos = FileInfoExtractor.extractNonBinaryFileInfos(config);
 
         List<FileResult> fileResults = fileInfos.stream()
-                .map(fileInfo -> FileInfoAnalyzer.analyzeFile(config, fileInfo))
+                .map(fileInfo -> FileInfoAnalyzer.analyzeNonBinaryFile(config, fileInfo))
```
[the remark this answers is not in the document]
ccyccyccy: Yep. Fixed.

src/main/java/reposense/authorship/FileInfoExtractor.java
ccyccyccy: Thanks for the review! Fixed.

src/main/java/reposense/authorship/FileInfoExtractor.java
ccyccyccy: Thanks for pointing that out.

This is the output for non binary file and binary file respectively.
```
3	0	src/test/resources/repoconfig_merge_test/repo-config.csv
-	-	docs/diagrams/architecture.pptx
```
So both types of file need it to work.

ccyccyccy: I will change the variable name to `file`

frontend/src/tabs/authorship.pug
ccyccyccy: fixed.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
ccyccyccy: Do you mean just here, or everywhere else as well?

src/test/java/reposense/authorship/FileInfoExtractorTest.java
ccyccyccy: Yep didn't realise that. Fixed.

src/systemtest/resources/sinceBeginningDateRange/expected/reposense_testrepo-Beta_master/authorship.json
ccyccyccy: It's possible. I'll just need to add a new type adapter [here](https://github.com/reposense/RepoSense/blob/0e39bb7ffa5094e13fb120a36e3c2666f2566572/src/main/java/reposense/util/FileUtil.java#L94). Should I do it?

ccyccyccy:
@eugenepeh @fzdy1914 @jylee-git Do you have any more opinions on this PR?

src/main/java/reposense/authorship/FileInfoAnalyzer.java
eugenepeh: everywhere. A binary file is commonly known as non-text file. Hence a nonbinary can be simplified as textfile.

src/systemtest/resources/sinceBeginningDateRange/expected/reposense_testrepo-Beta_master/authorship.json
eugenepeh: or another way is, we make it a default null field.
iirc, null object fields wont be outputted

src/main/java/reposense/authorship/FileInfoAnalyzer.java
ccyccyccy: Ok changed.

src/systemtest/resources/sinceBeginningDateRange/expected/reposense_testrepo-Beta_master/authorship.json
ccyccyccy: Thanks for the suggestion. Did just that.

frontend/src/static/js/v_authorship.js
```diff
@@ -294,6 +312,10 @@ window.vAuthorship = {
           });
       return numLinesModified;
     },
+
+    numberOfBinaryFiles() {
```
eugenepeh, line 316: `getBinaryFilesCount` will probably be better

frontend/src/tabs/authorship.pug
```diff
@@ -46,31 +46,46 @@
               span.bloc.medium-font ({{ totalBlankLineCount }})
           template(v-for="fileType in Object.keys(getFileTypeExistingLinesObj)")
             label(v-bind:key="fileType")
-              input.mui-checkbox--fileType(type="checkbox", v-bind:id="fileType", v-bind:value="fileType",
+              input.mui-checkbox--fileType(
+                type="checkbox",
```
eugenepeh, line 50: is this change necessary?

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -123,15 +132,17 @@
     }
 
     /**
-     * Returns a {@code Set} of non-binary files for the repo {@code repoConfig}.
+     * Returns a {@code Set} of non-binary files for the repo {@code repoConfig}
+     * if {@code isBinaryFiles} is set to `false`.
+     * Otherwise, returns a {@code Set} of binary files for the repo {@code repoConfig}
      */
-    public static Set<Path> getNonBinaryFilesList(RepoConfiguration repoConfig) {
+    public static Set<Path> getFilesList(RepoConfiguration repoConfig, boolean isBinaryFile) {
```
eugenepeh, line 139: `List` can be omitted, since what its really returning is a Set anyway.

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -123,15 +132,17 @@
...
         List<String> modifiedFileList = GitDiff.getModifiedFilesList(Paths.get(repoConfig.getRepoRoot()));
 
-        // Gets rid of binary files and files with invalid directory name.
```
eugenepeh, line 131: i think the original comment is simpler, should keep it

src/main/java/reposense/authorship/FileInfoExtractor.java
```diff
@@ -224,7 +243,7 @@ private static int getStartingLineNumber(String linesChanged) {
     /**
      * Returns true if {@code filePath} is valid and the file is not in binary.
      */
-    private static boolean isValidAndNonBinaryFile(String filePath, Set<Path> nonBinaryFilesSet) {
-        return FileUtil.isValidPath(filePath) && nonBinaryFilesSet.contains(Paths.get(filePath));
+    private static boolean isValidAndTextFile(String filePath, Set<Path> textFilesSet) {
```
eugenepeh, line 246: maybe just isValidTextFile

frontend/src/tabs/authorship.pug
ccyccyccy: Nope. Will revert.

src/main/java/reposense/authorship/FileInfoExtractor.java
ccyccyccy: Changed.

src/main/java/reposense/authorship/FileInfoExtractor.java
ccyccyccy: But this function can either get rid of binary files, or keep only binary files. The original comment would be misleading.

src/main/java/reposense/authorship/FileInfoExtractor.java
ccyccyccy: FIxed.

frontend/src/static/js/v_authorship.js
ccyccyccy: I don't think the `get` is necessary since it is a computed value. How about just `binaryFilesCount`?

src/main/java/reposense/authorship/FileInfoExtractor.java
eugenepeh: i see then maybe reword the sentence, `filters by the {@code isBinaryFile} flag.` will be better

frontend/src/static/js/v_authorship.js
eugenepeh: yes, that would be nice; keep it consistent to the naming convention

src/main/java/reposense/authorship/FileInfoExtractor.java
eugenepeh: i see then maybe reword the sentence, `filters by the {@code isBinaryFile} flag.` will be better

frontend/src/static/js/v_authorship.js
ccyccyccy: Will do.

src/main/java/reposense/authorship/FileInfoExtractor.java
ccyccyccy: Ok changed.

fzdy1914:
Can you help merge the conflicts, please?

fzdy1914:
> 2/3 observations, maybe @fzdy1914 can guide.
> 
> 1. Maybe we should update the UG to inform that **Binary files** count is not added in **All**, so that users get to know about it.

Yes, we can.

> 2. Couldn't find any discussion on this. So just a question - Why are we not displaying the `Binary` label? Is it because we want to treat them differently? Should still show the label in my opinion.

I think we should display the label.

> 1. We are using `+` for string interpolation in one file in this PR. Should we leave it here and fix it in #1308 or get it fixed in this PR.

Just leave it here.

Tejas2805:
@ccyccyccy Could you just make the required changes? Then we can merge this PR.

Tejas2805:
@fzdy1914 @jamessspanggg I think it might be a good idea for someone to take over this PR and complete it. Can ask the new students trying for CS3281.

fzdy1914:
> @fzdy1914 @jamessspanggg I think it might be a good idea for someone to take over this PR and complete it. Can ask the new students trying for CS3281.

Yes, we can.

PR #1057, opened 2020-02-05
[#910] Save authorship tab controls as hash parameters

### LINKED ISSUE #910: Code view: controls not restored when reloading page
Authorship page with modified controls are not restored when reloading the page:

reload

ang-zeyu:
Resolves #910 

```
Save authorship tab controls as hash parameters

The filter and sorting options for the authorship tab are not persisted
across page reloads.

This results having to re-select the options whenever the page is
reloaded, leading to a less pleasant user experience.

Let's save the options for these controls as hash parameters, and use
them across page reloads when available.
```

### DIFF
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -191,14 +191,7 @@ window.app = new window.Vue({
       window.addHash('tabOpen', this.isTabActive);
       window.removeHash('tabType');
       this.removeZoomHashes();
-      this.removeAuthorshipHashes();
       window.encodeHash();
-    },
-
-    removeAuthorshipHashes() {
-      window.removeHash('tabAuthor');
-      window.removeHash('tabRepo');
-      window.removeHash('authorshipIsMergeGroup');
     },
 
     removeZoomHashes() {
@@ -222,7 +215,6 @@ window.app = new window.Vue({
       this.activateTab('authorship');
     },
     updateTabZoom(obj) {
-      this.removeAuthorshipHashes();
       this.tabInfo.tabZoom = Object.assign({}, obj);
       this.activateTab('zoom');
     },
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -41,23 +41,56 @@ window.vAuthorship = {
       totalBlankLineCount: '',
       filesSortType: 'lineOfCode',
       toReverseSortFiles: true,
-      filterSearch: '*',
+      searchBarValue: '',
     };
   },
 
   watch: {
-    filterType() {
-      if (this.filterType === 'checkboxes') {
-        const searchBar = document.getElementById('search');
-        searchBar.value = '';
-        this.filterSearch = '*';
-      } else {
-        this.selectedFileTypes = this.fileTypes.slice();
+    filesSortType() {
+      window.addHash('authorshipSortBy', this.filesSortType);
+      window.encodeHash();
+    },
+
+    toReverseSortFiles() {
+      window.addHash('reverseAuthorshipOrder', this.toReverseSortFiles);
+      window.encodeHash();
+    },
+
+    isLoaded() {
+      if (this.isLoaded) {
+        this.retrieveHashes();
       }
     },
   },
 
   methods: {
+    retrieveHashes() {
+      window.decodeHash();
+      const hash = window.hashParams;
+
+      switch (hash.authorshipSortBy) {
+      case 'path':
+      case 'fileName':
+      case 'fileType':
+        this.filesSortType = hash.authorshipSortBy;
+        break;
+      default:
+        // Invalid value, use the default value of 'lineOfCode'
+      }
+
+      if (hash.reverseAuthorshipOrder === 'false') {
+        this.toReverseSortFiles = false;
+      }
+
+      if ('authorshipFilesGlob' in hash) {
+        this.indicateSearchBar();
+        this.searchBarValue = hash.authorshipFilesGlob;
+      } else if ('authorshipFileTypes' in hash) {
+        const parsedFileTypes = hash.authorshipFileTypes.split('~');
+        this.selectedFileTypes = parsedFileTypes.filter((type) => this.fileTypes.includes(type));
+      }
+    },
+
     initiate() {
       const repo = window.REPOS[this.info.repo];
 
@@ -247,11 +280,22 @@ window.vAuthorship = {
       filesInfoObj[fileType] += lineCount;
     },
 
-    updateFilterSearch(evt) {
-      if (this.filterType === 'checkboxes') {
-        this.indicateSearchBar();
-      }
-      this.filterSearch = (evt.target.value.length !== 0) ? evt.target.value : '*';
+    updateSearchBarValue() {
+      this.searchBarValue = this.$refs.searchBar.value;
+
+      window.addHash('authorshipFilesGlob', this.searchBarValue);
+      window.removeHash('authorshipFileTypes');
+      window.encodeHash();
+    },
+
+    updateFileTypeHash() {
+      const fileTypeHash = this.selectedFileTypes.length > 0
+          ? this.selectedFileTypes.reduce((a, b) => `${a}~${b}`)
+          : '';
+
+      window.addHash('authorshipFileTypes', fileTypeHash);
+      window.removeHash('authorshipFilesGlob');
+      window.encodeHash();
     },
 
     indicateSearchBar() {
@@ -260,12 +304,9 @@ window.vAuthorship = {
     },
 
     indicateCheckBoxes() {
-      if (this.filterType === 'search') {
-        const searchBar = document.getElementById('search');
-        searchBar.value = '';
-        this.filterSearch = '*';
-        this.filterType = 'checkboxes';
-      }
+      this.searchBarValue = '';
+      this.filterType = 'checkboxes';
+      this.updateFileTypeHash();
     },
 
     getFileLink(file, path) {
@@ -298,20 +339,19 @@ window.vAuthorship = {
         return this.selectedFileTypes.length === this.fileTypes.length;
       },
       set(value) {
-        if (this.filterType === 'search') {
-          this.indicateCheckBoxes();
-        }
         if (value) {
           this.selectedFileTypes = this.fileTypes.slice();
         } else {
           this.selectedFileTypes = [];
         }
+
+        this.indicateCheckBoxes();
       },
     },
 
     selectedFiles() {
       return this.files.filter((file) => this.selectedFileTypes.includes(file.fileType)
-          && minimatch(file.path, this.filterSearch, { matchBase: true, dot: true }))
+          && minimatch(file.path, this.searchBarValue || '*', { matchBase: true, dot: true }))
           .sort(this.sortingFunction);
     },
 
@@ -333,6 +373,14 @@ window.vAuthorship = {
   created() {
     this.initiate();
     this.setInfoHash();
+  },
+
+  beforeDestroy() {
+    window.removeHash('authorshipFileTypes');
+    window.removeHash('authorshipFilesGlob');
+    window.removeHash('authorshipSortBy');
+    window.removeHash('reverseAuthorshipOrder');
+    window.encodeHash();
   },
 
   components: {
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -59,6 +59,7 @@ window.vAuthorship = {
     isLoaded() {
       if (this.isLoaded) {
         this.retrieveHashes();
+        this.setInfoHash();
       }
     },
   },
@@ -78,17 +79,35 @@ window.vAuthorship = {
         // Invalid value, use the default value of 'lineOfCode'
       }
 
-      if (hash.reverseAuthorshipOrder === 'false') {
-        this.toReverseSortFiles = false;
-      }
+      this.toReverseSortFiles = hash.reverseAuthorshipOrder !== 'false';
 
       if ('authorshipFilesGlob' in hash) {
         this.indicateSearchBar();
         this.searchBarValue = hash.authorshipFilesGlob;
       } else if ('authorshipFileTypes' in hash) {
-        const parsedFileTypes = hash.authorshipFileTypes.split('~');
+        const parsedFileTypes = hash.authorshipFileTypes.split(window.HASH_FILETYPE_DELIMITER);
         this.selectedFileTypes = parsedFileTypes.filter((type) => this.fileTypes.includes(type));
       }
+    },
+
+    setInfoHash() {
+      const { addHash, encodeHash } = window;
+      // We only set these hashes as they are propagated from summary_charts
+      addHash('tabAuthor', this.info.author);
+      addHash('tabRepo', this.info.repo);
+      addHash('authorshipIsMergeGroup', this.info.isMergeGroup);
+      encodeHash();
+    },
+
+    removeAuthorshipHashes() {
+      window.removeHash('authorshipFileTypes');
+      window.removeHash('authorshipFilesGlob');
+      window.removeHash('authorshipSortBy');
+      window.removeHash('reverseAuthorshipOrder');
+      window.removeHash('tabAuthor');
+      window.removeHash('tabRepo');
+      window.removeHash('authorshipIsMergeGroup');
+      window.encodeHash();
     },
 
     initiate() {
@@ -140,14 +159,6 @@ window.vAuthorship = {
           this.filesLinesObj[type] = cnt;
         }
       });
-    },
-
-    setInfoHash() {
-      const { addHash, encodeHash } = window;
-      addHash('tabAuthor', this.info.author);
-      addHash('tabRepo', this.info.repo);
-      addHash('authorshipIsMergeGroup', this.info.isMergeGroup);
-      encodeHash();
     },
 
     expandAll() {
@@ -257,7 +268,6 @@ window.vAuthorship = {
 
       Object.keys(this.filesLinesObj).forEach((file) => {
         if (this.filesLinesObj[file] !== 0) {
-          this.selectedFileTypes.push(file);
           this.fileTypes.push(file);
         }
       });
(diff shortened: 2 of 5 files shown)


ang-zeyu:
> A quick look: shouldn't we store is checkbox or is search bar too? However, another ongoing PR is touching on that, so the rest part of this PR can continue after #1040 is merged.

Missed that out, thanks! I'll update this PR again once #1040 is merged

ang-zeyu:
> A quick look: shouldn't we store is checkbox or is search bar too? However, another ongoing PR is touching on that, so the rest part of this PR can continue after #1040 is merged.

Updated; Realised it isn't necessary to store isCheckbox/SearchBar as well, since we can just use the presence of either `authorshipFileTypes` or `authorshipFilesGlob` ( which are mutually exclusive ) to determine `isCheckbox/SearchBar`

For #947, I would reintroduce the watchers but leave the computed property change, so this PR should be more or less the same even after that is merged.

frontend/src/static/js/v_authorship.js
```diff
@@ -224,24 +270,30 @@ window.vAuthorship = {
...
     },
 
     indicateCheckBoxes() {
-      if (this.filterType === 'search') {
-        const searchBar = document.getElementById('search');
-        searchBar.value = '';
-        this.filterSearch = '*';
-        this.filterType = 'checkboxes';
-      }
+      this.filterType = 'checkboxes';
```
ang-zeyu: I removed this as the filterType watcher already executes this chunk with `updateXxHash` as well, to avoid duplication, likewise for the line in `indicateSearchbar`

frontend/src/static/js/v_authorship.js
```diff
@@ -42,32 +42,74 @@ window.vAuthorship = {
...
+        // Invalid value, use the default value
+      }
+
+      if (hash.reverseAuthorshipOrder === 'false') {
+        this.toReverseSortFiles = false;
+      }
+
+      if (hash.authorshipFilesGlob) {
+        this.indicateSearchBar();
+        this.searchBarValue = hash.authorshipFilesGlob === '*'
```
ang-zeyu: A new variable is needed as the dom might not have mounted when `retrieveHashes` is executed.
In such a case, `document.getElementById('search')` will fail.

frontend/src/static/js/v_authorship.js
```diff
@@ -42,32 +42,74 @@ window.vAuthorship = {
...
+      }
     },
     selectedFiles() {
       setTimeout(this.updateCount, 0);
     },
     filterType() {
       if (this.filterType === 'checkboxes') {
-        const searchBar = document.getElementById('search');
-        searchBar.value = '';
+        this.searchBarValue = '';
```
ang-zeyu: Consolidated `document.getElementById('search')` calls into to use `this.searchBarValue` instead as per below comment

ang-zeyu:
Updated on top of #1074

fzdy1914:
For the reviewing process, please request a review from @reposense/stage0-reviewers first. After getting at least one approval, request a review from me and jgit.

fzdy1914:
Meanwhile, just a question, would you like to join stage0 reviewer?

ang-zeyu:
> For the reviewing process, please request a review from @reposense/stage0-reviewers first. After getting at least one approval, request a review from me and jgit.

Noted on the order of reviewers, thanks for requesting the review as well!

> Meanwhile, just a question, would you like to join stage0 reviewer?

Yes, I'd be glad to!

fzdy1914:
> Yes, I'd be glad to!

Please accept the invitation first to be a member of reposense

ang-zeyu:
Updated on latest, no conflicts/changes

ang-zeyu:
@Tejas2805 could I get your review on this?

Tejas2805:
@ang-zeyu  Does this overlap with [issue](https://github.com/reposense/RepoSense/issues/1151)

ang-zeyu:
> @ang-zeyu Does this overlap with [issue](https://github.com/reposense/RepoSense/issues/1151)?

Nope, the file type hashed here is the one for the authorship panel only

frontend/src/static/js/v_authorship.js
0blivious: I dont really get the reason for having this new variable? Can just set filterSearch = '' in data and bind it to the search bar, then in `selectedFiles` do sth like `minimatch(file.path, this.filterSearch || '*'`

0blivious:
I was wondering why need to update the hash every time the variable change? Is it ok to only do this when we detect window reloading?

frontend/src/static/js/v_authorship.js
ang-zeyu: Good idea, I think I'll make `this.filterSearch || '*'` a computed property, thanks!

ang-zeyu:
> I was wondering why need to update the hash every time the variable change? Is it ok to only do this when we detect window reloading?

That's one option, I went with following the current behaviour in other places such as the summary view. Might be nice for a user to be able to share a link without having to reload the window.

0blivious:
Bug occured when
1. Go to any authorship with some commits
2. Enter some random character in Filter by Glob (eg 'd') -> no file should be displayed
3. Clear the Filter by Glob bar then hit enter, still no file is displayed. I would expect all files to be displayed

ang-zeyu:
> Bug occured when
> 
> 1. Go to any authorship with some commits
> 2. Enter some random character in Filter by Glob (eg 'd') -> no file should be displayed
> 3. Clear the Filter by Glob bar then hit enter, still no file is displayed. I would expect all files to be displayed

Thanks for catching this! Fixed it

0blivious:
The saved hash parameter affect other authorship page
1. open `Yong Hao` authorship page
2. untick all to hide all files
3. open `Eugene` authorship page
4. no file will be displayed as `all` is unticked
I would expect `all` to be ticked though as the selection I made on other authorship page should not affect the new authorship page I just opened

frontend/src/static/js/v_authorship.js
```diff
@@ -253,6 +293,10 @@ window.vAuthorship = {
   },
 
   computed: {
+    filterSearchDisplay() {
+      return this.filterSearch === '*' ? '' : this.filterSearch;
```
0blivious: I do not feel we even need this new computed property. Just filterSearch should be enough and the logic can be simplified. The only reason we want the initial value of filterSearch to be * is because we need this for the minimatch library. Thus, it is all right to set filterSearch = '' in data and bind input v-model = filterSearch. Then in the `selectedFiles` we change the filter function to `minimatch(file.path, this.filterSearch || '*', { matchBase: true, dot: true }))` so empty string will match all file names.

ang-zeyu: 👍 Changed the 'computed property' to `this.filterSearch || '*'` instead.
Changed `filterSearch` -> `searchBarValue` as well since its not referring to the value used in `minimatch` anymore as well.

> bind input v-model = filterSearch

This works, but the likely reason it was done otherwise (before this pr) is that it would cause `selectedFiles` to change on every key type, which causes performance to tank

ang-zeyu:
> The saved hash parameter affect other authorship page

> When selecting to view another author's code, if the other author's code has filetypes that are not within the filetypes of the previous author, those will not be checked.

Thanks for bringing this up, forgot to discuss this. I considered the following:

- A simple apparent fix would be to clear the authorship tab hashes before destroying the component. Bue due to race conditions, the creation of the other component may finish before the `beforeDestroy` of the current authorship component finishes, which causes the hashes to still be transferred over.
- A second fix would be to move hash application from `created` to `mounted`. The issue here again is that we have asynchronous operations in `created` ( `loadAuthorship` ). This would require introducing some data variable to know when said asychronous operation is finished, which there fortunately already is ( `isLoaded` ). This is what's used currently.

- A third option would be to use hashes as a means to persist options data authorship tab switches. If we are sticking with the `v-bind:key` approach for authorship, this can be very useful. Otherwise if we eventually aim to reintroduce #967, this is not so good an option.
  - If we go with this, I could make `authorshipFileTypes=java~md~csv~...` be the _excluded_ file types instead of included, which would fix the problem you mentioned above. ( files excluded before will still be excluded, but files not present before would be re-included on switching authors )

Which would be preferred? :o @0blivious @jamessspanggg

ang-zeyu:
Do ignore the earlier post ( 2nd point explains the current implementation of not persisting hashes across authors though ), I realised it would be better to eventually move hashes to global state management to solve the problem of persisting options across tabs.

Hence this PR does not modify that behaviour in anyway ( when the authorship tabs are switched, _all_ options are reset ), and only hashes the authorship tab options as per the title :)

0blivious:
Can you test the file detail display based on your implementation here and #1065. It should be that if I collapse some file, after reloading the page, the file still remains collapsed.

ang-zeyu:
> Can you test the file detail display based on your implementation here and #1065. It should be that if I collapse some file, after reloading the page, the file still remains collapsed.

This PR dosen't target file collapsing, I don't think having it would be worth the amount of data we need to add to the hash `collapsedFiles=frontend/src/static/css/_colors.scss~...~...~*quite a few files`.

frontend/src/tabs/authorship.pug
```diff
@@ -29,13 +29,32 @@
...
-        form.file-picker.mui-form--inline(onsubmit="return false")
-          input#search.medium-font(type="search", placeholder="Filter by glob", v-on:change="updateFilterSearch",
-            v-on:click="indicateSearchBar")
-          button#submit-button.medium-font(type="submit", v-on:click="indicateSearchBar") Filter
+        input.radio-button--search(
+          type="radio",
+          value="search",
+          v-model="filterType",
+          v-on:change="indicateSearchBar(); updateSearchBarValue();"
+          )
```
0blivious: should this closing bracket be on the same col as input? was checking the style in index.pug

ang-zeyu: Yup, thanks! Updated it

ang-zeyu:
ps: *Travis is passing but the status isn't being reported back, have tried rebuilding here and on travis quite a few times

frontend/src/static/js/v_authorship.js
```diff
@@ -335,6 +375,14 @@ window.vAuthorship = {
     this.setInfoHash();
   },
 
+  beforeDestroy() {
+    window.removeHash('authorshipFileTypes');
```
jamessspanggg, line 379: hmm, should here also remove `tabAuthor`, `tabRepo` and `authorshipIsMergeGroup` instead of removing in `main.js`?

ang-zeyu: Good idea, didn't notice this. Would be good to centralize where authorship hashes are removed

I couldn't find a `authorshipIsMergeGroup` hash though, was this removed recently?

jamessspanggg: `authorshipIsMergeGroup` was added recently in PR #1175. Look at `main.js`.

ang-zeyu: Ahh my bad, thought I was on the latest already, thanks!

ang-zeyu: updated, should we open an issue to investigate doing the same for the zoom tab?

jamessspanggg: > updated, should we open an issue to investigate doing the same for the zoom tab?

Yes please do. I believe it will be much cleaner for the component themselves to remove their hashes.

frontend/src/static/js/main.js
```diff
@@ -237,7 +237,7 @@ window.app = new window.Vue({
       if (Object.keys(info).length === tabInfoLength) {
         this.updateTabAuthorship(info);
       } else if (hash.tabOpen === 'false' || tabInfoLength > 2) {
-        window.app.isTabActive = false;
+        this.isTabActive = false;
```
jamessspanggg, line 240: is this change needed? Also, what was the cause of the previous issue?

ang-zeyu: > hmm, should here also remove tabAuthor, tabRepo and authorshipIsMergeGroup instead of removing in main.js?

still not fixed, this actually exposed a race condition I haven't thought of before.
I may need to move all hashes (including ones introduced here) back to `main.js` - still thinking if there are better solutions though

ang-zeyu: sorry false alarm 😞

the issue was simply that I didn't move `setInfoHash` into/after `retrieveHashes` as well per my comment here https://github.com/reposense/RepoSense/pull/1057#issuecomment-604363267, which was the 'race condition' 🤦‍♂️ 
( it causes the hashes to be removed after `setInfohash` is called, which in turn causes `window.app.isTabActive = false` above to execute - hence the tab disappears  )

frontend/src/static/js/v_authorship.js
```diff
@@ -257,7 +259,6 @@ window.vAuthorship = {
 
       Object.keys(this.filesLinesObj).forEach((file) => {
         if (this.filesLinesObj[file] !== 0) {
-          this.selectedFileTypes.push(file);
```
ang-zeyu, line 227: redundant call, could this in a separate pr if preferred.
( there is a `this.selectedFileTypes = this.fileTypes.slice();` right below it )

frontend/src/static/js/v_authorship.js
```diff
@@ -372,14 +373,16 @@ window.vAuthorship = {
 
   created() {
     this.initiate();
-    this.setInfoHash();
```
ang-zeyu, line 335: moved into `retrieveHashes` as per the false alarm

frontend/src/static/js/v_authorship.js
```diff
@@ -41,23 +41,66 @@ window.vAuthorship = {
...
+        // Invalid value, use the default value of 'lineOfCode'
+      }
+
+      this.toReverseSortFiles = hash.reverseAuthorshipOrder !== 'false';
+
+      if ('authorshipFilesGlob' in hash) {
+        this.indicateSearchBar();
+        this.searchBarValue = hash.authorshipFilesGlob;
+      } else if ('authorshipFileTypes' in hash) {
+        const parsedFileTypes = hash.authorshipFileTypes.split('~');
```
jamessspanggg, line 93: perhaps can make this a one-liner? `this.selectedFileTypes = hash.authorshipFileTypes.split('~').filter...`

jamessspanggg:
@eugenepeh may I get your final review on this?

frontend/src/static/js/v_authorship.js
```diff
@@ -41,23 +41,66 @@ window.vAuthorship = {
...
+        this.selectedFileTypes = parsedFileTypes.filter((type) => this.fileTypes.includes(type));
+      }
+
+      // We only add these hashes since they are propagated from summary_charts
+      addHash('tabAuthor', this.info.author);
```
jamessspanggg, line 98: I believe the adding of hashes can be in a separate method (since it's not suited to be in a `retrieveHashes` method), and called during the `mounted` hook. I believe u added line 59-61 to avoid the adding of hashes being called before the `beforeDestroy` hook of the previous authorship component right?

frontend/src/static/js/v_authorship.js
```diff
@@ -332,7 +373,17 @@ window.vAuthorship = {
 
   created() {
     this.initiate();
-    this.setInfoHash();
+  },
+
+  beforeDestroy() {
+    window.removeHash('authorshipFileTypes');
+    window.removeHash('authorshipFilesGlob');
```
jamessspanggg, line 380: Can factor out all the removing of hashes to a separate method, and call the method in `beforeDestroy`.

frontend/src/static/js/v_authorship.js
jamessspanggg, line 93: Do you think we can define a global constant for delimiters? In the future, when we need to restore the checked file types in v_summary etc, it may be useful.

ang-zeyu: Updated except for this. Might be better to split long one liners into multiple `consts`, which are self-documenting.

frontend/src/static/js/v_authorship.js
jamessspanggg: The `setInfoHash` method can be moved to its original position (below `initiate`) to only show ur addition for the method

frontend/src/static/js/v_authorship.js
```diff
@@ -41,23 +41,75 @@ window.vAuthorship = {
...
+    },
+
+    toReverseSortFiles() {
+      window.addHash('reverseAuthorshipOrder', this.toReverseSortFiles);
+      window.encodeHash();
+    },
+
+    isLoaded() {
+      if (this.isLoaded) {
+        this.retrieveHashes();
```
jamessspanggg, line 61: would it be simpler to call these two methods in `mounted` hook instead?

ang-zeyu: won't work as explained here https://github.com/reposense/RepoSense/pull/1057#issuecomment-604363267 unfortunately

as for the above I moved it to group the 'lifecycle' hash methods together 😮

jamessspanggg: Ahhh I see, sorry for not referring back to the comment.

jamessspanggg:
Seems like the UI changed slightly after the merge conflict. Can fix it?
Screen Shot 2020-04-20 at 8 11 08 PM

ang-zeyu:
> Seems like the UI changed slightly after the merge conflict (the checkboxes). Can fix it?

fixed

frontend/src/tabs/authorship.pug
```diff
@@ -30,13 +30,32 @@
...
-          input#search.medium-font(type="search", placeholder="Filter by glob", v-on:change="updateFilterSearch",
-            v-on:click="indicateSearchBar")
-          button#submit-button.medium-font(type="submit", v-on:click="indicateSearchBar") Filter
+        input.radio-button--search(
+          type="radio",
+          value="search",
+          v-model="filterType",
+          v-on:change="indicateSearchBar(); updateSearchBarValue();"
+        )
+        .file-picker.mui-form--inline
```
jamessspanggg, line 39: Is the `form` removed on purpose?

ang-zeyu: yup, that was the cause of this, I removed it to prevent writing `v-on:xx.prevent` all over since we aren't using it as a form per se

> Seems like the UI changed slightly after the merge conflict (the checkboxes). Can fix it?
> Screen Shot 2020-04-20 at 8 11 08 PM

PR #1059, opened 2020-02-06
[#576] Refactor summary charts from summary

0blivious:
Fix #576
```
The entire summary-charts template resides in the summary pug file.

Let's refactorsummary-charts to move it into its own component so
that  reposense's dashboard is much more extendable in the future.
```

PR #1062, opened 2020-02-08
[#576] Refactor summary from index

0blivious:
Part of #576
```
The entire v_summary template resides in the index pug file.

Let's refactor v_summary to move it into its own component so that 
reposense's dashboard is much more extendable in the future.
```


0blivious:
#1059 will be reopened after this PR is merged

0blivious:
@fzdy1914 hi can I get your opinion on this so we can proceed to do the refactoring of charts from summary?

frontend/src/summary.pug
```diff
@@ -0,0 +1,192 @@
...
+
+        .summary-chart__contribution
+          template(v-if="filterBreakdown")
+            .summary-chart__contrib(
+              v-for="(widths, fileType) in getFileTypeContributionBars(user.fileTypeContribution)"
+            )
+              .summary-chart__contrib--bar(
+                v-for="width in widths",
+                v-bind:style="{ width: width + '%',\
+                              'background-color': contributionBarFileTypeColors[fileType] }",
```
fzdy1914: Weird line wrap/indentation starting from here. Can you check so that the original line wrap/indentation is not affected? Or the line wrap/indentation is configured according to the #987?

0blivious: fixed

0blivious:
> I understand the effort to refactor according to the standard. However, seems there are some difference between the standard discuss. How about let's foget about the line wrap but only consider the line indentation?

fixed, @fzdy1914 may I get your review on this?

PR #1064, opened 2020-02-08
[#1004] Make Authorship Panel consistent with Chart Panel

### LINKED ISSUE #1004: Inconsistency between Chart Panel and Authorship Panel
The current Chart Panel and Authorship Panel use different font size, color, background color and header height. As suggested by Prof, we can try to make the 2 panels more similar to make the website look more consistent.

Tejas2805:
Fixes #1004

### DIFF
--- a/frontend/src/static/css/style.scss
+++ b/frontend/src/static/css/style.scss
@@ -181,7 +181,7 @@ header {
   #tabs-wrapper {
     $tabs-width: 50%;
 
-    background: mui-color('grey', '200');
+    background: mui-color('white');
     flex: 0 0 $tabs-width;
     overflow-y: scroll;
     position: relative;
@@ -193,7 +193,7 @@ header {
       z-index: z-index('center-divider');
 
       svg {
-        background-color: mui-color('white');
+        background-color: mui-color('grey', '200');
         color: mui-color('black');
         padding: 3rem .25rem;
       }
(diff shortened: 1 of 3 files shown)


Tejas2805:
@damithc What do you think of the Ui?
 Based on our discussion, have followed github's style for file-header and made panels white

damithc:
GitHub UI has a way to clearly separate each file. We need to have something similar. As of now we have a horizontal line but that seems to separate the filename from the code, rather than showing where a file ends.

image

Tejas2805:
Screen Shot 2020-02-10 at 8 47 45 AM

@damithc What do you think

damithc:
> @damithc What do you think?

As Github does, use a very light grey for the box containing the file name?

Tejas2805:
@damithc I have added the Github colour. I have also suggested two options for borders. Let me know how to proceed.

1) 
Screen Shot 2020-02-11 at 2 59 05 AM


2) 
Screen Shot 2020-02-11 at 3 21 18 AM

damithc:
Option 2 looks better. May be reduce the thickness of the border between the filename and the content as there is no reason for it to be thicker than others?

Tejas2805:
Screen Shot 2020-02-11 at 7 25 22 PM

@damithc For review. Will proceed accordingly then.

damithc:
Are you able increase the curvature at the top and bottom corners like this? No curve for the middle one.
image

Tejas2805:
The initial one was 2.

Current top: 4
Current bottom: 8

Screen Shot 2020-02-11 at 8 23 46 PM

damithc:
> Current top: 4
> Current bottom: 8

let's make both 4 as there is no logical reason to make them different.

Tejas2805:
Closing this PR due to some issues. Have re=opened in #1080

Issue #202, opened 2018-07-17
Dashboard: prevent crash due to excessive diff size


damithc:
When an author's contribution is beyond a certain size, we need a way to prevent the dashboard crashing if the user tries to view the code of that author.

jamessspanggg:
Should we impose a limit on the number of lines we can display per file? e.g. if the number of lines of a file has > 5000 lines, we show a message such as:
`unable to display large file > 5000 lines, please click on the [here] to view the code`, where `[here]` links to the blame view of the file on GitHub.

damithc:
Yes, we can limit. If the file is bigger, we can open the code view of that file only in a separate tab? Redirecting to GitHub is not nice because the details we display is different from what GitHub displays.

jamessspanggg:
Hmm there might still be cases where the commits contain extremely large files as below and loading them might still not be feasible:

<img width="1097" alt="Screen Shot 2019-07-08 at 2 36 00 PM" src="https://user-images.githubusercontent.com/32864116/60788028-46fe2700-a18e-11e9-970b-a796304a1360.png">
<img width="991" alt="Screen Shot 2019-07-08 at 2 37 36 PM" src="https://user-images.githubusercontent.com/32864116/60788030-482f5400-a18e-11e9-8e31-796b2b6b88f6.png">

damithc:
It's fine if an extremely large file cannot be opened but we cannot pre-determine the threshold because it depends on the user's machine. If we warn about the big file and allow the user to open it in a separate tab, if the page fails to load, the user can always kill the tab and go back without losing the entire report?

chan-j-d:
Closed in view of both https://github.com/reposense/RepoSense/pull/1750 and https://github.com/reposense/RepoSense/pull/1647 being merged which puts limits on the file size being analyzed and limits on the number of characters being displayed when viewing the report which were aimed at solving a similar issue in https://github.com/reposense/RepoSense/issues/1504.

PR #1065, opened 2020-02-09
[#202, #1081] Authorship: improve file collapsing mechanism

### LINKED ISSUE #351: Implement progressive loading in code view
as per title, loading a huge change often takes long time. 

_"It's important to deliver something meaningful to the user as soon as possible — the longer they wait for the page to load, the bigger the chance they will leave before waiting for everything to finish."_
Quoted from the Mozilla

### LINKED ISSUE #1081: Authorship view: bug in `hide all file details` button
As raised by @Tejas2805 in #1040, the file detail will be automatically displayed after untick->tick the file type, though it is supposed to be hidden.

Step to reproduce:
1) Click on `hide all file details`
2) Untick `frontend`
3) Tick `frontend`
4) Scroll all the way to the bottom

Expected result: Files should still be hidden due to Step 1
Actual result: Files affected by Step 3 are not hidden

Screen Shot 2020-02-15 at 12 49 15 AM

_Originally posted by @Tejas2805 in https://github.com/reposense/RepoSense/pull/1040#issuecomment-586433653_

ang-zeyu:
Fixes #1081
Alleviates #202

```
Improve file collapsing in authorship view

The authorship tab relies on several dom querying methods to toggle
whether a file is collapsed from view.
This hurts the performance of file collapsing, and detracts from the
use of vue as RepoSense’s front-end framework, decreasing code
readability.

Let's refactor this to use a new variable in the file object instead,
allowing vue reactivity to take over this behaviour.
This also fixes a bug with files not staying collapsed after changing
the file filter [#1081].

With this change, we are able to conveniently display both the expand
and collapse all buttons at the same time. Let's do so as well.

Occasionally, a small number of files have a large number of line
contributions. This slows the initial loading time at best, and causes
the application to crash at worst.
Let's use the new variable to collapse files with more than
2000 lines of code by default, alleviating this.

```

### DIFF
--- a/frontend/src/static/css/v_authorship.scss
+++ b/frontend/src/static/css/v_authorship.scss
@@ -10,6 +10,14 @@
     border-width: 0 1px 1px 1px;
   }
 
+  .toolbar {
+    > a {
+      display: block;
+      margin: 1px 0;
+      text-align: end;
+    }
+  }
+
   .title {
     .contribution {
       color: mui-color('red');
@@ -90,23 +98,17 @@
 
     .file {
       pre {
-        display: none;
+        display: grid;
         // GitHub's font family and font size
         font-family: SFMono-Regular, Consolas, Liberation Mono, Menlo, monospace;
         font-size: 12px;
-      }
+        margin-top: 0;
 
-      &.active {
-        pre {
-          display: grid;
-          margin-top: 0;
-
-          .hljs {
-            // overwrite hljs library
-            display: inherit;
-            padding: 0;
-            white-space: pre-wrap;
-          }
+        .hljs {
+          // overwrite hljs library
+          display: inherit;
+          padding: 0;
+          white-space: pre-wrap;
         }
       }
     }
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -40,16 +40,11 @@ window.vAuthorship = {
       totalBlankLineCount: '',
       filesSortType: 'lineOfCode',
       toReverseSortFiles: true,
-      hasActiveFile: true,
       filterSearch: '*',
     };
   },
 
   watch: {
-    selectedFiles() {
-      setTimeout(this.updateCount, 0);
-    },
-
     filterType() {
       if (this.filterType === 'checkboxes') {
         const searchBar = document.getElementById('search');
@@ -103,19 +98,22 @@ window.vAuthorship = {
       encodeHash();
     },
 
-    expandAll(hasActiveFile) {
-      const renameValue = hasActiveFile ? 'file active' : 'file';
-
-      const files = document.getElementsByClassName('file');
-      Array.from(files).forEach((file) => {
-        file.className = renameValue;
+    expandAll() {
+      this.selectedFiles.forEach((file) => {
+        file.active = true;
+        file.wasCodeLoaded = true;
       });
-
-      this.hasActiveFile = hasActiveFile;
     },
 
-    updateCount() {
-      this.hasActiveFile = document.getElementsByClassName('file active').length > 0;
+    collapseAll() {
+      this.selectedFiles.forEach((file) => {
+        file.active = false;
+      });
+    },
+
+    toggleFileActiveProperty(file) {
+      file.active = !file.active;
+      file.wasCodeLoaded = file.wasCodeLoaded || file.active;
     },
 
     hasCommits(info) {
@@ -163,6 +161,7 @@ window.vAuthorship = {
     },
 
     processFiles(files) {
+      const COLLAPSED_VIEW_LINE_COUNT_THRESHOLD = 2000;
       const res = [];
       const fileTypeBlanksInfoObj = {};
       let totalLineCount = 0;
@@ -175,6 +174,8 @@ window.vAuthorship = {
           const out = {};
           out.path = file.path;
           out.lineCount = lineCnt;
+          out.active = lineCnt <= COLLAPSED_VIEW_LINE_COUNT_THRESHOLD;
+          out.wasCodeLoaded = lineCnt <= COLLAPSED_VIEW_LINE_COUNT_THRESHOLD;
           out.fileType = file.fileType;
 
           const segmentInfo = this.splitSegments(file.lines);
@@ -280,6 +281,10 @@ window.vAuthorship = {
           .sort(this.sortingFunction);
     },
 
+    activeFilesCount() {
+      return this.selectedFiles.filter((file) => file.active).length;
+    },
+
     getFileTypeExistingLinesObj() {
       const numLinesModified = {};
       Object.entries(this.filesLinesObj)
@@ -295,6 +300,7 @@ window.vAuthorship = {
     this.initiate();
     this.setInfoHash();
   },
+
   components: {
     vSegment: window.vSegment,
   },
--- a/frontend/src/tabs/authorship.pug
+++ b/frontend/src/tabs/authorship.pug
@@ -1,8 +1,8 @@
 #authorship
   span.large-font Code Panel
   .toolbar(v-if="hasCommits(info)")
-    a(v-if="hasActiveFile", v-on:click="expandAll(false)") hide all file details
-    a(v-else, v-on:click="expandAll(true)") show all file details
+    a(v-if="activeFilesCount < this.selectedFiles.length", v-on:click="expandAll()") show all file details
+    a(v-if="activeFilesCount > 0", v-on:click="collapseAll()") hide all file details
   .panel-heading
     a.group-name(
       v-bind:href="info.location", target="_blank",
@@ -55,9 +55,9 @@
   .files(v-if="isLoaded")
     .empty(v-if="files.length === 0") nothing to see here :(
     template(v-for="file in selectedFiles")
-      .file.active(v-bind:key="file.path")
+      .file(v-bind:key="file.path")
         .title
-          span.path(onclick="toggleNext(this.parentNode)", v-on:click="updateCount") {{ file.path }}&nbsp;
+          span.path(v-on:click="toggleFileActiveProperty(file)") {{ file.path }}&nbsp;
           span.loc ({{ file.lineCount }} lines)
           a(
             v-bind:href="getFileLink(file, 'commits')", target="_blank",
@@ -69,7 +69,7 @@
             title="click to view the blame view of file"
           )
             i.button.fas.fa-user-edit
-        pre.hljs.file-content
+        pre.hljs.file-content(v-if="file.wasCodeLoaded", v-show="file.active")
           template(v-for="segment in file.segments")
             v-segment(v-bind:segment="segment", v-bind:path="file.path")
 


damithc:
It's a good idea to render/load the code 'above the fold' first rather than load everything upfront. But there is no need to wait for scrolling to load the rest?. The page can continue to load the remaining code while the user is still looking at the top of the page.

ang-zeyu:
> It's a good idea to render/load the code 'above the fold' first rather than load everything upfront. But there is no need to wait for scrolling to load the rest?. The page can continue to load the remaining code while the user is still looking at the top of the page.

I just tried this, but the issue is that some authors have very large, singular contributions ( e.g. test file updates with > 5k loc ). When lazy loading these files, the client still freezes in the meantime.

We may render the collapsed view for such files on first load instead, in addition to progressively rendering more files at a regular interval ( 1s perhaps ). A good threshold for showing the collapsed view first may be 2k loc.

Alternatively, we can further lazy load each file by itself.

Which option should we go with?

damithc:
> I just tried this, but the issue is that some authors have very large, singular contributions ( e.g. test file updates with > 5k loc ). When lazy loading these files, the client still freezes in the meantime.

Perhaps we can ignore these rare cases? The user will be more accepting of slow page loads in such cases. But yes, we can collapse such big files by default, similar to what GitHub does.

ang-zeyu:
> Perhaps we can ignore these rare cases? The user will be more accepting of slow page loads in such cases. But yes, we can collapse such big files by default, similar to what GitHub does.

Noted 👍, I'll go with collapsing big files by default first

ang-zeyu:
Updated, here's the new behaviour:

A single new file will be rendered every 200ms, and only when the previous file has been fully rendered.

In addition, large files ( > 2000 lines ) are initially displayed with the collapsed view, and its code is not rendered into the dom.

------------
Implementation details :

Rendering:

- The `selectedFiles` attribute is watched ( which is changed when opening an authorship view or changing the filter ), which upon changing calls the `displayMoreFiles` method.
- The method updates the an array of files to be displayed `displayedFiles`, and only if the method was not already executed before, and has yet to complete ( e.g. some files have yet to load, but the user changed the filter again ).
- The callback then recursively calls itself once `nextTick` ( for ensuring the previous dom update has finished ) is fufilled and the throttle timeout ( 200ms ) has elapsed.

Collapsed files:

- I needed to rewrite the way collapsed files were shown, since currently, a simple css style of `display: none` is applied over collapsed files. This causes the dom for these files to be rendered still.
  - I removed the `document.getElementsByClassName` approach in favour of an additional variable `file.active`, and rewrote the "toggle collapse" functions to trigger an additional cache flag `file.wasLoaded` .
  - Removed some old methods as a result ( `getSelectedFiles` - which did not have any usages before either, and `updateCount` )

------------
Some decisions that need more reviewing:
- The value of the hardcoded throttle duration - is 200ms too big / small?
- Whether we should lazily load large files ( which are likely to be non-functional though ) over collapsing them
- :grey_question:

ang-zeyu:
Ready for review.

The new behaviour:

A single new file will be rendered every 300ms, and only when the previous file has been fully rendered.

In addition, large files ( > 2000 lines ) are initially displayed with the collapsed view, and its code is not rendered into the dom until the user expands it.

Clicking `show all file details` or expanding a single file is still synchronous however.

------------

For Collapsed files:

I needed to rewrite the way collapsed files were shown, since currently, only a css style of `display: none` is applied over collapsed files. This causes the dom for these files to be rendered still.
- I removed the `document.getElementsByClassName` approach in favour of an additional variable `file.active`, and rewrote the "toggle collapse" functions to trigger an additional cache flag `file.wasCodeLoaded`.
- I also removed some unnecessary methods as a result ( `getSelectedFiles` - which did not have any usages before either, and `updateCount` )

damithc:
Nice work. 
When I check the Netlify preview for this PR, the first loading of the code view takes quite long and there is no `loading...` indicator. The subsequent loadings (for other authors) are quite fast though. I wonder if that's a netlify-specific problem or something to do with the change in this PR.

Also, should we keep the `loading...` indicator until the first file shows up?

ang-zeyu:
> When I check the Netlify preview for this PR, the first loading of the code view takes quite long and there is no `loading...` indicator. The subsequent loadings (for other authors) are quite fast though. I wonder if that's a netlify-specific problem or something to do with the change in this PR.

It may be the large amount of animated spinners ( one per file ), or some other things.
I'll try to micro optimize it!

> Also, should we keep the `loading...` indicator until the first file shows up?

👍

ang-zeyu:
Updated,
Its about 1s ( 7 down to 6s ) faster to first interaction for a 2000 line author in the netlify preview, and should scale the more lines there are.
The previous timings was simply due to showing 1 animated spinner per file, which was much more expensive than I expected.

frontend/src/tabs/authorship.pug
```diff
@@ -70,8 +70,8 @@
           title="click to view the blame view of file"
           )
             i.button.fas.fa-user-edit
-        pre.hljs
+        pre.hljs(v-if="doLoadFile(file)" v-show="file.active")
```
0blivious: Why v-show and v-if together? Also, the attribute `wasCodeLoaded` seems not useful.

ang-zeyu: Hi!
`v-if` is to prevent vue from unloading it from the dom ( caching it essentially ) once it has been loaded for the first time.
In here, `doLoadFile` returns `wasCodeLoaded || file.active`, and `wasCodeLoaded` is set to true and remains permenantly so once the file is set to `active` for the first time, and has been loaded 
( either by `displayMoreFiles` or the user re-expanding a by-default collapsed file )

`v-show` is just to toggle `display: none` on/off on it

0blivious: I understand. But isn't this what v-show does? It make the elements remain in the DOM and are only hidden by CSS.

ang-zeyu: We don't want all of the code lines' DOM to be loaded first ( rendered ) though, which is the purpose of this PR, hence we also need `v-if` to control the loading.

0blivious: Just condition on `wasCodeLoaded` then. Will there be any cases that a file is active but not loaded?

ang-zeyu: Yup, when the authorship view is initially loaded, `wasCodeLoaded` is false for all files.
Files with < 2000 loc are also not `active` in addition as discussed here https://github.com/reposense/RepoSense/pull/1065#issuecomment-583850317

`wasCodeLoaded` is toggled bit by bit for each file, which loads the page asynchronously.
It's also switched on when the user expands files with >= 2000 loc ( by the expand all button or toggling the file individually )

frontend/src/static/js/v_authorship.js
```diff
@@ -32,6 +32,8 @@ window.vAuthorship = {
     return {
       isLoaded: false,
       files: [],
+      isLoadingMoreFiles: false,
```
jamessspanggg: one question: is `isLoadingMoreFiles` the inverse of `hasAllLoaded`?

frontend/src/static/js/v_authorship.js
```diff
@@ -171,6 +181,9 @@ window.vAuthorship = {
           const out = {};
           out.path = file.path;
           out.lineCount = lineCnt;
+          out.active = lineCnt <= COLLAPSED_VIEW_LINE_COUNT_THRESHOLD;
+          out.wasCodeLoaded = false;
+          out.isLoading = true;
```
jamessspanggg: Perhaps `shouldLoad` / `shouldRender` may be a better variable name instead of `isLoading`?

frontend/src/static/js/v_authorship.js
ang-zeyu: yup, except `hasAllLoaded` is only toggled once to avoid triggering too many dom updates from `isLoadingMoreFiles`

frontend/src/tabs/authorship.pug
ang-zeyu: @0blivious I see your point now, thanks!

Just to clarify again, you meant that only the `||` in `file.wasCodeLoaded || file.active` is unecessary right?
ie. It should just be `file.wasCodeLoaded`

ang-zeyu:
Realised some additional things: 
1. If the user resorts the files before the loading has finished, the loading spinner would still be at the bottom, but files would pop in throughout since the loading order follows the new sorted order. Should we show a small `loading...` text in place of a spinner per file instead? ( as was before, but text should be less expensive ) ( there's also an added benefit of being able to remove one or two state variables )

2. For authors with a small number of total loc, it would be tiresome to wait for all the files to load. ( though the user can click expand all button to do so ) We could restrict progressive loading to files with a minimal amount of lines of code as well ( 50 maybe ), which should give a good balance of TTI and time spent waiting for files to be lazily loaded

what do you guys think? @jamessspanggg @0blivious @damithc

damithc:
> what do you guys think? @jamessspanggg @0blivious @damithc

No objections from me.

frontend/src/tabs/authorship.pug
0blivious: Yes, you can do sth like v-if="file.wasCodeLoaded"

frontend/src/tabs/authorship.pug
```diff
@@ -1,8 +1,8 @@
 #authorship
   span.large-font Code Panel
   .toolbar(v-if="hasCommits(info)")
-    a(v-if="activeFilesCount === 0", v-on:click="expandAll(true)") show all file details
-    a(v-else, v-on:click="expandAll(false)") hide all file details
+    a(v-if="activeFilesCount < this.selectedFiles.length", v-on:click="expandAll()") show all file details
+    a(v-else, v-on:click="collapseAll()") hide all file details
```
0blivious: Since we are using vue, is it better we just watch the activeFilesCount instead of implementing 2 click functions? Also, why revamp the previous logic for button change? I feel display `hide all` as long as there is at least one active file would be better since the first thing I do when open the authorship view is usually to hide all file details.

frontend/src/tabs/authorship.pug
0blivious: Yes, sth like v-if="file.wasCodeLoaded"

0blivious:
> what do you guys think? @jamessspanggg @0blivious @damithc

I agree, would prefer a single line of text instead of animation.

Also, can you link this PR to #1081 in your commit message?

frontend/src/tabs/authorship.pug
ang-zeyu: > Since we are using vue, is it better we just watch the activeFilesCount instead of implementing 2 click functions?

Hmm I don't follow, isn't vue already watching `activeFilesCount` implicitly and showing/hiding the `a(...)`'s as needed with the `v-if`'s?


> Also, why revamp the previous logic for button change? I feel display hide all as long as there is at least one active file would be better since the first thing I do when open the authorship view is usually to hide all file details.

Fixed :+1:. Only meant to change the condition for the first `a(...)`

jamessspanggg:
> what do you guys think? @jamessspanggg @0blivious @damithc

Sounds good to me. I prefer the `loading...` text too

ang-zeyu:
Updated, removed `isLoading` and `hasLoadedAll` flags.

For #202,
I tried loading the 400k loc commit as well, it works initially, but still crashes once you expand the 200k loc file. 
Will leave it at "alleviate", since we can at least now view the smaller files though.

frontend/src/static/js/v_authorship.js
```diff
@@ -99,19 +102,24 @@ window.vAuthorship = {
...
-      this.activeFilesCount = isActive ? this.selectedFiles.length : 0;
+    collapseAll() {
+      this.selectedFiles.forEach((file) => {
+        file.active = false;
+      });
     },
 
-    updateCount() {
-      this.activeFilesCount = document.getElementsByClassName('file active').length;
+    toggleCollapsed(file) {
```
jamessspanggg: perhaps `toggleFile` would be a better method name? as the file may not be collapsed

frontend/src/static/js/v_authorship.js
```diff
@@ -319,6 +351,7 @@ window.vAuthorship = {
     this.initiate();
     this.setInfoHash();
   },
+
```
jamessspanggg: maybe the new line is not necessary

frontend/src/static/css/v_authorship.scss
```diff
@@ -2,6 +2,14 @@
 
 /* Authorship */
 #tab-authorship {
+  .toolbar {
+    > a {
```
jamessspanggg: may I ask why this CSS change is needed?

ang-zeyu: for `show all file details` and `hide all file details` to show up on top / below of each other.
I changed the condition for `show all file details` to `activeFilesCount > 0` from `activeFilesCount === 0`, which can be helpful since we now collapse files with > 2000 loc by default.

frontend/src/static/js/v_authorship.js
ang-zeyu: Would `toggleFileActive` be fine?

frontend/src/static/css/v_authorship.scss
jamessspanggg: ahhh okay, @damithc are u ok with the following, where if there's at least one file that is collapsed and at least one file that is showed, then the `show all file details` and `hide all file details` options will appear.

Screen Shot 2020-02-20 at 11 23 46 PM

frontend/src/static/js/v_authorship.js
jamessspanggg: how about `toggleFileActiveProperty`?

frontend/src/static/js/v_authorship.js
ang-zeyu: Hmm, most of the "sections" are currently seperated by newlines though.
e.g. between `watch` and `methods` and `components`.
Not sure if github is displaying this correctly, I wouldn't mind changing it otherwise though

frontend/src/static/js/v_authorship.js
ang-zeyu: Sounds good :+1:

frontend/src/static/css/v_authorship.scss
damithc: Yes I'm OK with it.

frontend/src/static/js/v_authorship.js
jamessspanggg: Alright, then it is fine

ang-zeyu:
Updated `toggleCollapsed` naming

0blivious:
@ang-zeyu can fix the conflicts?

ang-zeyu:
> @ang-zeyu can fix the conflicts?

:+1: Updated

ang-zeyu:
Resolved minor conflict with #1074

ang-zeyu:
Resolved some more conflicts with merged PRs

ang-zeyu:
> there seems to be some memory leak and heavier memory usage compare to master branch

Thanks for testing this!

What was the method used to produce the leak in your example?
I tried switching between authorship tabs and different types of tabs, but wasn't able to produce it exclusively on this PR ( 'master' has it as well ).

I've updated this with a fix for another potential memory leak though ( clearing the timeout ), but the leak still occurs with it.

( PR preview of [1169](https://deploy-preview-1169--reposense.netlify.com/) after switching through some amount of authorship tabs )
memoryleak

Memory usage wise, I looked at the usage after entering the page and opening an author with ~2000 loc. In both cases the usages was within error of margin ( ~5 mb ); Could the increased observed usage previously be due to the memory leak?

eugenepeh:
>What was the method used to produce the leak in your example?

Memory leak observed was from constantly switching between the authorship view of different authors. 

Although there are some leakage from master branch, it wasn't as severe as the PR.
After some times of repetition, the PR's memory usage doubled compare to master branch.

This has happened before in the past where hljs were used inadequately; perhaps you can look into that area. 

Meanwhile, can other reviewers confirm this as well? In case I am the only one who experience degradation.

ang-zeyu:
> Memory leak observed was from constantly switching between the authorship view of different authors.
> 
> Although there are some leakage from master branch, it wasn't as severe as the PR.
> After some times of repetition, the PR's memory usage doubled compare to master branch.

Ah I see. It might be due to the potential memory fix I just updated this with. My timings between switching tabs weren't perfectly consistent before as well, which may have led to different results from yours.
I'm actually getting consistently lower heap usage on this netlify preview versus master now otherwise, which makes sense since we aren't rendering everything at once.

@jamessspanggg @0blivious 
would appreciate if you could help double check if the leak in this pr is more serious than master!

0blivious:
For my case it is a bit more serious than master, also the JS heap usage tend to be slightly higher (20MB more).

ang-zeyu:
> For my case it is a bit more serious than master, also the JS heap usage tend to be slightly higher (20MB more).

Thanks for testing it out!
I can now reproduce it strangely, and its actually + ~30% usage after a while, which seems rather significant.

I have a solution in mind, being to move the timer logic to the main application ( since it stays in memory all the time, it should prevent leaks from this ), but I think this is something better done after projectifying the frontend with better global state management.

In the meantime, I could repurpose this PR to preserve these behaviours
- refactoring of the `document.getElementsByClassName('...')` method of toggling file collapse to vue data variables to decrease reliance on dom querying
- showing both the `show all files` and `hide all files` button appropriately

Would these changes be fine on its own? @jamessspanggg @0blivious

frontend/src/static/js/v_authorship.js
```diff
@@ -40,14 +40,13 @@ window.vAuthorship = {
...
       toReverseSortFiles: true,
-      hasActiveFile: true,
       filterSearch: '*',
     };
   },
 
   watch: {
     selectedFiles() {
-      setTimeout(this.updateCount, 0);
+      this.displayMoreFiles();
```
0blivious, line 49: unresolved function

ang-zeyu: removed

ang-zeyu: thanks for re-reviewing this as well!

I suppose the changes mentioned above are fine on their own?
I've updated the commit message, title and linked issues to reflect, let me know otherwise!

0blivious: yes I feel the current implementation are fine

ang-zeyu:
@jamessspanggg could I get your review here again? 

The lazy rendering parts are removed and only the following parts are preserved:
- refactoring of the `document.getElementsByClassName('...')` method of toggling file collapse to vue data variables to decrease reliance on dom querying
- showing both the `show all files` and `hide all files` button appropriately
- collapsing files > 2000 loc by default ( alleviating #202 )

As a sidenote my all my PRs seem to be passing on travis currently https://travis-ci.org/github/reposense/RepoSense/pull_requests?utm_medium=notification&utm_source=github_status but the status isn't getting reported back to github o_o

jamessspanggg:
> As a sidenote my all my PRs seem to be passing on travis currently https://travis-ci.org/github/reposense/RepoSense/pull_requests?utm_medium=notification&utm_source=github_status but the status isn't getting reported back to github o_o

There are times where it happens to my PRs as well. Not sure what causes it :/

ang-zeyu:
Thanks for the re-review!

> There are times where it happens to my PRs as well. Not sure what causes it :/

Ahh ok, will try triggering a rebuild tomorrow again. ( been triggering for two days already tho =/ )

ang-zeyu:
> Clean fix. Can you also explore to apply the same fix to all other methods that use `toggleNext` function?

👍, seems like only `zoom.pug` is still using it. Should it be in a separate PR?

fzdy1914:
> > Clean fix. Can you also explore to apply the same fix to all other methods that use `toggleNext` function?
> 
> 👍, seems like only `zoom.pug` is still using it. Should it be in a separate PR?

Yes, please do it in a separate PR.

PR #1066, opened 2020-02-10
[#1060] Report: Record time taken for report generation in hidden element

### LINKED ISSUE #1060: Add time taken for analysis in report
I think we should add a hidden time taken value in the report for our own benchmark purposes.
This allows us to do a rough gauge on the RepoSense's performance from e.g. automated nightly build of dashboard demo or CS2113 dashboard etc.

anubh-v:
Fixes #1060 
```
Report: Add property to track time taken for report generation

It is sometimes not possible / inconvenient to retrieve the
RepoSense log output, if we want to view the time taken for
generating a report.

Let's record the time taken for report generation, as a hidden
element in the report as well as a field in the summary.json file.
This will allow developers to find out the amount of time taken
in generating the report, from the browser's javascript console.
```


src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -81,34 +81,40 @@
...
+    }
 
-        List<Path> reportFoldersAndFiles = cloneAndAnalyzeRepos(configs, outputPath);
-
+    /**
+     * Generates the summary JSON file for all repos in {@code configs} at {@code outputPath}.
+     *
+     * @return the path of the generated file.
+     */
+    public static List<Path> generateSummary(List<RepoConfiguration> configs, String outputPath,
```
anubh-v: I split the original `generateReposReport` method into 2 methods:

- `generateReposReport` : creates the `authorship.json` and `commits.json` file for each repo
- `generateSummary`: creates the single `summary.json` file.

I broke up `generateReposReport` to make it easier to test the functionality in this PR.
By having `generateSummary` as a separate method, the test code can pass an arbitrary value of `reportGenerationTime`.

anubh-v:
image

@eugenepeh is this close to what you meant?

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -17,23 +16,31 @@ public static void startTimer() {
...
+    public static String getElapsedTime() {
         long endTime = System.nanoTime();
         double elapsedTime = (double) (endTime - startTime) / 1_000_000_000.0;
         int elapsedHours = (int) elapsedTime / 3600;
         int elapsedMinutes = (int) (elapsedTime % 3600) / 60;
         double elapsedSeconds = elapsedTime % 60;
+        String formattedElapsedTime = "";
 
         if (elapsedHours > 0) {
-            elapsedTimeMessage += String.format(" %d hour(s)", elapsedHours);
```
eugenepeh: why are the spaces removed as well?

src/main/java/reposense/RepoSense.java
```diff
@@ -65,11 +65,17 @@ public static void main(String[] args) {
             RepoConfiguration.setDatesToRepoConfigs(configs, cliArguments.getSinceDate(), cliArguments.getUntilDate());
             RepoConfiguration.setStandaloneConfigIgnoredToRepoConfigs(configs,
                     cliArguments.isStandaloneConfigIgnored());
+
             List<Path> reportFoldersAndFiles = ReportGenerator.generateReposReport(configs,
+                    cliArguments.getOutputFilePath().toAbsolutePath().toString());
+
+            reportFoldersAndFiles.addAll(ReportGenerator.generateSummary(configs,
```
eugenepeh: doesn't quite make sense to split it into 2; this should be the underlying structure.
Think of it from another angle, shouldn't the `generateReposReport` comes with a summary?

src/main/java/reposense/util/TimeUtil.java
anubh-v: My apologies. I mistakenly thought the original code had some redundant spaces.
I have reverted this to the original style.

src/main/java/reposense/RepoSense.java
anubh-v: Thanks for the feedback. I have reverted to the original `generateReposReport` method.
My original reason for splitting this method is to allow the test code to provide its own report generation time.

However, this can be done by accepting a supplier for the report generation time, as done in the current solution.

anubh-v:
Current solution:
Time taken for report generation is stored in a hidden element

Report generation time = Time to parse user input / configs + cloning / analysing repos + generate report files (excludes time spent zipping the report files)

image

anubh-v:
> The elapsed time shown in the logger is different from that in the generated webpage. Would that be a source of confusion?

Thanks for the comment.
The logger's time includes the time taken to zip the report files - that's why it is different.
I think this is ok, given we are aware of the difference.

Including the zipping time in `summary.json` is not so straightforward because we would have to edit `summary.json` after the report files have been compressed.

@fzdy1914 would it be possible for the zip folder generation time to contribute significantly to overall execution time? In that case, should we explore ways of including the zip folder generation time in the report?

frontend/src/index.pug
```diff
@@ -50,6 +50,8 @@ html
             )
               strong User Guide - {{ repoSenseVersion }}
             span ) on {{ creationDate }}
+          .report-generation-time(style="visibility:hidden;font-size: 0;")
```
jamessspanggg: perhaps only `display: none`?

ccyccyccy:
> Since it the zip folder generation time does not contribute much to overall execution, I think it would be ok not to record it in the hidden element.

Then maybe we can use the same timing for both of them? i.e. don't factor in zipping for both of them.

anubh-v:
>  Then maybe we can use the same timing for both of them? i.e. don't factor in zipping for both of them.

It might still be useful to record the zip folder generation time in the log output, for debugging purposes.

In terms of performance measurement, this should not be a major source of confusion as the timing difference is small (~1 to 2 seconds).

@fzdy1914 may I clarify this with you? Would it be better not to record the zip folder generation time in both the logs and the hidden report element? Or should we keep the current method where it is only recorded in the logs?

eugenepeh:
Are both using the same elapsed function? If it is then its fine to retain both.
There shouldn't be any confusion since this report value is hidden and meant for development use only.

anubh-v:
> Are both using the same elapsed function?

Yes, the time measurement for both the hidden element and the logger uses `System.nanoTime()`

frontend/src/index.pug
anubh-v: Thanks for looking through! I agree; `display: none` is simpler and achieves the same purpose.

src/main/java/reposense/RepoSense.java
```diff
@@ -70,7 +70,8 @@ public static void main(String[] args) {
                     cliArguments.getOutputFilePath().toAbsolutePath().toString(),
                     formatter.format(ZonedDateTime.now(cliArguments.getZoneId())),
                     cliArguments.getSinceDate(), cliArguments.getUntilDate(),
-                    cliArguments.isSinceDateProvided(), cliArguments.isUntilDateProvided());
+                    cliArguments.isSinceDateProvided(), cliArguments.isUntilDateProvided(),
+                    TimeUtil::getElapsedTime);
```
fzdy1914: Can I know why this is necessary, why can't we call this function in `ReportGenerator::generateReposReport`? It seems there is no possibility we will use other time functions in the same place?

anubh-v: I chose to pass `TimeUtil::getElapsedTime` as an argument so that the system tests can provide their own timing rather than using `TimeUtil::getElapsedTime`.

In the system tests, I am passing in a fixed elapsed time to `generateReposReport` instead of using `TimeUtil::getElapsedTime`

fzdy1914: Then, that's fine if there is no other solution.

anubh-v: There is another possible solution.

We can change the API of `ReportGenerator` slightly.
Currently, we are calling `generateReposReport` from both `Reposense.java` and `ConfigSystemTest`.

We could break `generateReposReport` into 2 public methods: 

- `generateCommitsAndAuthorshipReport` : this will contain code for generating `commits.json` and `authorship.json`
-  `generateSummary`: this will contain code for generating `summary.json`.

We **will still have** the `generateReposReport` method, but it will simply call `generateCommitsAndAuthorshipReport` first, followed by `generateSummary`.

The elapsed time will only be an argument of `generateSummary` and **not** an argument of `generateReposReport`.

In `RepoSense.java`, we can continue to call `generateReposReport` as per normal.
In `ConfigSystemTest`, we will not call `generateReposReport`.
We will call `generateCommitsAndAuthorshipReport` and `generateSummary`  instead, so that we can pass in a fixed elapsed time.

anubh-v: @fzdy1914 should I refactor the solution as such ^ ?

fzdy1914: I think the current approach is fine. Anyway, the system test is meant to test the system as a whole. There is no reason to test only a part of the code just for this small issue.

anubh-v:
@fzdy1914 can this be merged?

PR #1067, opened 2020-02-10
[#823, #981] AnnotatorAnalyzer: Fix code disowning mechanism

### LINKED ISSUE #823: AnnotatorAnalyzer: Bug in code disowning mechanism

### LINKED ISSUE #981: Inaccurate Testing of AnnotatorAnalyzerTest
As per my conversation with @eugenepeh in #789,  

`AnnotatorAnalyzerTest` is meant to test the functionality of the author tag - `//@@author`. However, the test currently analyses `blameTest.java` within `testrepo-Alpha`, which does not contain any author tags: https://github.com/reposense/testrepo-Alpha/blob/136c6713fc00cfe79a1598e8ce83c6ef3b878660/blameTest.java#L1-L4

Hence, both tests in `AnnotatorAnalyzerTest` are unable to test the functionality of the author tag and yields the same output.

Code in `AnnotatorAnalyzerTest`: 
https://github.com/reposense/RepoSense/blob/9f10e7aff83b620595976c25a828be15e9863ece/src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java#L9-L27

### LINKED ISSUE #1115: Investigate case of @@author tag not being honored
Shouldn't this code block be skipped as it is attributed to a different author?
image

https://nus-cs2113-ay1920s2.github.io/tp-dashboard/#search=&sort=groupTitle&sortWithin=title&since=2020-03-01&timeframe=commit&mergegroup=false&groupSelect=groupByRepos&breakdown=false&tabOpen=true&tabType=authorship&tabAuthor=jinfayap&tabRepo=AY1920S2-CS2113T-M16-2%2Ftp%5Bmaster%5D

configs https://github.com/nus-cs2113-AY1920S2/tp-dashboard

anubh-v:
Fixes #823
Fixes #1115 
Fixes #981 too

```
AnnotatorAnalyzer: Fix code disowning mechanism

Suppose that "random-user" is not the git author name of any author
in a git repository (i.e. it is just a placeholder name).

RepoSense does not correctly mark code as disowned when a user marks
a code block with the "@@author random-user" tag.
Only the line containing the "@@author random-user" tag is treated
as disowned, while the remaining code block is still attributed to
the author deduced from the git blame analysis.

This bug occurs because the code in AnnotatorAnalyzer uses
a constant UNKNOWN_AUTHOR to represent both the case where
there is no currently active annotation override AND the case
where the user has used a "@@author random-user" tag.

Let's use Optionals to store the author names obtained from the
@@author tags, so we can identify the case where user has used
a "@@author random-user" tag.

```


anubh-v:
@fzdy1914, may I get your thoughts on this? Thanks very much for your time and effort

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
```diff
@@ -23,43 +24,45 @@
...
-                Author newAuthor = findAuthorInLine(lineInfo.getContent(), authorAliasMap);
+                Optional<Author> newAnnotatedAuthor = findAuthorInLine(lineInfo.getContent(), authorAliasMap);
 
-                if (newAuthor.equals(Author.UNKNOWN_AUTHOR)) {
-                    //end of an author tag should belong to this author too.
-                    lineInfo.setAuthor(currentAuthor);
-                } else if (newAuthor.getIgnoreGlobMatcher().matches(filePath)) {
-                    newAuthor = Author.UNKNOWN_AUTHOR;
+                if (!newAnnotatedAuthor.isPresent()) {
+                    // end of an author tag should belong to the current author too.
```
fzdy1914: The comment seems can be improved. Other LGTM

anubh-v: I've updated the comment.

anubh-v:
Here's a breakdown of the logic used in processing the `@@author` tags.

- If we encounter a start tag (i.e. `@@author GIT_AUTHOR_NAME`), then:
  - If `GIT_AUTHOR_NAME` belongs to an author within the `AuthorConfiguration`, then `AnnotatorAnalyzer` will attribute the subsequent lines to this author (until an end tag `@@author` is encountered, or till end of file if no end tag is encountered).

  - If `GIT_AUTHOR_NAME` does not belong to any author within `AuthorConfiguration`, then `AnnotatorAnalyzer` will attribute the subsequent lines to a `UNKNOWN_AUTHOR` (until an end tag is encountered, or till end of file if not end tag is encountered)

- If we encounter an end tag (i.e. `@@author`), then:
  - If there has been a start tag (i.e. `@@author GIT_AUTHOR_NAME`) encountered previously, the line with the end tag will be attributed to author inferred from that start tag.
  - If no start tag has been encountered yet, then the line with the end tag is attributed to `UNKNOWN_AUTHOR`.

anubh-v:
@fzdy1914 may I get your help to activate Codacy analysis for the `1035_feature_branch` ?
image

Thanks very much for your time

fzdy1914:
Sorry, I also do not have the permission.

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
```diff
@@ -23,43 +24,46 @@
...
     }
 
     /**
      * Extracts the author name from the given {@code line}, finds the corresponding {@code Author}
-     * in {@code authorAliasMap},and returns the result.
-     * @return {@code Author#UNKNOWN_AUTHOR} if no matching {@code Author} is found.
+     * in {@code authorAliasMap}, and returns this {@code Author} stored in an {@code Optional}.
+     * @return {@code Optional.of(Author#UNKNOWN_AUTHOR)} if no matching {@code Author} is found,
+     *         {@code Optional.empty()} if an end author tag is used (i.e. "@@author"),
+     *         and {@code Optional.empty()} if the extracted author name is too short.
```
jylee-git: Why return `Optional.empty()` if the extracted author name is too short?

jylee-git:
Btw @anubh-v , I suggest that you actively merge updates from the main branch in if you are using this feature-branch method. You may risk having a mountain of merge conflicts to deal with if you don't; now that there are no indicators in GitHub that there's updates to the code base.

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
anubh-v: **In the original codebase**, the extracted author name is validated against this regex: 

If the extracted author name does not match this regex, then the author is set to `UNKNOWN_AUTHOR` and the annotation override is not applied to the code block.  This regex ensures the author name has at least 2 consecutive alphanumeric characters.

In order to keep this behaviour, I chose to return `Optional.empty()` when the extracted author name fails the regex check.

I am returning `Optional.of(UNKNOWN_AUTHOR)` in cases where we detect the author wants to disown the code block (by using a `@@author random-user` tag)

jylee-git: Hmmm... Regarding the regex, I suppose the main intention of the founding developer who adopted this regex is to conform to the github's GitHub ID.

FYI, GitHub ID requirement is:
* Max 39 characters long
* Username must me alphanumeric and can contain up to a single hyphen, but username cannot begin or end with a hyphen

However the thing is, authors can use any type of character that they want when they commit in git, such as this
image

Since we are relying on Git author name to attribute code authorship, I think it's better if we change/remove the author regex to suit this current situation. You might want to create an issue on this and/or put this on hold for now.

If you want a better understanding on this, try this:

* use an author tag with an author name that doesn't conform to github id convention (e.g.: `-my author name-`)
* Run the program

anubh-v: Thanks for the thoughts @jylee-git 
Yes, we could update our regex to be more flexible (allow things like `@@author --test--`)
I'll open up an issue on this.

What about this PR itself? Do you mean this PR should be placed on hold?
What are your thoughts on returning `Optional.empty()` for cases where the regex validation fails?

jylee-git: I think you can continue fixing this bug.

One more thing to ask: is the current intended solution, when a person does not want to attribute the code to anyone, to use an author name that hasn't been used to commit the code before?
> e.g.: the git id `some-randomUser` did not make any commits to the code before. In order to **not** attribute the code to anyone, the person will use `//@@author some-randomUser` to avoid attributing the code below to anyone.

anubh-v: Yes, that is the intended solution (as mentioned by Prof @damithc [here](https://github.com/reposense/RepoSense/issues/823#issuecomment-513143467))

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
```diff
@@ -23,43 +24,45 @@
...
+    private static Optional<Author> findAuthorInLine(String line, Map<String, Author> authorAliasMap) {
         try {
             String[] split = line.split(AUTHOR_TAG);
             String name = extractAuthorName(split[1]);
             if (name == null) {
-                return Author.UNKNOWN_AUTHOR;
+                return Optional.empty();
             }
-            return authorAliasMap.getOrDefault(name, Author.UNKNOWN_AUTHOR);
+            return Optional.of(authorAliasMap.getOrDefault(name, Author.UNKNOWN_AUTHOR));
```
anubh-v: @jylee-git, @fzdy1914, I realised there's another aspect that will be good to discuss now.

When the `AUTHOR_NAME` in `@@author AUTHOR_NAME` is not provided in our `AuthorConfiguration`, `AnnotatorAnalyzer` will attribute the subsequent code block to `UNKNOWN_AUTHOR`.

This fixes the *current* bug in the disowning mechanism. Users will be able to disown code.

However, as mentioned in #1082, *in future* we would like to have a ramp chart for the "virtual authors" who have been credited with all the disowned code. This will allow us to view all the disowned code via the `Code Panel` button.

It is currently undecided whether we want a single ramp chart for all the virtual authors (`some-randomuser1`, 
`some-randomuser2`, etc) or separate ramp charts for each virtual author.  Prof mentioned that we can adopt the solution that requires less development effort.

Given our future goals, perhaps I should adjust this PR to prevent alot of rework in future?
If we want a **separate** ramp chart for each virtual author, then we cannot attribute all the disowned code blocks to `UNKNOWN_AUTHOR`. We would need to create an `Author` instance for each virtual author
(e.g. `some-randomuser1`, `some-randomuser2`, etc).

anubh-v: My view is that we should create a single ramp chart for all virtual authors, as this should involve less development work.

anubh-v:
> I suggest that you actively merge updates from the main branch in if you are using this feature-branch method.

@jylee-git, thanks for the feedback.

I have changed the target branch of this PR to be `master`. I will no longer be using the feature branch approach. I realised that approach will increase the effort needed from reviewers as they have to review related work multiple times. A single PR with all changes consolidated would be more efficient.

anubh-v:
As discussed in #1082, we would like to have a ramp chart for each virtual author. This means the implementation of this bugfix will also need to change as we can no longer attribute all the disowned code to `UNKNOWN_AUTHOR`

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
anubh-v: As discussed in #1082, it would be preferred to have separate ramp charts for virtual authors. I'll hence be tweaking the implementation of this bugfix.

jylee-git:
> As discussed in #1082, we would like to have a ramp chart for each virtual author. This means the implementation of this bugfix will also need to change as we can no longer attribute all the disowned code to `UNKNOWN_AUTHOR`

Hmm, I think this can be dealt with in another PR.

Meanwhile, let's get the code disowning mechanism fixed ASAP (see #1115)

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
```diff
@@ -23,43 +24,45 @@
...
-                if (newAuthor.equals(Author.UNKNOWN_AUTHOR)) {
-                    //end of an author tag should belong to this author too.
-                    lineInfo.setAuthor(currentAuthor);
-                } else if (newAuthor.getIgnoreGlobMatcher().matches(filePath)) {
-                    newAuthor = Author.UNKNOWN_AUTHOR;
+                if (!newAnnotatedAuthor.isPresent()) {
+                    // end of an author tag should belong to the current author too.
+                    // if an end author tag was used without a corresponding starting tag, attribute the
+                    // line to UNKNOWN_AUTHOR
+                    lineInfo.setAuthor(currentAnnotatedAuthor.orElse(Author.UNKNOWN_AUTHOR));
```
jylee-git: I don't quite get this particular comment; can you give an example for this part?

anubh-v:
> Hmm, I think this can be dealt with in another PR.

Ok, I will proceed to update the system tests then, without making further changes for #1082

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
anubh-v: The `orElse` caters for cases where the user used an end author tag (`@@author`) without having any preceding start author tag (e.g. `@@author tom`). This is most likely because the user forgot to provide a start author tag.

In such cases, we will attribute the line with the end author tag to `UNKNOWN_AUTHOR`

This logic is consistent with the current implementation:
https://github.com/reposense/RepoSense/blob/9725a25d4e34164cefbb51f7f6e39964cb0a521a/src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java#L25-L34

src/systemtest/resources/author-config.csv
```diff
@@ -1,5 +1,5 @@
 Repository's Location,Branch,Author's GitHub ID,Author's Email,Author's Display Name,Author's Git Author Name,Ignore Glob List
-https://github.com/reposense/testrepo-Beta.git,,nbr,neilbrian.nl@gmail.com,"Ashwin Mukadaan Singhania, Rathod",,
+https://github.com/reposense/testrepo-Beta.git,,nbr,neilbrian.nl@gmail.com,"Ashwin Mukadaan Singhania, Rathod",nbriannl,
```
anubh-v: @jylee-git I updated the `author-config.csv` used for the system tests.
Reason: Within `testRepo-Beta`, some files contain author tags within the name `nbriannl` (i.e. `@@author nbriannl`).
However, none of the authors in `author-config.csv` have the alias `nbriannl`.

From an inspection of the standalone config in testRepo-Beta ([here](https://github.com/reposense/testrepo-Beta/blob/master/_reposense/config.json)), it seems that the `nbriannl` alias belongs to the author whose github id is `nbr` in the `author-config`.

I updated the aliases list for the author `nbr` in `author-config`, so that `AnnotatorAnalyzer` will attribute code to him / her in the system tests.

Tejas2805: Alright, makes sense.

anubh-v:
In addition to the system tests, we can check that the bug in #1115 is fixed:

Current RepoSense release:
image


With this PR, the code block is disowned:
image

To create this report locally, checkout this PR and run RepoSense with:
[author-config.txt](https://github.com/reposense/RepoSense/files/4296670/author-config.txt)
[repo-config.txt](https://github.com/reposense/RepoSense/files/4296671/repo-config.txt)

(I had to upload *txt* files instead of *csv*, as GitHub does not allow *csv* file uploads.)

Use the command:
```
./gradlew run -Dargs="--config [PATH]--since 01/03/2020 --formats java md fxml sh bat gradle txt --timezone UTC+08"
```

(Command is taken from [CS2113 dashboard script](https://github.com/nus-cs2113-AY1920S2/tp-dashboard/blob/master/run.sh))

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
jylee-git: I see. Thanks

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
```diff
@@ -23,43 +24,45 @@
...
 
     /**
      * Extracts the author name from the given {@code line}, finds the corresponding {@code Author}
-     * in {@code authorAliasMap},and returns the result.
-     * @return {@code Author#UNKNOWN_AUTHOR} if no matching {@code Author} is found.
+     * in {@code authorAliasMap}, and returns this {@code Author} stored in an {@code Optional}.
+     * @return {@code Optional.of(Author#UNKNOWN_AUTHOR)} if no matching {@code Author} is found,
+     *         {@code Optional.empty()} if an end author tag is used (i.e. "@@author"),
+     *         and {@code Optional.empty()} if the extracted author name is too short.
      */
```
jylee-git:
```suggestion
     * @return {@code Optional.of(Author#UNKNOWN_AUTHOR)} if no matching {@code Author} is found,
     *         {@code Optional.empty()} if an end author tag is used (i.e. "@@author"), or the extracted author name is too short.
     */
```

src/systemtest/resources/author-config.csv
jylee-git: I don't think this is in the scope of this PR.

Since in the `repo-config.csv` of the systemtest, notice that the standalone config file is ignored. So i'm of the opinion that we should leave it as it isl

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -0,0 +1,84 @@
...
+import reposense.authorship.model.FileInfo;
+import reposense.authorship.model.FileResult;
+import reposense.authorship.model.LineInfo;
+import reposense.git.GitClone;
+import reposense.model.Author;
+import reposense.model.RepoConfiguration;
+import reposense.model.RepoLocation;
+import reposense.util.TestUtil;
+
+public class AnnotatorAnalyzerTest {
```
jylee-git: Why don't you sub-class from `GitTestTemplate` which already has the testrepo-alpha cloned?

You can have a look at `FileInfoExtractorTest` class to see how test cases are done

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
anubh-v: Done

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
anubh-v: Done

src/systemtest/resources/author-config.csv
anubh-v: Sure, I've removed the changes to the `author config csv`

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -0,0 +1,79 @@
...
+                Assert.assertEquals(originalAuthor, lineAuthor);
+            } else {
+                Assert.assertEquals(overridingAuthor, lineAuthor);
+            }
+        }
+    }
+
+    @Before
+    public void before() throws InvalidLocationException {
+        super.before();
```
anubh-v: I have overriden the `before` method in `GitTestTemplate` because I have some additional logic to perform before each test case. Specifically, I need to set the same sinceDate and untilDate for the config, before each test case.

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -0,0 +1,79 @@
...
+            Author overridingAuthor) {
+        for (LineInfo line : fileResult.getLines()) {
+            Author lineAuthor = line.getAuthor();
+            if (line.getContent().startsWith("fake")) {
+                Assert.assertEquals(originalAuthor, lineAuthor);
+            } else {
+                Assert.assertEquals(overridingAuthor, lineAuthor);
+            }
+        }
+    }
```
jylee-git: These two can be put below the test cases

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Before
+    public void before() throws InvalidLocationException {
+        super.before();
+        config.setSinceDate(SINCE_DATE);
+        config.setUntilDate(UNTIL_DATE);
+    }
+
+    @Test
+    public void ignoreAnnotationTest() {
```
jylee-git: Can you help me check if in the **main codebase**, are there any instances where `annotationOverwrite` is set to false?

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
anubh-v: Done.

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
anubh-v: Sorry, I overlooked this.
`annotationOverwrite` is never set to false in the codebase, and it is always true. We already decided to remove this flag in issue #1083.

I've hence removed the `ignoreAnnotationTest` test case entirely.

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Before
+    public void before() throws InvalidLocationException {
+        super.before();
+        config.setSinceDate(SINCE_DATE);
+        config.setUntilDate(UNTIL_DATE);
+    }
+
+    @Test
+    public void applyAnnotationTest() {
```
eugenepeh: This was from what we removed previously? Perhaps can apply our test method naming conventions this time

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public FileResult getFileResult(String relativePath) {
+        FileInfo fileInfo = FileInfoExtractor.generateFileInfo(config.getRepoRoot(), relativePath);
+        return FileInfoAnalyzer.analyzeFile(config, fileInfo);
+    }
+
+    /**
+     * Asserts the correctness of authorship overriding with regards to @@author tags given
+     * in the test file.
+     */
+    public void assertAnnotationAnalysisCorrectness(FileResult fileResult, Author originalAuthor,
```
eugenepeh: this can be a private method

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -0,0 +1,69 @@
...
+            Author overridingAuthor) {
+        for (LineInfo line : fileResult.getLines()) {
+            Author lineAuthor = line.getAuthor();
+            if (line.getContent().startsWith("fake")) {
```
eugenepeh: perhaps the test methods need a little rework; this would limit the usage of this method (for future test cases) isnt it?

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
Tejas2805: Looks like you need to add the `org.junit.jupiter.api` dependency for it. Don't know if its a feasible option. Can ask the senior devs.

fzdy1914: Looks fine here.

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
anubh-v: Done

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
anubh-v: Done

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -0,0 +1,69 @@
+package reposense.authorship;
+
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.Date;
+
+import org.junit.Assert;
```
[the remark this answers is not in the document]
anubh-v: Thanks for the suggestion. Done

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
anubh-v: > JUnit provides a BeforeEach() method. So if it needs to run before each method you can use that, instead of over-riding?

In this case, the logic only needs to be executed once before all tests.

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
anubh-v: I've modified the test method. The test method now accepts the actual `FileResult` obtained  from analyzing the test file, and a `List<Author>` that contains the expected author for each line in the test file.

anubh-v: @eugenepeh  @fzdy1914  may I get your review on the changes?

anubh-v:
@fzdy1914 @eugenepeh may I get your review on the changes?

anubh-v:
Thanks very much everyone for working with me on this PR.
I would suggest we merge this soon, as this bugfix is important for the current module.

Issue #937, opened 2019-11-03
Test: use generic throw Exception in all tests


anubh-v:
The `repoConfig_overrideKeyword_success` test case has the `HelpScreenException` specified in its `throws` clause.

However, `HelpScreenException` is not thrown by the code in the body of the test case.
We should remove this exception from the `throws` clause.

![image](https://user-images.githubusercontent.com/35621759/68088448-54a48e80-fe9a-11e9-94d4-16acef7ad6ad.png)

damithc:
Actually, it is OK to use a generic `throws Exception` in test cases, as all exceptions result in failure i.e., there is no handling of exceptions

anubh-v:
I agree that we can just specify a single `Exception` in the `throws` clause of all test cases.

Another issue with the current approach: we have to modify alot of test code if we change the `throws` clause of any method that is used in many test cases.

@fzdy1914, would it be ok to start specifying a generic `throws Exception` in all future test cases?

fzdy1914:
Actually, I got a different idea from prof. Although there is no exception handler, the Exception type itself can give us some information. For example, for the above code, if we got a `ParseException`, we may easily conclude that this Exception may not occur in the test method itself but the `Parser` class it calls. Using the specific exception is a kind of defining `"normal"` exception apart from `"abnormal"` exception.

It is just my own opinion anyway.
@eugenepeh @yong24s Can you kindly provide your idea?

yong24s:
> For example, for the above code, if we got a ParseException, we may easily conclude that this Exception may not occur in the test method itself but the Parser class it calls. Using the specific exception is a kind of defining "normal" exception apart from "abnormal" exception.

But those exceptions on the `throws` clause are needed to be there by the compiler, so it may or may not be thrown, right?
It should be more consistent to read expected exceptions by writing it explicitly like this:
```
@Test(expected = ParseException.class)
assertThrows(ParseException.class, () -> new TestMethod(null));
```
This way we are clear that these exceptions are being tested or expected. Also, referencing `se-edu` practices, they are using a generic `throw Exception`.

I'm supporting the use of a generic `throw Exception`, for simplicity.

anubh-v:
Thanks everyone for the discussion. 
Since we agree on using a generic `throws Exception` clause, I will close this issue.

We can starting using the generic `throws Exception` style for new tests in future PRs.

How should we proceed for the existing tests?
I don't think there is a need to immediately make a PR to switch all tests to this new style.
[...]

PR #1072, opened 2020-02-11
[#937] Use generic 'throw Exception' in all tests

yash-chowdhary:
Part of #937 

```
It is alright to use a generic "throws Exception" in test cases,
as all exceptions result in failure.

Let's use a generic "throws Exception" clause in all tests.
```


src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -101,7 +101,7 @@ private String getInputWithDates(String sinceDate, String untilDate) {
      * by the {@code ArgumentParser} hence this is to signal to the caller that the program is safe to exit.
      */
     private void generateReport(String inputDates)
-            throws IOException, InvalidCsvException, URISyntaxException, ParseException, HelpScreenException {
+            throws Exception {
```
anubh-v: Let's move the `throws Exception` onto the previous line, with `private void generateReport..` ?

src/test/java/reposense/git/GitBranchTest.java
```diff
@@ -18,14 +15,14 @@
...
-    public void getCurrentBranch_masterBranch_success() throws GitBranchException {
+    public void getCurrentBranch_masterBranch_success() throws Exception {
         String currentBranch = GitBranch.getCurrentBranch(config.getRepoRoot());
         Assert.assertEquals("master", currentBranch);
     }
 
     @Test
     public void getCurrentBranch_uncommonDefaultBranch_success()
-            throws GitCloneException, InvalidLocationException, GitBranchException {
+            throws Exception {
```
anubh-v: Same here, the `throws Exception` can be moved up to previous line

src/test/java/reposense/git/GitLsTreeTest.java
```diff
@@ -20,13 +20,13 @@ public void after() {
...
     @Test
-    public void repo_validFilePaths_success() throws InvalidFilePathException, IOException {
+    public void repo_validFilePaths_success() throws Exception {
         validateFilePaths(config);
     }
 
     @Test(expected = InvalidFilePathException.class)
     public void windows_cloneInvalidWindowsFilePaths_throwsInvalidFilePathException()
-            throws InvalidFilePathException, IOException {
+            throws Exception {
```
anubh-v: Same here

yash-chowdhary:
> We were initially planning to make this change gradually

I noticed that while reading through the issue, but when I went through the code, it didn't seem like much.

> I think we should do this to remove unnecessary extra lines.

Addressed.

anubh-v:
Minor comment @yash-chowdhary, could you remove the words `as discussed` from your commit message? They are mostly redundant.

yash-chowdhary:
> You seem to have missed removing a few of the `import Exception` statements.

If you're referring to the Netlify build failure, it is because [Checkstyle has deprecated resolving classpath](https://github.com/checkstyle/checkstyle/issues/5088#issuecomment-432092641). 

On a side note, removing the description following `@throws` in the javadoc seems to work fine (which is quite weird).

Another thing that can be done is to just mention `@throws Exception` instead of the particular exception class since that's the crux of this PR. This change can be made wherever there are javadoc statements being used for tests. Can I proceed with this?

Tejas2805:
Sorry for being unclear. @yash-chowdhary I was referring to this:

Screen Shot 2020-02-18 at 10 06 40 AM

There are still quite a few redundant import statements like this and was asking you to remove them

yash-chowdhary:
> There are still quite a few redundant import statements like this and was asking you to remove them

Wouldn't this have been flagged in the `checkstyleTest`/ `checkstyleMain`/ `checkstyleSystemtest` check? Just clarifying.

Tejas2805:
> Wouldn't this have been flagged in the `checkstyleTest`/ `checkstyleMain`/ `checkstyleSystemtest` check? Just clarifying.

Looks like we don't have a check for unused imports. Can you confirm this once?

Tejas2805:
> Looks like we don't have a check for unused imports. Can you confirm this once?

Alright. I checked. We have a check for unused imports. The reason `checkstyle` didn't pick it up was that the exceptions still exist in the code below.

Tejas2805:
Example: `ConfigSystemTest.java`

Screen Shot 2020-02-22 at 7 35 47 PM

Tejas2805:
@yash-chowdhary Can you solve the conflict, so that I can approve the PR?

yash-chowdhary:
Yup, sure

anubh-v:
@yash-chowdhary 

I realised the commit message and title of this PR may be confusing for those who are not familiar with the issue being solved.

Could you change this sentence in the commit message: 
`Let's make use of the generic Exception class across tests.`
to `Let's use a generic "throws Exception" clause in all tests`

For the PR title, perhaps we can re-use the title of issue 937?

Tejas2805:

fzdy1914:
Netlify build is failing, can you kindly recheck and then request a review from me again.

yash-chowdhary:
@Tejas2805 @anubh-v  I was wondering if there is still any use of mentioning the types of exceptions in the javadoc comments of the tests since it's all under the `Exception` class.

fzdy1914:
No need

yash-chowdhary:
@fzdy1914 ready for review

src/test/java/reposense/git/GitLsTreeTest.java
```diff
@@ -81,11 +79,8 @@ public void unix_cloneInvalidWindowsFilePaths_success() {
     /**
      * Clones a bare repo in {@code config} and verifies that the repo contains only file paths that are
      * compatible in Windows.
-     * @throws IOException if it fails to create/delete the folders.
-     * @throws InvalidFilePathException if the repository contains invalid file paths that are not compatible with
-     * Windows.
      */
-    private void validateFilePaths(RepoConfiguration config) throws IOException, InvalidFilePathException {
+    private void validateFilePaths(RepoConfiguration config) throws Exception {
```
Tejas2805: LGTM! Just one nit. Isn't the javadoc still suppose to mention `@throws Exception`? Not sure about this. Can ask someone else for this and proceed accordingly.

anubh-v: Actually, in this case, `validateFilePaths` is a utility function that is used by test code (it's not a test itself).
Would it be better to keep the existing `throws` clause for such utility functions?
Keeping the existing `throws` clause may provide more clarity to developers who are writing tests using these functions.

We can use the generic `throws` clause for the tests themselves.
@fzdy1914 may we get your thoughts on this?

yash-chowdhary: I think that we should stick to the generic `Exception` class because the test method that calls it also throws only the generic `Exception`. 
There is little use of mentioning the specific exceptions thrown by the utility method if the test method itself doesn't mention them, is there?

anubh-v: In my view, it may be useful to specify the exceptions thrown by the utility method. 
Reason: sometimes, test cases check whether certain exceptions are thrown during testing. When a developer is writing such tests cases, it could be useful to know what exceptions are thrown by the utility method.

src/test/java/reposense/git/GitLsTreeTest.java
```diff
@@ -85,7 +84,7 @@ public void unix_cloneInvalidWindowsFilePaths_success() {
      * @throws InvalidFilePathException if the repository contains invalid file paths that are not compatible with
      * Windows.
      */
-    private void validateFilePaths(RepoConfiguration config) throws IOException, InvalidFilePathException {
+    private void validateFilePaths(RepoConfiguration config) throws Exception {
```
anubh-v: For this utility method, let's keep the original `throws` clause as well?  (i.e. specify IOException and InvalidFilePathException in the clause)

anubh-v:
@fzdy1914 may we get your thoughts on this? Thanks for your time

src/test/java/reposense/git/GitLsTreeTest.java
fzdy1914: I agrees with what @anubh-v says.

anubh-v: @yash-chowdhary has resolved these changes. 
May we get your review on the PR @fzdy1914 ? Thanks for your time

anubh-v:
@fzdy1914, @jylee-git may we get your review for this? Thanks for your time

anubh-v:
@yash-chowdhary, there were some new tests added in [AnnotatorAnalyzerTest](https://github.com/reposense/RepoSense/blob/master/src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java).

Could you help to update the `throws` clause for them too? Thanks!

anubh-v:
@jylee-git may we get your review for this?

src/test/java/reposense/git/GitLsTreeTest.java
jylee-git: @anubh-v are you referring to the javadocs in your comment?

I'm thinking of whether to include the line
```java
@throws Exception if it fails to create/delete the folders, or if the repository contains invalid file paths .....
```

in the javadocs

anubh-v:
@jylee-git may we get your review on this? Thanks for your time

src/test/java/reposense/git/GitLsTreeTest.java
anubh-v: My apologies - I missed this message.
I was referring to the `throws` clause of the method, not its the javadoc.
For this method,`validateFilePaths`, I think both the `throws` clause and the javadoc should not be modified.

This method is a utility method used for writing tests. Hence, it would be good to have the
specific exceptions listed in `throws` clause, so developers who write tests can quickly understand its behaviour.

@yash-chowdhary has reversed his changes, so the method is OK now

jamessspanggg:
@yash-chowdhary can u resolve the conflicts?

src/test/java/reposense/git/GitRevParseTest.java
```diff
@@ -10,13 +10,13 @@
...
     @Test
-    public void assertBranchExists_withExistingBranch_success() throws GitBranchException {
+    public void assertBranchExists_withExistingBranch_success() throws Exception {
         config.setBranch("master");
         GitRevParse.assertBranchExists(config, Paths.get(config.getRepoRoot()));
     }
 
     @Test (expected = GitBranchException.class)
-    public void assertBranchExists_withNonExistentBranch_throwsGitBranchException() throws GitBranchException {
+    public void assertBranchExists_withNonExistentBranch_throwsGitBranchException() throws Exception {
```
jamessspanggg, line 19: The `throws Exception` here is not needed right?

src/test/java/reposense/model/StandaloneConfigTest.java
```diff
@@ -41,51 +40,51 @@
...
+    public void standaloneConfig_trailingCommasInList_success() throws Exception {
         StandaloneConfig standaloneConfig = new StandaloneConfigJsonParser().parse(AUTHORS_TRAILING_COMMAS_CONFIG);
         config.update(standaloneConfig);
 
         Assert.assertEquals(validStandaloneConfig, standaloneConfig);
     }
 
     @Test(expected = JsonSyntaxException.class)
-    public void standaloneConfig_trailingCommasInMaps_throwsJsonSyntaxException() throws IOException {
+    public void standaloneConfig_trailingCommasInMaps_throwsJsonSyntaxException() throws Exception {
```
jamessspanggg, line 70: same goes for this

src/test/java/reposense/model/StandaloneConfigTest.java
```diff
@@ -41,51 +40,51 @@
...
         new StandaloneConfigJsonParser().parse(LITHIUMLKID_TRAILING_COMMAS_CONFIG);
     }
 
     @Test(expected = IllegalArgumentException.class)
-    public void standaloneConfig_invalidIgnoreGlob_throwIllegalArgumentException() throws IOException {
+    public void standaloneConfig_invalidIgnoreGlob_throwIllegalArgumentException() throws Exception {
```
jamessspanggg, line 75: same here

src/test/java/reposense/model/StandaloneConfigTest.java
```diff
@@ -41,51 +40,51 @@
...
         StandaloneConfig standaloneConfig = new StandaloneConfigJsonParser().parse(INVALID_IGNOREGLOB_CONFIG);
         config.update(standaloneConfig);
     }
 
     @Test(expected = IllegalArgumentException.class)
-    public void standaloneConfig_invalidFormats_throwIllegalArgumentException() throws IOException {
+    public void standaloneConfig_invalidFormats_throwIllegalArgumentException() throws Exception {
```
jamessspanggg, line 81: same here

src/test/java/reposense/model/StandaloneConfigTest.java
```diff
@@ -41,51 +40,51 @@
...
         StandaloneConfig standaloneConfig = new StandaloneConfigJsonParser().parse(INVALID_FORMATS_CONFIG);
         config.update(standaloneConfig);
     }
 
     @Test(expected = IllegalArgumentException.class)
-    public void standaloneConfig_invalidIgnoreCommit_throwIllegalArgumentException() throws IOException {
+    public void standaloneConfig_invalidIgnoreCommit_throwIllegalArgumentException() throws Exception {
```
jamessspanggg, line 87: same here

src/test/java/reposense/parser/StandaloneConfigJsonParserTest.java
```diff
@@ -76,39 +75,39 @@ public static void setUp() throws InvalidLocationException {
...
     @Test
-    public void standaloneConfig_githubIdOnlyConfig_success() throws IOException, InvalidLocationException {
+    public void standaloneConfig_githubIdOnlyConfig_success() throws Exception {
         StandaloneConfig config = new StandaloneConfigJsonParser().parse(STANDALONE_CONFIG_GITHUBID_ONLY);
         assertSameConfig(expectedGithubIdOnlyRepoconfig, config);
     }
 
     @Test(expected = JsonSyntaxException.class)
-    public void standaloneConfig_malformedJsonFile_throwsJsonSyntaxException() throws IOException {
+    public void standaloneConfig_malformedJsonFile_throwsJsonSyntaxException() throws Exception {
```
jamessspanggg, line 105: same here

yash-chowdhary:
> For those with `expected = xxx.Class` the `throws Exception` is not needed right?

In this case, can we remove the `expected = xxx.Class`? It would standardise the code. From the previous comments, going ahead with `throws Exception` would be preferred

yash-chowdhary:
> For those with `expected = xxx.Class` the `throws Exception` is not needed right?

I think we should remove the `expected = xxx.Class` in these places. Removing `throws Exception` wouldn't cut it as the `parse()` method can throw an `IOException`, but that's not what is in the "expected" clause.
Also, it would help standardise the code.

jamessspanggg:
> I think we should remove the expected = xxx.Class in these places. Removing throws Exception wouldn't cut it as the parse() method can throw an IOException, but that's not what is in the "expected" clause.

Hmm, it's fine to just remove the `throws Exception`. If the parse() method throws an IOException, the test case will simply fail since the expected exception does not match with the actual?

anubh-v:
> For those with expected = xxx.Class the throws Exception is not needed right?

Correct me if I'm wrong here - but I think both of these are still needed.

I think the `throws Exception` is still needed as the compiler expects the test method to either catch or throw any (checked) exception that is possibly thrown by the functions called in the test method.

I believe the `expected === xxx.Class` is also needed because those tests are asserting whether that particular exception is thrown.

jamessspanggg:
> I think the throws Exception is still needed as the compiler expects the test method to either catch or throw any (checked) exception that is possibly thrown by the functions called in the test method.

I don't think the throws Exception is needed as the `expected=xxx` handles the thrown exception and checks if the exception type matches. https://junit.org/junit4/javadoc/4.12/org/junit/Test.html

> The first, expected, declares that a test method should throw an exception. If it doesn't throw an exception or if it throws a different exception than the one declared, the test fails.

yash-chowdhary:
> The first, expected, declares that a test method should throw an exception. If it doesn't throw an exception or if it throws a different exception than the one declared, the test fails.

Doesn't this refer to runtime exceptions? as @anubh-v mentioned, there are compile-time errors if the `throws Exception` clause is missing

anubh-v:
> I don't think the throws Exception is needed as the expected=xxx handles the thrown exception and checks if the exception type matches

I see. Is this info available at compile time though? (perhaps I setup the project incorrectly)

jamessspanggg:
> Doesn't this refer to runtime exceptions? as @anubh-v mentioned, there are compile-time errors if the `throws Exception` clause is missing
> 
> Screenshot 2020-04-18 at 7 31 47 PM

Ahhh I see, yes @yash-chowdhary @anubh-v you guys are right. Ignore my comments above.

PR #1075, opened 2020-02-12
[#1071] Zoom view: provide a way to highlight the commit

### LINKED ISSUE #1071: Provide a way to highlight the commit
Current: clicking a tag jumps to a list of commits full of tags. Users may not be able to find the commit with that tag easily.

image

Suggested: provide a way to highlight the resulting commit with that particular tag (something like the image below - notice the "comment" highlighted in blue

image

0blivious:
Fixes #1071
```
Clicking a tag jumps to a list of commits full of tags. Users may
not be able to find the commit with that tag easily.

We can highlight the resulting commit with that particular tag
by providing a light blue border around the message div.

Let's add a focus property to the commit message body.
```


Tejas2805:
Is it possible to highlight just the commit message border ? Maybe looks neater.

damithc:
> Is it possible to highlight just the commit message border ? Maybe looks neater.

Wouldn't it be more appropriate to highlight the tag?

Tejas2805:
Alright, this is how it currently works:

Screen Shot 2020-02-13 at 3 33 03 AM

You can select a commit message and it too gets highlighted. It doesn't need to necessarily have a tag (screenshot above). If we plan to keep this functionality, we either go ahead with the current implementation or just highlighting the commit message border. If we don't plan to keep this functionality (and just selection through tags), then Prof's idea is subtle and neat. 

Can proceed accordingly now.

My suggestion:
Let's keep the functionality and highlight the commit message border. 

@damithc @fzdy1914 Your point of view.

0blivious:
If we highlight the tag will be sth like below. However I feel circle the whole commit would be better as the user is interested in which commit has that specific tag?
image

fzdy1914:
I feels that only the tag border would be better. Meanwhile, the red color border is a bit weird.

jylee-git:
My take on this: I personally feel that the blue border over the entire commit is better. 

The reason why users want to click on that tag is because they want to know the commit which that tag is assigned to. 

So highlighting the whole commit message "sort of” gives users what they want directly; as compared to just highlighting the tag itself, where the user might have to "eyeball" a little just to identify the resulting commit with that tag.

damithc:
> The reason why users want to click on that tag is because they want to know the commit which that tag is assigned to.

That's a fair point. Highlight both tag and the commit? Can use the same blue style for both. Highlighting the tag provides the additional information of which tag was clicked to arrive at the commit.

frontend/src/tabs/zoom.pug
```diff
@@ -56,8 +56,9 @@
       h3(v-else) {{ day.date }}
       template
         .commit-message(
-            v-for="slice in day.commitResults"
-            v-bind:key="slice.hash"
+            tabindex="-1",
```
fzdy1914: May I now the usage of this line?

0blivious: To enable focus for div

frontend/src/static/css/style.scss
```diff
@@ -266,6 +266,14 @@ header {
 
 /* Commit Message Body in Zoom Tab */
 .commit-message {
+  padding: 5px;
+
+  &:focus {
+    .tag{
+      border: mui-color('blue', '500') solid;
```
jamessspanggg: can use the same colour as the border for the highlighted commit?

0blivious: done

jamessspanggg: The colour still looks different tho. Is it possible to use the default blue colour for the tag?

Screen Shot 2020-02-15 at 7 56 38 PM

jamessspanggg: Currently, clicking the tag within the commit message also focus on the commit:

Kapture 2020-02-15 at 20 03 15

0blivious: This is kinda the additional functionality. You can select a commit message (by clicking anywhere within the commit-message div) and it too gets highlighted.

0blivious: > The colour still looks different tho. Is it possible to use the default blue colour for the tag?

The border color all forced to #2196f3 alr, I feel should be the screen display problem.
Screenshot (16)
Screenshot (17)

frontend/src/tabs/zoom.pug
eugenepeh: maybe inline a comment then, so that other devs will be reminded what it is for

frontend/src/static/css/style.scss
fzdy1914: If it is truly a problem, can consider the bind style functionality of vue?

0blivious: Dont think it is truly a problem. On my computer the preview page the color look the same.
image

Tejas2805:
I feel highlighting the tag is looking a bit redundant with the entire commit message highlighted. Everything else, LGTM!

damithc:
> A question to ask: if there are multiple tags for one commit, what will happen if we focus on the commit body? All the tags will be also focused?
> 
> If the answer is yes. My thought is that: the intuition to highlight the tag is to distinguish which tag leads to this commits. If all the tags are highlighted, it seems to be meaningless to highlight tag again?

From what I remember, the intention was to highlight the tag that was clicked to reach the commit. I did not think it was meant to work in the other direction. i.e., tags should not get highlighted when the focus is no the commit message as doing so does not add any value.

0blivious:
> at I remember, the intention was to highlight the tag that was clicked to reach the commit. I did not think it was meant to work in the other direction. i.e., tags should not get highlighted when the focus is no the commit message as doing so does not add any value.

I have updated the UI. The new logic now is
1. clicking on the tag will highlight both the particular tag and the relevant commit.
image

2. clicking on the commit will highlight the commit border but not the tag
image

PR #1079, opened 2020-02-14
[#960] Refactor Zoom css

### LINKED ISSUE #960: Split CSS file into its relevant part

Tejas2805:
Fixes #960

```
The JS/HTML code of each component has already rested in its 
own file, but all the CSS are in one file. 

Let's refactor the Zoom CSS into a new file for better 
scalability and readability.
```


Tejas2805:

fzdy1914:
Please resolves the conflict and then request a review from j-git.

PR #1085, opened 2020-02-15
[#1076] Author: Add a `isIgnoringFile` method

### LINKED ISSUE #1076: Author: Add a `isIgnoringFile` method
Currently, there are a few instances where we obtain the `ignoreGlobMatcher` from an `Author` instance and then manually check if this matches with a given file path.

We could simplify the design by allowing the `Author` instance to tell clients if it is ignoring a particular file (given a filepath) - instead of giving clients the `ignoreGlobMatcher` and letting clients figure out the answer on their own.

Here are the current usages of `getIgnoreGlobMatcher`:

**FileInfoAnalyzer**
https://github.com/reposense/RepoSense/blob/cc28942504f513fc7c277c6eda0ac01b353ae962/src/main/java/reposense/authorship/FileInfoAnalyzer.java#L145-L152

**AnnotatorAnalyzer**
https://github.com/reposense/RepoSense/blob/cc28942504f513fc7c277c6eda0ac01b353ae962/src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java#L32-L36

**AuthorTest**
https://github.com/reposense/RepoSense/blob/cc28942504f513fc7c277c6eda0ac01b353ae962/src/test/java/reposense/model/AuthorTest.java#L84-L87

**AuthorTest**
https://github.com/reposense/RepoSense/blob/cc28942504f513fc7c277c6eda0ac01b353ae962/src/test/java/reposense/model/AuthorTest.java#L44-L48

anubh-v:
Fixes #1076 
```
Author: Add a `isIgnoringFile` method

There are a few instances where we obtain the
ignoreGlobMatcher from an Author instance and then manually
check if this matches with a given file path.

Let's allow the Author instance
to tell clients if it is ignoring a particular file
(given a filepath) - instead of giving clients the
ignoreGlobMatcher and letting clients figure
out the answer on their own.
```


anubh-v:
@fzdy1914 can this be merged? Thanks very much for your time

PR #1088, opened 2020-02-16
Change totalLineCount and totalBlankLineCount to computed values

ccyccyccy:
```
The variables `totalLineCount` and `totalBlankLineCount` are 
calculated in `processFiles()`, a method that executes a big chunk
of logic.

Their values are calculated in a loop through every files, which is
inefficient when we can instead calculate it from
`getFileTypeExistingLinesObj` and `fileTypeBlankLinesObj` which already
have most of the the information available.

Let's extract out the calculations into computed values from
`getFileTypeExistingLinesObj` and `fileTypeBlankLinesObj` which is
clearer and more efficient.
```


ang-zeyu:
Hmm, correct me if I'm wrong, but isn't `totalLineCount` and `totalBlankLineCount` only calculated once (ie. in `processFiles)?

I think its good to utilize computed properties, but since these properties don't change, wouldn't it be even more performance heavy to have `Object.values(this.getFileTypeExistingLinesObj).reduce((acc, val) => acc + val, 0)` in the computed property than exploiting the `processFiles` implementation? ( computed properties aren't free after all )

ccyccyccy:
> 
> 
> Hmm, correct me if I'm wrong, but isn't `totalLineCount` and `totalBlankLineCount` only calculated once (ie. in `processFiles)?

Yes I think they are calculated only once every time the authorship tab is loaded

> 
> I think its good to utilize computed properties, but since these properties don't change, wouldn't it be even more performance heavy to have `Object.values(this.getFileTypeExistingLinesObj).reduce((acc, val) => acc + val, 0)` in the computed property than exploiting the `processFiles` implementation? ( computed properties aren't free after all )

Actually this implementation is more efficient because it calculates the values on a per FileType basis, while the current approach calcuates it on a per file basis.

But I didn't really make this change for the efficiency. I did it more so to make the code clearer. Currently there's a lot of things going on in `processFiles` and I feel that separating out some calculations would `processFiles` more concised.

ang-zeyu:
> Actually this implementation is more efficient because it calculates the values on a per FileType basis, while the current approach calcuates it on a per file basis.

Ahh, you're right! Didn't notice `getFileTypeExistingLinesObj` was from `commits.json` directly.

Perhaps this line in the commit message could be edited to reflect the actual cause of the efficiency as you mentioned `Let's extract out the calculations into computed values which is more clear and efficient.`

> But I didn't really make this change for the efficiency. I did it more so to make the code clearer.

:+1:

ccyccyccy:
> Perhaps this line in the commit message could be edited to reflect the actual cause of the efficiency as you mentioned `Let's extract out the calculations into computed values which is more clear and efficient.`

Ah thanks I see the confusion. Fixed.

Tejas2805:
@ccyccyccy Can you resolve the conflicts?

ccyccyccy:
> @ccyccyccy Can you resolve the conflicts?

Ok resolved.

PR #1090, opened 2020-02-16
Add auto scrolling when a segment is collapsed

ang-zeyu:
```
Add auto scrolling when a segment is collapsed

When a segment is collapsed, the scroll position of the authorship view
remains at the same position.
This can cause the user to feel lost after collapsing the segment.

Let's implement auto scrolling for the segment collapse button.
If the top button is out of the user’s view, the authorship view will be
scrolled to the segment’s top.
```

### DIFF
--- a/frontend/src/tabs/segment.pug
+++ b/frontend/src/tabs/segment.pug
@@ -1,10 +1,10 @@
 .segment(v-bind:class="{ untouched: !segment.authored, active: segment.lines.length < 5 }")
-  .closer(v-if="!segment.authored && segment.lines.length > 4", v-on:click="loadCode()")
+  .closer(v-if="!segment.authored && segment.lines.length > 4", v-on:click="loadCode()", ref="topButton")
     i.fas.fa-plus-circle.icon.open(v-bind:title="'Click to open code'")
     i.fas.fa-chevron-circle-down.icon.hide(v-bind:title="'Click to hide code'")
   div(v-if="loaded", v-hljs="path")
     .code(v-for="(line, index) in segment.lines")
       .line-number {{ segment.lineNumbers[index] + "\n" }}
       .line-content {{ line + "\n" }}
-  .closer.bottom(v-if="!segment.authored && segment.lines.length > 4", v-on:click="loadCode()")
+  .closer.bottom(v-if="!segment.authored && segment.lines.length > 4", v-on:click="collapseCode")
     i.fas.fa-chevron-circle-up.icon.hide(v-bind:title="'Click to hide code'")
--- a/frontend/src/static/js/v_segment.js
+++ b/frontend/src/static/js/v_segment.js
@@ -14,5 +14,13 @@ window.vSegment = {
         window.toggleNext(this.$el.childNodes[0]);
       });
     },
+    collapseCode() {
+      const segmentTop = this.$refs.topButton.getBoundingClientRect().top;
+      if (segmentTop < 0) {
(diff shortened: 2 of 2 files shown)


jamessspanggg:
can't really tell the difference before or after this PR. Can you explain w screenshots/screen record?

ang-zeyu:
> can't really tell the difference before or after this PR. Can you explain w screenshots/screen record?

Apologies for the confusion, here it is: 

With "auto scroll" ( the top of the authorship view jumps back to the top button of the segment if the top button is out of view, so the user can continue scrolling from there ( LineInfo.java ) ):
autoscrolldemo

---

Without: ( the top of the authorship view stays at where it was after collapsing the segment, which may leave the user wondering where the collapsed segment jumped to )
noautoscrolldemo

frontend/src/tabs/segment.pug
```diff
@@ -1,10 +1,10 @@
...
-  .closer(v-if="!segment.authored && segment.lines.length > 4", v-on:click="loadCode()")
+  .closer(v-if="!segment.authored && segment.lines.length > 4", v-on:click="loadCode()" ref="topButton")
     i.fas.fa-plus-circle.icon.open(v-bind:title="'Click to open code'")
     i.fas.fa-chevron-circle-down.icon.hide(v-bind:title="'Click to hide code'")
   div(v-if="loaded", v-hljs="path")
     .code(v-for="(line, index) in segment.lines")
       .line-number {{ segment.lineNumbers[index] + "\n" }}
       .line-content {{ line + "\n" }}
-  .closer.bottom(v-if="!segment.authored && segment.lines.length > 4", v-on:click="loadCode()")
+  .closer.bottom(v-if="!segment.authored && segment.lines.length > 4", v-on:click="collapseCode")
```
jamessspanggg: this change is not needed right?

frontend/src/tabs/segment.pug
```diff
@@ -1,10 +1,10 @@
 .segment(v-bind:class="{ untouched: !segment.authored, active: segment.lines.length < 5 }")
-  .closer(v-if="!segment.authored && segment.lines.length > 4", v-on:click="loadCode()")
+  .closer(v-if="!segment.authored && segment.lines.length > 4", v-on:click="loadCode()" ref="topButton")
```
jamessspanggg: can place a comma before `..ref`

frontend/src/tabs/segment.pug
ang-zeyu: Not absolutely needed, but i separated it so that its less confusing.
e.g. if both are combined, it would be
```
loadCode() {
  this.isLoaded = true

  const segmentTop = this.$refs.topButton.getBoundingClientRect().top;
  if (segmentTop < 0) {
    this.$refs.topButton.scrollIntoView();
  }

  this.nextTick(() => {
    window.toggleNext(this.$el.childNodes[0]);
  }
},
```

It becomes ambiguous which action actually needs what 
Expanding it requires `nextTick` and `this.isLoaded = true`,
but not the scrolling to the top.

Collapsing it conversely dosen't require `nextTick` as the dom is already loaded, nor `this.isLoaded = true`

ang-zeyu:
> From the screen shot, it seems that it does not jump to the top button of that particular button?

Hmm, I just double confirmed in the netlify preview it does; Its the first gif

ang-zeyu:
Updated on latest, no changes / conflicts

PR #1091, opened 2020-02-17
[#892] Ramp chart: allow filtering by file type/group

0blivious:
#892 

allow sort contribution by selected file type


0blivious:
merged with #1025

PR #1095, opened 2020-02-23
[#1061] Remove unused dependency

Tejas2805:
Fixes part of #1061 

```
Netlify build takes a lot of time due to a large number of 
dependencies. 

Let's remove the unused dependencies to reduce the build
time.
```


Tejas2805:
Main focus on #1096

fzdy1914:
Is there more dependency to remove?

Tejas2805:
This is the only dependency that we can remove. There might be other optimizations in Gradle possible though

Tejas2805:
@fzdy1914 Have checked for other Gradle lint rules. All seem to be okay.

fzdy1914:
OK

anubh-v:
I just have a question to better understand the gradle build file. Do we know how this dependency was being used in the past? Or was it always unused?

anubh-v:
Based on a search, it seems that this Java Parser library was previously used in the `MethodAnalyzer` class which was removed in [PR 628](https://github.com/reposense/RepoSense/pull/628/)

PR #1096, opened 2020-02-23
Add gradle lint

Tejas2805:
We can add a gradle lint to check unused dependencies and deprecated packages and update them accordingly. 

PR #1095 was based on this

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,6 +8,7 @@ plugins {
     id 'com.liferay.node' version '4.4.0'
     id 'com.github.psxpaul.execfork' version '0.1.8'
     id 'com.palantir.git-version' version '0.11.0'
+    id 'nebula.lint' version '16.1.1'
 }
 
 mainClassName = 'reposense.RepoSense'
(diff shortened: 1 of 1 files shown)


Tejas2805:
@fzdy1914 @eugenepeh If you both think having a gradle lint is a good idea, I can try out more useful rules and apply them accordingly. Or I will just stick with PR #1095 Let me know your thoughts on it.

eugenepeh:
Hmm, if its a one-time thing then its better not to integrate it into the repo as it is also a plugin itself that may somewhat affect the build time. We can however, leave it as an active PR to help us detect misuse from time to time.

jamessspanggg:
It seems like the Gradle lint isn't detecting the appropriate unused dependencies. The `com.github.javaparser` isn't even detected. How was #1095 based on this?

Screen Shot 2020-03-03 at 10 23 37 AM

fzdy1914:
@Tejas2805, I accidentally mess up this PR, can you help fix it?

PR #1098, opened 2020-02-23
[#528] author-config csv: Allow multiple repos in one row

### LINKED ISSUE #528: author-config.csv: allow multiple repos in one row
Current:

Repo | branch | ...
----- | --------- | ---
https://github.com/reposense/reposense.git | master | ...
https://github.com/markbind/markbind.git | release| ...

Suggested:

Repo/branch | ... |
-------------- | --- |
https://github.com/reposense/reposense.git/master; https://github.com/markbind/markbind.git/release | ...|

Reason: to avoid duplicating remaining columns, when one person in working in multiple repos

Example: https://github.com/nus-cs3281/2019-dashboard/blob/master/configs/author-config.csv

anubh-v:
Fixes #528

```
author-config csv files: allow multiple repos per row

Users have to create one row for each repository-branch pair in the
author-config csv file. This is tedious when the same author details
need to be duplicated for several repository-branch pairs.

To reduce the duplication, let's allow the following:

(1) Allow multiple repository locations to be specified
in the 'Repository Location` field of the author-config csv

(2) Allow users to optionally append the branch name to the URL
of the repository's git directory

Example: https://github.com/reposense/RepoSense.git#release
(corresponds to the release branch on RepoSense repo)

(3) Allow users to optionally append the branch name to the path
of a local repository

Example: home/tom/Desktop/RepoSense#master
(corresponds to master branch on local RepoSense repo)

(4) Allow users to specify the repository location and branch name
using the GitHub URL of the repo's branch.

Example: https://github.com/reposense/RepoSense/tree/release
(corresponds to the release branch on RepoSense repo)

To allow all this new syntax, let's add three new parser classes:
UrlLocationParser, PathLocationParser and RepoLocationParser.
UrlLocationParser and PathLocationParser implement the logic for
parsing the repo URLs and local repo paths respectively.
Both of these are controlled by the RepoLocationParser.
The RepoLocationParser itself is invoked from the RepoLocation
constructor.
```


anubh-v:
Status update: Completed implementation, tidying it up to prepare for review

anubh-v:
Here is a breakdown of the new `author-config.csv` format:

1. The `Repository Location` column accepts a list of repository locations. The list of locations can contain not only git URLs and file paths **but also the GitHub URL to a branch**

Example of the last case: https://github.com/reposense/RepoSense/tree/release
(corresponds to the `release` branch of RepoSense repo)

2. For the git URLs / file paths provided in the `Repository Location` column, users can **optionally** append the branch name to the URL, (as shown below). For such URLs, the `Branch` column is ignored.

Example:

Repo/branch | Branch
-- | --
https://github.com/reposense/reposense.git#release; https://github.com/markbind/markbind.git#gh-pages | master

In the above csv, the `AuthorConfiguration` instances created for `reposense` and `markbind` will use the `release` and `gh-pages` branches respectively, and not the `master` branch.

3. If a particular git URL does not have the branch name appended onto it, then the branch field is taken from the `Branch` column.

Example:

Repo/branch | Branch
-- | --
https://github.com/reposense/reposense.git#release; https://github.com/markbind/markbind.git | master

The `AuthorConfiguration` instance created for `markbind` will use the `master` branch, as specified in the `Branch` column

anubh-v:
Currently, the new syntax does not apply to `repo-config.csv` or `group-config.csv`.
However, it is possible to allow this syntax in all csv files.

anubh-v:
@damithc may I get your approval for the new syntax for specifying the repo location in `author-config.csv` ? (given in [this comment](https://github.com/reposense/RepoSense/pull/1098#issuecomment-606698219))

The syntax for other csv configs is unchanged.

src/main/java/reposense/util/SystemUtil.java
```diff
@@ -11,4 +14,18 @@
...
     }
+
+    /**
+     * Returns true if the given String is a valid URL.
+     */
+    public static boolean isValidUrl(String possibleUrl) {
+        boolean isValidUrl;
+        try {
+            new URL(possibleUrl);
+            isValidUrl = true;
```
jamessspanggg, line 25: would just `return true` and `return false` be cleaner without creating a new variable?

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -0,0 +1,63 @@
...
+import reposense.util.AssertUtil;
+
+public class RepoLocationTest {
+
+    @Test
+    public void repoLocation_parseGitUrl_success() throws Exception {
+        String gitUrl = "https://github.com/reposense/RepoSense.git";
+        RepoLocation repoLocation = new RepoLocation(gitUrl);
+        assertEquals(gitUrl, repoLocation.toString());
+        assertEquals("RepoSense", repoLocation.getRepoName());
```
jamessspanggg, line 18: should we still assert the repo name and organization when u already asserted the `repoLocation.toString()`? Since it consists both repo name and organization

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -0,0 +1,63 @@
...
+        assertEquals("https://github.com/reposense/testrepo-Alpha.git", repoLocation.toString());
+        assertEquals("testrepo-Alpha", repoLocation.getRepoName());
+        assertEquals("reposense", repoLocation.getOrganization());
+        assertEquals("release", repoLocation.getParsedBranch().get());
+    }
+
+    @Test
+    public void repoLocation_parseInvalidGitUrl_throwsInvalidLocationException() {
+        // non GitHub url should rejected
+        AssertUtil.assertThrows(InvalidLocationException.class, () ->
```
jamessspanggg, line 33: Should u assert the exception message as well? apart from the exception class

src/test/java/reposense/parser/AuthorConfigParserTest.java
```diff
@@ -183,4 +186,30 @@ public void parse_multipleColumnsWithCommasAndDoubleQuotes_success() throws Exce
...
     }
+
+    @Test
+    public void parse_multipleRepos_success() throws Exception {
+        AuthorConfigCsvParser authorConfigCsvParser =
+                new AuthorConfigCsvParser(AUTHOR_CONFIG_MULTIPLE_REPOS);
+        List<AuthorConfiguration> configs = authorConfigCsvParser.parse();
+
+        Assert.assertEquals(4, configs.size());
+        AuthorConfiguration config = configs.get(0);
```
jamessspanggg, line 197: perhaps can make use of the constants created in `RepoLocation`?

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -53,31 +54,108 @@ public String getOrganization() {
...
+        throw new InvalidLocationException("The given location is invalid");
+    }
+
+    /**
+     * Parses a given repo URL and returns an array containing the following info:
+     * { url, repository name, organisation name, branch name (if any }
+     *
+     * @return null if the given String is an invalid URL
+     */
+    private static String[] tryParsingAsRepoUrl(String location) {
```
jamessspanggg, line 95: I'm thinking whether if u can merge the methods `tryParsing...` as one as they have a lot of repeated logic.

damithc:
> @damithc may I get your approval for the new syntax for specifying the repo location in `author-config.csv` ? (given in [this comment](https://github.com/reposense/RepoSense/pull/1098#issuecomment-606698219))

I'm OK with it.

A separate question. Is this change backward-compatible with older config files?

anubh-v:
> A separate question. Is this change backward-compatible with older config files?

Yes, it is backward-compatible.

@jamessspanggg thanks for the review. I will address your feedback shortly.

src/test/java/reposense/model/RepoLocationTest.java
anubh-v: Yes, we must still assert the repo name and organisation.
The objective of the test is to check whether the repo name, organisation and location are correctly parsed from the raw string given to the RepoLocation constructor.

Example: 
If I call `new RepoLocation("https://github.com/nus-oss/reposense.git#release")`
then the location, repo name and organisation must be as follows:
- location: `https://github.com/reposense/reposense.git`
- repo name : `reposense`
- organisation: `nus-oss`

jamessspanggg: my bad, didn't look through it clearly

src/test/java/reposense/model/RepoLocationTest.java
anubh-v: Good point. Done

src/main/java/reposense/util/SystemUtil.java
anubh-v: Yes, that's simpler. Done

src/main/java/reposense/model/RepoLocation.java
anubh-v: I've merged the logic responsible for parsing both types of URLs into a single method called
`tryParsingAsUrl`. 

The two types of URLs are 
(1) URLs to a branch on GitHub, (2) URLs to the `.git` folder on GitHub

src/test/java/reposense/parser/AuthorConfigParserTest.java
anubh-v: Those constants would not apply here, yes?
In this test case, we are parsing a `author-config.csv` file, which gives us a list of `AuthorConfiguration` objects.

The rest of the test case simply asserts various properties for each `AuthorConfiguration` object in the list.

jamessspanggg: Yep my bad.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -51,31 +59,91 @@ public String getOrganization() {
...
+
     /**
-     * Verifies {@code location} can be presented as a {@code URL} or {@code Path}.
-     * @throws InvalidLocationException if otherwise.
+     * Given a String representing a repo location, returns
+     * an array containing the following details of the repository (in order):
+     * { location, repository name, organisation name, branch name (if any) }
+     *
+     * @param location a repository location, which is a file path or URL with the branch
+     *                 name optionally appended
```
jamessspanggg, line 72: indentation

jamessspanggg:
Also, there seems to be some checkstyle errors:

Screen Shot 2020-04-04 at 9 03 31 PM

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -20,25 +19,27 @@
...
-            repoName = matcher.group("repoName");
-        } else {
-            repoName = Paths.get(location).getFileName().toString().replace(GIT_LINK_SUFFIX, "");
-        }
+        this(parse(location));
+    }
+
+    private RepoLocation(String[] repoLocationDetails) {
+        assert repoLocationDetails.length == 4;
+        location = repoLocationDetails[0];
```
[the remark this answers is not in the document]
anubh-v: Thanks for the suggestion. Done

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -51,31 +54,91 @@ public String getOrganization() {
...
         }
 
-        if (!isValidPathLocation && !isValidGitUrl) {
-            throw new InvalidLocationException(location + " is an invalid location.");
+        throw new InvalidLocationException(MESSAGE_INVALID_LOCATION);
+    }
+
+    /**
+     * Parses a given path to a repo and returns an array containing the following info:
+     * { url, repository name, organisation name, branch name (if any }
```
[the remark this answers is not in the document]
anubh-v: Fixed

src/main/java/reposense/model/RepoLocation.java
anubh-v: Fixed

anubh-v:
@fzdy1914 @jylee-git may I get your feedback on this?
Thanks a great deal for your guidance

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -51,31 +59,97 @@ public String getOrganization() {
...
     }
 
+    public Optional<String> getParsedBranch() {
+        return parsedBranch;
+    }
+
     /**
-     * Verifies {@code location} can be presented as a {@code URL} or {@code Path}.
-     * @throws InvalidLocationException if otherwise.
+     * Given a String representing a repo location, returns
```
Tejas2805, line 67: I don't think the JavaDoc highlights what the method does. It just informs what it gets and what it returns - not sure if that implies what it does.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -15,28 +14,37 @@
...
-            organization = matcher.group("org");
-            repoName = matcher.group("repoName");
-        } else {
-            repoName = Paths.get(location).getFileName().toString().replace(GIT_LINK_SUFFIX, "");
-        }
+        this(parse(location));
+    }
+
+    private RepoLocation(String[] repoLocationDetails) {
+        assert repoLocationDetails.length == 4;
```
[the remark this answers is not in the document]
Tejas2805: I think Error Handling or logging might be better for inputs provided by user.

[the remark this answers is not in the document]
anubh-v: The `repoLocationDetails` array is created internally, it isn't provided by the user.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -215,6 +215,20 @@ public static boolean isEmptyFile(String directoryPath, String relativeFilePath)
         return (new File(directoryPath, relativeFilePath).length() == 0);
     }
 
+    /**
+     * Returns true if there a file exists at the path corresponding to the given String.
```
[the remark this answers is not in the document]
anubh-v: Thanks! Done

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -51,31 +59,97 @@ public String getOrganization() {
...
-            // Ignore exception
+        parsedInfo = tryParsingAsPath(location);
+        if (parsedInfo != null) {
+            return parsedInfo;
+        }
+
+        parsedInfo = tryParsingAsBranchUrl(location);
+        if (parsedInfo != null) {
+            return parsedInfo;
+        }
```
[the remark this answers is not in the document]
anubh-v: Thanks for the review - could you clarify what you mean by a single try catch block? There isn't an exception being caught here?

src/main/java/reposense/model/RepoLocation.java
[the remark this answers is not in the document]
Tejas2805: Ah I see. My bad then. Assertions are the way to go. But yeah, if I remember correctly, they are disabled by default.

src/main/java/reposense/model/RepoLocation.java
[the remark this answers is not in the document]
Tejas2805: Hmm, that is true. An exception is not actually being caught, we just throw if all cases fail.

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -0,0 +1,76 @@
...
+        assertInvalidLocation("https://github.com/reposense/RepoSense/tree");
+        // url without the 'tree' keyword should be rejected
+        assertInvalidLocation("https://github.com/reposense/RepoSense/feature_branch_issue#1010");
+        assertInvalidLocation("https://github.com/reposense/RepoSense/TREE/feature_branch_issue#1010");
+        assertInvalidLocation("https://github.com/reposense/RepoSense/tre/feature_branch_issue#1010");
+        // non GitHub url should be rejected
+        assertInvalidLocation("https://gitlab.com/reposense/RepoSense/tree/feature_branch_issue#1010");
+    }
+
+    private static void assertInvalidLocation(String rawLocation) {
```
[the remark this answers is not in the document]
anubh-v: There are no tests for `tryParsingAsPath` currently.
It is not ideal to test `tryParsingAsPath` by instantiating a new `RepoLocation` and then asserting its properties (this is what I have done to test `tryParsingAsRepoUrl` and `tryParsingAsBranchUrl`)

Reason: RepoLocation is supposed to reject paths that do not exist on the filesystem.
Hence, we cannot pass in a mock path for testing purposes.

The 1st option is to clone repositories and then test `tryParsingAsPath` using the paths to these repositories. This is not efficient as the cloning step will increase the time needed for tests to execute.

The 2nd option is to use reflection and test the specific logic within `RepoLocation` that parses the filepaths. I will go with this approach.

src/main/java/reposense/model/RepoLocation.java
anubh-v: I've used the verb `Parses` instead of `Given` and added a bit of information about the repo location. Do you think it is suitable now?

src/test/java/reposense/model/RepoLocationTest.java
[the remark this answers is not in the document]
anubh-v: @Tejas2805, I've added tests for `tryParsingAsPath` using  reflection

src/main/java/reposense/model/RepoLocation.java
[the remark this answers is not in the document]
anubh-v: Yes, assertions are disabled by default, but they are enabled when we run our tests.
Hence, I think we should keep the assertion here.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -51,31 +59,104 @@ public String getOrganization() {
...
+     *
+     * @param path A path to a repository
+     * @param isPathValidationNeeded If this flag is set to true, then the method will check
+     *         verify that {@code path} is a valid path and refers to an actual directory on the
+     *         filesystem.
+     *
+     * @return null if the given String is an invalid path, or no directory exists at the path,
+     *         and {@code isPathValidationNeeded} was set to true.
+     */
+    private static String[] tryParsingAsPath(String path, boolean isPathValidationNeeded)  {
```
anubh-v, line 107: For the `tryParsingAsPath` method, I have modified it to accept a boolean flag `isPathValidationNeeded`.
This allows us to test `tryParsingAsPath` with mock filepaths that do not exist on the file system 
(since we only want to test whether the mock filepaths are parsed correctly)

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertInvalidLocation("https://gitlab.com/reposense/RepoSense/tree/feature_branch_issue#1010");
+    }
+
+    @Test
+    public void repoLocation_parsePath_success() throws Exception {
+        Method tryParsingAsPath = RepoLocation.class.getDeclaredMethod("tryParsingAsPath", String.class,
+                boolean.class);
+        tryParsingAsPath.setAccessible(true);
+
+        String pathToGitDirectory = Paths.get("home", "users", "tom", "Desktop",
```
anubh-v, line 71: For this test, I am using `Paths.get` to create the path used for testing.
This ensures that the path created is valid on whichever OS is used to run the test.
This way, the test can be run on any OS.

anubh-v:
@jamessspanggg @Tejas2805, currently we have 2 distinct types of logic for parsing:
- parsing URLs (either the URL to the `.git` file or the URL of the branch on github)
- parsing a filepath to a repo

I think it may be better to move 2 different types of logic into separate classes, since the two forms of parsing are independent. This should make the code more readable.

So I might have a `UrlLocationParser` class that has the public methods `tryParsingAsRepoUrl` and `tryParsingAsBranchUrl`.  There will be also be a `PathLocationParser` class that has the public method `tryParsingAsPath`.

Would this be cleaner?

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -1,42 +1,32 @@
...
 
-        if (matcher.matches()) {
-            organization = matcher.group("org");
-            repoName = matcher.group("repoName");
-        } else {
-            repoName = Paths.get(location).getFileName().toString().replace(GIT_LINK_SUFFIX, "");
-        }
+    private RepoLocation(String[] repoLocationDetails) {
+        assert repoLocationDetails.length == 4;
+        location = repoLocationDetails[RepoLocationParser.LOCATION_INDEX];
```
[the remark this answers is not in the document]
anubh-v: I did not do this initially, as I thought such a change may be out of scope of the PR.
However, this may help to catch bugs in the implementation of this PR.

src/main/java/reposense/parser/RepoLocationParser.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static final String MESSAGE_INVALID_LOCATION = "The given location is invalid";
+    public static final int LOCATION_INDEX = 0;
+    public static final int REPO_NAME_INDEX = 1;
+    public static final int ORG_INDEX = 2;
+    public static final int BRANCH_NAME_INDEX = 3;
+
+    /**
+     * Parses a String representing a repo location (which could be a repo URL, branch URL
+     * or a filepath), and returns an array containing the following repo location details:
```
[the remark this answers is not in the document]
anubh-v: Sorry, are you referring to `filepath)` ? 
The `)` is closing the parenthesis from the previous line `(which could be a ...`

src/main/java/reposense/parser/AuthorConfigCsvParser.java
```diff
@@ -87,7 +90,9 @@ protected void processLine(List<AuthorConfiguration> results, CSVRecord record)
      */
     private static AuthorConfiguration findMatchingAuthorConfiguration(
             List<AuthorConfiguration> results, String location, String branch) throws InvalidLocationException {
-        AuthorConfiguration config = new AuthorConfiguration(new RepoLocation(location), branch);
+        RepoLocation repoLocation = new RepoLocation(location);
+        AuthorConfiguration config = new AuthorConfiguration(repoLocation,
+                repoLocation.getParsedBranch().orElse(branch));
```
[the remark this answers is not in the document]
anubh-v: Done

anubh-v:
> (Just a suggestion) However, perhaps moving the url pattern matching to the RepoLocationParser would be neater and easier to understand? i.e. within RepoLocationParser#parse:

Thanks a lot for your time and the suggestion.

In my view, the current design should be kept. Currently, the `UrlLocationParser` provides the ability to parse Repo URLs (i.e. the url to the `.git` folder) and parse Branch URLs (i.e. the url to a branch on GitHub).

I think it is more readable if the details of how the parsing is done (via regex) are kept hidden within `UrlLocationParser`.

`RepoLocationParser` is responsible for controlling *which* types of parsing we support, but `UrlLocationParser` and `PathLocationParser` are responsible for *how* the parsing is implemented.

Just another note  -  the code snippet you provided would not directly fit the situation (correct me if I'm wrong). The reason is that it is not sufficient to check whether the `Matcher.matches()` method returns true, we also need to keep the `Matcher` object in order to extract the repo name, organization name, and branch name from the string.

So the snippet would be like this:

```java
if (isValidUrl(locationToParse)) {
    if (gitRepositoryLocationMatcher.matches() {
        return UrlLocationParser.parse(gitRepositoryLocationMatcher, locationToParse);
    } else if (gitBranchUrlMatcher.matches()) {
        return UrlLocationParser.parse(gitBranchUrlMatcher, locationToParse);
   }
}
```

This can be done, but my view is that the usage of regex should be hidden in `UrlLocationParser`.
What do you think about my reasoning?

jamessspanggg:
> I think RepoLocationParser should only be responsible for controlling which types of parsing we support, but UrlLocationParser and PathLocationParser should be responsible for how the parsing is implemented.

Makes sense.

> The reason is that it is not sufficient to check whether the Matcher.matches() method returns true, we also need to keep the Matcher object in order to extract the repo name, organization name, and branch name from the string.

Yes I think you are right. Pardon me as I wrote the pseudocode without looking into the details.

> This can be done, but my view is that the usage of regex should be hidden in UrlLocationParser. What do you think about my reasoning?

Yeah it's good for the usage of regex to be encapsulated within the corresponding `XXXParser` object. Can keep the current implementation.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -96,7 +63,11 @@ public boolean equals(Object other) {
         }
 
         RepoLocation otherLocation = (RepoLocation) other;
-        return this.location.equals(otherLocation.location);
+        return location.equals(otherLocation.location)
+                && repoName.equals(otherLocation.repoName)
+                && (organization == null ? otherLocation.organization == null
```
jamessspanggg, line 68: perhaps can place the `?` at a new line to make it clearer?

src/main/java/reposense/parser/PathLocationParser.java
```diff
@@ -0,0 +1,71 @@
...
+                return null;
+            }
+        }
+
+        String repoName = Paths.get(pathToRepo).getFileName().toString().replace(GIT_LINK_SUFFIX, "");
+        return new String[] { pathToRepo, repoName, null, branch };
+    }
+
+    /**
+     * Given a string of the form stringA#stringB, breaks the string at the # character
```
jamessspanggg, line 51: Would this Javadoc be providing too much info? Usually we only document what and why aspects of the code, rather than how.

Perhaps `Returns a string array containing the {@code pathAndBranch} that is split based on the last '#' character`? or something like this would suffice?

src/test/java/reposense/parser/PathLocationParserTest.java
```diff
@@ -0,0 +1,32 @@
...
+import static org.junit.Assert.assertArrayEquals;
+
+import java.nio.file.Paths;
+
+import org.junit.Test;
+
+public class PathLocationParserTest {
+
+    @Test
+    public void pathLocationParser_parsePath_success() throws Exception {
```
Tejas2805, line 12: Can you give one test for a failed path? For example - Path doesn't exist, or if it does exist it doesn't contain Git Directory?

src/main/java/reposense/parser/PathLocationParser.java
anubh-v: Thanks for the feedback. I actually think the javadoc is appropriate, and I have not tried to document *how* the method works. The description of *what* the method does could be summarised - I will do this.

src/main/java/reposense/model/RepoLocation.java
anubh-v: Thanks for the feedback. I will change this.

src/test/java/reposense/parser/PathLocationParserTest.java
anubh-v: I can add the following tests:
- Test that `InvalidLocation` exception is thrown if repo directory does not exist
- Test that `InvalidLocation` exception is thrown if the `.git` directory specified by user does not exist.

Note: If the user simply provides the path to a repo directory (not the repo's `.git` directory), then we do not actually validate that the repo directory contains a `.git` directory. We simply catch the exception that is thrown during the cloning step. I have not modified this behaviour; this is the current RepoSense behaviour.

anubh-v:
@fzdy1914 may I get your thoughts on this when you are free?

I've explained the author-config format in the commit message as well as in [this comment](https://github.com/reposense/RepoSense/pull/1098#issuecomment-606698219)

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -170,6 +170,11 @@ protected String getOrDefault(final CSVRecord record, int colNum, String default
                 .collect(Collectors.toList());
     }
 
+    protected List<String> getAsListOrDefault(final CSVRecord record, int colNum, List<String> defaultList) {
```
fzdy1914, line 173: Just a question, why you want to have a default list argument for this method?  Why cannot initialize the default list here in the method body? Since no other is call this method?

src/main/java/reposense/parser/PathLocationParser.java
```diff
@@ -0,0 +1,72 @@
...
+     * @param pathAndBranch A path to a repository, with a branch name optionally appended at the end
+     *         e.g. /home/users/tom/Desktop/reposense#release
+     * @param isPathValidationNeeded If this flag is set to true, then the method will check
+     *         verify that {@code path} is a valid path and refers to an actual directory on the
+     *         filesystem.
+     *
+     * @return null if the given String is an invalid path, or no directory exists at the path,
+     *         and {@code isPathValidationNeeded} was set to true.
+     */
+    public static String[] tryParsingAsPath(String pathAndBranch, boolean isPathValidationNeeded)  {
```
fzdy1914, line 28: No need to specify `try` in the method name, `parseAsPath` is clear enough.  A normal workflow should be:
```
try:
  `parseAsPath`
```

src/main/java/reposense/parser/UrlLocationParser.java
```diff
@@ -0,0 +1,65 @@
...
+            Pattern.compile("(http|https)://github.com/(?<org>.+?)/(?<repoName>.+?)/tree/(?<branch>.+?)");
+
+    /**
+     * Parses a given repo URL and returns an array containing the following info:
+     * { url, repository name, organisation name, branch name (if any) }
+     *
+     * @return null if the given String is an invalid URL, or does not match the
+     *         format of a GitHub repo URL.
+     */
+    public static String[] tryParsingAsRepoUrl(String repoUrl) {
```
fzdy1914, line 25: Same here, no need try.

anubh-v: Done.

src/main/java/reposense/parser/PathLocationParser.java
anubh-v: Sure, I've changed the name. Just to be clear, the `parseAsPath` method doesn't throw an exception.

src/main/java/reposense/parser/CsvParser.java
anubh-v: I can indeed initialize the default list in the method body. Done.
I was originally thinking that this method may be re-used, but there may not be a need to predict future usages.

anubh-v:
I'll start updating the user guide to document the new author config csv syntax

src/test/java/reposense/parser/AuthorConfigParserTest.java
```diff
@@ -183,4 +185,32 @@ public void parse_multipleColumnsWithCommasAndDoubleQuotes_success() throws Exce
             Assert.assertEquals(AUTHOR_ALIAS_COMMAS_AND_DOUBLE_QUOTES_MAP.get(author), author.getAuthorAliases());
         });
     }
+
+    @Test
+    public void parse_multipleRepos_success() throws Exception {
```
jylee-git, line 190: I assume this tests the case where multiple repo locations are in a single row.

If so, make the test name prominent such that we know multiple locations in the same row are supposed to be captured

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -1,42 +1,32 @@
...
+    }
 
-        if (matcher.matches()) {
-            organization = matcher.group("org");
-            repoName = matcher.group("repoName");
-        } else {
-            repoName = Paths.get(location).getFileName().toString().replace(GIT_LINK_SUFFIX, "");
-        }
+    private RepoLocation(String[] repoLocationDetails) {
+        assert repoLocationDetails.length == 4;
```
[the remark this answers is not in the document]
anubh-v: That's a good point. Thanks for the suggestion.

src/test/java/reposense/parser/AuthorConfigParserTest.java
anubh-v: Thanks for the suggestion. Done.

fzdy1914:
Hi, Can you help resolve the conflicts pls? @anubh-v

anubh-v:
My thanks to everyone for their feedback on this PR. Sorry for the delay from me, on this PR.
Current status: The implementation is complete, but I need to document the new author-config syntax in the user guide.

Issue #675, opened 2019-05-17
Show commit details in code view


damithc:
Current: when the user clicks on a ramp, we open the GitHub page showing the commit

Suggested: show the commit in the code view instead

Possibly related: #63

fzdy1914:
Possibly related: #148

damithc:
@myteo did some exploration on this front. It seems the cost is too high to achieve this, in terms of processing time as well as storage space.

0blivious:
I found the github API that will return the commit diff, sth like the image below, and there are open source diff parser but havnt tested. If we only make the api query when clicking the ramp I guess the cost might be ok? The problem is we need to import the parser and deal with cors issue. 
![image](https://user-images.githubusercontent.com/39231750/76061682-07a70500-5fbf-11ea-92f9-f7a04caa00c2.png)

damithc:
> I found the github API that will return the commit diff, sth like the image below, and there are open source diff parser but havnt tested. If we only make the api query when clicking the ramp I guess the cost might be ok? The problem is we need to import the parser and deal with cors issue.

Unauthenticated users have severe API quota limitations though.

fzdy1914:
I will leave two related issues here for reference but close them.
Issues are: #63, #148.

Issue #11, opened 2018-02-01
Tidy up UI controls for view config parameters


damithc:
This part can be tidied/polished up to make a good first impression. Also, may want to optimize the layout like a nav bar at the top of the page (to save vertical space).

![image](https://user-images.githubusercontent.com/1673303/35667627-5ba7dfa4-0769-11e8-95d8-bdb71ba7c0ff.png)

PR #1101, opened 2020-02-27
[#155] Display date ranged commit in side bar

### LINKED ISSUE #155: Explore using Github Web API to display date ranged commits in the sidebar
At the moment, in the daily and weekly view for the summary ramp chart, the
users will be redirect to a new tab where they will see the information of the
individual commits from GitHub itself.

This requires the users to switch between different tabs (Github and the
dashboard). This can be distracting and troublesome for the user, especially
when trying to traverse through multiple weeks of commits.

Github offers Web API which allows us to pull the relevant information for the
specific date range.

Lets include a new module to pull and display relevant commit data from GitHub.

The display of the commits will follow GitHub's exisiting dashboard interface
which groups the commits by the dates that they are created.

0blivious:
Fixes #155
```
When the granularity is set to daily or weekly, clicking on the ramp slice
redirect the user to a new tab where they will see the information of the
individual commits from GitHub itself.

This requires the users to switch between different tabs (Github and the
dashboard). This can be distracting and troublesome for the user,
especially when trying to traverse through multiple weeks of commits.
We can improve this by opening the zoom tab for the corresponding
date ranged commits instead of open another Github page.

Let's open the zoom tab when the user clicking the ramp with the
granularity set to daily or week, instead of opening new page. 
```


jamessspanggg:
The following occurs when trying to clicking ramp slice with granularity `day`:

Kapture 2020-02-28 at 10 51 32

0blivious:
> The following occurs when trying to clicking ramp slice with granularity `day`:

fixed

jamessspanggg:
@0blivious can fix CI?

@damithc may I get your approval for the following in functionality before I start reviewing?

damithc:
Should the ramp chart in the zoom view use per-commit view, give that the rest of the page shows commit details? Otherwise it always show just one ramp at the corner.

0blivious:
> Should the ramp chart in the zoom view use per-commit view, give that the rest of the page shows commit details? Otherwise it always show just one ramp at the corner.

Can but the color is calculated by the date so if we choose `week` or `date` we will have some slices of the same color displayed in zoom. Would that be ok or we prefer different color for each commit slick.

damithc:
> Can but the color is calculated by the date so if we choose `week` or `date` we will have some slices of the same color displayed in zoom. Would that be ok or we prefer different color for each commit slick.

Use the same color for all commit ramps to look like the same week/day ramp is being split into multiple commit ramps.

0blivious:
> Use the same color for all commit ramps to look like the same week/day ramp is being split into multiple commit ramps.

Is this one ok?
image

damithc:
> Is this one ok?

yes, assuming the user clicked on the green ramp to begin with.

frontend/src/ramp.pug
```diff
@@ -17,6 +17,26 @@
           }"
       )
 
+  template(v-else-if="zoomfromramp")
```
jamessspanggg: Instead of adding another `v-else-if` template that consists of mostly duplicate code, how about just add on to the `v-else` template where it looks smth like this:

`v-bind:href="zoomfromramp ? getLink(user, slice)", target="_blank" : javascript:void(0)`
`v-on:click="zoomfromramp ? {} : openTabZoom(user, slice)"`

frontend/src/ramp.pug
```diff
@@ -26,18 +46,13 @@
...
-      v-bind:href="getLink(user, slice)", target="_blank",
+      v-on:click="openTabZoom(user, slice)",
       v-bind:class="'ramp__slice--color' + getSliceColor(slice.date)",
       v-bind:style="[\
           {\
             zIndex: user.commits.length - j,\
             borderLeftWidth: getWidth(slice) + 'em',\
             right: (getSlicePos(tframe === 'day' ? slice.date : slice.endDate) * 100) + '%' \
-          },\
-          /* disallow clickable ramp slices when merging groups that are grouped by authors \
```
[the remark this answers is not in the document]
0blivious: The reason for making the slice unclickable is that we are unable to find a github website displaying the commits from the particular user for multiple repos. Now with this new sidebar implementation we can display this in zoom panel.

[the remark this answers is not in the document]
jamessspanggg: Yes I understand, but now currently in the zoom panel, slices with merge group are seen to be clickable when it's not:
Kapture 2020-03-01 at 16 59 40

To reproduce: apply zoom in to a ramp with granularity `week` and `merge group` checked, go to zoom tab and hover over slices.

jamessspanggg:
Another issue, when using the zoom feature, ctrl/cmd clicking the first `zoom since` region opens up the zoom tab, which shouldn't be the case:
Kapture 2020-03-02 at 12 38 03

frontend/src/ramp.pug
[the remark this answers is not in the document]
0blivious: I intend to make them clickable, to display the date ranged commit at zoom. I am facing some problems with the tab design though. How can we refresh/reload the zoom tab?

0blivious:
> Another issue, when using the zoom feature, ctrl/cmd clicking the first `zoom since` region opens up the zoom tab, which shouldn't be the case:

fixed

frontend/src/ramp.pug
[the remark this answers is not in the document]
0blivious: @fzdy1914 Can I get some help for this? Or we should stick to the previous design that ramp slice in zoom panel should not be clickable.

jamessspanggg:
> fixed

cmd clicking still has the same behaviour tho

frontend/src/ramp.pug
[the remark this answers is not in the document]
jamessspanggg: > Or we should stick to the previous design that ramp slice in zoom panel should not be clickable.

Note that the previous implementation is ramp slices in zoom panel is not clickable only when `merge group` is checked

0blivious:
> cmd clicking still has the same behaviour tho

Should due to the different cmd key in mac. Try again?

jamessspanggg:
> Should due to the different cmd key in mac. Try again?

works fine now

frontend/src/ramp.pug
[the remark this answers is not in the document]
0blivious: got it solved but use this.$parent.emit.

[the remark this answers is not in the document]
jamessspanggg: > I intend to make them clickable, to display the date ranged commit at zoom.

I'm not sure if this should be the intended behaviour, where clicking at `weekly` and `daily` ramps in the **zoom tab** will further show the commits within the `week` or `day`. The behaviour feels weird to me.

Kapture 2020-03-03 at 16 30 21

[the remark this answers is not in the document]
0blivious: My reason is, if the user want to see commits in a certain slice, he might want to zoom in to make the ramp slice more easily selectable then clicking on the slice in zoom to get a view of the commits.

@fzdy1914 @damithc How do you feel about this?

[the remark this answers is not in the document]
damithc: > I'm not sure if this should be the intended behaviour, where clicking at `weekly` and `daily` ramps in the **zoom tab** will further show the commits within the `week` or `day`.

What the the alternatives?

jamessspanggg:
@damithc this PR causes clicking ramps to show the commits in the zoom view. But in a previous issue #675,  you wanted to show the commit in the code view instead. Which behaviour do you prefer?

damithc:
> @damithc this PR causes clicking ramps to show the commits in the zoom view. But in a previous issue #675, you wanted to show the commit in the code view instead. Which behaviour do you prefer?

#675 might not be feasible after all. In the meantime, we can go with this solution.

frontend/src/ramp.pug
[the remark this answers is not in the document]
jamessspanggg: It's either clicking at weekly and daily ramps in the **zoom tab** will:
- further show the commits within the week or day in zoom tab. or
- follow previous implementation, bring to GitHub commits with the corresponding duration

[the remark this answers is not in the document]
0blivious: @damithc your opinion?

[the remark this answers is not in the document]
damithc: I don't mind giving the first option a try as it feels like a slightly better user experience than going to GitHub directly, unless you guys see some downside.

frontend/src/static/js/v_ramp.js
```diff
@@ -57,6 +36,42 @@ window.vRamp = {
       const newSize = 100 * (slice.insertions / this.avgsize);
       return Math.max(newSize * this.rampSize, 0.5);
     },
+    getTitle(slice, commit) {
+      let title = '';
+      if (this.tframe === 'commit') {
+        return `[${slice.date}] ${commit.messageTitle}: ${commit.insertions} lines`;
+      }
+
+      if (this.zoomfromramp) {
```
jamessspanggg: This is not necessary as all ramp slice when zoom from ramp will be of granularity commit

frontend/src/ramp.pug
```diff
@@ -1,13 +1,13 @@
 .ramp
-  template(v-if="tframe === 'commit'")
+  template(v-if="tframe === 'commit' || zoomfromramp")
```
jamessspanggg: Is it really needed to have another `zoomfromramp` field? Can't we just use tframe? I.e. when we call openTabZoom from the ramp, we pass the tframe to be `commit`.

frontend/src/static/js/v_ramp.js
```diff
@@ -57,6 +36,42 @@ window.vRamp = {
...
+      }
+
+      title = this.tframe === 'day'
+            ? `[${slice.date}] Daily `
+            : `[${slice.date} till ${slice.endDate}] Weekly `;
+      title += `contribution: ${slice.insertions} lines`;
+      return title;
+    },
+    openTabZoom(user, slice, evt) {
+      if (window.isMacintosh ? evt.metaKey : evt.ctrlKey) {
```
jamessspanggg: can put a comment that states the skipping of opening zoom tab when cmd/ctrl click

frontend/src/static/js/v_ramp.js
0blivious: No the tframe will still be week/day accordingly. That's why I need the zoomfromramp as the title of ramp will be slightly different. If it is zoomfromramp and with week granularity the title will be for eg `week of 2018-06-04 [#11] Implement logger`. If it is not from ramp and commit then will be `[2018-06-04] [#11] Implement logger`.

frontend/src/ramp.pug
0blivious: Need, otherwise in zoom will have a list of commit contributed on the same day which is not true.

frontend/src/static/js/v_ramp.js
jamessspanggg: Currently, when clicking on ramps with granularity `day` or `week`, it shows the zoom tab with a ramp chart consists of ramps with granularity `commit`. 

Instead of showing `week of 2018-06-04 [#11] Implement logger` as the title, standardising it with normal ramps with granularity commits (i.e. displays the true commit date) would be a better behaviour.

frontend/src/ramp.pug
jamessspanggg: > Instead of showing week of 2018-06-04 [#11] Implement logger as the title, standardising it with normal ramps with granularity commits (i.e. displays the true commit date) would be a better behaviour.

If the implementation follows this, then `zoomfromramp` won't be needed.

frontend/src/static/js/v_ramp.js
```diff
@@ -74,7 +85,13 @@ window.vRamp = {
       return new Date(untilDate) - new Date(sinceDate);
     },
     getSliceColor(date) {
-      const timeMs = (new Date(date)).getTime();
+      let timeMs;
+      if (window.hashParams.timeframe !== this.tframe) {
```
jamessspanggg: can explain the need to add this logic?

frontend/src/static/js/v_zoom.js
```diff
@@ -8,7 +8,7 @@ window.vZoom = {
   template: window.$('v_zoom').innerHTML,
   data() {
     return {
-      filterTimeFrame: window.hashParams.timeframe,
+      filterTimeFrame: this.info.timeFrame ? this.info.timeFrame : window.hashParams.timeframe,
```
jamessspanggg: can just pass the `timeFrame` at `openTabZoom` in `v_summary`?

frontend/src/static/js/v_ramp.js
```diff
@@ -57,6 +36,38 @@ window.vRamp = {
...
+      title += `contribution: ${slice.insertions} lines`;
+      return title;
+    },
+    openTabZoom(user, slice, evt) {
+      // prevent opening of zoom tab when cmd/ctrl click
+      if (window.isMacintosh ? evt.metaKey : evt.ctrlKey) {
+        return;
+      }
+
+      const newUser = { ...user };
```
jamessspanggg: Perhaps zoomUser?

frontend/src/static/js/v_zoom.js
0blivious: Though zoomfromramp is removed, I still need a mean to distinguish the two for the `getSliceColor` function. As the color is calculated using date and we want same color for all commit ramps to look like the same week/day ramp, when the zoom is opened by clicking ramp with week granularity, we need to use sdate instead of date in the  `getSliceColor`.

frontend/src/static/js/v_ramp.js
jamessspanggg: > Though zoomfromramp is removed, I still need a mean to distinguish the two for the getSliceColor function. As the color is calculated using date and we want same color for all commit ramps to look like the same week/day ramp, when the zoom is opened by clicking ramp with week granularity, we need to use sdate instead of date in the getSliceColor.

Ok noted. @0blivious this comment addresses this question instead right?

frontend/src/static/js/v_zoom.js
0blivious: ya, sorry under wrong section

frontend/src/static/js/v_ramp.js
0blivious: ya, sorry under wrong section

frontend/src/static/js/v_zoom.js
jamessspanggg: I'm sorry I don't quite get your point.

Why is zoom panel reactive to the change in granularity in summary? I thought they are independent? i.e. after opening zoom tab, changing the granularity in summary won't affect the zoom tab whatsoever?

0blivious: They are not independent. Change granularity in summary will change display in zoom. You can try preview of other PR.

jamessspanggg: Ahhh I see. Ok noted.

jylee-git:
When `Merge Group` is checked and `Granularity` is set to day/week, try clicking on the ramps in the summary panel. Notice that in the commits panel, none of the commits appear.

frontend/src/static/js/v_zoom.js
jamessspanggg: @fzdy1914 PR #830 is currently allowing zoom tab to be restorable after reloading, which requires the zoom tab and summary to be completely independent. So should we completely make them independent here? i.e. changing the filterTimeFrame in summary wont affect the zoom tab.

0blivious:
> When `Merge Group` is checked and `Granularity` is set to day/week, try clicking on the ramps in the summary panel. Notice that in the commits panel, none of the commits appear.

fixed

jamessspanggg:
When merge group, it is also seen that some commits are in the wrong dates:

Commit with date 01/09/2018
Screen Shot 2020-03-13 at 3 39 20 PM

Real committed date of 21/02/2019
Screen Shot 2020-03-13 at 3 35 07 PM

jamessspanggg:
@0blivious can you explain why the addition of this PR caused the previous regression?

0blivious:
> @0blivious can you explain why the addition of this PR caused the previous regression?

The `mergeCommits` function. Previously we push the original commit inside. So when calling mergeCommits for the second time, the commitResults of the original commit got updated again.
`const commitWithSameDate = merged[dateToIndexMap[date]];`
`commitWithSameDate.commitResults.push(commitResult);`

frontend/src/static/js/v_zoom.js
jamessspanggg: let's put this on hold until #967 is merged, as it affects the implementation of this PR

frontend/src/static/js/v_summary.js
```diff
@@ -465,13 +465,20 @@ window.vSummary = {
...
         const mergedFileTypeContribution = {};
         let mergedVariance = 0;
         let totalMergedCommits = 0;
 
         group.forEach((user) => {
-          this.mergeCommits(user, mergedCommits, dateToIndexMap);
+          user.commits.forEach((commit) => {
+            this.mergeCommits(commit, user, dateToIndexMap, mergedCommits);
+          });
+          user.dailyCommits.forEach((commit) => {
```
jamessspanggg: From my discussion with you last time, this extra step is so that you can get the `trueCommittedDate` for commits with granularity `week`? How about we just have an extra property `trueCommittedDate` for those commits?

0blivious: The problem is the ramp.pug file and the difference in structure of commits. The ramp.pug will try to read commit.date. When set to week, the commits belong to the same week will all be in commit.commitResult. We thus cannot read the true date of each commit unless we change all related code in ramp and zoom

jylee-git:
Regression in this PR:

1. Open commits panel of any author
2. click on `hide all commit messages`
3. Open commits panel of another author

Notice that commit message body expands for all, but the button still stays as `show all commit messages`. I need to click it twice to hide them again.

0blivious:
> Regression in this PR:
> 
> 1. Open commits panel of any author
> 2. click on `hide all commit messages`
> 3. Open commits panel of another author
> 
> Notice that commit message body expands for all, but the button still stays as `show all commit messages`. I need to click it twice to hide them again.

This should not be caused by this branch but due to #967. After merging the latest master branch this pr should be working fine now.

jamessspanggg:
Travis is failing because u are missing a comma

Screen Shot 2020-04-03 at 12 16 52 PM

0blivious:
waiting for #1162

jamessspanggg:
Can update the PR message body to fix issue #1213 as well

jamessspanggg:
@0blivious Can you open a separate PR to fix #1213 instead? Since this PR is stalled by another PR and the point of this PR is not fixing #1213.

frontend/src/static/js/v_ramp.js
```diff
@@ -51,6 +30,43 @@ window.vRamp = {
       const newSize = 100 * (slice.insertions / this.avgsize);
       return Math.max(newSize * this.rampSize, 0.5);
     },
+    getTitle(slice, commit) {
```
jamessspanggg, line 33: perhaps `getContributionMessage` might be better? `getTitle` seems to general

frontend/src/ramp.pug
```diff
@@ -20,24 +20,16 @@
...
       v-for="(slice, j) in user.commits",
       v-if="slice.insertions > 0",
-      v-bind:title="`${tframe === 'day' \
-        ? '[' + slice.date + '] Daily' \
-        : '[' + slice.date + ' till ' + slice.endDate + '] Weekly'} contribution: ${slice.insertions} lines`",
-      v-bind:href="getLink(user, slice)", target="_blank",
+      v-bind:title="getTitle(slice)",
+      v-on:click="openTabZoom(user, slice, $event)",
       v-bind:class="'ramp__slice--color' + getSliceColor(slice.date)",
       v-bind:style="[\
```
jamessspanggg, line 28: can remove the square brackets now

0blivious:
> Seems like the zoom from ramp is not restorable after reloading:
> To replicate, `group by authors`, `granularity week`.

I could not replicate this. The reloading works fine on my computer?

ezgif com-video-to-gif

jamessspanggg:
> > Seems like the zoom from ramp is not restorable after reloading:
> 
> > To replicate, `group by authors`, `granularity week`.
> 
> 
> 
> I could not replicate this. The reloading works fine on my computer?
> 
> 
> 
> ezgif com-video-to-gif
> 
> 

Note that merge group is checked as well. Forgot to mention that.

0blivious:
> Note that merge group is checked as well. Forgot to mention that.

The problem actually occurs at the restoring of zUser when mergeGroup is selected. The zUser is not restored properly and is actually irrelevant to this PR (you can try open any zoom page then refresh). One weird thing is that this only occurs on the surge preview, my local html file works fine.

jamessspanggg:
> The problem actually occurs at the restoring of zUser when mergeGroup is selected. The zUser is not restored properly (you can try open any zoom page then refresh). One weird thing is that this only occurs on the surge preview, my local html file works fine.

I see, I realised what was the problem. It is because there's two author's (eugenepeh) with the same name.

> Tested on other PR's preview, the problem is in the master branch. But when I open the web on localhost the restoring works fine.

I think you need to copy the exact same analysis in order to replicate it.

jamessspanggg:
Yes, the bug is unrelated to this PR. I created a PR to solve that issue. Can continue with the comments i gave above.

0blivious:
> A question/suggestion:
> 
> When the user refreshes the page with the date ranged commits being displayed on the side bar, should the commits tab pertaining to the date ranged commits be restored?

Are you saying the zoom tab should retain the previously opened page? It is actually designed to be restorable, just need to update based on latest master branch

jylee-git:
> > A question/suggestion:
> > When the user refreshes the page with the date ranged commits being displayed on the side bar, should the commits tab pertaining to the date ranged commits be restored?
> 
> Are you saying the zoom tab should retain the previously opened page? It is actually designed to be restorable, just need to update based on latest master branch

Yes. I think your e337539 addressed the issue

fzdy1914:
One suggestion from me is that we should standardize the behavior when we click the ramp on the v_zoom page to be all open the github page.

fzdy1914:
Codewise OK for me

fzdy1914:
~> One suggestion from me is that we should standardize the behavior when we click the ramp on the v_zoom page to be all open the github page.~

You may ignore this, actually, the current logic is Good.

jamessspanggg (approved):
LGTM

PR #1099, opened 2020-02-23
[#950] Use surge as netlify alternative (testing, WIP)

jamessspanggg:
Fixes #950 

```
Netlify's build quota is not enough for RepoSense's build usage.
```

PR #1105, opened 2020-02-29
[#950] Use surge as netlify alternative (#1099)

### LINKED ISSUE #950: Find a netlify alternative
Netlify is implementing a limit of 300 build minutes per month; it may not be sufficient for RepoSense as we have been using it as a PR review tool.
Hence, we might need to look out for an alternative.

https://www.netlify.com/pricing/#features

jamessspanggg:
Fixes #950 

```
Netlify's build quota is not enough for RepoSense's build usage.

Let's deploy on Surge instead, using Travis to run the build.
```


eugenepeh:
Travis automatic deployment: https://1105-pr-RepoSense-reposense.surge.sh

jamessspanggg:
@eugenepeh the auto commenting of deployment link works now. Thanks!

Now I just have to figure out a way to either:
- only commenting once for the link even with consecutive builds, or
- edit the pr message to include the deployment link

eugenepeh:
Travis automatic deployment: https://1105-pr-RepoSense-reposense.surge.sh

eugenepeh:
Travis automatic deployment: https://-RepoSense-reposense.surge.sh

eugenepeh:
@jamessspanggg can consider using status check too https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-status-checks

jamessspanggg:
@eugenepeh I have one question, should we only deploy for every PR? or branches and tag pushes should be considered as well?

eugenepeh:
>@eugenepeh I have one question, should we only deploy for every PR? or branches and tag pushes should be considered as well?

Hmm, for this surge solution, it only works for PR? It would be nice if it can work for branch and tag as well. Although we could probably continue to use netlify just for the branch and tag, but that would mean that we would need to settle the outstanding over usage fee first.

jamessspanggg:
> Hmm, for this surge solution, it only works for PR? It would be nice if it can work for branch and tag as well. Although we could probably continue to use netlify just for the branch and tag, but that would mean that we would need to settle the outstanding over usage fee first.

Surge works for branch and tags as well. Was just asking because wasn't sure if deployments with branch and tags are necessary.

jamessspanggg:
@eugenepeh @damithc managed to use GitHub Statuses to provide a link to the deployed domain. Would this way of deploying be okay?

Screen Shot 2020-03-01 at 12 22 15 AM
Note that the messages for the status can be modified.

damithc:
> @eugenepeh @damithc managed to use GitHub Statuses to provide a link to the deployed domain. Would this way of deploying be okay?

Nice!

jamessspanggg:
Note that there is a limit with GitHub Statutes where each commit sha and context is limited to 1000 statuses, which should be sufficient.

Screen Shot 2020-03-01 at 12 34 17 AM

eugenepeh:
>Surge works for branch and tags as well. Was just asking because wasn't sure if deployments with branch and tags are necessary.

It helps if we want to make visual comparison against master or specific version

>@eugenepeh @damithc managed to use GitHub Statuses to provide a link to the deployed domain. Would this way of deploying be okay?

Nice, great work 👍

.travis.yml
```diff
@@ -60,3 +57,8 @@ cache:
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+  - git fetch --all && git config --global user.email "-" && git config --global user.name "-"
```
jamessspanggg: did not add `gradlew clean build` here as it is unnecessary and will re-run the tests.

.travis.yml
```diff
@@ -47,9 +47,6 @@ before_cache:
   - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
   - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
 
-after_success:
```
jamessspanggg: Removed the following as I think it is unnecessary to have two `after_success` hook.

jamessspanggg:
@yong24s may I get your thoughts on this? I'm still new to deploying stuffs so unsure if I'm doing it correctly

.travis.yml
jamessspanggg: `git fetch --all && git config --global user.email "-" && git config --global user.name "-"` the previous netlify build executes this, but I'm not sure what does it do? May I clarify with you guys @eugenepeh @yong24s

.travis.yml
```diff
@@ -60,3 +57,8 @@ cache:
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+  - git fetch --all && git config --global user.email "-" && git config --global user.name "-"
+    && ./gradlew run -Dargs="--since d1"
+  - chmod ugo+x ./config/travis/deploy-surge.sh
```
anubh-v: Just want to clarify this -  there a reason we do not need to set the permissions for our script: `./config/travis/run-checks.sh`, but it's needed for this ?

jamessspanggg: Good catch, it is not needed.

.travis.yml
anubh-v: Agreed; will be good to determine if this line is necessary.

jamessspanggg:
> Secure environment variables are not included in pull request builds for security reasons. If they were, then anyone could submit a pull request and simply print out all the environment variables.
> 
> Unfortunately, we haven't been able to find a good way to get around this (there may not be one).

I found out an issue where environment variables can only be accessed from PRs that are from the same repository, that's why PRs from the forked wouldn't be able to access the environment variables. This explains why in #1099 we need to merge to a branch in the team repo in order to access the environment variables. Will try to find a workaround for this.

jamessspanggg:
Screen Shot 2020-03-05 at 7 55 26 PM

The only way for surge to work is to merge the forked branch into a branch from team's repo 😢 @damithc @eugenepeh thoughts on this?

fzdy1914:
Doesn't seems to be so promising.

.travis.yml
fzdy1914: `git fetch --all` is necessary because we need it to have the full git history to get the version correctly. The git config one I am not sure either.

damithc:
> Doesn't seems to be so promising.

Yes, this seems to be a hard problem. Perhaps the fork owner can set up something? e.g., whenever the fork owner pushes a commit to a branch in the fork, travis can build it, upload the site to surge, and post the link in the corresponding PR in the upstream repo (if there is a matching PR). The fork owner has the ability to update the PR description for example.

jamessspanggg:
> and post the link in the corresponding PR in the upstream repo (if there is a matching PR). The fork owner has the ability to update the PR description for example.

I've tried this and apparently posting the link automatically in the upstream repo is not feasible as it requires the upstream's environment variables.

damithc:
> I've tried this and apparently posting the link automatically in the upstream repo is not feasible as it requires the upstream's environment variables.

How about post as a comment, not as a status?

Also, would using GitHub actions help? I assume the actions for a PR run in the upstream repo, not in the fork.

jamessspanggg:
> How about post as a comment, not as a status?

Yes I was referring to posting comments.

> Also, would using GitHub actions help? I assume the actions for a PR run in the upstream repo, not in the fork.

I'll look up on it.

jamessspanggg:
> Also, would using GitHub actions help? I assume the actions for a PR run in the upstream repo, not in the fork.

Apparently the same issue goes for Github Actions. :(

> GitHub explicitly disables giving secrets to workflows that originate from a fork. That means that when a user opens a pull request on your project from a fork, no secrets are provided to this workflow.

Tejas2805:
> Apparently the same issue goes for Github Actions. :(

I think this is true. But I have been reading up on it, and I think we can trigger a workflow to cause it to happen.

jamessspanggg:
> **UPDATE:** It might be possible. If we do this, and it's possible to do this - We use Github Actions to merge our pull requests to another branch and open a pull request from that branch to the master.

Hmm, the merging of pull requests requires the environment variables as well. So it may not be feasible to automate the merging and opening of pull requests.

eugenepeh:
>How about post as a comment, not as a status?

If we post as a comment, we should be able to setup and utilize a toon account instead. 
That way we should be able to expose the env variables we need

jamessspanggg:
> If we post as a comment, we should be able to setup and utilize a toon account instead.
> That way we should be able to expose the env variables we need

Okay, so with this solution:
- Fork owner must set up his/her own surge account, whenever forked owner pushes a commit to his/her own branch, travis builds and deploys it to surge.
- When PR is made, we utilise the dummy account to post the corresponding link as a comment on the PR? (without knowing if the deployment is successful or not?)

Instead of creating a dummy account to post the link, how about we enforce our PR message convention to include the link to the surge deployment?

eugenepeh:
is there any risk in exposing the surge token as well? if not we can expose the surge token too to save all the trouble.

jamessspanggg:
Other users can just use our surge token to deploy their own stuffs, maybe the only concern is that they may deploy bad stuffs? (Illegal stuffs lol)

eugenepeh:
>Other users can just use our surge token to deploy their own stuffs, maybe the only concern is that they may deploy bad stuffs? (Illegal stuffs lol)

Hmm, for the surge token, is it exposed by default if it is a public env variable (given that its entered into a password field)? if it is not then i guess its fine

jamessspanggg:
> > Other users can just use our surge token to deploy their own stuffs, maybe the only concern is that they may deploy bad stuffs? (Illegal stuffs lol)
> 
> Hmm, for the surge token, is it exposed by default if it is a public env variable (given that its entered into a password field)? if it is not then i guess its fine

No it is not exposed unless the user tries to print it out using `echo`.

.travis.yml
jamessspanggg: @eugenepeh can I seek for your clarification on this?

damithc:
> No it is not exposed unless the user tries to print it out using `echo`.

I don't think we should leave room someone else to retrieve our token (i.e., it should be _impossible_ for them to retrieve it even if they tried) or use our token (without really knowing the actual value) to do something we didn't intend, such as deploy their own websites.

eugenepeh:
>I don't think we should leave room someone else to retrieve our token (i.e., it should be impossible for them to retrieve it even if they tried) or use our token (without really knowing the actual value) to do something we didn't intend, such as deploy their own websites.

Surge token seems rather insignificant; anyone can create one just by providing an email address. 
If i am not wrong (please correct me if i am), there's nothing preventing others to use their own surge token and deploy a website with similar name? since we're not using any custom names.

jamessspanggg:
> If i am not wrong (please correct me if i am), there's nothing preventing others to use their own surge token and deploy a website with similar name as ours? since we're not using any custom names.

I don't think so. I tried to deploy the website using the deployed surge domain of this PR and got the following message:
`Aborted - you do not have permission to publish to 1105-pr-reposense-reposense.surge.sh`

I think they check if user is the one who deployed the domain, if yes then he/she is able to continue deploying to that domain.

eugenepeh:
>I think they check if user is the one who deployed the domain, if yes then he/she is able to continue deploying to that domain.

I see, but they would be able to publish a child domain e.g. `1147-pr-reposense-reposense.surge.sh` that we didn't use right?

jamessspanggg:
> I see, but they would be able to publish a child domain e.g. `1147-pr-reposense-reposense.surge.sh` that we didn't use right?

Yes they should be able to.

eugenepeh:
>Fork owner must set up his/her own surge account, whenever forked owner pushes a commit to his/her own branch, travis builds and deploys it to surge.

Btw, I don't think this would work either. PRs from this repo would follow the travis configuration setup of this repository. Hence, the fork owners' configuration wouldn't be applied to the builds here.

>How about the GITHUB_API_TOKEN? Are we making it public as well? In order for GitHub Statuses to work, the token must only include repo scope in the repository.
Update: only the repo:status is needed.

@jamessspanggg if we use github actions instead, is repo:status the only permission needed?

eugenepeh:
>@jamessspanggg if we use github actions instead, is repo:status the only permission needed?

If its a yes then let's quickly move on to set it up, we only have 9 days till the netlify service gets suspended

jamessspanggg:
> @jamessspanggg if we use github actions instead, is repo:status the only permission needed?

You mean using github actions to run the `gradlew run` command, deploy to surge and create a Status?
Then it should be yes, repo:status is only needed.

But what is the rationale in using github actions instead of just using travis?

eugenepeh:
>But what is the rationale in using github actions instead of just using travis?

If prof has concern over the surge token, then I guess github actions would be a better alternative.

The other benefit would be that, the preview build would be in parallel to travis which leads to faster preview

jamessspanggg:
> If prof has concern over the surge token, then I guess github actions would be a better alternative.
> 
> The other benefit would be that, the preview build would be in parallel to travis which leads to faster preview

Hmm, I'm not so familiar with using GitHub actions, but I have tried few weeks before to use GitHub actions to deploy surge. I don't think GitHub actions is able to do the job where forked PRs are able to use the environment variables.

To pass environment variables to be used in github action workflows, you have to set the value of the environment variable as a secret. However, secrets cannot be made public so it can't be accessed by forked repos. [Link to relevant info](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets)

Another way of passing environment variables is to create a virtual environment, using docker. I am not so familiar with how to do this as well. [Link to relevant info](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#environment-variables)

eugenepeh:
@damithc perhaps we can use this solution (using public env var: surge token and repo:status token), given that we have limited time and the risk is low, while we try to employ other methods such as GitHub Action?

damithc:
> @damithc perhaps we can use this solution (using public env var: surge token and repo:status token), given that we have limited time and the risk is low, while we try to employ other methods such as GitHub Action?

Sure.

Tejas2805:
> To pass environment variables to be used in github action workflows, you have to set the value of the environment variable as a secret. However, secrets cannot be made public so it can't be accessed by forked repos.

I tried too. It doesn't work.

jamessspanggg:
@eugenepeh then you would have to set those env variables to public by toggling on the following when adding env variables:

Screen Shot 2020-03-18 at 12 10 06 AM

eugenepeh:
@jamessspanggg done, please give it a shot

jamessspanggg:
> @jamessspanggg done, please give it a shot

Ok, will make a PR from a forked branch to see if it works

jamessspanggg:
closing this as overlaps with #1165

PR #1100, opened 2020-02-25
[#873] Enable view commits option when merge group

jamessspanggg:
Fixes part of #873 

```
The view commits option is not available when merge group option
is checked.

Let's enable the view option when merge group, as the zoom in 
function is already working with merged ramps. Let's also fix
the regression from #1106 that removes the github icon when
merge group and filterGroupSelection is 'groupByRepos'.
```

PR #1106, opened 2020-02-29
Refactor the structure of summary-charts__title

jamessspanggg:
Discovered in #1100.

```
The summary-charts__title div is separated in two, each for when
filterGroupSelection is 'groupByRepos' and 'groupByAuthors' 
respectively.

There is an overlapping in code among the two, and adding a new 
common component in summary-charts__title needs to be done
twice, which is redundant.

Let's refactor the structure summary-charts__title div to share the
common divs/components. In this way, adding new components
would be easier.
```


frontend/src/summary.pug
```diff
@@ -90,34 +90,32 @@
...
         .summary-charts__title--contribution.mini-font(
           title="Total contribution of group"
         ) [{{ getGroupTotalContribution(repo) }} lines]
         template(v-if="isMergeGroup")
-          a(
-            v-bind:href="getAuthorProfileLink(repo[0].name)", target="_blank",
-            title="click to view author's profile"
-          )
-            i.icon-button.fab.fa-github
+          template(v-if="filterGroupSelection === 'groupByRepos'")
```
ccyccyccy: Maybe can put the `v-if` inside the `a` tag, getting rid of the `template`

frontend/src/summary.pug
```diff
@@ -90,34 +90,32 @@
...
+            a(
+              v-bind:href="getRepoLink(repo[0])", target="_blank",
+              title="click to view group's repo"
+            )
+              i.icon-button.fab.fa-github
+          template(v-else-if="filterGroupSelection === 'groupByAuthors'")
```
ccyccyccy: Same here

frontend/src/summary.pug
```diff
@@ -90,30 +90,27 @@
       .error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}
   #summary-charts
     .summary-charts(v-for="(repo, i) in filteredRepos")
-      .summary-charts__title(v-if="filterGroupSelection === 'groupByRepos'")
+      .summary-charts__title(
+        v-if="filterGroupSelection === 'groupByRepos' || filterGroupSelection === 'groupByAuthors'"
```
0blivious: maybe change to !== groupByNone?

frontend/src/summary.pug
```diff
@@ -90,30 +90,25 @@
       .error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}
   #summary-charts
     .summary-charts(v-for="(repo, i) in filteredRepos")
-      .summary-charts__title(v-if="filterGroupSelection === 'groupByRepos'")
+      .summary-charts__title(v-if="filterGroupSelection !== 'groupByNone'")
         .summary-charts__title--index {{ i+1 }}
-        .summary-charts__title--repo {{ repo[0].repoName }}
+        .summary-charts__title--repo
```
anubh-v: Just want to clarify, is there a reason we do not name this `.summary-charts__title--group` as it can be used for both grouping by authors or repos?

jamessspanggg: I don't think so. I agree `...--group` is a better name. Will change accordingly.

frontend/src/static/css/v_summary.scss
```diff
@@ -96,7 +96,7 @@
       vertical-align: middle;
     }
 
-    &--repo {
+    &--group {
```
Tejas2805: Will `group-name` be better ? Will be more specific I feel, and easier for future developers to understand ?

eugenepeh: agree, if its a name component then it should be `groupname`, otherwise may get misinterpreted as other things e.g. group container

Issue #795, opened 2019-07-08
Projectify frontend part


fzdy1914:
The way we use to develop frontend now is more like an attached component. They are less organized and formal. 

As the frontend part growing bigger, such organizations will make it harder to develop. Newcomers will also need more time to familiarize them with the code base.

Suggestion: Projectify frontend part to a formal project like `vue-cli 3` or some else.

Warning: This going to be a huge change to our frontend, which will potentially stall the front end development for some time.

Potential drawbacks: 
1. The framework itself may be too heavy for us.
2. It needs some time to test and integrate.

Potential benefits:
1. A hot re-load frontend development.
2. It may be easier to write tests.
3. The organization will be better.

fzdy1914:
@damithc @ongspxm @yamidark @yong24s 
Your inputs are needed.

damithc:
I have no objections. At least we can discuss and come up with a rough solution. A timeline depends on the availability of devs willing to take on the task.

eugenepeh:
>Suggestion: Projectify frontend part to a formal project like vue-cli 3 or some else.

Does that mean your proposition is to separate to front end off and form its own repository?

fzdy1914:
> Does that mean that your proposition is to separate to front end off and form its own repository?

I mean to have a project named frontend nested within our repo directory. It is not necessary to form a separate repo as we have PRs which will modify them together.

What I mean for projectify, is more like re-construct our codes. For now, the structure is less organized. The CSS and js do not stay together with the templates, making it harder to locate and change the code. 

To projectify, we abstract each component to one `.vue` file, which only contains the `.pug`, `.js`, `CSS` code related to this component. For example, if we want to change the style of a component, instead of doing a `control - f` in our around 1k CSS file, we can locate them easier in the `.vue` of this component. 

Meanwhile, we may also need to extract more component from our code(v_segment is a good example).

Tejas2805:
Labelling as epic as quite a few issues need to be solved to solve this issue.

gerhean:
Done in #1412

PR #1104, opened 2020-02-28
Refactor to Panels css

Tejas2805:
```
The JS/HTML code of each component has already rested in its 
own file, but all the CSS are in one file. 

Let's refactor the common components shared across panels to 
a common file for scalability and readability.
```

PR #1112, opened 2020-03-03
[#795] [For Discussion] projectify frontend

### LINKED ISSUE #719: Separate direct and indirect methods in javascripts
Methods that are use by the html are mixed with the indirect dependent methods. We should separate them for better code quality.

ang-zeyu:
To flesh out issues before going any further with this, and get feedback on the desired frontend workflow.

#795

---

Problems encountered thus far:

1)There is a heavy reliance on `window.xxx` functions ( #719 ). While this does not cause any issues with the migration, it pollutes the SFCs with global code.
  - Solution 1: Extract all such functions to a separate file. This is what was proposed in #719, but causes scope pollution, and defeats the purpose of SFCs.
  - Solution 2 ( preferred ): Many of these `window` methods are only being used by a single component, although they serve a utility like nature. We could move these methods into components' `methods`, separating only the ones that are used in multiple places into a separate utility file. (e.g. `window.toggleNext` ) This could be done during or before ( preferably ) projectifying it

2)`--view` option uses `net.freeutils.httpserver.HTTPServer` currently, which cant be used for the purposes of hot reload.
  - Nonetheless, since this is a user functionality, I suggest we preserve this, but also -
  - Maintain the state of this PR ( that is, you manually select `archive.zip` - see testing instructions )
    - This is crude, but allows you to quickly switch between built repos on the fly for testing
  - Hence, the frontend development workflow would be:
    1. `gradlew run -Dargs="... without --view"` * how many times for however many repos you want
    2.  `npm run dev`, then manually select `archive.zip` from any of the outputs. Make changes to frontend code and immediately observe changes.
        - To switch repos, simply refresh the page
    3. At the same time, provide a cli option to `npm run dev` allowing specifying the build directory to source for the `json` files if selecting `archive.zip` is not desired. This is useful when refresh-dependent features like the hashes have to be tested.
  - From the user standpoint, nothing would change. `net.freeutils.httpserver.HTTPServer` is still used to serve the build directory

3)Currently, the stylesheets are heavily reliant on child selectors ( nesting in scss ). It would be good if we could ensure that there are no side effects because of this across component scss files.

e.g.
```
// file 1
authorship {
  xx { - that also applies to some element inside the **segment** component
    ...
  }
}
```

This would be an issue if we want to use scoped styles in single file components, as child selectors aren't applied to child components ( by default ). It would be good if these styles could be completely side effect free

---

**Testing instructions:** (works enough for discussion, if needed)
`npm run dev`, and select your `archive.zip`
Make changes to frontend code and observe the changes


jamessspanggg:
hi @ang-zeyu, we are actually planning to do this incrementally. There are a couple existing PRs that are extracting components to its own file: #1059, #1104. Reason for this is because it is easier to review and detect regressions for smaller incremental changes. The effect on the existing PRs (in terms of merge conflict) will be lower if we do it incrementally as well.

Perhaps we can put this PR on hold first, then when we are done with the PRs that extract each component to its own file, we can proceed with this?

Also, it would be good if you can seek for permission before working on the issues, because some issues might not be relevant or contains more information that is not stated within the issue. Hope you can understand!

ang-zeyu:
> hi @ang-zeyu, we are actually planning to do this incrementally. There are a couple existing PRs that are extracting components to its own file: #1059, #1104. Reason for this is because it is easier to review and detect regressions for smaller incremental changes. The effect on the existing PRs (in terms of merge conflict) will be lower if we do it incrementally as well.
> 
> Perhaps we can put this PR on hold first, then when we are done with the PRs that extract each component to its own file, we can proceed with this?
> 
> Also, it would be good if you can seek for permission before working on the issues, because some issues might not be relevant or contains more information that is not stated within the issue. Hope you can understand!

Thanks for the info! I understand and am not expecting this to be merged anytime soon. ( working on a couple of other PRs that will help with this incrementally as well. 😄 )

This PR is currently only intended to flesh out more issues with the migration that could occur, besides 'componenizing' the files.

In summary so far,
- #719 could be rediscussed then resolved ( window methods -> component methods ) 
- how the desired front end workflow should be like in the end ( see the original post )
- Another thing I realised not mentioned in the OP was the issue of scoped styles. Currently, the stylesheets is heavily reliant on child selectors ( nesting in scss ). It would be good if we could ensure that there are no side effects because of this across component scss files.

e.g.
```
// file 1
authorship {
  xx { - that also applies to some element inside the **segment** component
    ...
  }
}
```

This would be an issue if we want to use scoped styles in single file components, as child selectors aren't applied to child components ( by default ). It would be good if these styles could be completely side effect free

jamessspanggg:
@fzdy1914 @eugenepeh need your inputs here.

fzdy1914:
That's great to see someone got to start on this issue. :) I previously decided to choose this issue as my FYP candidate in next semester, seems that I need to change a topic for now. :)

For the first question, the ways preferred are to either move it to one place first then put it back to its own component or put them to its own component first then do the projectify. Both are OK. The previous one I mentioned is a little bit more preferred.

For the second question, `npm run dev` can be used for temporary use. The desired workflow will have a Gradle comment that can handles both build and view. (Like move the archieve.zip to the build directory so we do not need to manually select, something like this).

For the third question, further refactor can be done after the main PR since it may be hard to predict the effect before the refactoring.

ang-zeyu:
> That's great to see someone got to start on this issue. :) I previously decided to choose this issue as my FYP candidate in next semester, seems that I need to change a topic for now. :)

It's still very much a WIP, I don't mind if anyone wants to pick this up eventually 😆
I'd be glad to finish it up otherwise though.

> For the first question, the ways preferred are to either move it to one place first then put it back to its own component or put them to its own component first then do the projectify. Both are OK. The previous one I mentioned is a little bit more preferred.

I think it may be better to chuck it into the component as far as component-specific logic goes, if this isn't high priority, so the migration is cleaner and is easier to review. ( no need to compare diffs for global methods )

> For the second question, `npm run dev` can be used for temporary use. The desired workflow will have a Gradle comment that can handles both build and view. (Like move the archieve.zip to the build directory so we do not need to manually select, something like this).

The gradle plugin should be able to push the directory as a cli option into the npm command.

> For the third question, further refactor can be done after the main PR since it may be hard to predict the effect before the refactoring.

👍 That's one option as well, and may be better since having hot reload will make testing stylesheet refactors a lot easier.

fzdy1914:
I have changed my FYP topic, so you can just go on to finish this PR step by step if you are willing to.

Reviewing this PR may take a long time, so you may need to merge a lot of conflict during the review. You may do the pre-steps first and leave the PR here and make a final conflict merge at last. 

@reposense/stage0-reviewers Please help fetch this PR locally and run it to see if there are any regressions so we can find the problem early.

Tejas2805:
@ang-zeyu did you have something like this in mind to remove nesting?

Screen Shot 2020-03-11 at 3 57 51 AM

ang-zeyu:
> @ang-zeyu did you have something like this in mind to remove nesting?

Not really, what I meant was migrating to vue-loader's [scoped styles](https://vue-loader.vuejs.org/guide/scoped-css.html#scoped-css), which makes the styles absolutely 'per-component'.

As @fzdy1914 mentioned though, doing it after the migration is one option as well, and may be better since having hot reload can make testing style refactors easier.
Before that, we could simply move each `.scss` file into the respective component file, but not scope it, so that its still applied globally.

> Please help fetch this PR locally and run it to see if there are any regressions so we can find the problem early.

The major problem right now is the mix of global functions inside existing `componentXx.js` files;
Until this is dealt with ( which makes the migration very difficult to review if done during the migration itself ), there's not much point in reviewing this!

I'll work on other components of the migration like preserving existing style check rules, integration with the `--view` option that are independent of this in the meantime though.

ang-zeyu:
After some more work on the `--view` option, I realised `npm` isn't a dependency for RepoSense, which means we can't use the same hot reload server in the final jar.
I've added the ability to specify a starting directory to the npm command, which works fine.

A separate option is needed to differentiate the server to start with, something like `--view-dev`.
The issue is that this would also be exposed to the user, but is something we can opt out of documenting in the user guide.

An alternative would of course to be just to run the `npm run dev` and `./gradlew run ...` commands separately during development.

What do you think? @jamessspanggg @fzdy1914

jamessspanggg:
> A separate option is needed to differentiate the server to start with, something like `--view-dev`.
> The issue is that this would also be exposed to the user, but is something we can opt out of documenting in the user guide.

I'm leaning more towards this as everything remains in one command.

fzdy1914:
> After some more work on the `--view` option, I realised `npm` isn't a dependency for RepoSense, which means we can't use the same hot reload server in the final jar.
> I've added the ability to specify a starting directory to the npm command, which works fine.
> 
> A separate option is needed to differentiate the server to start with, something like `--view-dev`.
> The issue is that this would also be exposed to the user, but is something we can opt out of documenting in the user guide.
> 
> An alternative would of course to be just to run the `npm run dev` and `./gradlew run ...` commands separately during development.
> 
> What do you think? @jamessspanggg @fzdy1914

As mentioned, this feature can be implemented after the main PR. For huge refactoring like this, it is better to do it step by step. No need to do it at one step.

ang-zeyu:
Thanks for the feedback!
I'll go with just `npm run dev + <path argument>` first then, the `--view-dev` option would be dealt with later on

ang-zeyu:
Closing in favour of #1412

PR #1114, opened 2020-03-04
Refactor ramp and zoom globals

ang-zeyu:
```
Refactor ramp and zoom globals

The ramp and zoom components have some component specific methods and
variables that can be moved inside the component definition.

Let’s do so, encapsulating component logic within the component itself.
```

ps: Travis is passing but the status isn't being reported back, have tried rebuilding quite a few times

### DIFF
--- a/frontend/src/ramp.pug
+++ b/frontend/src/ramp.pug
@@ -3,7 +3,7 @@
     template(v-for="(slice, j) in user.commits")
       a.ramp__slice(
         draggable="false",
-        onclick="rampClick(event);",
+        v-on:click="rampClick",
         v-for="(commit, k) in slice.commitResults",
         v-if="commit.insertions>0",
         v-bind:href="getLink(user, commit)", target="_blank",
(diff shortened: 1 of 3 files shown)


ang-zeyu:
> LGTM. Perhaps we can also refactor the `v_summary` and `v_authorship` globals here as well? But it is fine to split into few PRs as well

Hmm quite a few PRs are touching on the summary component at the moment; As for authorship, I’m thinking of moving window.toggleNext into the component definition itself ( likewise for segment.js ) to decrease reliance on direct dom manipulation. If that’s fine I’ll include it here as well!

jamessspanggg:
> Hmm quite a few PRs are touching on the summary component at the moment; As for authorship, I’m thinking of moving window.toggleNext into the component definition itself ( likewise for segment.js ) to decrease reliance on direct dom manipulation. If that’s fine I’ll include it here as well!

In that case, then let's move it to another PR!

fzdy1914:
@ang-zeyu Do mind the weird Chinese character `’` in your commit message comparing with the correct `'`.

ang-zeyu:
> @ang-zeyu Do mind the weird Chinese character `’` in your commit message comparing with the correct `'`.

My bad I wrote my commit messages in microsoft word 😅 Will change it for the other PRs

PR #1117, opened 2020-03-05
[#1083] RepoConfiguration: remove `annotationOverwrite` flag

### LINKED ISSUE #1083: RepoConfiguration: remove `annotationOverwrite` flag
The `RepoConfiguration` class has a `annotationOverwrite` attribute that controls whether the `@@author` tags are applied / ignored.

However, this attribute is always true.
Do we plan to have any cases where `@@author` tags are ignored?
If not, would it be preferable to remove `annotationOverwrite` attrbute to simplify the `RepoConfiguration` class?

Relevant code snippets:
https://github.com/reposense/RepoSense/blob/5cd4123966a51b1916858b08bc45d12014ad8157/src/main/java/reposense/model/RepoConfiguration.java#L32

The setter method below is not used anywhere.
https://github.com/reposense/RepoSense/blob/5cd4123966a51b1916858b08bc45d12014ad8157/src/main/java/reposense/model/RepoConfiguration.java#L331-L333

The `annotationOverwite` property is checked in `FileInfoAnalyzer`.
The check always passes since `annotationOverwrite` property is always true.

https://github.com/reposense/RepoSense/blob/5cd4123966a51b1916858b08bc45d12014ad8157/src/main/java/reposense/authorship/FileInfoAnalyzer.java#L63-L65

niqiukun:
Fixes #1083 

```
There is an `annotationOverwrite` flag in the RepoConfiguration. 
The flag is redundant because it defaults to `true` and is never 
ultilised in the main code.

Since there is currently no planned use case, for better code 
quality, we should remove the unnecessary code.

Let's remove this flag from the RepoConfiguration.
```


src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -57,10 +56,6 @@ public static FileResult analyzeFile(RepoConfiguration config, FileInfo fileInfo
         aggregateBlameAuthorInfo(config, fileInfo);
         fileInfo.setFileType(config.getFileType(fileInfo.getPath()));
 
-        if (config.isAnnotationOverwrite()) {
-            AnnotatorAnalyzer.aggregateAnnotationAuthorInfo(fileInfo, config.getAuthorEmailsAndAliasesMap());
```
anubh-v: @niqiukun, thanks for the PR!
Just want to highlight that there is no need to remove the call to `aggregateAnnotationAuthorInfo` (line 61).
We still need to call `aggregateAnnotationAuthorInfo` for parsing any `@@author` tags provided by users (you may read about this feature [here](https://github.com/reposense/RepoSense/blob/master/docs/UserGuide.md#provide-data-using-author-tags))

The original issue was simply that the `RepoConfiguration` class no longer needs the `isAnnotationOverwrite` attribute.
It is sufficient to remove  the`isAnnotationOverwrite` attribute and the `isAnnotationOverwrite` / `setAnnotationOverwrite` methods.

niqiukun: Thanks for the comment! Sorry I forgot that the default value for `isAnnotationOverwrite` is `true`. I have fixed this.

anubh-v:
The PR looks good now
@niqiukun could you also help to provide a commit message for the PR, following the guidelines [here](https://oss-generic.github.io/process/docs/FormatsAndConventions.html#commit-message)?

fzdy1914:
Can you help me update the branch?

PR #1127, opened 2020-03-08
[#850] Ignore git submodules when analyzing repos

### LINKED ISSUE #850: Fail to analyse repos which contain submodule
[Git submodule](https://git-scm.com/docs/git-submodule) is a tool which will be commonly used in large projects. 

When trying to analyze a repo which contains submodules, the following error occurs.

image

ccyccyccy:
Fixes #850 

```
Repos that contains git submodules are not handled properly and cannot
be analyzed.

This is due to the diff result including the submodule information.

Let set --ignore-submodules=all in the git diff command to ignore all
submodule information.
```

### DIFF
--- a/src/test/java/reposense/git/GitDiffTest.java
+++ b/src/test/java/reposense/git/GitDiffTest.java
@@ -31,4 +31,11 @@ public class GitDiffTest extends GitTestTemplate {
     public void diffCommit_nonexistentCommitHash_throwsRunTimeException() {
         GitDiff.diffCommit(config.getRepoRoot(), NONEXISTENT_COMMIT_HASH);
     }
+
+    @Test
+    public void diffCommit_commitContainingSubmodule_ignoresSubmodule() {
+        GitCheckout.checkout(config.getRepoRoot(), "850-GitDiffTest-commitWithSubmodule_success");
+        String diffResult = GitDiff.diffCommit(config.getRepoRoot(), LATEST_COMMIT_HASH);
+        Assert.assertFalse(diffResult.contains("Subproject commit e41deb84c36c2430caf742abdbd96cc2dc4a09ca"));
+    }
 }
--- a/src/test/java/reposense/git/GitDiffTest.java
+++ b/src/test/java/reposense/git/GitDiffTest.java
@@ -34,8 +34,9 @@ public class GitDiffTest extends GitTestTemplate {
 
     @Test
     public void diffCommit_commitContainingSubmodule_ignoresSubmodule() {
-        GitCheckout.checkout(config.getRepoRoot(), "850-GitDiffTest-commitWithSubmodule_success");
-        String diffResult = GitDiff.diffCommit(config.getRepoRoot(), LATEST_COMMIT_HASH);
-        Assert.assertFalse(diffResult.contains("Subproject commit e41deb84c36c2430caf742abdbd96cc2dc4a09ca"));
+        GitCheckout.checkout(config.getRepoRoot(),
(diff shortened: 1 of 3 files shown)


anubh-v:
Should we add a test for this?

ccyccyccy:
> Should we add a test for this?
> Perhaps one of the repos used for system tests can include a submodule, and we can check the output's correctness.

Yes I think we should.
I will create a new repo to be used as a submodule in testrepo alpha

ccyccyccy:
Actually since this pull request only affects the `GitDiff` class, I think we should only do unit test for  `GitDiff` instead of a system test.

src/test/java/reposense/git/GitDiffTest.java
```diff
@@ -31,4 +31,11 @@ public void diffCommit_latestCommitHash_emptyResult() {
     public void diffCommit_nonexistentCommitHash_throwsRunTimeException() {
         GitDiff.diffCommit(config.getRepoRoot(), NONEXISTENT_COMMIT_HASH);
     }
+
+    @Test
+    public void diffCommit_commitContainingSubmodule_ignoresSubmodule() {
```
eugenepeh: please keep the branch name the same as method name
E.g. IssueNumber-JavaClass-MethodName

src/test/java/reposense/git/GitDiffTest.java
```diff
@@ -31,4 +31,11 @@ public void diffCommit_latestCommitHash_emptyResult() {
...
+        GitCheckout.checkout(config.getRepoRoot(), "850-GitDiffTest-commitWithSubmodule_success");
+        String diffResult = GitDiff.diffCommit(config.getRepoRoot(), LATEST_COMMIT_HASH);
```
eugenepeh: not so ideal to diff it with `LATEST_COMMIT_HASH` because it does not make sense in this context - the latest hash is not the latest.

Perhaps use the id of the empty tree `4b825dc642cb6eb9a060e54bf8d69288fbee4904` to get the diff from initial till to-date

src/test/java/reposense/git/GitDiffTest.java
```diff
@@ -31,4 +31,11 @@ public void diffCommit_latestCommitHash_emptyResult() {
...
+        Assert.assertFalse(diffResult.contains("Subproject commit e41deb84c36c2430caf742abdbd96cc2dc4a09ca"));
```
eugenepeh: perhaps just `Subproject commit`?

src/test/java/reposense/git/GitDiffTest.java
ccyccyccy: I changed the name of the branch the branch pushed up.
However, the branches are protected and I could not delete the old branch, as well as another branch i accidentally pushed up with a typo.
Can you give me permission to delete these branches?

eugenepeh: all the branches are protected by default, hence only the admins can delete them.
Can just let any of the admins know `@reposense/admins` and we will delete them for you

ccyccyccy: Thank for the information :)

ccyccyccy:
@reposense/admins Need help deleting the branches `850-GitDiffTest-commitWithSubmodule_success` and `850-GitDIffTest-diffCommit_commitContainingSubmodule_ignoresSubmodule` (Note the capital `I` in `850-GitDIffTest`)

ccyccyccy:
@eugenepeh Thanks for the review! I have modified all the code as requested.

fzdy1914:
Can you help retrigger the build?

PR #1136, opened 2020-03-11
[#1137][#1138] Expand pug-lint glob and remove fix option in lint

### LINKED ISSUE #1137: Npm lint fails in windows command shell
Untitled

`pug-lint frontend/src/*.pug frontend/src/tabs/*.pug` fails in windows as pug-lint does not expand globs, instead relying on bash wildcard expansion to do so.

### LINKED ISSUE #1138: eslint in npm lint script malformed and uses the --fix option
`eslint -- --fix frontend/src/**/*js frontend/cypress/**.js`

causes `--fix` to be treated as a file ( tested ) and auto-fixable errors to be unfixed.

The main `lint` script should not have `--fix` either so that the errors get reported in CI.

I've searched the eslint documentation as well, and could not find mention of an empty `--` option anywhere. Was this a typo in https://github.com/reposense/RepoSense/pull/803#discussion_r304482267?

ang-zeyu:
Fixes #1137 
Fixes #1138

```
Expand pug-lint glob and remove fix option in lint

The pug-lint command does not expand globs provided as arguments,
relying instead on bash wildcard expansion. This causes the pug-lint
command to fail on windows.
The main linting script was also written with the --fix option, which
can increase the build time for CI.
This may also cause auto-fixable errors to go undetected, although the
lint script has an extra ‘--' that causes --fix to be recognised as a
file.

Let’s manually expand the globs for the pug-lint command, moving it to
a separate script for brevity.
Let’s also remove the --fix option used by the main linting script,
and move it to a separate lintfix script.
At the same time, let’s introduce --fix for stylelint as well.
```

### DIFF
--- a/docs/DeveloperGuide.md
+++ b/docs/DeveloperGuide.md
@@ -66,7 +66,7 @@ This project follows [oss-generic coding standards](https://oss-generic.github.i
 Optionally, you can follow the [Using Checkstyle](UsingCheckstyle.md) document to configure *Intellij* to check style-compliance as you write code.
 
 ### Configuring the JavaScript and CSS coding style
-Our project follows the [Airbnb Javascript Style Guide](https://github.com/airbnb/javascript) and [OSS CSS Coding Standard](https://oss-generic.github.io/process/codingStandards/CodingStandard-Css.html), which is governed by the Eslint and Stylelint respectively. We have our own customized style checker for the Pug files, which is governed by pug-lint. Their configuration files can be found at the root of the project. Please run a `npm run lint` from the project root directory and fix all of the lint errors before committing your code for final review.
+Our project follows the [Airbnb Javascript Style Guide](https://github.com/airbnb/javascript) and [OSS CSS Coding Standard](https://oss-generic.github.io/process/codingStandards/CodingStandard-Css.html), which is governed by the Eslint and Stylelint respectively. We have our own customized style checker for the Pug files, which is governed by pug-lint. Their configuration files can be found at the root of the project. Please run a `npm run lint` from the project root directory and fix all of the lint errors before committing your code for final review.  You can also use the `npm run lintfix` script to automatically fix some of the javascript and css lint errors!
 
 Eslint, Stylelint and their accompanying modules can be installed through NPM, so do ensure that you got it [installed](https://www.npmjs.com/get-npm) if you are working on the report.
 
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
     "doc": "docs"
   },
   "scripts": {
-    "lint": "eslint -- --fix frontend/src/**/*js frontend/cypress/**.js && stylelint frontend/src/**/*.scss && pug-lint frontend/src/*.pug frontend/src/tabs/*.pug",
+    "lint": "eslint frontend/src/**/*js frontend/cypress/**.js && stylelint frontend/src/**/*.scss && npm run puglint",
+    "puglint": "pug-lint frontend/src/index.pug frontend/src/ramp.pug frontend/src/summary.pug frontend/src/tabs/authorship.pug frontend/src/tabs/segment.pug frontend/src/tabs/zoom.pug",
+    "lintfix": "eslint --fix frontend/src/**/*js frontend/cypress/**.js && stylelint --fix frontend/src/**/*.scss",
     "browserify": "browserify -t vueify -e frontend/src/static/js/v_authorship.js -o frontend/build/static/js/v_authorship.js",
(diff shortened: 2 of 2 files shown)


ang-zeyu:
Fixed a typo in commit message

jamessspanggg:
@ccyccyccy @anubh-v  @0blivious can y'all help verify if the pug-lint works on Windows?

ccyccyccy:
Maybe instead of expanding the glob manually, we can do something like [this](https://github.com/pugjs/pug-lint/issues/144#issuecomment-453518170)?
It requires an npm library though...

ccyccyccy:
> 
> 
> @ccyccyccy @anubh-v @0blivious can y'all help verify if the pug-lint works on Windows?

Tested and works on Windows.

ang-zeyu:
> Tested and works on Windows.

What shell was this?

> Maybe instead of expanding the glob manually, we can do something like [this](https://github.com/pugjs/pug-lint/issues/144#issuecomment-453518170)?
> It requires an npm library though...

That’s one option as well; I opted for manually expanding it for now since after #795 pug-lint would be replaced by something like pug-lint-vue ( supports specifiying a directory ), and the number of pug files is small.
It’s still quite a while away though, I’m open to this!

ccyccyccy:
> What shell was this?

Tested on both CMD and WSL

> That’s one option as well; I opted for manually expanding it for now since after #795 pug-lint would be replaced by something like pug-lint-vue ( supports specifiying a directory ), and the number of pug files is small.

I think it's fine for now, considering that.

Tejas2805:
Can you update the commit message to wrap at 72 characters?

ang-zeyu:
> Can you update the commit message to wrap at 72 characters?

Missed this out, thanks!

Tejas2805:
Just a doubt. How are we ensuring that `lintfix` runs too? Does `npm run lint` ensure that?

ang-zeyu:
We don't want to ensure it since it would obscure errors in CI.
`lintfix` is a separate script just for fixing auto-fixable style errors

Tejas2805:
So this means the developer/contributor needs to run it if I have understood this correctly?

ang-zeyu:
> So this means the developer/contributor needs to run it if I have understood this correctly?

yup its meant to be only run on your local machine.
On this note unfortunately pug-lint dosen't have a `--fix` equivalent 😧

Tejas2805:
> yup its meant to be only run on your local machine.
> On this note unfortunately pug-lint dosen't have a `--fix` equivalent 😧

Then I think its better to add this in the DG. Otherwise, the new contributors/developers might not know about it and not be able to use the benefits.

ang-zeyu:
> Then I think its better to add this in the DG. Otherwise, the new contributors/developers might not know about it and not be able to use the benefits.

Good point, updated!

Noticed there isn't any info on the usage of pug-lint; Should we include that in this/another PR as well?

Tejas2805:
> Noticed there isn't any info on the usage of pug-lint; Should we include that in this/another PR (#1149 maybe?) as well?

Ah yes. I seem to have missed that my in my initial PR. Better to do it in PR #1149 since it is already updating the docs.

docs/DeveloperGuide.md
```diff
@@ -66,7 +66,7 @@ This project follows [oss-generic coding standards](https://oss-generic.github.i
...
 
 ### Configuring the JavaScript and CSS coding style
-Our project follows the [Airbnb Javascript Style Guide](https://github.com/airbnb/javascript) and [OSS CSS Coding Standard](https://oss-generic.github.io/process/codingStandards/CodingStandard-Css.html), which is governed by the Eslint and Stylelint respectively. Their configuration files can be found at the root of the project. Please run a `npm run lint` from the project root directory and fix all of the lint errors before committing your code for final review.
+Our project follows the [Airbnb Javascript Style Guide](https://github.com/airbnb/javascript) and [OSS CSS Coding Standard](https://oss-generic.github.io/process/codingStandards/CodingStandard-Css.html), which is governed by the Eslint and Stylelint respectively. Their configuration files can be found at the root of the project. Please run a `npm run lint` from the project root directory and fix all of the lint errors before committing your code for final review. You can also use the `npm run lintfix` script to automatically fix some javascript and css lint errors!
```
Tejas2805: some **of the** javascript and css

ang-zeyu: Fixed 👍

ang-zeyu:
@fzdy1914 could I get your review here?
Not too sure why codecov increased, since there aren't any code changes here 😮

PR #1148, opened 2020-03-15
[#1147] Generate random filetype colour when predefined colours are used up

### LINKED ISSUE #1147: Colour repetition in breakdown by filetype
When many file types are provided, colors begin to repeat themselves. 

Screen Shot 2020-03-15 at 4 48 26 AM

[Link to reproduce](https://nus-cs3281.github.io/2020-dashboard/#search=&sort=groupTitle&sortWithin=title&timeframe=commit&mergegroup=false&groupSelect=groupByAuthors&breakdown=true&since=2019-12-01&tabOpen=false)

jamessspanggg:
Fixes #1147 

```
The filetype colours are chosen through a rotation of the predefined 
colour values. So if the number of filetypes exceeds the number of 
predefined colours, there will be repetitions of the same colour.

The repeated filetypes may be hard to distinguish among one
another, in terms of their colours.

Let's generate a random colour for the filetypes after the standard
selection of predefined colours have been exhausted.
```


frontend/src/static/js/v_summary.js
```diff
@@ -516,8 +525,12 @@ window.vSummary = {
         repo.users.forEach((user) => {
           Object.keys(user.fileTypeContribution).forEach((fileType) => {
             if (!Object.prototype.hasOwnProperty.call(fileTypeColors, fileType)) {
-              fileTypeColors[fileType] = selectedColors[i];
-              i = (i + 1) % selectedColors.length;
+              if (i < selectedColors.length) {
+                fileTypeColors[fileType] = selectedColors[i];
+                i += 1;
```
jamessspanggg: note that eslint does not allow `i++`, if not I could have make this a one liner:

`fileTypeColors[fileType] = i < selectedColors.length ? selectedColors[i++] : this.getRandomColor();

frontend/src/static/js/v_summary.js
```diff
@@ -100,6 +100,8 @@ function dateRounding(datestr, roundDown) {
   return getDateStr(datems);
 }
 
+const hexLetters = '0123456789ABCDEF'.split('');
```
jamessspanggg: this is placed globally so that we only need to split once

jamessspanggg:
Can't view the difference in the netlify preview as it does not exhaust the predefined colors, but the following is a screenshot of the randomly generated colors:

Screen Shot 2020-03-15 at 1 31 07 PM

Note that the randomly generated colours will be different at every reload.

frontend/src/static/js/v_summary.js
```diff
@@ -505,6 +507,13 @@ window.vSummary = {
         merged[key] += value;
       });
     },
+    getRandomColor() {
+      let color = '#';
+      while (color.length < 7) {
+        color += hexLetters[Math.round(Math.random() * 15)];
+      }
+      return color;
+    },
```
0blivious: can use this one line code to generate random color since ffffff == 16777215
return '#' + Math.round(Math.random()*16777215).toString(16)

0blivious:
For the random color maybe we should exclude ffffff and 000000 as black is reserved for the `all` label and white is not visible

jamessspanggg:
> For the random color maybe we should exclude ffffff and 000000 as black is reserved for the `all` label and white is not visible

Not sure if worth the change as chances of getting `#fffff` or `#000000` is very low.

Added the change anyway, will revert if not needed.

jamessspanggg:
> I just want to clarify something before I approve. Have we made sure that colours don't get repeated? Or it's a one-off chance that it might happen?

It may happen but with a very very very slim chance.

> It is fine to use randomized colors after the standard selection of colors have been exhausted. There is a range of usage in which the software works optimally. Allowing to exceed that range is already a bonus (instead of limiting the number of categories, in this case); we don't have to provide an optimal experience beyond that range.

Tejas2805:
> It is fine to use randomized colors after the standard selection of colors have been exhausted. There is a range of usage in which the software works optimally. Allowing to exceed that range is already a bonus (instead of limiting the number of categories, in this case); we don't have to provide an optimal experience beyond that range.

Agreed. Just wanted to clarify since I didn't get that one-liner in the code.

frontend/src/static/js/v_summary.js
```diff
@@ -379,6 +379,12 @@ window.vSummary = {
       });
     },
 
+    getRandomColor() {
+      const maxHexColorValue = 16777213;
```
anubh-v, line 383: Should this be 16777214 since the `Math.random` is not inclusive of the upper bound?

jamessspanggg: Yes u are right

frontend/src/static/js/v_summary.js
```diff
@@ -379,6 +379,12 @@ window.vSummary = {
       });
     },
 
+    getRandomColor() {
+      const maxHexColorValue = 16777214;
+      // excludes #000000 and #FFFFFF as they are reserved
+      return `#${Math.round(Math.random() * maxHexColorValue + 1).toString(16)}`;
```
eugenepeh, line 385: instead of it being purely random, such that color collision can happen, can we reuse the existing template colors at different shade? Or apply an algorithm to check that the new colors are not similar to any of the existing colors?
https://stackoverflow.com/questions/5392061/algorithm-to-check-similarity-of-colors

jamessspanggg: I modified the implementation such that it will keep generating random colours until a colour that is not similar to the existing colours. The distance used to compare colours is Delta E, which is a perceptual colour distance. The threshold that was selected is 11, which strikes a balance between perceptibility and the time required generate a colour that satisfies the threshold. The threshold is based on the table below, available [here](http://zschuessler.github.io/DeltaE/learn/):

Screen Shot 2020-04-09 at 2 16 29 PM

Also, since we can't really reproduce predefined colours being used up in our preview, I created a preview just to show how the generated colours will look like [here](http://generatenonrepeatingcolors-deltae.surge.sh/#search=&sort=groupTitle&sortWithin=title&timeframe=commit&mergegroup=false&groupSelect=groupByRepos&breakdown=true).

Note that the colours will be generated **_once for every page reload_**, so I don't think it's a big overhead to add this algorithm.

damithc:
While it is not a big problem, it is desirable for a page to always have the same set of colors in the same order no matter how many times the page is loaded i.e., a deterministic algorithm is better than a random one.

frontend/src/static/js/v_summary.js
eugenepeh: Sounds good to me. If this works well, we can consider removing the existing `selectedColors` template. 

To make the algorithm more deterministic, we can consider using a const random seed.

jamessspanggg: I have utilised an external [seeded random generator](https://github.com/davidbau/seedrandom) to do this, since javascript does not have seeding capability for `Math.random`. I have also removed the existing color templates. Here is the [updated dashboard link](http://updatedrandomcolorgenerator.surge.sh/#search=&sort=groupTitle&sortWithin=title&timeframe=commit&mergegroup=false&groupSelect=groupByRepos&breakdown=true) that shows the generated file type colors. Reloading the same page will result to the same colors.

@eugenepeh @damithc do let me know if this is ok, or anymore changes required (such as increasing the threshold so that the colors are more different than one another)

damithc:
Are we discarding the previous set of preset colors. Those were especially selected by a previous FYP student to accommodate those with vision deficiencies. Perhaps we should keep them and fallback on the random colors only after exhausting the preset colors?

jamessspanggg:
> Are we discarding the previous set of preset colors? Those were especially selected by a previous FYP student to accommodate those with vision deficiencies. Perhaps we should keep them and fallback on the random colors only after exhausting the preset colors?

Oh I didn't know that. In that case, let's keep them and use random colors after exhausting.

frontend/src/index.pug
```diff
@@ -19,6 +19,7 @@ html
     script(src="https://cdn.jsdelivr.net/npm/jszip@3.1/dist/jszip.min.js")
 
     script(src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.14/build/highlight.min.js")
+    script(src="https://cdn.jsdelivr.net/npm/seedrandom@3.0.5/seedrandom.min.js")
```
eugenepeh, line 22: possible to group the scripts from same directory together? e.g. npm, gh

jamessspanggg: sure.

jamessspanggg: done.

PR #1149, opened 2020-03-15
[#1144, #1135] Docs: update outdated info

### LINKED ISSUE #1135: Invalid Location in example for config files in DG
The example that we show for the location of config files in the Developer Guide is invalid. There are no config files in that location to run.

Screen Shot 2020-03-11 at 3 32 23 AM

**Suggestion:** Since we display it as an example, let's either update the location or add config files in that location so that the location becomes valid.

### LINKED ISSUE #1144: Change example pictures in UG
The example pictures in UG still have a grey background for the right panel. 

Let's update the pictures to  show the new white background.

Tejas2805:
Fixes #1144 and fixes #1135 

```
There is a lot of outdated info in the User Guide and Developer Guide.

Let's update the UG and DG to show the current state of RepoSense
as this will help them better in using/developing RepoSense.
Let's also remove the warnings reported by Travis by removing the
trailing whitespaces. 
```


Tejas2805:
> Diagram for code panel?

@jamessspanggg Code Panel doesn't have a picture in the current UG. Want me to add one?

Tejas2805:
> Perhaps can update this to ./config as well?
https://github.com/reposense/RepoSense/pull/1149/files#diff-572f9bedcb201b96c74241fb42e29fcfR311-R312

Didn't get what you meant by this. Can clarify a bit?

jamessspanggg:
> > Perhaps can update this to ./config as well?
> > https://github.com/reposense/RepoSense/pull/1149/files#diff-572f9bedcb201b96c74241fb42e29fcfR311-R312
> 
> Didn't get what you meant by this. Can clarify a bit?

Line 311 in UG, can update the path to `./config` instead of `./my_configs`?

> @jamessspanggg Code Panel doesn't have a picture in the current UG. Want me to add one?
Yes I think we should put a diagram for each of the major component, including code panel

fzdy1914:
@jamessspanggg Can you investigate why report preview here is not successfully generated?

jamessspanggg:
> @jamessspanggg Can you investigate why report preview here is not successfully generated?

It is because he moved all the config files to another folder, we will need to change the build command to reflect those changes as well.

Tejas2805:
> It is because he moved all the config files to another folder, we will need to change the build command to reflect those changes as well.

Do we have to do this, whenever file locations are changed or only when few specific file locations are changed?

fzdy1914:
I think we should change the default gradlew run command, @Tejas2805 Can you include that change in this PR also?

jamessspanggg:
> > It is because he moved all the config files to another folder, we will need to change the build command to reflect those changes as well.
> 
> Do we have to do this, whenever file locations are changed or only when few specific file locations are changed?

Yes, because for deployment, we would need to run with the config files of RepoSense,  if ur path changed already of course we must also change the path specific in the run command

fzdy1914:
@Tejas2805 Please fix the issue I mentioned.

Tejas2805:
> @Tejas2805 Please fix the issue I mentioned.

Sorry a bit confused. Where to add exactly?

fzdy1914:
> > @Tejas2805 Please fix the issue I mentioned.
> 
> Sorry a bit confused. Where to add exactly?

The gradlew comment in the travis sript uses default config location:
image

So you should edit the following description:

https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/parser/ArgsParser.java#L137

And specify the default config folder place here:

https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/parser/ArgsParser.java#L204

fzdy1914:
An alternative will be editing the travis script only, however, it is not desired as it will still breaking our existing default `gradlew run`

Tejas2805:
@fzdy1914 Surge works now. 

@eugenepeh Once this PR gets merged, the Netlify build command needs to be updated based on the new location of the config files

fzdy1914:
Maybe we should directly disable netlify preview as the surge works fine for now.

fzdy1914:
> @fzdy1914 Surge works now.

I believe you should implement the previous method I have mentioned. You should set the config folder explicitly in the ArgsPaser to be `./config` if the path is not specified.

Tejas2805:
> I believe you should implement the previous method I have mentioned. You should set the config folder explicitly in the ArgsPaser to be `./config` if the path is not specified.

Okay, will. But will break a few of the tests, so will fix those too then.

Tejas2805:
@fzdy1914 Can I get your review on this. Have updated the default path.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -34,6 +34,7 @@
 public class ArgsParserTest {
 
     private static final Path PROJECT_DIRECTORY = Paths.get(System.getProperty("user.dir"));
+    private static final Path CONFIG_DIRECTORY = Paths.get(System.getProperty("user.dir") + "/config/");
```
jamessspanggg: would this work in windows machines? perhaps use File.separator instead?

Tejas2805: Aah sorry. Forgot to do that. Thanks for catching it!

docs/UserGuide.md
```diff
@@ -140,9 +141,10 @@ The `Tool Bar` at the top provides a set of configuration options that control t
 * `Breakdown by file type` : toggles the contribution bar to either display the bar by :
     * the total lines of codes added (if checkbox is left unchecked), or
     * a breakdown of the number of lines of codes added to each file type (if checkbox is checked). If sorting criteria is contribution, only lines contributed to selected file types will be taken into account.
+*`Merge group` : merges all the ramp charts of the same group together to form a single ramp chart.
```
eugenepeh: not properly formatted. 
Merges all the ramp charts of each group into a single ramp chart; aggregates the contribution of each group.

.travis.yml
```diff
@@ -60,7 +60,7 @@ after_success:
   - |
     if [ "$TRAVIS_JOB_NAME" = "Ubuntu 14.04 JDK 8" ];
     then
-      git fetch --all && git config --global user.email "-" && git config --global user.name "-" && ./gradlew run -Dargs="--since d1"
+      git fetch --all && git config --global user.email "-" && git config --global user.name "-" && ./gradlew run -Dargs="--since d1 --config ./config/"
```
fzdy1914: after changing the default location, this line edition is no longer needed right?

fzdy1914: @Tejas2805

Tejas2805: Doesn't make a difference, true. Will remove.

Tejas2805:
@fzdy1914 Can merge this now?

PR #1152, opened 2020-03-16
[#1150] Consistency between Code panel and Chart panel for file type selection

### LINKED ISSUE #1150: Consistency between Code panel and Chart panel for file type selection
When a user clicks on the </> icon, we can auto-set the file types of the code panel to match the file type selection of the ramp chart.

_Originally posted by @damithc in https://github.com/reposense/RepoSense/pull/1025#issuecomment-599303582_

0blivious:
Fix #1150 
```
The file type selection is always set to select all when opened at
first.

With the new sort by file type feature in Chart Panel, we can
auto-set the file types of the code panel to match the file type
selection of the ramp chart.

Let's pass the file type selection information to Code Panel.
```


jamessspanggg:
@0blivious can update the authorship file type hashes once you open the authorship tab? because now when u open the authorship tab with the transferred checked file types, the hashes are not set, and hence reloading the authorship tab will not retain the checked file types.

jamessspanggg:
when file types unchecked are unrelated to the authorship, the `all` checkbox is unchecked:
Kapture 2020-04-26 at 21 07 08

PR #1153, opened 2020-03-16
[#436] Give warning for duplicate alias

### LINKED ISSUE #436: Different author with same alias will not trigger any warning

Tejas2805:
Fixes #436 

```
Duplicate aliases in the csv files, do not trigger any warning.
As a result,the user might not get to know that there are duplicate 
aliases in the csv files provided by him/her.

Let's give a warning if there are duplicate aliases in the csv file
as the alias will belong to the last person who claims it. 
```


Tejas2805:
Despite it being a `firstTimers` issue, I have done the PR for two reasons:

1) I haven't done any backend change to RepoSense up to now. So I wished to start
with something small.
2) It had been quite a while since this issue had opened, and therefore I think it's better
to solve this.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -54,6 +56,14 @@ public void update(StandaloneConfig standaloneConfig, List<String> ignoreGlobLis
             List<String> aliases = new ArrayList<>(author.getAuthorAliases());
             List<String> emails = new ArrayList<>(author.getEmails());
             aliases.add(author.getGitId());
+            allAliases.addAll(aliases);
+
+            HashSet<String> uniqueAliasesSet = new HashSet<>(allAliases);
+            boolean hasDuplicateAliases = uniqueAliasesSet.size() != allAliases.size();
```
0blivious: why need allAliases? can just compare size of uniqueAliases with aliases

Tejas2805: aliases is just keeping a track of the aliases of a specific author and not all the aliases of all authors. This is why need allAliases.

0blivious: why not do this outside of the for loop then?

Tejas2805: Also can. Actually will be better I agree.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -54,6 +56,14 @@ public void update(StandaloneConfig standaloneConfig, List<String> ignoreGlobLis
...
+            if (hasDuplicateAliases) {
```
jamessspanggg: So once the `allAliases` has a duplicate alias, it keeps logging this warning all the way to the end?

Tejas2805: I was confused whether we want something specific to show up or a general warning. So implemented it there in case we wanted something specific. Will update.

jamessspanggg:
@fzdy1914 @eugenepeh need ur opinion here. Should the implementation for this PR be:
- a general warning warning if there exists at least one duplicate
- or display specific warning stating which name is being duplicated

fzdy1914:
> @fzdy1914 @eugenepeh need ur opinion here. Should the implementation for this PR be:
> 
> * a general warning warning if there exists at least one duplicate
> * or display specific warning stating which name is being duplicated

If an elegant way of showing all duplicates exists, it is better to show the specific warning. If it is hard to do so, just show a general warning.

Tejas2805:
@fzdy1914 How's this?

Tejas2805:
@fzdy1914 Looks like to trigger warning for duplicate aliases we might have to implement it at two different places if we want it for both - csv and standalone config. If we want just for csv, then can do accordingly too. 

As of now, it's implemented for standalone config. 

Do let me know how you want me to proeed.

fzdy1914:
My original issue only targets at the standalone config previously.  Can you check if it is also a problem in author-config.csv?

Tejas2805:
> My original issue only targets at the standalone config previously. Can you check if it is also a problem in author-config.csv?

Yes, in author-config.csv if there are multiple aliases, it will belong to the last one. And there are no warnings.

jamessspanggg:
> Yes, in author-config.csv if there are multiple aliases, it will belong to the last one. And there are no warnings.

Perhaps that could be targeted in this PR as well?

Tejas2805:

Warnings will be logged for both CSV and standalone config

anubh-v:
@Tejas2805, does the same issue apply to emails as well?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -54,6 +54,8 @@ public void update(StandaloneConfig standaloneConfig, List<String> ignoreGlobLis
             List<String> aliases = new ArrayList<>(author.getAuthorAliases());
             List<String> emails = new ArrayList<>(author.getEmails());
             aliases.add(author.getGitId());
+
+            checkDuplicateAliases(aliases);
```
anubh-v: There may be an issue here (let me know if I'm wrong)

In the method `checkDuplicateAliases`, we are checking a list of aliases against the aliases present in the *current* `authorEmailsAndAliasesMap` TreeMap of the AuthorConfiguration instance.

However, in the `update` method, we are adding the aliases to a *new* TreeMap.
i.e. this will not detect duplicate aliases that occur within the standalone configuration file?

anubh-v: I tried analysing a [test-repo with duplicate aliases in its standalone config](https://github.com/anubh-v/test-repo) using this PR. 
There do not seem to be any logged warnings when I run RepoSense:

Let me know if I made an error in testing.

Here is the [link](https://github.com/anubh-v/test-repo/blob/master/_reposense/config.json) to the standalone config in the repo.
There are 2 authors, `Anubhav` and `Bob` in the standalone config, both with the same alias.

Here is the output from running RepoSense:
image

Command used:
`./gradlew run -Dargs="--repo https://github.com/anubh-v/test-repo.git"`

Tejas2805: Yup, realized this too. Currently, fixing it in local. I will commit soon. 

I have found a quick fix for this. Trying to see if some better way possible.

Tejas2805:
> @Tejas2805, does the same issue apply to emails and git IDs as well?
> i.e. if authors provide the same email or git ID, it belongs to the last author that claimed it?
> (Though this may be impossible, compared with authors having the same alias)

For git ID, I will have to look into it. Will see and update. 
For emails, yes. It belongs to the last author who claims it.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -53,16 +54,32 @@ public void update(StandaloneConfig standaloneConfig, List<String> ignoreGlobLis
             newAuthorDisplayNameMap.put(author, author.getDisplayName());
             List<String> aliases = new ArrayList<>(author.getAuthorAliases());
             List<String> emails = new ArrayList<>(author.getEmails());
+            allAliases.addAll(aliases);
             aliases.add(author.getGitId());
+
             aliases.forEach(alias -> newAuthorEmailsAndAliasesMap.put(alias, author));
```
anubh-v: Just to clarify, is there a reason we're not doing the check for duplicates as we add the alias into the TreeMap?
The `TreeMap.put` method returns null if the key is a new key

Tejas2805: I agree with your suggestion that we can do it as we add the alias into the TreeMap. But the thing is it becomes a bit irrelevant. Cause even if the alias is duplicate, it will still have to be added, as the implementation decided is that the last person to claim the alias get it. 

Also, if I do it inside the function, then I will still have to check duplicates for CSV files. To prevent this duplication, I am doing it outside. This way I can use the same function for both.

anubh-v: > Cause even if the alias is duplicate, it will still have to be added, as the implementation decided is that the last person to claim the alias get it.

I don't quite see the issue. If we want, we can change the implementation such that we check whether each alias exists before adding it, and prevent existing aliases from being overwritten.

> Also, if I do it inside the function, then I will still have to check duplicates for CSV files. To prevent this duplication, I am doing it outside.

What if we create a new static method that accepts a TreeMap, a key and a value, and then:
- adds the key - value pair to the TreeMap
- checks if the key already existed by inspecting the return value of `TreeMap.put`
- uses the logger to issue a warning if the key already existed?

We can use this method for parsing both csv and standalone config?

Tejas2805: I am okay with changing the implementation. But I believe the senior devs and Prof in their original implementation wanted it to be overwritten. Can ask for their suggestion on this. I am not hard bent on any fixed way, but think last person should to claim alias should get it. 

@eugenepeh your suggestion for this?

anubh-v: Sorry, let me clarify.
I'm not suggesting that we *need* to change the current implementation *now*.
I'm suggesting that if we perform the check for duplicates as we add each alias into the TreeMap, then it'll be easier for us to change the handling of duplicates *if we want, in future*.

Tejas2805:
Alright, so I have looked into the issue deeply. There are three cases

1) If only standalone config run, warnings are generated correctly.
2) If only csv files run (ignore standalone), warnings are generated correctly.
3) If both are used, warnings are generated twice. First cause of csv, then due to standalone config. This happens because the same alias list is used for both. 

**Solutions:**

1) Implement specific warnings only for standalone config (which was originally in mind).

2) Implement general warnings for both csv and standalone config (only two warnings in total - specifying duplicate in standalone config or csv)

3) Implement specific warnings for both csv and standalone config, but I might have to change the backend code to some extent so that I can prevent alias warnings from getting duplicated and let it log only once.

Since this is not that important a feature, I think we should proceed with option 2. So that the user gets to know where the duplicate aliases are. Option 1 is incomplete and option 3 is too intricate for this. 

@eugenepeh @fzdy1914 @anubh-v What do you all suggest?

anubh-v:
Thanks for exploring the issue further. 

Correct me if I'm wrong, but doesn't the author information in `config.json` supersede the author information in `author-config.csv` ?

i.e. if `config.json` is present, and the `--ignore-standalone-config` flag is not used, then the authors' information is taken from `config.json` only?

If that's the case, then would it better to treat the duplicates in `author-config.csv` independently from the duplicates in `config.json`? 

i.e. we first show warnings for the duplicate aliases in `author-config.csv` (since csv is parsed before json config), then we show warnings for duplicate aliases in `config.json`? Any issues with this?

Tejas2805:
Alright! So, I have finally found the cause of the error. 👯‍♂ 

Screen Shot 2020-03-25 at 8 58 33 PM

Here, we set all the author details again. As a result of this, duplicate warnings are generated for aliases. 

Solution 1:
Let duplicate warnings be generated.

Pros: Easy to implement.
Cons: Might clog up the console. 

Solution 2:
Let warnings be generated only once.

Pros: User gets a good experience.
Cons: I will have to change backend a bit to bypass this. 

@eugenepeh @fzdy1914 @jylee-git Your opinions?

eugenepeh:
>Here, we set all the author details again. As a result of this, duplicate warnings are generated for aliases.

This is due to the standalone config? I don't really mind duplicate warnings as long as there's a clear separation to inform the user. E.g. warning: local config... or config.csv...

Tejas2805:
> This is due to the standalone config? I don't really mind duplicate warnings as long as there's a clear separation to inform the user. E.g. warning: local config... or config.csv...

So basically, first standalone config runs (unless ignored) and all authors based on that are added to the `authorList`. Then when it checks the author-config.csv, based on the records of the `IgnoreAuthorList` it again sets all the details. As a result of this,all authors despite not being in the csv, are set again. This results in duplicate warnings. So the warnings get generated as if coming from author-config.csv but in reality, they are not even there.

eugenepeh:
>So basically, first standalone config runs (unless ignored) and all authors based on that are added to the authorList. Then when it checks the author-config.csv, based on the records of the IgnoreAuthorList it again sets all the details. As a result of this,all authors despite not being in the csv, are set again. This results in duplicate warnings. So the warnings get generated as if coming from author-config.csv but in reality, they are not even there.

Standalone config itself should contains the author config component, why is it that it will check the author-config.csv? Is this a bug in our current implementation?

Tejas2805:
> Standalone config itself should contains the author config component, why is it that it will check the author-config.csv? Is this a bug in our current implementation?

Ah sorry, wrong choice of words. Will rephrase that. 

So basically, first standalone config runs (unless ignored) and all the authors based on that are added to the `authorList`. Then the CSV files run (if provided). This leads us to run the function `resetAuthorInformation` (due to ignoreAuthorsList). This function sets all the authors again (as seen in the screenshot attached in the previous comment). As a result of this, all authors from standalone config are set again. This results in duplicate warnings. So the warnings get generated as if coming from author-config.csv but in reality, they are not even present in the CSV files..

anubh-v:
Perhaps I can clarify.

The issue with the current implementation is that too many warnings are generated.
Let's use [this repo](https://github.com/anubh-v/test-repo) for discussion purposes.

Here is the [json config](https://github.com/anubh-v/test-repo/blob/master/_reposense/config.json).
Since the alias `anubh-v` is repeated 3 times, it would be reasonable to see 2 warnings from RepoSense.

When we run this PR, we get:
image

i.e. there are 2 more warnings than expected.

#### Reason for the extra warnings

As mentioned by @Tejas2805 [here](https://github.com/reposense/RepoSense/pull/1153#issuecomment-603826392), the `resetAuthorInformation` method in `AuthorConfiguration` causes these extra warnings

This method clears the *existing* author information in the `AuthorConfiguration` instance and replaces it with author information for a *new* list of authors.  It is called *after* we remove ignored authors from the list of authors in the `AuthorConfiguration` instance.

Because `resetAuthorInformation` causes the `AuthorConfiguration`'s internal TreeMap to be re-populated, it triggers the duplicate checking again --> resulting in the warnings being logged again.

Tejas2805:
@anubh-v Thanks for helping in clarifying. It's exactly like you said.

> One way to resolve the issue may be to change how we handle ignoring of authors.
Instead of resetting all author information and re-populating AuthorConfiguration 's internal TreeMap, we can try only resetting the author info of the ignored authors.

I was thinking more along the lines of bypassing the check the second time.

anubh-v:
> I was thinking more along the lines of bypassing the check the second time.

I see. I think this is fine. @fzdy1914  any thoughts on this approach?

anubh-v:
It seems that it should be doable to add the file name in the warning too?
So the warning would be like: 
`"Duplicate alias some-user found in config.json. The alias will belong to the last author who claims it.`

This may be worthwhile to add, since it helps user identify whether changes are needed in `author-config.csv` or `config.json`

Tejas2805:
Closing this PR. Work is continued in #1187 (bypass second check)

Issue #961, opened 2019-12-10
Explore using Vuex for globle variable management


fzdy1914:
We are getting more and more variables in our frontend, inter-component communication becomes more tedious. Some of our existing implementations can also be revamped using a more elegant way using global variables.

Let's explore the use of Vuex in our project and how it may benefit us.

Tejas2805:
Since we are thinking of exploring this, I was also thinking about revamping the front-end to Vue files. This will result in the JavaScript, HTML, CSS all associated with a particular component (eg: Authorship) to be in a single file and this will lead to better readability and understanding

jamessspanggg:
Yes I would agree to use Vuex as well.

fzdy1914:
> Since we are thinking of exploring this, I was also thinking about revamping the front-end to Vue files. This will result in the JavaScript, HTML, CSS all associated with a particular component (eg: Authorship) to be in a single file and this will lead to better readability and understanding

Yes, you may refer to #795 and #960 for more detail.

gerhean:
I am in favor of this.

fzdy1914:
Closed due to issue #1268

PR #1162, opened 2020-03-18
[#961] Add Vuex for event management

0blivious:
Part of #961 

```
The open of zoom and authorship tap is done through event
emit call now and the code will become ugly after #1101 and
#1059 get merged.

We can try to use Vuex to manage the event and call it in a
more elegant way.

Let's try to explore the use of vuex for event management.
```


jamessspanggg:
should we only consider vuex once we properly structure vue components?

0blivious:
> should we only consider vuex once we properly structure vue components?

Then very soon I will have sth like this.$parent.$parent.emit......I feel the projectify thing will take a very long time to merge. We can just shift the vuex part to a seperate file at that time.
@fzdy1914 mentioned can try use Vuex https://github.com/reposense/RepoSense/pull/1059#discussion_r390793866

fzdy1914:
I don't think we need to wait for the projectify because these two do not contradicts with each other. Meanwhile, can you put `tab open` and `tab type` into Vuex as well?

0blivious:
> I don't think we need to wait for the projectify because these two do not contradicts with each other. Meanwhile, can you put `tab open` and `tab type` into Vuex as well?

Can but why do we need to put them into Vuex? It seems they are only used in main.js.

frontend/src/static/js/main.js
```diff
@@ -133,6 +151,22 @@ window.app = new window.Vue({
...
+  computed: {
+    zoomTabInfo() {
+      return this.$store.state.zoomTabInfo;
+    },
+    authorshipInfo() {
+      return this.$store.state.authorshipInfo;
+    },
+  },
+  watch: {
+    zoomTabInfo() {
```
fzdy1914: Can the component watch the state directly? In my memory, it can.  Meanwhile, if we are already watching Info, it may be better to put code of that the method inside the watch code body directly for simplicity.

fzdy1914:
> > I don't think we need to wait for the projectify because these two do not contradicts with each other. Meanwhile, can you put `tab open` and `tab type` into Vuex as well?
> 
> Can but why do we need to put them into Vuex? It seems they are only used in main.js.

My bad, didn't looked into your code clearly enough.

frontend/src/static/js/main.js
0blivious: Do you mean put the updateTabZoom inside the watch method? The zoom one can but the authorship one is also used somewhere else so cannot shift.

0blivious:
@fzdy1914 hi can I get your review on this?

frontend/src/static/js/main.js
```diff
@@ -1,3 +1,20 @@
+const store = new Vuex.Store({
```
fzdy1914: Better to put in a separate file. It is kind of standard.

0blivious: I intended to put it in a separate file but for our current structure, I failed to use import. So if in a separate file I have to use sth like window.store. I am not very sure if I am correct on this. Can tell me how to separate the files correctly? Thx in advance!

fzdy1914: Why can't you import the store? Did you export it?

fzdy1914: image
That's how I use vuex in another program.

0blivious: My js file is exactly the same. After that in the `main.js` need to have sth like `import store from *.js` right? Then I will have this error `main.js:1 Uncaught SyntaxError: Cannot use import statement outside a module`. I dont know what's wrong with the structure.

jamessspanggg:
Travis is failing with some errors. Could you fix is so that the preview comes up?

frontend/src/static/js/main.js
```diff
@@ -115,6 +116,16 @@ window.app = new window.Vue({
 
     errorMessages: {},
   },
+  watch: {
+    '$store.state.zoomTabInfo': function () {
```
jamessspanggg, line 120: hmm, with these two additions, is the `updateTabZoom` and `updateTabAuthorship` methods still necessary?

frontend/src/static/js/store.js
```diff
@@ -0,0 +1,15 @@
+/* global Vuex */
+window.store = new Vuex.Store({
+  state: {
+    authorshipTabInfo: {},
```
jamessspanggg, line 4: Should we standardise the naming? the above uses tabAuthorship and here uses authorshipTab

frontend/src/static/js/main.js
ccyccyccy: > My js file is exactly the same. After that in the `main.js` need to have sth like `import store from *.js` right? Then I will have this error `main.js:1 Uncaught SyntaxError: Cannot use import statement outside a module`. I dont know what's wrong with the structure.

I played around with it. This should work if you change `script(src="static/js/main.js")` to `script(type="module" src="static/js/main.js")`.
Just use `import store from './store.js';` in `main.js`.

frontend/src/static/js/store.js
0blivious: done

frontend/src/static/js/main.js
0blivious: it was removed, accidentally added back after merging latest master

jamessspanggg:
@0blivious I think you need to add commas to separate them:
Screen Shot 2020-04-04 at 10 28 45 PM

frontend/src/index.pug
```diff
@@ -16,6 +16,8 @@ html
 
     script(src="https://use.fontawesome.com/releases/v5.0.13/js/all.js", defer=True)
     script(src="https://cdn.jsdelivr.net/npm/vue@2.5/dist/vue.js")
+    script(src="https://unpkg.com/vuex@3.1.3/dist/vuex.js")
+    script(src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.auto.js")
```
eugenepeh, line 20: whats the reason for adding this script?

0blivious: from https://vuex.vuejs.org/installation.html
Vuex requires Promise. If your supporting browsers do not implement Promise (e.g. IE), you can use a polyfill library, such as es6-promise.

0blivious: @eugenepeh can I get your opinion on this?

0blivious:
> Code Quality seems to be failing. Could you rectify it and ask for a review again?

That one cannot be resolved. Our CI style rule conflicts with Code Quality rule. Might need another PR to address this.

frontend/src/index.pug
0blivious: @eugenepeh can I get your opinion on this?

eugenepeh: I see. Can we separate the indirect dependencies, grouping them together on top?
E.g. 
```
// dependencies 
script(src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.auto.js")

script(src="https://unpkg.com/vuex@3.1.3/dist/vuex.js")
```

Others should then be arranged lexicographically according to cdn.

0blivious: I have put the es6 dependencies to the top. I moved the font awesome package to indirect dependencies also as now we use vue-font-awesome icon, and it exploits font-awesome package. Is this ok?

jamessspanggg:
@0blivious can u resolve the conflicts?

PR #1165, opened 2020-03-18
[#950] Use surge as alternative for netlify

### LINKED ISSUE #950: Find a netlify alternative

jamessspanggg:
Fixes #950

```
Netlify's build quota is not enough for RepoSense's build usage.

Let's deploy on Surge instead, using Travis to run the build and provide
a Github Status on each PR that links users to the deployed Surge site.
```


jamessspanggg:
@eugenepeh the github status is working now. Thanks!

.travis.yml
```diff
@@ -60,3 +57,7 @@ cache:
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
+  - git fetch --all && git config --global user.email "-" && git config --global user.name "-"
```
jamessspanggg: @eugenepeh may i get the purpose of `git config --global user.email "-" && git config --global user.name "-"` in the run command?

eugenepeh: There are some git commands that cannot be performed without the user.name and user.email value. We simply use the `-` value to bypass that need.

jamessspanggg: ahh ok, noted.

fzdy1914:
Can I propose your commit message? I will merge the PR now.

jamessspanggg:
> Can I propose your commit message? I will merge the PR now.

done

fzdy1914:
The travis fails, should we only run the surge for one sdk only?

jamessspanggg:
> The travis fails, should we only run the surge for one sdk only?

hmm, but it may be the case that the first sdk fails and the second is successful right. In this way we still get a preview, if we run both.

eugenepeh:
>hmm, but it may be the case that the first sdk fails and the second is successful right. In this way we still get a preview, if we run both.

I think its fine if we do the preview in just one env to slim down the build time; either way passing the CI test is a requirement for the PR to be merged.

jamessspanggg:
@eugenepeh @fzdy1914 updated to only run on the linux build.

PR #1169, opened 2020-03-20
[#1146] Add caret icon to indicate collapsible authorship file panel

### LINKED ISSUE #1146: Not intuitive that authorship file detail is collapsible
Currently it is not clear that the title is clickable to collapse the code view.

image

Let's add a small arrow such that it is clear that the panel is collapsible, similar to github's one

image

It should change when the panel is collapsed.

image

ccyccyccy:
Fixes #1146 
```
It is not apparent that the authorship file details can be collapsed by
clicking on the title.

Let's add caret icons to show that the title is clickable to collapse the
file details view.
```

### DIFF
--- a/frontend/src/static/css/v_authorship.scss
+++ b/frontend/src/static/css/v_authorship.scss
@@ -91,11 +91,8 @@
       margin-top: 1rem;
       padding: .3em .5em;
 
-      .collapsible {
-        cursor: pointer;
-      }
-
       .path {
+        cursor: pointer;
         overflow-wrap: break-word;
       }
 
--- a/frontend/src/static/css/v_authorship.scss
+++ b/frontend/src/static/css/v_authorship.scss
@@ -91,8 +91,11 @@
       margin-top: 1rem;
       padding: .3em .5em;
 
-      .path {
+      .collapsable {
         cursor: pointer;
+      }
+
+      .path {
         overflow-wrap: break-word;
       }
 
--- a/frontend/src/static/css/v_authorship.scss
+++ b/frontend/src/static/css/v_authorship.scss
@@ -91,7 +91,7 @@
       margin-top: 1rem;
       padding: .3em .5em;
 
-      .collapsable {
+      .collapsible {
         cursor: pointer;
       }
 
--- a/frontend/src/tabs/authorship.pug
+++ b/frontend/src/tabs/authorship.pug
@@ -82,7 +82,7 @@
     template(v-for="file in selectedFiles")
       .file(v-bind:key="file.path")
         .title
-          span.collapsable(v-on:click="toggleFileActiveProperty(file)")
+          span.collapsible(v-on:click="toggleFileActiveProperty(file)")
             font-awesome-icon(icon="caret-down", fixed-width, v-show="file.active")
             font-awesome-icon(icon="caret-right", fixed-width, v-show="!file.active")
             span.path {{ file.path }} &nbsp;
--- a/frontend/src/tabs/authorship.pug
+++ b/frontend/src/tabs/authorship.pug
@@ -82,10 +82,10 @@
     template(v-for="file in selectedFiles")
       .file(v-bind:key="file.path")
         .title
-          span.collapsible(v-on:click="toggleFileActiveProperty(file)")
+          span.path(v-on:click="toggleFileActiveProperty(file)")
             font-awesome-icon(icon="caret-down", fixed-width, v-show="file.active")
             font-awesome-icon(icon="caret-right", fixed-width, v-show="!file.active")
-            span.path {{ file.path }} &nbsp;
+            span {{ file.path }} &nbsp;
           span.icons
             a(
               v-bind:href="getFileLink(file, 'commits')", target="_blank",
--- a/frontend/src/tabs/authorship.pug
+++ b/frontend/src/tabs/authorship.pug
@@ -82,7 +82,10 @@
     template(v-for="file in selectedFiles")
       .file(v-bind:key="file.path")
         .title
-          span.path(v-on:click="toggleFileActiveProperty(file)") {{ file.path }}&nbsp;
+          span.collapsable(v-on:click="toggleFileActiveProperty(file)")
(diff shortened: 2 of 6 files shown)


frontend/src/static/css/v_authorship.scss
```diff
@@ -120,6 +120,30 @@
...
+            }
+
+            .fa-caret-right {
+              display: none;
+            }
+          }
+        }
+      }
+
+      .title {
```
Tejas2805: Is it necessary to have this twice? Can you try with having this only once? If no suitable solution, we will go ahead with this implemention

ccyccyccy: I can't think of a better solution.

But there are several instances of similar patterns in the repo, such as [here](https://github.com/reposense/RepoSense/blob/bf77c18fd41e53676458e80fd07922a215c7472c/frontend/src/static/css/v_authorship.scss#L103-L124) and [here](https://github.com/reposense/RepoSense/blob/bf77c18fd41e53676458e80fd07922a215c7472c/frontend/src/static/css/v_authorship.scss#L196-L235)

If we ever find a more elegant solution we can improve on all of them. But for now I think this is fine.

frontend/src/tabs/authorship.pug
```diff
@@ -56,7 +56,10 @@
     template(v-for="file in selectedFiles")
       .file.active(v-bind:key="file.path")
         .title
-          span.path(onclick="toggleNext(this.parentNode)", v-on:click="updateCount") {{ file.path }}&nbsp;
+          span.clickable(onclick="toggleNext(this.parentNode)", v-on:click="updateCount")
+            span.fas.fa-caret-down.fa-fw
+            span.fas.fa-caret-right.fa-fw
+            span.path {{file.path }}&nbsp
```
0blivious: spacing here

ccyccyccy: fixed

frontend/src/static/css/v_authorship.scss
jylee-git: Why not utilize the `file` (collapsed) and `file active` (expanded) classes to deal with this?

ccyccyccy: Could you elaborate more on that? I thought that is what I am currently doing. My current implementation is to change the display based on whether the class `active` exists.

frontend/src/tabs/authorship.pug
```diff
@@ -56,7 +56,10 @@
     template(v-for="file in selectedFiles")
       .file.active(v-bind:key="file.path")
         .title
-          span.path(onclick="toggleNext(this.parentNode)", v-on:click="updateCount") {{ file.path }}&nbsp;
+          span.clickable(onclick="toggleNext(this.parentNode)", v-on:click="updateCount")
+            span.fas.fa-caret-down.fa-fw
+            span.fas.fa-caret-right.fa-fw
+            span.path {{ file.path }} &nbsp;
```
Tejas2805: Is it necessary for the path to still help in toggling? Now that we the toggle button, maybe better to just allow that to toggle the files.

ccyccyccy: Prof Damith said [here](https://github.com/reposense/RepoSense/issues/1146#issuecomment-599176215) that it's fine to allow both.

frontend/src/static/css/v_authorship.scss
jylee-git: > Could you elaborate more on that? I thought that is what I am currently doing. My current implementation is to change the display based on whether the class `active` exists.

Yes now I see it. Sry about that, couldn't see the indentation properly.

Aside from that, would it be better if you move lines 137-146 to be inside the existing `.title` class (lines 126 and beyond)?

ccyccyccy: I guess we can split them apart, but I feel that keeping the 2 sections together makes it more immediately obvious to the reader about their behaviours.

jylee-git:
One more thing: Notice that in the commits panel, commits with commit message body will have that `***` to collapse and expand the message body.

If possible, let's standardize it by effecting the collapsible indicator over to the commits panel as well. Can either deal with it in this PR or make a new issue to be dealt with in another PR.

frontend/src/static/css/v_authorship.scss
jylee-git: Leaving this open for others to make their input if necessary.

ccyccyccy:
> 
> 
> One more thing: Notice that in the commits panel, commits with commit message body will have that `***` to collapse and expand the message body.
> 
> If possible, let's standardize it by effecting the collapsible indicator over to the commits panel as well. Can either deal with it in this PR or make a new issue to be dealt with in another PR.

I think we can deal with it in another PR, since currently their styles are very different already. Personally also don't think the same arrows will look nice for the commit panels, since it's on a white background without any borders.

0blivious:
Do you still intend to make the arrow facing downwards when expanded? It is not working in the surge preview.

ccyccyccy:
> Do you still intend to make the arrow facing downwards when expanded? It is not working in the surge preview.

Thanks for catching this :) The current implementation is affected by #1065 due to the file collapsing mechanism being changed, so it doesn't work anymore.

As such, there is another better approach by making use of the new `file.active` property to properly toggle the arrows. However, this approach is affected by #1191, and will have to wait until #1186 to be merged for this PR to be implemented.

frontend/src/static/css/v_authorship.scss
```diff
@@ -91,8 +91,11 @@
       margin-top: 1rem;
       padding: .3em .5em;
 
-      .path {
+      .collapsable {
```
jamessspanggg, line 94: collapsable -> collapsible

0blivious: is it possible to use 1 css class only instead of 2 (path and collapsible) ?

ccyccyccy: > is it possible to use 1 css class only instead of 2 (path and collapsible) ?

Yep it is very possible. Thanks for the suggestion. I changed to that.

PR #1056, opened 2020-02-05
[#1046] Refactor duplicate code

yash-chowdhary:
Fixes #1046 

```
There is an instance of code duplication in 'v_summary.js'. 
Specifically, the definition of a function that is passed to the comparator 
has been duplicated in the case of grouping by repos and by authors.

Let's extract this duplicated portion of code into a helper function so that
it may be referenced by the comparator.
```

PR #1171, opened 2020-03-20
[#1170] Update total contribution to reflect based on checked file types

### LINKED ISSUE #1170: Total line contribution not updated when filetypes are unchecked
When filetypes are unchecked, the contribution bar length adjusts to the correct length, while the total line contribution of each group does not (shown in diagram below). 

Hence, when toggling sorting controls to decreasing order of contribution, groups with higher total contribution are seen to be in lower position than groups with lower contribution (as they are sorted based on total contribution among checked filetypes). This may cause confusion in users.

Screen Shot 2020-03-20 at 2 30 02 PM

jamessspanggg:
Fixes #1170 

```
When some filetypes are unchecked, the contribution bar length adjusts 
to the correct length, while the total line contribution of each group 
does not. Additionally, when toggling sorting controls to decreasing 
order of contribution, groups with higher total contribution are seen 
to be in lower position than groups with lower contribution. This may 
cause confusion among users.

Let's update the total contribution to reflect the total contribution 
among those file types that are checked.
```


jamessspanggg:
Putting this on hold until #1056 merges.

jamessspanggg:
note that the variance will not be adjusted according to the filetype checkboxes, this is because the variance is calculated in the backend based on the author, not based on the files.

jamessspanggg:
> Better to add the variance thing to the UG. Otherwise, the users might expect that to change to.

Hmm, the current UG does not state that the contribution of each group will be based on the filetypes checked tho. So should I still even state about the variance?

Also, the current UG does not have the updates of filetype checkboxes in summary, as the changes are in #1149

Tejas2805:
> Hmm, the current UG does not state that the contribution of each group will be based on the filetypes checked tho. So should I still even state about the variance?

How about adding this too? In the same place where we talk about sorting each group based on contribution? 
Cause I think this feature might be somewhat intuitive for a new user but an original/old user might get confused about what is exactly happening

> Also, the current UG does not have the updates of filetype checkboxes in summary, as the changes are in #1149

I had ensured that @0blivious adds it in the same PR, so its there. Can just edit that. 

Screen Shot 2020-03-25 at 8 47 12 PM

frontend/src/static/js/v_summary_charts.js
```diff
@@ -167,6 +167,24 @@ window.vSummaryCharts = {
       return fileTypes;
     },
 
+    getFileTypeContribution(ele) {
```
jamessspanggg, line 170: Adding these 3 methods to `v_summary_charts` even though the methods exist in `v_summary`.

This is because both vue components are utilising these methods. And vue doesn't allow the child component to _call parent's method with a return value_.

Not sure if this is the best way of doing it.

docs/UserGuide.md
```diff
@@ -141,6 +141,7 @@ The `Tool Bar` at the top provides a set of configuration options that control t
 * `Breakdown by file type` : toggles the contribution bar to either display the bar by :
     * the total lines of codes added (if checkbox is left unchecked), or
     * a breakdown of the number of lines of codes added to each file type (if checkbox is checked). If sorting criteria is contribution, only lines contributed to selected file types will be taken into account.
+    * the total contribution of each group will update based on changes in file type checkboxes, while their variance does not.
```
Tejas2805, line 144:
```suggestion
    * the total contribution of each group will get updated based on the selected file type checkboxes, while their variance will not.
```

frontend/src/static/js/v_summary_charts.js
Tejas2805: If we refactor out these three functions to `util.js`, and then call them in `v_summary` and `v_summary_charts` will it work then?

jamessspanggg:
> If we refactor out these three functions to util.js, and then call them in v_summary and v_summary_charts will it work then?

It will work but it's not ideal as the methods are specifically used only in `v_summary` and `v_summary_charts`.

Tejas2805:
> It will work but it's not ideal as the methods are specifically used only in `v_summary` and `v_summary_charts`.

I think it's okay for `util.js` to have a few methods that are being used by certain components only, as this prevents duplication of code. I think **The DRY Principle** kind of fights for this cause. And if we end up modifying the functionality, we will need to make the same change at two places. So, I am not too pro for duplicate code.  

Let's ask one of the senior devs regarding this? And then just proceed accordingly?

jamessspanggg:
> Let's ask one of the senior devs regarding this? And then just proceed accordingly?

Ok. In the meantime, let me find if there's a better solution to this.

jamessspanggg:
@Tejas2805 I have simplified the implementation such that there will be no duplicate methods. Basically I just added a new `checkedFileTypeContribution` property to each repo object, in this way the property is able to pass from `v_summary` to `v_summary_charts`.

You can proceed with the review.

docs/UserGuide.md
```diff
@@ -127,6 +127,7 @@ The `Tool Bar` at the top provides a set of configuration options that control t
 * `Breakdown by file type` : toggles the contribution bar to either display the bar by :
     * the total lines of codes added (if checkbox is left unchecked), or
     * a breakdown of the number of lines of codes added to each file type (if checkbox is checked). If sorting criteria is contribution, only lines contributed to selected file types will be taken into account.
+    * the total contribution of each group will get updated based on the selected file type checkboxes, while their variance will not.
```
eugenepeh, line 130: I think it would be better if this is combined into the 2nd point as the original doc emphasize the difference between checked and unchecked.

>If sorting criteria is contribution, only lines contributed to selected file types will be taken into account.

This sentence can be moved to bottom as remark or moved to the notes.

jamessspanggg: I combined both the added sentence and 
> If sorting criteria is contribution, only lines contributed to selected file types will be taken into account.

as one,  placed them into the notes section.

This is because the `Breakdown by file type` was talking about how it affects contribution bar, so i think additional infos like these should be in the notes section.

PR #1172, opened 2020-03-21
[#1167] Revert PR #967 and make zoom tab independent of summary

### LINKED ISSUE #1167: Authorship / Zoom tab bug from regressions introduced by PR #967
The PR #967 caused some regressions because it did not re-render the component, causing the previous state to be preserved.

This caused the following bugs
1. Duplicate values getting pushed into `fileTypes`, causing filter selections to not work properly. This occurs even when the same authorship icon is clicked multiple times.
image
2. `getFileTypeExistingLinesObj` is not invoked when the prop `info` is changed as `filesLinesObj` is not watched by Vue.  As such, `getFileTypeExistingLinesObj` keeps the value the first author selected, even after selecting another author. This caused the filetype linecount to be not presented correctly.
image

There might be more bugs due to the old state being preserved. Will continue to investigate.

jamessspanggg:
Fixes #1167 

```
PR #967 causes multiple bugs in zoom tab and authorship tab, this is
because the side panels are not re-rendered when prop `info` changes, 
which causes other values in the panel to not be refreshed when `info`
changes.

Fixing these bugs may delay our upcoming release, so reverting the changes
that lead to these bugs may be the best option.

Let's revert that PR while only keeping the changes to make the zoom tab
to be independent of summary.
```


jamessspanggg:
@eugenepeh @fzdy1914 this PR reverts the changes as discussed in #1167

PR #1175, opened 2020-03-22
[#996] Add functionality to view code for entire repository

### LINKED ISSUE #996: Give a way to view code for the entire repository
Give a way to view the code of the entire repository at once and not just author wise. 

Use case: to see all the code of a repository as one page
Pros: easier than viewing one file at a time as one has to do via the GitHub UI
Cons: might not scale to big repositories.

Description edited by @damithc

jamessspanggg:
Fixes #996

```
There's no way of viewing the code authored for the entire repository.
This is particularly useful when users want to see the code contribution
of the repository as a whole.

We can allow this functionality by enabling the code icon when merge
group.

Let's add the functionality to view code for the entire repository, when
merge group is checked.
```


jamessspanggg:
On hold until #1172 merges

frontend/src/static/js/v_authorship.js
```diff
@@ -88,10 +89,23 @@ window.vAuthorship = {
...
     getRepoProps(repo) {
       if (repo) {
-        const author = repo.users.filter((user) => user.name === this.info.author);
-        if (author.length > 0) {
-          this.info.name = author[0].displayName;
-          this.filesLinesObj = author[0].fileTypeContribution;
+        if (this.info.isMergeGroup) {
+          // sum of all users' file type contribution
+          repo.users.forEach((author) => {
+            Object.entries(author.fileTypeContribution).forEach(([type, cnt]) => {
```
anubh-v: Just a suggestion: would it help readability if we abstract out the logic in the inner loop into a separate method?
Perhaps a method that updates the filesLinesObj given a fileTypeContribution object?

jamessspanggg: Yep that would be much cleaner. Refactored

frontend/src/summary.pug
```diff
@@ -122,11 +122,19 @@
           title="Total contribution of group"
         ) [{{ getGroupTotalContribution(repo) }} lines]
         template(v-if="isMergeGroup")
+          a(
+            v-if="filterGroupSelection === 'groupByRepos'",
+            title="click to view repo's code",
+            onclick="deactivateAllOverlays()",
+            v-on:click="openTabAuthorship(repo[0], repo, i)"
```
ccyccyccy: Can't you just pass in a `0` here for the `index` parameter, instead of modifying the `openTabAuthorship` method? Seems like the argument `i` is not used in `openTabAuthorship`.

frontend/src/static/js/main.js
```diff
@@ -214,10 +215,11 @@ window.app = new window.Vue({
       const info = {
         author: hash.tabAuthor,
         repo: hash.tabRepo,
+        isMergeGroup: hash.authorshipIsMergeGroup === 'true',
```
ccyccyccy: Might want to rename it to `isMergeAuthorshipGroup`? Since there are 2 ways of merging group.

jamessspanggg: Hmm don't think this is necessary as `isMergeGroup` is contained in the `info` object, which will be passed to authorship tab

frontend/src/summary.pug
jamessspanggg: Yep, didn't realise that

frontend/src/static/js/main.js
ccyccyccy: Ahh you are right. Was forgetting which file I was looking at.

frontend/src/static/js/v_authorship.js
```diff
@@ -169,7 +195,11 @@ window.vAuthorship = {
       let totalBlankLineCount = 0;
 
       files.forEach((file) => {
-        const lineCnt = file.authorContributionMap[this.info.author];
+        const contributionMap = file.authorContributionMap;
+        const lineCnt = this.info.isMergeGroup
+          ? Object.entries(contributionMap).reduce((acc, [author, cnt]) => (author !== '-' ? acc + cnt : acc), 0)
```
ccyccyccy: Could consider refactoring this to another function, even though it is just a 1 liner, since this is not too easy to digest. Maybe call it `sumContributionsFromAllAuthors`.

frontend/src/static/js/v_authorship.js
```diff
@@ -88,18 +89,36 @@ window.vAuthorship = {
...
-        if (author.length > 0) {
-          this.info.name = author[0].displayName;
-          this.filesLinesObj = author[0].fileTypeContribution;
+        if (this.info.isMergeGroup) {
+          // sum of all users' file type contribution
+          repo.users.forEach((author) => {
+            this.updateTotalFileTypeContribution(author.fileTypeContribution);
+          });
+        } else {
+          const author = repo.users.filter((user) => user.name === this.info.author);
```
ccyccyccy: I know it's not part of this PR, but can consider using [find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here

frontend/src/static/js/v_authorship.js
jamessspanggg: Yep refactored

frontend/src/static/js/v_authorship.js
jamessspanggg: Done

frontend/src/static/js/v_authorship.js
```diff
@@ -88,18 +89,36 @@ window.vAuthorship = {
...
-          this.info.name = author[0].displayName;
-          this.filesLinesObj = author[0].fileTypeContribution;
+        if (this.info.isMergeGroup) {
+          // sum of all users' file type contribution
+          repo.users.forEach((author) => {
+            this.updateTotalFileTypeContribution(author.fileTypeContribution);
+          });
+        } else {
+          const author = repo.users.find((user) => user.name === this.info.author);
+          if (author.length > 0) {
```
anubh-v: Is `author` a single value now, if we are using `find`? Would that mean we do not need to check the length?

jamessspanggg: Yes author is a single value, we still need to check if `length > 0` just in case

anubh-v: I see. Just to clarify, if there are no matches, `find` would return undefined? In that case, there wouldn't be a length property?

jamessspanggg: ahhh i see, good catch, thanks for the investigation

anubh-v: Also, within the `if` branch, should we be using `author.displayName` instead of `author[0].displayName` ?

0blivious:
Bug for the icon
1. Group by Repo -> merge group
2. Select Group by Author
3. Weird icon placement. Both of the icons below lead to author's repo.
image

jamessspanggg:
> Bug for the icon

Fixed. Related to #1186

0blivious:
Can do this when merge by author?

jamessspanggg:
> Can do this when merge by author?

That's going to need quite a lot of changes because of how the current data is being structured now, which groups based on repositories. Perhaps that can be a separate PR in the future, this only targets a way of viewing all authored code for repositories.

frontend/src/static/js/v_authorship.js
```diff
@@ -119,8 +138,12 @@ window.vAuthorship = {
     },
 
     hasCommits(info) {
-      if (window.REPOS[info.repo]) {
-        return window.REPOS[info.repo].commits.authorFinalContributionMap[info.author] > 0;
+      const { isMergeGroup, author } = info;
+      const repo = window.REPOS[info.repo];
+      if (repo) {
+        return isMergeGroup
+          ? Object.entries(repo.commits.authorFinalContributionMap).some(([name, cnt]) => name !== '-' && cnt > 0)
```
anubh-v: A suggestion: we're comparing against the name `-` at several points.
Would it be more readable to wrap this in a method like `isUnknownAuthor` ?

jamessspanggg: Yep that would be clearer. Perhaps it could be listed as a first time issue?

anubh-v: Sure, I'll open an issue on it

jamessspanggg:
> Should we update the UG to highlight the functionality of viewing authored code as a group?

Done

Tejas2805:
There are two PRs which might affect this PR - #1171 and #1152 

Because the best use case for this feature I think is - Breakdown by file type (selected files) + merge group + (click on the total repo code) -> shows code for the full repo with the selected file

Maybe we can keep this PR on hold till these two get merged? They are in the reviewing stage too.

jamessspanggg:
> There are two PRs which might affect this PR - #1171 and #1152

#1171 will not affect this PR as it does not touch on the authorship side. #1152 does not touch on the current changes as well. There wouldn't be merge conflicts and I believe we can review it together. I think the order doesn't matter as well as it wouldn't make a difference.

jamessspanggg:
@eugenepeh may i get a final review?

frontend/src/static/js/v_authorship.js
eugenepeh, line 145: should be double indented

frontend/src/static/js/v_authorship.js
```diff
@@ -169,7 +195,11 @@ window.vAuthorship = {
       let totalBlankLineCount = 0;
 
       files.forEach((file) => {
-        const lineCnt = file.authorContributionMap[this.info.author];
+        const contributionMap = file.authorContributionMap;
+        const lineCnt = this.info.isMergeGroup
+          ? this.getContributionFromAllAuthors(contributionMap)
```
eugenepeh, line 200: same for here

jamessspanggg:
> minor nits

done.

jamessspanggg:
@fzdy1914 ready to merge

jamessspanggg:
> Should due to the regression of this PR

This PR is only available when merge group and `group by repo`.

> Let's add the functionality to view code for the entire repository, when
merge group is checked.

jamessspanggg:
> Should due to the regression of this PR
> 
> image
> 
> image

This is due to the reactivity of the icons, which is related to #1191 and being fixed in #1186

jamessspanggg:
I added a Cypress front end test for the opening of code view when `merge group` and `group by repos`. This is to further prevent any regressions.

PR #1164, opened 2020-03-18
[#1129][#1157] A draft to check CI tests for #1157

shuiyao-sg:

PR #1176, opened 2020-03-23
[#1164] Have a common js file for global function

### LINKED ISSUE #719: Separate direct and indirect methods in javascripts

### LINKED ISSUE #1161: Have a common util file for global constant and function
Currently global functions and constants are registered in multiple js files which make them hard to trace. We can have a common util.js file for now to make the code more readable and easier to refactor in the future.

0blivious:
Fixes #1161
Fixes #719
```
Global functions and constants are registered in multiple js files
which make them hard to trace.

We can move all the global functions to api.js to
make the code more readable and easier to refactor in the future.

Let's use api.js to store all the global functions and
constants.
```


frontend/src/ramp.pug
```diff
@@ -3,9 +3,9 @@
     template(v-for="(slice, j) in user.commits")
       a.ramp__slice(
         draggable="false",
-        onclick="rampClick(event);",
         v-for="(commit, k) in slice.commitResults",
         v-if="commit.insertions>0",
+        v-on:click="rampClick",
```
ang-zeyu: #1114

frontend/src/static/js/v_ramp.js
```diff
@@ -21,6 +14,12 @@ window.vRamp = {
   },
 
   methods: {
+    rampClick(evt) {
```
ang-zeyu: #1114

0blivious:
@fzdy1914 can we get your opinion on this?

fzdy1914:
Just a question, Have you put all the methods that can be put inside a component already?

0blivious:
> Just a question, Have you put all the methods that can be put inside a component already?

Yes, some of them will be moved into another component in #1179. For now I put them in api.js temporarily.

> This looks good to me. I also agrees that utils and api js files seem to have the same usage now.

Now all the global function in api.js

frontend/src/static/js/v_summary_charts.js
```diff
@@ -1,94 +1,3 @@
...
-}
-
-function deactivateAllOverlays() {
-  document.querySelectorAll('.summary-chart__ramp .overlay')
-      .forEach((x) => {
-        x.className = 'overlay';
-      });
-}
-
-function dragViewDown(evt) {
```
ang-zeyu, line 44: for all the changes in this file except `window.comparator`: Would it be better to move it to the component definition? Since this isn't really a utility-like function and relies heavily on `summary`'s structure. 
Likewise for `dismissTab`, but in that case it would be moving it to `summary` instead of `summary_chart`'s definition

0blivious: all moved into respective component except deactivatealloverlay which is used in both summary and summarycharts

0blivious: all moved into respective component except deactivateAllOverlay which is used in both summary and summaryCharts

frontend/src/static/js/main.js
```diff
@@ -329,7 +290,7 @@ window.app = new window.Vue({
   updated() {
     this.$nextTick(() => {
       if (window.$('tabs-wrapper')) {
-        window.$('tabs-wrapper').style.flex = `0 0 ${flexWidth * 100}%`;
+        window.$('tabs-wrapper').style.flex = `0 0 ${window.flexWidth * 100}%`;
```
ang-zeyu, line 293: missed one out

frontend/src/static/js/v_summary.js
```diff
@@ -1,49 +1,6 @@
-// date functions //
-const DAY_IN_MS = (1000 * 60 * 60 * 24);
-window.DAY_IN_MS = DAY_IN_MS;
-const WEEK_IN_MS = DAY_IN_MS * 7;
+const WEEK_IN_MS = window.DAY_IN_MS * 7;
```
jamessspanggg, line 1: Perhaps can put this in the util.js as well? since it's pretty much a constant that might be useful for multiple classes

0blivious: According to previous discussion we decided to move all util function into api.js, I will update the commit message

frontend/src/summary_charts.pug
```diff
@@ -88,8 +88,7 @@
         ) {{ getPercentile(i) }} %
 
       .summary-chart__ramp(
-        onclick="viewClick(event)",
```
jamessspanggg, line 91: Why removed this? Am unable to `cmd-cliick` to zoom into ramps anymore.

frontend/src/static/js/v_summary.js
jamessspanggg: Can place WEEK_IN_MS as a window constant as well

frontend/src/summary_charts.pug
0blivious: the click is controlled by opentabzoom. and the feature is working on my computer.
Can help me check if it is due to the mac system? anything printed on your console?

Tejas2805: Could be a possibility. Doens't work for me too.

frontend/src/static/js/v_summary.js
0blivious: noticed after updating the constant is only used in one method. Has shifted it inside the component

frontend/src/summary_charts.pug
0blivious: Fixed. I forgot to change this.isMacintosh to window.isMacintosh

jamessspanggg:
@fzdy1914 can get one more review from you?

Issue #1185, opened 2020-03-26
Link to author profile not working when config.json not provided


ccyccyccy:
When `group by` Author and `merge group` is selected, the author profile linked by the octocat icon is shown. However, when `config.json` is not provided by the target repo, the hyperlink does not work.

This is due to the `name` of the author being the same as the `displayName` of the author when the config file is not provided, where the `name` is supposed to be the github id of the author.

HCY123902:
Actually, is there a way to reproduce this now? I think currently the profile can still be accessed even if the configuration file is not provided.

gerhean:
Suspect is the same issue as #1484

PR #1186, opened 2020-03-26
[#1178] Make github profile and repo hyperlinks more consistent

### LINKED ISSUE #1178: UI for viewing repo confusing
Currently we have an octocat icon to redirect users to the respective repos, as well as the same ocotcat icon directing users to github profiles when grouped by `Authors` and `merge group` selected. 

This icon is duplicated multiple times in the default setting of grouping by `Repo/Branch`, with them all pointing to the same repo.

image

This is confusing as one would think that the icon is related to the author, only to find out that they all point to the same repo.

Also, the other 2 icons changes the tab, while the octocat icon is the odd one out that redirects users to elsewhere, and is not consistent with its neighbours.

Hence, I would suggest that we make the author names and the repo names clickable to redirect users to the author profile and repo url respectively, replacing the need for the octocat icons. This would make the UI consistent no matter how we `group by` or if `merge group` is selected.

ccyccyccy:
Fixes #1178 and  #1191.
```
When grouped by repo, the same icon to link to the repo is duplicated
multiple times for no reason. Also, the icon to link to the author's
profile is hard to find.

We can make the links more intuitive and consistent by putting links to
the author's profile beside author names, and links to repo beside repo
names.

Let's reorder the icon group such that the hyperlinks are closer to the
respective names, as well as always showing the icons beside their
respective names. As this change heavily relies on the icons being
responsive, let's resolve issue #1191 as well to make use of
vue-fontawesome to make the icons responsive.
```

Note that there are some existing bugs that affect this PR.
1. Affected by #1185, but should work with the previews as the `config.json` is specified. Preferably merge after #1185 is resolved.

### DIFF
--- a/frontend/src/summary.pug
+++ b/frontend/src/summary.pug
@@ -72,7 +72,7 @@
     .error-message-box__close-button(onclick="dismissTab(this)") &times;
     .error-message-box__message The following issues occurred when analyzing the following repositories:
     .error-message-box__failed-repo(v-for="errorBlock in errorMessages")
-      i.fas.fa-exclamation.mini-font
+      font-awesome-icon.mini-font(icon="exclamation")
       span.error-message-box__failed-repo--name {{ errorBlock.repoName }}
       .error-message-box__failed-repo--reason(
         v-if="errorBlock.errorMessage.startsWith('Unexpected error stack trace')"
--- a/frontend/src/index.pug
+++ b/frontend/src/index.pug
@@ -15,6 +15,7 @@ html
     link(rel="stylesheet", href="static/css/v_zoom.css")
 
     script(src="https://use.fontawesome.com/releases/v5.0.13/js/all.js", defer=True)
+    script(src="https://cdn.jsdelivr.net/npm/@fortawesome/vue-fontawesome@0.1.9/index.min.js")
     script(src="https://cdn.jsdelivr.net/npm/vue@2.5/dist/vue.js")
     script(src="https://cdn.jsdelivr.net/npm/jszip@3.1/dist/jszip.min.js")
 
--- a/frontend/src/index.pug
+++ b/frontend/src/index.pug
@@ -59,7 +59,7 @@ html
         #tabs-wrapper(v-if="isTabActive")
           #tab-resize(onmousedown="registerMouseMove()")
           .tab-close(v-on:click="deactivateTab")
-            i.fas.fa-caret-right
+            font-awesome-icon(icon="caret-right")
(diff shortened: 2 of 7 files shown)


frontend/src/summary.pug
```diff
@@ -154,23 +158,34 @@
             v-bind:class="{ warn: user.name === '-' }"
           ) {{ user.displayName }} ({{ user.name }})
           .summary-chart__title--contribution.mini-font [{{ user.totalCommits }} lines]
+          a(
+            v-if="filterGroupSelection !== 'groupByRepos'"
```
anubh-v: I think you're missing a comma at the end of the line, which is causing the errors from pug-lint
Here is more info on the error: [link](https://github.com/pugjs/pug-lint/blob/master/docs/rules.md#validateattributeseparator-string--object)

ccyccyccy: Oops I forgot to check the travis result. Thanks for the catch :)

jamessspanggg:
One qn: what would be the downsides if we were to use [vue fontawesome icons](https://github.com/FortAwesome/vue-fontawesome), which alleviates the need for the span workaround?

ccyccyccy:
> 
> 
> One qn: what would be the downsides if we were to use [vue fontawesome icons](https://github.com/FortAwesome/vue-fontawesome), which alleviates the need for the span workaround?

Depends on how the library works, but we'll need to remember to remove the useless spans after using the new library.

jamessspanggg:
> Depends on how the library works, but we'll need to remember to remove the useless spans after using the new library.

I was thinking whether we should just switch all the icons to vue fontawesome, in this way we do not need to add and remove again. @eugenepeh @fzdy1914 may I get ur opinion on this?

fzdy1914:
I prefer to use vue fontawesome directly.

eugenepeh:
>I was thinking whether we should just switch all the icons to vue fontawesome, in this way we do not need to add and remove again. @eugenepeh @fzdy1914 may I get ur opinion on this?

Sounds good to me.

ccyccyccy:
I changed to use vue-fontawesome for all the icons in the repo in this PR. Should I split it to another PR? I think it's good to merge it into this PR since this PR heavily relies on the icons being reactive.

jylee-git:
Functionality seems to be fine.

Resolve the conflicts and request for review again

ccyccyccy:
> Functionality seems to be fine.
> 
> Resolve the conflicts and request for review again

Fixed.

PR #1187, opened 2020-03-26
[#436] Give warning for duplicate alias

### LINKED ISSUE #436: Different author with same alias will not trigger any warning

Tejas2805:
Fixes #436 

```
Duplicate aliases in the csv files or standalone config, 
do not trigger any warning. As a result,the user might 
not get to know that there are duplicate aliases in the csv 
files provided by him/her.

Let's give a warning if there are duplicate aliases in the csv 
file or standalone config as the alias will belong to the last 
person who claims it. 

```


Tejas2805:
The lines which I have removed were redundant.

Tejas2805:
@fzdy1914 Need your review on this. 

You can also see #1189 for another way to fix this issue, instead of bypassing the second check.

anubh-v:
Will be ideal to let #1189 be resolved first

fzdy1914:
@Tejas2805 Why  #1189 can solve this issue?

Tejas2805:
@fzdy1914 If we think #1189 is a good change to make, then I don't need to bypass the duplicate check since the duplicate check will be removed with PR #1189 . This is why #1189 needs to be decided first.

Let me know how we plan to proceed.

anubh-v:
@fzdy1914, please take a look at [this comment which explains the relationship between issue #436 and issue #1188](https://github.com/reposense/RepoSense/pull/1189#issuecomment-605432570)

Tejas2805:
On hold, till PR #1189 merges.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -198,7 +213,7 @@ public void setAuthorList(List<Author> authorList) {
     public void resetAuthorInformation() {
         authorEmailsAndAliasesMap.clear();
         authorDisplayNameMap.clear();
-        authorList.forEach(this::setAuthorDetails);
+        authorList.forEach(author -> setAuthorDetails(author));
```
jamessspanggg, line 216: is this change needed?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -63,6 +66,18 @@ public void update(StandaloneConfig standaloneConfig, List<String> ignoreGlobLis
...
 
+    /**
+     * Checks for duplicate aliases in {@code authorEmailsAndAliasesMap} and generates warnings
+     * @param authorEmailsAndAliasesMap
+     * @param alias
+     */
+    public void checkDuplicateAliases(Map<String, Author> authorEmailsAndAliasesMap, String alias) {
+        if (authorEmailsAndAliasesMap.containsKey(alias)) {
+            logger.warning(String.format(
+                    "Duplicate alias %s found. The alias will belong to the last author who claims it", alias));
```
jamessspanggg, line 77: I think can provide the last author's git id in the log as well, just to let the user know the specific author that gets the alias

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -213,8 +228,16 @@ public void setAuthorDisplayName(Author author, String displayName) {
...
     }
 
-    public void addAuthorEmailsAndAliasesMapEntry(Author author, List<String> values) {
-        values.forEach(value -> authorEmailsAndAliasesMap.put(value, author));
+    /**
+     * Adds {@code aliases} of {@code author} into the map
+     * @param author
+     * @param aliases
+     */
+    public void addAuthorEmailsAndAliasesMapEntry(Author author, List<String> aliases) {
```
jamessspanggg, line 236: why the change from values -> aliases? If I remember correctly, the map is not just `alias -> author` right?

src/main/java/reposense/parser/AuthorConfigCsvParser.java
```diff
@@ -105,7 +104,6 @@ private static AuthorConfiguration findMatchingAuthorConfiguration(
      */
     private static void setEmails(AuthorConfiguration config, Author author, List<String> emails) {
         author.setEmails(new ArrayList<>(emails));
-        config.addAuthorEmailsAndAliasesMapEntry(author, author.getEmails());
```
jamessspanggg, line 108: may i get the reason on why these `config.addAuthor..` lines are removed?

Tejas2805: All these are redundant as all of these are already set by the parent function before these functions are called. And these functions are called only one by one parent function.  Therefore, removed them.

src/main/java/reposense/model/AuthorConfiguration.java
Tejas2805: I might have to send one extra variable through all functions for this. Works? 
Not sure if I have the information currently to display the gitId without the extra variable

src/main/java/reposense/model/AuthorConfiguration.java
Tejas2805: I did the change thinking code will be more readable. But can remove too.

src/main/java/reposense/model/AuthorConfiguration.java
jamessspanggg: yes or you can just add the `if` check within the `update` method, instead of creating another method. Don't think it adds much value to separate little logic to another method.

Tejas2805: Actually I am calling the function from two places, therefore, extracted out to a function. One for csv, one for config.json

jamessspanggg: ok, then you'll have to pass the `author id` to the checkDuplicateAliases method

src/main/java/reposense/model/AuthorConfiguration.java
jamessspanggg: dont think its necessary for this change

src/main/java/reposense/parser/AuthorConfigCsvParser.java
anubh-v: By "parent function", are you referring to `processLine` ? 
I think the removed lines are not redundant.
In `processLine`, we indeed call `config.addAuthor` which technically does store the author's emails and aliases via the `addAuthorEmailsAndAliasesMapEntry` method.

However, the complication is that the `processLine` method calls `config.addAuthor`  **before** the `Author` object is initialised with any emails and aliases.  This means that no emails and aliases are stored in the `AuthorConfiguration` when `config.addAuthor` is called - because the `Author` object is not fully initialised yet.

Hence, the removed lines 108, 124, 130 are not redundant. 


**Unrelated to this PR** -  the flow of events mentioned above should be refactored in my view.
We should initialise the `Author` instances before calling `config.addAuthor`.

Tejas2805: Alright! I just checked this. I agree. We cannot remove this as of now. Will open an issue for the refactor.

anubh-v:
@Tejas2805, could you update the proposed commit message, so all reviewers know that the warnings apply to duplicate aliases in both the standalone json config, and the author config csv ? 
Thanks!

Tejas2805:
> @Tejas2805, could you update the proposed commit message, so all reviewers know that the warnings apply to duplicate aliases in both the standalone json config, and the author config csv ?
> Thanks!

Done.

Tejas2805:
@anubh-v Can I get your review on this?

anubh-v:
@Tejas2805, there seem to be some false positives in the warnings.
I tried this PR with these configs:

**repo-config.csv**
```
Repository's Location,Branch,File Formats,Ignore Glob List,Ignore standalone config,Ignore Commit List,Ignore Authors List
https://github.com/anubh-v/test-repo.git, master,,,yes,,
```

**author-config.csv**
```
Repository's Location,Branch,Author's GitHub ID,Author's Emails,Author's Display Name,Author's Git Author Name,Ignore Glob List
https://github.com/anubh-v/test-repo.git, master, tom,,anubhav,anubh-v,
https://github.com/anubh-v/test-repo.git, master, bob,,bob,anubh-v,
```

The user **bob** has the same alias as the user **tom**.

You can download the configs here:
[author-config.txt](https://github.com/reposense/RepoSense/files/4587256/author-config.txt)
[repo-config.txt](https://github.com/reposense/RepoSense/files/4587257/repo-config.txt)


When I ran RepoSense, there are some extra warnings generated for both users' emails:
RepoSense output:
```
10:07:48 - [WARNING] Duplicate alias tom@users.noreply.github.com found. The alias will belong to the last author - tom
10:07:48 - [WARNING] Duplicate alias tom found. The alias will belong to the last author - tom
10:07:48 - [WARNING] Duplicate alias bob@users.noreply.github.com found. The alias will belong to the last author - bob
10:07:48 - [WARNING] Duplicate alias bob found. The alias will belong to the last author - bob
10:07:48 - [WARNING] Duplicate alias anubh-v found. The alias will belong to the last author - bob
10:07:48 - [WARNING] Duplicate alias anubh-v found. The alias will belong to the last author - bob
10:07:48 - Cloning in parallel from https://github.com/anubh-v/test-repo.git...
10:07:48 - Waiting for cloning of https://github.com/anubh-v/test-repo.git to complete...
10:07:50 - Cloning of https://github.com/anubh-v/test-repo.git completed!
10:07:50 - [1/1] Analyzing https://github.com/anubh-v/test-repo.git (master)...
10:07:50 - Ignoring standalone config file in https://github.com/anubh-v/test-repo.git (master).
10:07:50 - Extracting commits info for https://github.com/anubh-v/test-repo.git (master)...
10:07:50 - Analyzing commits info for https://github.com/anubh-v/test-repo.git (master)...
10:07:50 - Extracting relevant file info from https://github.com/anubh-v/test-repo.git (master)...
10:07:50 - Analysis of https://github.com/anubh-v/test-repo.git (master) completed!
10:07:50 - The report is generated at /home/bhav/Desktop/RepoSense/reposense-report
10:07:50 - Elapsed processing time: 2.17 second(s)

```

I think this is because the internal Map of the AuthorConfiguration is being updated twice (once in `AuthorConfiguration#addAuthor` and once in the `AuthorConfigCsvParser#processLine`)

fzdy1914:
To be merged after #1270

fzdy1914:
@Tejas2805 Can u have a check to see if the error warning is fixed or not? If yes,  I will merge it now.

Tejas2805:
> @Tejas2805 Can u have a check to see if the error warning is fixed or not? If yes, I will merge it now.

Well it's fixed. All false positives gone. But, the original issue of double warnings still persist. The PR #1189 which we thought should fix this issue, hasn't.

PR #1189, opened 2020-03-26
[#1188] Improve ignore authors implementation 

### LINKED ISSUE #1188: AuthorConfiguration: all authors' information reset when ignored authors are removed
In the snippet below, we see that the `resetAuthorInformation` method is called after ignored authors are removed.

`resetAuthorInformation` will re-populate the `AuthorConfiguration` with the information of all authors except the ignored authors.

I wish to clarify:
- Is there a reason we reset **all** authors' information, rather than only removing the ignored authors' information?
- It seems that `resetAuthorInformation` will run even when no authors are ignored. Is this intended?

To me, it seems simpler to selectively remove the ignored authors' information, rather than resetting all authors' information. Hence, I wanted to clarify if there's a reason for the current design

https://github.com/reposense/RepoSense/blob/e0e912a58a3b4391522b1107dcbd5c789e04af11/src/main/java/reposense/model/AuthorConfiguration.java#L142-L152

Tejas2805:
Fixes #1188 

```
The ignored authors are removed from the list by clearing the 
authorDisplayNameMap and authorEmailsAndAliasesMap and 
setting the authorList all again.

Let's improve this implementation by just removing the authors to
be ignored.

All the key-value bindings of the ignoredAuthor need to be removed.
```


Tejas2805:
I have updated the ignore author list for test purposes. Will remove it once PR is approved.

Tejas2805:
@jylee-git @fzdy1914 Can I get your take on this. Thank you for your time.

jylee-git:
> @jylee-git @fzdy1914 Can I get your take on this. Thank you for your time.
> 
> This will affect the implementation of PR #1187 which is also ready

Can you provide a summary as to why the previous implementation didn't work which require this PR to improve the ignore author implementation? That would help us in providing feedback

Tejas2805:
> Can you provide a summary as to why the previous implementation didn't work which require this PR to improve the ignore author implementation? That would help us in providing feedback

Sorry for not doing this earlier.

> Perhaps I can help clarify.
> 
> The issue with the current implementation is that too many warnings are generated.
> Let's use [this repo](https://github.com/anubh-v/test-repo) for discussion purposes.
> 
> Here is the [json config](https://github.com/anubh-v/test-repo/blob/master/_reposense/config.json).
> Since the alias `anubh-v` is repeated 3 times, it would be reasonable to see 2 warnings from RepoSense.
> 
> When we run this PR using the `--repo ` parameter (no CSV configs are involved), we get:
> image
> 
> i.e. there are 2 more warnings than expected.
> 
> #### Reason for the extra warnings
> As mentioned by @Tejas2805 [here](https://github.com/reposense/RepoSense/pull/1153#issuecomment-603826392), the `resetAuthorInformation` method in `AuthorConfiguration` causes these extra warnings
> 
> This method clears the _existing_ author information in the `AuthorConfiguration` instance and replaces it with author information for a _new_ list of authors. It is called _after_ we remove ignored authors from the list of authors in the `AuthorConfiguration` instance.
> 
> Because `resetAuthorInformation` causes the `AuthorConfiguration`'s internal TreeMap to be re-populated, it triggers the duplicate checking again --> resulting in the warnings being logged again.
> 
> One way to resolve the issue may be to change how we handle ignoring of authors.
> Instead of resetting **all** author information and re-populating `AuthorConfiguration` 's internal TreeMap, we can try only resetting the author info of the ignored authors.
> 
> This would ensure the duplicates' check isn't run twice

As can be seen from above, the implementation of `resetAuthorInformation` is problematic (not wrong) and can be simplified to just removing the ignored author. This will also help in PR #1187  as after this new implementation, I will just have to add a check for the duplicate aliases. Whereas, right now I am trying to bypass the check when we set everything for the second time. 

So we can proceed ahead in two ways:

1) We change the implementation of `resetAuthorInformation` and just add checks for duplicate warnings.

If we proceed with this, we merge this PR first and then the other PR.

2) We keep the implementation of `resetAuthorInformation` same and bypass the check the second time. 

If we proceed with this, we just close this PR and go ahead with PR #1187 which already has this implementation.

Tejas2805:
@fzdy1914 @jylee-git Any updates? Thank you for your time.

jylee-git:
Implementation wise, I would prefer to go with point 1 as we deal with the root of the issue (i.e.: deal with this PR and also #1187).

Tejas2805:
@anubh-v Can you help me figure why the test is failing? Cause the logic seems to be fine

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,21 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
+    }
+
+    /**
+     * Removes information of the {@code author} along with its {@code aliases}
+     */
+    public void removeIgnoredAuthorsInformation(String author, List<String> aliases, Author ignoredAuthor) {
+        authorList.remove(ignoredAuthor);
+        authorDisplayNameMap.remove(ignoredAuthor);
+        authorEmailsAndAliasesMap.remove(author);
+        aliases.forEach(alias -> authorEmailsAndAliasesMap.remove(alias, ignoredAuthor));
```
anubh-v: Related to the test failure: 

- I think we'll need to remove the key-value bindings for ignoredAuthors' emails as well
- Is there a guarantee that the `author` parameter == the git ID of the `ignoredAuthor`? From the code, it seems `author` could actually be the email or an alias or ignoredAuthor. If that's the case, we need to remove the key-value binding for ignoreAuthor's git ID

- Actually, do we need the `author` parameter at all? I think all the info needed to be removed (aliases, emails, git id) is available in `ignoredAuthor` right?

anubh-v: Line 161 and 162 use different variants of the `remove` method
Both should be suitable, but I think `Map.remove(Object key)` is simpler here. What do you think?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,21 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
     public void removeIgnoredAuthors(List<String> ignoredAuthorsList) {
         for (String author : ignoredAuthorsList) {
             if (authorEmailsAndAliasesMap.containsKey(author)) {
-                authorList.remove(authorEmailsAndAliasesMap.get(author));
+                Author ignoredAuthor = authorEmailsAndAliasesMap.get(author);
+                List<String> aliases = new ArrayList<>(ignoredAuthor.getAuthorAliases());
```
anubh-v: Why create a new ArrayList here?

src/main/java/reposense/model/AuthorConfiguration.java
Tejas2805: So, the thing is - Author, alias, email all are set as `keys`. So we need to remove them all. 
As pointed by you, email has been missed out. So I will include that. 

We can use `Map.remove(Object key)` as we don't have duplicates anyway.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,26 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
+     * Removes all information of the {@code author} from the configs
+     */
+    public void removeIgnoredAuthorsInformation(String author) {
+        Author ignoredAuthor = authorEmailsAndAliasesMap.get(author);
+        authorList.remove(ignoredAuthor);
+        authorDisplayNameMap.remove(ignoredAuthor);
+        authorEmailsAndAliasesMap.remove(author);
+        authorEmailsAndAliasesMap.remove(ignoredAuthor.getGitId());
+
+        List<String> aliases = new ArrayList<>(ignoredAuthor.getAuthorAliases());
```
anubh-v: a clarification: why do we need to create a new arraylist?

Tejas2805: I felt the code would be more readable if we create an `aliases` list first. And then use the list. Instead of working directly by calling it.

anubh-v: I see. Doesn't `getAuthorAliases` return a list though?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,26 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
+                removeIgnoredAuthorsInformation(author);
             }
         }
-        resetAuthorInformation();
+    }
+
+    /**
+     * Removes all information of the {@code author} from the configs
+     */
+    public void removeIgnoredAuthorsInformation(String author) {
```
anubh-v: Is it simpler to name the method `removeAuthor`?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,26 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
             if (authorEmailsAndAliasesMap.containsKey(author)) {
-                authorList.remove(authorEmailsAndAliasesMap.get(author));
+                removeIgnoredAuthorsInformation(author);
             }
         }
-        resetAuthorInformation();
+    }
+
+    /**
+     * Removes all information of the {@code author} from the configs
```
anubh-v: I think the Javadoc should say whether `author` is a git ID, email, alias (or any of these)

Tejas2805: Done.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,27 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
+     * @param author Can be an author's git ID, email, or alias
+     */
+    public void removeAuthor(String author) {
+        Author ignoredAuthor = authorEmailsAndAliasesMap.get(author);
```
anubh-v: Not sure if `ignoredAuthor` is the best name here, perhaps `authorToRemove` ?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,27 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
-                authorList.remove(authorEmailsAndAliasesMap.get(author));
+                removeAuthor(author);
             }
         }
-        resetAuthorInformation();
+    }
+
+    /**
+     * Removes all information of the {@code author} from the configs
+     * @param author Can be an author's git ID, email, or alias
```
anubh-v: Perhaps we should state that `author` must be in the `authorEmailsAndAliasesMap` as a pre-condition

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,27 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
+     */
+    public void removeAuthor(String author) {
+        Author ignoredAuthor = authorEmailsAndAliasesMap.get(author);
+        authorList.remove(ignoredAuthor);
+        authorDisplayNameMap.remove(ignoredAuthor);
+        authorEmailsAndAliasesMap.remove(author);
+        authorEmailsAndAliasesMap.remove(ignoredAuthor.getGitId());
```
anubh-v: A clarification: if `author` is the same as `gitID`, then we'll be removing the key-value binding twice.
Would this be an issue?
Is it simpler to omit line 161 entirely, since we're also removing the key-value bindings for git ID, aliases and emails?

src/main/java/reposense/model/AuthorConfiguration.java
Tejas2805: Ya, can include.

src/main/java/reposense/model/AuthorConfiguration.java
Tejas2805: We can do this, I think. Just that the current one makes sure, nothing is left, at all. Let me take a check on the code and get back on this.

Tejas2805: **Update:** Have removed it. We are removing the rest explicitly. So should be fine.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,27 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
+     * @param author Can be an author's git ID, email, or alias
+     */
+    public void removeAuthor(String author) {
+        Author authorToRemove = authorEmailsAndAliasesMap.get(author);
+        authorList.remove(authorToRemove);
+        authorDisplayNameMap.remove(authorToRemove);
+        authorEmailsAndAliasesMap.remove(authorToRemove.getGitId());
+
+        List<String> aliases = authorToRemove.getAuthorAliases();
+        aliases.forEach(alias -> authorEmailsAndAliasesMap.remove(alias));
```
jamessspanggg, line 165: why is there a need to remove both alias and emails in the map? are the key value pairs not always `email:alias`?

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,27 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
...
         }
-        resetAuthorInformation();
+    }
+
+    /**
+     * Removes all information of the {@code author} from the configs
+     * Precondition: {@code author} must be present in {@code authorEmailsAndAliasesMap}
+     * @param author Can be an author's git ID, email, or alias
+     */
+    public void removeAuthor(String author) {
```
jamessspanggg, line 158: perhaps `removeAuthorInformation`?

src/main/java/reposense/model/AuthorConfiguration.java
Tejas2805: Nope. The key values are: `alias`, `emails` and `gitID`

jamessspanggg: Then the variable name does not reflect the underlying structure of the map? According to the name, it feels like it should be `email:alias` instead. Perhaps can change the variable name to reflect the key -> value pairs?

Tejas2805: If referring to `authorEmailsAndAliasesMap`, it's the one used everywhere. I have just continued the convention. It's a pre-defined Map.

jamessspanggg: Perhaps can create an issue for the inaccurate naming of variable `authorEmailsAndAliasesMap`, since the underlying structure is:
```
git id -> author
email -> author
alias -> author
```

Tejas2805:
@fzdy1914 @jylee-git  Can I get your review on this? Thank you for your time.

Tejas2805:
@jylee-git Can I get your review on this. Thank you for your time.

Tejas2805:
@fzdy1914 Can be merged now.

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -145,10 +145,27 @@ public void addAuthor(Author author, List<String> ignoreGlobList) {
     public void removeIgnoredAuthors(List<String> ignoredAuthorsList) {
         for (String author : ignoredAuthorsList) {
             if (authorEmailsAndAliasesMap.containsKey(author)) {
-                authorList.remove(authorEmailsAndAliasesMap.get(author));
+                removeAuthorInformation(author);
             }
         }
-        resetAuthorInformation();
```
jamessspanggg, line 151: I noticed that `resetAuthorInformation()` is used in `RepoConfiguration#setAuthorList`, could we utilise your `removeAuthorInformation` instead?

anubh-v: I think `removeAuthorInformation` is not a direct substitute for `resetAuthorInformation`, since resetting also involves re-building the internal maps in author config for the authors in the author list.

jamessspanggg: ok noted

anubh-v: As we discussed in issue #1204, the current naming of `resetAuthorInformation` is a bit confusing. We may break it into 2 methods: `clear` and `buildFromAuthorList` (as mentioned in the isue)

PR #1192, opened 2020-03-27
[#1181] Allow filtering of ramps by file type

### LINKED ISSUE #1181: Filter ramps by file type
Current: all ramps are shown even when some file types have been filtered out
image

Suggestion: show ramps that are relevant to the selected file types only

jamessspanggg:
Fixes #1181 

```
All ramps are shown even when some file types are filtered out, the
sizes of ramps are also not adjusted when their filetypes are 
filtered out.

This may cause confusion in users as they would think that the ramps 
will be filtered out accordingly.

Let's add the filtering of ramps by the file types that are selected, 
the size of the ramp will also be adjusted based on that.

Additional Info:
1. The current git log results does not include filetypes that are
modified, so I decided to get from the backend by changing the
git log format. The filetype details of each commit are also added
to the commit.json

2. 'git log --numstat --shortstat' is used instead of 'git log --stat'
to retrieve the log with files changed. This is because --stat only 
shows the sum of lines changed in file whereas --numstat displays 
addition and deletion separately. Addition and deletion of binary
files will be '-' and '-'.

3. When removing a filetype from a ramp, the corresponding addition
and deletion will be removed from the ramp. A ramp will only be 
removed when all of it's filetypes are removed.
```


src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -29,8 +32,9 @@
     private static final Logger logger = LogsManager.getLogger(CommitInfoAnalyzer.class);
     private static final String MESSAGE_START_ANALYZING_COMMIT_INFO = "Analyzing commits info for %s (%s)...";
 
-    private static final String LOG_SPLITTER = "\\|\\n\\|";
+    private static final String LOG_SPLITTER = "\\?\\n\\?";
```
jamessspanggg: Need advice here, I am switching the delimiter to be `?` instead of `|`, this is because the `git log ... --stat` returns the files changed in the following format:
`<file path>    | <num of lines changed>`,
which overlaps if we stick with using `|` as a delimiter.

Don't think there's a way to change the format of the `git log ... --stat`, which is why i decided to change the delimiter.

jamessspanggg:
prof @damithc, I have modified the implementation such that the unchecking the filetypes will decrement the corresponding addition and deletion of the ramp correspondingly.

Do note that, _the size of the ramp is proportionate to the total contribution of all ramp charts_. So it doesn't mean that unchecking a filetype that is contained within the ramp will decrease the size of that ramp. It depends.

Would you prefer:
- the current implementation, or 
- the previous implementation where the size of the ramp will not be affected as you uncheck filetypes, it will only remove the ramp when all of it's filetypes are unchecked

damithc:
> Would you prefer:
> 
> * the current implementation, or
> * the previous implementation where the size of the ramp will not be affected as you uncheck filetypes, it will only remove the ramp when all of it's filetypes are unchecked

The first one, as it is closer to the truth.

jamessspanggg:
> The first one, as it is closer to the truth.

Noted.

frontend/src/static/js/v_summary.js
```diff
@@ -650,12 +650,28 @@ window.vSummary = {
       user.dailyCommits.forEach((commit) => {
         const { date } = commit;
         if (date >= sinceDate && date <= untilDate) {
-          user.commits.push(commit);
+          const filteredCommit = JSON.parse(JSON.stringify(commit));
```
jamessspanggg: This is to ensure that the filtering of `commitResults` within `commit` does not affect the original `commit`.

src/main/java/reposense/git/GitLog.java
```diff
@@ -43,7 +43,7 @@ public static String getWithFiles(RepoConfiguration config, Author author) {
 
         String command = "git log --no-merges -i ";
         command += GitUtil.convertToGitDateRangeArgs(config.getSinceDate(), config.getUntilDate());
-        command += " --pretty=format:" + PRETTY_FORMAT_STRING + " --stat";
+        command += " --pretty=format:" + PRETTY_FORMAT_STRING + " --numstat --shortstat";
```
jamessspanggg: I changed from `--stat` to `--numstat --shorstat` is because:
- `--stat` only shows the sum of lines changed in file (instead of display addition and deletion separately`
- `--numstat --shortstat`'s files changed is easier to parse as the format is `<#addition>\t<#deletion>\t<file path>`

src/main/java/reposense/commits/CommitInfoExtractor.java
```diff
@@ -59,9 +60,9 @@
             Matcher matcher = TRAILING_NEWLINES_PATTERN.matcher(rawCommitInfos[i]);
             String rawCommitInfo = matcher.replaceAll("");
 
-            int statLineSeparatorIndex = rawCommitInfo.lastIndexOf("|");
+            int statLineSeparatorIndex = rawCommitInfo.lastIndexOf(INFO_STAT_SEPARATOR);
             String infoLine = rawCommitInfo.substring(0, statLineSeparatorIndex);
-            String statLine = rawCommitInfo.substring(statLineSeparatorIndex + 1);
+            String statLine = rawCommitInfo.substring(statLineSeparatorIndex + 1).trim();
```
jamessspanggg: `.trim()` is to remove the first new line character in `rawCommitInfo.substring(statLineSeparatorIndex + 1)`

jamessspanggg:
@anubh-v may I get your review on the back end implementation?

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +101,64 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
 
-        int insertion = getInsertion(statLine);
-        int deletion = getDeletion(statLine);
-        return new CommitResult(author, hash, date, messageTitle, messageBody, tags, insertion, deletion);
+        if (statLine.isEmpty()) { // empty commit, no files changed
+            return new CommitResult(author, hash, date, messageTitle, messageBody, tags, 0, 0, new HashMap<>());
+        }
+
+        String[] statInfos = statLine.split(NEW_LINE_SPLITTER);
+        String[] fileTypeContributions = Arrays.copyOfRange(statInfos, 0, statInfos.length - 1);
```
anubh-v: Clarification, the array is copied here because you do not want to include the last element of `statInfos` ?

jamessspanggg: yes. I did so that it is clearer

src/main/java/reposense/commits/model/CommitResult.java
```diff
@@ -29,6 +32,7 @@ public CommitResult(Author author, String hash, Date time, String messageTitle,
         this.tags = tags;
         this.insertions = insertions;
```
anubh-v: Unrelated to this PR, but should we use assertions to check if insertions and deletions are non negative?

jamessspanggg: the insertions and deletions are taken from git log, which they probably won't be negative. But for defensive programming, that would be good

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +101,64 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
+        }
+        return fileTypesAndContribution;
+    }
+
+    /**
+     * Extracts the correct file path from the pre-processed git log {@code filePath}
+     */
+    private static String extractFilePath(String filePath) {
+        String filteredFilePath = filePath;
+        if (filteredFilePath.contains(MOVED_FILE_INDICATION)) { // moved file has the format: fileA => newPosition/fileA
```
anubh-v: For this method, if `filePath` does not contain the `=>` string, then we'll just return `filePath` right?
In that case, I think it'll be simpler if we change the if condition to 
`if (!filteredFilePath.contains(MOVED_FILE_INDICATION))` and then just return `filePath` in the block.

Then, the rest of the method can handle the logic for extracting the filepath further

jamessspanggg: Yep changed accordingly

anubh-v: Do we need the moved file indication? Any idea if this info can be omitted in git log?

jamessspanggg: Don't think it can be omitted in git log as it does not have an option to edit the format of the _file paths_. More info [here](https://git-scm.com/docs/git-log)

0blivious:
1. Select breakdown by file type
2. untick java
The 1st ramp of Eugene changes though java is not in the group file type

jamessspanggg:
> 1. Select breakdown by file type
> 2. untick java
>    The 1st ramp of Eugene changes though java is not in the group file type

It changes because the size of all ramps are proportionate to the **_total contribution of all ramp charts_**.

0blivious:
> It changes because the size of all ramps are proportionate to the **_total contribution of all ramp charts_**. So even if u untick a file type that is not related to the ramp, it may change in size as the unchecking decreases the total contribution.

So the size depends on total charts from all groups? Is it very troublesome to make it depends on only the total contribution of its own group? As by intuition I thought the size should not change if I am not excluding anything from the ramp.

jamessspanggg:
> So the size depends on total charts from all groups? 

Yes it is. Both length of contribution bar and ramp size are dependent on the total contribution of all groups.

> Is it very troublesome to make it depends on only the total contribution of its own group? As by intuition I thought the size should not change if I am not excluding anything from the ramp.

I think the decision to make the size dependent on all charts is because you want to standardise a base line on how big/long the ramp/contribution bar will be. 

This is because you don't want a ramp (let's say 1000 lines) from a group (let's say total group contribution of 10,000 lines) that has the **same size** as another ramp (let's say 10 lines) from another group (let's say total group contribution of 100 lines).

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +101,65 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
+        return new CommitResult(author, hash, date, messageTitle, messageBody, tags, totalCommitInsertions,
+                totalCommitDeletions, fileTypeAndContributionMap);
+    }
+
+    /**
+     * Extract the additions and deletions of file types that have been modified
+     */
+    private static Map<FileType, Map<String, Integer>> getFileTypesAndContribution(String[] filePathContributions,
+            RepoConfiguration config) {
+        Map<FileType, Map<String, Integer>> fileTypesAndContribution = new HashMap<>();
```
ang-zeyu: Since the inner `Map<String, Integer>` only contains two possible keys, perhaps it could be extracted into a new model?

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +101,65 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
+        for (String filePathContribution : filePathContributions) {
+
+            String[] infos = filePathContribution.split(TAB_SPLITTER);
+            int addition = Integer.parseInt(infos[0]);
+            int deletion = Integer.parseInt(infos[1]);
+            String filePath = extractFilePath(infos[2]);
```
ang-zeyu: I'm personally fine with magic numbers here as the variable names are quite clear, could declare them as constants if it isn't intentional though

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +101,65 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
+            if (fileTypesAndContribution.containsKey(fileType)) {
+                // update existing file type contribution
+                Map<String, Integer> prevContribution = fileTypesAndContribution.get(fileType);
+                prevContribution.replace(INSERTIONS, prevContribution.get(INSERTIONS) + addition);
+                prevContribution.replace(DELETIONS, prevContribution.get(DELETIONS) + deletion);
+            } else {
+                Map<String, Integer> contributionMap = new HashMap<>();
+                contributionMap.put(INSERTIONS, addition);
+                contributionMap.put(DELETIONS, deletion);
+                fileTypesAndContribution.put(fileType, contributionMap);
```
ang-zeyu: following the `Map<String, Integer>` model above, perhaps this could be rewritten into 'guard-clause' style
```
if fileTypesAndContriibution does not have fileType
  initiate fileType model
common logic
```

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +101,65 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
+
+    /**
+     * Extracts the correct file path from the unprocessed git log {@code filePath}
+     */
+    private static String extractFilePath(String filePath) {
+        if (!filePath.contains(MOVED_FILE_INDICATION)) {
+            return filePath;
+        }
+        // moved file has the format: fileA => newPosition/fileA
+        String filteredFilePath = filePath.substring(filePath.indexOf(MOVED_FILE_INDICATION) + 3);
```
ang-zeyu: how about `MOVED_FILE_INDICATION.length()`?

src/main/java/reposense/commits/CommitInfoAnalyzer.java
jamessspanggg: Yep, declaring them as constants will be clearer

src/main/java/reposense/commits/CommitInfoAnalyzer.java
jamessspanggg: Yep extracting to a new model will be much cleaner.

Extracted this to a new model `ContributionPair`. Note that there are other places where we can utilise this `ContributionPair` class such as `AuthorDailyContribution`, `CommitResult` etc. (since they both use insertion and deletion variables). But this can be a future PR to refactor them to use `ContributionPair`.

Tejas2805: If this is really the case, can create an issue on this. Will help keep a track.

jamessspanggg:
> Some of the lines in the commit message are crossing 72 characters. Can you correct them?

done

src/main/java/reposense/commits/CommitInfoAnalyzer.java
jamessspanggg: Will do when this is merged

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +104,62 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
+    }
+
+    /**
+     * Extract the additions and deletions of file types that have been modified
+     */
+    private static Map<FileType, ContributionPair> getFileTypesAndContribution(String[] filePathContributions,
+            RepoConfiguration config) {
+        Map<FileType, ContributionPair> fileTypesAndContributionMap = new HashMap<>();
+        for (String filePathContribution : filePathContributions) {
+
```
ang-zeyu: extra new line here

src/test/java/reposense/util/TestUtil.java
```diff
@@ -199,7 +199,7 @@ public static boolean compareNumberFilesChanged(int expectedNumberFilesChanged,
      * Returns the file changed given a {@code rawFileChangedString}.
      */
     private static String getFileChanged(String rawFileChangedString) {
-        String fileChanged = rawFileChangedString.split("\\| ")[0].trim();
+        String fileChanged = rawFileChangedString.split("\t")[2].trim();
```
anubh-v: Would it be suitable to reuse the constants `TAB_SPLITTER` and `STAT_FILE_PATH_INDEX` here?

src/main/java/reposense/commits/model/ContributionPair.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public int getInsertions() {
+        return insertions;
+    }
+
+    public int getDeletions() {
+        return deletions;
+    }
+
+    public void setInsertions(int insertions) {
```
anubh-v: Is there a need for these setter / getters in the current PR? If not, perhaps we can introduce them when they are needed in future?

jamessspanggg: Removed

src/test/java/reposense/util/TestUtil.java
jamessspanggg: yep utilised them

jamessspanggg:
@jylee-git can i get ur review on this?

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +104,61 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
+        int totalCommitDeletions = getDeletion(contributionStat);
+
+        return new CommitResult(author, hash, date, messageTitle, messageBody, tags, totalCommitInsertions,
+                totalCommitDeletions, fileTypeAndContributionMap);
+    }
+
+    /**
+     * Extract the additions and deletions of file types that have been modified
+     */
+    private static Map<FileType, ContributionPair> getFileTypesAndContribution(String[] filePathContributions,
```
jylee-git, line 127:
```suggestion
    /**
     * Returns the number of lines added and deleted for the specified file types in {@code config}.
     */
    private static Map<FileType, ContributionPair> getFileTypesAndContribution(String[] filePathContributions,
```
?

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +104,61 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
...
+    private static String extractFilePath(String filePath) {
+        if (!filePath.contains(MOVED_FILE_INDICATION)) {
+            return filePath;
+        }
+        // moved file has the format: fileA => newPosition/fileA
+        String filteredFilePath =
+                filePath.substring(filePath.indexOf(MOVED_FILE_INDICATION) + MOVED_FILE_INDICATION.length());
+        if (filteredFilePath.charAt(filteredFilePath.length() - 1) == '}') { // renamed file has ending '}' char
+            filteredFilePath = filteredFilePath.substring(0, filteredFilePath.length() - 1);
+        }
```
jylee-git, line 160:
```suggestion
        // Removes the trailing '}' character from the file name if it exists.
        filteredFilePath = filteredFilePath.replaceAll("}$", "");
```

Btw, will there be any issue when a file is renamed to one ending with a `}`?

src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
```diff
@@ -25,6 +29,10 @@
     private static final int NUMBER_EUGENE_COMMIT = 1;
     private static final int NUMBER_MINGYI_COMMIT = 1;
     private static final int NUMBER_EMPTY_MESSAGE_COMMIT = 1;
+    private static final String INSERTIONS = "insertions";
+    private static final String DELETIONS = "deletions";
+    private static final FileType JAVA = new FileType("java", Collections.singletonList("**java"));
+    private static final FileType MD = new FileType("md", Collections.singletonList("**md"));
```
jylee-git, line 35: Would be better to rename the constants as `FILETYPE_JAVA` and `FILETYPE_MD`.

src/main/java/reposense/commits/CommitInfoAnalyzer.java
jamessspanggg: Yes, that may happen.

But the filePath is only used for getting the filetype, so its fine to remove the ending `}`, as a file type that includes `}` at the end wouldn't make sense.

frontend/src/static/js/v_summary.js
```diff
@@ -492,13 +492,52 @@ window.vSummary = {
...
+          if (this.filterBreakdown) {
```
ccyccyccy, line 496: Should we filter by checked file types even when `filterBreakdown` is not checked? 

i.e. Should the user be able to check `filterBreakdown`, uncheck some file types, and then uncheck `filterBreakdown`, and still view the filtered ramp chart?

frontend/src/static/js/v_summary.js
```diff
@@ -492,13 +492,52 @@ window.vSummary = {
...
+      }).filter((result) => Object.values(result.fileTypesAndContributionMap).length > 0);
+
+      commit.insertions = filteredCommitResults.reduce((acc, result) => acc + result.insertions, 0);
+      commit.deletions = filteredCommitResults.reduce((acc, result) => acc + result.deletions, 0);
+      commit.commitResults = filteredCommitResults;
+    },
+
+    getFilteredFileTypes(commitResult) {
+      return Object.keys(commitResult.fileTypesAndContributionMap)
+          .filter((fileType) => this.isFileTypeIncluded(fileType))
```
ccyccyccy, line 522: Can rename `isFileTypeIncluded` to `isFileTypeChecked` to be clearer.

Also, I think you can just rewrite `.filter((fileType) => this.isFileTypeIncluded(fileType))` as `.filter(this.isFileTypeIncluded)`

frontend/src/static/js/v_summary.js
jamessspanggg: No I don't think we should filter by checked file types when `filterBreakdown` is not checked.

ccyccyccy:
Also, the backend crashes when I run `./gradlew run -Dargs="--repos https://github.com/reposense/RepoSense.git --output output_path/ --ignore-standalone-config --view"`, with the following error log
```
12:41:23 - Cloning in parallel from https://github.com/reposense/RepoSense.git...
12:41:23 - Waiting for cloning of https://github.com/reposense/RepoSense.git to complete...
12:41:40 - Cloning of https://github.com/reposense/RepoSense.git completed!
12:41:40 - [1/1] Analyzing https://github.com/reposense/RepoSense.git (master)...
12:41:40 - Ignoring standalone config file in https://github.com/reposense/RepoSense.git (master).
12:41:40 - https://github.com/reposense/RepoSense.git (master) has no authors specified, using all authors by default.
12:41:40 - Extracting commits info for https://github.com/reposense/RepoSense.git (master)...
12:41:40 - Analyzing commits info for https://github.com/reposense/RepoSense.git (master)...
12:41:40 - [ERROR] java.lang.NumberFormatException: For input string: "-"
        at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
        at java.base/java.lang.Integer.parseInt(Integer.java:642)
        at java.base/java.lang.Integer.parseInt(Integer.java:770)
        at reposense.commits.CommitInfoAnalyzer.getFileTypesAndContribution(CommitInfoAnalyzer.java:132)
        at reposense.commits.CommitInfoAnalyzer.analyzeCommit(CommitInfoAnalyzer.java:114)
        at reposense.commits.CommitInfoAnalyzer.lambda$analyzeCommits$0(CommitInfoAnalyzer.java:68)
        at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
        at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1654)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
        at reposense.commits.CommitInfoAnalyzer.analyzeCommits(CommitInfoAnalyzer.java:72)
        at reposense.commits.CommitsReporter.generateCommitSummary(CommitsReporter.java:21)
        at reposense.report.ReportGenerator.analyzeRepo(ReportGenerator.java:247)
        at reposense.report.ReportGenerator.analyzeRepos(ReportGenerator.java:206)
        at reposense.report.ReportGenerator.cloneAndAnalyzeRepos(ReportGenerator.java:172)
        at reposense.report.ReportGenerator.generateReposReport(ReportGenerator.java:100)
        at reposense.RepoSense.main(RepoSense.java:69)

```

jamessspanggg:
> Also, the backend crashes when I run `./gradlew run -Dargs="--repos https://github.com/reposense/RepoSense.git --output output_path/ --ignore-standalone-config --view"`, with the following error log

This is due to not handling binary file contributions. Will be coming up with a fix.

frontend/src/static/js/v_summary.js
jamessspanggg: Done. Thanks for the feedback.

jamessspanggg:
> This is due to not handling binary file contributions. Will be coming up with a fix.

Added a test case for analysing commits binary file contribution.

src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
```diff
@@ -201,7 +231,7 @@ public void analyzeCommits_emptyCommits_success() throws ParseException {
 
         expectedCommitResults.add(new CommitResult(author, "016ab87c4afe89a98225b96c98ff28dd4774410f",
                 parseGitStrictIsoDate("2020-01-27T22:20:51+08:00"), "empty commit",
-                "", null, 0, 0));
+                "", null, 0, 0, new HashMap<>()));
```
jylee-git, line 234: If the expected fileType contribution here is empty, then is it possible to use `Collections.emptyMap()` here?

```suggestion
                "", null, 0, 0, Collections.emptyMap()));
```

src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
```diff
@@ -215,6 +245,28 @@ public void analyzeCommits_emptyCommits_success() throws ParseException {
...
 
+    @Test
+    public void analyzeCommits_commitsWithBinaryFileContribution_success() throws ParseException {
+        Author author = new Author(JAMES_AUTHOR_NAME);
+        List<CommitResult> expectedCommitResults = new ArrayList<>();
+
+        // binary file contribution will have 0 contribution and won't be added to fileTypesAndContributionMap
+        expectedCommitResults.add(new CommitResult(author, "a00c51138cbf5ab7d14f52b52abb182c8a369169",
+                parseGitStrictIsoDate("2020-04-06T16:41:10+08:00"), "Add binary file",
+                "", null, 0, 0, new HashMap<>()));
```
jylee-git, line 256: Same as the one above

```suggestion
        expectedCommitResults.add(new CommitResult(author, "a00c51138cbf5ab7d14f52b52abb182c8a369169",
                parseGitStrictIsoDate("2020-04-06T16:41:10+08:00"), "Add binary file",
                "", null, 0, 0, Collections.emptyMap()));
```

src/systemtest/resources/sinceBeginningDateRange/expected/reposense_testrepo-Delta_master/commits.json
```diff
@@ -781,13 +1279,29 @@
...
+              "java": {
+                "insertions": 1,
+                "deletions": 0
+              },
+              "adoc": {
+                "insertions": 165,
+                "deletions": 62
+              }
+            },
             "insertions": 166,
```
eugenepeh, line 1305: are the total value still in used? if not can consider removing them

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -26,11 +30,19 @@
 public class CommitInfoAnalyzer {
     public static final DateFormat GIT_STRICT_ISO_DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
 
+    public static final String TAB_SPLITTER = "\t";
+    public static final String MOVED_FILE_INDICATION = "=> ";
+    public static final String BINARY_FILE_CONTRIBUTION = "-";
+    public static final int STAT_ADDITION_INDEX = 0;
```
eugenepeh, line 36: can these be private?

src/main/java/reposense/commits/CommitInfoAnalyzer.java
```diff
@@ -93,9 +105,73 @@ public static CommitResult analyzeCommit(CommitInfo commitInfo, RepoConfiguratio
             extractTagNames(tags);
         }
 
-        int insertion = getInsertion(statLine);
-        int deletion = getDeletion(statLine);
-        return new CommitResult(author, hash, date, messageTitle, messageBody, tags, insertion, deletion);
+        if (statLine.isEmpty()) { // empty commit, no files changed
+            return new CommitResult(author, hash, date, messageTitle, messageBody, tags, 0, 0, new HashMap<>());
```
eugenepeh, line 109: consider adding a new constructor that only takes in `CommitResult(author, hash, date, messageTitle, messageBody, tags)` to handle the zero/default values.

jamessspanggg:
> The breakdown has already calculated the lines of each file type, why do we still need to modify the git command to perform this task?

Because we need to re-calculate the size of each ramp based on the checked filetypes, having the total lines of each file type is not enough. Also, each commit may contribute to multiple file types, so we need to know the lines contributed for all file types in that commit.

src/main/java/reposense/commits/CommitInfoAnalyzer.java
jamessspanggg: The reason I did not place this as private is because I am utilising these constant fields in `TestUtil`. Should I put this as private and declare the constant again in `TestUtil`?

src/systemtest/resources/sinceBeginningDateRange/expected/reposense_testrepo-Delta_master/commits.json
jamessspanggg: Yes, we can remove them since we can sum up all filetype contributions. Will do so.

src/main/java/reposense/commits/CommitInfoAnalyzer.java
jamessspanggg: Noted.

src/systemtest/resources/sinceBeginningDateRange/expected/reposense_testrepo-Delta_master/commits.json
jamessspanggg: total insertions and deletions value of all commitResults are removed. Added some logic in the front end to get those data initially instead.

jamessspanggg: @eugenepeh Note that the insertions and deletions of each commit (each commit contains multiple commitResults) are still used in the commits.json files. Should we also remove them and populate the numbers in `api.js` during initialization?

jamessspanggg: Update: can't really remove the insertions and deletions of each **commit** as they are required in the backend to calculate the variance.

src/main/java/reposense/commits/CommitInfoAnalyzer.java
jamessspanggg: Have set them to be private.

jamessspanggg:
> Please raise an issue for this, we should aim to resolve this in the near future

Will do after this is merged.

PR #1208, opened 2020-04-03
Docs: Update current members' list

anubh-v:
```
Let's update the list of current members to include
the new contributors.
```


anubh-v:
@reposense/stage0-reviewers @yash-chowdhary @ang-zeyu @niqiukun please take a look if the list is ok

docs/Team.md
```diff
@@ -61,6 +61,28 @@ Contributor [2018 Dec - 2019 May]<br/>
 ![](https://avatars3.githubusercontent.com/u/35756209?s=150&v=4)<br/>
 **Role**: Contributor [2019 May - ]<br/>
 
+### [Liu Yiwen](https://github.com/0blivious)
```
jamessspanggg, line 64: Might be better to sort based on start date?

jamessspanggg:
I think can move the contributor jin minjia to a past member. She was working on the project from May 2019 to Aug 2019.

anubh-v:
> I think can move the contributor jin minjia to a past member. She was working on the project from May 2019 to Aug 2019.

Thanks for the response @jamessspanggg 

I will make this change, as well as sort by start date.

docs/Team.md
anubh-v: Done

eugenepeh: Yes, i think sorting by start date is better

eugenepeh:
jun an and shupeng have also withdrawn from RepoSense, can move them to the past members.
The end date can be based on their last month of activity.

fzdy1914:
@eugenepeh We should also remove some of the inactive developers from stage1, stage2 reviewers so they do not get spammed by our activity.

eugenepeh:
>@eugenepeh We should also remove some of the inactive developers from stage1, stage2 reviewers so they do not get spammed by our activity.

Have removed the inactive developers. They should be able to leave the team too on their own if they felt that theyre spammed i think

anubh-v:
> jun an and shupeng have also withdrawn from RepoSense, can move them to the past members.

Done

PR #1210, opened 2020-04-03
[#1163] GitLsTreeTest: Improve unix_cloneInvalidWindowsFilePaths_success

### LINKED ISSUE #1163: GitLsTreeTest: improve testing in `unix_cloneInvalidWindowsFilePaths_success` test case
I am referring to this test: 
https://github.com/reposense/RepoSense/blob/bf77c18fd41e53676458e80fd07922a215c7472c/src/test/java/reposense/git/GitLsTreeTest.java#L73-L79

Shouldn't the test call `validateFilePaths(config)` after setting the branch?

For comparison, here is the test for Windows OS
https://github.com/reposense/RepoSense/blob/bf77c18fd41e53676458e80fd07922a215c7472c/src/test/java/reposense/git/GitLsTreeTest.java#L63-L71

anubh-v:
Fixes #1163 

```
When RepoSense is run on Windows OS, it uses the 'validateFilePaths' 
method of the GitLsTree class to verify  if all the file paths in 
the repo are valid file paths on Windows OS.

Within unix_cloneInvalidWindowsFilePaths_success test case, it 
checks whether 'validateFilePaths' method runs correctly on 
non Windows OS. However, this test does not actually call
the `validateFilePaths` method.

Let's call 'validateFilePaths' in the test case, so we can 
confirm that the method works correctly on non Windows OS.

```

### DIFF
--- a/src/test/java/reposense/git/GitLsTreeTest.java
+++ b/src/test/java/reposense/git/GitLsTreeTest.java
@@ -70,11 +70,12 @@ public class GitLsTreeTest extends GitTestTemplate {
     }
 
     @Test
-    public void unix_cloneInvalidWindowsFilePaths_success() {
+    public void unix_cloneInvalidWindowsFilePaths_success() throws Exception {
         // Runs test only on non Windows (Unix) operating systems
         Assume.assumeTrue(!SystemUtil.isWindows());
 
         config.setBranch("391-invalid-filepaths");
+        validateFilePaths(config);
     }
 
     /**


src/test/java/reposense/git/GitLsTreeTest.java
```diff
@@ -71,11 +71,12 @@ public void windows_cloneInvalidWindowsFilePaths_throwsInvalidFilePathException(
     }
 
     @Test
-    public void unix_cloneInvalidWindowsFilePaths_success() {
+    public void unix_cloneInvalidWindowsFilePaths_success() throws Exception {
         // Runs test only on non Windows (Unix) operating systems
         Assume.assumeTrue(!SystemUtil.isWindows());
 
         config.setBranch("391-invalid-filepaths");
+        validateFilePaths(config);
```
jamessspanggg, line 79: hmmm.. but the validateFilePaths method checks for filepaths compatible to windows tho. This TC is for unix path.
Screen Shot 2020-04-04 at 1 49 08 AM

anubh-v: Sorry, could you clarify what you mean by the test case being for unix paths?

From what I understand - this test case is checking whether a repo that contains file paths that are invalid on Windows, can pass the validation step in `GitLsTree.validateFilePaths`

jamessspanggg: Ah I see. 

I was a lil confused with the test case above, it says that it runs on non windows (unix) OS-es, while the file paths are invalid **Windows** file paths. So I thought it should have been invalid Unix file paths instead.

anubh-v: Yes, perhaps the branch that contains the test files should be named `invalid-windows-filepaths` instead

anubh-v:
@fzdy1914  can I get a review for this?

anubh-v:
I may have accidentally merged the master branch when it had several un-squashed commits. 
I need to rebase this PR

PR #1214, opened 2020-04-04
[#1213] Pass sorting props to v_summary_charts

### LINKED ISSUE #1213: Regressions from #1059
PR #1059 separates extracts `v_summary_charts` out of `v_summary`, while passing data through props from `v_summary` to `v_summary_charts`.

Props such as `sortingOption`, `sortingWithinOption`, `isSortingDsc`, `isSortingWithinDsc` were not passed to `v_summary_charts`, even though they were used in `openTabZoom`.

Hence, in some cases, the zoom tab is not restorable after reloading the page.

jamessspanggg:
Fixes #1213 

```
With the recent merged PR #1059, the data related to sorting were not
passed from v_summary to v_summary_charts, even though v_summary_charts
uses those data within openTabZoom method.

This results in zoom tab to not be restorable at some situations, as it 
is unable to utilise the sorting options.

Let's pass all data related to sorting from v_summary to 
v_summary_charts as props.
```

### DIFF
--- a/frontend/src/static/js/v_summary_charts.js
+++ b/frontend/src/static/js/v_summary_charts.js
@@ -92,7 +92,7 @@ window.viewClick = function viewClick(evt) {
 window.vSummaryCharts = {
   props: ['checkedFileTypes', 'filtered', 'fileTypeColors', 'avgContributionSize', 'filterBreakdown',
       'filterGroupSelection', 'filterTimeFrame', 'filterSinceDate', 'filterUntilDate', 'isMergeGroup',
-      'minDate', 'maxDate'],
+      'minDate', 'maxDate', 'sortingOption', 'sortingWithinOption', 'isSortingDsc', 'isSortingWithinDsc'],
   template: window.$('v_summary_charts').innerHTML,
   computed: {
     avgCommitSize() {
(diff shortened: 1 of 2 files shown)


jamessspanggg:
@0blivious need ur review on this, since you authored #1059.

0blivious:
Can give me a case when zoom is not restored correctly (except the case there is error reported for missing parameters)? Actually I removed all of these 4 parameters in my pr #1101 and it seems the restoration works fine. Can you check the implementation there also? Thx

jamessspanggg:
> Can give me a case when zoom is not restored correctly (except the case there is error reported for missing parameters)? Actually I removed all of these 4 parameters in my pr #1101 and it seems the restoration works fine. Can you check the implementation there also? Thx

Click on merge group -> view all commits of the first group -> change the sort groups by option -> reload

[Link to reproduce](https://reposense.netlify.com/)

jamessspanggg:
> Actually I removed all of these 4 parameters in my pr #1101 and it seems the restoration works fine. Can you check the implementation there also? Thx

I looked through the implementation and indeed the sorting options and the call for `sortFiltered` is not needed within the `restoreZoomFiltered` method. This is because the `matchZoomUser` within `restoreZoomFiltered` will already filter out other groups/repos.

Good catch. This PR is no longer necessary as passing more data will complicate the structure.

PR #1218, opened 2020-04-10
[#1213] Regressions from #1059

### LINKED ISSUE #1213: Regressions from #1059

0blivious:
Fixes #1213

```
With the recent merged PR #1059, the data related to sorting were not
passed from v_summary to v_summary_charts, even though
v_summary_charts uses those data within openTabZoom method.

This results in zoom tab to not be restorable at some situations, as it 
is unable to utilise the sorting options. However, the sorting options
and the call for sortFiltered is actually not needed
within the restoreZoomFiltered method. This is because the
matchZoomUser within restoreZoomFiltered will already filter out
other groups/repos. Thus, there is no need to pass those props into
v_summary_charts.

Let's remove the unnecessary parameters and methods.
```


jamessspanggg:
@fzdy1914 @eugenepeh may i get your review on this? This PR will affect the usage for current users

jamessspanggg:
@Tejas2805 @anubh-v @ang-zeyu @ccyccyccy may I ask your reviews on this to ensure that the bug no longer exists and no other bugs are introduced?

frontend/src/static/js/v_summary.js
```diff
@@ -803,10 +803,9 @@ window.vSummary = {
 
     restoreZoomFiltered(info) {
       const {
-        zSince, zUntil, zFilterGroup, zTimeFrame, zIsMerge, zSorting, zSortingWithin,
-        zIsSortingDsc, zIsSortingWithinDsc,
+        zSince, zUntil, zTimeFrame, zIsMerge,
       } = info;
-      let filtered = [];
+      const filtered = [];
```
Tejas2805, line 808: Can you just leave a comment here as to why the shift from `let` to `const`?

frontend/src/static/js/v_summary.js
```diff
@@ -828,15 +827,6 @@ window.vSummary = {
         }
       });
 
-      const filterControl = {
```
Tejas2805, line 831: This also exists around line number 300. Do you know why did we have duplicate code?

frontend/src/static/js/v_summary.js
0blivious: it will not be reassigned so by definition const would be better classifier. Actually this is the result of js auto correction

frontend/src/static/js/v_summary.js
0blivious: The one in 300 is needed as summary_charts will receive the filtered repo and display them without sorting. The line here is for zoom, which will perform sorting in its own js file.

PR #1223, opened 2020-04-11
[#1221] Allow merging of individual groups

### LINKED ISSUE #1221: Improve the flexibility of merge group feature
Currently the merge group feature only supports merging/expanding all groups. However, it may be useful if the user can merge all groups first, find a particular group, and expand that group to see the detailed distribution within the group. This is especially the case when there are many people and repos involved in the dashboard (e.g. [this one](https://nus-cs3281.github.io/2020-dashboard/)).

One solution is:

- Add a button after each group's title to expand/merge a particular group 
- Keep the existing merge group checkbox with the caption changed to "merge all groups"
- Change the `mergegroup` parameter in the page's url from `false`/`true` to  `all`/`none`/`0,1,2` (indices of merged groups joined together)

Shall we discuss whether there is a need for this feature?

niqiukun:
Resolves #1221 
```
Currently the `merge group` checkbox only supports merging/expanding all 
groups. 

It is inconvenient for users who want to have an overview across all 
repos first, then take a detailed look within a particular group, because
there is a need to relocate the group on the page after expanding all 
groups. This is especially the case when there are many authors and repos
on the dashboard.

Let's add the feature to allow merging/expanding individual groups 
instead.
```


frontend/src/static/js/v_summary.js
```diff
@@ -45,7 +45,7 @@ window.vSummary = {
       isSortingWithinDsc: '',
       filterTimeFrame: 'commit',
       filterBreakdown: false,
-      isMergeGroup: false,
+      mergedGroups: 'none',
```
0blivious, line 48: It seems the mergedGroups type is switch btw string and list of repos. Maybe it is better to standardise it. e.g., use null instead of string 'none'.

frontend/src/static/js/v_summary.js
```diff
@@ -188,7 +194,7 @@ window.vSummary = {
       }
 
       addHash('timeframe', this.filterTimeFrame);
-      addHash('mergegroup', this.isMergeGroup);
+      addHash('mergegroup', this.mergedGroups);
```
0blivious, line 197: The merged groups are not restored after refreshing

frontend/src/summary_charts.pug
```diff
@@ -27,9 +32,12 @@
         a(
           title="click to view breakdown of commits",
           onclick="deactivateAllOverlays()",
-          v-on:click="openTabZoom(repo[0], filterSinceDate, filterUntilDate, repo, 0)"
+          v-on:click="openTabZoom(repo[0], filterSinceDate, filterUntilDate, true)"
```
0blivious, line 35: instead of passing the boolean value here, is it possible to determine it in js?

frontend/src/summary_charts.pug
```diff
@@ -41,9 +49,9 @@
             v-bind:style="{ 'color' : fileTypeColors[fileType] }"
           )
           span &nbsp; {{ fileType }} &nbsp;
-    .summary-chart(v-for="(user, i) in repo")
-      .summary-chart__title(v-if="!isMergeGroup")
-        .summary-chart__title--index {{ i+1 }}
+    .summary-chart(v-for="(user, j) in repo")
```
0blivious, line 52: is it really needed to change all the i to j?

0blivious:
Sth wrong with the sorting, for img below Reposense is the only repo I expand
image
Then untick other, alpha become the group expanded and Reposense get merged
image

niqiukun:
> Sth wrong with the sorting, for img below Reposense is the only repo I expand
> Then untick other, alpha become the group expanded and Reposense get merged

Didn't notice that changing checked file types will change the sorting as well. I have fixed this but for now, any changes to the sorting will reset the merged groups since the merged groups are identified by the indices but not the repos themselves.

frontend/src/static/js/v_summary.js
0blivious: Then maybe better to make it a list of int? And also can change the name to mergedGroupIndexes

0blivious: Btw is it possible to have a attribute like isMerged for each repo instead of record down the indexes?

niqiukun: > Btw is it possible to have a attribute like isMerged for each repo instead of record down the indexes?

That is feasible but doing so will require change to the current representation of repos. Currently they are just arrays without any other attributes. Changing that may require large refactoring while benefits of it may not be so much.

niqiukun: > Then maybe better to make it a list of int? And also can change the name to mergedGroupIndexes

When I first thought about the type of this variable I used `string` because it can easily bind to the `merge all groups` checkbox with `v-model`, i.e. `all` represents checked while anything else represents unchecked. Let's see whether others have any opinions on this?

frontend/src/static/js/v_summary.js
niqiukun: Fixed.

Tejas2805:
Not very happy with the Ui aspect of it. Can it be something more subtle? In front of all the icons, it kind of stands out (without any good reason)

niqiukun:
> Not very happy with the Ui aspect of it. Can it be something more subtle? In front of all the icons, it kind of stands out (without any good reason)

Does this look better now?
image
image

0blivious:
@damithc hi prof can we get your opinion on the UI?

damithc:
> @damithc hi prof can we get your opinion on the UI?

Looks OK to me. 👍

frontend/src/static/js/v_summary.js
jamessspanggg: Since changing any of the sorting options or checking any file types will reset all the merge groups, perhaps we can store the merged references as the repoName (when `group by repos`) or authorName (when `group by authors`) instead of indexes? 

In this way, the sorting options or checking any file types will not affect the merged groups.

niqiukun: > Since changing any of the sorting options or checking any file types will reset all the merge groups, perhaps we can store the merged references as the repoName (when group by repos) or authorName (when group by authors) instead of indexes?

Thanks for the suggestion. I have changed the implementation for `mergedGroups`.

niqiukun:
> However, there seems to be some weird behaviour when I try to change the sort within group option:

This issue seems to exist on master and not introduced by my changes. I'm not able to verify on master yet but it does exist on the preview of another PR: https://dashboard-1218-pr-reposense-reposense.surge.sh/

frontend/src/static/js/v_summary.js
```diff
@@ -305,38 +324,83 @@ window.vSummary = {
...
+      mergeGroupStatus.forEach((isMergeGroup, groupIndex) => {
+        if (isMergeGroup) {
+          this.mergeGroupByIndex(this.filtered, groupIndex);
+        }
+      });
     },
 
     mergeGroup(filtered) {
       filtered.forEach((group, groupIndex) => {
-        const dateToIndexMap = {};
```
jamessspanggg, line 315: following the suggestion previously, you just need to add a check here if the current group name within `group name -> isMerge` leads to false, if false then you skip the merging by `continue;`. You don't need to additionally create the `mergeGroupByIndex` method.

frontend/src/static/js/v_summary.js
```diff
@@ -212,7 +213,8 @@ window.vSummary = {
 
       if (hash.timeframe) { this.filterTimeFrame = hash.timeframe; }
       if (hash.mergegroup) {
-        this.isMergeGroup = convertBool(hash.mergegroup);
+        // make a copy to prevent custom merged groups from overwritten
+        this.customMergedGroups = hash.mergegroup;
```
jamessspanggg, line 217: the mapping of hash values to the `group name -> isMerge` map can be done here

frontend/src/static/js/v_summary_charts.js
```diff
@@ -260,6 +260,35 @@ window.vSummaryCharts = {
...
+      case 'groupByRepos':
+        return group[0].repoName;
+      case 'groupByAuthors':
+        return group[0].name;
+      default:
+        return '';
+      }
+    },
+
+    isGroupMerged(groupName) {
```
jamessspanggg, line 275: you no longer need this anymore as you can just pass the `group name -> isMerged` map as props from `v_summary` to `v_summary_charts`

frontend/src/static/js/v_summary_charts.js
```diff
@@ -260,6 +260,35 @@ window.vSummaryCharts = {
...
+      if (this.mergedGroups === 'all') {
+        return true;
+      }
+      if (this.mergedGroups === 'none') {
+        return false;
+      }
+      return this.mergedGroups.split(',').includes(groupName);
+    },
+
+    handleMergeGroup(groupName) {
```
jamessspanggg, line 285: no longer need this method and the parent's `handleMergeGroup` as modifying the `group name -> isMerge` object here will affect the parent's as well.

frontend/src/static/js/v_summary_charts.js
niqiukun: I tried but it seems that changing the value of a map without reassignment does not trigger the watcher of `v_summary` to merge to repo. I still need to reassign value for the entire map attribute.

jamessspanggg: Hmm how about using `vue watch deep`? https://vuejs.org/v2/api/#watch

frontend/src/static/js/v_summary_charts.js
jamessspanggg: Ignore this, this may not be able to achievable as vue doesn't allow child component to update the parent's data. Using `$emit` is the only way. Perhaps when we transition to `vuex` then we won't be needing this anymore.

frontend/src/static/js/v_summary_charts.js
niqiukun: > Hmm how about using `vue watch deep`?

Hey this works. Thanks so much!

frontend/src/static/js/v_summary.js
jamessspanggg: it doesn't matter whether `this.filtered` is initialised or not right? Here is just to populate the data from `hash.mergegroup` to `mergedGroupsMap`?

niqiukun: But the keys of `mergedGroupsMap` need to be the group names, which comes from `this.filtered`. `hash.mergegroup` only has the merged groups' names, not the unmerged ones.

frontend/src/static/js/v_summary.js
```diff
@@ -724,7 +801,7 @@ window.vSummary = {
       filtered = this.sortFiltered(filtered, filterControl);
 
       if (zIsMerge) {
-        this.mergeGroup(filtered);
+        this.mergeGroupByIndex(filtered, 0);
```
jamessspanggg, line 804: Is this still applicable?

frontend/src/static/js/v_summary.js
jamessspanggg: Ahh i see.

frontend/src/static/js/v_summary.js
niqiukun: Sorry I missed this out. But this means that I still need `mergeGroupByIndex` because here merging has to be done regardless of the `mergedGroupsMap` status.

niqiukun:
> Can you add a corresponding front-end test for this?

There are front-end tests for the merge group feature (although not for individual groups) in PR https://github.com/reposense/RepoSense/pull/1228. I will work on the front-end test later when that PR is merged. But I hope to make a separate PR for front-end tests so that this PR can be reviewed first.

Tejas2805:
> There are front-end tests for the merge group feature (although not for individual groups) in PR #1228. 

Saying for individual groups specifically.

> I will work on the front-end test later when that PR is merged. But I hope to make a separate PR for front-end tests so that this PR can be reviewed first.

Alright! Can.

frontend/src/static/js/v_summary.js
```diff
@@ -98,16 +98,27 @@ window.vSummary = {
     },
 
     filterGroupSelection() {
+      this.getFiltered();
```
jamessspanggg, line 101: why is this moved to the top?

frontend/src/static/js/v_summary.js
```diff
@@ -98,16 +98,27 @@ window.vSummary = {
...
+
       // merge group is not allowed when group by none
+      // also reset merged groups
       if (this.filterGroupSelection === 'groupByNone') {
-        this.isMergeGroup = false;
+        this.mergedGroupsMap = {};
```
jamessspanggg, line 106: seems like the `mergedGroupsMap` will be modified once every time `filterGroupSelection` is changed. Which means that `getFiltered` will be called twice (extra time when `mergeGroupsMap` watcher is called).

frontend/src/static/js/v_summary.js
```diff
@@ -249,6 +268,14 @@ window.vSummary = {
       window.decodeHash();
     },
 
+    restoreMergedGroups() {
+      if (this.customMergedGroups) {
+        this.customMergedGroups.split(',').forEach((x) => {
```
jamessspanggg, line 273: could u use window.HASH_FILETYPE_DELIMITER as the delimiter? can refactor the name to window.HASH_DELIMITER

frontend/src/static/js/v_summary.js
niqiukun: Because to `resetMergedGroupsMap()` the info of repo names is needed (to populate the map).

frontend/src/static/js/v_summary.js
niqiukun: I suppose that is ok? Because you need to call `getFiltered` the first time to reset the map and populate it with repo names, and the second time to merge repos.

frontend/src/static/js/v_summary.js
```diff
@@ -98,16 +98,27 @@ window.vSummary = {
...
+      this.getFiltered();
+
       // merge group is not allowed when group by none
+      // also reset merged groups
       if (this.filterGroupSelection === 'groupByNone') {
-        this.isMergeGroup = false;
+        this.mergedGroupsMap = {};
+      } else if (!this.isAllGroupsMerged()) {
+        this.resetMergedGroupsMap();
+      } else {
```
0blivious, line 109: why need else if statement here?

frontend/src/static/js/v_summary.js
```diff
@@ -98,16 +98,27 @@ window.vSummary = {
...
+        this.resetMergedGroupsMap();
+        Object.keys(this.mergedGroupsMap).forEach((x) => {
```
0blivious, line 111: can use a better name instead of `x`

frontend/src/static/js/v_summary.js
```diff
@@ -207,7 +218,14 @@ window.vSummary = {
       }
 
       addHash('timeframe', this.filterTimeFrame);
-      addHash('mergegroup', this.isMergeGroup);
+
+      let mergedGroupsMapHash = Object.keys(this.mergedGroupsMap)
+          .filter((x) => this.mergedGroupsMap[x])
```
0blivious, line 223: can use better name instead of `x`

frontend/src/static/js/v_summary.js
```diff
@@ -324,41 +362,85 @@ window.vSummary = {
...
+      filtered[groupIndex] = filtered[groupIndex].slice(0, 1);
+    },
 
-        mergedCommits.sort(window.comparator((ele) => ele.date));
-        group[0].commits = mergedCommits;
-        group[0].fileTypeContribution = mergedFileTypeContribution;
-        group[0].totalCommits = totalMergedCommits;
-        group[0].variance = mergedVariance;
-        group[0].checkedFileTypeContribution = totalMergedCheckedFileTypeCommits;
+    isNoGroupMerged() {
```
0blivious, line 400: can change this to hasMergedGroup and reverse the logic accordingly?

0blivious:
Not very sure about this but since now we have the arrow indicating collapsible file for authorship page, is it better we standardise the arrow style (pointing down and pointing rightwards), and we can place the arrow icon at the right end of the icon group

niqiukun:
> Not very sure about this but since now we have the arrow indicating collapsible file for authorship page, is it better we standardise the arrow style (pointing down and pointing rightwards)

I feel that the arrow style for authorship page stands out to inform the user that it is collapsible, but we may not want the same degree of emphasis for the summary charts. 

> and we can place the arrow icon at the right end of the icon group

The main reason that it is placed as the first icon is that it does not shift its place when user collapse/expand it.

frontend/src/static/js/v_summary.js
jamessspanggg: I see

frontend/src/static/js/v_summary.js
0blivious: still I feel we should try to avoid call gieFiltered twice though since this involves lots of merging and sorting methods. if only the name is needed maybe can just extract the logic out here?

frontend/src/summary.pug
```diff
@@ -64,10 +64,11 @@
         )
           input.summary-picker__checkbox.mui-checkbox(
             type="checkbox",
-            v-model="isMergeGroup",
-            v-bind:disabled="filterGroupSelection === 'groupByNone'"
+            v-bind:checked="isAllGroupsMerged()",
+            v-bind:disabled="filterGroupSelection === 'groupByNone'",
+            v-on:click="handleMergeGroupCheckboxClicked()"
```
0blivious, line 69: dont need the click function. can just use vue set and get in your computed property and bind the model to isAllGroupsMerged

frontend/src/static/js/v_summary.js
niqiukun: > still I feel we should try to avoid call gieFiltered twice though since this involves lots of merging and sorting methods. if only the name is needed maybe can just extract the logic out here?

I have refactored this part. Now it will only do filtering and sorting (since sorting includes grouping by repo/author) but not do merging.

Tejas2805:
Can you update the UG with the new summary chart look (having the individual merge sign)?

frontend/src/static/js/v_summary.js
```diff
@@ -323,42 +377,76 @@ window.vSummary = {
...
+      return Object.keys(this.mergedGroupsMap)
+          .map((key) => this.mergedGroupsMap[key])
+          .includes(true);
+    },
+
+    isAllGroupsMerged() {
+      if (Object.keys(this.mergedGroupsMap).length === 0) {
+        return false;
+      }
+      return !Object.keys(this.mergedGroupsMap)
```
jamessspanggg, line 431: Perhaps `Object.keys(this.mergedGroupsMap).every.((key) => this.mergedGroupsMap[key])` would be clearer?

frontend/src/static/js/v_summary.js
```diff
@@ -863,6 +951,14 @@ window.vSummary = {
       info.fileTypeColors = this.fileTypeColors;
     });
   },
+  mounted() {
+    this.resetMergedGroupsMap();
+
+    // restoring custom merged groups after watchers finish their job
+    setTimeout(() => {
```
jamessspanggg, line 958: Would setting the timeout to be 0ms work as well? Since the point is to push the `restoreMergedGroups()` to the end of the execution queue. 

Currently, when you try to reload a page with some groups merged, there seems to be a slight delay on the page before they merge.

frontend/src/static/js/v_summary.js
```diff
@@ -156,6 +167,17 @@ window.vSummary = {
 
       return totalLines / totalCount;
     },
+
+    allGroupsMerged: {
+      get() {
+        return this.isAllGroupsMerged();
```
jamessspanggg, line 173: is it necessary to separate into two (allGroupsMerged and isAllGroupsMerged)? seems like the is hard to distinguish the two of them based on their names, which may confuse developers in the future

frontend/src/static/js/v_summary.js
```diff
@@ -207,7 +229,14 @@ window.vSummary = {
       }
 
       addHash('timeframe', this.filterTimeFrame);
-      addHash('mergegroup', this.isMergeGroup);
+
+      let mergedGroupsMapHash = Object.keys(this.mergedGroupsMap)
+          .filter((key) => this.mergedGroupsMap[key])
+          .join(window.HASH_DELIMITER);
+      if (mergedGroupsMapHash.length === 0) {
+        mergedGroupsMapHash = 'none';
```
jamessspanggg, line 237: I guess this can be an empty string `''`, to follow the way on how checkedFileTypes in authorship tab is done.

frontend/src/static/js/v_summary.js
```diff
@@ -277,17 +315,33 @@ window.vSummary = {
       this.$emit('get-dates', [this.minDate, this.maxDate]);
     },
 
+    getGroupName(group) {
```
jamessspanggg, line 318: i guess this can be made global to reduce duplication?

frontend/src/static/js/v_summary.js
```diff
@@ -323,42 +377,76 @@ window.vSummary = {
...
-        group[0].commits = mergedCommits;
-        group[0].fileTypeContribution = mergedFileTypeContribution;
-        group[0].totalCommits = totalMergedCommits;
-        group[0].variance = mergedVariance;
-        group[0].checkedFileTypeContribution = totalMergedCheckedFileTypeCommits;
+    hasGroupMerged() {
+      if (Object.keys(this.mergedGroupsMap).length === 0) {
+        return true;
+      }
+      return Object.keys(this.mergedGroupsMap)
```
jamessspanggg, line 422: return `Object.keys(...).map(...).some(true)` would be cleaner?

frontend/src/static/js/v_summary.js
```diff
@@ -323,42 +377,76 @@ window.vSummary = {
...
 
-        mergedCommits.sort(window.comparator((ele) => ele.date));
-        group[0].commits = mergedCommits;
-        group[0].fileTypeContribution = mergedFileTypeContribution;
-        group[0].totalCommits = totalMergedCommits;
-        group[0].variance = mergedVariance;
-        group[0].checkedFileTypeContribution = totalMergedCheckedFileTypeCommits;
+    hasGroupMerged() {
+      if (Object.keys(this.mergedGroupsMap).length === 0) {
+        return true;
```
jamessspanggg, line 420: why would hasGroupMerged return true if the map's length is 0? shouldn't it return false instead?

frontend/src/static/js/v_summary.js
```diff
@@ -323,42 +377,76 @@ window.vSummary = {
...
+      });
+
+      mergedCommits.sort(window.comparator((ele) => ele.date));
+      filtered[groupIndex][0].commits = mergedCommits;
+      filtered[groupIndex][0].fileTypeContribution = mergedFileTypeContribution;
+      filtered[groupIndex][0].totalCommits = totalMergedCommits;
+      filtered[groupIndex][0].variance = mergedVariance;
+      filtered[groupIndex][0].checkedFileTypeContribution = totalMergedCheckedFileTypeCommits;
+
+      filtered[groupIndex] = filtered[groupIndex].slice(0, 1);
```
jamessspanggg, line 415: can place a comment here to explain why only the first subarray is taken. (aka the merged group only)

frontend/src/static/js/v_summary.js
```diff
@@ -323,42 +377,76 @@ window.vSummary = {
...
-        filtered[groupIndex].push(group[0]);
+    resetMergedGroupsMap() {
+      const mergedGroupsMap = {};
+      this.filtered.forEach((group) => {
+        mergedGroupsMap[this.getGroupName(group)] = false;
       });
+      this.mergedGroupsMap = mergedGroupsMap;
+    },
+
+    handleMergeGroup(groupName) {
```
jamessspanggg, line 444: `handleMergeGroup` sounds a bit vague when the logic is only setting the merged property to true. Perhaps `setGroupToMerge` may be a better name?

frontend/src/static/js/v_summary.js
```diff
@@ -323,42 +377,76 @@ window.vSummary = {
...
+      this.mergedGroupsMap[groupName] = true;
+    },
+
+    handleExpandGroup(groupName) {
```
jamessspanggg, line 448: `unsetGroupToMerge`? just giving a suggestion, u may think of a better name

frontend/src/static/js/v_summary_charts.js
```diff
@@ -262,6 +262,25 @@ window.vSummaryCharts = {
...
+      case 'groupByRepos':
+        return group[0].repoName;
+      case 'groupByAuthors':
+        return group[0].name;
+      default:
+        return '';
+      }
+    },
+
+    handleMergeGroup(groupName) {
```
jamessspanggg, line 277: same goes for this for the naming

frontend/src/summary.pug
```diff
@@ -23,7 +23,7 @@
       .mui-select.sort-within-group
         select(
           v-model="sortWithinGroupSelection",
-          v-bind:disabled="filterGroupSelection === 'groupByNone' || isMergeGroup"
+          v-bind:disabled="filterGroupSelection === 'groupByNone' || isAllGroupsMerged()"
```
jamessspanggg, line 26: Its better to have a single `allGroupsMerged` instead of having both `isAllGroupsMerged` and `allGroupsMerged`, might be a little confusing.

frontend/src/static/js/v_summary.js
niqiukun: Do you know if there is anyway I can get a computed variable (`allGroupsMerged`) from a method? If there is a way then I don't need the `isAllGroupsMerged` method anymore.

jamessspanggg: `this.allGroupsMerged`?

frontend/src/static/js/v_summary.js
niqiukun: The reason I use `handleXXX` is to signify that the function is specifically used for UI event handler. It just happens to be that the way which it handles the event is only setting the property to true.

niqiukun: Perhaps another way to make it more specific is `handleMergeGroupIconClicked` but that's a bit too long?

frontend/src/static/js/v_summary.js
niqiukun: Thanks for the advice. But I changed the timeout to 0 but the delay still seems to be there.

frontend/src/static/js/v_summary.js
niqiukun: Oh okay it works. I thought I tried that before but somehow it didn't work. Maybe I mistakenly added brackets after.

niqiukun:
Hi, may I know what is the status for this PR? Thanks!

frontend/src/static/js/v_summary.js
```diff
@@ -98,16 +98,27 @@ window.vSummary = {
...
+      // also reset merged groups
       if (this.filterGroupSelection === 'groupByNone') {
-        this.isMergeGroup = false;
+        this.mergedGroupsMap = {};
+      } else if (!this.allGroupsMerged) {
+        this.resetMergedGroupsMap();
+      } else {
+        this.resetMergedGroupsMap();
+        Object.keys(this.mergedGroupsMap).forEach((x) => {
+          this.mergedGroupsMap[x] = true;
```
fzdy1914, line 112: Why we have to use a map to store all the merged Groups? Why a list is not enough?

frontend/src/summary_charts.pug
```diff
@@ -23,18 +35,18 @@
         title="click to view author's profile"
       )
         font-awesome-icon.icon-button(icon="user")
-      template(v-if="isMergeGroup")
+      template(v-if="mergedGroupsMap[getGroupName(repo)]")
```
fzdy1914, line 38: If you believe getting a boolean using a map is easier, it is acceptable, however, it is more standard to use a set /  list.

frontend/src/summary_charts.pug
```diff
@@ -61,35 +73,35 @@
...
-          v-bind:href="getAuthorProfileLink(repo[i].name)", target="_blank",
+          v-bind:href="getAuthorProfileLink(repo[j].name)", target="_blank",
           title="click to view author's profile"
         )
           font-awesome-icon.icon-button(icon="user")
         a(
           title="click to view author's code",
           onclick="deactivateAllOverlays()",
-          v-on:click="openTabAuthorship(user, repo, i)"
+          v-on:click="openTabAuthorship(user, repo, j, mergedGroupsMap[getGroupName(repo)])"
```
fzdy1914, line 89: Moreover, `mergedGroupsMap[getGroupName(repo)])` is used so many time, is it able to set a variable inside the `repo` class to represent for it is a merged Group or not?

The logic will be clearer as it only needs to check the attribute of the repo object. 

@jamessspanggg Your idea?

frontend/src/static/js/v_summary.js
niqiukun: I suppose using a map makes the check and modification easier.

frontend/src/summary_charts.pug
niqiukun: I believe currently we do not have a `repo` class. Right now a `repo` is a list of groups. Changing it to a class will require some refactoring. I am not sure if we should change so much just for this feature. But I agree that it can be improved.

jamessspanggg: > Moreover, `mergedGroupsMap[getGroupName(repo)])` is used so many time, is it able to set a variable inside the `repo` class to represent for it is a merged Group or not?
> 
> The logic will be clearer as it only needs to check the attribute of the repo object.
> 
> @jamessspanggg Your idea?

yes I believe it will be much clearer that way. But the way our current `repo` (a list) is structured, it makes it harder/more complex to achieve certain features like this PR. 

I believe we need to re-think and consider changing the structure of our `repo`, it may incur a lot of changes, but it will definitely make things much easier in the long run. Maybe in a separate PR.

frontend/src/static/js/v_summary_charts.js
```diff
@@ -266,6 +266,22 @@ window.vSummaryCharts = {
...
+      return this.mergedGroups.includes(groupName);
+    },
+
+    handleMergeGroup(groupName) {
+      this.$emit('handle-merge-group', groupName);
+    },
+
+    handleExpandGroup(groupName) {
+      this.$emit('handle-expand-group', groupName);
+    },
```
0blivious, line 284: we now implemented vuex for event emitting, can try use vuex here?

niqiukun: But right now `mergedGroups` is a property of `vSummary`. Do you mean making it as a global variable in `Vuex.Store`?

0blivious: You can make the `groupName` to global variable and watch it in vSummary. I am not very sure about this though as you are only emitting this to the direct parent. @fzdy1914 do you feel we should handle all event emit in Vuex

niqiukun:
Any idea how to disable `detect-non-literal-fs-filename` on Codacy? I'm not using `fs` but the rule confuses anything else that ends with `.watch` with `fs.watch`.

frontend/src/static/js/v_summary.js
```diff
@@ -285,17 +330,26 @@ window.vSummary = {
       this.$emit('get-dates', [this.minDate, this.maxDate]);
     },
 
+    getGroupName(group) {
```
0blivious, line 333: if you are only return window.getGroupName why bother writiing this function

frontend/src/static/js/v_summary.js
```diff
@@ -285,17 +330,26 @@ window.vSummary = {
...
+    },
+
+    getFilteredRepos() {
       // array of array, sorted by repo
       const full = [];
 
       // create deep clone of this.repos to not modify the original content of this.repos
       // when merging groups
-      const groups = this.isMergeGroup ? JSON.parse(JSON.stringify(this.repos)) : this.repos;
+      const groups = this.hasGroupMerged() ? JSON.parse(JSON.stringify(this.repos)) : this.repos;
```
0blivious, line 352: I feel the name should change to hasMergedGoups

frontend/src/static/js/v_summary_charts.js
```diff
@@ -266,6 +269,25 @@ window.vSummaryCharts = {
       }
       return (Math.round((index + 1) * 1000 / this.filtered.length) / 10).toFixed(1);
     },
+
+    getGroupName(group) {
```
0blivious, line 273: same here, dont need this function

frontend/src/static/js/v_summary.js
```diff
@@ -285,17 +330,26 @@ window.vSummary = {
...
+      return window.getGroupName(group, this.filterGroupSelection);
+    },
+
     getFiltered() {
       this.setSummaryHash();
       this.getDates();
       window.deactivateAllOverlays();
 
+      this.getFilteredRepos();
+      this.getMergedRepos();
```
0blivious, line 343: better perform this merging inside getFilteredRepos since the function is not returning anything.

frontend/src/static/js/v_summary_charts.js
```diff
@@ -218,7 +221,7 @@ window.vSummaryCharts = {
         author: user.name,
         repo: user.repoName,
         name: user.displayName,
-        isMergeGroup: this.isMergeGroup,
+        isMergeGroup: isMerge,
```
0blivious, line 224: isMerged

frontend/src/static/js/v_summary.js
niqiukun: 之前有人跟我讲让我把它们分开orz 因为有时候getFiltereredRepos()没必要merge

niqiukun:
I have already changed the `mergedGroups` to a list. Could you please give another review when you are free? Thanks. @fzdy1914

niqiukun:
Hi may I know what is the latest status of this PR? Does it need more reviews?

niqiukun:
@jamessspanggg It is fixed now.

jamessspanggg:
@niqiukun just one last question for me, since currently you are restoring the merged groups within the `mounted()` hook, which basically only starts processing when the html dom is mounted (that causes the slight delay in the front end before merging). 

Is it possible to place the restoring of merged groups within `beforeMount()` hook and remove the `setTimeout`? This will provide a better UX and decrease the need to re-render the ramps again.

niqiukun:
> @niqiukun just one last question for me, since currently you are restoring the merged groups within the `mounted()` hook, which basically only starts processing when the html dom is mounted (that causes the slight delay in the front end before merging).
> 
> Is it possible to place the restoring of merged groups within `beforeMount()` hook and remove the `setTimeout`? This will provide a better UX and decrease the need to re-render the ramps again.

I think the reason is that if putting it at before mount the watch hook will gets executed after restoring, and then the restored merged groups information will get wiped out when the group by category is assigned.

frontend/src/static/js/v_summary.js
```diff
@@ -259,7 +259,7 @@ window.vSummary = {
       if (hash.breakdown) {
         this.filterBreakdown = convertBool(hash.breakdown);
       }
-      if (hash.checkedFileTypes !== null) {
+      if (hash.checkedFileTypes) {
```
jamessspanggg, line 262: this fix is not needed, it is intentional for when all file types are unchecked

jamessspanggg:
> I pushed a small fix because it seems that some errors are being printed in console.

What error is that?

niqiukun:
> What error is that?

`TypeError: Cannot read property 'split' of undefined`. You can go to master's preview and open console.

jamessspanggg:
> TypeError: Cannot read property 'split' of undefined. You can go to master's preview and open console.

I think that can be in a separate PR.

> I think the reason is that if putting it at before mount the watch hook will gets executed after restoring, and then the restored merged groups information will get wiped out when the group by category is assigned.

I see. I think we can try figure out a way to preserve the information after `group by` is changed, while removing `setTimeout` and placing the restoring of merged groups in `beforeMount` hook. (In a separate PR) Will merge this for now.

jamessspanggg:
> I am fine with the current implementation. But anyway, we can consider using an object attribute to stand if it is a merged group.

> TypeError: Cannot read property 'split' of undefined. You can go to master's preview and open console.

> I think we can try figure out a way to preserve the information after group by is changed, while removing setTimeout and placing the restoring of merged groups in beforeMount hook.

@niqiukun Good work done! Can you please create issues for the comments above?

Issue #1230, opened 2020-04-12
Show all commit messages link shown when no messages available


niqiukun:
The `show all commit messages` link still shows when there is no commit messages available for all commits. Clicking on the link gives no effect.

![image](https://user-images.githubusercontent.com/5328196/79066300-6ca1f900-7ce9-11ea-9e4f-2a352e53545b.png)

PR #1231, opened 2020-04-12
[#1230] Hide show all commit messages link when no messages available

niqiukun:
Fixes https://github.com/reposense/RepoSense/issues/1230

```
Currently the `show all commit messages` link on the zoom view displays
even when there is no messages available to show. 

This may cause confusion to the user because clicking on the link gives
no effect.

Let's hide the link when there are no available messages to show.
```

### DIFF
--- a/frontend/src/static/css/panels.scss
+++ b/frontend/src/static/css/panels.scss
@@ -90,6 +90,16 @@
 .toolbar {
   cursor: pointer;
   float: right;
+
+  &--multiline {
+    float: right;
+
+    > a {
+      display: block;
+      margin: 1px 0;
+      text-align: end;
+    }
+  }
 }
 
 .sorting {
--- a/frontend/src/static/css/v_zoom.scss
+++ b/frontend/src/static/css/v_zoom.scss
@@ -1,5 +1,13 @@
 @import 'colors';
 
+.toolbar {
+  > a {
+    display: block;
+    margin: 1px 0;
+    text-align: end;
+  }
+}
+
 .zoom {
   &__title {
     &--granularity {
--- a/frontend/src/static/css/v_zoom.scss
+++ b/frontend/src/static/css/v_zoom.scss
@@ -1,13 +1,5 @@
 @import 'colors';
 
-.toolbar {
-  > a {
-    display: block;
-    margin: 1px 0;
-    text-align: end;
-  }
-}
-
 .zoom {
   &__title {
     &--granularity {
--- a/frontend/src/tabs/zoom.pug
+++ b/frontend/src/tabs/zoom.pug
@@ -38,7 +38,6 @@
               v-for="tag in commitResult.tags",
               v-on:click="scrollToCommit(tag, `tag ${commitResult.hash}`)"
             )
-
               font-awesome-icon(icon="tags")
               span &nbsp;{{ tag }}
 
--- a/frontend/src/tabs/zoom.pug
+++ b/frontend/src/tabs/zoom.pug
@@ -1,8 +1,14 @@
 #zoom
   span.large-font Commits Panel
   .toolbar(v-if="filteredUser.commits.length")
-    a(v-if="expandedCommitMessagesCount === 0", v-on:click="toggleAllCommitMessagesBody(true)") show all commit messages
-    a(v-else, v-on:click="toggleAllCommitMessagesBody(false)") hide all commit messages
+    a(
+      v-if="totalCommitMessageBodyCount !== 0 && expandedCommitMessagesCount === 0",
+      v-on:click="toggleAllCommitMessagesBody(true)"
+    ) show all commit messages
+    a(
+      v-if="totalCommitMessageBodyCount !== 0 && expandedCommitMessagesCount !== 0",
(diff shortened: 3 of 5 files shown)


Tejas2805:
Is this PR ready for review?

niqiukun:
> Is this PR ready for review?

Yes

niqiukun:
@Tejas2805 I cannot add tags or request reviews because I wasn't assigned to RepoSense team initially. Could you help me tag this PR? Thanks.

frontend/src/tabs/zoom.pug
```diff
@@ -1,8 +1,14 @@
 #zoom
   span.large-font Commits Panel
   .toolbar(v-if="filteredUser.commits.length")
-    a(v-if="expandedCommitMessagesCount === 0", v-on:click="toggleAllCommitMessagesBody(true)") show all commit messages
-    a(v-else, v-on:click="toggleAllCommitMessagesBody(false)") hide all commit messages
+    a(
+      v-if="totalCommitMessageBodyCount !== 0 && expandedCommitMessagesCount === 0",
```
Tejas2805, line 5: LGTM! But I think it might be better to use `v-if` followed by `v-if` and `v-else`.

niqiukun:
I noticed that the authorship tab have similar show/hide all links and both links show when files are partially shown. We should standardised the behaviour of show/hide all links.

Tejas2805:
> I noticed that the authorship tab have similar show/hide all links and both links show when files are partially shown. We should standardised the behaviour of show/hide all links.

When we have no code to show, the message is displayed accordingly. But, I am okay with having it removed.

frontend/src/static/css/v_zoom.scss
```diff
@@ -1,5 +1,13 @@
 @import 'colors';
 
+.toolbar {
```
Tejas2805, line 3: We have CSS for this in `panels.css` maybe can utilize that? Otherwise, it's just duplicate.

niqiukun:
> When we have no code to show, the message is displayed accordingly. But, I am okay with having it removed.

What I mean was that both show and hide all files links show on authorship tab when some files are hidden and some are not, while for commit messages on zoom tab it was either show all or hide all, without both links (hence `v-else` used). I changed this behavior to be consistent with the authorship tab.

frontend/src/static/css/v_zoom.scss
niqiukun: The CSS here forces the links to appear on two lines. I think its use is different from panels.css?

frontend/src/static/css/panels.scss
```diff
@@ -90,6 +90,16 @@
 .toolbar {
   cursor: pointer;
   float: right;
+
+  &--multiline {
```
jamessspanggg, line 94: actually, is it needed to have this extra `--multiline` class that has a duplicate `float: right` property?

frontend/src/tabs/zoom.pug
```diff
@@ -1,8 +1,14 @@
...
-  .toolbar(v-if="filteredUser.commits.length")
-    a(v-if="expandedCommitMessagesCount === 0", v-on:click="toggleAllCommitMessagesBody(true)") show all commit messages
-    a(v-else, v-on:click="toggleAllCommitMessagesBody(false)") hide all commit messages
+  .toolbar--multiline(v-if="filteredUser.commits.length && totalCommitMessageBodyCount")
+    a(
+      v-if="expandedCommitMessagesCount < totalCommitMessageBodyCount",
+      v-on:click="toggleAllCommitMessagesBody(true)"
+    ) show all commit messages
+    a(
+      v-if="expandedCommitMessagesCount !== 0",
```
jamessspanggg, line 9: would `expandedCommitMessagesCount > 0` be a better condition?

frontend/src/static/css/panels.scss
niqiukun: This class is to let the links on the toolbar appear on separate lines. May I know what is your suggestion?

jamessspanggg: as in, there's no need for the `&--multiline` right? can just 
```
.toolbar {
    ...
    > a {
        ...
    }
}
```

niqiukun: My concern would be that would apply to links on all toolbars. While some cases we might not want the links to appear on separate lines (although for now we don't have any).

niqiukun:
Hi, may I know what is the status for this PR? Thanks!

frontend/src/static/css/panels.scss
fzdy1914, line 94: Just a question, why we need to move the CSS here?

niqiukun: This `toolbar--multiline` CSS is used by both zoom and authorship page. Moved here to avoid duplication of code in separate places.

eugenepeh: hmm, why is it named multiline?

niqiukun: This allows the links on the toolbars to appear on multiple lines instead of one line. Example: 
image

PR #1243, opened 2020-04-19
[#1242] Move removal of zoom hashes to be within v_zoom component

### LINKED ISSUE #1242: Investigate moving zoom tab hash removal to zoom component

Zoom hashes are removed when the tab is deactivated and when the authorship tab is switched to.

This likely means that like the authorship view hashes, the removal can be moved into the zoom component itself so that `main.js` is cleaner.

Let's investigate whether this is really the case, and if so, move it over.

na

**Additional context**

na

jamessspanggg:
Fixes #1242 

```
The removal of zoom hashes are done in main.js, while the removal
of authorship hashes is done within the v_authorship component
itself.

We can move the responsibility to remove zoom hashes to be within 
the v_zoom component itself, in the 'beforeDestroy' hook. This 
will result in a cleaner implementation and separation of concern.

Let's move the removal of zoom hashes to be within v_zoom
component itself.
```

### DIFF
--- a/frontend/src/static/js/v_zoom.js
+++ b/frontend/src/static/js/v_zoom.js
@@ -78,8 +78,8 @@ window.vZoom = {
     setInfoHash() {
       const { addHash, encodeHash } = window;
       const {
-        zAvgCommitSize, zSince, zUntil, zFilterGroup, zTimeFrame, zIsMerge, zSorting,
-        zSortingWithin, zIsSortingDsc, zIsSortingWithinDsc, zAuthor, zRepo,
+        zAvgCommitSize, zSince, zUntil, zFilterGroup,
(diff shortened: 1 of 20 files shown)


frontend/src/static/js/v_zoom.js
```diff
@@ -110,13 +109,29 @@ window.vZoom = {
...
+      window.removeHash('zFTF');
+      window.removeHash('zMG');
+      window.encodeHash();
+    },
   },
   created() {
     this.initiate();
   },
   mounted() {
+    this.setInfoHash();
```
jamessspanggg, line 129: The reason why `setInfoHash` is moved to the `mounted` hook is that the `created` hook of the newly created component runs before the `beforeDestroy` of the previously created component.

PR #1246, opened 2020-04-20
[#1244] Add handling of filenames wrapped in double quotes

### LINKED ISSUE #1244: Possibility of Regression Bug
The Travis build which hadn't been failing upto 3 days ago, has started failing. This is when we started merging all the PRs again. We might have a regression bug. Let's look into this and find out the reason.

Can check the travis build [here](https://travis-ci.org/github/nus-cs2103-AY1920S2/ip-dashboard/builds).

This is related to individual project dashboard for CS2103T

jamessspanggg:
Fixes #1244 

```
When a filename contains special characters such as double quotes
or the backslash character \, git log will escape those characters and
surround the entire file name with double quotes.

CommitInfoAnalyzer is not handling these cases, which will lead to an
assertion error that fails the entire analysis.

Let's add the handling of such cases, removing the surrounded quotes
from the filenames, if available.
```

An example to show how Git Log displays files with special characters:
actual: `"READ\ME"` 
how git log displays:`"\"READ\\ME\""`

Screen Shot 2020-04-19 at 7 27 58 PM

Reference to the failed analysis in [CS2103's travis build](https://travis-ci.org/github/nus-cs2103-AY1920S2/ip-dashboard/builds)

### DIFF
--- a/src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
+++ b/src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
@@ -30,6 +30,7 @@ public class CommitInfoAnalyzerTest extends GitTestTemplate {
     private static final FileType FILETYPE_JAVA = new FileType("java", Collections.singletonList("**java"));
     private static final FileType FILETYPE_MD = new FileType("md", Collections.singletonList("**md"));
     private static final FileType FILETYPE_JSON = new FileType("json", Collections.singletonList("**json"));
+    private static final FileType FILETYPE_TXT = new FileType("txt", Collections.singletonList("**txt"));
 
     @Before
     public void before() throws Exception {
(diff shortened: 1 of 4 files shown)


src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
```diff
@@ -285,6 +288,38 @@ public void analyzeCommits_commitsWithBinaryFileContribution_success() throws Ex
         Assert.assertEquals(expectedCommitResults, actualCommitResults);
     }
 
+    @Test
+    public void analyzeCommits_fileNameWithSpecialChars_success() throws Exception {
```
jamessspanggg, line 292: will be running this test on non-windows as the file paths contain invalid characters in Windows.

Note that in running main program, GitLsTree will first check if system is windows, and will validate the file paths before calling the `CommitInfoAnalyzer#analyzeCommits` method. So in Windows, those file paths with invalid windows characters will not reach this point.

jamessspanggg:
Travis is under maintenance, hence the status is not showing currently.

jamessspanggg:
I have ran the analysis with the same CS2103 repo. The analysis is successful and here's the [deployed site](https://1244-bug-fix-cs2103.surge.sh)

PR #1247, opened 2020-04-20
[#1244] Add handling of filenames wrapped in double quotes

### LINKED ISSUE #1244: Possibility of Regression Bug

jamessspanggg:
Fixes #1244 

```
When a filename contains special characters such as double quotes
or the backslash character \, git log will escape those characters and
surround the entire file name with double quotes.

CommitInfoAnalyzer is not handling these cases, which will lead to an
assertion error that fails the entire analysis.

Let's add the handling of such cases, removing the surrounded quotes
from the filenames, if available.
```

An example to show how Git Log displays files with special characters:
actual: `"READ\ME"` 
how git log displays:`"\"READ\\ME\""`

Screen Shot 2020-04-19 at 7 27 58 PM

Reference to the failed analysis in [CS2103's travis build](https://travis-ci.org/github/nus-cs2103-AY1920S2/ip-dashboard/builds)

### DIFF
--- a/src/main/java/reposense/commits/CommitInfoAnalyzer.java
+++ b/src/main/java/reposense/commits/CommitInfoAnalyzer.java
@@ -1,5 +1,7 @@
 package reposense.commits;
 
+import static reposense.util.StringsUtil.removeQuote;
+
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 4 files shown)


jamessspanggg:
> Can you send the link of the passed Travis Build with this change?

I can't and don't have the access to run a Travis build on the CS2103 repo. 

However, I have ran the analysis locally with the same `repo-config` and `author-config` as CS2103's repo. The analysis is successful and here's the [deployed site](https://1244-bug-fix-cs2103.surge.sh)

jamessspanggg:
@anubh-v may i seek one more review from you?

Tejas2805:
@anubh-v @ang-zeyu Can one of you review? Need to fix this bug.

PR #1255, opened 2020-04-23
[#1139] Fix no-param- reassign warnings

### LINKED ISSUE #1139: Investigate multiple npm lint warning
When running `npm lint` under `frontend` several warnings are output like following.

`warning  Assignment to property of function parameter 'user' no-param-reassign`

Let's investigate if there are more elegant way of doing the same thing.

Tejas2805:
Fixes #1139 

```
When running npm lint under frontend a large number of
`no-param-reassign` warnings are displayed.

Let's investigate the issue and fix it accordingly so that the
warnings are removed.
```

### DIFF
--- a/frontend/src/static/js/api.js
+++ b/frontend/src/static/js/api.js
@@ -108,9 +108,10 @@ window.api = {
   setContributionOfCommitResults(dailyCommits) {
     dailyCommits.forEach((commit) => {
       commit.commitResults.forEach((result) => {
-        result.insertions = Object.values(result.fileTypesAndContributionMap)
+        const copyResult = result;
+        copyResult.insertions = Object.values(result.fileTypesAndContributionMap)
             .reduce((acc, fileType) => acc + fileType.insertions, 0);
-        result.deletions = Object.values(result.fileTypesAndContributionMap)
+        copyResult.deletions = Object.values(result.fileTypesAndContributionMap)
             .reduce((acc, fileType) => acc + fileType.deletions, 0);
       });
     });
--- a/frontend/src/static/js/api.js
+++ b/frontend/src/static/js/api.js
@@ -109,9 +109,9 @@ window.api = {
     dailyCommits.forEach((commit) => {
       commit.commitResults.forEach((result) => {
         const copyResult = result;
-        copyResult.insertions = Object.values(result.fileTypesAndContributionMap)
+        copyResult.insertions = Object.values(copyResult.fileTypesAndContributionMap)
             .reduce((acc, fileType) => acc + fileType.insertions, 0);
-        copyResult.deletions = Object.values(result.fileTypesAndContributionMap)
+        copyResult.deletions = Object.values(copyResult.fileTypesAndContributionMap)
             .reduce((acc, fileType) => acc + fileType.deletions, 0);
       });
     });
--- a/frontend/src/static/js/safari_date.js
+++ b/frontend/src/static/js/safari_date.js
@@ -20,13 +20,14 @@ function validateInputDate(event) {
 }
 
 function deleteDashInputDate(event) {
+  const copyEvent = event;
   const key = event.keyCode;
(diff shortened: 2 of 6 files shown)


jamessspanggg:
@Tejas2805 There are several regressions due to this PR, and the regressions are obvious as your changes directly modify them. You would have noticed them if you had tested out the UI with your changes.

Can you please check through your work before asking for reviews? This is because doing so will simplify the review process, it will save your time as well as the reviewers' time. There's less of a need to go back and forth to point out simple issues that you could have figured out if you had put in the effort prior.

Tejas2805:
> @Tejas2805 There are several regressions due to this PR, and the regressions are obvious as your changes directly modify them. You would have noticed them if you had tested out the UI with your changes.
> 
> Can you please check through your work before asking for reviews? This is because doing so will simplify the review process, it will save your time as well as the reviewers' time. There's less of a need to go back and forth to point out simple issues that you could have figured out if you had put in the effort prior.

@jamessspanggg Not quite sure which regressions are you pointing to. I had checked the UI and PR before asking for the review. Could you help a bit by pointing out to them? Sorry for the inconvenience. Will be more mindful next time.

jamessspanggg:
> Not quite sure which regressions are you pointing to. I had checked the UI and PR before asking for the review. Could you help a bit by pointing out to them? Sorry for the inconvenience. Will be more mindful next time.

You can check through the components that are affected by these changes:
https://github.com/reposense/RepoSense/blob/6769a9b8747e19ce1d50c54dc692157f0f3226f6/frontend/src/static/js/v_summary.js#L7-L10

https://github.com/reposense/RepoSense/blob/6769a9b8747e19ce1d50c54dc692157f0f3226f6/frontend/src/static/js/v_summary_charts.js#L37-L42

https://github.com/reposense/RepoSense/blob/6769a9b8747e19ce1d50c54dc692157f0f3226f6/frontend/src/static/js/v_zoom.js#L102-L104

ccyccyccy:
Actually I feel that this PR makes the code very much more unreadable, with the introduction of many extra variables just to avoid infringing on the rules, when we do actually want to modify the properties of the param passed in without reassigning it.

I think a better way would be to change the config to `"no-param-reassign": ["warn", { "props": false }],`, so that it will still catch the error of actually reassigning the param, but will not give a warning for the modification of properties of the param.

Tejas2805:
> I think a better way would be to change the config to `"no-param-reassign": ["warn", { "props": false }],`, so that it will still catch the error of actually reassigning the param, but will not give a warning for the modification of properties of the param.

Agreed. Had thought of the same but then based on discussion with @fzdy1914 on the issue, looks like he wanted code to change. Let's wait to see what he has to say.

Tejas2805:
@fzdy1914 Need your review on this once. To fix all the warnings I had a lot of variables to avoid infringing the rules. Do you think its okay? If yes, then will fix the conflicts and progress the PR.

fzdy1914:
Actually, it depends on the situation. The main reason JS gives such a warning is that the argument is passed by its reference. So modifying it may lead to unwanted behavior. (Modifying the original value)

However, many of our code actually rely on such behavior. For example, for the same `repo` object, it can stand for a repo, or it can stand for an author.  So sometimes, we have to modify the passed in object directly. 

The idea behind that may because of the high memory occupancy of our data. Storing all the representations of the same data may greatly influence our performance. So we trade off the run-time speed for lesser memory occupancy.

Maybe, for now, it is better to just disable the warning.

I haven't come up with a better solution for this kind of issue.

A usual way we will do is `this.user = func(this.user)`. But it is indeed complicating the things.

Maybe let this hold is a better option.

We can think more to utilize our data representation first. Like not using repo to stand for both `repo` and `author`

Tejas2805:
Sure. Closing this for now then. And opening another PR to disable the warning.

PR #1256, opened 2020-04-24
[#1241] Restore checked file types

### LINKED ISSUE #1241: Save checked file types as hash parameters
So that reloading the page will show the same checked file types

jamessspanggg:
Fixes #1241 

```
The checked file types in summary panel is not restorable after
reloading the page.

This is due to the checked file types not being saved in the hash.

Let's add the checked file types hash to allow the checked file types
to be restored after reloading the page.
```


frontend/src/static/js/v_summary.js
```diff
@@ -849,9 +862,9 @@ window.vSummary = {
     },
   },
   created() {
+    this.processFileTypes();
```
jamessspanggg, line 865: the reason why `processFileTypes` method is moved up is because `renderFilterHash` needs to access `fileTypes` array that is populated in the `processFileTypes` method.

frontend/src/static/js/v_summary.js
```diff
@@ -246,6 +255,10 @@ window.vSummary = {
       if (hash.breakdown) {
         this.filterBreakdown = convertBool(hash.breakdown);
       }
+      if (hash.checkedFileTypes) {
+        const parsedFileTypes = hash.checkedFileTypes.split(window.HASH_FILETYPE_DELIMITER);
+        this.checkedFileTypes = parsedFileTypes.filter((type) => this.fileTypes.includes(type));
+      }
```
0blivious, line 261: is there really a need to filter?

jamessspanggg: yes, there may be a case where the hash contains unnecessary noise, so this filters it out.

PR #1261, opened 2020-05-09
Fix bug when restoring zoom tab

### LINKED ISSUE #1260: Bug when restoring zoom tab

* **RepoSense Version:**
Latest master branch.

* **OS and Version:**
MacOS 10.15.4

Chrome.

1. Go to any PR preview.
1. Select group by authors, check merge group and select granularity as week.
1. Zoom into the author Eugene (eugene peh) (note that there are two eugene authors)
1. Reload the page, observe that zoom tab commits are not restored

Zoom tab commits should be restored.

Zoom tab commits are not restored.
Kapture 2020-05-09 at 18 22 06

NIL

jamessspanggg:
Fixes #1260 

```
When group by authors and groups are merged, the commits are 
not shown correctly in the zoom tab after reloading the page.

This occurs especially for an author with contribution in several 
repositories, as the repositories of the author are treated as 
separate groups (should be pushed into same group) within the
'restoreZoomFiltered' method. This leads to the merging of group 
not being able to merge all the repositories of the author, as 
they are treated as separate groups. 

The searched field is also not handled by zoom tab as well, 
leading to the reloaded zoom tab containing commits that are
not by users that match the searched field.

Let's modify the 'restoreZoomFiltered' method to treat the
repositories of an author to be in the same group, instead of 
being separate groups. Let's also add the handling of the
search field when reloading the zoom tab.
```


jylee-git:
The specific setting does not reproduce the bug anymore, but I found an issue when playing around with it, though it may not be related to this:

Try these:
1. Search: `eugene`
2. Check `merge group`
3. View all commits for testrepo alpha (notice that there should only be 3 commits).
4. Refresh the page (notice that now there are more than 3 commits).

ezgif-5-7b2ff20a7fd2

jamessspanggg:
> The specific setting does not reproduce the bug anymore, but I found an issue when playing around with it, though it may not be related to this:

Ahh that's due to another issue where we did not take into account when search is used.

jamessspanggg:
@jylee-git Fixed the issue, can help check if it still occurs?

jamessspanggg:
> Can I know which part of PR also solve the issue you mentioned?

https://github.com/reposense/RepoSense/blob/0bc6aa840263761684a7158ebe6674c02c975b1c/frontend/src/static/js/v_summary.js#L818

https://github.com/reposense/RepoSense/blob/0bc6aa840263761684a7158ebe6674c02c975b1c/frontend/src/static/js/v_summary.js#L833-L835

here.

fzdy1914:
> > Can I know which part of PR also solve the issue you mentioned?
> 
> https://github.com/reposense/RepoSense/blob/0bc6aa840263761684a7158ebe6674c02c975b1c/frontend/src/static/js/v_summary.js#L818
> 
> https://github.com/reposense/RepoSense/blob/0bc6aa840263761684a7158ebe6674c02c975b1c/frontend/src/static/js/v_summary.js#L833-L835
> 
> here.

OK, got it. But I thought this bug was fixed sometime before. Maybe I recall wrongly.

PR #1264, opened 2020-05-11
[#1263] Tests: Simplify resource loading

### LINKED ISSUE #1263: Csv Parser tests: Code used to load resources can be simplified
In all the tests for our csv parsers (`AuthorConfigParserTest`, `GroupConfigParserTest`, `RepoConfigParserTest`), we make *repeated* use of the ClassLoader to obtain the path to each resource used in the tests. This results in verbose code.

Instead of repeatedly using the ClassLoader, we can get the path to the parent directory *once* and then use the file name of each resource to get the individual paths.
This method has been adopted in the `ArgsParserTest`.

Relevant snippet in `AuthorConfigParserTest` :
(though the issue is also present in other csv parsers' tests):
https://github.com/reposense/RepoSense/blob/454eda1b61564c068accb394084fa5ca6ee99548/src/test/java/reposense/parser/AuthorConfigParserTest.java#L20-L41

anubh-v:
Fixes #1263 

```
Tests: Simplify code used for loading resources

In many of our tests,
we make repeated use of the ClassLoader to obtain the path
to each resource used in the tests.
This has resulted in alot of repetitive code.

Let's create a helper method to reduce this repetition,
and simplify the resource loading code.
```


fzdy1914:
Do we get more places in our codebase that use this kind of class loader? Can abstract them also?

anubh-v:
> Do we get more places in our codebase that use this kind of class loader? Can abstract them also?

Ok, I will apply the `loadResource` method in all these cases.

anubh-v:
> Do we get more places in our codebase that use this kind of class loader? Can abstract them also?

@fzdy1914  I've made this change
The Travis checks have passed, but the status has not been reported back to this PR.
image

anubh-v:
Thanks @jamessspanggg and @fzdy1914 for working with me on this PR !

PR #1267, opened 2020-05-13
[#1139] Remove no-param-reassign warnings

### LINKED ISSUE #1139: Investigate multiple npm lint warning

Tejas2805:
Fixes #1139 

```
When running npm lint under frontend a large number of
`no-param-reassign` warnings are displayed.

Let's prevent the warnings from being displayed.
```


.eslintrc.json
```diff
@@ -4,7 +4,7 @@
     },
     "extends": "airbnb-base",
     "rules": {
-        "no-param-reassign": "warn",
+        "no-param-reassign": 0,
```
jamessspanggg, line 7: would removing the entire rule works as well?

Tejas2805: Nope, removing the rule doesn't work as this rule is present in `airbnb-base` so causes errors, if we remove it.

Tejas2805:
@jamessspanggg Can merge this.

Issue #1271, opened 2020-05-23
Missing radio button option to filter by filetype


jylee-git:
The option to filter by filetype is missing (see below for a snapshot).

![image](https://user-images.githubusercontent.com/35756209/82726994-3a3ce000-9d1a-11ea-91a7-65cfa8c36a3a.png)

damithc:
Oops. Good catch @jylee-git 
As this is an important feature, might not want to wait till a first timer fix it.

niqiukun:
But as a user I would be quite confused by the use of radio button here. Because selecting all would mean no filtering and selecting individual types would enable the filter, the only purpose of the radio button is to choose between filter by glob and filter by file type. But maybe both filters can be applied?

damithc:
> But as a user I would be quite confused by the use of radio button here. Because selecting all would mean no filtering and selecting individual types would enable the filter, the only purpose of the radio button is to choose between filter by glob and filter by file type. But maybe both filters can be applied?

Good point @niqiukun . There is some merit in using the filter value itself to determine if the user intends to apply that filter. That also means to 'not apply' a glob filter, the user has to clear the glob box. What if the user wants to alternate between the two for some reason e.g., apply the glob filter -> check for something -> apply the files filter -> check for something -> go back to the glob filter ...
That kind of behavior is easier when there is explicit enabling/disabling of the mode.

Do we want to let users apply both at the same time? There may be use cases for that too but that's a separate feature. If we go for that, we can use checkboxes instead of radio buttons?

niqiukun:
> That kind of behavior is easier when there is explicit enabling/disabling of the mode.

I see. 

It turns out to be a small css glitch that can be fixed easily. If there is really a need to apply both filters we can have a separate issue for that. But maybe on hold for now.

damithc:
> It turns out to be a small css glitch that can be fixed easily. If there is really a need to apply both filters we can have a separate issue for that. But maybe on hold for now.

Yes, the two things can be done as separate PRs.

PR #1273, opened 2020-05-25
[#1271] Missing radio button option to filter by filetype

niqiukun:
Fixes https://github.com/reposense/RepoSense/issues/1271

```
The radio button to filter by file type is missing because it's
invisible due to css.

Let's fix the css to make it visible.
```


0blivious:
For now one of your radio button appear in the middle of the div while another appear in the top of the row. Would prefer to standardise them.

Suggestion: use one css style to control the two radio button. Use display: flex and aligh-items:baseline for the position.

frontend/src/static/css/v_authorship.scss
```diff
@@ -19,7 +19,7 @@
 
       .radio-button--checkbox {
         float: left;
-        margin: 0 2.0rem 0 -3rem;
+        margin: 0 2.0rem 0 0;
```
0blivious, line 22: can combine `radio-button--checkbox` and `radio-button--search` to `radio-button`

niqiukun: Doing so will have to change other things like the upper margin of the search bar, etc. Would like to minimise changes if possible as it is a bug fix not refactoring.

Issue #1279, opened 2020-06-05
Provide a way to run analysis for a moving window


damithc:
Suggestion: give a CLI option that runs the analysis for a period of specific length e.g., `--period 7d` runs the analysis for the last 7 days

Rationale: some users might want to monitor repositories for a moving window rather than from a specific date

Notes: minimally, we can support specifying period using days. Ability to use weeks can be provided as a convenience feature although not strictly necessary. Not sure if we want to support months/years because they are not of fixed size (it doesn't make sense if the moving window length keeps changing over time?).

PR #1280, opened 2020-06-05
[#1279] Provide a way to run analysis for a moving window

niqiukun:
Resolves https://github.com/reposense/RepoSense/issues/1279.

```
Some users might want to monitor repositories for a moving window rather
than from a specific date.

We can give a CLI option that runs the analysis for a period of specific 
length e.g., `--period 7d` runs the analysis for the last 7 days, and
`--period 1w` runs the analysis for the last 1 week.

Period can combine with either since date or until date, but not both.
If both not supplied, it will be the period before the current date.
```


niqiukun:
UG to be updated after https://github.com/reposense/RepoSense/pull/1281 is merged.

damithc:
@reposense/stage0-reviewers this PR needs reviews

docs/ug/cli.md
```diff
@@ -88,6 +88,20 @@ This flag overrides the `Ignore standalone config` field in the CSV config file.
 
 <!-- ------------------------------------------------------------------------------------------------------ -->
 
+### `--period`, `-r`
```
0blivious, line 91: -r should be changed to -p

docs/ug/cli.md
```diff
@@ -88,6 +88,20 @@ This flag overrides the `Ignore standalone config` field in the CSV config file.
...
+
+**`--period, -p PERIOD`**: Specifies the period of analysis window.
```
0blivious, line 93: remove the -p here?

0blivious:
May need to validate the calculated until date? Now if enter large period number the until date will be some day in future.

niqiukun:
> May need to validate the calculated until date? Now if enter large period number the until date will be some day in future.

I think currently for `--until` we do not check if it is in future? Maybe we need to make the behaviour consistent.

fzdy1914:
> In fact, there may be a use case where you set up the reposense report at the beginning of the semester and set the period to be 14 weeks (one semester). The until date can be someday in future so that you don't have to always update it as time progresses. I think that makes some sense.
 
@damithc Your input for this?

damithc:
> > In fact, there may be a use case where you set up the reposense report at the beginning of the semester and set the period to be 14 weeks (one semester). The until date can be someday in future so that you don't have to always update it as time progresses. I think that makes some sense.
> 
> @damithc Your input for this?

Yes, it is good if we can cater for this use case. However, the report should be generated until _today_ or the _future end date_, whichever is earlier i.e., we should not show an empty space for the period from today to the future end date.

niqiukun:
I have set the until date to be the earlier of the until date specified or current date. Not sure why `gradlew downloadNode` keeps failing though. Does not seem to be introduced by this PR.

0blivious:
> I have set the until date to be the earlier of the until date specified or current date. Not sure why `gradlew downloadNode` keeps failing though. Does not seem to be introduced by this PR.

CI failing

niqiukun:
> CI failing

Any idea on the fix?

0blivious:
> Any idea on the fix?

@fzdy1914 @reposense/stage1-reviewers

fzdy1914:
> > Any idea on the fix?
> 
> @fzdy1914 @reposense/stage1-reviewers

Seems the downloadNode gradle extension is outdated.

fzdy1914:
> > > Any idea on the fix?
> > 
> > 
> > @fzdy1914 @reposense/stage1-reviewers
> 
> Seems the downloadNode gradle extension is outdated.

The issue happens in recent 2 days, and I am able to reproduce it with their latest plugins. Not sure if using java 11 will help.

fzdy1914:
> > > > Any idea on the fix?
> > > 
> > > 
> > > @fzdy1914 @reposense/stage1-reviewers
> > 
> > 
> > Seems the downloadNode gradle extension is outdated.
> 
> The issue happens in recent 2 days, and I am able to reproduce it with their latest plugins. Not sure if using java 11 will help.

https://issues.liferay.com/browse/LPS-116536?jql=text%20~%20%22Node%20Gradle%20Plugin%3A%22 This is the related issue. It may be a good time to bump our Java version to 11 in the same time.

fzdy1914:
@jamessspanggg need your review here

PR #1287, opened 2020-06-11
Add DG page for 'Project management'

damithc:
The main change is the addition of a new DG page for 'Project management'. Other changes are just housekeeping changes.

I suggest we do a rebase-and-merge for this PR as the commits are self-contained and independent.

Note: After the next release, let me know so that I can tweak the `site.json` in the `release` branch and deploy the production website myself. The person releasing can also deploy the production website in subsequent releases.


damithc:
Ready for review.

damithc:
@jamessspanggg @fzdy1914 As you are the current main audience for this part of the documentation, have a look and go ahead and merge if OK.

damithc:
> I am OK with that. But in such a case, our release branch cannot be as same as master anymore right? In the future, we may need to merge the master into the release branch for a release?

Yes, but only the site.json file will be different. There might be merge conflicts when master branch changes site.json but that would be rare. Also, it is better to have conflicts that we can eventually resolve rather than not realize the site.json has changed and forget to do a similar change to the site.json copy used for deploying the production site. That's why I chose to use the same file for both sites rather than keep two copies of site.json for the two sites. Let's try this for a while and see how it goes. I can help if there are merge conflicts.

docs/dg/projectManagement.md
```diff
@@ -0,0 +1,49 @@
...
+1. **Dev website** at https://reposense.org/RepoSense
+   * matches the latest `master` branch
+   * deployed automatically by Travis whenever the `master` branch is updated
+
+The production website differs from the dev website in some ways e.g.,
+
+* It has a `CNAME` file (to indicate that it is the target destination for the `reposense.org` domain name)
+* Its DG pages show a warning that it is not the latest version of the DG.
+
+These variations are mainly managed by MarkBind via th `site.config` file. That is why the `site.config` file in the `release` branch is slightly different from the one in the `master` branch.
```
jamessspanggg, line 42: ... via **the `site.json` config file**? .... That is why the **`site.json` config file** in the...

damithc:
@jamessspanggg fixed the typo. Also fixes another unrelated typo in 2d2d611 
force-pushed. Remember to use 'rebase and merge' option as this PR contains multiple unrelated changes.

PR #1288, opened 2020-06-11
Add an overview diagram to the landing page

damithc:
To review, see the surge preview.

Not sure if the landing page is better with or without a diagram. If you guys think the page looks cleaner without a diagram, we can drop this PR. I'm OK either way.

Proposed commit message:

```
Add an overview diagram to the landing page
```


damithc:
Ready for review

jamessspanggg:
I think it would be a plus to add the diagram! However, would it better if we move the diagram into the grey banner? In my opinion the transition from the text `Visualize programmer activities across git repositories...` to the diagram will be smoother.

Screen Shot 2020-06-12 at 1 51 16 PM

damithc:
> I think it would be a plus to add the diagram! However, would it better if we move the diagram into the grey banner? In my opinion the transition from the text `Visualize programmer activities across git repositories...` to the diagram will be smoother.

Good point. I'll try it out and post a screenshot later. I suspect that the diagram will not look as nice in a grey background though. But we can try.

damithc:
@jamessspanggg After some tweaks, the diagram can work the grey background as well. See updated surge preview. If this one is better than the previous, can go ahead and merge.

damithc:
> The diagrams look good to me. There's a small typo with the `reposenseOverview.pptx` file name.

Good catch. Fixed.

damithc:
https://reposense.org/ updated

PR #1291, opened 2020-06-19
Use variables for page titles and headings to reduce duplication

### LINKED ISSUE #1290: DG page title should be 'Project management'
Is it supposed to look like this? @damithc 

1) 
Screen Shot 2020-06-15 at 2 32 45 PM

The dark blue in the middle looks a bit weird to me. 

2) 
Screen Shot 2020-06-15 at 2 37 29 PM

They should probably be same for consistency and follow sentence case.

damithc:
Fixes #1290 and few other similar inconsistencies

Proposed commit message:
```
The page title and the main heading of a page is exactly the same.
Let's use a MarkBind variable to hold that value so that it can be
specified only once and used in both as title and heading.

At the same time, let's fix some inconsistencies between page titles
and headings that were partially caused by the duplication this
commit addresses.

Warning: Two pages (runSh.md, usingAuthorTags.md) do not use
{{ title }} inside the <frontmatter>. This is due to a flaw in the
current MarkBind processor that affects variable values that
have `.` and `@` in them. This flaw has been fixed in the next
upcoming MarkBind release. Those two pages need to be 
updated after that version of MarkBind has been released.
```

### DIFF
--- a/docs/about.md
+++ b/docs/about.md
@@ -1,8 +1,9 @@
+<variable name="title">About us</variable>
 <frontmatter>
-  title: "About us"
+  title: "{{ title | safe }}"
 </frontmatter>
 
-<h1 class="display-3">About us</h1>
+<h1 class="display-3"><md>{{ title }}</md></h1>
 
 RepoSense is a project based in the [National University of Singapore, School of Computing](http://www.comp.nus.edu.sg/), and is funded by a _Teaching Enhancement Grant_ from [NUS Center for Development of Teaching and Learning](http://www.cdtl.nus.edu.sg/).
 
--- a/docs/contact.md
+++ b/docs/contact.md
@@ -1,8 +1,9 @@
+<variable name="title">Contact us</variable>
 <frontmatter>
-  title: "Contact us"
+  title: "{{ title | safe }}"
 </frontmatter>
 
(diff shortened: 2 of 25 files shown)


damithc:
Ready for review. Had a few false starts because of a build failure that took several tries to figure out.

Tejas2805:
Few observations and nits:

1) Not sure if this should be **report** or **Report**. 

Screen Shot 2020-06-20 at 2 10 39 PM

2) This error still exists though I remember we fixed this.

Screen Shot 2020-06-20 at 2 11 31 PM

3) In the Appendix of User Guide, only three pages have **Appendix** in heading. Maybe we can remove from all, or add in all to make it consistent. The three are:
`CLI syntax reference`, `FAQ`, and `Config files format`.

damithc:
Thanks for the review @Tejas2805 

> 3\. In the Appendix of User Guide, only three pages have **Appendix** in heading. Maybe we can remove from all, or add in all to make it consistent. The three are:
>    `CLI syntax reference`, `FAQ`, and `Config files format`.

Fixed this one. 

Can create issues for the other two as they are unrelated to this PR?

Tejas2805:
@jamessspanggg Can take a look and merge.

PR #1292, opened 2020-06-20
Use nunjucks to generate siteNavs and the site map

damithc:
## Before:
image
image

## After:
image
image

```
<include>ing the (ug|dg)SiteNave.md in the landing page site-map avoids
duplication but prevents optimal use of siteNav as well as site-map.
For example, we cannot use `:expanded:` in the siteNav to make an item
collapsible because `:expanded:` appears as raw text in the landing page.

Let's use nunjucks to define siteNav items as an array and a macro to
render it optimally for both reuse contexts.

Pros:
* siteNav items are now collapsible
* landing page shows links as a hierarchy rather than a flat list

Cons:
* The code is more complicated than before: But, the complexity is
  within the expertise of RepoSense developers as nunjucks code is
  similar to JavaScript.
```

### DIFF
--- a/docs/scripts/macros.njk
+++ b/docs/scripts/macros.njk
@@ -1,3 +1,16 @@
+{% macro show_sitenav_items(items, is_flat=false) %}
+  {% for i in items %}
+    {% if i.level == 2 %}
+  * [{{ i.name }}]({{baseUrl}}/{{ i.link }})
+    {% elseif i.link %}
+* [**{{ i.name }}**]({{baseUrl}}/{{ i.link }})
+    {% else %}
+* **{{ i.name }}**{% if not is_flat%} :expanded:{% endif %}
+    {% endif %}
+  {% endfor %}
+{% endmacro %}
+
+
 {% macro embed(header, src, level="0") %}
 <div class="indented-level{{ level }}">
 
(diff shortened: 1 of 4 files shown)


damithc:
Ready for review

docs/scripts/macros.njk
```diff
@@ -9,6 +22,7 @@
 <p/>
 {% endmacro %}
 
+
```
Tejas2805, line 25: Not sure about this but are two blank lines required?

damithc:
> These changes can't be seen in the `Production Site`. Is it intended? Or is it due to different versions of MarkBind being used for the two sites?

There is no preview for the production site. You must have been looking at the actual production site which has not been updated yet with these changes.

docs/scripts/macros.njk
damithc: I usually use two blank lines between macros, similar to python convention of putting two blank lines between functions.

damithc:
@jamessspanggg have a look and merge if no issues. Documentation PRs can be merged faster. If any issues are discovered later, we can always fix as a separate PR.

damithc:
https://reposense.org/ updated

PR #1300, opened 2020-07-09
Update Gradle Version

Tejas2805:
Fixes #1302 

```
The downloadNode Gradle extension is outdated, as a result of which
AppVeyor integration is failing. 

Let's update the Gradle version to keep up to date with its development.

From Gradle 5.0 onwards, the tasks running with code coverage are configured to 
delete the execution data just before they start executing another task. As a result of 
this, our code coverage dropped significantly since we use the same file to store the
execution data of the different tests. 

Let's use different files to store the execution data of the different tests to fix 
this problem. 
```

PR #1303, opened 2020-07-10
Upgrade to Gradle 5.2.1

fzdy1914:
Upgrade to the latest version of Gradle to keep up to date with its
development.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ plugins {
     id 'idea'
     id 'jacoco'
     id 'java'
-    id 'com.github.johnrengelman.shadow' version '2.0.4'
+    id 'com.github.johnrengelman.shadow' version '4.0.4'
     id 'com.liferay.node' version '4.4.0'
     id 'com.github.psxpaul.execfork' version '0.1.8'
     id 'com.palantir.git-version' version '0.11.0'
@@ -47,7 +47,7 @@ sourceSets {
 }
 
 wrapper {
-    gradleVersion = '4.9'
+    gradleVersion = '5.2.1'
 }
 
 run {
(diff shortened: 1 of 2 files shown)


Tejas2805:
Just want to understand one thing. Any specific reason for not using the latest versions of Gradle and Shadow?

fzdy1914:
> Just want to understand one thing. Any specific reason for not using the latest versions of Gradle and Shadow

You can have a try. I got some previous experience to be sure that this setting is compatible with java 8. If you tried that the latest one also works, we can switch the latest stable version of them.

Tejas2805:
We can update the Shadow version to 5.2.0 but updating the Gradle Version causes errors. So maybe can just use that.

fzdy1914:
> We can update the Shadow version to 5.2.0 but updating the Gradle Version causes errors. So maybe can just use that.

OK, nice finding.

fzdy1914:
@Tejas2805 If you are free, can you help investigate why there is a huge drop in codecov coverage rate?

jylee-git:
No issues from my side when using the specified configs.

Tejas2805:
> @Tejas2805 If you are free, can you help investigate why there is a huge drop in codecov coverage rate?

Sorry, for the late reply. Saw this right now. 

I have figured out the issue. Not sure how to fix it. 

```
JaCoCo plugin now works with the build cache and parallel test execution

... the tasks running with code coverage are configured to delete the execution data just before they starts executing ...
```

Based on this update, looks like we need to use a different file, currently we use the same file, as seen in this [line](https://github.com/reposense/RepoSense/blob/6c59593168fb5df060df95d288b11db14421cb31/build.gradle#L86).

Therefore, need to just split that up.

fzdy1914:
> > @Tejas2805 If you are free, can you help investigate why there is a huge drop in codecov coverage rate?
> 
> Sorry, for the late reply. Saw this right now.
> 
> I have figured out the issue. Not sure how to fix it.
> 
> ```
> JaCoCo plugin now works with the build cache and parallel test execution
> 
> ... the tasks running with code coverage are configured to delete the execution data just before they starts executing ...
> ```
> 
> Based on this update, looks like we need to use a different file, currently we use the same file, as seen in this [line](https://github.com/reposense/RepoSense/blob/6c59593168fb5df060df95d288b11db14421cb31/build.gradle#L86).
> 
> Therefore, need to just split that up.
> 
> **UPDATE**: It's been fixed and is working in #1300

OK， close this

PR #1305, opened 2020-07-19
[#1232] Fix footer link

### LINKED ISSUE #1232: Link in footer to RepoSense UG
Once PR #1222 gets merged, we can point the link in the footer to the UG in the website.

Tejas2805:
Fixes #1232 

```
The link in the footer of the RepoSense report is outdated as we have moved
everything to the website.

Let's update the links in the footer to point to the website based on whether 
a specific version or master is being used for deployment. 

If the master branch is used for deployment, we point to the latest UG. If the 
latest release is used, we point to the UG in the production site.
```


jamessspanggg:
It still shows the user guide version even though we are bringing them to the latest version. Should we remove the commit revision?

Screen Shot 2020-07-20 at 9 47 21 PM

Tejas2805:
> It still shows the user guide version even though we are bringing them to the latest version. Should we remove the commit revision?

Not sure, what you exactly mean by this. Can elaborate?

jamessspanggg:
> Not sure, what you exactly mean by this. Can elaborate?

As in the `User Guide - 266370e`, does the `266370e` represent the master's commit hash or the commit hash that the report was run on?

Tejas2805:
> > Not sure, what you exactly mean by this. Can elaborate?
> 
> As in the `User Guide - 266370e`, does the `266370e` represent the master's commit hash or the commit hash that the report was run on?

It will either represent the version which it was built on, or it will represent the commit hash that it was built on which should technically be the master's commit hash. Since this is a PR, it represents the commit hash which we will get after merging this PR. 

For example, any live report will match the master's hash, if built on the master. 

Screen Shot 2020-07-26 at 3 24 58 AM

Tejas2805:
@jamessspanggg Any updates on this?

fzdy1914:
> > Not sure, what you exactly mean by this. Can elaborate?
> 
> As in the `User Guide - 266370e`, does the `266370e` represent the master's commit hash or the commit hash that the report was run on?

It represents the hash that the commit hash that the report was run on.

fzdy1914:
> @jamessspanggg Any updates on this?

I feel that it will make our version meaningless. It is previously used to determine the report version to help bug/feature analysis. And leading to the user-guide of that specific version is more appropriate. 

Since the link is already broken, a workaround is to only make `user guide` a hyperlink but not include the version number. It is the best way under current situation.

@jamessspanggg Your opinion?

Tejas2805:
> Since the link is already broken, a workaround is to only make `user guide` a hyperlink but not include the version number. It is the best way under current situation.

Makes sense. But we should still connect to the UG of the master/ latest version based on what is being used to generate the report.

jamessspanggg:
> Since the link is already broken, a workaround is to only make `user guide` a hyperlink but not include the version number. It is the best way under current situation.
> 
> @jamessspanggg Your opinion?

I would agree that setting just a general `user guide` without a version is better for our situation now. At first glance, i thought the user guide was specific for the build.

fzdy1914:
Then let the foot link to be ` RepoSense (User Guide) - 6cfa66a `

Tejas2805:
Is this okay? The commit number looks weird there I think. Better to remove? @fzdy1914 @jamessspanggg 

Screen Shot 2020-08-13 at 12 50 00 AM

fzdy1914:
> Is this okay? The commit number looks weird there I think. Better to remove? @fzdy1914 @jamessspanggg
> 
> Screen Shot 2020-08-13 at 12 50 00 AM

Actually I think it is OK. @damithc Your idea?

damithc:
> > Is this okay? The commit number looks weird there I think. Better to remove? @fzdy1914 @jamessspanggg
> > Screen Shot 2020-08-13 at 12 50 00 AM
> 
> Actually I think it is OK. @damithc Your idea?

OK to keep. Alternatively, we can hide it behind a link e.g., 
> Generated by [this version]() of [RepoSense]() ...

Tejas2805:
These are two options:

1)
Screen Shot 2020-08-14 at 3 45 06 PM

2)
Screen Shot 2020-08-13 at 12 50 00 AM

I personally prefer the first one. Find it it to be neater. 

@damithc @fzdy1914 @jamessspanggg Your thoughts?

fzdy1914:
> I personally prefer the first one. Find it it to be neater.
> 
> @damithc @fzdy1914 @jamessspanggg Your thoughts?

The issue that, the first one may lead to some ambiguity. The link now is not version specified. It will not work if it is a previous version's website. And we lose the meaning of `version` here. 

If you truly found that version is not neat. We can do the same trick as the report generating time and hide them in the HTML and not displaying it.

Anyway, the version is to help the developer to locate the issue when the user reports a bug by attaching his report.

damithc:
> The issue that, the first one may lead to some ambiguity. The link now is not version specified. It will not work if it is a previous version's website. And we lose the meaning of `version` here.

My intention was the [version]() link will point to a specific commit in the RepoSense repo, not a version of the UG. Is that not possible? Or will it still be misleading?

fzdy1914:
> My intention was the version link will point to a specific commit in the RepoSense repo, not a version of the UG. Is that not possible? Or will it still be misleading?

Do you mean something like `Generated by this version of RepoSense (User Guide)` ? And the `version` points to a specific commit? I think it will work.

damithc:
> Do you mean something like `Generated by this version of RepoSense (User Guide)` ? And the `version` points to a specific commit? I think it will work.

Yes. If can work, let's do that.

Tejas2805:
> Yes. If can work, let's do that.

Yup, can.

Tejas2805:
@fzdy1914 for your review.

PR #1308, opened 2020-07-30
[#1073] Modify frontend pug files to use ES6+ method of string interpolation

### LINKED ISSUE #1073: Use ES6+ way for string interpolation
In the front end pug files, strings are concatenated with variables using the `+` operator. This might be messy for lines with a lot of variables concatenated together. Let's use ES6's way for string interpolation: title="\`${fileType}: ${user.fileTypeContribution[fileType]} lines, ...\`"

https://github.com/reposense/RepoSense/blob/61f1fc4f4f854a176ab1f6672f6759c42712c9c3/frontend/src/index.pug#L267-L269

vig42:
Fixes #1073 

```
In some pug files, strings are concatenated with variables using the `+`
operator. 

This might be messy for lines with a lot of variables concatenated
together. 

Let's use ES6's way for string interpolation for better readability.
```


Tejas2805:
Hi, @vig42 thanks for contributing. Could you just follow the style for our commit message, like it's done for #1309 ?

vig42:
> 
> 
> I think there is an instance of this in `segment.pug` too. Could you just check and do accordingly, if there?

Thanks, I missed that one. Have edited it.

> 
> 
> Hi, @vig42 thanks for contributing. Could you just follow the style for our commit message, like it's done for #1309 ?

Is the commit message OK now?

vig42:
> 
> 
> Good job, one issue is that ES6 backticks are indentation sensitive and there are two instances where the output contains awkward spaces:

Good point. We could consider using something like [dedent](https://www.npmjs.com/package/dedent) to address those cases. Otherwise, it would be pretty inconvenient to use the ES6 string templates with multiline strings.

jamessspanggg:
> Good point. We could consider using something like [dedent](https://www.npmjs.com/package/dedent) to address those cases. Otherwise, it would be pretty inconvenient to use the ES6 string templates with multiline strings.
> 

I don't mind using a package to manage es6 multiline string templates as it will look way neater. @fzdy1914 what do you think? however, is there another package that is being maintained actively? seems like the one you provided isn't being maintained actively.

> Actually I had enclosed it in ``` initially, but it seems @fzdy1914 had edited the message to remove those. So just to be clear, should I keep the message as it is now, or should I add back the ```?

Yeah just surround the message body with ```.

fzdy1914:
> I don't mind using a package to manage es6 multiline string templates as it will look way neater. @fzdy1914 what do you think? however, is there another package that is being maintained actively? seems like the one you provided isn't being maintained actively.

I am OK with a new package for it.

vig42:
> I don't mind using a package to manage es6 multiline string templates as it will look way neater. @fzdy1914 what do you think? however, is there another package that is being maintained actively? seems like the one you provided isn't being maintained actively.

Having looked through some alternatives, I think [endent](https://www.npmjs.com/package/endent) seems to be a good choice. Last update was relatively recent (in April 2020), and has >300k weekly downloads on npm, so it should be fairly widely used. Do you think that would be alright?

jamessspanggg:
> > I don't mind using a package to manage es6 multiline string templates as it will look way neater. @fzdy1914 what do you think? however, is there another package that is being maintained actively? seems like the one you provided isn't being maintained actively.
> 
> Having looked through some alternatives, I think [endent](https://www.npmjs.com/package/endent) seems to be a good choice. Last update was relatively recent (in April 2020), and has >300k weekly downloads on npm, so it should be fairly widely used. Do you think that would be alright?

Looks good to me! Can proceed with this option

PR #1313, opened 2020-08-08
[#1204] AuthorConfiguration#resetAuthorInformation can be made more readable

### LINKED ISSUE #1204: Usage of AuthorConfiguration#resetAuthorInformation can be made more readable
I'm referring to the following snippet:

https://github.com/reposense/RepoSense/blob/1a34084e5379a86bd685c1f35847c1038a1c37de/src/main/java/reposense/model/RepoConfiguration.java#L385-L392

Line 390 can be confusing because it appears like we are
wiping out the author list that was set on the previous line (although the reader will realise this is not the case *if* they read the `resetAuthorInformation` method code)

We can consider renaming the method `resetAuthorInformation` to `resetFromAuthorList`.

Alternatively, we can consider breaking `resetAuthorInformation` into 2 methods: 
- `authorConfig.clear` : this will clear the author configuration's internal data structures
-  `authorConfig.buildFromAuthorList` : this will re-populate the author config's internal data structures based on the author list

Then, the above snippet can look like this:
```
public void setAuthorList(List<Author> authorList) { 
  authorConfig.clear();
  authorConfig.setAuthorList(authorList);
  authorConfig.buildFromAuthorList();
  authorList.forEach(author -> AuthorConfiguration.propagateIgnoreGlobList(author, this.getIgnoreGlobList())); 
}
```

Based on discussion with @jamessspanggg

lm5lm5:
Fixes #1204 

The name of the method can be confusing because it appears like we are
wiping out the author list that was set on the previous line.

Let's,
* Break resetAuthorInformation method into two methods to increase
code readability.

### DIFF
--- a/src/main/java/reposense/model/AuthorConfiguration.java
+++ b/src/main/java/reposense/model/AuthorConfiguration.java
@@ -193,11 +193,17 @@ public class AuthorConfiguration {
     }
 
     /**
-     * Clears author mapping information and resets it with the details of current author list.
+     * Clears author mapping information.
      */
-    public void resetAuthorInformation() {
+    public void clear() {
         authorEmailsAndAliasesMap.clear();
         authorDisplayNameMap.clear();
+    }
+
+    /**
+     * Resets author mapping information with the details of current author list.
+     */
+    public void buildFromAuthorList() {
         authorList.forEach(this::setAuthorDetails);
     }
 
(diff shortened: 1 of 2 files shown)


lm5lm5:
I am not sure how to resolve the issue that "Some checks haven’t completed yet".
image

jamessspanggg:
Hi @lm5lm5 thanks for contributing! 

> I am not sure how to resolve the issue that "Some checks haven’t completed yet".

Sometimes Travis does not return the status of the checks. Do an empty commit would perhaps solve it. i.e. `git commit --allow-empty -m "retrigger CI"`

lm5lm5:
Hi @jamessspanggg, I have made an empty commit but it did not retrigger CI. What else could I do?

jamessspanggg:
Hmm I'm not sure about this, there's no build for this PR in Travis as well. Sometimes this occurs for Travis. 

@fzdy1914 is it possible to merge without the status checks? seems like a straightforward PR

fzdy1914:
I think it is straight-forward, can just merge.

PR #1316, opened 2020-08-21
[#1276] Cache mergedGroups in new data structure

### LINKED ISSUE #1276: Remove the delay before restoring merged groups
Currently there is a delay while restoring merged groups as the user refreshes the page. This is because the use of `setTimeout` which allows the merged groups to be restored after the watch hooks finish executing, in order to prevent the merged groups information being reset when `group by` is restored from hash. We can try figure out a way to preserve the information after `group by` is changed and avoid that delay.

Boon-Jun:
A PR that aims to eventually fix #1276
Still a WIP as I have not yet finalise the design and there are some finicky issues that I do not yet know how to fix. 
Any comments on how to improve is appreciated.

### DIFF
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -226,5 +226,6 @@ window.app = new window.Vue({
   },
   created() {
     this.updateReportDir();
+    this.$store.dispatch('initialize');
   },
 });
--- a/frontend/src/static/js/store.js
+++ b/frontend/src/static/js/store.js
@@ -4,7 +4,7 @@ const store = new Vuex.Store({
     tabAuthorshipInfo: {},
     tabZoomInfo: {},
     summaryDates: {},
-    mergedGroups: [],
+    mergedGroups: {},
   },
   mutations: {
     updateTabZoomInfo(state, info) {
@@ -17,9 +17,29 @@ const store = new Vuex.Store({
       state.summaryDates = info;
     },
     updateMergedGroup(state, {group, info}) {
-      state.mergedGroups = info;
+      state.mergedGroups[group] = info;
+      state.mergedGroups = { ...state.mergedGroups };
     },
   },
+  actions: {
+    initialize(context) {
+      context.commit('updateMergedGroup',
+                     {
+                        group: 'groupByNone',
+                        info: []
+                     });
+      context.commit('updateMergedGroup',
+                     {
+                       group: 'groupByRepos',
+                       info: []
+                     });
+      context.commit('updateMergedGroup',
+                     {
+                       group: 'groupByAuthors',
+                       info: []
+                     });
+    }
+  }
 });
 
 export default store;
--- a/frontend/src/static/js/store.js
+++ b/frontend/src/static/js/store.js
@@ -16,7 +16,7 @@ const store = new Vuex.Store({
     updateSummaryDates(state, info) {
       state.summaryDates = info;
     },
-    updateMergedGroup(state, info) {
+    updateMergedGroup(state, {group, info}) {
       state.mergedGroups = info;
     },
   },
--- a/frontend/src/static/js/v_summary.js
+++ b/frontend/src/static/js/v_summary.js
@@ -139,10 +139,11 @@ window.vSummary = {
 
     allGroupsMerged: {
       get() {
-        if (this.mergedGroups.length === 0) {
+        if (this.mergedGroups[this.filterGroupSelection].length === 0) {
           return false;
         }
-        return this.mergedGroups.length === this.filtered.length;
+        return this.mergedGroups[this.filterGroupSelection].length ===
+               this.filtered.length;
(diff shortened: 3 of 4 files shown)


Tejas2805:
@Boon-Jun Why do we need a cache for `groupbyNone` since there are no merged groups in that filter criteria?

Boon-Jun:
> @Boon-Jun Why do we need a cache for `groupbyNone` since there are no merged groups in that filter criteria? Is it to just store that there are no merge groups?

Hi, So mergedGroups was represented as an empty array previously and it will never be undefined. In my design, I am retrieving an array for each possible filterGroupSelection by storing them in a dictionary. The key for each filterGroup might not exist if it is not initialized explicitly at creation. So my idea is to initialize the 'groupByNone' value with an empty array, so that I can skip the check for the key/undefined values before calling its methods.

Tejas2805:
> Hi, So mergedGroups was represented as an empty array previously and it will never be undefined. In my design, I am retrieving an array for each possible filterGroupSelection by storing them in a dictionary. The key for each filterGroup might not exist if it is not initialized explicitly at creation. So my idea is to initialize the 'groupByNone' value with an empty array, so that I can skip the check for the key/undefined values before calling its methods.

Rest looks okay to me as of now. Can proceed further. Maybe @jamessspanggg can raise any issues he finds.

jamessspanggg:
Hi @Boon-Jun could you explain the high level idea of this PR and how it aims to solve the issue within your PR description? 

One note from me: it seems like you are storing all the merged groups (`groupByNone`, `groupByRepo`, `groupByAuthors`). When the analysis involves a lot of repos and groups (such as CS2103 classes), it might not be the best choice as it consumes a lot of memory.

Also, it seems like you have lint errors, can try to run `npm run lint` to view and fix them

Tejas2805:
Another thing to take note of is: When `group by` changes the `mergeGroups` also go away. So you could maybe accordingly empty the data structures. But I agree with @jamessspanggg that for modules like CS2103 it will take up a lot of memory.

Boon-Jun:
@jamessspanggg 

https://github.com/reposense/RepoSense/blob/a53961510d07cf6f066363fcdf2244ca8a7ff88f/frontend/src/static/js/v_summary.js#L62-L71
Referring to the above code_snippet, the reason why we have the problem at #1276 is because the watchers are committing an empty array into mergedGroups.
So my idea is that by caching merged groups across all group types, we can load mergedGroups from the cache whenever `filterGroupSelection` is not `groupByNone`. So, the watchers themselves will be loading from the cache instead of clearing it, and there will no longer be a need to restore mergedGroups after all watchers executes.

However, I agree that this approach can take up more memory and processing time, as everything within the cache has to be recomputed whenever there is a change in filtered groups, so that the cache stays relevant. 
I do not have any numbers yet, but the computation of merged groups will definitely be slower. Are there any recommendations on how to benchmark this project?

@Tejas2805 
Also, with regards to emptying the data structure, my original intention is to allow the caching of merged_groups for each category so that they can be stored and restored across `group by` changes. If we are going to empty the data structure every time, this defeats the purpose of having a cache🤔. There are however alternative and simpler methods to fix the issue which does not require the concept of a cache though.  But the change is less intuitive and makes the code less understandable imo. I can push new commits into this PR for your consideration.

jamessspanggg:
@Boon-Jun  thanks for the detailed explanation.

> However, I agree that this approach can take up more memory and processing time, as everything within the cache has to be recomputed whenever there is a change in filtered groups, so that the cache stays relevant.

I'm wondering if we need to sacrifice our memory and processing time usage in order to solve this UX issue that may not be a serious issue. @fzdy1914 what do you think?

> Are there any recommendations on how to benchmark this project?

I think we do have some benchmark tests within cypress, for e.g. [this](https://github.com/reposense/RepoSense/blob/master/frontend/cypress/tests/codeView/codeView_load_benchmark.js). But specifically for the merge groups, I don't think we currently have a benchmark test for that. PRs are welcomed for that as we do need more front end tests.

Tejas2805:
> If we are going to empty the data structure every time, this defeats the purpose of having a cache🤔. 

Yup, I realized this. But since the main issue for the delay is caused due to the refreshing of the pages, and not changing the `groupBy` (currently) I think it's a good way to proceed ahead to save space; unless we plan to change the current implementation by storing data across different `groupBy` too. Maybe @fzdy1914 can comment on this.

> There are however alternative and simpler methods to fix the issue which does not require the concept of a cache though. But the change is less intuitive and makes the code less understandable imo.

Maybe you can create a new PR with the changes you have in mind; this way we can easily compare between the two PRs and decide the better way to proceed ahead.

Boon-Jun:
Alternative PR: #1318

Boon-Jun:
Closed as I have made an alternative PR: #1318 to resolve the same issue

PR #1318, opened 2020-08-28
[#1276] Delay execution of filterGroupSelection watcher

### LINKED ISSUE #1276: Remove the delay before restoring merged groups

Boon-Jun:
```
MergedGroups is cleared whenever filterGroupSelection watcher is
executed in the early stages after mount.

As such, mergedGroups has to be restored separately with a noticeable
delay on the UI.

Let's delay the execution of the watcher such that mergedGroups will
not be cleared and there is no need to restore it subsequently.
```
Fix #1276


frontend/src/static/js/v_summary.js
```diff
@@ -915,11 +913,10 @@ window.vSummary = {
...
   },
   mounted() {
-    this.$store.commit('updateMergedGroup', []);
-
-    // restoring custom merged groups after watchers finish their job
+    // Delay execution of filterGroupSelection watcher
+    // to prevent clearing of merged groups
     setTimeout(() => {
-      this.restoreMergedGroups();
+      this.watcherFlag = true;
```
jamessspanggg, line 919: Perhaps can rename this watcherFlag to be more specific `filterGroupSelectionWatcherFlag`? since only filterGroupSelection watcher uses it.

jamessspanggg (commented):
The issue seems to be fixed. Just one comment from me

jamessspanggg (approved):
LGTM

fzdy1914 (changes requested):
The contribution bar seems to disappear when `merge all groups` is clicked then refresh. 

Try this [link](https://dashboard-1318-pr-reposense-reposense.surge.sh/#search=&sort=groupTitle&sortWithin=title&since=2017-10-09&timeframe=commit&mergegroup=reposense%2FRepoSense%5Bmaster%5D~reposense%2Ftestrepo-Alpha%5Bmaster%5D~reposense%2Ftestrepo-Beta%5Badd-config-json%5D~reposense%2Ftestrepo-Beta%5Bmaster%5D~reposense%2Ftestrepo-Delta%5Bmaster%5D&groupSelect=groupByRepos&breakdown=false)

frontend/src/static/js/v_summary.js
```diff
@@ -119,6 +124,9 @@ window.vSummary = {
       let totalCount = 0;
       this.repos.forEach((repo) => {
         repo.users.forEach((user) => {
+          if (user.checkedFileTypeContribution === undefined) {
```
Boon-Jun, line 127: By issuing an extra check for the checkedFileTypeContribution property and defining it, the issue of missing contribution bars will be fixed. Even so, I'm not sure if its intentional for the checkedFileTypeContribution property to be undefined.

fzdy1914: You may try to found the root issue of why this property will be undefined here.

Boon-Jun: `avgContributionSize` is computed based off `checkedFileTypeContribution` of users in this.repos ever since the change from commit 46c0d8e3d9957aa6aa440bff1440d7a1eee1e6a3. 

As far as I know, `avgContributionSize` only affects the UI of the contribution bar. What I'm unsure of is whether if it is intentional for `avgContributionSize` to vary according to the of the availability of the `checkedFileTypeContribution` property, and does the change from the commit assumes that all `checkedFileTypeContribution` are defined?

fzdy1914: @0blivious Can you help answer the question?

Boon-Jun:
Hi all, I'll have to wrap up this PR before a module deadline this Sunday. Any comments on how to proceed are greatly appreciated!

fzdy1914 (approved):
I think it is fine to stay with the current implementation then.

frontend/src/static/js/v_summary.js
Boon-Jun: Closer inspection seems to indicate that the computation of `avgContributionSize` helps to set the limit on the length of the contribution bar on initialization, but any future changes in checked file types does not lead to a recomputation of this limit.

Below is an example of what happens as this limit is not updated when file types are checked/unchecked
image

In my opinion, we can stay with the current implementation as this property is not serving it's purpose yet. but we have to figure a new way on how we can recompute this limit whenever file types are checked/unchecked.

Issue #1268, opened 2020-05-15
Moving props to Vuex global constant


0blivious:
Currently we are passing many variables as props across the components. As now we implemented Vuex, we can store some of the variables as Vuex global variables to simplify the structure. Some of these variables include

- [ ] fileTypeColors
- [ ] fileTypeColors
- [ ] fileTypeColors

Boon-Jun:
Hi, I would like to submit some changes for this. But I think this could be large enough of refactoring that spans across multiple PRs.

damithc:
As this is a wide-spread refactoring, I think it's best done by someone who is already familiar with the codebase.

PR #1319, opened 2020-08-28
Store minDate and maxDate in Vuex global variable

vig42:
Partially addresses #1268 

```
Currently we are passing minDate and maxDate as props across the components.

As we have now implemented Vuex, let's store minDate and maxDate in a Vuex
global variable to simplify the structure.
```


vig42:
Regarding the Codacy issue, I'm a little unsure as to how to deal with it. Basically it says that I should be using the ES6 shorthand for method definitions. But the format I'm using for the watcher (line 45-50) is the same as the previous one (line 41-44):

https://github.com/reposense/RepoSense/blob/102bec2d3a091f1784da5950337c7f315b21c13e/frontend/src/static/js/main.js#L41-L50

I tried changing it to the shorthand, but got an ESLint error "Expected longform method syntax for string literal key".

Also, this seems to be the correct syntax according to the [Vue API](https://vuejs.org/v2/api/#watch).

jamessspanggg:
@eugenepeh do you know how we can alter the codacy checks?

fzdy1914 (commented):
Feels that there is no need to store a combined min-max date in Vuex, can be min date and max date separately.
@jamessspanggg Your input?

jamessspanggg:
> Feels that there is no need to store a combined min-max date in Vuex, can be min date and max date separately.
> @jamessspanggg Your input?

agreed as well

frontend/src/static/js/v_summary.js
```diff
@@ -532,6 +532,7 @@ window.vSummary = {
       });
 
       this.checkedFileTypes = this.fileTypes.slice();
+      this.$store.commit('updateFileTypeColors', this.fileTypeColors);
```
fzdy1914, line 535: Is this line related to this PR?

vig42: Oh sorry, that somehow got copied into this branch. I'll take it out.

vig42:
> 
> 
> > Feels that there is no need to store a combined min-max date in Vuex, can be min date and max date separately.
> > @jamessspanggg Your input?
> 
> agreed as well

Per your suggestion, I have separated them into 2 variables.

eugenepeh:
>@eugenepeh do you know how we can alter the codacy checks?

If we have eslint, we can probably do without codacy. Can ignore codacy errors in this case.
@jamessspanggg DM me your github email on skype so that I can add you to the codacy team.

fzdy1914 (approved):

jamessspanggg (approved):

fzdy1914:
The min and max date seems not to be passed to the authorship tab correctly.

@jamessspanggg Is min date and max date really do the role of filtering in authorship tab?

fzdy1914:
> The min and max date seems not to be passed to the authorship tab correctly.
> 
> @jamessspanggg Is min date and max date really do the role of filtering in authorship tab?

@jamessspanggg @Tejas2805 Looking forward to your reply. Do the min and max date really have some meaning in authorship? Or I missed something here?

vig42:
> 
> 
> > The min and max date seems not to be passed to the authorship tab correctly.
> > @jamessspanggg Is min date and max date really do the role of filtering in authorship tab?
> 
> @jamessspanggg @Tejas2805 Looking forward to your reply. Do the min and max date really have some meaning in authorship? Or I missed something here?

By my understanding, min and max date refer to the overall min and max date. The dates which are used to filter the summary charts are stored in "filterSinceDate" and "filterUntilDate". I don't think there is any filtering of the authorship tab by date.

jamessspanggg:
@fzdy1914 @vig42 sorry for not seeing this. Nope, the min and max date does not have any meaning in the authorship. It simply represents the since date and until date for generating the reposense report. It does not filter the contributions accordingly in the authorship tab. (as opposed to commits tab)

fzdy1914 (changes requested):
Sorry for my messy comment, pls see the newest version.

1.
image
This part should also be removed right?

2.
image
This is a legacy problem, the date should stand for the start date and the end date of the report in the authorship tab. However, if we change the date in the date picker and refresh, the min-max date shown here is also changed accordingly, which is not the case. The min-max date in authorship tab should stick to the report start date and end date instead of the filter start date and end date.

So, my suggestion is that: Investigate the usage of min-date and max-date in the whole project to see if you can 
1. Correct the legacy problem mentioned in part 2, so that the min-max date displayed in authorship tab is always correct.
2. Move the getDate function to another vue life cycle as it is designed to be called only once?
3. Instead of passing min-date and max-date to authorship tab by `info`, use the attribute directly in the store in authorship tab.

@vig42 Please help to investigate the issue I have mentioned.
@jamessspanggg Your input is also needed, correct me if I am wrong.

jamessspanggg:
> This part should also be removed right?

Yes this part should be removed since the component can directly access the variables from the store.

> So, my suggestion is that: Investigate the usage of min-date and max-date in the whole project to see if you can
> 
> 1. Correct the legacy problem mentioned in part 2, so that the min-max date displayed in authorship tab is always correct.
> 2. Move the getDate function to another vue life cycle as it is designed to be called only once?
> 3. Instead of passing min-date and max-date to authorship tab by `info`, use the attribute directly in the store in authorship tab.

Agree with the 3 points above.

PR #1320, opened 2020-08-28
Store fileTypeColors as Vuex global variable

vig42:
Partially addresses #1268 

```
Currently we are passing fileTypeColors as a prop across the components.

As we have now implemented Vuex, let's store fileTypeColors as a Vuex
global variable to simplify the structure.
```


jamessspanggg (approved):

fzdy1914 (approved):
LGTM

frontend/src/static/js/v_summary_charts.js
```diff
@@ -31,6 +31,7 @@ window.vSummaryCharts = {
     },
 
     ...Vuex.mapState(['mergedGroups']),
+    ...Vuex.mapState(['fileTypeColors']),
```
fzdy1914, line 34: Can this statement be merged with the previous one?

vig42: Yes, I have merged it.

frontend/src/static/js/v_authorship.js
```diff
@@ -121,7 +121,7 @@ window.vAuthorship = {
       }
 
       if (!this.info.fileTypeColors) {
-        this.$root.$emit('restoreFileTypeColors', this.info);
+        this.info.fileTypeColors = this.$store.state.fileTypeColors;
```
fzdy1914, line 124: image

I believe there is no need to store fileTypeColors in the info anymore. Let's set it to a computed map state value just like how we do it in `vSummaryCharts`

0blivious (approved):

fzdy1914 (approved):

jamessspanggg (commented):
Seems like the file types are not restored after reloading the page, is it related to ur PR?

filetype

Don't think so, I'm getting the same issue on master too. Specifically, if you select 2 or more file types, it does not remember your selection when you refresh. Strangely, if you only select 1 type (e.g. only selecting frontend), it remembers your selection. Issue is present on both Chrome and Firefox.

vig42:
> 
> 
> Seems like the file types are not restored after reloading the page, is it related to ur PR?
> 
> filetype


Don't think so, I'm getting the same issue on master too. Specifically, if you select 2 or more file types, it does not remember your selection when you refresh. Strangely, if you only select 1 type (e.g. only selecting frontend), it remembers your selection. Issue is present on both Chrome and Firefox.

Tejas2805:
From what I remember, we don't restore `fileTypes` currently. Based on the tests I wrote, I think this is the case.

jamessspanggg:
> Don't think so, I'm getting the same issue on master too. Specifically, if you select 2 or more file types, it does not remember your selection when you refresh. Strangely, if you only select 1 type (e.g. only selecting frontend), it remembers your selection. Issue is present on both Chrome and Firefox.

Thanks for investigating, perhaps can file this as an issue?

jamessspanggg (approved):

PR #1326, opened 2020-09-12
[#1239] Add filetype tags in commits panel

### LINKED ISSUE #1239: Commits Panel: add filetype tag to each commit
With the recent merged #1192, we have the file type info of each commit. 

With this info, we can easily add the filetype tag(s) to each commit and allowing the filter of commits based on filetypes.  Similar to the UI implemented [here](https://github.com/reposense/RepoSense/pull/1093), but in Commits Panel.

Boon-Jun:
Resolves #1239
```
With PR #1192, we have the file type info of each commit, 
but users are still unable to view the type of file changes 
made in each commit as this information is not displayed.

The addition of filetype tags in the commits panel gives 
users an easy view to the type of changes made in 
each commit.

Let's add file type tags to each commit in the 
commits panel along with an additional filtering 
functionality for easy viewing.
```

### DIFF
--- a/frontend/src/static/js/v_zoom.js
+++ b/frontend/src/static/js/v_zoom.js
@@ -72,7 +72,7 @@ window.vZoom = {
     },
 
     restoreZoomTab() {
-      // restore selected user's commits from v_summary
+      // restore selected user's commits and file type colors from v_summary
       this.$root.$emit('restoreCommits', this.info);
     },
 
--- a/frontend/src/static/js/v_zoom.js
+++ b/frontend/src/static/js/v_zoom.js
@@ -9,6 +9,7 @@ window.vZoom = {
       expandedCommitMessagesCount: this.totalCommitMessageBodyCount,
       commitsSortType: 'time',
       toReverseSortedCommits: true,
+      isCommitsFinalized: false,
       selectedFileTypes: [],
       fileTypes: [],
     };
@@ -33,7 +34,7 @@ window.vZoom = {
       filteredUser.commits = zUser.commits.filter(
           (commit) => commit[date] >= zSince && commit[date] <= zUntil,
       ).sort(this.sortingFunction);
-      this.updateFileTypes(filteredUser.commits);
+      this.isCommitsFinalized = true;
 
       return filteredUser;
     },
@@ -82,6 +83,15 @@ window.vZoom = {
     ...Vuex.mapState(['fileTypeColors']),
   },
 
+  watch: {
+    isCommitsFinalized() {
+      if (this.isCommitsFinalized) {
+        this.updateFileTypes(this.filteredUser.commits);
+        this.selectedFileTypes = this.fileTypes.slice();
+      }
+    },
+  },
+
   methods: {
     initiate() {
       if (!this.info.zUser) { // restoring zoom tab from reloaded page
@@ -125,7 +135,6 @@ window.vZoom = {
         });
       });
       this.fileTypes.sort();
-      this.selectedFileTypes = this.fileTypes.slice();
     },
 
     setInfoHash() {
--- a/frontend/src/static/js/v_zoom.js
+++ b/frontend/src/static/js/v_zoom.js
@@ -8,6 +8,8 @@ window.vZoom = {
       commitsSortType: 'time',
       toReverseSortedCommits: true,
       shouldShowFileTypes: false,
+      selectedFileTypes: [],
+      fileTypes: [],
     };
   },
 
@@ -30,8 +32,28 @@ window.vZoom = {
       filteredUser.commits = zUser.commits.filter(
           (commit) => commit[date] >= zSince && commit[date] <= zUntil,
       ).sort(this.sortingFunction);
+      this.selectedCommits = filteredUser.commits;
+      this.updateFileTypes(filteredUser.commits);
 
       return filteredUser;
+    },
+    selectedCommits() {
+      const commits = [];
+      this.filteredUser.commits.forEach((commitDay) => {
+        const filteredDay = { ...commitDay };
+        filteredDay.commitResults = [];
+        commitDay.commitResults.forEach((slice) => {
+          if (Object.keys(slice.fileTypesAndContributionMap).some(
+              (fileType) => this.selectedFileTypes.indexOf(fileType) !== -1,
+          )) {
+            filteredDay.commitResults.push(slice);
+          }
+        });
+        if (filteredDay.commitResults.length > 0) {
+          commits.push(filteredDay);
+        }
+      });
+      return commits;
     },
     totalCommitMessageBodyCount() {
       let nonEmptyCommitMessageCount = 0;
@@ -45,7 +67,20 @@ window.vZoom = {
 
       return nonEmptyCommitMessageCount;
     },
+    isSelectAllChecked: {
+      get() {
+        return this.selectedFileTypes.length === this.fileTypes.length;
+      },
+      set(value) {
+        if (value) {
+          this.selectedFileTypes = this.fileTypes.slice();
+        } else {
+          this.selectedFileTypes = [];
+        }
+      },
+    },
   },
+
   methods: {
     initiate() {
       if (!this.info.zUser) { // restoring zoom tab from reloaded page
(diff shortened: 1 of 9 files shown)


Tejas2805:
I was thinking for consistency we can display it like the `Code Panel`? Or if we consider fileTypes to not be that important for the `Commit Panel` then we proceed with this.

@damithc Your thoughts on the UI?

damithc:
@Boon-Jun can give a screenshot and point out the exact UI change?

Boon-Jun:
@damithc 
image
These are the UI changes made in this PR. The code panels displays line contributions for each file type, but they are not displayed here in the commits panel. Other than that, I have thought of additional functionalities which are also not included for now. (e.g hashing of filters for restoration, changing the displayed v-ramp according to the filters). But it will be great if I can get some opinions on them and potential use cases before continuing.

damithc:
Thanks for the screenshot @Boon-Jun 
Hashing of filters (and any other setting in the page) could be useful as we want to be able to share a report exactly as seen by the viewer by simply sending the current URL to someone else. But we should do it in a way that can be applied to all pages rather than  just the commits page.

damithc:
> @damithc Your thoughts on the UI?

I'm fine with the proposed UI given in the screenshot. I didn't find the need for such a feature myself, but may be OK to have if there is no big cost.

BTW, why doesn't the feature show up in the surge PR preview?

Tejas2805:
> BTW, why doesn't the feature show up in the surge PR preview?

It is there in the surge PR preview too. Just that currently by default its unticked, so the fileTypes don't show up when you open them.

Boon-Jun:
I think we can choose to implement the hashing of all settings(including the filetypes) within the code panels with a separate PR. Or is it more preferable to include it here?

Boon-Jun:
There also seems to be a bug with the restoration of fileTypes in the code panels.
To reproduce it, open up a code panel with any of the file types selected and refresh the page.

Tejas2805:
> I think we can choose to implement the hashing of all relevant settings(including the filetypes) within the commit panels with a separate PR. Or is it more preferable to include it here?

Better to create a new PR for this.

Tejas2805:
> There also seems to be a bug with the restoration of fileTypes in the code panels.
> To reproduce it, open up a code panel with any of the file types selected and refresh the page.

Can you check if there is an existing issue already or not? If not, do create one.

frontend/src/tabs/zoom.pug
```diff
@@ -62,8 +62,28 @@
...
+        span All&nbsp;
+      template(v-for="fileType in fileTypes")
+        label(
+          v-bind:key="fileType",
+          v-bind:style="{\
+                  'background-color': info.zFileTypeColors[fileType],\
+                  'color': this.getFontColor(info.zFileTypeColors[fileType])\
+                  }"
+        )
+          input.mui-checkbox--fileType(type="checkbox", v-bind:id="fileType", v-bind:value="fileType",
```
jamessspanggg, line 82: one question: do you need to v-bind `id` and `value`? would `id` and `value` suffice?

Boon-Jun: v-bind:value should be enough. Thanks for notifying

frontend/src/static/js/v_zoom.js
```diff
@@ -29,9 +32,29 @@ window.vZoom = {
       filteredUser.commits = zUser.commits.filter(
           (commit) => commit[date] >= zSince && commit[date] <= zUntil,
       ).sort(this.sortingFunction);
+      this.selectedCommits = filteredUser.commits;
```
fzdy1914, line 35: The logic here is quite confusing, why we need to update a computed value here?

frontend/src/static/js/v_zoom.js
```diff
@@ -29,9 +32,29 @@ window.vZoom = {
...
+      this.updateFileTypes(filteredUser.commits);
```
fzdy1914, line 36: We can make selectedFileTypes as a computed value, too

fzdy1914 (approved):

frontend/src/static/js/v_zoom.js
Boon-Jun: Thanks for the review. Can I clarify why you think selectedFileTypes can be made a computed value? In my implementation, the updateFileTypes method is only called once, and filetypes can only be determined when the commits are known. You can think of it as a form of "initialization" of required variables. selectedFileTypes is only dependent on user interaction thereafter.

frontend/src/static/js/v_zoom.js
```diff
@@ -7,6 +9,9 @@ window.vZoom = {
       expandedCommitMessagesCount: this.totalCommitMessageBodyCount,
       commitsSortType: 'time',
       toReverseSortedCommits: true,
+      shouldShowFileTypes: true,
```
0blivious, line 12: I feel we dont really need this option, can make it as the same as authorship one.

frontend/src/static/js/v_zoom.js
fzdy1914: Yes, I got what you mean. I didn't notice that this attribute will only dependent on user interaction after the initialization.

frontend/src/static/js/v_zoom.js
```diff
@@ -7,6 +9,9 @@ window.vZoom = {
       expandedCommitMessagesCount: this.totalCommitMessageBodyCount,
       commitsSortType: 'time',
       toReverseSortedCommits: true,
+      isFileTypesLoaded: false,
```
0blivious, line 12: is this used?

frontend/src/static/js/v_zoom.js
```diff
@@ -72,10 +111,52 @@ window.vZoom = {
...
+      commits.forEach((day) => {
+        day.commitResults.forEach((slice) => {
+          Object.keys(slice.fileTypesAndContributionMap).forEach((fileType) => {
+            if (this.fileTypes.indexOf(fileType) === -1) {
+              this.fileTypes.push(fileType);
+            }
+          });
+        });
+      });
+      this.fileTypes.sort();
```
0blivious, line 129: Better not use sorting as we would want the `others` type to be displayed as the last one.

frontend/src/static/js/v_zoom.js
```diff
@@ -72,10 +111,52 @@ window.vZoom = {
...
+      this.updateSelectedFileTypes();
+    },
+
+    updateSelectedFileTypes() {
+      this.retrieveSelectedFileTypesHash();
+      if (this.selectedFileTypes === undefined) {
+        this.selectedFileTypes = this.fileTypes.slice();
+      }
```
0blivious, line 137: the undefined case will only occur at initialization? then why not put it in filteredUser and then combine the rest of updateSelectedFileTypes with retrieveSelectedFileTypesHash

frontend/src/static/js/v_zoom.js
Boon-Jun: Yes, it will be.

Boon-Jun:
Screenshot 2020-10-03 at 6 37 53 PM

Following the suggestion by @0blivious, I have decided to remove the option for users to remove the display of filetypes.
Also, I have additionally introduced the hashing of filetypes with this PR.
Thanks.

frontend/src/static/js/v_zoom.js
```diff
@@ -7,6 +9,9 @@ window.vZoom = {
       expandedCommitMessagesCount: this.totalCommitMessageBodyCount,
       commitsSortType: 'time',
       toReverseSortedCommits: true,
+      isCommitsFinalized: false,
+      selectedFileTypes: undefined,
```
0blivious, line 13: would prefer initialize to be `[]`

frontend/src/static/js/v_zoom.js
```diff
@@ -29,9 +34,28 @@ window.vZoom = {
       filteredUser.commits = zUser.commits.filter(
           (commit) => commit[date] >= zSince && commit[date] <= zUntil,
       ).sort(this.sortingFunction);
+      this.isCommitsFinalized = true;
 
       return filteredUser;
     },
+    selectedCommits() {
+      const commits = [];
+      this.filteredUser.commits.forEach((commitDay) => {
```
0blivious, line 43: can change all the `commitDay`, `day` in code to `commit`? According to the granularity the commit here can actually be grouped by weeks etc

0blivious (approved):

jamessspanggg (approved):
LGTM 👍🏻

jamessspanggg:
After merging this PR, perhaps can add new tests for the file types in commits panel?

fzdy1914:
The CI is failing, can you have a check? @Boon-Jun

Boon-Jun:
> After merging this PR, perhaps can add new tests for the file types in commits panel?

I've now added the tests for commits panel file types in this PR.

frontend/src/static/js/v_zoom.js
```diff
@@ -1,3 +1,5 @@
+/* global Vuex */
```
fzdy1914, line 1: Why this comment is needed here?

fzdy1914 (approved):

frontend/src/static/js/v_zoom.js
Boon-Jun: I believe it was a technique to avoid lint errors.

fzdy1914: OK then

Issue #1205, opened 2020-04-02
Provide help page in dashboard


jamessspanggg:
The current workflow of RepoSense is that users can setup a RepoSense dashboard and share the link to his/her peers or students. 

However, the people that are shared to may not understand how to interpret the report (such as what does the ramp size and ramp position represent etc). They are also highly unlikely to go to our User Guide to find out how to interpret the report. 

Perhaps we can provide sort of a help page (in terms of a modal) that provides all information to interpret the report? 

_p.s. asking for this because there were several occasions where i shared the dashboard to my peers but they can't understand what the ramp represents._

jamessspanggg:
@damithc what do you think? since you are sharing to hundreds of students every semester, would this addition be good?

damithc:
Yes, that's a good idea @jamessspanggg

Tejas2805:
What all we plan to have in Help Page? Some specific sections of the UG?

Tejas2805:
Do we still intend to provide this after the change in our docs? If yes, then what's the intendedd plan? I can work on this accordingly.

damithc:
Let's add a `how to interpret this report` link somewhere appropriate and link it to the relevant place of the user guide.

tlylt:
> Let's add a how to interpret this report link somewhere appropriate and link it to the relevant place of the user guide.

Hi @damithc, currently the page does include a link to the `user guide > Using reports` here:
> See the [User Guide](https://reposense.org/ug/usingReports.html)  to get a better understanding of how to interpret the report.

Would this be sufficient already or is there anything else left for this issue?

Screenshot for reference:
![image](https://user-images.githubusercontent.com/41845017/166439249-e7065050-632c-4e54-83a8-e21f29c507df.png)

damithc:
> Would this be sufficient already or is there anything else left for this issue?

And there is another link to the user guide at the bottom of the left panel. Yes, I guess this is enough.

PR #1331, opened 2020-09-26
[#1205] Show a tooltip when hovering over ambiguous icons and text

dcshzj:
Part of #1205.

```
The report currently has existing tooltips that are rendered by
the browser. However, the behaviour is not consistent across
different browsers and in certain cases, not appear at all.

Let's introduce a CSS-based functionality for showing tooltips to
the user and provide a quick explanation on the use of some of the
report features.
```


frontend/src/tabs/zoom.pug
```diff
@@ -86,10 +86,11 @@
...
           v-if="slice.messageBody !== ''",
           v-on:click="updateExpandedCommitMessagesCount",
-          onclick="toggleNext(this)",
-          title="Click to show/hide the commit message body"
+          onclick="toggleNext(this)"
         )
-          font-awesome-icon.commit-message--button(icon="ellipsis-h")
+          .tooltip
+            font-awesome-icon.commit-message--button(icon="ellipsis-h")
+            span.tooltip-text Click to show/hide the commit message body
```
Tejas2805, line 93: Is it possible to show the message based on whether the commit message body is being shown or hidden? I feel it might be better than a general message.

0blivious: Also, possible to add the same message for the authorship panel?

dcshzj: Sure! I can make the message responsive depending on the state.

Tejas2805:
@damithc What do you think about adding tooltips to the UI. I think it might be more prominent and better than the way, we used to display it earlier.

damithc:
> @damithc What do you think about adding tooltips to the UI. I think it might be more prominent and better than the way, we used to display it earlier.

Can show me (explain) _before_ and _after_?

dcshzj:
This is what we have currently:

image

The tooltip is displayed depending on the browser, and sometimes it might not show up at all.

This is the proposed change to the tooltip:

image

damithc:
> Currently, the tooltip is displayed depending on the browser, and sometimes it might not show up at all. However, the proposed change allows the tooltip to appear more prominently (and also definitely). The user will then be able to get immediate feedback on what the button does, especially for instances where it can be ambiguous.

Sounds good. Go ahead.

dcshzj:
@Tejas2805 I tried to make the tooltip for showing and hiding the commit message body responsive, but apparently it is not trivial as we have to refactor the code such that we can take advantage of states. Is it possible to work on that as a separate pull request instead?

Tejas2805:
> @Tejas2805 I tried to make the tooltip for showing and hiding the commit message body responsive, but apparently it is not trivial as we have to refactor the code such that we can take advantage of states (currently, it is using the `window` functions). Is it possible to work on that as a separate pull request instead?

@fzdy1914 Will it be useful to refactor the code to take advantages of the states, otherwise, this is a very small case for refactoring the code

fzdy1914:
> > @Tejas2805 I tried to make the tooltip for showing and hiding the commit message body responsive, but apparently it is not trivial as we have to refactor the code such that we can take advantage of states (currently, it is using the `window` functions). Is it possible to work on that as a separate pull request instead?
> 
> @fzdy1914 Will it be useful to refactor the code to take advantages of the states, otherwise, this is a very small case for refactoring the code

If the state here means Vuex, I don't think it is necessary.  A simple v-if, v-else should do the job.

dcshzj:
> > > @Tejas2805 I tried to make the tooltip for showing and hiding the commit message body responsive, but apparently it is not trivial as we have to refactor the code such that we can take advantage of states (currently, it is using the `window` functions). Is it possible to work on that as a separate pull request instead?
> > 
> > 
> > @fzdy1914 Will it be useful to refactor the code to take advantages of the states, otherwise, this is a very small case for refactoring the code
> 
> If the state here means Vuex, I don't think it is necessary. A simple v-if, v-else should do the job.

Yep, so ideally this is done on a separate pull request as Vuex does not store any state information in this part yet. We will need to refactor this part so that Vuex knows which commit message is open.

fzdy1914:
> Yep, so ideally this is done on a separate pull request as Vuex does not store any state information in this part yet. We will need to refactor this part so that Vuex knows which commit message is open.

Yes, I know the difficulty you met. But usually, we don't put such kind of information in Vuex. You can see how `v-segment` in `v_authorship` handles similar logic. Instead of the current CSS style toggle, we change it to JS style. Can you raise an issue about this? I am OK with the current style actually.

dcshzj:
> > Yep, so ideally this is done on a separate pull request as Vuex does not store any state information in this part yet. We will need to refactor this part so that Vuex knows which commit message is open.
> 
> Yes, I know the difficulty you met. But usually, we don't put such kind of information in Vuex. You can see how `v-segment` in `v_authorship` handles similar logic. Instead of the current CSS style toggle, we change it to JS style. Can you raise an issue about this? I am OK with the current style actually.

I raised this issue in #1340 so that we can follow-up on how toggling is done.

frontend/src/static/css/style.scss
```diff
@@ -58,6 +58,48 @@ a {
...
 
+.tooltip {
+  display: inline-block;
+  position: relative;
+
+  .tooltip-text {
+    background-color: #000000;
+    border-radius: .25rem;
+    bottom: 125%;
+    color: #ffffff;
```
Tejas2805, line 69: Can import this from `colors.scss`

Tejas2805 (commented):
Some comments: 
Screen Shot 2020-10-07 at 2 23 31 PM

This tooltip is supposed to be for an author and not a group. Possible to make it author or group, based on the kind of "group by"?

frontend/src/static/css/style.scss
```diff
@@ -58,6 +58,48 @@ a {
...
+.tooltip {
+  display: inline-block;
+  position: relative;
+
+  .tooltip-text {
+    background-color: mui-color('black');
+    border-radius: .25rem;
+    bottom: 125%;
+    color: mui-color('white');
+    font-size: .875rem;
```
Tejas2805, line 70: Just realized can use @mixin medium-font here. It's in the utils.scss file.

frontend/src/static/css/style.scss
```diff
@@ -58,6 +58,48 @@ a {
...
+    font-weight: normal;
```
Tejas2805, line 71: Same for this.

frontend/src/summary_charts.pug
```diff
@@ -73,28 +85,32 @@
...
+            span.tooltip-text Click to view author's profile
         a(
-          title="click to view author's code",
           onclick="deactivateAllOverlays()",
           v-on:click="openTabAuthorship(user, repo, j, isGroupMerged(getGroupName(repo)))"
         )
-          font-awesome-icon.icon-button(icon="code")
+          .tooltip
+            font-awesome-icon.icon-button(icon="code")
+            span.tooltip-text Click to view author's code
```
Tejas2805, line 106:
```suggestion
            span.tooltip-text Click to view author's contribution
```
I think this might be more suitable and will work in both the contexts.

Tejas2805 (commented):
Some comments. Otherwise LGTM!

Tejas2805 (approved):

jamessspanggg (approved):
LGTM

jamessspanggg:
@fzdy1914

PR #1332, opened 2020-09-27
[#1328] Fix usage of non-existent delimiter

Boon-Jun:
```
File type hashes in authorship tab were delimited with a
non-existent delimiter and this causes errors when restoring
selected file types from the hash.

Let's use an appropriate delimiter so that file type hashes 
can be parsed correctly. 
```
Fixes #1328

PR #1334, opened 2020-09-29
[#1322] Fix loading page rendering issue

### LINKED ISSUE #1322: Refine RepoSense Loading Page
image

There are a few seconds of the weird message appearing before the actual loading/request for the upload page. Can we alter the Vue life cycle to prevent them from showing?

jamessspanggg:
Fixes #1322 

```
The contents within v-resizer shows at the beginning of page load.

This is because we write our code directly to index.pug directly 
instead of a separate file e.g. app.pug. The generated html from 
index.pug contains some body html, which is recognised by the 
browser and displayed first.

Let's move the v-if conditional out of v-resizer to a vue template,
since the template tag is not recognizable by the browser.
```


jamessspanggg:
Can check the surge dashboard preview to see if the issue still occurs.

frontend/src/index.pug
```diff
@@ -34,73 +34,73 @@ html
     script(src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.14/build/languages/yaml.min.js")
   body
     #app
-      v-resizer(
-        v-if="userUpdated",
-        v-bind:is-tab-active="isTabActive",
-        v-on:close-tab="deactivateTab"
-      )
-        #summary-wrapper(slot="left")
-          .summary-status(v-if="loadedRepo<repoLength") Loading repos ... {{ loadedRepo }} / {{ repoLength }}
```
fzdy1914, line 43: Can move this line and the `loadedRepo` related attributes. We are not showing this anymore?

jamessspanggg:
As discussed with @fzdy1914, the issue arises because we write our code directly to `index.pug`, instead of a separate `app.pug`. During the early times of initial page load, some of the generated HTML from `index.pug` is recognised by the browser and displayed first (in our case it is contents within `v-resizer` in issue #1332). 

Placing a vue template outside of the `v-resizer` will fix the issue by ensuring that the contents to be not recognisable by the browser.

frontend/src/index.pug
jamessspanggg: do u mean remove this line?

fzdy1914 (approved):

PR #1343, opened 2020-10-18
[#590] Configuring report generation

### LINKED ISSUE #590: Give a way to set website properties

Boon-Jun:
```
The generated report currently does not allow the customization of
some website properties such as page title and favicons

Let's add a new model ReportConfiguration that allows customization 
of page titles via a json config file. Report Generation also takes an 
assets folder into consideration for the customization of favicons. 
These concepts could be extended further for other configurable settings.
```
Resolves #590

Following is an example of the structure of the assets directory right now.
Screenshot 2020-10-31 at 3 37 19 PM
At the moment, it simply copies favicons from the assets directory into the output directory.
I'm sure we can perform more complex functions with this directory later on.
An alternative to having an assets folder in the path is to use an `-assets` flag for a different directory

Next is an example of where to place the report config json files (Its placed within the config directory together with other configs)
Screenshot 2020-10-31 at 3 32 38 PM

Right now, this config file only helps us customize the title, but I believe we can include other cli related configurations here too, and allow users to override it whenever needed. 
This can work as a permanent storage of all configurations.
image


src/main/java/reposense/util/FileUtil.java
```diff
@@ -175,6 +177,13 @@ public static void copyTemplate(InputStream is, String outputPath) throws IOExce
         FileUtil.unzip(is, Paths.get(outputPath));
     }
 
+    /**
+     * Copies assets from {@code sourcePath} to the {@code outputPath}.
+     * @throws IOException if {@code is} refers to an invalid path.
+     */
+    public static void copyDirectoryContents(String sourcePath, String outputPath) throws IOException {
+        FileUtils.copyDirectory(new File(sourcePath), new File(outputPath));
```
fzdy1914, line 185: It may be better to write our own `copyDirectoryContents` since the method itself is not so complicated.

src/main/java/reposense/parser/AssetsFolderArgumentType.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Checks the argument of {@code --assets} flag.
+ */
+public class AssetsFolderArgumentType implements ArgumentType<Path> {
+    @Override
+    public Path convert(ArgumentParser parser, Argument arg, String value) throws ArgumentParserException {
+        // Piggyback on library methods to do file existence checks
+        Arguments.fileType().verifyExists().verifyIsDirectory().verifyCanRead().convert(parser, arg, value);
+        System.out.println("12312312312" + value);
```
dcshzj, line 20: Is this print expected?

src/main/java/reposense/model/ReportConfiguration.java
```diff
@@ -0,0 +1,20 @@
...
+public class ReportConfiguration {
+    private static final String DEFAULT_TITLE = "RepoSense Report";
+    private String title;
+
+    public ReportConfiguration() {
+    }
+
+    public String getTitle() {
+        if (title == null) {
+            return DEFAULT_TITLE;
```
dcshzj, line 15: Would it be more ideal if we set the `DEFAULT_TITLE` in the constructor? The title variable can then be overridden when parsing the configuration file.

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -116,9 +120,10 @@ private void generateReport(String inputDates) throws Exception {
         RepoConfiguration.setDatesToRepoConfigs(
                 repoConfigs, cliArguments.getSinceDate(), cliArguments.getUntilDate());
 
-        ReportGenerator.generateReposReport(repoConfigs, FT_TEMP_DIR, TEST_REPORT_GENERATED_TIME,
-                cliArguments.getSinceDate(), cliArguments.getUntilDate(), cliArguments.isSinceDateProvided(),
-                cliArguments.isUntilDateProvided(), () -> TEST_REPORT_GENERATION_TIME);
+        ReportGenerator.generateReposReport(repoConfigs, FT_TEMP_DIR, FT_ASSETS_TEMP_DIR, new ReportConfiguration(),
+                TEST_REPORT_GENERATED_TIME, cliArguments.getSinceDate(), cliArguments.getUntilDate(),
+                cliArguments.isSinceDateProvided(), cliArguments.isUntilDateProvided(), () ->
```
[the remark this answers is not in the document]
Boon-Jun: It was a trick that I'm using to avoid checkstyle errors for now. Maybe we can update checkstyle configurations in the future to account for this scenario?

src/main/java/reposense/model/ReportConfiguration.java
Boon-Jun: I'm ok with this design as it is. 🤔 If there is a consensus that setting the default title in the constructor is better, then lets change it.

Boon-Jun:
With the introduction of these configurables, long parameter lists for several methods & constructors are being flagged by Codacy. I'm thinking of creating a new class for all dates related settings, or alternatively include them under the newly introduced report-config.json file. Meaning that they will be bundled with the ReportConfiguration model. Are there any other ideas we can consider?

fzdy1914:
> With the introduction of these configurables, long parameter lists for several methods & constructors are being flagged by Codacy. I'm thinking of creating a new class for all dates related settings, or alternatively include them under the newly introduced report-config.json file. Meaning that they will be bundled with the ReportConfiguration model. Are there any other ideas we can consider?

Or we can ignore Codacy if the suggestion is not ideal.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -118,9 +126,24 @@
...
+    }
+
+    private static void setReportConfiguration(ReportConfiguration config, String outputPath) throws IOException {
+        setLandingPageTitle(outputPath, config.getTitle());
+    }
+
+    /**
+     * Set title of template file located at {@code filePath} to {@code pageTitle}
+     */
+    private static void setLandingPageTitle(String filePath, String pageTitle) throws IOException {
```
fzdy1914, line 142: Can we set the page title using Vue's functionality? Like we pass the title in the JSON file so the Vue can read and update it?

fzdy1914: I agree that in the case of pageTitle, this way may be better. But, in the case that we need more customized settings, we still need to pass them to frontend using JSON.

fzdy1914: Or we can seek to generate the Vue config file using the given parameters, including the title, and then build the Vue project. [Link](https://forum.vuejs.org/t/how-to-pass-command-line-parameters-to-vue/38701)

Boon-Jun: Yes, I think it is possible to let Vue read in page title and and customize it easily even after report generation. I don't have a preference on usage of a custom json file or vue config right now though, but I will dig a little deeper and update this PR accordingly.

Boon-Jun: ezgif com-optimize
Here is an example of how dynamically loading a title from a JSON file in Vue is like. Basically, there can be a noticeable delay in the loading of page titles. I dont think there is a suitable Vue lifecycle event that allows us to inject a page title without this delay.

Boon-Jun: Also regarding vue config, and if I did my research right, vue config is used together with Vue CLI, which is mainly used for development purposes, and I’m not really comfortable with adding such a requirement on the users for the running of RepoSense.

Boon-Jun: I’ve also tried looking up on other solutions, but they are also mainly designed around web dev too.
Maybe someone has a better suggestion on how a page title can be changed via frontend means?

fzdy1914: Can't the config be used during building of vue project?

Boon-Jun: Correct me if I’m wrong. But i dont think running reposense right now, involves an extra build step for the building of the vue project, and most of the required templates/js/css files have already been pre-generated and aggregated into the JAR for usage?

Boon-Jun: Also, when we talk about building a vue project. do you mean letting the users build the project themselves? Or do you mean abstracting the build process within the executable?

fzdy1914: Technically, we have a self-defined building step for frontend called `spuild` (you can check in npm run). It is meant to be deprecated, but is still used now. In the near future, we are going to turn our frontend into a formal vue project as describe in #795.

fzdy1914: I will also check the logic of Jar file.

fzdy1914: Yes, you are correct. When running the Jar file, we don't need to build the report again. Seems that there are only two options remaining. One is to edit the title directly. The second is the dynamic loading. 

I am not sure if Prof is OK if we can view a noticeable delay in the loading of page titles. We may first show a `RepoSense Report` title rather than `Example Sample` at the first step of loading. So the following change of title won't cause much misunderstanding.

@damithc  The situation is like this, we want to pass the title defined in a user-defined config to frontend. And there are two ways.
1. We modify the relative part in the HTML code directly. But the extensibility is lower and we cannot extend it to further attributes.

2. We dynamically load the attribute from a JSON. But we can view a noticeable delay in the loading of page titles. The disadvantage maybe somehow mitigated by using a `RepoSense Report` as default title.

Boon-Jun:
Hi all, I'll have to wrap up this PR before a module deadline this Sunday. Any comments on how to proceed are greatly appreciated!

src/main/java/reposense/report/ReportGenerator.java
fzdy1914: @damithc Prof, your input is needed here.

damithc: Sorry, didn't notice it before. 
For 1, what are other attributes we could extend to? If we can't think of any, we can go with 1.
I'm OK with 2 too.

Boon-Jun: Currently, I dont have any plans to extend this functionality. But using html parsers can be a viable solution if we want to extend it to other attributes. But might be too complex given how simple our use case is right now?

damithc: I see. I'm OK with both options. You guys can choose.

fzdy1914: The related functionality is mentioned in #467. Then, let's go with option 1.

jamessspanggg (commented):
Looking at the PR briefly (sorry I have exams going on these few days, will look more thoroughly during the weekends), there are some initial comments:
 - Please update the UG on the new json file and the `--assets` command
- I feel like the timezone can be included in the new json file, since it is shared among all repositories under the same analysis (future work)
- Ensure that your commit message is wrapped within 120 chars
- Overall logic looks good to me atm

Boon-Jun:
a friendly ping for this PR!

jamessspanggg (approved):
LGTM. When this is merged, please create an issue on adding other settings that's shared by the entire report, i.e. `timezone` to the report-config.json.

fzdy1914 (approved):

PR #1346, opened 2020-10-26
[#816] Generate the report based on the user's timezone

### LINKED ISSUE #816: Generate since date and until date based on timezone
> The `timezone` via `-t` is only used for the generation time of the report, but not affecting to the `sinceDate` and `untilDate` of the report. The `sinceDate` and `untilDate` is generated based on the system's default.

It would be nice to have the timezone considered for all time-related things such as the start time. That can be a separate PR though.

_Originally posted by @damithc in https://github.com/reposense/RepoSense/pull/789_

Let's modify all time-related things (since time and until time) to be generated based on the specified `timezone` as well.

dcshzj:
Fixes #816.

```
The report is not currently configured to respect the --timezone
configuration setting, which may cause issues where a commit is
incorrectly attributed to the wrong date in terms of the user's
local time zone.

Let's introduce a functionality that makes the report generation
more aware about the --timezone configuration setting, such that
any date and time related functionality is made timezone-aware.
```

Due to the limitations of the PR preview, the following are the expected output for commits made by `jamessspanggg`.

In this case, the time zone is set to the default setting (`ZoneId.systemDefault()`), which is set to `Asia/Singapore` on my system. Hence, the two commits will be recognised as being committed on 12 May 2020 and 8 May 2020 respectively.

In this case, the time zone is set to UTC+11:00. As expected, the commit dates are recognised as being committed "on the next day", which is 13 May 2020 and 9 May 2020 respectively. This is because in the UTC+11:00 time zone, the commits were made "on the next day".

Previously, the report generation process did not make any distinction between both time zones. Other commits that do not span across different days for the two time zones are correctly recognised as being committed on the same day.

Also, due to the use of the `--timezone` flag, the `--since` and `--until` flags will be based on the time zone provided by the user.


src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -24,25 +27,26 @@
     private static final int DAYS_IN_MS = 24 * 60 * 60 * 1000;
 
     /**
-     * Returns the {@code CommitContributionSummary} generated from aggregating the {@code commitResults}.
+     * Returns the {@code CommitContributionSummary} generated from aggregating the {@code commitResults}, where the
+     * time is given in {@code zoneId} timezone.
      */
     public static CommitContributionSummary aggregateCommitResults(
-            RepoConfiguration config, List<CommitResult> commitResults) {
+            RepoConfiguration config, List<CommitResult> commitResults, ZoneId zoneId) {
```
jamessspanggg, line 34: I think we should place the `zoneId` as a new field within `RepoConfiguration`, instead of passing it directly as a parameter. Then to access the `zoneId`, can just call `config.getZoneId()`. This is to standardise with all the other types of configurations.

jamessspanggg (commented):
Some initial questions:
- Would the system test fails if the tester's default system time zone is different from our SG timezone? If so, we might need to set a specific timezone for these tests.

dcshzj:
> Some initial questions:
> 
> * Would the system test fails if the tester's default system time zone is different from our SG timezone? If so, we might need to set a specific timezone for these tests.

For system tests, it uses the tester's system default timezone which is declared in `SYSTEM_DEFAULT_ZONE_ID` in the relevant test files. This constant uses `ZoneId.systemDefault()` to adapt to the system that the test is running on.

src/main/java/reposense/commits/CommitResultAggregator.java
dcshzj: Okay sure, I will tweak the design of the configuration.

jamessspanggg:
> For system tests, it uses the tester's system default timezone which is declared in `SYSTEM_DEFAULT_ZONE_ID` in the relevant test files. This constant uses `ZoneId.systemDefault()` to adapt to the system that the test is running on.

Yep, that's exactly my point. If I change my laptop's timezone now and run the system tests, it would fail (since we verify the output with a pre-generated json that is in our sg timezone). That is why I suggested to specify a specific timezone for the system tests.

jamessspanggg:
Can u check why the CI is failing?

dcshzj:
> Can u check why the CI is failing?

Not fully complete yet. There's a weird bug where the `getDateMinusNDays` and related functions don't seem to be aware of the timezone despite setting it explicitly. Will continue debugging tomorrow and open to any suggestions on how to resolve this.

dcshzj:
Ideally, we should extend the system tests to consider different time zones as well. However, I felt that it might cause the testing time to be significantly increased due to the current setup of the test suite. More robust tests can probably be added after #1349 is resolved.

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -122,16 +127,64 @@ private static void addDailyContributionForNewDate(
...
-        cal.setTime(current);
+    /**
+     * Get the starting point of the {@code current} date with respect to the {@code zoneId} timezone.
+     */
+    private static Date getStartOfDate(Date current, ZoneId zoneId) {
+        if (current.equals(SinceDateArgumentType.ARBITRARY_FIRST_COMMIT_DATE)) {
+            return current;
+        }
+
+        Instant now = Instant.now();
```
jamessspanggg, line 138: Since line 138 until line 142 is being re-used to get the `systemRawOffset - zoneRawOffset` in multiple methods, perhaps can refactor to a separate method?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -174,7 +177,10 @@ public void parse_withExtraWhitespaces_success() throws Exception {
 
     @Test
     public void parse_configFolderOnly_success() throws Exception {
-        String input = new InputBuilder().addConfig(CONFIG_FOLDER_ABSOLUTE).build();
+        String input = new InputBuilder()
+                .addConfig(CONFIG_FOLDER_ABSOLUTE)
+                .addTimezone(DEFAULT_TIMEZONE)
```
jamessspanggg, line 182: is it necessary to add the timezone into the input builder as these tests are specifically for certain flags? i.e. for this test, its testing for the config flag

src/systemtest/resources/dateRange/expected/summary.json
```diff
@@ -10,7 +10,10 @@
       },
       "branch": "master",
       "displayName": "reposense/testrepo-Alpha[master]",
-      "outputFolderName": "reposense_testrepo-Alpha_master"
+      "outputFolderName": "reposense_testrepo-Alpha_master",
+      "zoneId": {
+        "id": "Asia/Singapore"
```
jamessspanggg, line 15: is there any reasoning in using an object for the `zoneId`? would just a key value pair `"zoneId": "Asia/Singapore" ` suffice?

jamessspanggg (commented):
Looks good to me. One more thing from me, can you add tests with timezones set to others such as `UTC-1100`? 

>In this case, the time zone is set to UTC+11:00. As expected, the commit dates are recognised as being committed "on the next day", which is 13 May 2020 and 9 May 2020 respectively. This is because in the UTC+11:00 time zone, the commits were made "on the next day".


Can be the example you gave in ur PR message

jamessspanggg (approved):
LGTM

src/test/java/reposense/authorship/FileAnalyzerTest.java
```diff
@@ -56,6 +66,8 @@ public void blameTestDateRange() throws Exception {
     @Test
     public void movedFileBlameTestDateRange() throws Exception {
         GitCheckout.checkoutDate(config.getRepoRoot(), config.getBranch(), MOVED_FILE_UNTIL_DATE);
+        System.out.println(MOVED_FILE_SINCE_DATE);
```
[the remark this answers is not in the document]
dcshzj: Oops, removed them.

src/systemtest/resources/sinceBeginningDateRange/expected/summary.json
```diff
@@ -30,7 +32,8 @@
       },
       "branch": "master",
       "displayName": "reposense/testrepo-Charlie[master]",
-      "outputFolderName": "reposense_testrepo-Charlie_master"
+      "outputFolderName": "reposense_testrepo-Charlie_master",
+      "zoneId": "Asia/Singapore"
```
fzdy1914, line 36: Just a question, why we need zoneId in `summary.json`? Can set the attribute not to display when generating the JSON.

fzdy1914 (changes requested):
I recall that the original `-t` should support all kind of timezone strings, do the current implementation all support them?

Meanwhile, it is better to have a test for different timezone apart from Singapore though.

dcshzj:
> I recall that the original `-t` should support all kind of timezone strings, do the current implementation all support them?

Yes, the original implementation has not been altered.

> Meanwhile, it is better to have a test for different timezone apart from Singapore though.

This is blocked by #1349. There are some basic tests for testing different time zones and how the `sinceDate` and `untilDate` change depending on the timezone used.

src/systemtest/resources/sinceBeginningDateRange/expected/summary.json
dcshzj: Without the zoneId, the dates used in generating the report can get ambiguous, especially since the dates will be reported differently for different time zones.

dcshzj:
Any other changes needed? I will have to wrap up this PR by this coming Sunday, thank you!

src/systemtest/resources/sinceBeginningDateRange/expected/summary.json
fzdy1914: But now, we do not support a different `zoneId` for different config, right? Even we need to pass it to frontend, there is no need to write it again and again in `summary.json`?

fzdy1914:
> This is blocked by #1349. There are some basic tests for testing different time zones and how the `sinceDate` and `untilDate` change depending on the timezone used.

So you mean that as long as we can test the correct generation of `sinceDate` and `untilDate` from different `timezones`, it is enough? Why it is related to #1349?

dcshzj:
> > This is blocked by #1349. There are some basic tests for testing different time zones and how the `sinceDate` and `untilDate` change depending on the timezone used.
> 
> So you mean that as long as we can test the correct generation of `sinceDate` and `untilDate` from different `timezones`, it is enough? Why it is related to #1349?

The use of time zones primarily affects 2 things. One is the way the `sinceDate` and `untilDate` are interpreted (the exact cut off timestamp) and the other being the dates that commits are attributed. The interpretation of the `sinceDate` and `untilDate` have some basic tests covered for them, which serves as a good indicator to see if the time zone supplied by the user is correctly adhered to.

The second way on the dates for the commits are blocked by #1349 as it requires adding more system tests, since it is dependent on the exact timing that a particular commit is made. However, adding more system tests will lengthen the testing time as the same repositories have to be cloned again, which is something that I hope to avoid first.

src/systemtest/resources/sinceBeginningDateRange/expected/summary.json
fzdy1914: @dcshzj Can you help answer this question? Why we cannot set it as a global attribute?

dcshzj: My apologies, I misunderstood the original intent. I moved it out to a global attribute.

fzdy1914 (approved):

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -174,30 +177,34 @@ public void parse_withExtraWhitespaces_success() throws Exception {
 
     @Test
     public void parse_configFolderOnly_success() throws Exception {
-        String input = new InputBuilder().addConfig(CONFIG_FOLDER_ABSOLUTE).build();
+        String input = new InputBuilder()
```
fzdy1914, line 180: No need to split it into multiple line?

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -174,30 +177,34 @@ public void parse_withExtraWhitespaces_success() throws Exception {
...
-        assertDateDiffOneMonth(cliArguments.getSinceDate(), cliArguments.getUntilDate());
-        assertDateDiffEndOfDay(cliArguments.getUntilDate());
+        assertDateDiffOneMonth(cliArguments.getSinceDate(), cliArguments.getUntilDate(), cliArguments.getZoneId());
+        assertDateDiffEndOfDay(cliArguments.getUntilDate(), cliArguments.getZoneId());
         Assert.assertEquals(ArgsParser.DEFAULT_REPORT_NAME, cliArguments.getOutputFilePath().getFileName().toString());
         Assert.assertEquals(FileTypeTest.NO_SPECIFIED_FORMATS, cliArguments.getFormats());
         Assert.assertFalse(cliArguments.isAutomaticallyLaunching());
 
-        input = new InputBuilder().addConfig(CONFIG_FOLDER_RELATIVE).build();
+        input = new InputBuilder()
```
fzdy1914, line 196: same here

fzdy1914 (changes requested):
Sorry for the last minute request. It will be better if u can edit it.

fzdy1914 (approved):

PR #1348, opened 2020-10-27
[#1345] Authorship: Add last modified date to each LoC if specified

### LINKED ISSUE #1345: Authorship.json: give per-line last-modified-date
Current:
image

I'm wondering is it possible to also give last-modified-date for each line. While not strictly not necessary for the dashboard, I can use that data in further processing when I grade students' work.

This will increase the size of the file significantly though. Perhaps we can disable by default with an option to enable it using a flag.

The time zone (if provided) needs to be considered as well.

jamessspanggg:
Fixes #1345 

```
The last modified date of each line of code is not present in the
authorship.json.

It would be beneficial for tutors to use the last modified date of
each line of code to do further processing.

Let's add the last modified date of each line of code within the
authorship.json. This functionality will only be enabled when the
user specifies the flag '--last-modified-date' or '-l' in short.
```

Snapshot of updated `authorship.json`:
Screen Shot 2020-10-27 at 8 31 01 PM


src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -75,11 +75,22 @@ public void test30DaysFromUntilDate() throws Exception {
...
         Path actualFiles = loadResource(getClass(), "dateRange/expected");
         verifyAllJson(actualFiles, FT_TEMP_DIR);
     }
 
+    /**
+     * System test with a specified since date and until date, with the last modified date time in each
+     * line of code.
+     */
+    @Test
+    public void testDateRangeWithModifiedDateTimeInLines() throws Exception {
```
jamessspanggg, line 88: Added another system test for this functionality, however, I feel that this is redundant and takes up space. Should I merge this test with the `testDateRange` as one?

jamessspanggg:
Hi Prof @damithc, 
- would it be okay if i use the `--last-modified-date` or `-l` to toggle this feature? the feature will be disabled by default.

With using the following args `--config ./config -v -s d1`,
- performance wise: 33.06 second(s) (without the `-l`), 35.06 second(s) (with the `-l`)
- total size of authorship.json: 190kb (without the `-l`), 213kb (with the `-l`)

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -58,25 +58,26 @@ public void tearDown() throws Exception {
...
     /**
-     * System test with a specified since date and until date.
+     * System test with a specified since date and until date, with the last modified date time in each
+     * line of code.
      */
     @Test
-    public void testDateRange() throws Exception {
-        generateReport(getInputWithDates("1/9/2017", "30/10/2017"));
-        Path actualFiles = loadResource(getClass(), "dateRange/expected");
+    public void testDateRangeWithModifiedDateTimeInLines() throws Exception {
```
jamessspanggg, line 78: Created this system test that merges with the previous `testDateRange` as it feels redundant to have another system test that may take up our build time and space.

Tejas2805:
> * performance wise: 33.06 second(s) (without the `-l`), 35.06 second(s) (with the `-l`)
> * total size of authorship.json: 190kb (without the `-l`), 213kb (with the `-l`)

Which repo are the above parameters for?

jamessspanggg:
> Which repo are the above parameters for?
> 
> 
> 
> 

The ones in our repo config.

damithc:
> Hi Prof @damithc,
> 
> * would it be okay if i use the `--last-modified-date` or `-l` to toggle this feature? the feature will be disabled by default.

Sounds good. In the documentation, mention `lower case L` to prevent someone reading it as upper case i.

jamessspanggg:
> Sounds good. In the documentation, mention `lower case L` to prevent someone reading it as upper case i.

Noted prof. Have done so.

docs/ug/cli.md
```diff
@@ -16,10 +16,10 @@ The command `java -jar RepoSense.jar` takes several flags.
 **Examples**:
 
 An example of a command using most parameters:<br>
-`java -jar RepoSense.jar --repo https://github.com/reposense/RepoSense.git --output ./report_folder --since 31/1/2017 --until 31/12/2018 --formats java adoc xml --view --ignore-standalone-config --timezone UTC+08`
+`java -jar RepoSense.jar --repo https://github.com/reposense/RepoSense.git --output ./report_folder --since 31/1/2017 --until 31/12/2018 --formats java adoc xml --view --ignore-standalone-config --last-modified-date --timezone UTC+08`
 
 Same command as above but using most parameters in alias format:<br>
-`java -jar RepoSense.jar -r https://github.com/reposense/RepoSense.git -o ./report_folder -s 31/1/2017 -u 31/12/2018 -f java adoc xml -v -i`
+`java -jar RepoSense.jar -r https://github.com/reposense/RepoSense.git -o ./report_folder -s 31/1/2017 -u 31/12/2018 -f java adoc xml -v -i -l`
```
Tejas2805, line 22: Just realised. We say same command as above but don't include `--timezone`. Can include it?

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -78,9 +79,10 @@ private static FileResult generateFileResult(FileInfo fileInfo) {
     }
 
     /**
-     * Sets the {@code Author} for each line in {@code fileInfo} based on the git blame analysis on the file.
+     * Sets the {@code Author} and {@code Date} for each line in {@code fileInfo} based on the git blame analysis
+     * on the file.
      */
-    private static void aggregateBlameAuthorInfo(RepoConfiguration config, FileInfo fileInfo) {
+    private static void aggregateBlameAuthorModifiedAndDateInfo(RepoConfiguration config, FileInfo fileInfo) {
```
Tejas2805, line 85: Not sure if this change is required. Since it's technically a part of the `aggregateBlame`. @fzdy1914 your thoughts?

Tejas2805 (commented):
Left 2 comments. 

Just one quick thought: Should we include in our documentation that it will affect the performance and size of the json file? Though it's obvious since extra stuff is being added, but might be a good thing to mention?

Everything else LGTM!

docs/ug/cli.md
jamessspanggg: I think that can be included in #1346 as it is more relevant to this change

src/main/java/reposense/authorship/FileInfoAnalyzer.java
jamessspanggg: how about `aggregateBlameInfo`?

docs/ug/cli.md
dcshzj: Yep, I think the last modified date needs to be made timezone-aware as well

jamessspanggg: Yep, placing this PR on hold until #1346 merges

dcshzj: #1346 was merged, I think we can move ahead with this PR?

jamessspanggg:
@dcshzj I refactored some parts of ur code to `TimeUtil` as there are multiple parts utilising the same method.

Since you have worked with timezone related issues, could you check if my implementation related to timezones are ok? especially the `convertDateWithZones` method.

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -66,4 +66,13 @@ public static int getZoneRawOffset(ZoneId zoneId) {
         ZoneOffset zoneOffset = zoneId.getRules().getOffset(now);
         return zoneOffset.getTotalSeconds() * 1000;
     }
+
+    /**
+     * Get the {@code current} date that is in {@code fromZoneId} timezone into {@code toZoneId} timezone.
+     */
+    public static Date convertDateWithZones(Date current, ZoneId fromZoneId, ZoneId toZoneId) {
+        int zoneRawOffset = getZoneRawOffset(toZoneId);
+        int systemRawOffset = getZoneRawOffset(fromZoneId);
```
dcshzj, line 75: I think the variable should not use `systemRawOffset` as it can be confused as referring to the actual system time zone.

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -66,4 +66,13 @@ public static int getZoneRawOffset(ZoneId zoneId) {
         ZoneOffset zoneOffset = zoneId.getRules().getOffset(now);
         return zoneOffset.getTotalSeconds() * 1000;
     }
+
+    /**
+     * Get the {@code current} date that is in {@code fromZoneId} timezone into {@code toZoneId} timezone.
+     */
+    public static Date convertDateWithZones(Date current, ZoneId fromZoneId, ZoneId toZoneId) {
```
dcshzj, line 73: We might be able to use the functionality in this function for `getSystemDateFromZonedDate`, since it is using the same logic.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -99,6 +103,13 @@ private static void aggregateBlameAuthorInfo(RepoConfiguration config, FileInfo
                 author = Author.UNKNOWN_AUTHOR;
             }
 
+            if (config.shouldIncludeLastModifiedDate()) {
+                // convert the commit date from the author time zone to cli-specified timezone
+                Date convertedCommitDate = TimeUtil.convertDateWithZones(new Date(commitDateInMs),
+                        ZoneId.of(authorTimeZone), ZoneId.of(config.getZoneId()));
```
dcshzj, line 109: I can't remember exactly, but if `commitDateInMs` is a number that is the number of milliseconds from epoch, then we need to convert from the system time zone to the user-specified time zone, not from the author's time zone.

The milliseconds represents a point in time, which is converted to a `Date` object based on the system time zone. We then convert the date (in the system time zone) to the user-specified time zone so that when we drop the hour, minute and seconds, the year, month and day is still correct.

dcshzj (commented):
> @dcshzj I refactored some parts of ur code to `TimeUtil` as there are multiple parts utilising the same method.
> 
> Since you have worked with timezone related issues, could you check if my implementation related to timezones are ok? especially the `convertDateWithZones` method.

For `ArgsParserTest.java`, there is also a `getZoneRawOffset` method but it is in the test suite. Should we also extract it out, since the test you introduced in `FileAnalyzerTest.java` will also require adapting it to the correct timezone?

src/main/java/reposense/util/TimeUtil.java
dcshzj: I think we should stick to using `getSystemDateFromZonedDate`? Right now I think the `fromZoneId` will always be the system default, which was the original implementation.

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -43,4 +48,22 @@ public static String getElapsedTime() {
...
+    public static int getZoneRawOffset(ZoneId zoneId) {
+        Instant now = Instant.now();
+        ZoneOffset zoneOffset = zoneId.getRules().getOffset(now);
+        return zoneOffset.getTotalSeconds() * 1000;
+    }
+
+    /**
+     * Get the {@code current} date that is in {@code fromZoneId} timezone into {@code toZoneId} timezone.
+     */
+    public static Date convertDateWithZones(Date current, ZoneId fromZoneId, ZoneId toZoneId) {
```
dcshzj, line 64: I think we should stick to using `getSystemDateFromZonedDate`? Right now the `fromZoneId` will always be the system default, which was the original implementation. I don't think we will have a case where the `Date` object is not in the system time zone.

dcshzj (approved):
LGTM 👍

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -99,6 +103,13 @@ private static void aggregateBlameAuthorInfo(RepoConfiguration config, FileInfo
                 author = Author.UNKNOWN_AUTHOR;
             }
 
+            if (config.shouldIncludeLastModifiedDate()) {
+                // convert the commit date from the author time zone to cli-specified timezone
+                Date convertedCommitDate = TimeUtil.getSystemDateFromZonedDate(new Date(commitDateInMs),
```
fzdy1914, line 108: Should the method be `getZonedDateFromSystemDate`?

src/main/java/reposense/model/LocationsCliArguments.java
```diff
@@ -13,14 +13,16 @@
 
     public LocationsCliArguments(List<String> locations, Path outputFilePath, Path assetsFilePath, Date sinceDate,
             Date untilDate, boolean isSinceDateProvided, boolean isUntilDateProvided, List<FileType> formats,
-            boolean isAutomaticallyLaunching, boolean isStandaloneConfigIgnored, ZoneId zoneId) {
+            boolean shouldIncludeLastModifiedDate, boolean isAutomaticallyLaunching, boolean isStandaloneConfigIgnored,
```
fzdy1914, line 16: Change the variable name to `isLastModifiedDateIncluded` to match the other variable name?

src/main/java/reposense/authorship/FileInfoAnalyzer.java
jamessspanggg: @dcshzj may i get ur previous rationale in naming the method?

dcshzj: Yes, I think it might be more appropriate to rename it to `getZonedDateFromSystemDate`. The Javadoc comment will need to be updated accordingly.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -99,6 +103,13 @@ private static void aggregateBlameAuthorInfo(RepoConfiguration config, FileInfo
                 author = Author.UNKNOWN_AUTHOR;
             }
 
+            if (config.isLastModifiedDateIncluded()) {
+                // convert the commit date from the author time zone to cli-specified timezone
```
fzdy1914, line 107: from system default time zone to cli specified timezone?

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -155,8 +154,8 @@ private static Date getStartOfDate(Date current, String zoneId) {
      * system's timezone.
      */
     private static Date getSystemStartOfDate(Date current, String zoneId) {
```
fzdy1914, line 156: Can double confirm the usage of this method?

fzdy1914 (commented):
Some comment

src/main/java/reposense/commits/CommitResultAggregator.java
fzdy1914: My bad for not looking into the detail of these two methods in the previous PR.

dcshzj: `getSystemStartOfDate` gets the instant where it says 00:00:00 in the system time zone for the given date. `getStartOfDate` gets the instant where it says 00:00:00 in the given `zoneId` time zone for the given date.

We still need the system time zone because Gson does not allow for a time zone to be specified, so they only recognise the system default time zone.

PR #1350, opened 2020-10-31
[#460] Allow specifying a range of commits to ignore

### LINKED ISSUE #460: Config files: allow specifying commit range to specify
Current: each commit to ignore needs to be specified individually in config files
image

Suggested: Allow specifying commit ranges

dcshzj:
Fixes #460.

```
Currently, each commit that needs to be ignored has to be
specified accordingly, one entry at a time. This may be
troublesome for use cases where a range of commits has to be
ignored.

Let's introduce a functionality that allows users to specify a
range of commits using the Git ".." notation, so as to ignore
commits that are within the given two commits inclusive in the
report generation.
```

The start and end commit hash represents the entire range of commits to ignore. All commits within that range, plus the start and the end commits will all be added to the ignore list.


dcshzj:
Any comments? I will have to wrap up this PR by this coming Sunday, thank you!

src/main/java/reposense/model/CommitHash.java
```diff
@@ -49,16 +52,31 @@ public int hashCode() {
...
         return commits.stream()
+                .flatMap(x -> getHashOrHashes(root, branchName, x))
                 .map(CommitHash::new)
                 .collect(Collectors.toList());
     }
 
+    /**
+     * Converts a commit {@code entry} into either itself, or a list of hashes if a range was provided.
+     */
+    public static Stream<String> getHashOrHashes(String root, String branchName, String entry) {
```
jamessspanggg, line 70: I think naming it as `getHashes` would be fine too

jamessspanggg (commented):
Just a question: what happens if the start hash is later than the end hash?

dcshzj:
> Just a question: what happens if the start hash is later than the end hash?

The underlying functionality uses `rev-list`, so if start hash is later than end hash, no output will be returned by the command. The function then returns only the start hash and the end hash.

I think the logic might be a bit weird, so let me come up with a fix where we try both ways and see which of the commands returns a result.

src/main/java/reposense/git/GitRevList.java
```diff
@@ -57,15 +57,28 @@ public static String getCommitHashInRange(String root, String branchName, String
             return startHash;
         }
 
-        Path rootPath = Paths.get(root);
-        String revListCommand = "git rev-list " + startHash + ".." + endHash + " " + branchName
-                + REVISION_PATH_SEPARATOR;
-        String output = runCommand(rootPath, revListCommand);
+        String fromStartHash = getAllCommitHashSince(root, branchName, startHash);
```
jamessspanggg, line 60: what happens if the start hash = end hash? might need to handle

src/main/java/reposense/git/GitRevList.java
```diff
@@ -57,15 +57,28 @@ public static String getCommitHashInRange(String root, String branchName, String
...
+        String fromEndHash = getAllCommitHashSince(root, branchName, endHash);
+        StringBuilder output = new StringBuilder();
 
-        if (output.equals("")) {
-            return endHash + "\n" + startHash;
+        if (fromStartHash.length() > fromEndHash.length()) {
```
jamessspanggg, line 64: I understand that this segment of code is trying to find which hash is earlier, might need to refactor or place a comment here as it is slightly unclear and hard to understand at first glance

src/main/java/reposense/git/GitRevList.java
```diff
@@ -57,15 +57,28 @@ public static String getCommitHashInRange(String root, String branchName, String
...
+            output.append(fromStartHash.substring(fromEndHash.length()));
```
jamessspanggg, line 65: should this be `fromStartHash.substring(0, fromStartHashLength - fromEndHashLength)` instead?

src/main/java/reposense/git/GitRevList.java
```diff
@@ -57,15 +57,28 @@ public static String getCommitHashInRange(String root, String branchName, String
...
+            output.append(startHash);
         } else {
-            return output + startHash + "\n";
+            output.append(fromEndHash.substring(fromStartHash.length()));
```
jamessspanggg, line 68: same goes for this, should this be `fromEndHash.substring(0, fromEndHash - fromStartHashLength)` instead?

jamessspanggg (commented):
Some more questions:

What happens if the start or end hash is invalid, i.e. the hash does not exist as a commit?

src/main/java/reposense/git/GitRevList.java
dcshzj: The `rev-list` command outputs the list of commits starting from `HEAD` until the given commit, so this part is removing the parts that are common between both `fromStartHash` list and `fromEndHash` list.

For example, if `fromStartHash` has:
```
abc
def
ghi
123
```

And `fromEndHash` has:
```
abc
def
```

Then, I want to get just `ghi` and `123`, since that is the result of the set difference, and that starts from the index which corresponds to the length of `fromEndHash`.

I think it is quite confusing at the first glance, so I have added a comment to explain what it is intending to do.

src/main/java/reposense/git/GitRevList.java
dcshzj: This is implicitly handled later on when computing the set difference, due to the additional step where we add back one of the hashes. If both sets of commits are the same, then the set difference will generate an empty list, then the original hash is added to the list again to output only the given hash.

dcshzj:
> Some more questions:
> 
> What happens if the start or end hash is invalid, i.e. the hash does not exist as a commit?

I overlooked this by accident. The `rev-list` command will throw an error, so I added some code to handle that exception.

src/main/java/reposense/git/GitRevList.java
jamessspanggg: noted, thanks for the clear explanation 👍🏻 abc -> def -> ghi -> 123 sorted based on the most recent commit right

jamessspanggg (approved):

src/main/java/reposense/git/GitRevList.java
dcshzj: Yep, the newest commit is `abc` and the oldest commit is `123`.

fzdy1914 (approved):

PR #1355, opened 2020-11-08
[#1341] Hash sort order in commits panel

### LINKED ISSUE #1341: Hashing of settings in commits panel
> Hashing of filters (and any other setting in the page) could be useful as we want to be able to share a report exactly as seen by the viewer by simply sending the current URL to someone else. But we should do it in a way that can be applied to all pages rather than just the commits page.

Originally posted by @damithc in #1239 
It would be good if we can hash the settings of commits panel for the above mentioned reasons.
Potential settings that can be hashed includes sort by attribute, sorting order, hiding of commit messages.

Boon-Jun:
```
Currently, commits panel already supports the hashing of several attributes
(e.g fileTypes, Author, Repo etc.), but not the sorting order of commits.

To allow the sharing of sort settings via the URL, and also for completeness, 
let's hash the sorting order of commits in the commits-panel.
```

Resolves #1341


Boon-Jun:
Hi all, I'll have to wrap up this PR before a module deadline this Sunday. Any comments on are greatly appreciated!

frontend/cypress/tests/zoomView/zoomView_reload.js
```diff
@@ -9,10 +9,10 @@ describe('reload page', () => {
         .click();
 
     cy.get('#tab-zoom > .sorting > .sort-by > select:visible')
-        .select('Time');
+        .select('LoC');
```
fzdy1914, line 12: Why the default controls will be affected here? I didn't see the related code change?

0blivious:
Restoring not working correctly
Try open zoom panel -> select sort by LOC + Ascending -> refresh the page

frontend/cypress/tests/zoomView/zoomView_reload.js
Boon-Jun: The default controls are not affected here. But the purpose of selecting LoC is to ensure that it remains selected upon reload.

frontend/src/static/js/v_zoom.js
```diff
@@ -145,6 +153,22 @@ window.vZoom = {
...
+    },
+
+    retrieveSortHash() {
+      window.decodeHash();
+      const hash = window.hashParams;
+      if (hash.zCST) {
+        this.commitsSortType = hash.zCST;
+      }
+      if (hash.zRSC) {
+        this.toReverseSortedCommits = (hash.zRSC === 'true');
```
Tejas2805, line 168: Why do we need the `hash.zRSC === 'true'` ?

Boon-Jun: hashes are stored as a string, I did this to convert the string into a boolean

0blivious (approved):

jamessspanggg (approved):

frontend/cypress/tests/zoomView/zoomView_reload.js
fzdy1914: OK, understood

Issue #1202, opened 2020-04-01
Missing javadoc for functions


Tejas2805:
There are many methods in the backend which don't have any javadoc This makes it difficult for first-time developers to understand the backend code. 

Suggestion: Let's add javadoc for missing functions so that it becomes easier for someone reading the code to do it.

Tejas2805:
Let's make this a `FirstTimer issue`? It will help them understand backend better.

anubh-v:
Which methods are these? (any specific class / package ? )

Tejas2805:
The `addAuthorEmailsAndAliasesMapEntry` in AuthorConfiguration for example doesn't have it. Though it is not something too important to add, such methods when edited, cause an error in Travis, due to having no javadoc. Therefore, I feel it is better to find such methods and add them.

anubh-v:
I think our checkstyle config does check for presence of javadocs.
Perhaps we can investigate if the config is sufficiently strict in detecting missing javadoc comments?

For the specific case you mentioned, I think it can be a first timer issue, but perhaps issue #1207 should be resolved first as it affects the same data structure

HCY123902:
Hello, I am looking for a first time issue, and I am wondering whether this issue is still open? I seem to have found some functions without documentation and I think this may be a point for me to start.

damithc:
I think it's best if someone who knows the codebase well writes the javadoc comments. The outcome can be of better quality that way.

yhtMinceraft1010X:
Do we also need javadoc for private functions like the one below from `CommitResultAggregator`?
```
private static float getContributionVariance(List<AuthorDailyContribution> contributions,
            Date startDate, Date lastDate, String zoneId) {
        << Actual code not shown >>
    }
```

I can add them in if needed since I've explored most of the backend code while fixing #1601.

PR #1364, opened 2020-12-10
[#1202] Add Javadoc for backend functions

HCY123902:
#1202 
Commit  message:
```
Add Javadoc to methods in StreamGobbler, CustomLogFormatter, 
FileType, and RepoConfiguration.

Some methods in these classes do not seem to have Javadoc 
although the methods are not standard getter or setter.

To enhance readability, it may be necessary to explain what 
they do, especially for run() in StreamGobbler since the method
name may not give much information.

Additionally, there seems to be a colon missing in the developer guide
architecture section.

Let's add Javadoc to these methods and add the missing punctuation
to the developer guide to make documentation more consistent.
```
The missing colon in the developer guide architecture section is something that I happen to notice when going through it. 
I previously plan to split this into 2 separate pull requests, but since only minor changes are made, I decide to combine them into one.


jamessspanggg:
@HCY123902 please follow the commit message guidelines located [here](https://se-education.org/guides/guidelines/PRs.html). In particular, wrap the commit message body within two ```.

fzdy1914:
>  In particular, wrap the commit message body within two ```.

@jamessspanggg Actually, it is OK not to wrap the commit message body with two ```, as long as they are easily distinguishable. Now, we use "`" for variable highlighting.

HCY123902:
Thanks for the reminder. I will change it according to the guideline.

src/main/java/reposense/system/StreamGobbler.java
```diff
@@ -24,6 +24,11 @@ public String getValue() {
         return value;
     }
 
+    /**
+     * Reads from input stream {@code is} and stores it into {@code value}.
+     * This method will attempt to iteratively read 8KB of data from {@code is}, until no
```
jamessspanggg, line 29: I think its not needed to state the second point `This method will attempt...`

HCY123902: Sure, I have removed it. It does seem to be a bit redundant.

jamessspanggg (approved):

PR #1368, opened 2020-12-11
[#1233] Use a more specific regex for URL validation

### LINKED ISSUE #1233: RepoLocation: URL validation can be simplified by using a more specific regex
Currently, we instantiate a `URL` object to check if the repo url is a valid URL, as seen in the snippet below:

https://github.com/reposense/RepoSense/blob/0f57aab1deb97699e3b8d59ddca0ef492bbdc89d/src/main/java/reposense/model/RepoLocation.java#L69-L74

However, this can also be done by using a stronger regex.
Currently, our regex is: `^.*github.com\\/(?<org>.+?)\\/(?<repoName>.+?)\\.git$`

We can change it to:
`^(http|https)://github.com/(?<org>.+?)/(?<repoName>.+?)\\.git$`

Since we are using a regex anyway to parse the repo url, we can make it stronger such that the validation in line 69 - line 74 is no longer needed.

wltan:
Resolves #1233.

```
Within RepoLocation, a Git URL is checked using both regex and by
instantiating a java.net.URL object.

With a stronger regex, we can avoid having to instantiate a URL object.

Let's use the stronger regex to replace creating a URL object when
checking a Git URL.
```

Notes:
- The original issue suggests a regex of `^(http|https)://github.com/(?<org>.+?)/(?<repoName>.+?)\\.git$`. There are some minor deviations from this.
    - The forward slashes `/` need to be escaped with `\\`, as done in the old regex.
    - `(http|https)` has been replaced with a more elegant `https?` pattern
    - Thus, the final regex to be used (as shown in the code) is `^https?:\\/\\/github.com\\/(?<org>.+?)\\/(?<repoName>.+?)\\.git$`
- The validation logic within `verifyLocation` ("line 69 - line 74") has been modified to use the regex pattern matching.
    - It cannot be entirely removed as `verifyLocation` must still check that the location is either a valid path or valid URL.
    - We have however eliminated the need to instantiate a `java.net.URL` object, which fits the spirit of the original issue.
- There is a slight duplication of code as a `Matcher` is created once in the `RepoLocation` constructor and once again in `verifyLocation`.
    - It should be possible to eliminate this duplication, but I would appreciate any guidance on how to best do so while maintaining readability (e.g. passing a `Matcher` object to `verifyLocation` might not be very intuitive to the reader).
- The initialization here is now redundant. We could either move the declaration down, or move the logic up to the declaration.
https://github.com/reposense/RepoSense/blob/60de6237039041e7b2dfe9af208d3775fa61fb51/src/main/java/reposense/model/RepoLocation.java#L60

### DIFF
--- a/src/main/java/reposense/model/RepoLocation.java
+++ b/src/main/java/reposense/model/RepoLocation.java
@@ -1,7 +1,5 @@
 package reposense.model;
 
-import java.net.MalformedURLException;
-import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
@@ -17,7 +15,7 @@ import reposense.parser.InvalidLocationException;
 public class RepoLocation {
     private static final String GIT_LINK_SUFFIX = ".git";
     private static final Pattern GIT_REPOSITORY_LOCATION_PATTERN =
-            Pattern.compile("^.*github.com\\/(?<org>.+?)\\/(?<repoName>.+?)\\.git$");
+            Pattern.compile("^https?:\\/\\/github.com\\/(?<org>.+?)\\/(?<repoName>.+?)\\.git$");
 
     private final String location;
     private final String repoName;
@@ -66,12 +64,8 @@ public class RepoLocation {
             // Ignore exception
         }
 
-        try {
-            new URL(location);
-            isValidGitUrl = location.endsWith(GIT_LINK_SUFFIX);
-        } catch (MalformedURLException mue) {
-            // Ignore exception
-        }
+        Matcher matcher = GIT_REPOSITORY_LOCATION_PATTERN.matcher(location);
+        isValidGitUrl = matcher.matches();
 
         if (!isValidPathLocation && !isValidGitUrl) {
             throw new InvalidLocationException(location + " is an invalid location.");


fzdy1914 (changes requested):
Is it possible to propose a commit message in your PR as per our [contributing guidelines](https://reposense.org/RepoSense/dg/workflow.html)?

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -17,7 +15,7 @@
 public class RepoLocation {
     private static final String GIT_LINK_SUFFIX = ".git";
```
fzdy1914, line 16: Is this attribute unused after the change?

wltan: There is still one usage here, which I have not touched.

https://github.com/reposense/RepoSense/blob/60de6237039041e7b2dfe9af208d3775fa61fb51/src/main/java/reposense/model/RepoLocation.java#L38

fzdy1914: OK

fzdy1914:
@jamessspanggg Can u give some advice on the proposed commit message?

wltan:
> Is it possible to propose a commit message in your PR as per our [contributing guidelines](https://reposense.org/RepoSense/dg/workflow.html)?

Added to the top comment.

I wasn't actually aware that we had to put proposed commit messages in the PR description. Perhaps that could be made clearer/more explicit in the DG for future contributors? I couldn't find anything specifically about this, either in the linked contributing guidelines or in the [SE-EDU guide](https://se-education.org/guides/guidelines/PRs.html).

fzdy1914:
> > Is it possible to propose a commit message in your PR as per our [contributing guidelines](https://reposense.org/RepoSense/dg/workflow.html)?
> 
> Added to the top comment.
> 
> I wasn't actually aware that we had to put proposed commit messages in the PR description. Perhaps that could be made clearer/more explicit in the DG for future contributors? I couldn't find anything specifically about this, either in the linked contributing guidelines or in the [SE-EDU guide](https://se-education.org/guides/guidelines/PRs.html).

Indeed, we are using a different merge method comparing to SE-EDU. For seedu, the commit message should be in each commit. But as we are using squash merging, the commit message should be better put at the top comment for us to edit. You may raise an issue for not clearly pointing out this in DG.

jamessspanggg:
@wltan remember to wrap your commit message body to be within 72 characters. Also there's no need of the `Currently`, as stated by the commit message guidelines [here](https://se-education.org/guides/guidelines/PRs.html).

wltan:
> @wltan remember to wrap your commit message body to be within 72 characters. Also there's no need of the `Currently`, as stated by the commit message guidelines [here](https://se-education.org/guides/guidelines/PRs.html).

Ok, I've edited the commit message above.

jamessspanggg (approved):

fzdy1914:
> > @wltan remember to wrap your commit message body to be within 72 characters. Also there's no need of the `Currently`, as stated by the commit message guidelines [here](https://se-education.org/guides/guidelines/PRs.html).
> 
> Ok, I've edited the commit message above. It should fit within 72 characters now.

`With a stronger regex, we can avoid having to instantiate a URL object.` This line is more than 72 characters. If you are using SourceTree, there is a build-in indicator to turn on in the setting so you can have a reference.

fzdy1914 (approved):

PR #1371, opened 2020-12-14
[#1356] Not use # in URLs 

### LINKED ISSUE #1356: Not use # in URLs
Looks like the # in the RepoSense URLs are causing some problems. Is it possible not to use # in URLs?

https://github.com/nus-cs2103-AY2021S1/forum/issues/404

gerhean:
Fixes #1356 

**Current Situation**

Parameters were stored in `window.location.hash` which automatically adds the `#` symbol to the url. (Undesirable due to issues with copying non-alphabetical characters in some pdf applications).

image

**Outline of Solution**

Use `window.history.replaceState()` to edit the url directly to store parameters.

URL uses `?` to separate the path and parameters, and the symbol used can be easily modified in the future.

image

**Possible Alternative**

Using `window.history.pushState()` will instead leave the old url in the history (so one can navigate back to the previous url by using the back button on the browser. This does not rerender the page however.)

**Proposed commit message:**

```
Parameters were stored in `window.location.hash` which automatically adds the `#` symbol to the url.

This leads to issues with copying non-alphabetical characters in some pdf applications.

Let's replace # with ? in the URL, and additionally make the symbol easy to change in the future.
```

### DIFF
--- a/frontend/src/static/js/api.js
+++ b/frontend/src/static/js/api.js
@@ -63,10 +63,10 @@ window.encodeHash = function encodeHash() {
 window.decodeHash = function decodeHash() {
   const hashParams = {};
 
-  const hash_index = window.location.href.indexOf(HASH_ANCHOR);
-  const parameter_string = hash_index == -1 ? '' : window.location.href.slice(hash_index + 1);
+  const hashIndex = window.location.href.indexOf(HASH_ANCHOR);
+  const parameterString = hashIndex === -1 ? '' : window.location.href.slice(hashIndex + 1);
 
-  parameter_string.split('&')
+  parameterString.split('&')
       .forEach((param) => {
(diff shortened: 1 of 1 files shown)


fzdy1914 (approved):
Codewise, LGTM.
 
> Use `window.history.pushState()` will instead leave the old url in the history (so one can navigate back to the previous url by using the back button on the browser. This does not rerender the page however.)

Yep, it is an inherent problem of Vue. Using `replaceState()` is fine for me. 

Do remember to edit your PR title and description according to our DG.

jamessspanggg (approved):

fzdy1914:
Do check [https://se-education.org/guides/guidelines/PRs.html#submitting-a-pr] for you commit body.

gerhean:
I have proposed a commit message.

Issue #746, opened 2019-06-13
CodeView: loading screen only shows when first time selecting to view code


jamessspanggg:
The loading screen only appears the first time selecting to view the code. 
After that, when users selects another code, they may not know that the code is loading as the loading screen does not appear.

![loading-screen](https://user-images.githubusercontent.com/32864116/59408797-6308e580-8de7-11e9-96bd-6f41aed64230.gif)

fzdy1914:
Wait for #967 to be merged.

damithc:
Seems to be fixed now.

PR #1372, opened 2020-12-14
[#746, #1351] Dashboard: show a spinner when the view is changing

### LINKED ISSUE #1351: Dashboard: show a spinner when the view is changing
For reports with large amount of data, the view can take some time to change.  Take https://nus-cs2103-ay2021s1.github.io/tp-dashboard/#breakdown=true for example. It takes about 5 seconds to change the view from group-by-repo to group-by-author. 

It would be good to show some sort of activity indicator (e.g., a spinner) during that period.

gerhean:
Fixes #1351, #746

Probably fixed, but unsure how to test as need large amount of data for the loading to actually be seen.

**Edit:**

Before:
image
Screen frozen and unresponsive.

After:
image

Example of loading spinner (will ensure link works until PR is merged): https://gerhean.github.io/tp-dashboard/

setTimeout() is used to force this.filtered to update only after browser rerender (loading screen is displayed).

Something to note is that the animated loading indicator uses css to be animated, so it spins even if the website is actually not working. It is basically just eye candy.

**Proposed commit message:**

```
For reports with large amount of data, the view can take some time to change.

Looking at a frozen browser tab causes anxiety and might lead one to believe that the tab has stopped working, while in reality it is simply doing background processing.

Let's add an animated loading indicator.

A portion of filterGroupSelection() has been abstracted out for better readability as it had to be wrapped in a setTimeout()

v-bind:key for v-authorship and v-zoom tabs in index.pug has been changed to prevent unnecessary rerendering.
```


fzdy1914:
The data can be found [here](https://github.com/nus-cs2103-AY2021S1/tp-dashboard/tree/gh-pages). Since your PR does not edit the file structure. You can just use the data directly. So you may need to clone this `tp-dashboard`, switch to github page branch, replace the HTML and attached the resultant Github Page link here for us to check your effort. 

An even better choice is that, we have a [repo](https://github.com/reposense/publish-RepoSense) for generating reposense using travis or github actions. By modifying the `get-reposense.py`, you can make it to download the version of your reposense and do the report generation. So you then can automate the process.

The second part can leave as a bonus part, that you submit a [PR](https://github.com/reposense/publish-RepoSense/blob/master/get-reposense.py) for our publish RepoSense so it can support `--branch` tag.

fzdy1914:
Meanwhile, do remember to follow the development guide in DG. Including the PR title and description.

gerhean:
Perhaps there should be an article in the DG about how to test frontend without repeatedly generating the data as I am still not entirely sure of the instructions.

fzdy1914:
> Perhaps there should be an article in the DG about how to test frontend without repeatedly generating the data as I am still not entirely sure of the instructions.

This is indeed in our plan. But currently, we have to achieve it using a workaround. You may just terminate the `gradlew run` process just at the time the console shows that it starts to analyzing the repos. Then, at that timestamp, the html is replaced with new one but the data is still the old one. That is the trick I use when I am a student. But you may also try `gradlew zipReport` command to see if it does the same job.

fzdy1914:
@damithc  Prof, can you help to check the UI? I feel OK, but we may add some text for it, like: `Loading Data` or `Processing Data` comparing with the `loading resources` shown in the first place.

frontend/src/static/js/main.js
```diff
@@ -12,6 +12,7 @@ Vue.directive('hljs', {
 });
 
 Vue.component('font-awesome-icon', window['vue-fontawesome'].FontAwesomeIcon);
+Vue.component('loading', window.VueLoading);
```
fzdy1914, line 15: We can have a better component name, like `loading-overlay`?

frontend/src/static/js/main.js
```diff
@@ -21,6 +22,7 @@ window.app = new window.Vue({
     users: [],
     userUpdated: false,
 
+    isLoadingOverlayEnabled: false,
     isLoading: false,
```
fzdy1914, line 26: Change the name to `isLoadingResources` two distinguish these two attribute.

frontend/src/static/js/v_summary.js
```diff
@@ -364,6 +361,20 @@ window.vSummary = {
       this.filtered = this.sortFiltered(this.filtered, filterControl);
     },
 
+    getMergedGroup(allGroupsMerged) {
```
fzdy1914, line 364: The name should better be `updateMergedGroup`?

fzdy1914 (changes requested):

fzdy1914:
Meanwhile, there is latency time when `merge all groups` button is clicked.

damithc:
> @damithc Prof, can you help to check the UI? I feel OK, but we may add some text for it, like: `Loading Data` or `Processing Data` comparing with the `loading resources` shown in the first place.

Mostly fine :+1:

Shall we use the same spinner used at the very start? No reason to use two different ones.

Yes, we can show something like `Working. Please wait ...` below the spinner.

fzdy1914:
@gerhean As pointed out by the Prof, let's use the new loading spinner for both resource loading and data processing. Meanwhile, please add some text below the spinner to notify the user of the different loading stages. 

Finally, please look into the loading of `v_authorship`.  If u found it easy, u may add the same loading overlay for the loading stage of  `v_authorship`

gerhean:
I am having a problem in which v-authorship is being created twice and destroyed once on startup, which causes `removeAuthorshipHashes()` to be called. This is causing problems in testing but i'm not really sure why it's happening.

gerhean:
May I ask why a key needs to be specified here:

image

I want to remove it as it is causing the tab to rerender multiple times (as the loading screen requires some data to be updated asynchronously)

frontend/src/static/js/main.js
```diff
@@ -21,7 +22,8 @@ window.app = new window.Vue({
     users: [],
     userUpdated: false,
 
-    isLoading: false,
+    isLoadingOverlayEnabled: false,
```
fzdy1914, line 25: Seems that the implementation can be simplified a bit? We are already using the same spinner, why we cannot only have one attribute for its state and another attribute for the text below it? Will it be more elegant in the implementation? And we can extend it to any scenario that needs a loading overlay?

frontend/src/static/js/v_authorship.js
```diff
@@ -298,6 +312,18 @@ window.vAuthorship = {
       window.encodeHash();
     },
 
+    updateSelectedFiles() {
+      this.$store.commit('updateIsLoadingOverlayEnabled', true);
+      const selectedFiles = this.files.filter(
```
fzdy1914, line 317: Shouldn't we put the whole computation inside `setTimeOut` so that the loading overlay will show up before the actual computation?

frontend/src/static/js/v_authorship.js
```diff
@@ -388,6 +408,10 @@ window.vAuthorship = {
     this.removeAuthorshipHashes();
   },
 
+  updated() {
+    this.$store.commit('updateIsLoadingOverlayEnabled', false);
```
fzdy1914, line 412: If we decided to make the attribute global, we may not able to make use of the `updated` life hook right? This may accidentally terminate a loading overlay of another component?

fzdy1914 (changes requested):
@gerhean Good observation of the issue! Actually, there is an attempt to fix it but fails. You may refer to #967 and #1167 for more details. Your proposed implementation is quite nice and can fix it as a workaround. However, I believe the ideal solution is to make use of the re-rendering of  Vue directly like what is done in #967 but fix the bug caused by it.

@gerhean Do request a review from me once the PR is finished. Good Work done also, I have recommended you to the Prof.

@jamessspanggg Can you check the implementation also to make sure that the attributes listed are the key components to re-render the sidebar?

frontend/src/static/js/main.js
gerhean: Done

frontend/src/static/js/v_authorship.js
gerhean: Changed loading overlay implementation

frontend/src/static/js/v_authorship.js
gerhean: Done

gerhean:
@fzdy1914  About the issue regarding the tab rerendering, I was wondering why `this.$store.state.tabAuthorshipInfo` is passed to `main.js` before being passed to `v_authorship.js` as props, when `v_authorship.js` could just get it direct from the store.

fzdy1914:
> @fzdy1914 About the issue regarding the tab rerendering, I was wondering why `this.$store.state.tabAuthorshipInfo` is passed to `main.js` before being passed to `v_authorship.js` as props, when `v_authorship.js` could just get it direct from the store.

Do you mean the usage here? 
image

I agree that in the ideal design, the main.js can use the info in the store to generate the key instead of keeping a local copy. But we are using the current workflow as it is inherited from the old design. In our old design, the sidebar will be completely destroyed and re-rendered when updating, so we designed that we use a `prop` to control the sidebar because at that time, we haven't implemented the Vue store.

Now, you may try to see if it is possible to use the info in the store to generate the key instead of keeping a local copy and get the info from Vue store directly for sidebar components. So each time we update, we destroy old component by the key, and the new component is generated based on the info in store instead of a `prop` passed in.

And you may open a new PR for it to see if it works or not. This kind of PR needs to be very careful as it is touching the root design of our project.

fzdy1914 (changes requested):
@damithc Can you check the new loading overlay for recourses, summary and authorship?

@gerhean Cau also fixes #746 in this PR by removing all old loading components? We only need one loading overlay to show that we are loading.

damithc:
> @damithc Can you check the new loading overlay for recourses, summary and authorship?

Seems there is a significant delay (2-5 seconds) between clicking on the `</>` icon and the spinner showing up. Is that something we can eliminate? Can do as a separate PR though.

I wonder if there is any value in using different phrases for different places. A generic `Working ...`  (or `Loading ...`) is enough? Otherwise we end up having to figure out the right term, exposing technical terms to the user, typos, upper/lower case issues etc. with no gain in return. Anyway, not a major problem.

Noticed this at some times. Are we able to replace that with a spinner as well (for consistency sake)?
image

fzdy1914:
> Seems there is a significant delay (2-5 seconds) between clicking on the `</>` icon and the spinner showing up. Is that something we can eliminate? Can do as a separate PR though.

The delay also occurs when clicking the merge group button.

> I wonder if there is any value in using different phrases for different places. A generic `Working ...` (or `Loading ...`) is enough? Otherwise we end up having to figure out the right term, exposing technical terms to the user, typos, upper/lower case issues etc. with no gain in return. Anyway, not a major problem.

OK, let's do the general `Loading...` then. Sorry for the inconvenience caused. @gerhean 


> Noticed this at some times. Are we able to replace that with a spinner as well (for consistency sake)?
> image

This is what I request in previous comment: 
> @gerhean Cau also fixes #746 in this PR by removing all old loading components? We only need one loading overlay to show that we are loading.

frontend/src/static/js/v_authorship.js
```diff
@@ -316,7 +314,7 @@ window.vAuthorship = {
 
     updateSelectedFiles() {
       this.$store.commit('incrementLoadingOverlayCount', 1);
-      this.$store.commit('updateLoadingOverlayMessage', renderAuthorshipMessage);
```
fzdy1914: So we don't need `updateLoadingOverlayMessage` attribute right?

gerhean: I think it can be useful if you ever need to change the message. I can add a generic message used as fallback when no message is specified though.

gerhean:
> The delay also occurs when clicking the merge group button.

Something I noticed is that clicking this causes a lot of warnings on the console so perhaps that should also be fixed?

> Seems there is a significant delay (2-5 seconds) between clicking on the </> icon and the spinner showing up. Is that something we can eliminate? Can do as a separate PR though.

The delay actually occurs when clicking anything.

I think a 2 second delay is still acceptable for users. The browser itself needs time to render the loading screen so I'm not actually sure if this delay can be reduced.

frontend/src/static/js/main.js
```diff
@@ -40,6 +46,12 @@ window.app = new window.Vue({
       this.tabInfo.tabAuthorship = Object.assign({}, this.$store.state.tabAuthorshipInfo);
       this.activateTab('authorship');
     },
+    '$store.state.loadingOverlayCount': function () {
+      this.isLoadingOverlayEnabled = this.$store.state.loadingOverlayCount > 0;
+    },
+    '$store.state.loadingOverlayMessage': function () {
+      this.loadingOverlayMessage = this.$store.state.loadingOverlayMessage;
```
fzdy1914, line 53: This one can be simplified with a computed value so we do not need to keep the local copy.  See how other components use `mapState` for vuex values.

frontend/src/static/js/v_authorship.js
```diff
@@ -16,6 +16,7 @@ window.vAuthorship = {
     return {
       isLoaded: false,
```
fzdy1914, line 17: This attribute should be also removed right?

gerhean: Actually I realized that without this, `nothing to see here :(` might be wrongly displayed due to async loading of data. Unless having it displayed for a few seconds while data is loading doesn't matter then I suppose it is fine to remove it.

gerhean:
> Good observation of the issue! Actually, there is an attempt to fix it but fails. You may refer to #967 and #1167 for more details. Your proposed implementation is quite nice and can fix it as a workaround. However, I believe the ideal solution is to make use of the re-rendering of Vue directly like what is done in #967 but fix the bug caused by it.

Tried to fix it in another branch, will PR once this PR is merged.

fzdy1914:
> > Seems there is a significant delay (2-5 seconds) between clicking on the </> icon and the spinner showing up. Is that something we can eliminate? Can do as a separate PR though.
> 
> The delay actually occurs when clicking anything.
> 
> I think a 2 second delay is still acceptable for users. The browser itself needs time to render the loading screen so I'm not actually sure if this delay can be reduced.

I think to overcome this issue, we need to put the handling of loading in two different components. So, before the new info is pushed to the store, we trigger the overlay first. And, after the authorship finishes loading, we finish the overlay. Putting the loading command in two different places may cause some difficulties in understanding the whole flow and future development.

@gerhean Let's make it an issue to record our discussion and to see if there are better choices to fix that. Actually, I have some idea now, but, it is out of the scope of this PR.

frontend/src/static/js/v_authorship.js
fzdy1914: Is it because you forgot to remove the `v-if` here:
image

If what you mention is because moving this `if`, to fix that, replace it with a mapped computed value from vue store.

frontend/src/index.pug
```diff
@@ -34,6 +34,14 @@ html
     script(src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.14/build/languages/yaml.min.js")
   body
     #app
+      loading-overlay.overlay-loader(
+        :active.sync="isLoadingOverlayEnabled",
```
jamessspanggg, line 38: should we stick to the same `v-bind` format as the rest instead of the shorthand for consistencies' sake?

jamessspanggg (commented):
> @jamessspanggg Can you check the implementation also to make sure that the attributes listed are the key components to re-render the sidebar?

Yep they are they key attributes to re-render the side panel.

> I think a 2 second delay is still acceptable for users. The browser itself needs time to render the loading screen so I'm not actually sure if this delay can be reduced.

I think it is acceptable for now, perhaps we can dive in deeper in future PRs to figure out how to reduce the delay.

frontend/src/static/js/v_authorship.js
gerhean: image

This is the async function which is used to load the files (from JSON). Because of how Vue works the browser will be rendered before this function ends. Since the `files` array depends on waiting for this function there isn't any computed value from vue store to map from until the function finishes.

frontend/src/index.pug
fzdy1914: Yep, better to be consistent.

frontend/src/static/js/v_authorship.js
fzdy1914: OK, then, we can just keep this attribute

fzdy1914 (approved):

jamessspanggg (approved):
Good work!

frontend/src/index.pug
```diff
@@ -34,6 +34,14 @@ html
     script(src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.14/build/languages/yaml.min.js")
   body
     #app
+      loading-overlay.overlay-loader(
```
fzdy1914, line 37: Can you try to wrap it with a template?  Otherwise, an unpleasant message will show when the network connection is slow.

image

fzdy1914 (changes requested):

Tejas2805:
I think it might be better to say - **Loading. Please wait.** - instead of - **Waiting. Please wait.**

Tejas2805 (changes requested):
Regression bug introduced - Sorting in Code Panel is not working. Sorting is still working in the Commit Panel.

frontend/src/index.pug
gerhean: Fixed with v-cloak

gerhean:
@Tejas2805 

> I think it might be better to say - Loading. Please wait. - instead of - Working. Please wait.`
Done

> Regression bug introduced - Sorting in Code Panel is not working. Sorting is still working in the Commit Panel.
Thanks for telling me. Fixed.

frontend/src/index.pug
```diff
@@ -34,6 +34,14 @@ html
...
+        v-bind:active.sync="isLoadingOverlayEnabled",
+        v-bind:opacity='loadingOverlayOpacity',
+        v-bind:is-full-page="true"
```
fzdy1914, line 40: Shouldn't v-cloak better to be added here?

fzdy1914: Meanwhile, the travis is failing.

gerhean: fixed

fzdy1914 (approved):

fzdy1914:
@Tejas2805, your review is needed.

Tejas2805:
In some cases, the part displayed in the image below shows up first. And then the loading spinner. In some cases, both come together. So if that is not much of an issue, LGTM. @damithc @fzdy1914 


https://user-images.githubusercontent.com/35946746/103080509-f80db100-4610-11eb-8763-a6840272c6d7.mp4

fzdy1914:
> In some cases, the part displayed in the image below shows up first. And then the loading spinner. In some cases, both come together. So if that is not much of an issue, LGTM. @damithc @fzdy1914

That is indeed what will happen in the current implementation.  Changing the implementation will be done in a separate PR.

damithc:
Memory usage on Edge:
* https://gerhean.github.io/tp-dashboard/  800-900Mb
* https://nus-cs2103-ay2021s1.github.io/tp-dashboard/ (generated on Dec 4) 600-700Mb

Did we do something since Dec 4 to degrade performance?

fzdy1914:
> Did we do something since Dec 4 to degrade performance?

Don't seem to have done anything that may downgrade the performance.

@gerhean Can you help to check this issue?

damithc:
> > Did we do something since Dec 4 to degrade performance?
> 
> Don't seem to have done anything that may downgrade the performance.
> 
> @gerhean Can you help to check this issue?

I checked again now. Seems the memory usage is similar for both cases. Goes to 1Gb at the start but comes down to 600Mb after some time. So, should be fine. OK to go ahead.

fzdy1914:
@Tejas2805 Can you review again?

Tejas2805:
Just reviewed again. I thought we had fixed issue #1081 in PR #1089 but I can see the bug still exists. Has it been re-introduced in this PR?


https://user-images.githubusercontent.com/35946746/103237109-a737e880-4981-11eb-8d34-7e7a928f795f.mov

The `frontent` files should still be hidden and not be shown as being seen in the video.

fzdy1914:
> Just reviewed again. I thought we had fixed issue #1081 in PR #1089 but I can see the bug still exists. Has it been re-introduced in this PR?

That PR is not even merged. So, it is not fixed.

Tejas2805:
> That PR is not even merged. So, it is not fixed.

The issue was closed so thought that it had been fixed. Just checked it was closed by PR #1065

Tejas2805 (approved):
LGTM!

PR #1373, opened 2020-12-14
[#1370] Use light theme for code

### LINKED ISSUE #1370: Use light theme for code
RepoSense docs were written when MarkBind supported only the light theme for code. Recently, MarkBind switched to the dark theme as the default. However, our documentation doesn't look nice in the dark theme as there are many inline code fragments in the docs. An example is given below.
Perhaps switch back to light theme? Instructions for adding the setting to the `site.json` are [here](https://markbind.org/userGuide/siteJsonFile.html#style).

image

kkangs0226:
Fixes #1370 

The code blocks in RepoSense documents have changed to dark theme after the recent MarkBind update. Changing the theme of code blocks back to light theme would improve the readability of the documents. 

Before:
image

After:
image

Proposed commit message:
```
Current code blocks are in dark theme after latest MarkBind
update.

Changing back to light theme would improve readability of code
blocks.

Let's update site.json file to change back to light theme.
```


jamessspanggg (approved):
LGTM

damithc (approved):

PR #1386, opened 2020-12-21
[#1227] Add minify option to reduce json files size in report generator

### LINKED ISSUE #1227: CLI: add minify option to reduce json files size
Spaces and indentation take up a huge amount of space in the file size.

HCY123902:
Fixes #1227
Commit message:
```
Make pretty printing disabled by default to reduce json file
size generated.

The pretty printing option of GsonBuilder seems to make
json file generated take more space as is mentioned in 
issue #1227 

This can cause issues regarding space consumption
when the report contains large amount of information from
the repository.

Let's make pretty printing disabled by default to make 
enhancement to the space consumption.
```


fzdy1914:
@damithc  Hi Prof, I am wondering if we truly need a flag for this? I recall that you may need to analyze the generated JSON file right? Can your parser correctly parse JSON not in pretty-printing mode? If your parser can recognize the minified version, maybe we can just set the default to be minified so we do not need to have a flag for it?

damithc:
Yes, no need for the flag. We can minify by default.

fzdy1914 (changes requested):
> Yes, no need for the flag. We can minify by default.

As mentioned by Prof, we may not need to have a flag for it. 

Can you make the default mode to be minified instead and focus more on the system test, and how our frontend will handle the minified JSON file?

HCY123902:
Sure, no problem

fzdy1914 (approved):

fzdy1914:
@HCY123902 Do say fixes #1227 so this PR will close the issue once get merged.

jamessspanggg (approved):

Tejas2805:
How about adding a flag to let the user get pretty printing if he wants to? The default will still be minified. @damithc @fzdy1914.

Otherwise LGTM!

damithc:
> How about adding a flag to let the user get pretty printing if he wants to? The default will still be minified. @damithc @fzdy1914.

That idea occurred to me earlier too. But then, I remembered that the user is not even supposed to know about that json file. Its format it also not optimized for human consumption anyway. I guess anyone willing to tinker an internal json file will know how to make it pretty.

Tejas2805 (approved):

PR #1390, opened 2020-12-31
[#1382] Avoid rerendering tabs unnecessarily

### LINKED ISSUE #1382: Avoid rerendering tabs unnecessarily.

Currently, tab info is passed from `store.js` to `main.js` to `v-authorship.js` and `v-zoom.js`.
**Suggestion:** to pass info directly from `store.js` to `v-authorship.js` and `v-zoom.js`. 

Each time tab information is updated, we destroy old component by the key, and the new component is generated based on the info in store instead of a prop passed in.
**Suggestion:** Remove the key and watch for info changes directly from the tabs.

 Might be related to #967, #1167 and #967

From PR #1372:

image

> 
> I agree that in the ideal design, the main.js can use the info in the store to generate the key instead of keeping a local copy. But we are using the current workflow as it is inherited from the old design. In our old design, the sidebar will be completely destroyed and re-rendered when updating, so we designed that we use a prop to control the sidebar because at that time, we haven't implemented the Vue store.
> 
> Now, you may try to see if it is possible to use the info in the store to generate the key instead of keeping a local copy and get the info from Vue store directly for sidebar components. So each time we update, we destroy old component by the key, and the new component is generated based on the info in store instead of a prop passed in.
> 
> And you may open a new PR for it to see if it works or not. This kind of PR needs to be very careful as it is touching the root design of our project.

Watch for specific changes in `info` using computed string as a key:

```
  watch: {
    authorshipKey() {
      this.initiate();
    },
```

```
 computed: {
    authorshipKey() {
      return `${this.info.author}|${this.info.repo}|${this.info.isMergeGroup}`;
    },
```

Calling `this.initiate();` is less destructive than rerendering the entire tab.

Not entirely sure which properties are important to watch however.

Proposed solution:
https://github.com/gerhean/RepoSense/tree/tabs-get-props-from-store

Will PR once PR #1372 is merged.

gerhean:
Fixes #1382

Should have no change from user perspective. Built on #1351.

Example (will try to keep this updated): https://gerhean.github.io/tp-dashboard/

Proposed commit message:

```
Currently tabs are destroyed and rerendered based on a key which can lead to mysterious bugs. Furthermore tab information is passed indirectly from store to main.js before being passed to the tab itself.

Let's clean up the code such that tabs do not need to completely rerender when information changes. Furthermore remove indirect passing of information.
```


gerhean:
image
These 2 variables seem to be the same thing should I remove 1?

0blivious:
Unwanted change introduced by this PR.

1. Go to report generated, open any authorship view in Group 1, notice that all file types are checked
image

2.  Open any authorship view in Group 2, by default, are are unchecked
image

0blivious:
Also, when merge group, line count in authorship view is not handled correctly.
image

frontend/src/static/js/main.js
```diff
@@ -104,8 +101,6 @@ window.app = new window.Vue({
 
     // handle opening of sidebar //
     activateTab(tabName) {
-      // changing isTabActive to trigger redrawing of component
-      this.isTabActive = false;
```
fzdy1914, line 108: This one may still be essential to make everything work. A simple workaround to destroy all the component attribute that we want to renew when re-rendering.

gerhean: Oh the tab doesn't rerender anymore in this PR. I'll try removing it and see if it causes any bugs.

frontend/src/static/js/v_zoom.js
```diff
@@ -1,17 +1,21 @@
 /* global Vuex */
 
+function initialState() {
```
fzdy1914, line 3: Can you provide some link/document for the rationale to convince us that it is a more elegant way of `resetting some default attribute` comparing with our previous methods?

fzdy1914: It is the first time for me to see such implementation.

frontend/src/static/js/main.js
fzdy1914: I got the logic as I tried to do so once. The new way `initialState()` seems to be working.

frontend/src/static/js/v_zoom.js
gerhean: Oh is there a previous method you used? I just copied this off Stackoverflow as a nice way to avoid rewriting the entire initial state since I need to reset to the initial state in some method.

frontend/src/static/js/main.js
gerhean: Actually I removed isCollapsed because it wasn't used anywhere after rewriting some stuff. I think it still works.

gerhean:
The unwanted change introduced by this PR pointed out by @0blivious  should be fixed.

Some things to test is:
Refreshing page works.
Switching to another author/repo's authorship/zoom tab does not retain anything from a previously visited authorship/zoom tab.

So far I haven't had any bugs with these but it may be good to make sure.

frontend/src/static/js/v_zoom.js
fzdy1914: Got it

fzdy1914:
> The unwanted change introduced by this PR pointed out by @0blivious should be fixed.
> 
> Some things to test is:
> Refreshing page works.
> Switching to another author/repo's authorship/zoom tab does not retain anything from a previously visited authorship/zoom tab.
> 
> So far I haven't had any bugs with these but it may be good to make sure.

By this chance, can you add the corresponding cypress test of the two points you have mentioned?

0blivious:
Seems your tab is rendering twice
ezgif com-gif-maker

gerhean:
> Seems your tab is rendering twice
It renders twice in the original too, but I made some changes so that the `nothing to see here :(` isn't displayed anymore.

> By this chance, can you add the corresponding cypress test of the two points you have mentioned?
Maybe in another PR.

fzdy1914:
> Maybe in another PR, I'm asking someone else to help me with this.

OK.

fzdy1914:
Found one regression:

steps, 
1. open preview
2. view the authorship for `Eugene(1.1)` 
3. unclick file type `frontend`
4. view the authorship for `A.M.S.Rathod(3.1)`

expect result: 
image

now:
image

gerhean:
Working on it

fzdy1914:
@reposense/21-22-active-developer Pls help review this PR and check if there are any other regressions. `Approve this PR` after your review is done.

frontend/src/static/js/v_authorship.js
```diff
@@ -6,26 +6,29 @@ const filesSortDict = {
...
+    selectedFileTypes: [],
+    fileTypes: [],
+    filesLinesObj: {},
+    fileTypeBlankLinesObj: {},
+    filesSortType: 'lineOfCode',
+    toReverseSortFiles: true,
+    searchBarValue: '',
+  };
+}
+
```
dcshzj, line 24: Should this be an object instead of a function?

gerhean: It should be a function as I need to reset the state of the component while it is still alive by creating a new object, as global objects persist.

frontend/src/static/js/v_authorship.js
```diff
@@ -108,7 +114,7 @@ window.vAuthorship = {
       window.encodeHash();
     },
 
-    initiate() {
+    async initiate() {
```
HCY123902, line 117: Actually, is it necessary to make this asynchronous here?

gerhean: No, but the alternative is `.then()` chaining is something I avoid for personal preference, as `loadAuthorship()` is an async function which I need to wait for.

gerhean: No, but the alternative is .then() chaining is something I avoid for personal preference, as loadAuthorship() is an async function which I need to wait for.

fzdy1914 (approved):
Codewise OK to me. Waiting for more regression tests

PR #1400, opened 2021-01-05
[#1358] Migrate tests from TravisCI to GitHub Actions

### LINKED ISSUE #1358: Find an alternative to TravisCI
TravisCI has [announced changes to its pricing model](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing) which will limit free and open source projects to a fixed amount of credits (currently 10,000), unless a higher and renewable limit is requested. We will be affected when we are forced to migrate to `travis-ci.com` at the end of the year.

We might have to look for an alternative for our CI needs, or request from Travis Support for a reasonable renewable limit per month.

dcshzj:
Fixes #1358.

```
TravisCI has moved towards a fixed quota to execute tests on their
website, which is not sustainable in the long term.

Let's migrate the testing infrastructure to GitHub Actions to take
advantage of their free unlimited minutes.
```

I have removed some code from `build.gradle` in checking the Git version as it was not working on GitHub Actions (they were returning `null`). I am not sure if there are any available alternatives?

There are also some settings that needs to be changed within this repository to enable some of the dependencies used.


Tejas2805:
Could you show the working of these actions in your forked repository or some other repository, before we review the code?

dcshzj:
For the continuous integration workflow, here is a sample run: https://github.com/dcshzj/RepoSense/actions/runs/463484966

For the GitHub Pages workflow, here is a sample run: https://github.com/dcshzj/RepoSense/actions/runs/463474225

build.gradle
```diff
@@ -173,9 +173,6 @@ task coverage(type: JacocoReport) {
 
 String getRepoSenseVersion() {
     String repoSenseVersion = project.property("version")
-    if (repoSenseVersion.equals("unspecified")) {
```
fzdy1914, line 176: This one is needed for our `surge` preview. Maybe, leave the surge deploy with travis first? And in the meantime, we look for an alternative for `id 'com.palantir.git-version' version '0.11.0'`? Anyway, the deployment won't take up so much time?

Tejas2805:
There's some issue with the action you are using. Maybe we need to give access or something in my opinion.

fzdy1914:
> There's some issue with the action you are using. Maybe we need to give access or something in my opinion.

Can I know what issue u are referring to? So I may do the fix?

dcshzj:
I'll try to see if we can have a workaround for the `com.palantir.git-version` plugin.

Also, @fzdy1914 I used the `cypress-io/github-action@v2` GitHub Action to handle the installation and setup of Cypress. I think it needs to be added to the whitelist of actions allowed under "Actions permissions". Thanks!

fzdy1914:
> I think it needs to be added to the whitelist of actions allowed under "Actions permissions". Thanks!

Done.
image

build.gradle
dcshzj: Upgraded this plugin to the latest version `0.12.3` seems to work now

fzdy1914:
@dcshzj Can you double-check for surge deployment? Seems it is a failing. We may need to specify some key attributes there.

dcshzj:
@fzdy1914 We will need to declare a `SURGE_LOGIN` secret with the email address of the surge.sh account, and a `SURGE_TOKEN` secret for the account's token.

I have also disabled the Travis CI check for this pull request, so the status from Travis will not be reported.

fzdy1914:
> @fzdy1914 We will need to declare a `SURGE_LOGIN` secret with the email address of the surge.sh account, and a `SURGE_TOKEN` secret for the account's token.
> 
> I have also disabled the Travis CI check for this pull request, so the status from Travis will not be reported. We will probably need to modify the required check from Travis to the GitHub Actions tests.

@jamessspanggg Can you provide some help here? As you have set up the surge?

jamessspanggg:
Previously there was an issue with running our tests and deploying our surge site using github actions, as external PRs will be unable to utilise our secrets:

> GitHub explicitly disables giving secrets to workflows that originate from a fork. That means that when a user opens a pull request on your project from a fork, no secrets are provided to this workflow.

They have since added a new feature where PR from forks are able to access the secrets, more details here: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

> @fzdy1914 We will need to declare a `SURGE_LOGIN` secret with the email address of the surge.sh account, and a `SURGE_TOKEN` secret for the account's token.

Yes we do need these secrets in order to deploy surge.

jamessspanggg:
@dcshzj I have added the `SURGE_LOGIN` and `SURGE_TOKEN` secrets

dcshzj:
> Previously there was an issue with running our tests and deploying our surge site using github actions, as external PRs will be unable to utilise our secrets:

Seems like a major blocker to having deployment to surge.sh via GitHub Actions. I don't think we want to build using the `pull_request_target`, as we want to actually build with the new code in the pull request.

Should we keep using Travis for deployment to surge.sh? This should save us in the short term while we look for alternatives, since Travis build minutes are non-renewable.

fzdy1914:
> Should we keep using Travis for deployment to surge.sh? This should save us in the short term while we look for alternatives, since Travis build minutes are non-renewable.

Let‘s go with this.

dcshzj:
I've moved the deployment to surge.sh back to Travis and the build took about 3 minutes. Once the new plan kicks in, we will be able to build about 333 times, which should last us for a couple of months.

I amended this pull request so that it does not close issue #1358, since we have not fully migrated.

damithc:
> I've moved the deployment to surge.sh back to Travis and the build took about 3 minutes. Once the new plan kicks in, we will be able to build about 333 times, which should last us for a couple of months.

Not sure if we can last that long. Once CS3281 students start pushing commits to their PRs, we might run out of build minutes quite fast :-)

fzdy1914:
@damithc 
Now, it seems that the issue is from `Github Action`. So there are two possible workarounds to continue work with `Github Action`.

1. We add the privilege of `push to other branches` for new devs (which they should already have). For the new workflow, new devs will need to
- Set up a local CI/Preview workflow on their own forked repo
- When making a new PR, they need to submit a PR to their own repo to preview their PR first
- When the PR is ready, push the new branch (with a proper name) to team repo and submit a PR from the team branch to team master
- The senior will review and go through the previous rounds again

2.  We just leak the secret as plain text in the repo because they may not be so important. So, leaking them is kind of acceptable as a workaround.

damithc:
Hmm... it's tricky.

> 1. We add the privilege of `push to other branches` for new devs (which they should already have). For the new workflow, new devs will need to

The workflow should work for any contributor (because anyone in the world should be able to contribute), not just those with write permissions.


> 2\. We just leak the secret as plain text in the repo because they may not be so important. So, leaking them is kind of acceptable as a workaround.

Doesn't sound very proper. What's the risk?

Is the problem caused by surge specifically?

dcshzj:
> Is the problem caused by surge specifically?

Surge requires authentication before allowing you to publish builds. This is done through secrets but unfortunately GitHub Actions do not allow secrets to be read for pull requests.

I was thinking if it is possible for us to use CircleCI for this instead? [They allow secrets to be passed to forked pull requests if explicitly enabled](https://circleci.com/docs/2.0/oss/#pass-secrets-to-builds-from-forked-pull-requests). They also provide about 250 build minutes per week, which works out to about 83 builds per week for our use case. We can also send them an email to be upgraded to their open source plan which gives about 40,000 build minutes per month.

If not, we can try to send an email to Travis CI to get some renewable build credits and continue using their platform.

damithc:
> They have since added a new feature where PR from forks are able to access the secrets, more details here: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/

@dcshzj confirmed the above doesn't help?

dcshzj:
> @dcshzj confirmed the above doesn't help?

No, that workflow will use code that is already in the repository. Instead, we want to build the site based on the code submitted in the PR.

damithc:
> No, that workflow will use code that is already in the repository. Instead, we want to build the site based on the code submitted in the PR.

If I understand correctly, it uses _only the workflow files_ from the base branch, the rest of the code from the PR code is used in the build. That is, it prevents the PR from doing malicious modifications to the workflow to hijack our secrets for their own purposes. Should work as long as the PR is not touching our workflow files?

dcshzj:
Hmm no, this event still uses the base branch's code and workflow files. From that blog post:

> However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request.

The code will still be based on the base branch, not from the code submitted in the PR. However, I [found a workaround to this problem](https://github.community/t/running-code-from-forks-with-pull-request-target/126688/6), let me try that and see if it works out.

dcshzj:
Whoops, I think we have to get the workflow file into the repository first before we can test out `pull_request_target`. Should I create another PR to test this?

fzdy1914:
> Whoops, I think we have to get the workflow file into the repository first before we can test out `pull_request_target`. Should I create another PR to test this?

You should have the privilege to create a branch in this Repo. Just do a PR against that branch to see if this workflow works?

fzdy1914:
@dcshzj I saw it is working. But can you check the warning below?
image

dcshzj:
> @dcshzj I saw it is working. But can you check the warning below?
> <snip>

Yep I'm very confused by this. The original code gives a Not Found error, even though the code is exactly the same as the one that used to run on Travis CI. I tried to see if `GITHUB_SHA` was set correctly (which it is) and I also tried to pass in a new environment variable that gives the HEAD commit for the PR, but the problem still persists.

I also checked that `GITHUB_TOKEN` has sufficient permissions to set the commit status, which makes it quite confusing as to what the error is.

fzdy1914:
> > @dcshzj I saw it is working. But can you check the warning below?
> > <snip>
> 
> Yep I'm very confused by this. The original code gives a Not Found error, even though the code is exactly the same as the one that used to run on Travis CI. I tried to see if `GITHUB_SHA` was set correctly (which it is) and I also tried to pass in a new environment variable that gives the HEAD commit for the PR, but the problem still persists.
> 
> I also checked that `GITHUB_TOKEN` has sufficient permissions to set the commit status, which makes it quite confusing as to what the error is. One possible way is to create yet another secret with someone's personal access token and to enable the `repo:status` scope, then using that token to update the commit status.

Can you try to print out the `curl` command that is actually running? And try to test the command locally? To see if it is working?

config/gh-actions/deploy-surge.sh
```diff
@@ -51,15 +60,15 @@ do
...
   then
     # Create github statuses that redirects users to the deployed dashboard and markbind docs
-    curl "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/statuses/${TRAVIS_PULL_REQUEST_SHA}?access_token=${GITHUB_API_TOKEN}" \
+    curl "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/statuses/${ACTIONS_PULL_REQUEST_HEAD}?access_token=${GITHUB_TOKEN}" \
     -H "Content-Type: application/json" \
     -X POST \
     -d "{\"state\": \"success\",\"context\": \"dashboard/surge/deploy/${DEPLOY_SUBDOMAIN}\", \"description\": \"Deploy domain: ${DASHBOARD_DEPLOY_DOMAIN}\", \"target_url\": \"${DASHBOARD_DEPLOY_DOMAIN}\"}"
 
-    curl "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/statuses/${TRAVIS_PULL_REQUEST_SHA}?access_token=${GITHUB_API_TOKEN}" \
+    curl "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/statuses/${ACTIONS_PULL_REQUEST_HEAD}?access_token=${GITHUB_TOKEN}" \
```
fzdy1914, line 71: Meanwhile, can you check if the repo owner here is changed due to our previous checkout command? (our new feature)

The `not found` command should come from here.

fzdy1914: Adding another environment variable should solve the problem.

dcshzj:
@fzdy1914 It should be working now, the `GITHUB_TOKEN` was not available in the workflow at all originally.

fzdy1914 (approved):
LGTM, it works.

jamessspanggg (approved):
LGTM

damithc:
Good work @dcshzj 
I'm actually interested (on the MarkBind site) to document how Surge+GitHubActions can be used to preview PRs for MarkBind sites. @jamessspanggg managed to do it with Surge+Travis and now you have managed to port it to Surge+GitHubActions. Feel free to submit a documentation PR to the MarkBind if you are interested.

fzdy1914:
Meanwhile, it seems the checking speed is a little bit faster than before.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,236 @@
...
+
+    - name: Build with Gradle
+      run: time ./gradlew clean checkstyleMain checkstyleTest test systemTest coverage
+
+    - name: Run code coverage
+      if: ${{ success() }}
+      run: bash <(curl -s https://codecov.io/bash)
+
+    - name: Deploy to surge.sh
+      if: ${{ success() }}
```
fzdy1914, line 88: `if: ${{ github.event_name == 'pull_request_target' }}`

Feels that we need to scope this as we only need it for PR?

PR #1413, opened 2021-01-18
[#674] Include binary files touched in authorship panel

### LINKED ISSUE #674: List binary files touched

HCY123902:
This is a continuation of pull request #1053.

fixes #674 

Commit message:
```
Include the binary files touched in authorship panel

The authorship panel currently does not show the binary files
touched by the author in the repository. This is because content
changes in binary files are usually unreadable by the user, as is
shown in #711. However, with the exclusion of binary files, the 
authorship panel cannot accurately reflect the authorship 
contribution.

Therefore, while it is still necessary to include the binary files 
touched, it will be appropriate to list the title of the binary files 
in the authorship panel without showing the content change to 
the user. This can make the authorship contribution summary 
accurate and organized.

Let's include the binary files touched in the authorship panel 
```


docs/ug/usingReports.md
```diff
@@ -140,6 +139,11 @@ The `Code panel` allows users to see the code attributed to a specific author. C
 * Code attributed to the author is highlighted in green.
 * Non-trivial code segments that are not written by the selected author are hidden by default, but you can toggle them by clicking on the %%:fas-plus-circle:%% icon.
 
+<box type="info" seamless>
+
+Changing of binary files, such as png and jpg files, is not counted towards the total number of lines of code contributed by an author.
```
Tejas2805, line 144:
```suggestion
Changing binary files, such as png and jpg files, is not counted towards the total number of lines of code contributed by an author.
```

Tejas2805 (changes requested):
1) I can't see a preview of the Binary label. Could you show it once?

2) Also, "troubleshooting in UG", see the last question and updated the UG accordingly.

HCY123902:
image
Thanks for the comment. Here is how the label looks like. There is no line count shown since it is not relevant to binary file. The font colour is green so that it is consistent with the font colour of the label shown beside the file type checkbox.

docs/ug/usingReports.md
HCY123902: Thanks for pointing it out. I have edited it accordingly. Is there any further comment on the documentation part?

Tejas2805: In "Troubleshooting" in the UG we say that we don't analyze such files, so that needs to be removed and the documentation regarding the files we analyze needs to be updated accordingly.

HCY123902: I have removed that section. I also changed the documentation regarding the formats flag and the File formats argument in the repo csv file, since adding extensions such as `png` and `jpg` will include those files grouped under the file type `Binary`

0blivious:
@damithc Hi prof can we get your opinion on this UI design?

damithc:
> @damithc Hi prof can we get your opinion on this UI design?

Looks good to me 👍

frontend/src/tabs/authorship.pug
```diff
@@ -98,17 +105,30 @@
...
                 span.tooltip-text Click to view the blame view of file
           span.fileTypeLabel(
+            v-if='!file.isBinary',
             v-bind:style="{\
               'background-color': fileTypeColors[file.fileType],\
               'color': this.getFontColor(fileTypeColors[file.fileType])\
               }"
-          ) {{ file.fileType }}&nbsp;{{ file.lineCount }}&nbsp;({{ file.lineCount - file.blankLineCount }})
-        pre.hljs.file-content(v-if="file.wasCodeLoaded", v-show="file.active")
+          ) {{ file.fileType }}&nbsp;{{ file.lineCount }}&nbsp;({{ file.blankLineCount }})
```
0blivious, line 121: Lines like this look like old code before we made the changes

HCY123902: Thanks for pointing it out. I have changed it accordingly.

frontend/src/static/js/v_authorship.js
```diff
@@ -245,22 +246,32 @@ window.vAuthorship = {
...
-          out.fileType = file.fileType;
+        // skip file not touched by author
+        if (!this.info.isMergeGroup && !(this.info.author in contributionMap)) {
+          return;
+        }
+
+        if (this.info.isMergeGroup && lineCnt === 0) {
+          return;
+        }
+
```
0blivious, line 257: And here also looks like legacy code. Could you please help go through the code once more and remove all these sections?

Also if possible can help squash your own changes into one/a few commits so it is easier for us to just look at the new changes.

HCY123902: This does seem to cause bug where the report does not show binary files when the group is merged. I have removed this section. So far I did not find other such sections, but please let me know if there is

HCY123902:
Please refer to commit 15d5c9f for the changes made based on the original pull request. I seem to have accidentally included changes from #1411, but it should not cause contribution problems since it is not included in the difference report of this pull request.

0blivious:
Got [Vue warn]: The computed property "selectedFiles" is already defined in data.

frontend/src/static/js/v_authorship.js
```diff
@@ -313,7 +328,8 @@ window.vAuthorship = {
       this.$store.commit('incrementLoadingOverlayCount', 1);
       setTimeout(() => {
         this.selectedFiles = this.files.filter(
-            (file) => this.selectedFileTypes.includes(file.fileType)
+            (file) => ((this.selectedFileTypes.includes(file.fileType) && !file.isBinary)
+            || (file.isBinary && this.isBinaryFilesChecked))
```
0blivious, line 332: just wondering why the condition so complicated. selectedfiletypes should not contain any binary file?

frontend/src/static/js/v_authorship.js
```diff
@@ -245,20 +246,29 @@ window.vAuthorship = {
             ? this.getContributionFromAllAuthors(contributionMap)
             : contributionMap[this.info.author];
 
-        if (lineCnt) {
+        // skip file not touched by author
+        if (!this.info.isMergeGroup && !(this.info.author in contributionMap)) {
+          return;
+        }
```
0blivious, line 252: what does this section do with the binary file change?

frontend/src/static/js/v_authorship.js
```diff
@@ -359,19 +376,29 @@ window.vAuthorship = {
 
     isSelectAllChecked: {
       get() {
-        return this.selectedFileTypes.length === this.fileTypes.length;
+        return this.selectedFileTypes.length === this.fileTypes.length && this.isBinaryFilesChecked;
```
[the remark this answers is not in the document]
HCY123902: @damithc Hello professor, can I get your opinion on this? Should all and binary be dependent on each other?

frontend/src/static/js/v_authorship.js
```diff
@@ -395,6 +422,11 @@ window.vAuthorship = {
     },
 
     ...Vuex.mapState(['fileTypeColors']),
+
+    binaryFilesCount() {
+      return this.files.filter((file) => file.isBinary).length;
+    },
+
```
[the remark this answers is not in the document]
HCY123902: Sure, no problem

frontend/src/static/js/v_authorship.js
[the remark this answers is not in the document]
damithc: What does it mean, from the user POV? What are the alternatives being considered and pros/cons of each?

frontend/src/tabs/authorship.pug
```diff
@@ -98,17 +105,30 @@
...
             v-bind:style="{\
               'background-color': fileTypeColors[file.fileType],\
               'color': this.getFontColor(fileTypeColors[file.fileType])\
               }"
           ) {{ file.fileType }}&nbsp;{{ file.lineCount }}&nbsp;({{ file.lineCount - file.blankLineCount }})
-        pre.hljs.file-content(v-if="file.wasCodeLoaded", v-show="file.active")
+          span.fileTypeLabel(
+            v-if='file.isBinary',
+            style='background-color: #FFF9C4; color: #1B5E20'
+          ) {{ 'Binary' }}&nbsp;
```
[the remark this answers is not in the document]
HCY123902: image

Uh will this be acceptable?

frontend/src/static/js/v_authorship.js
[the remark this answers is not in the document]
0blivious: From my point of view

1. `binary` is not displayed in the main summary panel when click break down by file types, making it like an independent feature we provided in authorship view.
2. the `all` label is followed by total LoC, which is independent of the binary files.
3. From a user's point of view, binary files are not really considered as code contributions and I would prefer it to be unticked by default.

frontend/src/static/js/v_authorship.js
HCY123902: Selected file types do not include binary files because binary file has 0 line count. I also think this should make sense because strictly speaking binary is not really a file type, rather it should be a file type group, containing file types such as `png` and `jpg`, and therefore it may be necessary to separate binary files with normal files.

frontend/src/tabs/authorship.pug
```diff
@@ -98,17 +105,30 @@
...
+        pre.hljs.file-content(v-if="file.wasCodeLoaded && !file.isBinary", v-show="file.active")
           template(v-for="segment in file.segments")
             v-segment(v-bind:segment="segment", v-bind:path="file.path")
+        pre.file-content(v-if="file.isBinary", v-show="file.active")
+          .binary-segment
+            .indicator BIN
+            .bin-text Binary file not shown.
+
+  .empty(v-else) loading...
```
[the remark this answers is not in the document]
HCY123902: Uh this is contributed by the original author. It does not seem to cause problem when I remove it.

frontend/src/static/js/v_authorship.js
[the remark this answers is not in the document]
HCY123902: Uh actually I think the current implementation already satisfies 1 and 2. Regarding 3, binary files are ticked by default when entering the author contribution view

frontend/src/static/js/v_authorship.js
HCY123902: This is contributed by the original author. It skips the files not touched by the current author. I am not sure but removing this section seem to cause bug where loading of authorship cannot terminate. I am looking into the problem but it may be necessary to keep it as it is at this moment.

frontend/src/static/js/v_authorship.js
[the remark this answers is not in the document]
damithc: > 3\. From a user's point of view, binary files are not really considered as code contributions and I would prefer it to be unticked by default.

I see your point. In that case, the binary files label should be visually different from others to clearly indicate that it is independent? e.g., bigger gap between it and other labels, make its appearance different (e.g., no bg color but light border) and text different (e.g., `[ ] 5 binary files`). What do you think?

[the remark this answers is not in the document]
0blivious: Yes, I think that will be a better design. Also, when I tick/untick `all`, the `binary` checkbox should not be affected

frontend/src/tabs/authorship.pug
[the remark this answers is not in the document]
0blivious: Pls remove it, I think that is part of legacy code

frontend/src/tabs/authorship.pug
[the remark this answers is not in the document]
0blivious: Please refer to Prof Damith's comment regarding making this label more independent. Also, for UI changes, please always seek approval from prof.

frontend/src/static/js/v_authorship.js
0blivious: So what if I specify jpg in file type? If I tick binary but untick jpg? Could you please set up a repo and respective argument for us to test?

frontend/src/tabs/authorship.pug
```diff
@@ -77,6 +77,13 @@
               span(v-bind:title="getFileTypeBlankLineInfo(fileType)")
                 span {{ fileType }}&nbsp;{{ fileTypeLinesObj[fileType] }}&nbsp;
                 span ({{ fileTypeLinesObj[fileType] - fileTypeBlankLinesObj[fileType] }})&nbsp;
+          label.binary-fileType(v-if="binaryFilesCount > 0")
+            input.mui-checkbox--fileType(type="checkbox", v-bind:checked="isBinaryFilesChecked",
```
0blivious, line 81: if it is independent now, should be able to directly bind the value to a vue model?

frontend/src/static/js/v_authorship.js
0blivious: Yes, but the line here looks very similar to the previous section. Instead of conditioning on linecount and isBinary, can we have a function like isValidFile? Also, you can calculate linecount after the if condition.

frontend/src/static/js/v_authorship.js
HCY123902: Thanks for pointing it out. `jpg` and `png` will be grouped under 'binary' if they are specified by the user. If the user specifies `jpg` in the format argument, the user will see the `jpg` files in under `binary`. Therefore, technically speaking the `jpg` checkbox will not appear.

frontend/src/static/js/v_authorship.js
HCY123902: Sure, I will refactor this section.

frontend/src/tabs/authorship.pug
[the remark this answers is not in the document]
HCY123902: image

@damithc Hello professor, can you give some comment on this design?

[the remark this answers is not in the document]
damithc: Perhaps increase the gap between the binary label and the other labels?
image

To see how it looks, we need some more file type labels though.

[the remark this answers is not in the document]
HCY123902: image

Here is how it looks after increasing the gap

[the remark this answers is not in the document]
damithc: Still looks too similar. Remove the border?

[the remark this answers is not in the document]
HCY123902: image

I made the border lighter. Can I ask for some recommendation on the font colour?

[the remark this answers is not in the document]
damithc: I think we have to remove it entirely if we want it to look like a _different_ thing. But I'm OK to keep the border also. Also, make it `32 binary file(s)`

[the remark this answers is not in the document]
HCY123902: Sure, no problem

Tejas2805 (changes requested):

docs/ug/troubleshooting.md
```diff
@@ -36,6 +36,6 @@ It is possible you may have some file names with [special characters](https://do
 
 <!-- ------------------------------------------------------------------------------------------------------ -->
 
-### Some files are not captured by RepoSense
+### Some file types are not shown in the file type filter even if I have included them in the file formats when generating the report
 
-The files may be [binary files](https://en.wikipedia.org/wiki/Binary_file). *RepoSense* does not analyze binary files. Common binary files include images (`.jpg`, `.png`), applications (`.exe`), zip files (`.zip`, `.rar`) and certain document types (`.docx`, `.pptx`).
+The files of these types may be [binary files](https://en.wikipedia.org/wiki/Binary_file). *RepoSense* will group binary files under one single file type `Binary`. Common binary files include images (`.jpg`, `.png`), applications (`.exe`), zip files (`.zip`, `.rar`) and certain document types (`.docx`, `.pptx`).
```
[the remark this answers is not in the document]
HCY123902: Sure, no problem

frontend/src/tabs/authorship.pug
HCY123902: Sure, no problem

docs/ug/cli.md
```diff
@@ -69,6 +69,10 @@ The section below provides explanations for each of the flags.
 * Alias: `-f`
 * Example:`--formats css fxml gradle` or `-f css fxml gradle`
 
+<box type="info" seamless>
+
+Binary file formats, such as `jpg`, `png`,`exe`,`zip`, `rar`, `docx`, and `pptx`, will all be labelled as the file type `binary` in the generated report.
```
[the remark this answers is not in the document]
Tejas2805: Not sure about this change though. Can decide based on your opinion.

docs/ug/configFiles.md
```diff
@@ -131,7 +131,7 @@ Note: all fields are optional unless specified otherwise.
 **Fields to provide _repository-level_ info**:
 
 * `ignoreGlobList`: Folders/files to ignore, specified using the [_glob format_](https://docs.oracle.com/javase/tutorial/essential/io/fileOps.html#glob).
-* `formats`: File formats to analyze. Default: all file formats
+* `formats`: File formats to analyze. Binary file formats, such as `png` and `jpg`, will be automatically labelled as the file type `Binary` in the generated report. Default: all file formats
```
Tejas2805, line 134:
```suggestion
* `formats`: File formats to analyze. Binary file formats, such as `png` and `jpg`, will be automatically labelled as the file type `binary` in the generated report. Default: all file formats
```

Tejas2805 (changes requested):

Tejas2805:
Maybe we can add a test for -> Binary files affect the number of files changed but not the number of lines of code changed. What do you think @fzdy1914 ?

docs/ug/configFiles.md
HCY123902: Sorry I did not seem to notice it previously. I have changed it accordingly.

docs/ug/cli.md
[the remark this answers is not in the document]
HCY123902: Sure, no problem

frontend/src/static/js/v_authorship.js
```diff
@@ -245,20 +246,25 @@ window.vAuthorship = {
             ? this.getContributionFromAllAuthors(contributionMap)
             : contributionMap[this.info.author];
 
-        if (lineCnt) {
+        if (this.isValidFile(this.info.isMergeGroup,
+            this.info.author, contributionMap, lineCnt, file.isBinary)) {
```
0blivious, line 250: why passing this.info as parameter?

frontend/src/static/js/v_authorship.js
```diff
@@ -245,20 +246,25 @@ window.vAuthorship = {
             ? this.getContributionFromAllAuthors(contributionMap)
             : contributionMap[this.info.author];
```
0blivious, line 247: can check valid file first then calculate linecnt

frontend/src/static/js/v_authorship.js
```diff
@@ -372,6 +385,30 @@ window.vAuthorship = {
...
+        } else {
+          this.isBinaryFilesChecked = false;
+        }
+
+        this.updateSelectedFiles();
+        this.indicateCheckBoxes();
+      },
+    },
+
+    selectedFiles() {
```
0blivious, line 404: make this either data or computed property

0blivious:
Can you enable restoring of isBinaryChecked?

frontend/src/static/js/v_authorship.js
HCY123902: Thanks for the suggestion. I tried to refactor it but I think the original intention is that file that is not binary and has 0 line count will not be included as valid files, so it may be necessary to compute the line count before checking whether the file is valid. Still, please let me know if there is a workaround as I am still learning this section

frontend/src/static/js/v_authorship.js
HCY123902: Actually I think this section is redundant and I will remove it accordingly.

frontend/src/static/js/v_authorship.js
HCY123902: Thanks for pointing it out. I will remove it accordingly.

HCY123902:
Sorry by restoring do you mean restoring when reloading the page?

0blivious:
@fzdy1914 can we add a test repo containing binary files for future dashboard deploy pipeline?

frontend/src/static/js/v_authorship.js
0blivious: Using line count is just some hacking when the group is merged since we did not have this binary type which always has 0 line count. It is ok to compute the lineCount first, but then your isValidFile logic has overlapping part with the lineCount section. You may want to read `authorship.json` to better understand the logic here.

frontend/src/static/js/v_authorship.js
```diff
@@ -245,20 +251,24 @@ window.vAuthorship = {
...
 
-        if (lineCnt) {
+        if (this.isValidFile(contributionMap, lineCnt, file.isBinary)) {
           const out = {};
           out.path = file.path;
           out.lineCount = lineCnt;
           out.active = lineCnt <= COLLAPSED_VIEW_LINE_COUNT_THRESHOLD;
           out.wasCodeLoaded = lineCnt <= COLLAPSED_VIEW_LINE_COUNT_THRESHOLD;
           out.fileType = file.fileType;
+          out.isBinary = file.isBinary;
```
0blivious, line 261: Now all the non-binary file will have this section as undefined, pls make them have false value

frontend/src/static/js/v_authorship.js
```diff
@@ -304,7 +320,12 @@ window.vAuthorship = {
           ? this.selectedFileTypes.reduce((a, b) => `${a}~${b}`)
           : '';
 
+      const binaryHash = this.isBinaryFilesChecked
```
0blivious, line 323: the name suggesting this should be some boolean value, even if in string

frontend/src/static/js/v_authorship.js
0blivious: So let me check the logic here, the `this.selectedFileTypes` will never include any binary file types according to your explanation? Then what does the condition `this.selectedFileTypes.includes(file.fileType) && !file.isBinary` do?

frontend/src/tabs/authorship.pug
```diff
@@ -77,6 +77,13 @@
               span(v-bind:title="getFileTypeBlankLineInfo(fileType)")
                 span {{ fileType }}&nbsp;{{ fileTypeLinesObj[fileType] }}&nbsp;
                 span ({{ fileTypeLinesObj[fileType] - fileTypeBlankLinesObj[fileType] }})&nbsp;
+          span &nbsp;&nbsp;&nbsp;&nbsp;
```
0blivious, line 80: why the spacing is controlled by &nbsp instead of css?

frontend/src/tabs/authorship.pug
```diff
@@ -98,17 +105,28 @@
...
             v-bind:style="{\
               'background-color': fileTypeColors[file.fileType],\
               'color': this.getFontColor(fileTypeColors[file.fileType])\
               }"
           ) {{ file.fileType }}&nbsp;{{ file.lineCount }}&nbsp;({{ file.lineCount - file.blankLineCount }})
-        pre.hljs.file-content(v-if="file.wasCodeLoaded", v-show="file.active")
+          span.fileTypeLabel(
+            v-if='file.isBinary',
+            style="background-color: #FFF; color: #424242; border: 1px solid #424242"
+          ) binary&nbsp;
```
0blivious, line 125: Instead of in-line style, use css

frontend/src/tabs/authorship.pug
HCY123902: Uh this is to prevent the situation where the binary checkbox appears as the first checkbox on a new line, in which case setting left margin in css may cause unnecessary indentation of the checkbox

frontend/src/tabs/authorship.pug
HCY123902: Sure, no problem

frontend/src/static/js/v_authorship.js
HCY123902: Thanks for pointing it out. I will change it accordingly.

frontend/src/static/js/v_authorship.js
HCY123902: Sure, no problem

frontend/src/static/js/v_authorship.js
HCY123902: I have moved the validation to the position before the calculation of line count. However, I am still not sure about whether the changes made correspond to the given suggestion. Please let me know if the changes made are still inconsistent with the suggestion given and I will look into it again.

frontend/src/tabs/authorship.pug
0blivious: I think it is possible to make this binary label automatically aligned to right. Can you give some screenshots of

1.  binary first of row with nbsp
2.  binary  first of row with css
3.  binary aligned to right

And seek comments from Prof?

frontend/src/static/js/v_authorship.js
0blivious: Following your current implementation, the logic here should be

1. if not merged group -> check if in contribution map
2. if merged group -> check for all authors in the group, if > 0 summed contribution
3.  if satisfy either 1 or 2 -> a valid file for consideration, do the following computation

Also, don't use return within ForEach, do a filter first

frontend/src/tabs/authorship.pug
HCY123902: image

image

image

@damithc Hello, professor. In the situation where the binary checkbox appears as the first checkbox in a row, which design do you think is suitable?

damithc: I'm OK with aligning right (even in cases where it appears as the last label of a row)

Tejas2805:
> @fzdy1914 can we add a test repo containing binary files for future dashboard deploy pipeline?

Need your update on this @fzdy1914

fzdy1914:
> @fzdy1914 can we add a test repo containing binary files for future dashboard deploy pipeline?

Of course, we can. Just add it to a branch of one of our test repo.

HCY123902:
It seems that I cannot create new branch in test repository. Do I need to first gain permission to create new branch?

fzdy1914:
> It seems that I cannot create new branch in test repository. Do I need to first gain permission to create new branch?

image
Should be fine now.

frontend/src/static/js/v_authorship.js
HCY123902: Sure, no problem

frontend/src/static/js/v_authorship.js
HCY123902: After some trial, I found that this check is necessary because there can be a situation where the user specifies the file group for a repository. In this case, it is possible that a binary file will be included in both the file group and the binary group without this condition.

HCY123902:
Is there any other suggestion or comment on this pull request? Many thanks

frontend/src/static/js/v_authorship.js
```diff
@@ -277,6 +291,16 @@ window.vAuthorship = {
       this.updateSelectedFiles();
     },
 
+    isValidFile(file) {
+      file.groupLineCnt = this.info.isMergeGroup
+          ? this.getContributionFromAllAuthors(file.authorContributionMap)
```
0blivious, line 296: is `getContributionFromAllAuthors` really needed? In `authorContributionMap` (except binary ones), do we put in any author with 0 contributions? If not, why do you just check if there exists any named author other than `-`?

frontend/src/static/js/v_authorship.js
```diff
@@ -304,7 +329,12 @@ window.vAuthorship = {
           ? this.selectedFileTypes.reduce((a, b) => `${a}~${b}`)
           : '';
 
+      const binaryFileTypeHash = this.isBinaryFilesChecked
+          ? 'binary'
+          : '';
```
0blivious, line 334: can you make this 'true' and 'false'?

frontend/src/static/js/v_authorship.js
0blivious: Could you please provide an example for this?

HCY123902: image

There can be issue caused by special name, especially related to the `other` category. When the extension of the binary file cannot be recognized, it will be both in the `other` group and in the binary group without this condition.

frontend/src/static/js/v_authorship.js
HCY123902: Sure, no problem

frontend/src/static/js/v_authorship.js
HCY123902: image


Thanks for pointing it out. After some trial, I found that marking a file as valid based on whether there exists named author instead of the line count may allow invalid file to sneak in. I am not sure about the exact reason but I assume that it is possible for there to be author with no contribution, which causes this issue.

frontend/src/static/js/v_authorship.js
0blivious: @Tejas2805 @fzdy1914 any idea for the file type handling here？

frontend/src/static/js/v_authorship.js
0blivious: Could you please provide the repo name and the argument you used so that I can have a look into this problem<

HCY123902: Actually the issue seems to have been resolved now. Sorry for the inconvenience. I have changed the implementation accordingly.

frontend/src/static/js/v_authorship.js
```diff
@@ -277,6 +291,16 @@ window.vAuthorship = {
       this.updateSelectedFiles();
     },
 
+    isValidFile(file) {
+      if (this.info.isMergeGroup) {
+        const fileAuthorCnt = Object.entries(file.authorContributionMap)
+            .filter((authorCount) => !this.isUnknownAuthor(authorCount[0]))
```
0blivious, line 297: can we use some() here instead of computing the length?

HCY123902: Sure, no problem

frontend/src/static/js/v_authorship.js
Tejas2805: I think if not recognized, we can just keep it in 'other' and not in 'binary'. This will be consistent with what we mention in the docs regarding which file formats are recognized as `binary`.

Tejas2805: @fzdy1914 What do you think?

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -62,19 +66,68 @@ public static FileResult analyzeFile(RepoConfiguration config, FileInfo fileInfo
...
+     * Returns null if the file is missing from the local system, or none of the
+     * {@code Author} specified in {@code config} contributed to the file in {@code fileInfo}.
+     */
+    public static FileResult analyzeBinaryFile(RepoConfiguration config, FileInfo fileInfo) {
+        String relativePath = fileInfo.getPath();
+
+        if (Files.notExists(Paths.get(config.getRepoRoot(), relativePath))) {
+            logger.severe(String.format(MESSAGE_FILE_MISSING, relativePath));
+            return null;
+        }
```
Tejas2805, line 84: Can add a test that ends up displaying this image? Need not be an elaborate one.

frontend/src/static/js/v_authorship.js
Tejas2805: I am okay with the current implementation based on what I saw in the deployed report.

HCY123902: Uh I think the intention of the original author is that binary file should not be recognized only by its extension but also by the actual file content, because there can be binary file with extension intentionally changed or deleted, and normal file with binary extension. In this case it may not be sufficient to decide whether it is a binary file based on the extension. One example that I can think of is in https://github.com/reposense/testrepo-Alpha.git, specifically in the `My picture` folder and the `binaryFileTest` folder, where whether a file is binary or not should not be decided simply on the extension. This should be part of the reason why the file above without recognizable extension is still recognized as a binary file here, which seems to be correct implementation on my personal opinion. However, I am not sure about this and please let me know if you have opinon on this specific issue.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
HCY123902: Sure, no problem

0blivious (approved):
Frontend LGTM

HCY123902:
@fzdy1914 Hello, can I ask for some opinion on this pull request? Many thanks

Tejas2805:
Is there any specific reason for the `binary` files to be open by default? Because even when I click `hide all details` and then click on the binary files `checkbox` they open by default. Since they actually don't have anything to show, I think it is better if they are closed by default or atleast follow the `hide all details/show all details` button.

HCY123902:
> Is there any specific reason for the `binary` files to be open by default? Because even when I click `hide all details` and then click on the binary files `checkbox` they open by default. Since they actually don't have anything to show, I think it is better if they are closed by default or atleast follow the `hide all details/show all details` button.

Sure, no problem

fzdy1914 (changes requested):
image

Did Prof mention that it should be `float: right`? Seems weird though.

fzdy1914:
Logic wise seems to be OK for me. But I may need more time to check the logic to avoid potential regression.

HCY123902:
> image
> 
> Did Prof mention that it should be `float: right`? Seems weird though.

Yeah it was suggested by @0blivious and approved by the professor

fzdy1914 (approved):
LGTM

Tejas2805 (approved):
I know it was a lot of work to put in especially with all the changes and add-ons. But job well done!

fzdy1914:
@HCY123902 Final question before merging. Now, the binary files are not included in `ALL`. Like if you uncheck `ALL` in authorship panel, the binary file will not be unchecked.

Is it intended?

HCY123902:
> @HCY123902 Final question before merging. Now, the binary files are not included in `ALL`. Like if you uncheck `ALL` in authorship panel, the binary file will not be unchecked.
> 
> Is it intended?

Yes, it was also suggested by @0blivious and approved by the professor

Issue #1340, opened 2020-10-06
Deprecate use of window.toggleNext in zoom panel


dcshzj:
Currently, `window.toggleNext` is being used by the ~segment, summary chart and~ zoom panels to toggle between displaying and hiding information. However, the authorship panel uses `v-show` to handle the same functionality. Even for the segment, summary chart and zoom panels, they each have their own ways of handling the toggling of the content.

We should deprecate the use of `window.toggleNext` and provide a common functionality for handling toggling of information. This would also allow other components to be aware of whether a particular element is hidden or is displayed.

Potentially related to #795.

LuoZhijie-tom:
Hi is this issue still open? Is it ok to work on it?

LuoZhijie-tom:
Hi @dcshzj, I wonder if this is still ongoing and if I can attempt it?

damithc:
> Hi @dcshzj, I wonder if this is still ongoing and if I can attempt it?

@dcshzj what do you think?

LuoZhijie-tom:
Hi @dcshzj, is this issue still unsolved? I wonder can I have a try?

vvidday:
Hi, could I try taking this issue?

PR #1416, opened 2021-01-19
[#1340] v_segment: rewrite toggling logic

gerhean:
Does not fully fix [#1340], toggle next still exists in `zoom`.

Rewrite `v_segment.js`

`v_segment.js` is written confusingly. It is not clear how exactly it works.

Note: v-segment is a component of v-authorship.

```
v-segment currently uses window.toggleNext() 
to show and hide code.

window.toggleNext() directly modifies html 
(add classes and elements directly), 
and so might conflict with how Vue works.

As a result, using window.toggleNext() 
creates code which is hard to debug and maintain.

Lets rewrite v-segment to use the Vue way 
of hiding and showing elements by using v-if.

In order to do this, some css must be rewritten, 
and parts of the html must be changed.
```


fzdy1914:
The topic is under #1340. Do check if there is any toggle next code in the frontend. If not, pls delete the function also.

gerhean:
It still exists in `zoom.pug`. Maybe can remove in another commit?

fzdy1914:
> It still exists in `zoom.pug`. Maybe can remove in another PR?

OK

Tejas2805:
@gerhean In such cases, it might be a good thing to explain which part of the code are you removing and why (if you have removed some part completely unless very obvious)

Also, can you update your commit message based on our guidelines.

gerhean:
@Tejas2805 I have rewritten the commit message.

Tejas2805:
> @Tejas2805 I have rewritten the commit message.

@gerhean Same guideline here. Max 70 characters in a line in the commit message.

frontend/src/static/js/v_segment.js
```diff
@@ -3,24 +3,14 @@ window.vSegment = {
   template: window.$('v_segment').innerHTML,
   data() {
     return {
-      loaded: this.segment.authored || this.segment.lines.length < 5,
+      isOpen: this.segment.authored || this.segment.lines.length < 5,
+      hasCloser: !this.segment.authored && this.segment.lines.length > 4,
```
Tejas2805, line 7: `hasCloser` kind of seems like a misnomer. Can name it more appropriately?

Tejas2805 (changes requested):

frontend/src/static/js/v_segment.js
gerhean: I can't really think of a better name since its a boolean which determines if the component has a `closer` which I got from how the html class is named. If I remember from my cs2103 booleans usually should start with 'is' or 'has', google says 'can' is usually used too, so would `canOpen` be a better name?

fzdy1914:
> > @Tejas2805 I have rewritten the commit message.
> 
> @gerhean Same guideline here. Max 70 characters in a line in the commit message.

72 actually.

frontend/src/static/js/v_segment.js
fzdy1914: Both OK to me.

Tejas2805: `canOpen` sounds better to me.

fzdy1914: @gerhean Can you fix the issue of this comment?

fzdy1914 (approved):

Tejas2805 (approved):
LGTM!

Issue #1417, opened 2021-01-19
Change the way api.js loads json files


gerhean:
**What feature(s) would you like to see in RepoSense?**
Use fetch over XMLHttpRequest in `api.js`.
Use async await over promises.
Also make it possible to load zip files.

**Is the feature request related to a problem?**

![image](https://user-images.githubusercontent.com/16630400/105028223-b9420e00-5a8b-11eb-8fc3-bc98377a3ad3.png)
`const` so it can't actually be modified.

**If possible, describe the solution**
In PR.

PR #1418, opened 2021-01-19
[#1417] Change the way api.js loads json files

gerhean:
Use fetch over XMLHttpRequest in `api.js`.
Use async await over promises.
Also make it possible to load zip files.

image
`const` so it can't actually be modified.

```
Currently, api.js uses XMLHttpRequest and promises to load json files from the server.
I believe that using the fetch api with async-await results in cleaner and more easily maintainable code.

Furthermore, uploading zip files does not work, 
as the variable used to denote the zip file is a const null value,
which cannot be changed when using the website.

Lets use fetch and async-await to get the json file.

Lets fix the uploading of zip files.
```


Tejas2805:
```
Furthermore, uploading zip files does not work, 
as the variable used to denote the zip file is a const null value,
which cannot be changed when using the website.

Lets make uploading zip files work.
```

How has it been working till now? What does this affect?

Tejas2805 (approved):
The changes LGTM. But better to get the view of @0blivious for js files.

gerhean:
> How has it been working till now? What does this affect?

Uploading zip files just never worked. I don't think anyone ever tested what happens when `summary.json` isn't also hosted. You can test this by removing `summary.json` from `reposense-report` and opening the server, uploading `.zip` files doesn't do anything. This isn't relevant in production but still makes one wonder what is the point of the page (which will never be seen in production) which asks you to upload a `.zip` file.

0blivious (approved):

0blivious:
LGTM. Please change the commit message according to our guidelines.

frontend/src/static/js/api.js
```diff
@@ -138,26 +138,27 @@ window.getGroupName = function getGroupName(group, filterGroupSelection) {
...
-      };
-      xhr.send(null);
-    });
+    try {
+      const response = await fetch(fname);
+      // Not directly returned in case response is not actually json.
+      const json = await response.json();
+      return json;
+    } catch (e) {
+      throw new Error('Json file not found.');
```
fzdy1914, line 160: Codewise OK to me. However, this one should not be an error as uploading a ZIP file is also in our normal use cases.

gerhean: Oh the error is just to tell the functions calling it that it didn't find any json to load.
Returning an empty string would cause some things to crash.

For example, it would probably be fine for `this.loadJSON(`${REPORT_DIR}/summary.json`)` to throw an error if `${REPORT_DIR}/summary.json` doesn't exist.

Using a different implementation would require further changes.

fzdy1914: The issue is that, this error will cause an alert dialog to be opened saying (json file not found), which should not appear in our design.

frontend/src/static/js/api.js
```diff
@@ -138,96 +138,101 @@ window.getGroupName = function getGroupName(group, filterGroupSelection) {
 };
 
 window.api = {
-  loadJSON(fname) {
-    if (REPORT_ZIP) {
-      const zipObject = REPORT_ZIP.file(fname.slice(2));
```
gerhean, line 143: I realized that `.slice(2)`  actually removes `./` from the file name, however that is unintuitive as the REPORT_DIR is explicitly defined as a const, and so I changed the implementation slightly.

frontend/src/static/js/main.js
```diff
@@ -62,35 +62,38 @@ window.app = new window.Vue({
             window.alert('Either the .zip file is corrupted, or you uploaded a .zip file that is not generated '
                 + 'by RepoSense.');
           })
-          .then(() => this.updateReportView().then(() => this.renderTabHash()));
```
gerhean, line 65: The original implementation is actually wrong as calling `renderTabHash()` when `updateReportView` encounters an error in parsing the JSON does not make sense.

frontend/src/static/js/api.js
gerhean: Should be resolved

gerhean:
One thing I noticed is that if you upload a different `archive.zip`, you might encounter console errors because the information in the url might specify things in the old zip which does not exist in the new zip.

fzdy1914 (approved):
Tested

PR #1419, opened 2021-01-19
Rewrite how zoom refreshes

gerhean:
`this.$root.$emit` is deprecated in Vue 3 so I removed it, and slightly updated how the zoom tab loads data on page refresh.

```
Currently the zoom tab uses this.$root.$emit 
to obtain information from summary.js 
during a page refresh.

However I am looking into upgrading to Vue3 
in the projectify frontend PR which 
does not support this.

Let's update the way zoom tab loads data 
on page refresh to be compatible with Vue3.
```


Tejas2805:
@gerhean Can you write a commit message explaining why the change and why is the new code better?

frontend/src/static/js/main.js
```diff
@@ -159,7 +159,7 @@ window.app = new window.Vue({
       };
       const tabInfoLength = Object.values(zoomInfo).filter((x) => x !== null).length;
       if (Object.keys(zoomInfo).length === tabInfoLength) {
-        this.$store.commit('updateTabZoomInfo', zoomInfo);
+        this.$refs.summary.restoreZoomFiltered(zoomInfo);
```
0blivious, line 162: We tried to avoid use this.$ref/$parent etc, could you try to refactor this part since you are on the way projectify the app?

gerhean: I can't think of any easy way to do this.

gerhean: Actually this can be done by moving some more code out of summary.js

0blivious: Since you actually have another PR for modifying summary.js, you can do the changes needed there and we will merge this PR later.

gerhean: Actually there's data stored in `summary.js` which cannot really be moved out unless you move out a lot from `summary.js`. I can try to remove `this.ref` after #1418 is merged however.

fzdy1914: Merged #1418

frontend/src/static/js/v_zoom.js
```diff
@@ -133,11 +135,6 @@ window.vZoom = {
       }
     },
 
-    restoreZoomTab() {
-      // restore selected user's commits and file type colors from v_summary
-      this.$root.$emit('restoreCommits', this.info);
```
fzdy1914, line 138: Since this method is not used, should it also be removed?

fzdy1914 (changes requested):

fzdy1914:
PLS help to solve the conflicts

frontend/src/static/js/v_zoom.js
gerhean: I don't understand what you mean. The method is not used so it has already been removed?

fzdy1914: My bad, didn't see this.
image

fzdy1914 (approved):
LGTM

Issue #1420, opened 2021-01-19
Abstract code out from summary.js


gerhean:
**What feature(s) would you like to see in RepoSense?**
Move code which does not rely on the state of `summary.js` out of `summary.js` to somewhere else.
Also can apply to other files.

**Is the feature request related to a problem?**
`summary.js` has over 900 lines and is very messy. Looking for relevant code can be hard
Also would help if we decide to move handling of repo data out of `summary.js` (maybe to Vuex).

fzdy1914:
Yes, can go ahead. And you can choose to do it before or after morphing.

damithc:
@gerhean is there more to do for this issue or done already?

gerhean:
I'll close this first, not planning to do much about this at the moment.

PR #1421, opened 2021-01-19
[#1420] Extract some code from summary.js

gerhean:
Some of [#1420], but more can certainly be done.

```
Currently summary.js is very long (900+ lines).

This makes maintaining code hard 
as it not always clear which code causes
changes in the components state.

Ideally, all code which do not modify
state should be moved out.

This further helps to decouple code,
as it forces the dependencies of the functions
to be written clearly as arguments,
allowing the function to act as a reliable
black box which is also easier to test.

Lets move some code out of summary.js.
```


frontend/src/static/js/utils/repo_sorter.js
```diff
@@ -0,0 +1,134 @@
...
+  //     ? element.reduce(getGroupCommitsVariance, 0)
+  //     : element[0][sortingOption];
+  if (sortingOption === 'totalCommits') {
+    return element.reduce(getTotalCommits, 0);
+  }
+  if (sortingOption === 'variance') {
+    return element.reduce(getGroupCommitsVariance, 0);
+  }
+  return element[0][sortingOption];
+}
```
gerhean, line 29: The change I made to the code. Will remove comment once verified is ok.

Tejas2805: Any reason for making this change?

frontend/src/static/js/utils/ramp_colour_generator.js
```diff
@@ -0,0 +1,65 @@
+function getRandomHex() {
```
Tejas2805, line 1: There are no changes in the code in this file right? Except what is required for the integration.

Tejas2805 (commented):
Left a few initial comments.

frontend/src/static/js/utils/repo_sorter.js
gerhean: Actually I realized that the commented code was not the original code sorry.

The reason why I changed the original code is because it was using `this.sortingOption` which is not available when extracted, and that returning an anonymous function is not the best idea (I think there was a warning).

frontend/src/static/js/utils/ramp_colour_generator.js
gerhean: There are no changes besides removing `this.` from the function calls and loading them into window.

frontend/src/static/js/v_summary.js
```diff
@@ -361,7 +361,8 @@ window.vSummary = {
         isSortingDsc: this.isSortingDsc,
         isSortingWithinDsc: this.isSortingWithinDsc,
       };
-      this.filtered = this.sortFiltered(this.filtered, filterControl);
+      this.getOptionWithOrder();
```
0blivious, line 364: why this extra line?

fzdy1914:
@gerhean Pls answer the questions and proceed with PR.

gerhean:
Would it be good to merge this first and extract more stuff in another PR?

0blivious:
What are the other changes you plan to extract? Can you give some line number?

frontend/src/static/js/v_summary.js
0blivious: @gerhean Pls answer this

gerhean:
> What are the other changes you plan to extract? Can you give some line number?
At the moment I'm not really focusing on this so nothing yet.

0blivious (approved):

fzdy1914 (approved):

PR #1432, opened 2021-01-25
Handle CLI documentation bug

HCY123902:
Commit message:
```
There seems to be a bug in the CLI documentation

I assume that the original intention is to warn
the users not to use --repos with --config according
to the config section.

Let's change it to reduce confusion for the users.
```


Tejas2805 (commented):
LGTM!

Tejas2805 (approved):

PR #1434, opened 2021-01-25
[#1429] Make GitHub Actions workflows more specific for artifacts

### LINKED ISSUE #1429: GitHub Actions conflict if multiple deployments are done in a short amount of time
This issue is a follow up from #1408.

Due to the way the `dawidd6/action-download-artifact` action handles the detection of artifacts, it often results in a race condition where the wrong artifacts are used for certain `workflow_run` runs.

We should explore how to obtain the correct artifact for each `workflow_run`.

dcshzj:
Fixes #1429.

```
Currently, GitHub Actions can encounter race conditions where
multiple pull requests are deployed at once, causing confusion as
to the exact artifacts to download. This is caused by not
specifying the pull request numbers when downloading artifacts.

Let's make it such that the correct artifacts are downloaded even
when multiple deployments occur concurrently.
```

Testing pull requests:
- https://github.com/dcshzj/RepoSense/pull/4
- https://github.com/dcshzj/RepoSense/pull/5

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -79,7 +79,7 @@ jobs:
         npm install -g markbind-cli
         (cd docs && markbind build)
 
-    - name: Save PR number and HEAD commit
+    - name: Save PR number and HEAD commit (pull request)
       if: ${{ success() && github.event_name == 'pull_request' }}
       run: |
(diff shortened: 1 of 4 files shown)


fzdy1914 (approved):
LGTM

Issue #1426, opened 2021-01-22
Zoom reloading not function properly


0blivious:
**Tell us about your environment**

* **RepoSense Version:** master
* **OS and Version:** windows10
* **Web Browser and Version (if applicable):** Chrome

**Please include the steps to reproduce the bug.**

1. Visit https://nus-cs2103-ay2021s1.github.io/tp-dashboard/#breakdown=true&search=&sort=groupTitle&sortWithin=title&since=2020-08-14&timeframe=week&mergegroup=&groupSelect=groupByRepos&checkedFileTypes=docs~functional-code~test-code~other&until=2020-11-10&tabOpen=true&tabType=zoom&zFR=true&zA=sc-arecrow&zR=AY2021S1-CS2103-F09-1%2Ftp%5Bmaster%5D&zACS=2476&zS=2020-09-21&zFS=undefined&zU=2020-09-27&zMG=false&zFTF=commit&zFGS=groupByRepos
1. Open any zoom panel
2. Click on the ramp at the zoom panel
3. Refresh

**What was expected to happen?**
The page should reload with the zoom panel
![image](https://user-images.githubusercontent.com/39231750/105511790-43e47080-5d0b-11eb-92a2-250baecd72bc.png)

**What actually happened? Please include a screenshot of the output.**
Error in event handler for "restoreCommits": "TypeError: Cannot read property '0' of undefined"

**If possible, include the URL to your RepoSense report or log files (if any).**

fzdy1914:
Seems can not reproduce. Any other parameter specified?

gerhean:
Can guess what happened, but cannot reproduce either.

fzdy1914:
> Can guess what happened, but cannot reproduce either.

We need to reproduce it by `groupByRepos` option. I successfully reproduced it. That link is useful.

gerhean:
I still cannot reproduce it. In any case this link does not reflect the current master branch.

In the current master branch, clicking on the ramp at the zoom panel brings one to the commit on the github website and should not cause any changes to state. Hence this issue might not be relevant anymore.

fzdy1914:
> I successfully reproduced it. However this link does not reflect the current master branch
> 
> In the current master branch, clicking on the ramp at the zoom panel brings one to the commit on the github website and should not cause any changes to state. Hence this issue might not be relevant anymore.

Step to reproduce on current master:
1. open report
2. breakdown by filetyple
3. granularity by week
4. view any commit pannel
[...]

PR #1437, opened 2021-01-29
[#1426] Zoom reloading not function properly

gerhean:
```
When following the steps mentioned in #1426,
we get an error.

The error is caused by the zFilterSearch parameter of the url
being the string 'undefined', which causes restoreZoomFiltered
in v_summary.js to crash as 'isMatchSearchedUser(zFilterSearch, user)'
does not match any user leading to an array index out of bounds error.

Lets fix this by ensuring that zFilterSearch is defined in v_ramp, 
such that zFilterSearch='undefined' is not generated in the url,
when one clicks on the ramp.

This does not fix the fact that a bad url can still cause the error.
```


0blivious (approved):
LGTM, please write a complete PR message, including what causes the error and the explanation for your solution. Then you can ask review from stage2

fzdy1914:
@gerhean Just a question, as mentioned in your commit message, a bad URL can still cause an error.  How do you define a bad URL? Can we ignore such parameters if they are provided?

gerhean:
@fzdy1914 
I would define a bad URL as a URL which would cause the reposense website to crash, even if the URL cannot possibly be generated by the website itself.
Example of bad URL: go to any reposense, look for `&sort=groupTitle` in the url and change it to `&sort=a`, on refresh it will completely not work.

fzdy1914:
> @fzdy1914
> I would define a bad URL as a URL which would cause the reposense website to crash, even if the URL cannot possibly be generated by the website itself.
> Example of bad URL: go to any reposense, look for `&sort=groupTitle` in the url and change it to `&sort=a`, on refresh it will completely not work.

Seems that this can be solved easily by a pre-verification step? Can you put it as an issue to solve such invalid parameters?

Issue #1423, opened 2021-01-20
Fix vue build warning


fzdy1914:
![image](https://user-images.githubusercontent.com/35621726/105150592-ebad4300-5b3f-11eb-8100-9fb43b397810.png)

We should try to look into these Vue warnings and try to fix them.

gerhean:
Is it ok to just tell eslint to ignore the alert? Unless there's any better alternatives to telling users that a zip file is bad?

PR #1438, opened 2021-01-29
[#1423] Fix vue build warning

gerhean:
Not a complete fix.

```
Currently, ESLint complains about unnamed functions.

Lets name all unnamed functions.
```

### DIFF
--- a/frontend/src/static/js/main.js
+++ b/frontend/src/static/js/main.js
@@ -38,15 +38,15 @@ window.app = new window.Vue({
     errorMessages: {},
   },
   watch: {
-    '$store.state.tabZoomInfo': function () {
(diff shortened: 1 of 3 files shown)


Tejas2805:
@gerhean I am not that good with js but is there any other way to write it? In the sense can the code be modified so that the error is not shown and we avoid duplicating -> window.x = function x()?

gerhean:
> I am not that good with js but is there any other way to write it? In the sense can the code be modified so that the error is not shown and we avoid duplicating -> window.x = function x()?

Possible, but there needs to be consistency. Currently `api.js` uses window.x = function x() so should I change these too?
image

Issue #1440, opened 2021-01-29
Call `this.getFiltered()` only once during initialization.


gerhean:
**What feature(s) would you like to see in RepoSense?**

Call `this.getFiltered()` only once during initialization.

**Is the feature request related to a problem?**

![image](https://user-images.githubusercontent.com/16630400/106261970-41c76800-625d-11eb-9139-c42904400d68.png)

When opening the website, `getFiltered()` is being called multiple times by `v-summary`.

This is due to watched properties changing during initialization.
![image](https://user-images.githubusercontent.com/16630400/106262150-894df400-625d-11eb-872e-27c51944b5bb.png)

![image](https://user-images.githubusercontent.com/16630400/106262846-8acbec00-625e-11eb-8e6d-eac76abfa72c.png)

`getFiltered()` is a very expensive operation so it should ideally only be called once during initialization.

This is causing other problems in projectify frontend.

**If possible, describe the solution**

Remove `getFiltered` from watchers, and directly call it from the template using `onChange`

**If applicable, describe alternatives you've considered**

I'm open to other ideas as well.

PR #1441, opened 2021-01-29
Use computed `filterSinceDate`

gerhean:
Is linked to [#1440]

```
Currently in `v-summary.js`, `filterSinceDate` is changed 
almost exclusively by the watcher of `tmpFilterSinceDate`.

The watcher of `tmpFilterSinceDate` also calls `getFiltered()`
which is a very expensive operation.

Lets turn `filterSinceDate` into a computed function which
is easier to understand.

Lets move `getFiltered()` out of the watcher of `tmpFilterSinceDate`.

Lets make similar changes to `tmpFilterUntilDate` and `filterUntilDate`.
```


0blivious:
Could you please elaborate more on the rationale of this PR? The `getFiltered` will need to be called when `filterSinceDate` etc changed. Why not just watch this property?

gerhean:
The biggest problem is here when `renderFilterHash` is called on `created()`.


Can refer to #1447 for more explaination.

fzdy1914 (approved):

PR #1442, opened 2021-01-30
[#1423] Fix Vue build warning

### LINKED ISSUE #1423: Fix vue build warning

gerhean:
Fixes #1423
```
Currently, ESLint complains about unnamed functions,
alerts and linebreaks.

Lets stop ESLint from complaining, 
and standardize function declarations.
```


.eslintrc.json
```diff
@@ -13,6 +13,9 @@
             "ignoredNodes": [
                 "ConditionalExpression"
             ]
-        }]
+        }],
+        "func-names": ["error", "as-needed"],
+        "no-alert": 0,
+        "linebreak-style": 0
```
fzdy1914, line 19: What is this for?

gerhean: `"func-names": ["error", "as-needed"],`
- stops eslint complaining about unnamed functions which it can infer the name of.

`"no-alert": 0,`
- stops eslint from complaining about alerts

`"linebreak-style": 0`
- stops eslint from complaining about windows linebreak.
Pushing to github should standardize the linebreak style on github (if it has already been set up).
Checking for linting errors in windows is fairly annoying in any case without this rule.

Tejas2805: > stops eslint from complaining about alerts

Does this mean we won't get if there are alerts but get to know if there are erroers?

gerhean: I don't really understand what you mean? Alerts are just these pieces of UI which many people find annoying:
image

Tejas2805: Ah, my bad. I confused "alert" with "warning" in a rush.

Tejas2805 (approved):
LGTM!

PR #1443, opened 2021-02-01
Cypress: add test for tabs rendering

rolandyuwy:
Add Cypress tests for #1390.

```
The above mentioned PR introduces non-trivial front-end changes.

Front-end tests for these changes are required to prevent regression 
bugs.

Tests are added specific to the front-end changes for the above 
mentioned PR.

Let's,
- add tests to check if default states remain upon page refreshing and 
tab switching for authorship and zoom tabs
- add tests to ensure information from previously visited tabs are not 
retained.

### CODE REVIEW

File: frontend/cypress/tests/codeView/codeView_switchAuthorship.js [hunk 1]
```diff
@@ -0,0 +1,69 @@
+describe('switch authorship', () => {
+  it('switch authorship view should restore all default controls', () => {
+    Cypress.wait();
+
+    // open the code panel
+    cy.get('.icon-button.fa-code')
+        .should('be.visible')
+        .first()
+        .click();
+
```
HCY123902, line 10: Hello, I have a question regarding this. Will it be necessary to have a check on the sorting and file types before the switch happens?

frontend/cypress/tests/codeView/codeView_switchAuthorship.js
```diff
@@ -0,0 +1,69 @@
...
+    // switch authorship view
+    cy.get('.icon-button.fa-code')
```
HCY123902, line 12: Hello, I have a question regarding this. Will it be necessary to have a check on the sorting and file types before the switch happens?

gerhean:
Can cypress check url? Maybe can use to check the url is correct?

frontend/cypress/tests/codeView/codeView_switchAuthorship.js
rolandyuwy: This test fundamentally checks if default controls are restored after switching between authorship tabs, so I don't see a need to perform checks before the switch. However, if the first authorship view is regarded as "switching of authorship view", then it can be included.

rolandyuwy:
> Can cypress check url? Maybe can use to check the url is correct?

Noted, will include checks for url.

frontend/cypress/tests/codeView/codeView_switchAuthorship.js
HCY123902: Yeah that makes sense. However, I am wondering whether it will be necessary to check if previous authorship tab control can affect the control after the switch. Will it be necessary to alter some of the tab control setting first before making the switch, and then check if the default control is restored? Still, it is just some personal thought.

rolandyuwy: Thanks for the input @HCY123902, altering the tab control settings before the switch is a valid consideration and has been implemented.

Tejas2805:
@rolandyuwy What's the status of this PR?

Tejas2805:
LGTM! @0blivious for your review

Tejas2805 (approved):

frontend/cypress/tests/codeView/codeView_switchAuthorship.js
```diff
@@ -0,0 +1,93 @@
...
+    Cypress.wait();
+
+    // open the code panel
+    cy.get('.icon-button.fa-code')
+        .should('be.visible')
+        .first()
+        .click();
+
+    cy.get('#tab-authorship .panel-heading .author span')
+        .should('contain.text', 'eugenepeh');
```
0blivious, line 66: The implementation here is assuming that eugenepeh will be the first in the group. However, this may not always be true as we might edit the configuration (add author, etc.). Could you change this part of the test to allow it to handle dynamic configuration?

0blivious (changes requested):
Also for the zoom part

frontend/cypress/tests/codeView/codeView_switchAuthorship.js
gerhean: I think it is hard to handle dynamic configuration as we at least need a guarantee that the first author in the repos we switch from and to is not the same. Otherwise there is nothing to test.

gerhean:
I'll be taking over this PR

0blivious (approved):

fzdy1914:
```
PR #1390 introduces non-trivial front-end changes to the way to 
render tabs. Frontend tests for these changes are required to 
prevent regressions.

Let's,
- add tests to check if default states remain upon page refreshing 
  and tab switching for authorship and zoom tabs
- add tests to ensure information from previously visited tabs are 
  not retained.
```

PR #1447, opened 2021-02-09
[#1440] Call `this.getFiltered()` only once during initialization.

gerhean:
```
Currently, in v_summary.js, state such as checkedFileTypes 
and filterBreakdown  are being watched in order to rerender 
v_summary when users select options.

However, renderFilterHash() also modifies these states when 
v_summary is created, leading to unneeded calls to getFiltered().

In other words getFiltered() is being redundantly called
multiple times during initialization.

getFiltered() is a very expensive operation so it should ideally 
only be called once during initialization.

This is also causing other problems in projectify frontend.

Lets remove such watchers, and directly call getFiltered() 
from the template.
```


dcshzj:
Sorry, to clarify, you mentioned that "getFiltered() is a very expensive operation so it should ideally only be called once during initialization", but it looks like getFiltered() is called on every change (`v-on:change`), which is contradictory. Will this be changed in the future to match your commit description?

Also, is it possible to put this in the same PR as #1441? They seem to be working on the same thing, which is the handling of filters on the frontend.

gerhean:
I don't think there is any contradictions since getFiltered() was previously called multiple times during initialisation and now it's being called once during initialisation. `v-on:change` is not called during initialisation so there is no contradictions?

PR #1441 is making a more complex change than this PR hence I believe it is better to seperate them.

frontend/src/summary.pug
```diff
@@ -56,8 +57,11 @@
         a(v-on:click="resetDateRange") Reset date range
       .summary-picker__checkboxes.summary-picker__section
         label.filter-breakdown
-          input.mui-checkbox(type="checkbox", v-model="filterBreakdown",
-            v-on:change="setSummaryHash()")
```
fzdy1914, line 60: Is `setSummaryHash` contained in getFiltered so there is no need to call it again?

gerhean: image
Yes

PR #1449, opened 2021-02-11
Implement shallow cloning

### LINKED ISSUE #1450: Implement shallow cloning to speed up cloning
Typically, RepoSense is used to display the contributions made to a repo over a relatively short period, e.g. a few months. This means the commit history prior to these months is not required to generate the report.

However, RepoSense still downloads the full commit history of the repo when cloning it. For repos with a long history of commits, especially those which have frequent changes to binary files, the size of the cloned repo can end up being very large. This also means that more time is needed to clone the repo.

We could use Git’s shallow cloning functions to reduce the size of the cloned repo, and consequently speed up the cloning process for large repos.

Git’s shallow cloning functionality allows us to obtain a cloned repo with truncated history. This can be done using the `--shallow-since=<date>` parameter of the `git clone` command. This results in a shallow clone of the repo with history going back only up till the specified date. The earliest commits in the shallow clone repo will be designated as “grafted” commits, and all ancestor commits of the grafted commits are inaccessible from the shallow clone repo.

Using the shallow clone functionality, we can cut down the size of the cloned repo by instead making a shallow clone and hence only downloading commit history up to the `--since` date passed to RepoSense. However, there are 2 issues that we would face.

The first issue is with regards to the grafted commits. The behaviour of grafted commits is a little unusual; e.g. if a merge commit ends up being the grafted commit, then it will show up in the output of the `git log --no-merges` command. This means that RepoSense’s commit and file info extraction process will be affected. The solution to this is to create a shallow clone with slightly more history, such that all of the commits that we need (i.e. all commits between the “since” and “until” dates) are not grafted commits.

The second issue is an edge case that causes some problems with the shallow cloning functionality in Git itself. If we are obtaining a shallow clone with history up to a particular “since” date, we would face a problem if there was a merge commit made after the “since” date, having 1 parent commit that was made before the “since” date and another parent commit that was made after the “since” date. In this edge case, neither of the 2 parent commits will be accessible in the shallow clone. (see the image below for a clearer illustration).

image

To address this, we need to first detect if the edge case is present and, if so, set the `--shallow-since` parameter date to the earlier of the 2 parent commits.

In order to address these 2 issues, we could use Git’s partial clone functionality. The partial clone functionality allows us to produce a clone repo without downloading any file blobs, using the `--filter=blob:none` parameter of the `git clone` command. Subsequently, when we run a command that requires the file contents for a particular commit (e.g. `git diff`), the required blobs will be downloaded on demand. As a result, the partial clone repo produced will be much smaller in size than a normal clone, and as such can typically be downloaded in seconds even for larger repos.

We can utilize the partial clone functionality to address the 2 aforementioned issues. First, we perform a shallow, partial clone with the history truncated to the `--since` date input to RepoSense. From this repo, we can use `git rev list` to obtain the commit hashes of all the grafted commits (i.e. the earliest ones in the commit tree). Then, we perform a partial clone with the full history. On this repo, we can search for the parent commits of the grafted commits obtained earlier, and then check the dates of these parent commits. Finally, we obtain the date of the earliest parent, and proceed to perform the shallow clone with the history truncated to this date. This takes care of both of the 2 issues, since all of our desired commits (i.e. all commits between the “since” and “until” dates) in the resulting shallow repo will not be grafted commits, and we will ensure that the aforementioned edge case will be taken care of.

**Performance improvement**

I have made a PR with an implementation of this solution (#1449) and tested it on a few large repos for a roughly 6 month period from last August to now. The solution significantly reduced the download size of the cloned bare repo, and consequently also sped up the time taken for RepoSense to clone the repos, which resulted in the report being generated faster. The results are shown here:

Repo | Current Size | Reduced Size | Current Clone   Time | Reduced Clone   Time
-- | -- | -- | -- | --
[RaspberryPi  Firmware](https://github.com/raspberrypi/firmware) | 16.6GB | 2.27GB <br> (86% lower) | 19min 20sec | 4min 30sec <br> (77% faster)
[Nerd Fonts](https://github.com/ryanoasis/nerd-fonts) | 6.21GB | 1.08GB <br> (83% lower) | 8min 25sec | 3min 5sec <br> (64% faster)
[Apollo](https://github.com/ApolloAuto/apollo) | 2.32GB | 1.27GB <br> (45% lower) | 2min 40sec | 1min 40sec <br> (38% faster)
[Selenium](https://github.com/SeleniumHQ/selenium) | 1.80GB | 116MB <br> (93% lower) | 2min 10sec | 40sec <br> (69% faster)
[Servo](https://github.com/servo/servo) | 868MB | 100MB <br> (88% lower) | 1min 40sec | 1min 5sec <br> (35% faster)

**Shortcomings**

There are 2 main shortcomings with this solution.

The first is that there is a time overhead associated with the 2 partial clones that are performed. For large repos, this is significantly outweighed by the time saved due to the reduction in size of the cloned repo. However, in case of small repos, this results in overhead of about 5-10 seconds per repo. When RepoSense is run with a large number of small repos, this adds up to a significant overhead. 

I tested this with 2 of the existing demos on the RepoSense websites: the CS2113 AY2019/20 Sem 2 individual project (72 repos) and team project (17 repos). On the 17 group project repos, RepoSense currently takes 2 min 40 sec to generate the report on my local machine, but with this solution implemented it goes up to around 4 min 5 sec. The effect is even more pronounced on the 72 individual project repos; RepoSense currently takes around 5 min 45 sec to generate the report, but it goes up to 12 min 55 sec with this solution implemented.

One possible way to address the shortcoming could be to implement this as an option in the configuration for each repo, rather than enabling it by default on every repo. Since the user is likely to know in advance if a particular repo is very large, they would be able to set the config accordingly to enable this optimization so as to speed up cloning and reduce the size of the cloned repo.

The second shortcoming is that the partial cloning functionality may not be compatible with all remote repositories, since it requires support from the server. While GitHub and GitLab repos support this functionality, there is an [as yet unresolved Jira ticket]( https://jira.atlassian.com/browse/BSERV-11639) asking Bitbucket to support this, which means that Bitbucket probably does not support it. In this case, if we detect an error with the partial clone, we would be forced to fall back to the normal cloning process.

vig42:
Fixes #1450

```
Currently, RepoSense downloads the entire commit history of the repo
when cloning it. This results in a lot of unnecessary history being
downloaded, which increases the size of the cloned repo and slows
down the time taken to clone the repo.

Let’s implement Git’s shallow cloning functionality so as to truncate
the downloaded history thereby reducing the size of the cloned repo
and speeding up the cloning process.
```


vig42:
System test is failing because the "lastModifiedDate" does not match. This is because of the shallow clone functionality - since the commit history of the cloned repo is truncated, RepoSense does not see any changes made prior to the `--since` date. As such, for lines that were last modified before the `--since` date, the "lastModifiedDate" value is incorrect.

The difference is only visible in the authorship.json when the `--last-modified-date` option is used. The actual reports generated are identical to those generated by the master branch.

Tejas2805:
> The difference is only visible in the authorship.json when the --last-modified-date option is used

If I understand this correctly, it means that the `--last-modified-date` option will not work if we switch to shallow cloning?

vig42:
> 
> 
> > The difference is only visible in the authorship.json when the --last-modified-date option is used
> 
> If I understand this correctly, it means that the `--last-modified-date` option will not work if we switch to shallow cloning?

Yes. With shallow cloning, the `--last-modified-date` would provide wrong values for all commits made prior to the `--shallow-since` date passed to the `git clone` command.

Tejas2805:
> As I had mentioned in issue #1450, I feel that it might be better to make the shallow cloning an option in the repo config. Then we could also specify in the documentation that the --last-modified-date option does not work on repos that have this option turned on. Would that be a better way to handle this?

Yup, we can proceed with it, I think. Plus it would be good to state that shallow cloning should only be used for large repositories and not for a large number of repositories considering the overhead. 

Also, based on what I read in the issue, are we avoiding partial cloning unless the 2 cases occur? Or will the partial cloning occur irrespective of the two cases?

vig42:
>
> Yup, we can proceed with it, I think. Plus it would be good to state that shallow cloning should only be used for large repositories and not for a large number of repositories considering the overhead.

Will do.

> 
> Also, based on what I read in the issue, are we avoiding partial cloning unless the 2 cases occur? Or will the partial cloning occur irrespective of the two cases?

Actually we need to perform partial cloning to detect the 2 cases in the first place, so we would have to do it regardless.

damithc:
@vig42 what's the verdict on costs and benefits? Can give a summary?

vig42:
I have made some changes to implement this functionality as a CLI flag `--optimize-cloning` and an option in repo-config.csv. By default, the option would not be enabled, i.e. RepoSense would not use shallow cloning.

> 
> 
> @vig42 what's the verdict on costs and benefits? Can give a summary?

The benefit is the ability to significantly reduce the download size of a large repository, thereby also reducing the time taken to clone the repo. I had provided some examples in #1450 where the download size of a large repo can be reduced by more than 50%, and as much as 90%, if we only look at the most recent 6 months of commits.

The first issue is that there is some overhead since we need to perform a partial clone to check for edge cases. While this overhead is insignificant relative to the massive time savings on large repos, it is quite noticeable on small repos and especially results in significantly slower performance when run on a large number of small repos.

However, given the current implementation as a config option, which can be turned on or off for each repo, this issue is not really relevant. As @Tejas2805 suggested, I have specified in the documentation that the option should not be used for smaller repos. As such, users should not experience the overhead.

The second issue is that partial cloning - which is needed to check for the edge cases - may not be compatible with all remote repositories. While it is supported by GitHub and GitLab, BitBucket notably does not support this functionality yet. If that is the case, then RepoSense will fall back to the usual method of cloning, which would also result in some overhead.

damithc:
Thanks for the summary @vig42 
I wonder if the flag name is too general e.g., what if we introduce other ways to optimize cloning in future? Perhaps make it more specific? e.g., `--partial-cloning`?

docs/ug/cli.md
```diff
@@ -107,6 +107,17 @@ The last modified dates will be in the same timezone specified with the `--timez
 
 <!-- ------------------------------------------------------------------------------------------------------ -->
 
+### `--optimize-cloning`, `-m`
+
+**`--optimize-cloning`**: Clones repositories using Git's shallow cloning functionality, which can significantly reduce the time taken to clone large repositories. However, the flag should not be used for smaller repositories. Note: this flag is incompatible with the `--last-modified-date`.
```
damithc, line 112: You might want to explain why it should not be used for small repos. also, what's considered a 'small' repo?

docs/ug/configFiles.md
```diff
@@ -38,6 +38,7 @@ Ignore Glob List<sup>*+</sup> | The list of file path globs to ignore during ana
 Ignore standalone config | To ignore the standalone config file (if any) in the target repository, enter **`yes`**. If the cell is empty, the standalone config file in the repo (if any) will take precedence over configurations provided in the csv files.
 Ignore Commit List<sup>*+</sup> | The list of commits to ignore during analysis. For accurate results, the commits should be provided with their full hash. Additionally, a range of commits can be specified using the `..` notation, e.g., `abc123..def456` (both inclusive).
 Ignore Authors List<sup>*+</sup> | The list of authors to ignore during analysis. Authors should be specified by their [Git Author Name](#a-note-about-git-author-name).
+Optimize Cloning | Enter **`yes`** to clone the repository using git's shallow cloning functionality. This option can significantly reduce the time taken to clone large repositories. However, the option should ideally be disabled for smaller repositories. Note: this is incompatible with the "Ignore standalone config" option.
```
damithc, line 41: Is this a breaking change? i.e., will this break config files in the old format? In that case, see if you can make it backward compatible.

vig42: > 
> 
> Is this a breaking change? i.e., will this break config files in the old format? In that case, see if you can make it backward compatible.

As it stands, it is a breaking change, because the CsvParser validates the repo-config.csv file by checking the number of columns. The expected number of columns in the csv file is now 8 rather than 7, so the CsvParser would throw an exception if there are only 7 columns.

We can make it backward compatible by loosening the restriction on the number of columns - we can set a minimum and maximum number of columns in the repo-config.csv file, so we would accept a csv with at least 7 and at most 8 columns. Since the new column is the rightmost, the position of the other columns in the repo-config.csv would all still be the same.

Alternatively, we could instead make CsvParser validate the csv file by checking the column headers. We could make it so that this particular column header is unnecessary, so it just sets a default value of "false". This would also handle cases where a user makes a config csv with the columns in a different order, since the CsvParser currently only cares about the position of the columns (e.g. "File Formats" is assumed to be in column 2) and ignores the headers entirely.

damithc: > Alternatively, we could instead make CsvParser validate the csv file by checking the column headers. We could make it so that this particular column header is unnecessary, so it just sets a default value of "false". This would also handle cases where a user makes a config csv with the columns in a different order, since the CsvParser currently only cares about the position of the columns (e.g. "File Formats" is assumed to be in column 2) and ignores the headers entirely.

I like this idea. In fact, making some columns optional and not caring about the order of columns would be even better. This may have to be a separate PR. But unless it is done, I'm afraid we can't merge this PR because it will break currently deployed dashboards that use the latest build of RepoSense every time the report is updated.

vig42:
> 
> 
> Thanks for the summary @vig42
> I wonder if the flag name is too general e.g., what if we introduce other ways to optimize cloning in future? Perhaps make it more specific? e.g., `--partial-cloning`?

I was worried about that too. I ended up making it general (`--optimize-cloning`) to make it simpler for users to understand, but I think your point is more relevant. I will change it to `--shallow-cloning`, since the main reason for the reduced size of the cloned repo is that it is a shallow clone.

docs/ug/configFiles.md
vig42: > 
> I like this idea. In fact, making some columns optional and not caring about the order of columns would be even better. This may have to be a separate PR. But unless it is done, I'm afraid we can't merge this PR because it will break currently deployed dashboards that use the latest build of RepoSense every time the report is updated.

Understood, I will make a separate PR with that change. It would also make it easier to add other configuration options going forward without breaking backward compatibility.

docs/ug/cli.md
vig42: > 
> 
> You might want to explain why it should not be used for small repos.

Noted, will add a point about the overhead.

>
>
> also, what's considered a 'small' repo?

The exact threshold isn't very clear, since the reduction in size due to shallow cloning varies depending on when the commits were made relative to the time window of the shallow clone. But I think it is safe to say that shallow cloning would not be very helpful if the size of the output of the bare clone (i.e. the ".git" file) is less than 1GB.

damithc: > The exact threshold isn't very clear, since the reduction in size due to shallow cloning varies depending on when the commits were made relative to the time window of the shallow clone. But I think it is safe to say that shallow cloning would not be very helpful if the size of the output of the bare clone (i.e. the ".git" file) is less than 1GB.

I see. You may want to mention this in the documentation as well.

fzdy1914 (changes requested):
Logic wise, it seems to be OK for me. However, can you add some tests to cover the newly added code?

docs/ug/configFiles.md
vig42: > 
> 
> > Alternatively, we could instead make CsvParser validate the csv file by checking the column headers. We could make it so that this particular column header is unnecessary, so it just sets a default value of "false". This would also handle cases where a user makes a config csv with the columns in a different order, since the CsvParser currently only cares about the position of the columns (e.g. "File Formats" is assumed to be in column 2) and ignores the headers entirely.
> 
> I like this idea. In fact, making some columns optional and not caring about the order of columns would be even better. This may have to be a separate PR. But unless it is done, I'm afraid we can't merge this PR because it will break currently deployed dashboards that use the latest build of RepoSense every time the report is updated.

Prof, I have implemented this change and ran into an issue while testing. Unfortunately, a bunch of the JUnit tests fail since the config files that are being used for testing themselves do not follow the column header formats exactly.

- AuthorConfigParserTest: test config files have the column header "Author's Email", while the user guide specifies that it should be "Author's Emails"
- GroupConfigParserTest: test config files have the column header "Group label" while the user guide specifies that it should be "Group Name"
- RepoConfigParserTest: test config files have the column header "Ignore Commits List" while the user guide specifies that it should be "Ignore Commit List"

These are relatively small discrepancies. Right now I am using case-insensitive string matching to check the headers, but we can account for these differences quite easily using regex instead. Otherwise, we can just change the headers to match the user guide.

However, my worry is that users may also be using config files that do not match the user guide specification, which would mean that we would still end up breaking backward compatibility.

damithc: > Prof, I have implemented this change and ran into an issue while testing. Unfortunately, a bunch of the JUnit tests fail since the config files that are being used for testing themselves do not follow the column header formats exactly.
> 
> * AuthorConfigParserTest: test config files have the column header "Author's Email", while the user guide specifies that it should be "Author's Emails"
> * GroupConfigParserTest: test config files have the column header "Group label" while the user guide specifies that it should be "Group Name"
> * RepoConfigParserTest: test config files have the column header "Ignore Commits List" while the user guide specifies that it should be "Ignore Commit List"
> 
> These are relatively small discrepancies. Right now I am using case-insensitive string matching to check the headers, but we can account for these differences quite easily using regex instead. Otherwise, we can just change the headers to match the user guide.
> 
> However, my worry is that users may also be using config files that do not match the user guide specification, which would mean that we would still end up breaking backward compatibility.

Let's make this a separate PR as it is a non-trivial change. Yes, this change itself can be a breaking change, but at least it will make it easier for future changes to be backward compatible.

vig42: > 
> Let's make this a separate PR as it is a non-trivial change. Yes, this change itself can be a breaking change, but at least it will make it easier for future changes to be backward compatible.

I have made the PR: #1459.

vig42:
> 
> 
> Logic wise, it seems to be OK for me. However, can you add some tests to cover the newly added code?

I have added JUnit tests (ArgsParserTest, RepoConfigParserTest, RepoConfigurationTest) as well as system tests.

> 
> 
> Thanks for the summary @vig42
> I wonder if the flag name is too general e.g., what if we introduce other ways to optimize cloning in future? Perhaps make it more specific? e.g., `--partial-cloning`?

I have adjusted the flag to `--shallow-cloning`.

> 
> I see. You may want to mention this in the documentation as well.

I have updated the documentation accordingly.

docs/ug/cli.md
```diff
@@ -107,6 +107,15 @@ The last modified dates will be in the same timezone specified with the `--timez
 
 <!-- ------------------------------------------------------------------------------------------------------ -->
 
+### `--shallow-cloning`, `-S`
```
Tejas2805, line 110: This page is arranged alphabetically as of now. So better to push it down accordingly.

docs/ug/cli.md
```diff
@@ -107,6 +107,15 @@ The last modified dates will be in the same timezone specified with the `--timez
...
+
+**`--shallow-cloning`**: Clones repositories using Git's shallow cloning functionality, which can significantly reduce the time taken to clone large repositories. However, the flag should not be used for smaller repositories where the .git file is smaller than 500 MB, as it would create overhead. Note: this flag is incompatible with the `--last-modified-date` flag.
```
Tejas2805, line 112: Instead of `Note:` better to use the `MarkBind info box` we use throughout the page.

docs/ug/configFiles.md
```diff
@@ -38,6 +38,7 @@ Ignore Glob List<sup>*+</sup> | The list of file path globs to ignore during ana
 Ignore standalone config | To ignore the standalone config file (if any) in the target repository, enter **`yes`**. If the cell is empty, the standalone config file in the repo (if any) will take precedence over configurations provided in the csv files.
 Ignore Commit List<sup>*+</sup> | The list of commits to ignore during analysis. For accurate results, the commits should be provided with their full hash. Additionally, a range of commits can be specified using the `..` notation, e.g., `abc123..def456` (both inclusive).
 Ignore Authors List<sup>*+</sup> | The list of authors to ignore during analysis. Authors should be specified by their [Git Author Name](#a-note-about-git-author-name).
+Shallow Cloning | Enter **`yes`** to clone the repository using Git's shallow cloning functionality. This option can significantly reduce the time taken to clone large repositories. However, the option should ideally be disabled for smaller repositories, where the .git file is smaller than 500 MB. Note: this is incompatible with the "--last-modified-date" CLI flag.
```
Tejas2805, line 41: Same suggestion for `Note` here

src/main/java/reposense/git/GitClone.java
```diff
@@ -119,4 +169,32 @@ public static void cloneFromBareAndUpdateBranch(Path rootPath, RepoConfiguration
...
+     * into the folder {@code outputFolderName}.
+     */
+    private static String getCloneShallowBareCommand(RepoConfiguration config,
+                                                    String outputFolderName, Date shallowSinceDate) {
+        return "git clone --bare --shallow-since=\"" + shallowSinceDate + "\" "
+                + config.getLocation() + " " + outputFolderName;
+    }
+
+    /**
+     * Constructs the command to clone a bare repo specified in the {@code config}
```
Tejas2805, line 184: You have two functions with a duplicate comment header. Better to modify to show what each function exactly does. Will be easier for other devs to understand.

src/main/java/reposense/git/GitClone.java
```diff
@@ -119,4 +169,32 @@ public static void cloneFromBareAndUpdateBranch(Path rootPath, RepoConfiguration
...
+     * into the folder {@code outputFolderName}.
+     */
+    private static String getClonePartialBareCommand(RepoConfiguration config, String outputFolderName) {
+        return "git clone --bare --filter=blob:none " + config.getLocation() + " " + outputFolderName;
+    }
+
+    /**
+     * Constructs the command to clone a bare repo specified in the {@code config}
```
Tejas2805, line 192: Same suggestion for this.

Tejas2805 (changes requested):
Left comments regarding some small changes. Overall, great work done!

Just a doubt - What happens if both the CLI arguments are passed - S and last-modified-date?

vig42:
> 
> 
> Left comments regarding some small changes. Overall, great work done!

Have made all the suggested changes.

> 
> Just a doubt - What happens if both the CLI arguments are passed - S and last-modified-date?

Right now it just puts the incorrect "last modified" dates in the JSON. My assumption is that users would read the docs and avoid doing that.

We could make the CLI arguments mutually exclusive so we would not face that conflict. However, we would still face an issue if the "Shallow Clone" option is enabled in the repo config file. One way to handle that could be to modify the FileInfoAnalyzer part here:

https://github.com/reposense/RepoSense/blob/fc288b461f87a14fc7004c0503fe4adbe12360b4/src/main/java/reposense/authorship/FileInfoAnalyzer.java#L106-L112

We could change line 106 to:

```java
if (config.isLastModifiedDateIncluded() && !config.isShallowCloningPerformed()) 
```

So for any repo where the Shallow Cloning config option is enabled, RepoSense would not put the last modified date in the JSON (we could also give the users a warning log if this happens).

Tejas2805:
> Right now it just puts the incorrect "last modified" dates in the JSON. My assumption is that users would read the docs and avoid doing that.

It is possible that the user wants to shallow clone one repo out of say 5. And wants the last-modified-date for all (though we won't get for the one with shallow clone obviously. Therefore, better to generate a warning. 

> So for any repo where the Shallow Cloning config option is enabled, RepoSense would not put the last modified date in the JSON. We could also give the users a warning log if this happens.

I think it is better to give the warning log from our side in case both shallow-cloning and last-modified-date are given? Saying that the last modified dates for this repo will be incorrect or invalid since shallow cloning has been enabled. @damithc your thoughts?

damithc:
> I think it is better to give the warning log from our side in case both shallow-cloning and last-modified-date are given? Saying that the last modified dates for this repo will be incorrect or invalid since shallow cloning has been enabled. @damithc your thoughts?

Sure, we can give a warning for that.

vig42:
> 
> 
> > Right now it just puts the incorrect "last modified" dates in the JSON. My assumption is that users would read the docs and avoid doing that.
> 
> It is possible that the user wants to shallow clone one repo out of say 5. And wants the last-modified-date for all (though we won't get for the one with shallow clone obviously. Therefore, better to generate a warning.
> 
> > So for any repo where the Shallow Cloning config option is enabled, RepoSense would not put the last modified date in the JSON. We could also give the users a warning log if this happens.
> 
> I think it is better to give the warning log from our side in case both shallow-cloning and last-modified-date are given? Saying that the last modified dates for this repo will be incorrect or invalid since shallow cloning has been enabled. @damithc your thoughts?

I have added this warning. I have also made the CLI parameters for `--shallow-cloning` and `--last-modified-date` mutually exclusive.

Tejas2805 (commented):
Looks okay to me. I will need some time to go through the whole code in detail.

docs/ug/configFiles.md
```diff
@@ -38,6 +38,11 @@ Ignore Glob List<sup>*+</sup> | The list of file path globs to ignore during ana
 Ignore standalone config | To ignore the standalone config file (if any) in the target repository, enter **`yes`**. If the cell is empty, the standalone config file in the repo (if any) will take precedence over configurations provided in the csv files.
 Ignore Commit List<sup>*+</sup> | The list of commits to ignore during analysis. For accurate results, the commits should be provided with their full hash. Additionally, a range of commits can be specified using the `..` notation, e.g., `abc123..def456` (both inclusive).
 Ignore Authors List<sup>*+</sup> | The list of authors to ignore during analysis. Authors should be specified by their [Git Author Name](#a-note-about-git-author-name).
+Shallow Cloning | Enter **`yes`** to clone the repository using Git's shallow cloning functionality. This option can significantly reduce the time taken to clone large repositories. However, the option should ideally be disabled for smaller repositories where the .git file is smaller than 500 MB, as it would create overhead.
```
fzdy1914, line 41: Quote .git as `.git` would be better.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -104,6 +107,10 @@ private static void aggregateBlameAuthorModifiedAndDateInfo(RepoConfiguration co
             }
 
             if (config.isLastModifiedDateIncluded()) {
+                if (config.isShallowCloningPerformed()) {
```
fzdy1914, line 110: If the two options already mutually exclusive by the argparser, will it still be possible to go to the branch here?

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -78,9 +80,18 @@ protected void processLine(List<RepoConfiguration> results, CSVRecord record) th
                     "Ignoring unknown value " + ignoreStandaloneConfig + " in ignore standalone config column.");
         }
 
+        String shallowCloningConfig = get(record, SHALLOW_CLONING_CONFIG_POSITION);
```
fzdy1914, line 83: If I understand correctly, in your current design, users cannot choose the depth of shallow cloning?

fzdy1914: We may extend to this shallow cloning argument in a nicer latest-n-commit command if we allow customizing the depth right? Correct me if I am wrong.

fzdy1914 (commented):
Is this PR conflicting with the multithreaded one? If so, which PR would you like for me to merge first?

Are we lacking system tests for shallow cloning so there are a lot of warning for uncovered line?

src/main/java/reposense/authorship/FileInfoAnalyzer.java
vig42: It is possible, since the shallow cloning can also be enabled as an option in the repo-config.csv. So as @Tejas2805 had mentioned above, it is possible for a user to specify 1 repo out of 5 to be shallow cloned while also using the `--last-modified-date` flag.

src/main/java/reposense/parser/RepoConfigCsvParser.java
vig42: > 
> 
> If I understand correctly, in your current design, users cannot choose the depth of shallow cloning?

Yes, that is correct. It is designed to replicate the report normally generated by RepoSense, so it just pulls the minimum amount of history possible while still capturing all commits between the "since" and "until" date.

> 
> 
> We may extend to this shallow cloning argument in a nicer latest-n-commit command if we allow customizing the depth right? Correct me if I am wrong.

I feel that this would be a separate feature altogether. This PR is aimed at speeding up the cloning process for large repos by reducing the size of the cloned repo. From the user's POV, the generated report would be identical (well, besides the incompatibility with `--last-modified-date`).

If I understand you correctly, the last-n-commit command would generate a report containing info about the last N commits only, i.e. it would be an alternative to the "since" and "until" dates. We can implement this for the commits report by modifying `GitLog.getWithFiles()` to look at the last N commits rather than the date range. As for the authorship report, we can implement this by just using the N-th latest commit for the `lastCommitHash`.

This means we wouldn't need to use shallow cloning, meaning we can implement the last-n-commits feature while maintaining compatibility with the `--last-modified-date` flag.

docs/ug/configFiles.md
vig42: I have made this change.

vig42:
> 
> 
> Is this PR conflicting with the multithreaded one? If so, which PR would you like for me to merge first?

There are some minor conflicts but they are quite trivial to resolve. They are mainly in the Location/ConfigCliArguments and the ArgsParser, since both PRs make changes to those files. Basically just need to make sure the parameters added by each of the 2 PRs are both included in the merged commit. It shouldn't make a difference which PR is merged first.

> 
> Are we lacking system tests for shallow cloning so there are a lot of warning for uncovered line?

I had added 2 system tests that replicate the existing system tests, except with shallow cloning enabled: `testSinceBeginningDateRangeWithShallowCloning` and `test30DaysFromUntilDateWithShallowCloning`. I feel that this should be sufficient to cover the shallow cloning code in `RepoCloner.java`, so I am not sure why the lines are listed as uncovered.

src/main/java/reposense/parser/RepoConfigCsvParser.java
fzdy1914: Sorry, my bad, I previously thought you will only clone depth 1 only for analyze. It is fine in such a case.

fzdy1914:
Can you resolve the conflicts and then request a review from the stage-1 active reviewer? Thanks.

src/main/java/reposense/git/GitCatFile.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Returns parent commits for the commit associated with the input commit hash.
+     */
+    public static List<String> getParentCommits(String root, String commitHash) throws CommitNotFoundException {
+        Path rootPath = Paths.get(root);
+        String catFileCommand = "git cat-file -p " + commitHash;
+        try {
+            String output = runCommand(rootPath, catFileCommand);
+            List<String> parents = new ArrayList();
```
Tejas2805, line 30: Can name `parentCommits` instead of `parents`?

src/main/java/reposense/git/GitCatFile.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * Given the input list of commit hashes, return a list of commit hashes for the parent
+     * commits of all the commits.
+     */
+    public static List<String> getParentsOfCommits(String root, List<String> commitHashes) {
+        List<String> parents = new ArrayList();
+        for (String commitHash : commitHashes) {
+            try {
+                parents.addAll(getParentCommits(root, commitHash));
```
Tejas2805, line 50: Same here

src/main/java/reposense/git/GitRevList.java
```diff
@@ -98,4 +100,24 @@ private static String getAllCommitHashSince(String root, String branchName, Stri
...
+        String revListCommand = "git rev-list --max-parents=0 HEAD";
+        Path rootPath = Paths.get(root);
+        String output = runCommand(rootPath, revListCommand);
+        return Arrays.asList(output.split("\n"));
+    }
+
+    /**
+     * Returns true if the repository is empty.
+     */
+    public static boolean getIsEmptyRepo(String root) {
```
Tejas2805, line 117: Doesn't sound like a boolean returning function with the `get`.  Maybe `check` or `is` will make it sound better.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -159,6 +159,15 @@ private static ArgumentParser getArgumentParser() {
                         + "If not provided, default file formats will be used.\n"
                         + "Please refer to userguide for more information.");
 
+        parser.addArgument(TIMEZONE_FLAGS)
```
Tejas2805, line 162: Any reason for shifting it here? Or just arrangement?

src/test/java/reposense/git/GitCatFileTest.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Test
+    public void getParentCommits_rootCommit_success() throws Exception {
+        List<String> parentsList = GitCatFile.getParentCommits(config.getRepoRoot(), ROOT_COMMIT_HASH);
+        Assert.assertEquals(0, parentsList.size());
+    }
+
+    @Test(expected = CommitNotFoundException.class)
+    public void getParentCommits_nonexistentCommit_throwsEmptyCommitException() throws Exception {
```
Tejas2805, line 28:
```suggestion
    public void getParentCommits_nonExistentCommit_throwsEmptyCommitException() throws Exception {
```

src/test/java/reposense/git/GitShowTest.java
```diff
@@ -0,0 +1,48 @@
...
+
+    @Test
+    public void getCommitDate_normalCommit_success() throws Exception {
+        Date commitDate = GitShow.getCommitDate(config.getRepoRoot(), TEST_COMMIT_HASH);
+        Date expectedDate = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss Z").parse("2018-02-09 22:17:39 +0800");
+        Assert.assertEquals(expectedDate, commitDate);
+    }
+
+    @Test(expected = CommitNotFoundException.class)
+    public void getCommitDate_nonexistentCommit_throwsEmptyCommitException() throws Exception {
```
Tejas2805, line 23:
```suggestion
    public void getCommitDate_nonExistentCommit_throwsEmptyCommitException() throws Exception {
```

Tejas2805 (commented):
Left some comments. The code and flow LGTM!

src/main/java/reposense/parser/ArgsParser.java
vig42: This was originally at the bottom after the mutex flags. When I added in the 2 new mutex flags, I moved this above all the mutex flags so it would be a bit clearer.

src/main/java/reposense/git/GitRevList.java
vig42: I have changed it to `checkIsEmptyRepo`

src/main/java/reposense/git/GitCatFile.java
vig42: I have changed this.

src/main/java/reposense/git/GitCatFile.java
vig42: I have changed this.

src/test/java/reposense/git/GitCatFileTest.java
vig42: I have changed this.

src/test/java/reposense/git/GitShowTest.java
vig42: I have changed this.

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -47,7 +63,41 @@
...
+            if (!didPartialCloneSucceed) {
+                isCurrentRepoCloned = spawnCloneProcess(config);
+                return;
+            }
+
+            String partialBareRoot = FileUtil.getPartialBareRepoPath(config).toString();
+            Date sinceDate;
+            try {
+                List<String> distinctParents = graftedCommitParents.stream().distinct().collect(Collectors.toList());
+                sinceDate = GitShow.getEarliestCommitDate(partialBareRoot, distinctParents);
```
fzdy1914, line 91: Just a question, will just check the date one by one be very time-consuming? Or the number of parents will be very limited given that the date range given is not so much?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -64,7 +66,7 @@ public static void zipFoldersAndFiles(List<Path> pathsToZip, Path sourceAndOutpu
      * {@code sourcePath} into {@code outputPath}.
      */
     public static void zipFoldersAndFiles(List<Path> pathsToZip,
-            Path sourcePath, Path outputPath, String... fileTypes) {
+                                          Path sourcePath, Path outputPath, String... fileTypes) {
```
fzdy1914, line 69: Just a question, does it violate our code standard? I kind of forgot.

src/main/java/reposense/parser/ArgsParser.java
fzdy1914: Yes, it will be clearer now.

fzdy1914 (approved):
Mostly OK for me by just checking the code logic online.

src/main/java/reposense/report/RepoCloner.java
vig42: The number of parents is quite limited. For the cases I have manually checked, the number of grafted commits has been < 5. So even if all of them are merge commits with 2 parents, the number of parents to check is at most 10.

src/main/java/reposense/util/FileUtil.java
vig42: Oh, good catch. I have fixed the indentation. Not sure why it got changed in the first place.

docs/ug/cli.md
```diff
@@ -107,7 +107,11 @@ This flag overrides the `Ignore standalone config` field in the CSV config file.
 * Alias: `-l` (lowercase L)
 * Example:`--last-modified-date` or `-l`
 
-The last modified dates will be in the same timezone specified with the `--timezone` flag.
+<box type="info" seamless>
+
+* Cannot be used with ``--shallow-cloning`.
```
Tejas2805, line 112: There are two `` in this line. Needs to be fixed.

docs/ug/cli.md
```diff
@@ -150,6 +154,20 @@ Cannot be used with `--config`.
...
+### `--shallow-cloning`, `-S`
+
+**`--shallow-cloning`**: Clones repositories using Git's shallow cloning functionality, which can significantly reduce the time taken to clone large repositories. However, the flag should not be used for smaller repositories where the `.git` file is smaller than 500 MB, as it would create overhead.
+* Default: RepoSense does not clone repositories using Git's shallow cloning functionality.
+* Alias: `-S` (uppercase S)
+* Example:`--shallow-cloning` or `-S`
+
+<box type="info" seamless>
+
+Cannot be used with ``--last-modified-date`.
```
Tejas2805, line 166: Same here.

Tejas2805 (changes requested):
LGTM. Last 2 errors need to be fixed.

docs/ug/cli.md
vig42: Fixed

docs/ug/cli.md
vig42: Fixed

Tejas2805 (approved):
LGTM!

PR #1451, opened 2021-02-11
[#1452] Implement multi-threading

### LINKED ISSUE #1452: Implement multi-threading

Currently, RepoSense performs cloning and analysis in parallel; i.e. while Repo 1 is being analyzed, Repo 2 is being cloned. When we are running RepoSense on a large number of repos, we could achieve better performance if we clone and analyse more repos simultaneously using multi-threading.

We could implement multi-threading using Java’s ExecutorService and thread pooling. Each thread could do 3 things:

1.	Clone a single repo
2.	Perform analysis on that repo and generate the outputs 
3.	Delete all the data associated with that repo

Each repo has its own distinct folder to store the cloned repo data and a distinct folder to store the generated outputs (the commits and authorship JSON files). Since each thread handles 1 repo, there should be no conflicts with the file system between the threads. Furthermore, as there is no shared data modified by the threads, there should be no issue of race conditions.

**Performance improvement**

I made a rough implementation of this in #1451 as a proof of concept to test out the performance.

I tested it with 2 of the existing demos on the RepoSense websites: the CS2113 AY2019/20 Sem 2 individual project (72 repos) and team project (17 repos). With 4 threads, the time taken is cut down significantly.

For the 17 group project repos, the time taken to generate the report on my local machine went down from around 2 min 40 sec to 1 min 20 sec. For the 72 individual project repos, the time taken to generate the report went down from around 5 min 45 sec to 2 min 30 sec. This means the time taken is reduced by about 50-55% in each case.

Prof Damith had previously mentioned in [this comment](https://github.com/reposense/RepoSense/issues/540#issuecomment-460142643) on issue #540 that threading should not be considered unless there is significant time saving of 20%+, due to issues with debugging. However I feel that the potential savings of around 50% may make this viable.

vig42:
Fixes #1452

```
Currently, RepoSense runs cloning and analysis of 1 repo each in
parallel.

Let’s implement multi-threading to increase the number of repos being
analyzed in parallel, which can speed up the report generation when
running RepoSense on a large number of repos.
```


src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -173,38 +180,52 @@ private static void setLandingPageTitle(String filePath, String pageTitle) throw
...
-            if (nextRepoLocation != null) {
-                repoCloner.cloneBare(repoLocationMap.get(nextRepoLocation).get(0));
-            }
-
-            if (clonedRepoLocation == null) {
-                handleCloningFailed(configs, currRepoLocation);
-            } else {
-                generatedFiles.addAll(analyzeRepos(outputPath, configs, repoLocationMap.get(clonedRepoLocation),
-                        repoCloner.getCurrentRepoDefaultBranch()));
+        ExecutorService executor = Executors.newFixedThreadPool(4);
```
dcshzj, line 191: Would it be more ideal to adapt this number to match the number of available threads that the user's machine has? Something along the lines of:

```java
int threads = Runtime.getRuntime().availableProcessors();
ExecutorService executor = Executors.newFixedThreadPool(threads);
```

We should also impose an upper limit so that we don't open too many Git clones at once coming from a single host, so that we do not accidentally trigger any auto throttling by GitHub.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -216,7 +237,7 @@ private static void setLandingPageTitle(String filePath, String pageTitle) throw
      * @return A list of paths to the JSON report files generated for the repositories in {@code configsToAnalyze}.
      */
     private static List<Path> analyzeRepos(String outputPath, List<RepoConfiguration> configs,
-            List<RepoConfiguration> configsToAnalyze, String defaultBranch) {
+                                           List<RepoConfiguration> configsToAnalyze, String defaultBranch) {
```
dcshzj, line 240: I think the indentation is incorrect.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -359,7 +380,7 @@ private static void handleCloningFailed(List<RepoConfiguration> configs, RepoLoc
      * removes {@code failedConfig} from the list of {@code configs}.
      */
     private static void handleAnalysisFailed(List<RepoConfiguration> configs, RepoConfiguration failedConfig,
-            String errorMessage) {
+                                             String errorMessage) {
```
dcshzj, line 383: I think the indentation is incorrect.

dcshzj (changes requested):

src/main/java/reposense/report/ReportGenerator.java
Tejas2805: Agree with @dcshzj suggestion.

src/main/java/reposense/report/ReportGenerator.java
gerhean: Is it possible to add a checkstyle rule to ensure this indentation rule is followed instead of manually warning about it in the reviews?

src/main/java/reposense/report/ReportGenerator.java
vig42: I have fixed the indentation here.

src/main/java/reposense/report/ReportGenerator.java
vig42: I have fixed the indentation here.

src/main/java/reposense/report/ReportGenerator.java
vig42: > 
> 
> Would it be more ideal to adapt this number to match the number of available threads that the user's machine has? Something along the lines of:
> 
> ```java
> int threads = Runtime.getRuntime().availableProcessors();
> ExecutorService executor = Executors.newFixedThreadPool(threads);
> ```
Agreed, this makes more sense.

I tried this out; it seems that if the CPU supports hyper-theading, the number of "available processors" is doubled, but the performance increase is more marginal. On my PC, with 8 threads, it took 1 min 8 sec to generate the report for the CS2113 AY2019/20 Sem 2 group project (17 repos). This is a slight improvement over the 1 min 20 sec taken with 4 threads. For reference, vanilla RepoSense takes about 2 min 40 sec.

>
> 
> We should also impose an upper limit so that we don't open too many Git clones at once coming from a single host, so that we do not accidentally trigger any auto throttling by GitHub.

According to [their docs](https://docs.github.com/en/github/getting-started-with-github/troubleshooting-connectivity-problems#troubleshooting-slow-downloads-and-intermittent-slow-connections), GitHub does not throttle bandwidth per user. But I think we could still end up with bandwidth issues with many simultaneous connections. How many connections do you think is a suitable upper limit?

src/main/java/reposense/report/ReportGenerator.java
gerhean: I believe it is better to open a separate PR for this

rolandyuwy: As pointed out, enforcing this checkstyle rule will violate the coding standard of `Indentation for wrapped lines should be 8 spaces (i.e. twice the normal indentation of 4 spaces) more than the parent line.` I think it's best to leave it for now and focus on non trivial aspects of this issue.

src/main/java/reposense/report/ReportGenerator.java
dcshzj: Sorry I missed this, I don't have an exact figure to suggest but I think 4 threads should be a good maximum.

damithc:
@vig42 So, what's the verdict on costs, benefits?

vig42:
As mentioned on the [related issue](https://github.com/reposense/RepoSense/issues/1452#issuecomment-782841443), I have parallelized the cloning and analysis steps separately, using Java's CompletableFuture class. I kept 4 threads available for cloning and set the number of threads for analysis to equal the number of CPU threads.

> 
> 
> @vig42 So, what's the verdict on costs, benefits?

The main benefit is that the time taken to generate the report is reduced when running RepoSense on a large number of repos. e.g. for the CS2113 AY2019/20 Sem 2 individual project demo (which consists of 72 repos), it took about 1 min 40 sec to generate the report on my local machine, relative to around 5 min with the master branch, i.e. an improvement of over 65%. The improvement was less dramatic when testing it out on Github Actions; it went down from around 25 seconds to around 20 seconds, i.e. improvement of about 20%.

One obvious reason for the improvement is because we are making use of more CPU cores. This is most clearly seen on my local machine, which has a quad-core CPU with hyper-threading, so effectively there are 8 cores available; in comparison, Github Actions or Travis CI only allocate us 2 cores.

Besides that, as @gerhean had mentioned, the bottleneck resource is different for cloning (IO-heavy) and analysis (CPU-heavy). Currently, RepoSense clones repos in parallel to analysis, i.e. while repo 1 is being analyzed, repo 2 is being cloned. However, once cloning of repo 2 is finished, cloning of repo 3 will not begin until analysis of repo 1 is done. With multi-threading enabled, we can now ensure that the IO-heavy cloning process will not have to wait for a CPU-heavy analysis process to finish. Additionally, even with 2 cores, we would still see improvement since we can run 2 CPU-heavy analysis jobs simultaneously.

The main drawback of implementing multi-threading is the potential bugs it could introduce and the difficulty in reproducing the bugs, e.g. in case of race conditions.

I feel that we should not face race conditions as the only objects shared between threads are the `repoLocationMap` and the `outputPath`, which are both not modified. Besides that, the `RepoConfiguration` objects are modified by the threads, but since each `RepoConfiguration` object is associated with 1 repo, and since each repo is handled by a single thread, we would never have multiple threads attempting to access the same data. If we want to be even more careful, rather than modifying the `RepoConfiguration` objects, we can create new instances with the necessary changes and return them as outputs from the threads.

Additionally, the disk IO should not cause any conflicts since both the cloned data and the output JSON files are stored in separate folders for each repo. Since there is 1 thread for each repo, we should not have any conflicts where 2 threads are attempting to write to the same folder.

damithc:
@vig42 thanks for the summary. Some follow up questions:

1. What's the impact of threading when running on a CI environment such as GitHub actions?
2. Is there an easy way to go back to a single thread mode (e.g., by setting thread count to 1), for troubleshooting purposes (e.g., to figure out if a bug is caused specifically due to multi-threading)?

vig42:
> 
>     1. What's the impact of threading when running on a CI environment such as GitHub actions?

I tried testing a few times with the 2 demos showcased on the RepoSense website, i.e. the CS2113 AY2019/20 individual and group projects. It seems that the time taken varies significantly, both with or without threading.

e.g. for the individual project, with threading enabled, on 3 separate tries it took 19.83 sec, 8.25 sec and 11.71 sec. Without threading, on 3 tries it took 24.82 sec, 25.86 sec and 66.83 sec. So while it was consistently faster with threading, I can't really say how much faster. I only had 1 Github Action running at a time, so I'm really not sure why some tries took much longer than others.

Given the massive variance, I am not sure how to make a fair comparison. Should I just run the test with the same configs for 50-100 times, and take the average time?

> 
>     2. Is there an easy way to go back to a single thread mode (e.g., by setting thread count to 1), for troubleshooting purposes (e.g., to figure out if a bug is caused specifically due to multi-threading)?

There are 2 ways we can do this. First way is to set the thread count to 1 when initializing the `ExecutorService` objects; this would mean that there is max 1 thread doing cloning and 1 thread doing analysis at any given time. Second way is simply to reuse the old function to clone and analyze repos sequentially.

We could add a flag, e.g. `--debug`, to trigger this behaviour (either of the 2 options).

damithc:
> I tried testing a few times with the 2 demos showcased on the RepoSense website, i.e. the CS2113 AY2019/20 individual and group projects. It seems that the time taken varies significantly, both with or without threading.

I think it is still good to know if one is consistently faster than the other. And yes, you can do a higher number of tests to find a median/mean.


> There are 2 ways we can do this. First way is to set the thread count to 1 when initializing the `ExecutorService` objects; this would mean that there is max 1 thread doing cloning and 1 thread doing analysis at any given time. Second way is simply to reuse the old function to clone and analyze repos sequentially.
> We could add a flag, e.g. `--debug`, to trigger this behaviour (either of the 2 options).

I don't think it makes sense to make it a flag or maintain the old code if one is to be used just for debugging. That old code itself can break without us noticing as it will not be covered by tests. However, you can make the thread count easy to configure e.g., set it as a constant and document how to change it (in developer documentation?)

fzdy1914:
> I don't think it makes sense to make it a flag or maintain the old code if one is to be used just for debugging. That old code itself can break without us noticing as it will not be covered by tests. However, you can make the thread count easy to configure e.g., set it as a constant and document how to change it (in developer documentation?)

Maybe we can have a parameter for users to choose the number of threads to use in cmd?

vig42:
> 
> 
> > I tried testing a few times with the 2 demos showcased on the RepoSense website, i.e. the CS2113 AY2019/20 individual and group projects. It seems that the time taken varies significantly, both with or without threading.
> 
> I think it is still good to know if one is consistently faster than the other. And yes, you can do a higher number of tests to find a median/mean.

I ran the 2 showcase demos on the RepoSense website on Github Actions, with and without threading, 50 times each. These are the results:

Name | Threading? | Mean Time | Median Time | Std Dev
-- | -- | -- | -- | --
Individual Project | Yes| 14.08 | 14.07 | 5.29
Individual Project | No | 35.90 | 34.70 | 12.38
Group Project | Yes | 8.30 | 7.84 | 2.64
Group Project | No | 17.44 | 18 | 6.73

So for the individual project (with 72 repos), on average there is a 60% improvement with threading, while for the group project (with 17 repos) on average there is a 52% improvement.

>
> I don't think it makes sense to make it a flag or maintain the old code if one is to be used just for debugging. That old code itself can break without us noticing as it will not be covered by tests. However, you can make the thread count easy to configure e.g., set it as a constant and document how to change it (in developer documentation?)

I have fixed the number of threads for cloning to be 4, so that can be easily converted to a constant. However, as @dcshzj had suggested, the number of threads for analysis is set to match the number of processor cores are available. In order to override this, we could have a constant "thread cap":

```java
private static final int ANALYSIS_MAX_THREADS = 8;

...

int threads = Runtime.getRuntime().availableProcessors();
ExecutorService analyzeExecutor = Executors.newFixedThreadPool(Math.min(threads, ANALYSIS_MAX_THREADS));
```

This way, if we set the ANALYSIS_MAX_THREADS constant to be 1, then we would effectively disable multi-threading. Would that be sufficient?

> 
> Maybe we can have a parameter for users to choose the number of threads to use in cmd?

This is also possible. Do you think we should provide 2 separate parameters for the number of threads for cloning vs analysis?

damithc:
> This way, if we set the ANALYSIS_MAX_THREADS constant to be 1, then we would effectively disable multi-threading. Would that be sufficient?

I think so.

> This is also possible. Do you think we should provide 2 separate parameters for the number of threads for cloning vs analysis?

I'm not sure whether we want to tell users about these internal things (i.e., cloning vs analysis). Perhaps let the user set the thread count and recommend not to set it higher than the core count?

vig42:
> 
> 
> > This way, if we set the ANALYSIS_MAX_THREADS constant to be 1, then we would effectively disable multi-threading. Would that be sufficient?
> 
> I think so.
> 

I have done this, and added this in the javadoc for the cloneAndAnalyzeRepos function

```java
/**
 * By default, runs in multi-threaded mode. To turn off multi-threading, set the constants
 * {@code NUM_THREADS_CLONING} and {@code NUM_THREADS_ANALYSIS_MAX} to 1.
 */
```

fzdy1914:
> > This way, if we set the ANALYSIS_MAX_THREADS constant to be 1, then we would effectively disable multi-threading. Would that be sufficient?
> 
> I think so.
> 
> > This is also possible. Do you think we should provide 2 separate parameters for the number of threads for cloning vs analysis?
> 
> I'm not sure whether we want to tell users about these internal things (i.e., cloning vs analysis). Perhaps let the user set the thread count and recommend not to set it higher than the core count?

fzdy1914:
@vig42 
> > This is also possible. Do you think we should provide 2 separate parameters for the number of threads for cloning vs analysis?
> 
> I'm not sure whether we want to tell users about these internal things (i.e., cloning vs analysis). Perhaps let the user set the thread count and recommend not to set it higher than the core count?

Or maybe we can have one command for users to choose to utilize multi-threading or not? I.e. Default we use multi-threaded process. If the user provide with a flag, then, we run in non-multithread mode?

damithc:
> Or maybe we can have one command for users to choose to utilize multi-threading or not? I.e. Default we use multi-threaded process. If the user provide with a flag, then, we run in non-multithread mode?
> 
> So no need for users to know the exact number of threads they have used.

Any reason why the user should use single-threaded? If the multi-threaded is not buggy and faster, there is no need to burden the user with this choice, right?

fzdy1914:
> > Or maybe we can have one command for users to choose to utilize multi-threading or not? I.e. Default we use multi-threaded process. If the user provide with a flag, then, we run in non-multithread mode?
> > So no need for users to know the exact number of threads they have used.
> 
> Any reason why the user should use single-threaded? If the multi-threaded is not buggy and faster, there is no need to burden the user with this choice, right?

That command may be used for debugging purposes? So, we do not need to manually set the thread to be 1 when testing. But it is OK to always have multi-threaded enabled.

damithc:
> That command may be used for debugging purposes? So, we do not need to manually set the thread to be 1 when testing. But it is OK to always have multi-threaded enabled.

OK, but we can keep it as an internal flag. i.e., only in the DG but not in the UG.

vig42:
> 
> 
> > That command may be used for debugging purposes? So, we do not need to manually set the thread to be 1 when testing. But it is OK to always have multi-threaded enabled.
> 
> OK, but we can keep it as an internal flag. i.e., only in the DG but not in the UG.

Do you think we would need to allow the devs to specify the exact number of threads (e.g. putting the flag `--threads 2` to use 2 threads), or would it be sufficient to just add a flag `--single-threaded` to disable multi-threading? Also, if we let them specify the number of threads, should that apply to both the cloning and analysis steps? i.e. for the flag `--threads 8`, should we set 8 threads for cloning and 8 threads for analysis?

damithc:
> Do you think we would need to allow the devs to specify the exact number of threads (e.g. putting the flag `--threads 2` to use 2 threads), or would it be sufficient to just add a flag `--single-threaded` to disable multi-threading? 

Allowing to specify the exact number is more flexible, if the implementation is not more complex.

>Also, if we let them specify the number of threads, should that apply to both the cloning and analysis steps? i.e. for the flag `--threads 8`, should we set 8 threads for cloning and 8 threads for analysis?

If there is a use case for splitting, go ahead.

vig42:
> 
> > Also, if we let them specify the number of threads, should that apply to both the cloning and analysis steps? i.e. for the flag `--threads 8`, should we set 8 threads for cloning and 8 threads for analysis?
> 
> If there is a use case for splitting, go ahead.
Since these flags would only be used by devs, the use case would be to test the performance with varying numbers of threads for cloning and analysis. So e.g. to test whether 2 threads for cloning is better than 4, when running on a machine with bad internet connection.

vig42:
I have implemented the 2 hidden flags: `--cloning-threads` and `--analysis-threads`. The PR should be ready for review.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -182,6 +188,18 @@ private static ArgumentParser getArgumentParser() {
                         + "One kind of valid timezones is relative to UTC. E.g. UTC, UTC+08, UTC-1030. \n"
                         + "If not provided, system default timezone will be used.");
 
+        parser.addArgument(CLONING_THREADS_FLAG)
+                .dest(CLONING_THREADS_FLAG[0])
+                .type(int.class)
```
dcshzj, line 193: Trivial, but I think we should create a `CloningThreadsArgumentType` instead of just using `int`.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -182,6 +188,18 @@ private static ArgumentParser getArgumentParser() {
...
+                .setDefault(DEFAULT_NUM_CLONING_THREADS)
+                .help(FeatureControl.SUPPRESS);
+
+        parser.addArgument(ANALYSIS_THREADS_FLAG)
+                .dest(ANALYSIS_THREADS_FLAG[0])
+                .type(int.class)
```
dcshzj, line 199: Same here for `AnalysisThreadsArgumentType`.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -173,96 +177,147 @@ private static void setLandingPageTitle(String filePath, String pageTitle) throw
...
-                StringWriter sw = new StringWriter();
-                e.printStackTrace(new PrintWriter(sw));
-                logger.log(Level.SEVERE, sw.toString());
-                handleAnalysisFailed(configs, configToAnalyze,
-                        String.format(LOG_UNEXPECTED_ERROR, configToAnalyze.getLocation(), sw.toString()));
+        List<AnalysisErrorInfo> analysisErrors = new ArrayList<>();
+        if (cloneSuccessful) {
+            Iterator<RepoConfiguration> itr = configsToAnalyze.iterator();
+            while (itr.hasNext()) {
+                progressTracker.incrementProgress();
```
dcshzj, line 275: One question: Will this increment be multithread safe?

dcshzj (changes requested):
Any reason why the ConfigSystemTest is failing? There seems to be some `NumberFormatException` happening in `testSinceBeginningDateRange`.

> OK, but we can keep it as an internal flag. i.e., only in the DG but not in the UG.

As Prof mentioned, we should add some information in the DG about the presence of these two flags in this PR.

src/main/java/reposense/report/ReportGenerator.java
vig42: Yes, this should be thread safe. The iterator is operating on the list `configsToAnalyze`. This list is obtained from the `repoLocationMap` for a particular `location`:

```java
List<RepoConfiguration> configsToAnalyze = repoLocationMap.get(location);
```

Since each `location` is handled by a separate thread, logically, we would never have 2 different threads trying to access the same list.

vig42:
> 
> > OK, but we can keep it as an internal flag. i.e., only in the DG but not in the UG.
> 
> As Prof mentioned, we should add some information in the DG about the presence of these two flags in this PR.

I have added this to the DG under `architecture.md`.

> 
> 
> Any reason why the ConfigSystemTest is failing? There seems to be some `NumberFormatException` happening in `testSinceBeginningDateRange`.

I'm really not sure about this. For one thing, the error only occurred on Ubuntu. I ran the system tests several times on my machine (Windows 10), and could not reproduce the error, so I figured maybe the platform might be the issue.

But now, after pushing the changes to the docs, it seems that the system tests have all passed, even on Linux. So I'm not sure what to make of it...maybe a random bug with the CI?

docs/dg/architecture.md
```diff
@@ -65,9 +65,12 @@
 ## ReportGenerator(Main)
 
 [`ReportGenerator`](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/report/ReportGenerator.java)
- 1. uses `GitClone` API to clone the repository from *GitHub*.
- 1. copies the template files into the designated output directory.
- 1. uses `CommitReporter` and `AuthorshipReporter` to produce the commit and authorship summary, respectively.
+ 1. clones repositories from *GitHub* using the `GitClone` API, in a multi-threaded fashion.
```
damithc, line 68: I know this is not related to this PR, but is the GitClone API provided by GitHub? If not, probably no need to mention *GitHub* in this step?

docs/dg/architecture.md
```diff
@@ -65,9 +65,12 @@
...
+    * by default, 4 threads are used for cloning; the number of threads can be specified using the CLI argument `--cloning-threads <threads>`.
+ 1. analyzes the repositories using the `CommitReporter` and `AuthorshipReporter`, in a multi-threaded fashion.
+    * first, copies the template files into the designated output directory.
```
damithc, line 71: As I was reading this, I was wondering "What are these _template_ files?"

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -173,96 +177,158 @@ private static void setLandingPageTitle(String filePath, String pageTitle) throw
...
-                        Author.NAME_NO_AUTHOR_WITH_COMMITS_FOUND));
-                generateEmptyRepoReport(repoReportDirectory.toString(), Author.NAME_NO_AUTHOR_WITH_COMMITS_FOUND);
-            } catch (Exception e) {
-                StringWriter sw = new StringWriter();
-                e.printStackTrace(new PrintWriter(sw));
-                logger.log(Level.SEVERE, sw.toString());
-                handleAnalysisFailed(configs, configToAnalyze,
-                        String.format(LOG_UNEXPECTED_ERROR, configToAnalyze.getLocation(), sw.toString()));
+        List<AnalysisErrorInfo> analysisErrors = new ArrayList<>();
+        if (cloneSuccessful) {
```
damithc, line 283: Use a guard clause to reduce nesting?

damithc (commented):
Added a few comments. The code/comments looks fine; I didn't check the logic though.

damithc:
Guys, let's try to merge this by this weekend.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -173,96 +177,158 @@ private static void setLandingPageTitle(String filePath, String pageTitle) throw
...
+
+        RepoCloner repoCloner = new RepoCloner();
         repoCloner.cleanup();
         return generatedFiles;
     }
 
     /**
-     * Analyzes all repos in {@code configsToAnalyze} and generates their report.
-     * Also removes {@code configsToAnalyze} that failed to analyze from {@code configs}.
+     * Clones repo at {@code locaation}.
```
fzdy1914, line 249: Typo here.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -173,96 +177,158 @@ private static void setLandingPageTitle(String filePath, String pageTitle) throw
...
+        if (clonedRepoLocation != null) {
+            String defaultBranch = repoCloner.getCurrentRepoDefaultBranch();
+            return new CloneJobOutput(location, defaultBranch);
+        } else {
+            return new CloneJobOutput(location);
+        }
+    }
+
+    /**
+     * Analyzes all repos in the configs associated with  and generates their report.
```
fzdy1914, line 269: Two spaces between `with` and `and`

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -410,6 +410,49 @@ public void formats_inAlphanumeric_success() throws Exception {
         Assert.assertEquals(expectedFormats, cliArguments.getFormats());
     }
 
+    @Test
+    public void numCloningThreads_default_success() throws Exception {
+        String input = DEFAULT_INPUT_BUILDER
```
fzdy1914, line 415: Can be put in one single line.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -410,6 +410,49 @@ public void formats_inAlphanumeric_success() throws Exception {
...
+        CliArguments cliArguments = ArgsParser.parse(translateCommandline(input));
+        Assert.assertTrue(cliArguments instanceof ConfigCliArguments);
+        int expectedNumThreads = 2;
+        Assert.assertEquals(expectedNumThreads, cliArguments.getNumCloningThreads());
+    }
+
+    @Test
+    public void numAnalysisThreads_default_success() throws Exception {
+        String input = DEFAULT_INPUT_BUILDER
+                .build();
```
fzdy1914, line 438: Same here, can be put in one single line.

fzdy1914 (approved):
Logic-wise OK to me. Good comment in both code and DG!
Several small issues are found.

Last question, Now the system test will be run in multi-threading by default right?

docs/dg/architecture.md
```diff
@@ -65,9 +65,12 @@
 ## ReportGenerator(Main)
 
 [`ReportGenerator`](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/report/ReportGenerator.java)
- 1. uses `GitClone` API to clone the repository from *GitHub*.
- 1. copies the template files into the designated output directory.
- 1. uses `CommitReporter` and `AuthorshipReporter` to produce the commit and authorship summary, respectively.
+ 1. clones repositories from *GitHub* using the `GitClone` API, in a multi-threaded fashion.
+    * by default, 4 threads are used for cloning; the number of threads can be specified using the CLI argument `--cloning-threads <threads>`.
```
dcshzj, line 69: The sub-bullets should capitalise the first letter. "By default" instead of "by default".

docs/dg/architecture.md
dcshzj: Same for this part.

dcshzj (changes requested):
LGTM! No other comments.

docs/dg/architecture.md
vig42: I think it's referring to the files in src/main/resources/templateZip.zip. So basically . Seems like they are copied over to the report directory by the `prepareTemplateFile` method.

src/main/java/reposense/report/ReportGenerator.java
vig42: I have added this.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -173,56 +177,118 @@ private static void setLandingPageTitle(String filePath, String pageTitle) throw
...
+            // The `thenApplyAsync` method is used to analyze the cloned repo in parallel.
+            // This ensures that the analysis job for each repo will only be run after the repo has been cloned.
+            // Note that the `analyzeExecutor` is passed as a parameter to ensure that the number of threads used
+            // for analysis is no more than `numAnalysisThreads`.
+            CompletableFuture<AnalyzeJobOutput> analyzeFuture = cloneFuture.thenApplyAsync(
+                    cloneJobOutput -> analyzeRepos(outputPath, configsToAnalyze, cloneJobOutput), analyzeExecutor);
+
+            analyzeJobFutures.add(analyzeFuture);
         }
+
```
vig42, line 218: FYI, I changed this portion a bit to make it more readable. Instead of the streams, I switched bacj to a for loop. I feel that it makes the 2 lines relating to the `CompletableFutures` a little easier to parse.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -173,56 +177,118 @@ private static void setLandingPageTitle(String filePath, String pageTitle) throw
...
+        // Finally, we collect the list of outputs from all the analyze jobs
+        List<AnalyzeJobOutput> jobOutputs = analyzeJobFutures.stream()
+                .map(CompletableFuture::join)
+                .collect(Collectors.toList());
+
+        // Finally, the ExecutorService objects are shut down to prevent memory leaks
```
damithc, line 224: Minor: `Finally` used in the previous step as well.

damithc (approved):
No further comments from me.

dcshzj (approved):
LGTM!

fzdy1914:
@Tejas2805 Your review is needed here.

fzdy1914:
@vig42 PLS, resolve the conflicts.

vig42:
> 
> 
> @vig42 PLS, resolve the conflicts.

Conflicts are resolved, and tests are all passing

Tejas2805 (approved):
LGTM!

docs/dg/architecture.md
```diff
@@ -65,9 +65,12 @@
 ## ReportGenerator(Main)
 
 [`ReportGenerator`](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/report/ReportGenerator.java)
- 1. uses `GitClone` API to clone the repository from *GitHub*.
- 1. copies the template files into the designated output directory.
- 1. uses `CommitReporter` and `AuthorshipReporter` to produce the commit and authorship summary, respectively.
+ 1. clones repositories using the `GitClone` API, in a multi-threaded fashion.
```
Tejas2805, line 68:
```suggestion
 1. clones repositories using the `GitClone` API in a multi-threaded fashion.
```

docs/dg/architecture.md
```diff
@@ -65,9 +65,12 @@
...
+    * By default, 4 threads are used for cloning; the number of threads can be specified using the CLI argument `--cloning-threads <threads>`.
+ 1. analyzes the repositories using the `CommitReporter` and `AuthorshipReporter`, in a multi-threaded fashion.
```
Tejas2805, line 70:
```suggestion
 1. analyzes the repositories using the `CommitReporter` and `AuthorshipReporter` in a multi-threaded fashion.
```

Tejas2805 (commented):
Two minor nits.

One question - What happens if more threads are specified? If my understanding is right, the MAX....... will get invoked right?

vig42:
> 
> 
> Two minor nits.
> 
> One question - What happens if more threads are specified? If my understanding is right, the MAX....... will get invoked right?

Actually, since the `--cloning-threads` and `--analysis-threads` flags were implemented as dev flags, I removed the "MAX" constants. This would allow devs to test more freely with varying numbers of threads.

Tejas2805:
> Actually, since the --cloning-threads and --analysis-threads flags were implemented as dev flags, I removed the "MAX" constants. This would allow devs to test more freely with varying numbers of threads.

So, what happens now if a dev puts in more number of threads than what his computer allows?

vig42:
> 
> So, what happens now if a dev puts in more threads than what his computer allows? Does it just work based on the max supported by the system?

I am honestly not sure, but I think it would be basically impossible to hit the limit. As far as I know, the theoretical limit for the number of threads per process for 32-bit Windows is around 2,000 (based on the max memory per process of 2GB and min stack size of 1MB); for 64-bit the limit would be 8 million threads. For Linux, I believe the only limit that applies is the max number of processes, which is ~32 thousand on 32-bit and 4 million on 64-bit. Not very sure about Mac, but I imagine it is somewhat similar.

In order for RepoSense to create that many threads, you would not only have to set the parameter, e.g. `--analysis-threads 2000`, but you would also have to analyze 2000 different repos. Given this extremely rare scenario, I think most likely you would run out of memory before hitting the limit on threads.

fzdy1914:
Will merge this today

PR #1361, opened 2020-12-09
[#1352] Add highlight for </> when clicked

ZhangYiping126:
Fixes #1352

Issue #1352, opened 2020-11-01
Dashboard: the chart view should indicate which author is shown in code view


damithc:
![image](https://user-images.githubusercontent.com/1673303/97795999-1a560480-1c48-11eb-97f5-839d1f12cdee.png)

Steps:
* Click on the `</>` icon on the chart view
* Scroll down the code view

Problem: need to scroll to the top of the code view to recall who's code is being viewed

Suggestion: Highlight the `</>` icon when clicked? Not sure about the coding and performance costs though.

fzdy1914:
It should be easy to implement.

Tejas2805:
> Suggestion: Highlight the </> icon when clicked?

If the user scrolls down the summary panel too, the error will still remain. Better to have some mechanism of knowing whose code we are viewing at all points of time?

PR #1453, opened 2021-02-12
[#1352] Add highlight to selected authorship or commit tab

HCY123902:
This is a continuation of #1361

Commit message:
```
Add highlight to selected authorship or commit tab

Given an authorship or commit panel with a large 
number of file changes involved, it may be 
inconvenient for the user to see which author is 
currently inspected upon when the user scrolls down 
the panel.

Adding highlight to the selected authorship or commit
panel in the summary chart can make it more 
convenient for the user.

Let's add the highlight to the selected panel.
```


dcshzj:
Thanks for the PR! I have 2 comments.

Firstly, I think the yellow colour might be a little bit light and is not immediately obvious why the colour has changed. Perhaps use a darker colour and maybe use the tooltips to explain to the user?

Secondly, the appearance of the yellow does not seem to be consistent. When a user accesses the RepoSense report directly using the URL, the icon does not appear yellow when it should be. To reproduce this error, try clicking on the icon to make it yellow, then go to the browser's address bar, click the URL and press enter (to simulate accessing the report directly). The icon does not appear yellow. The same issue also occurs when you try to refresh the page after clicking on the icon.

HCY123902:
Thank you for the comment. I have changed it accordingly.

gerhean:
Is it possible to add highlighting to the zoom tab as well? Since it is currently inconsistent that only authorship tab has highlighting.

gerhean:
image

If I merge all groups the highlighting is not enabled.

gerhean:
Bug:
When I change the sorting of the groups, the highlighting disappears.
Then when I refresh the page, the highlighting is based on the index of the authorship which I clicked, and not the actual author.
image

rolandyuwy:
> Is it possible to add highlighting to the zoom tab as well? Since it is currently inconsistent that only authorship tab has highlighting.

- Agreed with point on consistency. 
- Can consider highlighting on hover as well, similar to other components.
- Personally, additional tooltip text is a little too wordy for readability. Coloured icons seems intuitive enough to indicate selection. Open for discussion.
image

damithc:
> * Personally, additional tooltip text is a little too wordy for readability. Coloured icons seems intuitive enough to indicate selection. Open for discussion.

Yes, I think no need to add it to the tool tip.

HCY123902:
> Bug:
> When I change `sort within groups by`, the highlighting disappears (similar with merge groups and other toolbar changes).
> Then when I refresh the page, the highlighting is based on the index of the authorship which I clicked, and not the actual author.
> image

Thanks for pointing it out. I have changed it accordingly.

HCY123902:
> If I merge all groups the highlighting is not enabled when icon is clicked.
> image

I will enable the highlighting for merged group. I think it may also be necessary to highlight the icon for the merged group if individual author within the group is selected for inspection of contribution

frontend/src/static/js/v_summary_charts.js
```diff
@@ -251,6 +260,38 @@ window.vSummaryCharts = {
...
+
+    getCurrentRepo() {
+      const hash = window.hashParams;
+
+      if (hash.activeIndex) {
+        this.activeRepo = hash.activeRepo;
+      }
+
+      return this.activeRepo;
+    },
```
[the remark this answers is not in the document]
HCY123902: Thanks for pointing it out. I have changed it accordingly.

frontend/src/static/js/v_summary_charts.js
```diff
@@ -116,7 +119,9 @@ window.vSummaryCharts = {
       if (Object.prototype.hasOwnProperty.call(location, 'organization')) {
         return `${window.BASE_URL}/${location.organization}/${location.repoName}/tree/${branch}`;
       }
-
+      this.activeIndex = null;
+      this.activeRepo = null;
+      this.removeSelectedTabHash();
```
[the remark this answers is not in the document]
HCY123902: This may require the method to include more work, but it makes it more encapsulated. I will take a look and make the adjustment if necessary.

HCY123902:
> > * Personally, additional tooltip text is a little too wordy for readability. Coloured icons seems intuitive enough to indicate selection. Open for discussion.
> 
> Yes, I think no need to add it to the tool tip.

Thanks for the comment. I have changed it accordingly.

gerhean:
Not sure if it is intended behavior but when I click a person's repo, and tick to `merge all groups`, the group contribution icon is highlighted as well.
image

Also when I refresh page with the group authorship open and have `merge all groups` unticked, the first person has group contribution icon highlighted (this is quite minor tho).
image

HCY123902:
> Not sure if it is intended behavior but when I click a person's repo, and tick to `merge all groups`, the group contribution icon is highlighted as well.
> image
> 
> Also when I refresh page with the group authorship open and have `merge all groups` unticked, the first person has group contribution icon highlighted (this is quite minor though).
> image

Thanks for the comment. 
The first behavior is intended because the panel shows the contribution of an individual within the group, and it seems more consistent to me to highlight the group instead of not highlighting anything.
I will look into the second situation.

HCY123902:
I have made the change accordingly. Is there any other comment or suggestion on this pull request?

Tejas2805:
Not sure about the status of this PR. 

Initial comments - Only the Code Panel icon is being highlighted. Better to duplicate the same for Commit Panel as well. Also, I think the green colour is not too visible. I am searching for it specifically and therefore I can find it, I think better to make it more visible smh. Can you show an example of how it looks if you draw a box or something around it? Or shade the background of the icon?

HCY123902:
Thanks for the suggestion. I have made the changes regarding the commit tab. I also changed the highlight by changing the background of the icon selected

frontend/src/static/css/v_summary.scss
```diff
@@ -9,6 +9,8 @@
   .icon-button {
     color: mui-color('grey');
     cursor: pointer;
+    padding-left: 1.2px;
+    padding-right: 1.2px;
```
0blivious, line 13: why introducing this additional padding?

0blivious:
TBH I am not sure if it is a good idea to add so many things into hashing just for this simple function. @fzdy1914?

frontend/src/static/css/v_summary.scss
HCY123902: This is to add additional padding for the background of the selected icon since I have set the background color to be green. The coloured background may be bounded too tightly by the border without this padding, which can make the icon look a bit strange. If there is any alternative design of the highlight effect, then the padding may not be needed.

HCY123902: @damithc Hello professor, can I ask for your opinion on the current design of the highlight effect? Many thanks

damithc: Sorry, I didn't think about this earlier. The current one feels too subtle. It defeats the purpose if the highlighted item is not easy to spot. What about doing something to the entire box rather than just the item only e.g., make the background of the entire area light yellow?

image

HCY123902: Thanks for the suggestion. However, in order to distinguish whether it is the authorship contribution or the commit panel that is currently being opened, will it be more suitable to highlight the icon with a more obvious visual effect instead?

damithc: Well, in the unlikely case the user forgot which one, he only has to look to the right panel to see which of the two it is, yes? :-)
In any case, no harm doing both i.e., colorize the icon and well as the full box.

HCY123902: Sure, no problem

0blivious:
A special use case for consideration
1. merge the group, open the zoom panel, the group title is highlighted with yellow
2. unmerged the group, the highlight disappeared, while the zoom panel still display the info about the group

So should we keep the yellow bg for the group title in this case? @damithc

Tejas2805:
> A special use case for consideration
> 
> 1. merge the group, open the zoom panel, the group title is highlighted with yellow
> 2. unmerged the group, the highlight disappeared, while the zoom panel still display the info about the group

3. When the group is merged again, the highlight comes back. 

Maybe can highlight all the authors in the group for step 2.

damithc:
> So should we keep the yellow bg for the group title in this case? @damithc

I think it is fine to lose the highlighting in that case. It is just a convenience feature only. Not having it is fine, as long as we are not highlighting the wrong author/group.

frontend/src/static/css/v_summary.scss
0blivious, line 13: can make this to one line padding: a b c d?

frontend/src/summary_charts.pug
```diff
@@ -52,14 +56,28 @@
           v-on:click="openTabAuthorship(repo[0], repo, 0, isGroupMerged(getGroupName(repo)))"
         )
           .tooltip
-            font-awesome-icon.icon-button(icon="code")
+            font-awesome-icon.icon-button.active-icon(
+              icon="code",
+              v-if="isSelectedTab(repo[0].name, repo[0].repoName, 'authorship', true)"
```
0blivious, line 61: can use class binding if no other content changed here?

frontend/src/static/css/v_summary.scss
HCY123902: Sure, no problem

frontend/src/summary_charts.pug
HCY123902: Sure, no problem

0blivious (approved):

fzdy1914 (approved):
Good Work Done!

Tejas2805 (approved):
LGTM!

gerhean (approved):
LGTM

Issue #442, opened 2018-12-18
Deprecate the CommandRunner class


eugenepeh:
Our tool has the pre-requisite of existing git installation and command line dependency which makes it more tedious to set up compare to a general tool. 

Additionally, this reduce the maintainability due to security issues such as command injection, and may also prevent us from developing it into a integrate-able tool such as gradle plugin.

Let's move towards alternative such as jgit and deprecate the use of terminal / command prompt.

dcshzj:
This is quite an epic issue and in the interest of reporting my progress so far, here are some notes on my attempt to perform this migration.

**Potential replacements**

The following is a table of the various CommandRunner usages and their potential replacements in JGit:

| Usage            | JGit API                        | JGit Class               | Remarks                                                                  |
|------------------|---------------------------------|--------------------------|--------------------------------------------------------------------------|
| GitBlame.java    | `Git.blame()`                     | BlameCommand, BlameResult | Need to build `git blame` result format                                  |
| GitBranch.java   | `Git.getRepository().getBranch()` | Repository               |                                                                          |
| GitCheckout.java | `Git.checkout()`                  | CheckoutCommand          |                                                                          |
| GitClone.java    | `Git.cloneRepository()`           | CloneCommand             |                                                                          |
| GitDiff.java     | `Git.diff()`                      | DiffCommand, DiffEntry    | Need to build `git diff` result format                                   |
| GitLog.java      | `Git.log()`                       | LogCommand               | Need to build `git log` result format, might not have certain information |
| GitLsTree.java   | _None_                          | _None_                   | Need to build using `RevWalk` and `TreeWalk` ([more info](https://stackoverflow.com/questions/17655619/get-a-list-of-files-in-a-revision-with-jgit))                            |
| GitRevList.java  | _None_                          | _None_                   | Need to build using `LogCommand` instead ([more info](https://stackoverflow.com/questions/27278328/use-jgit-to-get-the-git-commit-count))                                |
| GitRevParse.java | _None_                          | _None_                   | Potentially no replacements available ([more info](https://stackoverflow.com/questions/39356463/is-there-a-jgit-method-equivalent-to-git-rev-parse-short))                                   |
| GitShortlog.java | `Git.log()`                       | LogCommand               | Need to build `git shortlog` result format                               |

**Resources**
1. A [jgit-cookbook](https://github.com/centic9/jgit-cookbook) that provides examples and code snippets for using JGit
[...]

PR #1454, opened 2021-02-13
[#442] Replace CommandRunner with JGit for GitBranch

dcshzj:
Part of #442.

```
CommandRunner should be deprecated and replaced by JGit as it
requires Git to be installed and has poor exception handling.
JGit is able to handle Git operations natively, which makes it
more robust to use.

Let's replace usages of CommandRunner with their equivalents in
JGit, starting with GitBranch.
```

This PR is a proof-of-concept of what JGit can potentially do, instead of relying on CommandRunner.

### DIFF
--- a/src/main/java/reposense/report/RepoCloner.java
+++ b/src/main/java/reposense/report/RepoCloner.java
@@ -68,11 +68,18 @@ public class RepoCloner {
             String bareRepoPath = FileUtil.getBareRepoPath(configs[currentIndex]).toString();
             currentRepoDefaultBranch = GitBranch.getCurrentBranch(bareRepoPath);
         } catch (GitBranchException gbe) {
-            // GitBranch will throw this exception when repository is empty
             logger.log(Level.WARNING, String.format(MESSAGE_ERROR_GETTING_BRANCH,
                     configs[currentIndex].getLocation(), configs[currentIndex].getBranch()), gbe);
             return null;
         }
+
+        if (currentRepoDefaultBranch == null) {
+            // Current repository is empty
+            logger.log(Level.WARNING, String.format(MESSAGE_ERROR_GETTING_BRANCH,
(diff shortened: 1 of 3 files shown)


src/main/java/reposense/report/RepoCloner.java
```diff
@@ -68,11 +68,18 @@ public RepoLocation getClonedRepoLocation() {
             String bareRepoPath = FileUtil.getBareRepoPath(configs[currentIndex]).toString();
             currentRepoDefaultBranch = GitBranch.getCurrentBranch(bareRepoPath);
         } catch (GitBranchException gbe) {
-            // GitBranch will throw this exception when repository is empty
             logger.log(Level.WARNING, String.format(MESSAGE_ERROR_GETTING_BRANCH,
```
Tejas2805, line 71: Why do we need this given you have added code for this already? Or did you forget to remove it?

dcshzj: I didn't remove this exception handling as I think that there might be some other errors that may occur. For instance, GitBranchException is thrown because an IOException occurred, which probably happened if the directory given is not a Git repository (which shouldn't happen).

damithc:
Good to do a light exploration of this direction at first before going hard at it. For example, we can try to get it working first before deciding whether to merge (hence, no need for a detailed review of the PR first)
I'm particularly worried about not being able to use the all/latest features/improvements of Git and not particularly worried about having to install Git first. It's not unreasonable to expect users to have Git if they want to run RepoSense locally.
If we decide not to go forward, we should still document pros and cons found.

dcshzj:
I have updated my original comment in #442 with a pros and cons list based on the research that I have done on JGit, so that we can make a more informed decision.

fzdy1914:
As the discussion, I will make it `DoNotMerge`

PR #1456, opened 2021-02-19
[#1414] Add message before closing a stale PR

### LINKED ISSUE #1414: add a message when the bot closes a stale issue/PR
We can add an explanatory message when the bot closes a stale PR e.g., https://github.com/reposense/RepoSense/pull/1308#event-4220302825

dcshzj:
Fixes #1414.

```
The current workflow is set up such that a PR that is inactive for
30 days will be marked as stale with a comment. However, after the
7 days have elapsed with no further activity, the workflow just
closes the PR without any comment.

Let's add a comment to the PR when closing it, so that it is
clearer to the PR author on why their PR was closed.

Let's also specify only the major version for the dependency to
remain forward-compatible.
```


gerhean:
LGTM but is there a reason for upgrading the version of the dependencies? Will it break anything?

.github/workflows/stale.yml
```diff
@@ -7,20 +7,23 @@ on:
 
 jobs:
   stale:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/stale@v3.0.7
+    - uses: actions/stale@v3
```
gerhean, line 13: Is there a reason for upgrading the version of the dependencies? Will it break anything? Maybe write that in your commit message too?

gerhean (commented):
Otherwise LGTM

.github/workflows/stale.yml
dcshzj: It is to remain forward-compatible. As per semver, changes in the minor and patch versions will not cause breaking changes. I have also modified the commit message accordingly.

gerhean (commented):
LGTM

gerhean (approved):

gerhean:
Actually is there a way to test that this works?

dcshzj:
> Actually is there a way to test that this works?

The only way I can think of is to create a PR in a test repository and wait for 30 + 7 days. However, I think that the code is straightforward enough to avoid over-testing.

HCY123902:
LGTM

0blivious (approved):

.github/workflows/stale.yml
```diff
@@ -7,20 +7,23 @@ on:
 
 jobs:
   stale:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
Tejas2805, line 10: Always better to use a specific version in my opinion. As the `latest` can keep changing.

Tejas2805 (commented):
One nit. Rest LGTM.

.github/workflows/stale.yml
dcshzj: Yep I wanted it to change along with the use of `actions/stale@v3`, so we don't have to keep updating this.

Tejas2805: It's not about having to continuously update it. If it's a fixed version, we will always know up to which version it worked well with. @fzdy1914 your opinion.

fzdy1914: From my view, we should not use `an unspecified version` unless it is truly needed.

dcshzj: I disagree in this context, but nonetheless I have reverted this change based on the consensus reached.

fzdy1914: @dcshzj It is to avoid unwanted bugs that may happen due to the latest version while the environment changes.

PR #1459, opened 2021-03-02
CSVParser: check column headers when parsing config CSV files

vig42:
```
Currently, RepoSense parses the config CSV files purely based on
column order. This means that every column needs to be present in the
correct order (even if the column is optional) in order for the CSV
file to be parsed. This additionally means that every time we add a
new column to a config CSV, we would break compatibility with existing
config CSV files for all users.

Let’s change the CSVParser to verify and parse the config CSV files
based on the column headers. This would allow us to parse config CSVs
with columns in a different order or with some (optional) columns
missing, and would allow us to add new columns without breaking
compatibility.
```

### DIFF
--- a/src/main/java/reposense/parser/CsvParser.java
+++ b/src/main/java/reposense/parser/CsvParser.java
@@ -17,6 +17,7 @@ import java.util.StringJoiner;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.commons.csv.CSVFormat;
 import org.apache.commons.csv.CSVRecord;
@@ -42,6 +43,8 @@ public abstract class CsvParser<T> {
     private static final String MESSAGE_EMPTY_CSV_FORMAT = "The CSV file, %s, is empty.";
     private static final String MESSAGE_MANDATORY_HEADER_MISSING = "Required column header, %s, not found in "
             + "CSV file, %s";
+    private static final String MESSAGE_DUPLICATE_COLUMN_HEADER = "Column header. %s, appears twice in CSV file, %s. "
+            + "The rightmost occurrence will be used.";
     private static final String MESSAGE_COLUMNS_RECOGNIZED = "Parsed header of CSV file, %s, and recognized columns: "
             + "%s";
     private static final String MESSAGE_ZERO_VALID_CONFIGS = "No valid configurations in the %s.";
@@ -227,14 +230,12 @@ public abstract class CsvParser<T> {
         headerSize = possibleHeader.length;
         for (int i = 0; i < headerSize; i++) {
             String possible = possibleHeader[i].trim();
-            for (String parsedHeader : mandatoryHeaders()) {
+            for (String parsedHeader : mandatoryAndOptionalHeaders()) {
                 if (possible.equalsIgnoreCase(parsedHeader)) {
-                    headerMap.put(parsedHeader, i);
-                    break;
-                }
-            }
-            for (String parsedHeader : optionalHeaders()) {
-                if (possible.equalsIgnoreCase(parsedHeader)) {
+                    if (headerMap.containsKey(parsedHeader)) {
+                        logger.warning(String.format(
+                                MESSAGE_DUPLICATE_COLUMN_HEADER, parsedHeader, csvFilePath.getFileName()));
+                    }
                     headerMap.put(parsedHeader, i);
                     break;
                 }
@@ -259,6 +260,14 @@ public abstract class CsvParser<T> {
      * Gets the list of optional headers that can be parsed.
      */
     protected abstract String[] optionalHeaders();
+
+    /**
+     * Gets the list of all mandatory and optional headers that can be parsed.
+     */
+    protected List<String> mandatoryAndOptionalHeaders() {
+        return Stream.concat(Arrays.stream(mandatoryHeaders()), Arrays.stream(optionalHeaders()))
+                .collect(Collectors.toList());
+    }
 
     /**
      * Processes the csv file line by line.
--- a/src/main/java/reposense/parser/CsvParser.java
+++ b/src/main/java/reposense/parser/CsvParser.java
@@ -50,7 +50,6 @@ public abstract class CsvParser<T> {
     private static final String MESSAGE_ZERO_VALID_CONFIGS = "No valid configurations in the %s.";
 
     private Path csvFilePath;
-    private int headerSize;
     private Map<String, Integer> headerMap = new HashMap();
     private int numOfLinesBeforeFirstRecord = 0;
 
@@ -227,7 +226,7 @@ public abstract class CsvParser<T> {
      * @throws InvalidCsvException if {@code possibleHeader} does not contain all the mandatory headers.
      */
     private void validateHeader(String[] possibleHeader) throws InvalidCsvException {
-        headerSize = possibleHeader.length;
+        int headerSize = possibleHeader.length;
         for (int i = 0; i < headerSize; i++) {
             String possible = possibleHeader[i].trim();
             for (String parsedHeader : mandatoryAndOptionalHeaders()) {
--- a/src/main/java/reposense/parser/CsvParser.java
+++ b/src/main/java/reposense/parser/CsvParser.java
@@ -6,12 +6,7 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
-import java.util.StringJoiner;
+import java.util.*;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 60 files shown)


damithc:
As we are adding a restriction on column names, we'll also need good error reporting e.g., RepoSense should inform the user which column is missing, which column name is incorrect, etc.

vig42:
> 
> 
> As we are adding a restriction on column names, we'll also need good error reporting e.g., RepoSense should inform the user which column is missing, which column name is incorrect, etc.

Right now if a mandatory column is missing, RepoSense would throw an InvalidCsvException with the name of the missing column. For optional columns, if the column is missing, there is no exception thrown (because they are optional). This allows us to add in new columns without breaking compatibility.

However, this also means that if the column header has a typo in it, the user would not get any indication. If a column header does not match any of the specified mandatory or optional columns, we could do some form of typo detection using string distance and warn the user if it is a close match.

Otherwise, we could provide a list of all the columns that were successfully parsed. This would allow the user to manually check if 1 of the columns was not recognized, and would probably help with debugging as well.

damithc:
> Otherwise, we could provide a list of all the columns that were successfully parsed. This would allow the user to manually check if 1 of the columns was not recognized, and would probably help with debugging as well.

Yes, we should do this.

> However, this also means that if an optional column header has a typo in it, the user would not get any indication. If a column header does not match any of the specified mandatory or optional columns, we could do some form of typo detection using string distance and warn the user if it is a close match.

Probably no need to cater for typos (at least not urgent) but should be able to deal with case-differences and whitespace differences (at least leading/trailing spaces).

vig42:
> 
> 
> > Otherwise, we could provide a list of all the columns that were successfully parsed. This would allow the user to manually check if 1 of the columns was not recognized, and would probably help with debugging as well.
> 
> Yes, we should do this.

I have added this. The log output looks like this:
`12:01:31 - Parsed header of CSV file, group-config.csv, and recognized columns: Globs,  Repository's Location,  Group Name`

> 
> Probably no need to cater for typos (at least not urgent) but should be able to deal with case-differences and whitespace differences (at least leading/trailing spaces).

Right now we're taking care of case differences by using the `equalsIgnoreCase` method to compare, and taking care of leading/trailing spaces by trimming the header before comparing.

Tejas2805:
@vig42 Is this ready for review?

vig42:
> 
> 
> @vig42 Is this ready for review?

Yes, should be ready.

I have added some checks to account for leading/trailing spaces and upper/lower case, and I have added an info log to show the user which columns have been parsed from the CSV, so I think Prof @damithc's concerns should be taken care of.

fzdy1914 (commented):
The idea is quite good, will proceed to review later

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -211,22 +220,45 @@ private String getRowContentAsRawString(final CSVRecord record) {
...
-            throw new InvalidCsvException(String.format(
-                    MESSAGE_WRONG_HEADER_SIZE, csvFilePath.getFileName(), actualNumberOfColumns,
-                    expectedHeaderSize));
+        headerSize = possibleHeader.length;
+        for (int i = 0; i < headerSize; i++) {
+            String possible = possibleHeader[i].trim();
+            for (String parsedHeader : mandatoryHeaders()) {
+                if (possible.equalsIgnoreCase(parsedHeader)) {
+                    headerMap.put(parsedHeader, i);
+                    break;
```
Tejas2805, line 233: Why is the 'break' been added here? Is it not supposed to be checking for all the mandatory headers?

vig42: The break is there because I am assuming that the parsed column header would match with at most 1 of the mandatory headers, i.e. that we would not have a case where there are 2 mandatory headers which are identical.

fzdy1914: Then, it seems that in a certain scenario, duplicated column names cannot be found out?

vig42: > 
> 
> Then, it seems that in a certain scenario, duplicated column names cannot be found out?

I have added a warning if there are duplicate headers detected in the CSV.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -211,22 +223,51 @@ private String getRowContentAsRawString(final CSVRecord record) {
...
-                    MESSAGE_WRONG_HEADER_SIZE, csvFilePath.getFileName(), actualNumberOfColumns,
-                    expectedHeaderSize));
+        headerSize = possibleHeader.length;
+        for (int i = 0; i < headerSize; i++) {
+            String possible = possibleHeader[i].trim();
+            for (String parsedHeader : mandatoryAndOptionalHeaders()) {
+                if (possible.equalsIgnoreCase(parsedHeader)) {
+                    if (headerMap.containsKey(parsedHeader)) {
+                        logger.warning(String.format(
+                                MESSAGE_DUPLICATE_COLUMN_HEADER, parsedHeader, csvFilePath.getFileName()));
```
dcshzj, line 237: We should add a test where the CSV file has multiple headers of the same name.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -211,22 +223,51 @@ private String getRowContentAsRawString(final CSVRecord record) {
...
+                    }
+                    headerMap.put(parsedHeader, i);
```
dcshzj, line 239: I think that we should take the first instance instead of the last instance of the same header, as we primarily read from left to right and we would usually register the first instance. Open for discussion.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -37,26 +41,29 @@
...
+            + "CSV file, %s";
+    private static final String MESSAGE_DUPLICATE_COLUMN_HEADER = "Column header. %s, appears twice in CSV file, %s. "
+            + "The rightmost occurrence will be used.";
+    private static final String MESSAGE_COLUMNS_RECOGNIZED = "Parsed header of CSV file, %s, and recognized columns: "
+            + "%s";
     private static final String MESSAGE_ZERO_VALID_CONFIGS = "No valid configurations in the %s.";
 
     private Path csvFilePath;
-    private int expectedHeaderSize;
+    private int headerSize;
```
dcshzj, line 53: This is used only once, I think we should keep it inside the `validateHeader` function as a local variable.

dcshzj (changes requested):

src/main/java/reposense/parser/CsvParser.java
vig42: I have changed `headerSize` to a local variable.

vig42:
@dcshzj @fzdy1914 FYI, I found out that the existing code would already throw a runtime exception if there were any duplicate headers passed. The exception is thrown by line 75 here:

https://github.com/reposense/RepoSense/blob/8c052946cee72d757fcdf013c42d31ff27fc28c5/src/main/java/reposense/parser/CsvParser.java#L69-L76

This issue is also present in the master branch. If you modify the repo-config.csv and edit 1 of the headers to be a duplicate of another header (e.g. change "Ignore Glob List" to "File formats"), then you would see this:
```
Exception in thread "main" java.lang.IllegalArgumentException: The header contains a duplicate entry: 'File formats' in [Repository's Location, Branch, File formats, File formats, Ignore standalone config, Ignore Commit List, Ignore Authors List]
        at org.apache.commons.csv.CSVFormat.validate(CSVFormat.java:1466)
        at org.apache.commons.csv.CSVFormat.<init>(CSVFormat.java:700)
        at org.apache.commons.csv.CSVFormat.withHeader(CSVFormat.java:1746)
        at reposense.parser.CsvParser.parse(CsvParser.java:75)
        at reposense.RepoSense.getRepoConfigurations(RepoSense.java:109)
        at reposense.RepoSense.main(RepoSense.java:64)
```

By default, this only picks up exact duplicates, i.e. it is case sensitive. I have added `.withTrim()` and `.withIgnoreHeaderCase()` so that it will catch duplicates even if they are not in the same case and if they have leading/trailing spaces. I have also added a try/catch block to catch any runtime exceptions and throw an `InvalidCsvException` with the respective filename.

This means that RepoSense won't allow duplicate headers in the config files, so there is no need to worry about whether we take the leftmost or rightmost duplicate.

src/main/java/reposense/parser/CsvParser.java
vig42: I have added 2 tests, for case-sensitive and case-insensitive duplicates.

dcshzj (approved):
LGTM!

Tejas2805 (approved):
LGTM!

damithc:
Guys, shouldn't this PR have been merged before the shallow cloning one? Anyway, let's merge this soon.

fzdy1914:
> Guys, shouldn't this PR have been merged before the shallow cloning one? Anyway, let's merge this soon.



> Guys, shouldn't this PR have been merged before the shallow cloning one? Anyway, let's merge this soon.

I don't think that two conflicts though.

damithc:
> I don't think that two conflicts though.

That one introduced a new column into the csv file that broke all the existing dashboards. This PR is supposed to ignore optional columns. In fact, this PR was created to prevent the former from happening. Anyway, it's not a big problem. I'll wait till this is merged to use the `master` again. Just don't do another release until this is merged. @HCY123902

fzdy1914:
> > I don't think that two conflicts though.
> 
> That one introduced a new column into the csv file that broke all the existing dashboards. This PR is supposed to ignore optional columns. In fact, this PR was created to prevent the former from happening. Anyway, it's not a big problem. I'll wait till this is merged to use the `master` again. Just don't do another release until this is merged. @HCY123902

I am sorry that I was not informed of this issue. Anyway, I have merged it right now.

PR #1412, opened 2021-01-18
[#795] Projectify frontend part

gerhean:
Working on this.

Up to date with master branch as of 29/6/2021.

Things to do in another PR: 
- Add linter for pug
- Resolve eslint warnings
- Remove all `window.`

How to port code into new-frontend:
- Copy directly into `.vue` or same extension file of same name
  - `pug` file copy into \<template> portion
  - `js` file copy into \<script> portion
  - `css` file copy into \<style> portion 
- If there is `import aaa from bbb` at the top of the \<script> portion, search for `windows. utilsAaa` in the copied code and replace it with `aaa`
- For other exceptions may refer to the differences to the new-frontend file and old file

How to test:
1. First generate the desired report json using `gradlew run`
2.  (Working for windows, not tested on mac, not working on linux)
```
gradlew hotReloadFrontend
```

Build at `./reposense-report`:
```
gradlew run -Dargs="..."
```

Commit message:
```
Let's convert frontend to a standard Vue project.
```

PR #1462, opened 2021-03-07
Extract zoom commit messages as a separate component

gerhean:
Fixes [#1340]
```
v-zoom currently uses window.toggleNext() to show and
hide commit messages.

window.toggleNext() directly modifies html (add classes and
elements directly), and so might conflict with how Vue works.

As a result, using window.toggleNext() creates code which is
hard to debug and maintain.

Lets rewrite v-zoom to use the Vue way of hiding and showing
elements by using v-if.

In order to do this, the code which handles the display of zoom
commit messages has been moved into a separate file. This also
helps with code readability.

Let's remove window.toggleNext().
```

### DIFF
--- a/frontend/src/static/js/v_zoom.js
+++ b/frontend/src/static/js/v_zoom.js
@@ -16,6 +16,7 @@ window.vZoom = {
   template: window.$('v_zoom').innerHTML,
   data() {
     return {
+      defaultExpansionStateSignal: false,
       expandedCommitMessagesCount: 0,
       ...initialState(),
     };
@@ -214,6 +215,7 @@ window.vZoom = {
 
     toggleAllCommitMessagesBody(isActive) {
       this.showAllCommitMessageBody = isActive;
+      this.defaultExpansionStateSignal = !this.defaultExpansionStateSignal;
 
       // const toRename = this.showAllCommitMessageBody
       // ? 'commit-message message-body active' : 'commit-message message-body';
--- a/frontend/src/static/js/v_zoom.js
+++ b/frontend/src/static/js/v_zoom.js
@@ -2,8 +2,8 @@
 
 function initialState() {
   return {
-    showAllCommitMessageBody: true,
-    // expandedCommitMessagesCount: 0,
+    showAllCommitMessageBody: 1,
+    // 1 means all expanded, 0 means all collapsed, -1 means inconsistent
     commitsSortType: 'time',
     toReverseSortedCommits: true,
     isCommitsFinalized: false,
@@ -16,7 +16,6 @@ window.vZoom = {
   template: window.$('v_zoom').innerHTML,
   data() {
     return {
-      defaultExpansionStateSignal: false,
       expandedCommitMessagesCount: 0,
       ...initialState(),
     };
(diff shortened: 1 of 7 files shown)


fzdy1914:
@gerhean Is this part of #1412?

gerhean:
no

damithc:
The width of the commit message body seems narrower than necessary? Or is my eye fooling me?

gerhean:
I can't see any difference in width, is it possible to provide a screenshot?

damithc:
Compare these two:
image
image

Yours seems noticeably narrower. Perhaps following a wrong width limit?

HCY123902:
Can you change the commit message to make it a bit wider according to the suggestion given? Thanks

gerhean:
I'm not sure if I'm blind but I literally can't see any difference...
image
image

damithc:
> I'm not sure if I'm blind but I literally can't see any difference...

I think @HCY123902 was referring to https://github.com/reposense/RepoSense/pull/1462#issuecomment-792237869
Your propose commit message is line wrapping too early.

gerhean:
Ok nevermind I think I might actually be blind, I'll be fixing it right away.

0blivious:
I agree with the idea of extracting out the component. However, this approach seems to introduce some even more complicated communication between zoom and the message body component. Why don't you follow the approach we used in the authorship panel, which uses v-show to control the toggle of the individual elements?

gerhean:
@0blivious I don't really want to introduce a new property into the `slice` object, since it is already hard to keep track of what properties are in what object. Plus it means I have to shallow copy the `slice` object as well.

I don't believe that the communication process is particularly complicated, given that I only added about 4 lines into v_zoom.js, while most of the communication logic is in v_zoom_commit.js, which I believe is a reasonable abstraction.

Following the approach we used in the authorship panel would increase the amount of code in v_zoom.js, making it overall harder to read.

frontend/src/static/js/v_zoom_commit.js
```diff
@@ -0,0 +1,71 @@
...
+  created() {
+    if (this.defaultExpansionState) {
+      this.toggleExpansion();
+    }
+  },
+  beforeDestroy() {
+    if (this.isExpanded) {
+      this.$emit('decrement-expanded-count');
+    }
+  },
```
0blivious, line 70: May I know that is the case for this beforeDestroy?

0blivious, line 70: May I know that is the case for this beforeDestroy?

frontend/src/static/js/v_zoom.js
```diff
@@ -15,7 +15,12 @@ function initialState() {
 window.vZoom = {
   template: window.$('v_zoom').innerHTML,
   data() {
-    return initialState();
+    return {
+      defaultExpansionStateSignal: false,
```
0blivious, line 19: Can we have some other naming like toExpandAll?

frontend/src/static/js/v_zoom_commit.js
```diff
@@ -0,0 +1,71 @@
+/* global Vuex */
+
+window.vZoomCommit = {
+  props: [
+      'defaultExpansionState',
+      'defaultExpansionStateSignal',
```
0blivious, line 6: Do we really need both state and signal? Is it possible to just use one?

PR #1466, opened 2021-03-07
[#1464] Remove group highlight when merging group

### LINKED ISSUE #1464: Incorrect highlighting after merging/unmerging groups
Example:

Highlighting is correct before merging groups:
image

After clicking on 'merge groups', the high-lighting is incorrect.
image

I think highlighting should be removed upon merging/unmerging groups. It's better not to have a highlighting than to highlight the wrong one.

HCY123902:
Resolves #1464 

Commit message:
```
Remove group highlight in summary view when the tab
is not on the merged group

Currently, a merged group will be highlighted in the summary
view even if the tab is not on the merged group but a member of
the group.

Removing the group highlight in this situation can make the 
highlight provide more accurate information.

Let's remove the group highlight when merging a group.
```


0blivious (approved):

gerhean (approved):
LGTM

fzdy1914 (approved):

PR #1469, opened 2021-03-07
[#1468] Dashboard loading empty if base URL is given

### LINKED ISSUE #1468: Dashboard loading empty if base URL is given
Any idea why https://nus-cs3281.github.io/2021-dashboard is loading empty? The data shows up only if I click on the 'reset date' button

image

Any recent change to URL parameters that could have caused this?

gerhean:
Fixes #1468
Regression caused by #1421.
```
Repositories with more than 19 different types of files will crash,
which can be traced from line 448 of v_summary.js

Bug is caused by a function not being copied properly in #1421,
leading to an undefined function call.

Let's copy the function over properly.
```

### DIFF
--- a/frontend/src/static/js/utils/ramp_colour_generator.js
+++ b/frontend/src/static/js/utils/ramp_colour_generator.js
@@ -1,7 +1,10 @@
+// eslint-disable-next-line new-cap
+const randomGenerator = new Math.seedrandom('Seeded Random Generator');
+
 function getRandomHex() {
   const maxHexColorValue = 16777214;
   // excludes #000000 and #FFFFFF as they are reserved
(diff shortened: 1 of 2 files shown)


gerhean:
Need add test cases for why this bug did not affect the test deployment.

fzdy1914 (approved):

fzdy1914:
@gerhean Can you re-generate the affected board? If the issue is resolved, I will merge this PR.

fzdy1914 (approved):

damithc:
It's better if test cases are updated to catch similar regressions that can happen in the future, unless you think it is too rare to cater for.

fzdy1914:
It is indeed something very rare to happen again. But we can add more file types in test env.

fzdy1914:
@gerhean Can you put it in the issue? To add more file types in default config.

PR #1475, opened 2021-03-12
[#1473] Checkbox not working properly in break down by file type

### LINKED ISSUE #1473: Checkbox not working properly in break down by file type

* **RepoSense Version:** master
* **OS and Version:** Windows10
* **Web Browser and Version (if applicable):** Edge

1. Open a report
1. Tick the `break down by file type` checkbox
1. Tick the `All` checkbox
1. Untick the `All` checkbox 
1. Tick any checkbox of a specific file type
1. Untick the checkbox
1. Tick the `All` checkbox again 

The ramp chart should be empty and the line count of each author should be changed to 0 after unticking the `All` checkbox in step 4.

The commit and line count of each file type should be restored after ticking the `All` checkbox in step 7.

The commit and line count are not changed after unticking the `All` checkbox in step 4.

image

The commit is empty and the line count is 0 after ticking the `All` checkbox in step 7.

image

gerhean:
Fixes #1473
```
this.getFiltered() is not called when checkAllFileTypes
is modified.

This is regression introduced by PR [#1440]

Let's call this.getFiltered() when checkAllFileTypes
is modified.
```


Tejas2805:
@gerhean there still seems to be some bug as noticed in deployment.

Untick on any filetype that is not displayed in a specific repository. The line count doesn't change but the length of the contribution bar still changes for the specific repository.

1) Java ticked.

ticked

2) Java unticked.

unitcked

Tejas2805 (changes requested):
The initial bug is fixed but can notice other bugs.

gerhean:
| Untick on any filetype that is not displayed in a specific repository. The line count doesn't change but the length of the contribution bar still changes for the specific repository.

This other bug is caused by the computed property `avgContributionSize` in `summary.js` not being a constant when report is filtered (it is a variable which depends on the line count of every single repository displayed).

I don't think this bug will be as simple to fix at least, since `avgContributionSize` is a computed property with side effects.

dcshzj (approved):
The commit message should be `getFiltered` instead of just `filtered`, otherwise LGTM.

0blivious (approved):
LGTM, in the PR message, please also specify this is regression introduced by PR [#1440]

0blivious:
@Tejas2805 I think I went through this before when I implemented this feature and at that time we agreed that this problem is tolerable.

gerhean:
@Tejas2805 So can the other bugs be fixed in a separate PR if needed?

Tejas2805 (approved):
LGTM!

fzdy1914:
```
`this.getFiltered()` is not called when `checkAllFileTypes` is 
modified. This is regression introduced by PR #1440.

Let's call `this.getFiltered()` when `checkAllFileTypes` is 
modified.
```

Issue #1383, opened 2020-12-19
Add virtual scrolling


gerhean:
**What feature(s) would you like to see in RepoSense?**

`v-ramp` is clearly very expensive to render. More likely than not the user will not scroll through the entire list of repos rendered, making it a waste.

Let's try to render components only when they become visible. This will hopefully improve front end performance.

**If possible, describe the solution**

Can consider using: https://github.com/Akryum/vue-virtual-scroller

**Additional context**

Might need to upgrade Vue to 2.6

on visible rendering might prevent searching the page using ctrl-f

damithc:
> on visible rendering might prevent searching the page using ctrl-f
> 
> Possible worse experience as users may need to wait for entries to render while scrolling

Alternative: render to top part first, and continue to render the rest in the background (no need to wait for user to scroll).

ckcherry23:
Another alternative we discussed is collapsing the ramp charts by default so that the ramp charts are only rendered when the user decides to expand the chart(s) they are interested in. However, we've opted against this approach because it diminishes the visual nature of data on the code dashboard, hindering comparison between charts.

PR #1477, opened 2021-03-12
v-ramp render on visible

gerhean:
Possible solution to [#1383]

Original (18s):
https://nus-cs2103-ay2021s1.github.io/tp-dashboard/

Demo (4s):
https://gerhean.github.io/tp-dashboard/

```
v-ramp is clearly very expensive to render.  The user
is unlikely to scroll through the entire list of repos 
rendered, making it a waste to render everything.

Let's render v-ramp only when visible, hence
massively reducing initial load time.
```


HCY123902:
Can you show the time taken to load this particular report with and without this change to draw a comparison?

gerhean:
I think it is better for you to test it yourself as there are some visual differences.

dcshzj (approved):
LGTM. RAM usage looks significantly lesser on first load.

frontend/src/index.pug
```diff
@@ -25,6 +25,7 @@ html
     script(src="https://cdn.jsdelivr.net/npm/vue-loading-overlay@3")
     script(src="https://cdn.jsdelivr.net/npm/vue@2.5/dist/vue.js")
     script(src="https://unpkg.com/vuex@3.1.3/dist/vuex.js")
+    script(src="https://unpkg.com/vue-observe-visibility/dist/vue-observe-visibility.min.js")
```
dcshzj, line 28: One thing, can you specify the exact version to use?

dcshzj (changes requested):
As above.

dcshzj (approved):
LGTM!

gerhean:
Actually I was wondering about the purpose of specifying the rampSize in the state previously, if anyone can answer.

fzdy1914:
@damithc Do you think the render on visible OK to you?

damithc:
> @damithc Do you think the render on visible OK to you?

Seems significantly faster. And seems memory usage is less too. Nice. 
1. But this comparison may not be fair as the first dashboard is using a much older version of RepoSense. Can we compare with the current master?
2. Are there any known downsides?

gerhean:
One obvious downside:
image

If you scroll too fast the ramps take about 1s to actually appear.

gerhean:
Actually when you load every single ramp, the filtering becomes slow again.

damithc:
> If you scroll too fast the ramps take about 1s to actually appear (aesthetic factor).

Might be OK if they appear fast enough, without misleading the user into thinking there are no ramps. The alternative is to show some placeholder first e.g., a `loading ...` but not sure if it will make the page slow again. 
BTW, it seems to take longer (even 2-3 seconds at times) for the ramps to appear in my home computer.

fzdy1914:
@gerhean Can we do a mix-up? We first only render the ramps that are visible and display them to the users. Then, we start to render all other ramps.

damithc:
> @gerhean Can we do a mix-up? We first only render the ramps that are visible and display them to the users. Then, we start to render all other ramps regardless of visible or not.

Yes, that feels like best of both worlds.

fzdy1914:
@gerhean Any update?

gerhean:
There's still a few issues to solve like the loading spinner not displaying properly.

PR #1478, opened 2021-03-14
[#1349] Clone test repositories only once when testing

### LINKED ISSUE #1349: Test repositories should be cloned only once when testing
Currently, the test repositories are cloned for every test that performs cloning. This results in lengthy tests as a significant amount of time is spent on cloning the same repository for different tests. This can be reproduced by running `systemTest` to see that the same repositories are cloned multiple times.

The repositories should be cloned only once when running the entire test suite, whether it is for a single test or for multiple tests under the same class, or all the tests. This can reduce the time spent testing and extend the scope of existing tests, especially with regards to testing different time zones. This is to also avoid hitting any rate limiting imposed by GitHub due to high frequency of testing.

dcshzj:
Fixes #1349.

Tests and system tests are treated separately, as they represent different commands within Gradle.

```
The test repositories are cloned for every test that performs
cloning, which results in tests taking a long time to complete as
a significant amount of time is spent cloning the same repository
for every test.

Let's prevent the deletion of the cloned test repositories until
we are sure that all tests have completed, thus allowing the test
repositories to be cloned only once during each test run.
```

### Additional information

The following table represents [the comparison in build times](https://gist.github.com/dcshzj/f5da11ac14b675d311dee9029cf6a3af) before and after this change is implemented:

Test type | Build time before change | Build time after change | Impact
---------- | ----------- | ----------- | -----------
Normal (`test`) | 94s | 52s | Decrease by 44.68%
System test (`systemTest`) | 149s | 127s | Decrease by 14.77%


src/main/java/reposense/git/GitClone.java
```diff
@@ -84,8 +91,15 @@ public static void clone(RepoConfiguration config) throws GitCloneException {
      */
     public static void cloneBare(RepoConfiguration config, String outputFolderName) throws IOException {
         Path rootPath = Paths.get(FileUtil.REPOS_ADDRESS, config.getRepoFolderName());
-        FileUtil.deleteDirectory(Paths.get(rootPath.toString(), outputFolderName).toString());
-        Files.createDirectories(rootPath);
+        Path outputFolderPath = Paths.get(rootPath.toString(), outputFolderName);
+
+        if (!SystemUtil.isTestEnvironment()) {
+            FileUtil.deleteDirectory(outputFolderPath.toString());
+            Files.createDirectories(rootPath);
```
HCY123902, line 98: Is it possible that you can add a test to cover this branch here?

dcshzj: I don't think so, as this branch can only run if it is a non-test environment. The test environment is defined in an environment variable and [it won't be easy to modify it just for one test](https://stackoverflow.com/questions/580085/is-it-possible-to-set-an-environment-variable-at-runtime-from-java).

HCY123902 (approved):
LGTM

src/main/java/reposense/report/RepoCloner.java
```diff
@@ -230,6 +241,12 @@ private boolean spawnShallowPartialCloneProcess(RepoConfiguration config) {
      */
     private boolean waitForCloneProcess(RepoConfiguration config) {
         try {
+            Path repoPath = Paths.get(FileUtil.REPOS_ADDRESS, config.getRepoFolderName(), config.getRepoName());
+
+            if (SystemUtil.isTestEnvironment() && Files.exists(repoPath)) {
+                return true;
```
Tejas2805, line 247: Doesn't this mean that the repo is already cloned? Then why are we returning true for `wait`?

Tejas2805 (commented):
One doubt.

src/main/java/reposense/report/RepoCloner.java
dcshzj: Yes, the repo is already cloned at this stage. The `true` is to indicate that the function is successful, otherwise it will return `false` if an exception has occurred.

dcshzj: @Tejas2805

fzdy1914:
@Tejas2805 You input is needed

src/main/java/reposense/report/RepoCloner.java
dcshzj: @Tejas2805 Ping

fzdy1914:
@Tejas2805 Ping again

Tejas2805 (approved):
LGTM! Sorry for the late reply.

fzdy1914:
Will try to merge it asap.

dcshzj:
@fzdy1914 Will this be merged soon?

fzdy1914:
Can you help check why the deployment fail? @dcshzj

dcshzj:
> Can you help check why the deployment fail? @dcshzj

@fzdy1914 It is likely some weird bug that occurred at random, it should be working now.

PR #1480, opened 2021-03-15
[#1476] Remove the highlighting when the panel is closed

### LINKED ISSUE #1476: Highlighting persists even when tab closed.

https://dashboard-1466-pr-reposense-reposense.surge.sh/

When I close the side tab the highlighting persists.

image

When I close the side tab the highlighting is removed.

HCY123902:
Resolves #1476 

Commit message:
```
Remove highlighting when the panel is closed by the user

The highlighting of the selected tab on the summary 
view is present when the panel is closed.

Removing the highlighting will make the indicated tab 
information more consistent.

Let's remove the highlighting when closing the panel
```


gerhean (approved):
LGTM

frontend/src/static/js/store.js
```diff
@@ -8,6 +8,7 @@ const store = new Vuex.Store({
     fileTypeColors: {},
     loadingOverlayCount: 0,
     loadingOverlayMessage: '',
+    isTabActive: true,
```
0blivious, line 11: shouldn't this set to false by default?

HCY123902: Thanks for pointing it out. Currently, `isTabActive` is set to true by default in the main window during its initialization. It is after a tab is selected and then closed that `isTabActive` will be set to false, so I set `isTabActive` to be true here in the store initially.

0blivious (approved):

frontend/src/static/js/main.js
```diff
@@ -112,14 +112,15 @@ window.app = new window.Vue({
 
       this.isTabActive = true;
       this.tabType = tabName;
-
+      this.$store.commit('updateTabState', this.isTabActive);
```
fzdy1914, line 115: Since the tab state in Vuex is always the same with this.isTabActive. We can just remove this.isTabActive and use the one in Vuex.

fzdy1914 (changes requested):

frontend/src/static/js/main.js
HCY123902: Sure, no problem

fzdy1914 (approved):

fzdy1914:
The highlighting of the selected tab on the summary view is present 
even when the panel is closed.

Removing the highlighting will make the indicated tab information 
more consistent.

Let's remove the highlighting when closing the panel.

Issue #1485, opened 2021-03-18
Error in connection between chart panel and code panel. 


Tejas2805:
Can check the video below. 
1: The file-breakdown doesn't go away from the commits panel though the commit goes.
2: On restoring, the ramp appears first but the commits don't appear. 

https://user-images.githubusercontent.com/35946746/111701934-81322e00-8876-11eb-8e08-d61844cf85b9.mov

gerhean:
Regression caused by #1390.

PR #1488, opened 2021-03-19
[#1485] Make file types selection consistent across panel

HCY123902:
Part of #1485 

Commit message:
```
Synchronize the file types in zoom panel with the file types in
the summary view

The file types in zoom panel currently do not change actively
when the file type filtering condition in the summary view is
changed.

Making the file types on zoom panel change with the file types
in the summary view can enhance the consistency between
different interface components.

Let's synchronize the zoom panel with the summary view.
```

### DIFF
--- a/config/repo-config.csv
+++ b/config/repo-config.csv
@@ -1,9 +1,3 @@
 Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config,Ignore Commits List,Ignore Authors List
 https://github.com/reposense/testrepo-Alpha.git,master,,,,2fb6b9b2dd9fa40bf0f9815da2cb0ae8731436c7;c5a6dc774e22099cd9ddeb0faff1e75f9cf4f151;cd7f610e0becbdf331d5231887d8010a689f87c7;768015345e70f06add2a8b7d1f901dc07bf70582,
-https://github.com/reposense/testrepo-Beta.git,master,fxml,docs**,yes,,
-https://github.com/reposense/testrepo-Beta.git,add-config-json,fxml,docs**,yes,,
-https://github.com/reposense/testrepo-Delta.git,master,override:java;md,,,,
-https://github.com/reposense/testrepo-Delta.git,add-binary-file,,,,,
-https://github.com/reposense/testrepo-Delta.git,nonExistentBranch,,,,,
-https://github.com/reposense/RepoSense.git,master,,,,,
-https://github.com/reposense/testrepo-Empty.git,master,,,,,
+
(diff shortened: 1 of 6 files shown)


gerhean:
Don't really agree, I believe my fix #1490 is cleaner.

fzdy1914 (commented):
PR duplicated with #1490 and #1491 ?

HCY123902:
> PR duplicated with #1490 and #1491 ?

There seems to be some slight difference between #1490 and #1491 in the sense that #1491 restores the original implementation of how zoom tab is updated when `tabZoomInfo` is changed and it restores the original behavior. #1490 synchronizes the change in break down file types with the file types in the zoom panel immediately after the checkbox in break down by file types are ticked or unticked. This change is made because I interpret that #1485 points out a need to synchronize the file types in the summary view and the file types in the panel responsively.

0blivious:
This disconnection was actually the intended design. The idea is, we wish to separate the tab from the summary view and not updating it until we click to open a new zoom panel. Thus, I am not sure if this PR is necessary.

@damithc may I get your opinion on this change?

damithc:
> This disconnection was actually the intended design. The idea is, we wish to separate the tab from the summary view and not updating it until we click to open a new zoom panel. Thus, I am not sure if this PR is necessary.
> 
> @damithc may I get your opinion on this change?

Are we talking about the chart panel and the code panel? Let's use these terms consistently (those are the names used in the UG https://reposense.org/ug/usingReports.html#report-structure), irrespective of the terms used in the code.

Yes, no need to link them up. The more we link different things, the more susceptible they are to break. 

On a related note, it is better to discuss and get the greenlight before implementing a feature. That will reduce the chance of a PR not getting merged down the line.

Tejas2805:
> This disconnection was actually the intended design. The idea is, we wish to separate the tab from the summary view and not updating it until we click to open a new zoom panel. 

If you see the video I have posted for the issue, the error is in the fact, that the commits are removed but the file types are still on display. Therefore it should either be that both are not removed or both are removed together. Removal of just one might cause the user to be confused, is my opinion. And that's why I created an issue calling it a bug.

0blivious:
> If you see the video I have posted for the issue, the error is in the fact, that the commits are removed but the file types are still on display in the Commits Panel when I click to open the Commits Panel. Therefore it should either be that both are not removed or both are removed together. Removal of just one might cause the user to be confused, is my opinion. And that's why I created an issue calling it a bug.

The bug should be handled in #1490. This PR looks more like a new feature to me.

damithc:
> If you see the video I have posted for the issue, the error is in the fact, that the commits are removed but the file types are still on display in the Commits Panel when I click to open the Commits Panel. Therefore it should either be that both are not removed or both are removed together. Removal of just one might cause the user to be confused, is my opinion. And that's why I created an issue calling it a bug.

Yes, that sounds like a bug. Either the panel should be fully disconnected, or fully connected. It's simpler if we keep it disconnected? Is there a significant value in keeping it connected?

0blivious:
> Yes, that sounds like a bug. Either the panel should be fully disconnected, or fully connected. It's simpler if we keep it disconnected? Is there a significant value in keeping it connected?

I don't really see much value in making them connected. Would prefer the fully disconnected one.

fzdy1914:
Closing the PR as the discussion should have reached a conclusion.

PR #1489, opened 2021-03-20
remove duplicate initiate

gerhean:
```
Regression by #1419.

Code was accidentally duplicated.

Let's remove the duplicated code.
```

### DIFF
--- a/frontend/src/static/js/v_zoom.js
+++ b/frontend/src/static/js/v_zoom.js
@@ -118,12 +118,7 @@ window.vZoom = {
 
   methods: {
     initiate() {
-      if (this.info.zUser) {
-        // This code should always run since zUser must be defined
-        this.updateFileTypes();
-        this.selectedFileTypes = this.fileTypes.slice();
-      }
-
+      // This code crashes if info.zUser is not defined
       this.updateFileTypes();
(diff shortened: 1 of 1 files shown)


frontend/src/static/js/v_zoom.js
```diff
@@ -118,12 +118,7 @@ window.vZoom = {
 
   methods: {
     initiate() {
-      if (this.info.zUser) {
-        // This code should always run since zUser must be defined
-        this.updateFileTypes();
-        this.selectedFileTypes = this.fileTypes.slice();
-      }
-
```
fzdy1914, line 126: Why do you remove the need for checking `zUser`? So it may crash now?

gerhean: I think the code crashes regardless if zUser is not defined since the html portion also uses it.

In some previous PR I made it that zUser should be defined the moment the v_zoom.js component is loaded. As such checking for zUser should be redundant as zUser not being defined at the moment which this piece of code runs is a bug. Hence I thought that it would perhaps be better for the code to crash earlier.

fzdy1914: OK

HCY123902 (approved):
LGTM

0blivious (approved):

PR #1490, opened 2021-03-20
[#1485] Watch tab info directly

### LINKED ISSUE #1485: Error in connection between chart panel and code panel.

gerhean:
Resolves #1485
```
Regression by #1390.

Currently zoom tab does not update properly if one of the
zoom watchables do not change.

Let's watch the tabZoomInfo stored in vuex instead of 
zoomOwnerWatchable.

This works as we ensure that every change to tabZoomInfo 
stored in vuex replaces tabZoomInfo with a new object. 

Hence watching the object pointer stored in tabZoomInfo 
is sufficient.

Everything here also applies to the authorship tab with
authorshipOwnerWatchable and tabAuthorshipInfo instead.
```

Changes in behavior:
Clicking on an already highlighted zoom/authorship tab icon on the summary chart will reset the tab even if there is no changes to summary chart.

### DIFF
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -54,7 +54,7 @@ window.vAuthorship = {
       this.updateSelectedFiles();
     },
 
-    authorshipOwnerWatchable() {
+    info() {
       Object.assign(this.$data, initialState());
       this.initiate();
     },
@@ -393,10 +393,6 @@ window.vAuthorship = {
   },
 
   computed: {
-    authorshipOwnerWatchable() {
-      return `${this.info.author}|${this.info.repo}|${this.info.isMergeGroup}`;
-    },
-
     sortingFunction() {
       return (a, b) => (this.toReverseSortFiles ? -1 : 1)
         * window.comparator(filesSortDict[this.filesSortType])(a, b);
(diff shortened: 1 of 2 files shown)


damithc:
Lines in the proposed commit message seem too short i.e., premature line wrapping

gerhean:
I have increased the length thanks.

HCY123902:
Not sure about the original specified behavior. However, unticking a checkbox in the break down file type does not seem to synchronize the corresponding file type in the zoom panel

gerhean:
@HCY123902 I believe that the zoom tab has never actually been responsive to changes in summary-charts until the zoom tab has been reloaded. Hence your suggestions seems more like a new feature? Similarly, changing the date range in summary does not affect the zoom tab.

HCY123902 (approved):

HCY123902 (approved):
I think this pull request can be merged before #1490 because it seems that #1490 is indeed handling a slightly different issue

0blivious (approved):
Please change the PR title according to the DG

PR #1491, opened 2021-03-20
Remove authorshipOwnerWatchable 

gerhean:
Regression by #1390.
```
Currently authorship tab does not update properly if one of the
authorship watchables do not change.

Let's watch the tabAuthorshipInfo stored in vuex instead of 
authorshipOwnerWatchable.

This works as we ensure that every change to tabAuthorshipInfo
stored in vuex replaces tabAuthorshipInfo with a new object. 

Hence watching the object pointer stored in tabAuthorshipInfo
is sufficient.
```
Changes in behavior:
Clicking on an already highlighted authorship tab icon on the summary chart will reset the tab even if there is no changes to summary chart.

### DIFF
--- a/frontend/src/static/js/v_authorship.js
+++ b/frontend/src/static/js/v_authorship.js
@@ -54,7 +54,7 @@ window.vAuthorship = {
       this.updateSelectedFiles();
     },
 
-    authorshipOwnerWatchable() {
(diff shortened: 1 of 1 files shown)


fzdy1914:
Can combine with #1490

PR #1503, opened 2021-03-23
[#1435] Expand continuous integration testing to Ubuntu 18.04 and 20.04

### LINKED ISSUE #1435: Expand integration testing to other environments
Now that #1358 has been resolved, the continuous integration tests run much faster than before. We should explore expanding the testing to other environments such as Windows as well as other versions of Ubuntu and MacOS.

Currently, the test environments are limited to `ubuntu-16.04` and `macos-10.15` and testing using JDK 8.

dcshzj:
Fixes #1435.

Sample run for commits: https://github.com/dcshzj/RepoSense/actions/runs/691897897

The required checks needs to be modified to reflect the change in jobs.

```
The continuous integration test suite only runs on Ubuntu 16.04,
which is a relatively older version.

Let's expand the testing to both Ubuntu 18.04 and 20.04 to make
sure that RepoSense works for both environments as well. This is
done by defining a job matrix that can automatically run the same
build steps for all versions of Ubuntu.
```


gerhean:
Would doing these extra tests would slow down deployment? Furthermore, there is a limit to the amount of free github action minutes for free tier. 
image

Hence I would if possible try to avoid redundant tests, especially since these tests run every time a commit is push to a branch which is being PR.

What kinds of errors are you hoping to catch with these tests? Are there any sources documenting any changes in Java or node runtime behavior across the ubuntu versions which necessitates these extra tests?

dcshzj:
GitHub Actions has free unlimited minutes for public repositories. The jobs run concurrently, so there is very little impact on the time taken for doing these tests. In fact, the slowest job is the frontend tests which takes 5 minutes.

There are environment changes that we hope to catch quickly and early with having more test environments. Otherwise, we would not know of them until someone reports it on our issue tracker. By then I would say it's a little late.

.github/workflows/integration.yml
```diff
@@ -69,25 +72,25 @@ jobs:
       run: time ./gradlew clean checkstyleMain checkstyleTest test systemTest coverage
 
     - name: Run code coverage
-      if: ${{ success() }}
+      if: ${{ success() && matrix.os == 'ubuntu-20.04' }}
```
gerhean, line 75: the `&& matrix.os == 'ubuntu-20.04'` feels a little repetitive, not sure if you can like `if` this once run all these jobs.

dcshzj: I moved it out to a separate job that will only run if the original build job was successful.

gerhean: Do you think that something like matrix.os[last index] should be used instead of 'ubuntu-20.04'?

dcshzj: It is possible, but I think it might be a bit too complicated, because we need to count the number of entries in the build matrix first to determine the index of the last OS. When updating the build matrix in the future, it would be easier to search for occurrences of the string `ubuntu-20.04`.

gerhean: If not the last element, maybe the first element is good?

dcshzj: I think we shouldn't run it on the first element which is not Ubuntu 20.04. This way of specifying ensures that we know the exact OS version that this step is running on.

gerhean (approved):

.github/workflows/integration.yml
```diff
@@ -214,3 +196,52 @@ jobs:
 
     - name: Run frontend tests
       run: ./gradlew frontendTest -Pci
+
+  preview:
```
Tejas2805, line 200: How is this change related to ubuntu integration testing? I can see that you have added `runs-on` which wasn't there before; but you also seem to have modified parts of code.

Tejas2805 (commented):
Left comment

.github/workflows/integration.yml
dcshzj: It is not related to Ubuntu specifically, but should only be run once. Hence, I extracted it out to a different job that will run after the integration tests have been completed. Code-wise nothing has been changed, only the `if` statements have been consolidated.

Tejas2805: I will prefer to see this change in a new PR or mention this in the commit message. I agree it's a small tweak but PRs should always just perform one specific change. This helps in keeping track and finding PRs and commits if ever required.

Tejas2805 (changes requested):

.github/workflows/integration.yml
dcshzj: @Tejas2805 I have reverted the PR back to e4e832d, as the original design did not accompany this change.

Tejas2805 (approved):
LGTM!

fzdy1914 (changes requested):
Feeling that instead of expanding to more ubuntu versions, it is more urgent to have a windows test environment.

dcshzj:
> Feeling that instead of expanding to more ubuntu versions, it is more urgent to have a windows test environment.

AppVeyor currently does our Windows integration tests. Should we move it to GitHub Actions?

damithc:
> AppVeyor currently does our Windows integration tests. Should we move it to GitHub Actions?

Pros: everything on one platform
Cons: everything on one platform i.e., one point of failure :-p

fzdy1914:
OK, then, I will merge this PR soom.

fzdy1914:
@dcshzj 
image

Why did the old two actions fail?

dcshzj:
> Why did the old two actions fail?

The test names have been renamed. Previously it is called "Ubuntu 16.04 JDK 8" but now it is "ubuntu-16.04 JDK 8". Same for the macOS one.

PR #1511, opened 2021-03-29
Upgrade minimatch version

gerhean:
Regression by #1482,

```
minimatch@1.0.0 is deprecated and is causing some problems.

Lets upgrade to minimatch@3.0.4.
```

### DIFF
--- a/frontend/src/index.pug
+++ b/frontend/src/index.pug
@@ -19,7 +19,6 @@ html
     script(src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.auto.js")
     script(src="https://use.fontawesome.com/releases/v5.0.13/js/all.js", defer=True)
 
-    script(src="https://cdn.jsdelivr.net/npm/minimatch@1.0.0/minimatch.min.js")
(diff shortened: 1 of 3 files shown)


fzdy1914 (approved):

0blivious (approved):

HCY123902 (approved):

dcshzj:
Are we able to continue using jsDeliver instead? Using skypack is a deviation from what we have been using previously.

https://cdn.jsdelivr.net/npm/minimatch@3.0.4

gerhean:
So skypack is I think the easiest way to use ES6 modules in the current project.

First understand that there are two types of js modules, CommonJS and ES6 modules. ES6 modules are originally made for nodejs, but since some of them are useful and have no CommonJS alternatives like minimatch, we run into some problems where the browser doesn't support some functions of ES6 modules.

So in order to mitigate this issue, people use js module bundlers like browsersify, which I removed because it was kinda not being maintained properly (https://github.com/vuejs/vueify).

Skypack is just an easy plug and forget thing so naturally I used it.

image
I think this sums it up nicely.

dcshzj:
I see, thanks for the summary! No other comments from me.

dcshzj (approved):

PR #1516, opened 2021-03-31
[#1515] UG: update documentation on using RepoSense at a specific commit

### LINKED ISSUE #1515: run.sh: Support using a version at a specific commit
See https://reposense.org/ug/runSh.html

Current: We can configure run.sh to use a specific tag (as long as there is a matching RepoSense release).

Suggested: Is it possible to support running RepoSense at a specific commit?

Reason: If the master branch breaks, we can use this feature to freeze run.sh at the commit immediately before the broken commit until we can use the latest tip of the branch again. Currently, we have to fall back on the released version which may be many commits behind the master branch.

dcshzj:
Associated pull request: https://github.com/reposense/publish-RepoSense/pull/10
Fixes #1515.

```
The script to obtain RepoSense does not support a way to use
RepoSense at a specific commit.

Let's update the documentation to showcase the functionality for
checking out and using RepoSense at a specific commit.
```


docs/ug/runSh.md
```diff
@@ -27,4 +27,5 @@ You can update the RepoSense command (i.e., the last line) in the `run.sh` to ma
 Depending on which version you wish to use for report generation, add one of the following flags to the line `./get-reposense.py` in `run.sh` (e.g., `./get-reposense.py --release`):
 * `--release`: Use the latest release (Stable)
 * `--master`: Use the latest version of the master branch
-* `--tag TAG` e.g., `--tag v1.6.1`: use the version identified by the git tag given
+* `--tag TAG` (e.g. `--tag v1.6.1`): Use the version identified by the Git tag given
+* `--commit COMMIT` (e.g. `--commit abc123`): Use the version identified by the Git commit SHA given
```
HCY123902, line 31: Actually, from the perspective of the user, will the user be able to know the difference in behavior of each commit version?

dcshzj: I think it really depends on the commits that the user is skipping, but at the bottom of the RepoSense report, there is a link to the commit version that the RepoSense report is built on.

image

HCY123902 (approved):
LGTM

damithc:
feature, docs, and tests are best put in the same PR as they all have to merge at the same time, preferably in the same merge commit. Otherwise it will be hard to revert the feature later if we wanted to remove it.

0blivious (approved):

fzdy1914:
```
The script to obtain RepoSense in `publish-RepoSense` supports a way 
to use RepoSense at a specific commit. 

Let's update the documentation to showcase the functionality for
checking out and using RepoSense at a specific commit.
```

PR #1518, opened 2021-03-31
[#1184] Test: improve assertion of authors for file analysis

### LINKED ISSUE #1184: GitTestTemplate: Method of asserting file analysis correctness can be improved
Currently, it is difficult to change the test files used in some tests of `FileAnalyzerTest`.

Reason: the underlying method `assertFileAnalysisCorrectness` only asserts whether each line in the test file is / is not attributed to a dummy author called `FAKE_AUTHOR`

It doesn't actually assert against the expected author of each line

Relevant snippet:
 https://github.com/reposense/RepoSense/blob/9609c591f7e3a4c8d3a1f8acf2f71f7aec15b043/src/test/java/reposense/template/GitTestTemplate.java#L127-L139

Proposed solution:
Let's generalise the `assertFileAnalysisCorrectness` so that we can change the test files more easily.

Within `AnnotatorAnalyzerTest` , there **already exists a similar method** which may be more suitable than `assertFileAnalysisCorrectness` and **could be used in its place**

https://github.com/reposense/RepoSense/blob/9609c591f7e3a4c8d3a1f8acf2f71f7aec15b043/src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java#L69-L82

Shall we replace `assertFileAnalysisCorrectness` with `assertAnnotationAnalysisCorrectness`  ?
(replace the logic, but keep the method name)

dcshzj:
Fixes #1184.

```
The assertion used in checking the correctness of the attribution
of authors is not entirely robust and can be prone to other errors
which may not be detected.

Let's refactor the assertion used in file and annotation analysis
such that they use the same function and make the tests more
robust.
```


HCY123902 (approved):
LGTM

0blivious (approved):

Tejas2805 (approved):
Good neat work!

PR #1523, opened 2021-04-03
UG: reflect new method of CSV parsing

vig42:
```
Currently, the User Guide states that "RepoSense ignores the first row (i.e.,
column headings) of CSV config files". However, as of #1459, this is no longer
true.

Let's change the User Guide to reflect the new method of CSV parsing.
```


docs/ug/configFiles.md
```diff
@@ -15,7 +15,7 @@ Given below are the details of the various config files used by RepoSense.
 
 <box type="info" seamless>
 
-**RepoSense ignores the first row (i.e., column headings) of CSV config files.** It is used simply to provide more information to human readers. This also means the ==columns in your config files should be in the exact order specified here==.
+**RepoSense ignores the order of column headers in the CSV config files. However, the spelling of the column headers must match the specified headers exactly.**
```
dcshzj, line 18: LGTM so far. I think can add the phrase "Otherwise, the column will be ignored when the report is generated" or something similar.

dcshzj (changes requested):

docs/ug/configFiles.md
vig42: I have changed it to "However, the spelling of the column headers must match the specified headers exactly, or the column would be ignored by RepoSense."

dcshzj (approved):
LGTM!

fzdy1914:
```
User Guide states that "RepoSense ignores the first row (i.e., 
column headings) of CSV config files". However, as of #1459, this is 
no longer true.

Let's change the User Guide to reflect the new method of CSV 
parsing.
```

Issue #1506, opened 2021-03-23
Fail gracefully when an author has a large number of files


damithc:
Similar to #1504

vig42:
One possible consequence of this is that the `authorship.json` file could exceed the limit of 100MB imposed by GitHub, which means that the dashboard cannot be deployed using GitHub Actions.

This bug can be observed when running RepoSense on the [servo](https://github.com/servo/servo) repository, using GitHub Actions.

PR #1524, opened 2021-04-03
Add warning when there are a large number of files to be analyzed

vig42:
```
Currently, if RepoSense is run on a repo with a large number of text files
(with a large number of lines overall), it can take a long time to generate
the authorship report. Additionally, the size of the authorship.json file can
be very large, possibly even exceeding the file size limit of 100 MB for
GitHub, which would prevent the report from being deployed on GitHub Pages.

Let's add a warning when there are a large number of text files with a large
number of lines overall. This would let users know the reason why they may
experience slow performance and would also advise users on how they can reduce
the number of files to be analyzed using repo-config.csv.
```


src/main/java/reposense/authorship/AuthorshipReporter.java
```diff
@@ -37,4 +60,28 @@ public static AuthorshipSummary generateAuthorshipSummary(RepoConfiguration conf
...
+        try {
+            reader = new BufferedReader(new FileReader(path));
+        } catch (FileNotFoundException fnfe) {
+            return 0;
+        }
+        int lines = 0;
+        try {
+            while (reader.readLine() != null) {
+                lines++;
+            }
```
dcshzj, line 80: This may be quite time consuming as it loads every file once to count the number of lines. Instead, I found [a way to count total number of lines in Git](https://stackoverflow.com/questions/4822471/count-number-of-lines-in-a-git-repository/4822890#4822890). It uses the following command:

```shell
git diff --shortstat `git hash-object -t tree /dev/null`
```

It's a Unix command but I think it should be possible to generalize to Windows as well. Using the [servo](https://github.com/servo/servo) repository, it seems to run reasonably fast:

```shell
$ time git diff --shortstat `git hash-object -t tree /dev/null`
 101260 files changed, 6570287 insertions(+)

real    0m2.601s
user    0m1.887s
sys     0m0.715s
```

Perhaps we can consider using this method instead?

dcshzj (changes requested):

gerhean:
Instead of counting the lines why not just get the total size of the repo being downloaded?

src/main/java/reposense/authorship/AuthorshipReporter.java
vig42: I tried implementing this `git diff` based method in RepoSense. I was able to get a Windows-compatible version by splitting this into 2 commands, first using a `git hash-object` command to get the hash and then running `git diff`.

I timed both methods within Java using `System.nanotime()`. Based on this, it seems like the `git diff` method is much slower - on my Windows 10 machine, on the servo repository, `git diff` took around 20 seconds on average, compared to less than 1 sec using the current method of reading every file.

Additionally, the `git diff` method counts the lines in every single file in the repository. However, the current method only looks at the files that are going to be analyzed for the authorship report. Since users can filter the files that should be included in the authorship report (e.g. based on the file format) I feel that the current method would provide a more accurate count.

vig42:
> 
> 
> Instead of counting the lines why not just get the total size of the repo being downloaded?

The main reason for the slow performance here is because RepoSense runs a `git blame -w --line-porcelain` command on each of the text files. This command produces some output for every line in the file. So the overall time taken is a function of the number of files to be analyzed and the total number of lines in the files.

It is possible to have a large sized repo without having a lot of text files, e.g. if there are large binary files in the repo. In this case, we would not experience slow performance when extracting and analyzing the file info.

If you want to reduce the overhead, rather than counting the lines, we could simply use the number of files to be analyzed, i.e. the size of the `textFileInfos` list here:

https://github.com/reposense/RepoSense/blob/5d0768d1c6df9925b89527ee0253ae86e53f0475/src/main/java/reposense/authorship/AuthorshipReporter.java#L37-L39

This would be a little less accurate, but I feel that it could still provide a good gauge. However, the amount of overhead is also not very significant; e.g. on my local machine, it took less than 1 second to count all the lines on the servo repository, even though there were 11.5 thousand files and 1.5 million lines.

src/main/java/reposense/authorship/AuthorshipReporter.java
dcshzj: I think it is a good point that there may be some files that are filtered, but it is provided that users specify the file globs to ignore. 

> I timed both methods within Java using `System.nanotime()`. Based on this, it seems like the `git diff` method is much slower - on my Windows 10 machine, on the servo repository, `git diff` took around 20 seconds on average, compared to less than 1 sec using the current method of reading every file.

On this point, I tried to reproduce this on my Windows machine using PowerShell and on Intellij (with SSD storage). I got the following output for PowerShell:

```powershell
PS C:\Users\user\Documents\servo> Measure-Command {git diff --shortstat 4b825dc642cb6eb9a060e54bf8d69288fbee4904}


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 10
Milliseconds      : 965
Ticks             : 109652154
TotalDays         : 0.000126912215277778
TotalHours        : 0.00304589316666667
TotalMinutes      : 0.18275359
TotalSeconds      : 10.9652154
TotalMilliseconds : 10965.2154
```

I took a snippet of the output log from Intellij:
```
03:31:27 - [7/8] Analyzing https://github.com/servo/servo.git (master)...
03:32:39 - https://github.com/servo/servo.git (master) does not contain a standalone config file.
03:32:39 - https://github.com/servo/servo.git (master) has no authors specified, using all authors by default.
03:32:41 - [WARNING] Duplicate alias camelid found. The alias will belong to the last author - camelid
03:32:41 - [WARNING] Duplicate alias camelid@users.noreply.github.com found. The alias will belong to the last author - camelid
03:32:41 - [WARNING] Duplicate alias cybai found. The alias will belong to the last author - cybai
03:32:41 - [WARNING] Duplicate alias cybai@users.noreply.github.com found. The alias will belong to the last author - cybai
03:32:41 - Extracting commits info for https://github.com/servo/servo.git (master)...
03:50:00 - Analyzing commits info for https://github.com/servo/servo.git (master)...
03:50:04 - Extracting relevant file info from https://github.com/servo/servo.git (master)...
03:50:52 - [WARNING] There are a large number (97401) of text files to be analyzed, comprising a total of 6571182 lines. As such, RepoSense may take a long time to generate the report. Performance may be improved by setting values for the "File formats" and  "Ignore Glob List" columns in the repo-config.csv so as to reduce the number of files to be  analyzed.
```

The code on Intellij is running the branch for this PR, and it takes about 48 seconds compared to the `git diff` method of about 10 seconds. Can I check how you are able to get less than 1 second using this method that you are proposing?

dcshzj (commented):
Figures aside, my concern with this implementation is that it runs for every repository to be analysed. It is likely rare for us to encounter repositories that are as large as the servo repository we are testing with. However, CS2103 has an increasing enrollment number, which means that the number of repositories to be analysed will increase, which adds to the report generation time. Most of them will be under the threshold proposed in this PR, but we still incur the cost of calculating the number of lines.

Is it possible to adopt a cheaper, albeit less accurate figure for determining the number of lines? I think @gerhean proposed a good idea of getting the size of the repository. Although the point on having binary files is valid as well, I think it is alright for us to get an estimate first and if the repository is beyond a certain size, then we can go ahead with counting the actual number of lines to analyse.

src/main/java/reposense/authorship/AuthorshipReporter.java
vig42: I think the 48 seconds figure is not accurate. The "Extracting relevant file info..." log appears at the beginning of the `extractTextFileInfos` method in `FileInfoExtractor`, which means that the 48 seconds includes all the time taken to generate the whole list of `FileInfo` objects.

To get a more accurate measurement, I put a couple of logs at the start & end of both the `git diff` and the function that reads the files.

```
05:22:14 - [1/1] Analyzing https://github.com/servo/servo.git (master)...
05:24:00 - https://github.com/servo/servo.git (master) does not contain a standalone config file.
05:24:00 - https://github.com/servo/servo.git (master) has no authors specified, using all authors by default.
05:24:00 - Extracting commits info for https://github.com/servo/servo.git (master)...
05:24:36 - Analyzing commits info for https://github.com/servo/servo.git (master)...
05:24:37 - Extracting relevant file info from https://github.com/servo/servo.git (master)...
05:24:48 - [WARNING] "tests/wpt/web-platform-tests/origin-policy/policies/op10 cspfp-valid.json      " is an invalid file path. Skipping this directory.
05:24:48 - [WARNING] "tests/wpt/web-platform-tests/origin-policy/policies/op2 cspfp-double-top-level.json    " is an invalid file path. Skipping this directory.
05:24:48 - [WARNING] "tests/wpt/web-platform-tests/origin-policy/policies/op5 cspfp-non-object.json  " is an invalid file path. Skipping this directory.
05:24:48 - [WARNING] "tests/wpt/web-platform-tests/origin-policy/policies/op6 cspfp-non-string.json  " is an invalid file path. Skipping this directory.
05:24:49 - Starting GitDiff
05:25:06 - GitDiff complete, 6570263 lines found
05:25:06 - Starting to read files
05:25:07 - Reading files complete, 1701913 lines found
05:25:07 - [WARNING] There are a large number (11584) of text files to be analyzed, comprising a total of 1701913 lines. As such, RepoSense may take a long time to generate the report. Performance may be improved by setting values for the "File formats" and  "Ignore Glob List" columns in the repo-config.csv so as to reduce the number of files to be  analyzed.
```

Based on this, the `git diff` method took about 17 sec, and the other method took about 1 sec.

vig42:
> 
> Is it possible to adopt a cheaper, albeit less accurate figure for determining the number of lines? I think @gerhean proposed a good idea of getting the size of the repository. Although the point on having binary files is valid as well, I think it is alright for us to get an estimate first and if the repository is beyond a certain size, then we can go ahead with counting the actual number of lines to analyse.

Actually, I realized there's a better way to count the lines. Currently, the `FileInfoExtractor` already reads all of the files to be analyzed, and keeps track of all the lines. This means we can just add a method to return the number of lines in each `FileInfo` object:

https://github.com/reposense/RepoSense/blob/2eb98dc8a25bfa7db787cb6e9a7a88c66c7f17ba/src/main/java/reposense/authorship/model/FileInfo.java#L45-L47

This provides an accurate count of the number of lines at basically no extra cost.

src/main/java/reposense/authorship/AuthorshipReporter.java
```diff
@@ -2,24 +2,42 @@
...
+            + " analyzed.";
 
     /**
      * Generates and returns the authorship summary for each repo in {@code config}.
      */
     public static AuthorshipSummary generateAuthorshipSummary(RepoConfiguration config) {
         List<FileInfo> textFileInfos = FileInfoExtractor.extractTextFileInfos(config);
 
+        int numFiles = textFileInfos.size();
+        int totalNumLines = textFileInfos.stream().mapToInt(fileInfo -> fileInfo.getNumOfLines()).sum();
```
dcshzj, line 35: Great! This looks much neater now. Can you split this into multiple lines, like this?
```suggestion
        int totalNumLines = textFileInfos.stream()
                .mapToInt(fileInfo -> fileInfo.getNumOfLines())
                .sum();
```

dcshzj (changes requested):

dcshzj (approved):
LGTM!

However, the issue of having authorship.json being larger than 100 MB will still be present, but that should be worked on in another PR, as it is not trivial. @vig42 are you able to create an issue to take note of this?

vig42:
> 
> 
> LGTM!
> 
> However, the issue of having authorship.json being larger than 100 MB will still be present, but that should be worked on in another PR, as it is not trivial. @vig42 are you able to create an issue to take note of this?

Would this be covered by issue #1506? I had previously left [a comment regarding the 100 MB issue](https://github.com/reposense/RepoSense/issues/1506#issuecomment-810748603).

dcshzj:
> Would this be covered by issue #1506? I had previously left [a comment regarding the 100 MB issue](https://github.com/reposense/RepoSense/issues/1506#issuecomment-810748603).

Yep, I missed that. Thanks!

0blivious (approved):

Tejas2805 (commented):
LGTM!

Just one question. On what basis has the current threshold been set?

PR #1527, opened 2021-04-06
Remove `window.app` state changes and remove `this.` in .pug

gerhean:
```
Currently assignment to window.app is being done in api.js and
authorship.js to change the state of the topmost Vue component.

These state changes are hard to track and breaks projectify-frontend.

Let's remove all property changes to window.app.

Using this. in template files is unnecessary and causing some
problems in projectify-frontend.

Let's remove all this. in .pug files.

PR #1480 included some changes to how isTabActive is handled,
which can be optimized further.

Let's move hashing of tabOpen to store.js.
Let's have v_resizer get isTabActive directly from Vuex.
```


HCY123902 (approved):

frontend/src/static/js/v_zoom.js
```diff
@@ -130,6 +130,10 @@ window.vZoom = {
       this.selectedFileTypes = this.fileTypes.slice();
     },
 
+    getFontColor(color) {
+      return window.getFontColor(color);
+    },
+
```
0blivious, line 136: is `this.getFontColor` really big trouble? seems now you have to duplicate the common function for all subcomponent.

gerhean: A problem for projectify frontend, since `this.` does use the variables from `window.`

If the duplication is big trouble I can try using mixins.

0blivious: Yes, could you please try to use mixins for this? If we really gonna projectify this project properly, mixin will eventually be required.

gerhean: Actually after trying mixins, I don't think they are a good idea. Manually importing the required functions in would probably be a better choice for easier maintenance. I did change the implementation a little to make it less wordy though.

0blivious (approved):

frontend/src/static/js/main.js
```diff
@@ -136,7 +144,7 @@ window.app = new window.Vue({
       if (Object.keys(info).length === tabInfoLength) {
         this.$store.commit('updateTabAuthorshipInfo', info);
       } else if (hash.tabOpen === 'false' || tabInfoLength > 2) {
-        window.app.isTabActive = false;
```
fzdy1914, line 139: Recently, we changed `isTabActive` to a computed value from Vuex, can you do a check if the `isTabActive` here is affected or not?

fzdy1914 (changes requested):

frontend/src/static/js/main.js
gerhean: Which PR was it?

fzdy1914: https://github.com/reposense/RepoSense/pull/1480/files Can you check this PR, it is good to fix it once and for all.

gerhean: Made some changes to how it works now.

gerhean:
@fzdy1914 I need this to be merged for projectify frontend

fzdy1914 (approved):

PR #1530, opened 2021-04-08
[#1234] Move relevant test cases in ArgsParserTest to RepoSenseTest class

### LINKED ISSUE #1234: ArgsParserTest: Some tests should be moved to separate file, as they test logic in RepoSense driver class
The majority of the test cases in `ArgsParserTest` test whether the `CliArguments` object is correctly created by `ArgsParser` , in various situations.

However, there are a few tests which test whether the results of processing the `CliArguments` object in RepoSense's driver class are correct.

I am referring to the following tests:

- `parse_validGitRepoLocations_repoConfigurationListCorrectSize` - This verifies whether the RepoSense driver logic obtains the correct number of `RepoConfigs` when it processes the `CliArguments` object. This test indirectly involves the `RepoLocation` too.

-  `parse_configOrLocationsSimilar_success` - This verifies whether the RepoSense driver logic obtains the same `RepoConfigs`, regardless if the `CliArguments` was built using the `--config` argument or the `--repos` argument. This test indirectly involves `RepoConfigCsvParser` and `AuthorConfigCsvParser`.

- `parse_invalidRepoLocation_emptyRepoConfigurationList` - This verifies that the RepoSense driver logic ends up with an empty list of `RepoConfigs` when all repo locations in the `CliArguments` are invalid.

I think we should move these tests into a separate test file that is specifically meant for testing logic in the RepoSense driver class (i.e. `RepoSense.java`).

HCY123902:
Resolves  #1234 

Commit message:
```
Move the relevant test cases in ArgsParserTest to RepoSenseTest

The relevance of some test cases currently in ArgsParserTest 
is more on testing the behavior of methods whose main logic stays 
in the RepoSense class, such as the getRepoConfiguration method 
with LocationsCliArguments, rather than testing the behavior of 
ArgsParser.

Moving these test cases to RepoSenseTest will make the test cases
more organized. Additionally, when new methods are added to
the driver RepoSense class, there can be new test cases added.

Let's move the relevant test cases to a new RepoSenseTest class


gerhean (approved):
LGTM

dcshzj (commented):
I don't agree that we should move them to a separate class, but I leave it for the seniors to give their inputs.

Also, do assign the tasks to yourselves if you are intending to work on it, or if you have already submitted a PR. This makes the issue tracker easier to navigate.

0blivious:
Judging by the name and the aspects to be tested, I feel we should still assign them to ArgParser class. @fzdy1914 @Tejas2805

Tejas2805 (commented):
A majority of the testing involves the ArgsParserClass, therefore don't think this migration is necessary. @fzdy1914 your opinion?

fzdy1914:
Agree that it is trivial to do so. I would like to better classify it as ArgsParserTest also.

PR #1531, opened 2021-04-08
[#435] Convert special character in path to the corresponding white space

### LINKED ISSUE #435: Files that have whitespace in their paths lead to test failures
**Bug:**
For example, the following exception is encountered due to whitespace in **`Abc Xyz`**, which is in the file path to the test file `standardBlameOutput.txt`. 

```java 
   Path STRINGS_UTIL_TEST_DIRECTORY = ...

20 Path gitBlameStandardOutputFile =
21                Paths.get(STRINGS_UTIL_TEST_DIRECTORY.toString(), "standardBlameOutput.txt");
22 List<String> lines = Files.readAllLines(gitBlameStandardOutputFile);
```

```java
java.nio.file.NoSuchFileException: C:\Users\Abc%20Xyz\git\RepoSense\out\test\resources\StringsUtilTest\standardBlameOutput.txt

	at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:79)
	...
	at java.nio.file.Files.readAllLines(Files.java:3242)
	at reposense.util.StringsUtilTest.filterText_standardInput_success(StringsUtilTest.java:22)
```
**Cause:**
The exception occurs since `Path.toString()` returns the path with the spaces replaced by `%20`. But, `java.io.File` accepts spaces in the `String pathname`, not `%20`.

**Fix:**
I think it can be fixed by replacing such instances by 
` String decodedPath = URLDecoder.decode(Path.toString(), "UTF-8");`

HCY123902:
Resolves #435 

Commit message:
```
Convert the special character in path to corresponding white space

The white space in the path generated from the loadResource 
method may be replaced by %20 when the path is converted to
the corresponding string. This issue when the path string is used 
to open files and read content.

Adding a method to convert path to appropriate string with
the original white space present can make the opening of files
and testing less vulnerable to the local project directory.

Let's convert special character in path to corresponding white space
```


gerhean (commented):
LGTM but maybe resolve the code not covered by tests.

HCY123902:
Thanks for the suggestion. I have changed it accordingly

PR #1501, opened 2021-03-23
[#634] Support defining sinceDate, untilDate and period in report config

dcshzj:
Fixes #634.
Part of #1360.

```
The new report-config.json allows for the configuration of the
report generation. However, only the report title can be
configured at this time.

Let's expand the functionality of the report config file by
allowing sinceDate, untilDate and period to be defined as well.
If the CLI arguments were to contain the same settings as those in
the report config, then the CLI arguments will take precedence.
```

## Summary of changes
- The `report-config.json` file now supports setting `sinceDate`, `untilDate` and `period`. The equivalent CLI arguments are `--since`, `--until` and `--period` respectively. (this PR)
- The logic for getting the ReportConfiguration object has been moved to ArgsParser. (in PR #1534)
- DateTime functionality that previously existed in ArgsParser has now been moved to TimeUtil to reduce code duplication. (in PR #1535)

## :warning: Requires
- [x] PR #1534 to be merged first
- [x] PR #1535 to be merged first

Issue #1360, opened 2020-11-16
Report-config.json: Support more fields


Boon-Jun:
We have added a report-config.json file that can be used for the configuration of report generation. 
Right now, it only allows users to specify a report title.

We can support more fields with this config file, especially if they can be applied to all repositories during report generation.
The following are some of the fields I think we can include:
- [ ] sinceDate
- [ ] untilDate  
- [ ] timeZone

dcshzj:
We can also extend it to configure the introductory text as outlined in #467.

Boon-Jun:
I think we have to what extent should we allow the customization of the introductory text. If its just basic text replacement, then it can be done, but if the users are allowed to modify the html themselves, then it might not be suitable

damithc:
I think it would be support basic HTML but if that's too hard, even just text is good enough. We also need to figure out where we are going to display that text, in what format.

dcshzj:
Related task for setting `sinceDate` and `untilDate`: #634

logical-1985516:
Hi, may I check if this issue is resolved? If not, may I try this issue?

ckcherry23:
Hi @logical-1985516, I think this issue is not so relevant anymore as one of our ongoing PRs https://github.com/reposense/RepoSense/pull/2192 might make it irrelevant when merged, depending on the implementation. I'd recommend working on other issues for now, thanks!

Issue #634, opened 2019-03-28
Allow sinceDate and untilDate to be configured in config files


yong24s:

damithc:
It's a bit awkward to specify key-value pairs in a csv file though. Might have to add another json file for this purpose.

anubh-v:
This would allow users to easily compare git activity for repositories that were active over different time periods.

i.e. Repository A was actively developed 1 year ago, while Repository B has only been under development for the past month.

With regards to specifying the `sinceDate` / `untilDate` in the `repo-config.csv` file, I think we could simply add a "Data range" column to `repo-config.csv`, as shown below?

Repository's Location | Branch | File formats | Ignore Glob List | Ignore standalone config | Ignore Commits List | Date range
-- | -- | -- | -- | -- | -- | --
https://github.com/foo/bar.git | master | override:java;css | test/** | yes |  | 10/10/2018;10/12/2018


@damithc , I'm not clear what you meant by the "difficulty of specifying key-value pairs in a csv". 

Does the above proposal for the `repo-config.csv` seem suitable?

damithc:
If the value can differ across different repos, a csv file is the way to go.
But I'm not sure we can handle different periods for different repos. The current ReposSense ramp chart assumes a single timeline.

anubh-v:
Thanks for the response @damithc 
Yes, I overlooked that `sinceDate` and `untilDate` are *report-level* parameters, and not *repository/branch* level.

Perhaps these parameters can be configured in the ReportConfiguration that is being considered in PR #627.

FH-30:
Hi, can I try working on this issue?

damithc:
> Hi, can I try working on this issue?

Go ahead @FH-30

dcshzj:
Sorry @FH-30 @damithc I am currently working on this in #1501. This issue is currently assigned to myself.

dcshzj:
@FH-30 I have closed my PR as it is not progressing. Feel free to work on it if you wish.

FH-30:
Thanks for notifying @dcshzj , I would like to work on this issue!

dcshzj:
@FH-30 Please do so. You may refer to the original PR I have created if you wish to.

yhtMinceraft1010X:
I propose closing this issue since it has been encapsulated in #1360

PR #1534, opened 2021-04-10
[#634] Move logic for getting ReportConfiguration to ArgsParser

dcshzj:
Part of #634.
Part of #1360.
Part of #1501.

```
The current design of getting the ReportConfiguration object in
the main RepoSense class is flawed and the design is not ideal,
as it meant that configuration has to be set and parsed twice.

Let's move the logic for getting the ReportConfiguration object
into ArgsParser to avoid having to duplicate configuration parsing
in the future.
```

### DIFF
--- a/src/main/java/reposense/RepoSense.java
+++ b/src/main/java/reposense/RepoSense.java
@@ -10,8 +10,6 @@ import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-import com.google.gson.JsonSyntaxException;
-
 import net.sourceforge.argparse4j.helper.HelpScreenException;
 import reposense.model.AuthorConfiguration;
 import reposense.model.CliArguments;
@@ -29,7 +27,6 @@ import reposense.parser.InvalidCsvException;
 import reposense.parser.InvalidLocationException;
 import reposense.parser.ParseException;
 import reposense.parser.RepoConfigCsvParser;
-import reposense.parser.ReportConfigJsonParser;
 import reposense.report.ReportGenerator;
 import reposense.system.LogsManager;
 import reposense.system.ReportServer;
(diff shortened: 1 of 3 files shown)


src/main/java/reposense/RepoSense.java
```diff
@@ -62,7 +58,7 @@ public static void main(String[] args) {
                 return;
             } else if (cliArguments instanceof ConfigCliArguments) {
                 configs = getRepoConfigurations((ConfigCliArguments) cliArguments);
-                reportConfig = getReportConfigurations((ConfigCliArguments) cliArguments);
```
gerhean, line 65: If configuration has to be set and parsed twice, does it mean that this code here is run twice? Why so?

dcshzj: This PR needs to be viewed in the context of #1501. The configuration parsing would have to be done twice if we kept this function inside RepoSense.java.

Putting them together here allows us to get information from both sources, then do a final output (something like a parse-parse-set instead of parse-set-parse-set).

For this function in particular, the configuration does not get parsed twice. It’s just retrieving the ReportConfiguration object, which has already been prepared earlier in ArgsParser.parse().

HCY123902:
Actually, given that the responsibility of `ArgsParser` is to extract argument from the command, can you more justification for making this refactoring and adding additional responsibility of JSON file parsing to the class?

dcshzj:
> Actually, given that the responsibility of `ArgsParser` is to extract argument from the command, can you give more justification for making this refactoring that adds additional responsibility of JSON file parsing to the class?

Please refer to my comment above.

Tejas2805 (approved):
LGTM!

fzdy1914:
Good PR.

PR #1535, opened 2021-04-10
[#634] Extract time-based functionalities to TimeUtil

dcshzj:
Part of #634.
Part of #1360.
Part of #1501.

```
Many of the time-based functionality and logic are scattered
about in various files, which increases code duplication and is a
maintenance burden.

Let's abstract all time-based functionalities into the TimeUtil
class such that multiple places in the code can reuse the same
logic.
```

### DIFF
--- a/src/main/java/reposense/parser/DateArgumentType.java
+++ b/src/main/java/reposense/parser/DateArgumentType.java
@@ -1,50 +1,24 @@
 package reposense.parser;
 
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Optional;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import net.sourceforge.argparse4j.inf.Argument;
 import net.sourceforge.argparse4j.inf.ArgumentParser;
 import net.sourceforge.argparse4j.inf.ArgumentParserException;
 import net.sourceforge.argparse4j.inf.ArgumentType;
+import reposense.util.TimeUtil;
 
 /**
  * Verifies and parses a string-formatted date to a {@code Date} object.
  */
 public class DateArgumentType implements ArgumentType<Optional<Date>> {
-    protected static final String DATE_FORMAT_REGEX =
-            "^((0[1-9]|[12][0-9]|3[01])\\/(0[1-9]|1[012])\\/(19|2[0-9])[0-9]{2})";
     private static final String PARSE_EXCEPTION_MESSAGE_INVALID_DATE_STRING_FORMAT = "Invalid Date: %s";
-    private static final DateFormat CLI_ARGS_DATE_FORMAT = new SimpleDateFormat("dd/MM/yyyy HH:mm:ss");
-
-    static {
-        /*
-         * Setting setLenient to false prevents unexpected results.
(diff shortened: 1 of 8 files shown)


Tejas2805 (changes requested):
LGTM! Liked the tests added too.

Just a quick question. What about the other tests? The ones for the already implemented functions. Should migrate them to TimeUtilTest too for completeness.

dcshzj:
> Just a quick question. What about the other tests? The ones for the already implemented functions. Should migrate them to TimeUtilTest too for completeness.

Are you referring to functions like `getDateMinusAMonth` and `getDatePlusNDays`? Those currently do not have standalone unit tests, but are covered as part of the `ArgsParserTest` class.

src/main/java/reposense/parser/DateArgumentType.java
```diff
@@ -1,50 +1,24 @@
...
-     * Extracts the first substring that matches the {@code DATE_FORMAT_REGEX}.
-     */
-    protected String extractDate(String date) {
-        Matcher matcher = Pattern.compile(DATE_FORMAT_REGEX).matcher(date);
-        String extractedDate = date;
-        if (matcher.find()) {
-            extractedDate = matcher.group(1);
-        }
-        return extractedDate;
-    }
```
gerhean, line 42: Don't really agree with moving this code into TimeUtil since the date extraction seems like a parsing thing. Also this file is already quite short. TimeUtil seems to be for stuff related to Date objects, but this takes in a string and outputs a string.

dcshzj: This PR needs to be viewed in the context of #1501. The date parsing is also needed in report-config.json, which is why it is extracted out to this file.

Also, this file contains time-based functionalities, not Time-Object functionalities or Date-Object functionalities. Hence, it returns a string.

Tejas2805: Based on the logic I feel it's okay to extract it out to a different file otherwise there will be a lot of code duplication. Though it seems like a parsing thing, it is better to not have duplicate code in the codebase if it can be avoided. Therefore, in my opinion, I think this refactoring and moving is okay. 
@0blivious your opinion?

Tejas2805 (approved):
LGTM!

src/main/java/reposense/parser/PeriodArgumentType.java
```diff
@@ -14,7 +14,7 @@
 public class PeriodArgumentType implements ArgumentType<Optional<Integer>> {
     private static final String PARSE_EXCEPTION_MESSAGE_NOT_IN_NUMERIC =
             "Invalid format. Period must be in the format of nd (n days) or nw (n weeks), "
-            + "where n is a number greater than 0.";
+                    + "where n is a number greater than 0.";
```
fzdy1914, line 17: The modification is not needed here?

fzdy1914 (approved):

fzdy1914:
```
Many of the time-based functionality and logic are scattered about 
in various files, which increases code duplication and is a 
maintenance burden.

Let's abstract all time-based functionalities into the `TimeUtil`
class such that multiple places in the code can reuse the same
logic.
```

PR #1556, opened 2021-06-29
[#435] TestUtil: fix issue with whitespace in paths

### LINKED ISSUE #435: Files that have whitespace in their paths lead to test failures

chan-j-d:
Fixes #435 

Uses Java's URL to URI conversion for decoding of URL, as is recommended by Java's API over using the URLDecoder class.

Commit message:
```
TestUtil: fix issue with whitespace in paths

The test suite will convert spaces in URLs into %20, which can lead to
test case failures. Java API recommends managing encoding and decoding
of URLs by using URI, which would resolve the issue with spaces.

This could save new developers some trouble when encountering test
failures while attempting to build the project locally for the first
time.

Let's fix the issues caused by whitespace in the test cases.
```


chan-j-d:
Requesting a review from @dcshzj

dcshzj (changes requested):
Thanks @chan-j-d for your PR! Some comments for improvement:
1. The title and description of the PR contains what we need, that's great!
2. The proposed commit message you provided follows our convention, that's great too! However, we need to keep each line to a maximum of 80 characters so that it is easier to read on smaller screens.
3. Some test cases are needed in your PR, especially those with URLs that contain spaces in their paths. @HCY123902 mentioned in #435 that there may be parsing issues if the path contains special characters, some test cases on this needs to be added as well.

Overall, good job on your first PR! Let us know if you need any assistance.

chan-j-d:
Thanks for the review! I will update the commit message accordingly. Regarding the test cases, I do need some advice on how to approach this. At least when I took up the issue, the idea was to address the way the test cases build up their path strings so as to prevent test failure from the reposense's location locally. 

The actual source of error I believe comes with the use of Java's `URL` class to get the `Path` object in `TestUtil::loadResource` method which encodes the space character along the way. However, (and please correct me on this) I don't think the actual backend code uses this method to get the `Path` object and so the issue does not occur at all in normal usage. So to clarify, is it to add test cases for the test method or to add new test cases for the existing methods involving the repo's path.

For some context, this problem caused about 80% of my test cases to fail so I could not really check what all the test cases affected were actually testing and if this change affects the accuracy of the test.

dcshzj:
> The actual source of error I believe comes with the use of Java's `URL` class to get the `Path` object in `TestUtil::loadResource` method which encodes the space character along the way. However, (and please correct me on this) I don't think the actual backend code uses this method to get the `Path` object and so the issue does not occur at all in normal usage. So to clarify, is it to add test cases for the test method or to add new test cases for the existing methods involving the repo's path.

The `TestUtil::loadResource` method is used by various files in the test suite to load files in the `test/resources` folder, and this method provides the full file path to that resource.

As for the test case, I was thinking more of creating new test cases to test possible paths that have spaces and/or special characters in them. There doesn't seem to be a good place to put the tests, so maybe you can create a new file `TestUtilTest.java` in the same folder to do loadResource tests. You can also create a new resource folder that contains some dummy files with different types of file names to test each case. Each test can be a simple one such as reading the contents of the file and ensuring that it matches an expected value.

> For some context, this problem caused about 80% of my test cases to fail so I could not really check what all the test cases affected were actually testing and if this change affects the accuracy of the test.

It might be because your computer's user name contains spaces. If this fix resolves your issue, then it sounds like a good fix. I think the other test cases shouldn't be affected by this.

chan-j-d:
Thanks for the response! I know how to proceed now.

chan-j-d:
Hi @dcshzj, requesting another review as I have included the changes you have proposed. The tests cover the potentially problematic symbols of spaces and the plus sign. An additional test case was added to test a large collection of symbols. 

Test files were standardized to be `.txt` in order to test for "."s within the file names in the case of windows. There are a few symbols which UNIX and Mac will accept but not windows so those have been left out.

src/test/java/reposense/util/TestUtilTest.java
```diff
@@ -0,0 +1,59 @@
...
+import java.io.FileReader;
+import java.nio.file.Path;
+import java.util.Scanner;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TestUtilTest {
+
+    private static final String FILE_CONTENTS = "1";
```
dcshzj, line 14: Let's rename this to `EXPECTED_FILE_CONTENTS` to make it more explicit.

src/test/java/reposense/util/TestUtilTest.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final String PATH_SYMBOLS_TEST = DIRECTORY_NAME + "/symbols +-,.;'[]{}~`!@#$%^&()_-+=test.txt";
+
+
+    @Test
+    public void loadResource_validFileWithNoSpecialCharacters_success() throws Exception {
+        Path path = loadResource(TestUtilTest.class, PATH_NO_SPECIAL_CHAR_TEST);
+        FileReader fileReader = new FileReader(path.toString());
+        String actualLine = new Scanner(fileReader).nextLine();
+        String expectedLine = FILE_CONTENTS;
+        Assert.assertEquals(expectedLine, actualLine);
```
dcshzj, line 29: After the rename above, we can directly use `EXPECTED_FILE_CONTENTS` instead of creating a new `expectedLine` variable. Same for the other test cases as well.

src/test/java/reposense/util/TestUtilTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+    @Test
+    public void loadResource_validFileWithManySymbols_success() throws Exception {
+        Path path = loadResource(TestUtilTest.class, PATH_SYMBOLS_TEST);
+        FileReader fileReader = new FileReader(path.toString());
+        String actualLine = new Scanner(fileReader).nextLine();
+        String expectedLine = FILE_CONTENTS;
+        Assert.assertEquals(expectedLine, actualLine);
+    }
+
```
dcshzj, line 58: Extra newline can be removed.

dcshzj (changes requested):
The checkstyle is not passing for the test resource files as well.

chan-j-d:
The issues have been addressed and fixed

dcshzj (approved):
LGTM!

dcshzj:
@chan-j-d Just a [note on the branch name](https://se-education.org/guides/conventions/git.html#branch-names), usually we start with the issue number that we are fixing, following by some keywords of that issue, written in kebab case. For instance, this issue can use a branch name like `435-whitespace-test-failures`, so that it is more descriptive when we look at the branches list.

Don't worry about it for this pull request, it can be done for next time.

chan-j-d:
Thanks! I'll take note.

fzdy1914 (approved):
LGTM, Good Job

PR #1561, opened 2021-07-08
ConfigParser: fix Ignore Commits List header typo in CSV files

FH-30:
Commit Message:
```
ConfigParser: fix Ignore Commits List header typo in CSV files

In RepoConfigCsvParser.java, the processLine method is unable to read
from Ignore Commit List column in config/repo-config.csv due to a
typo. Typo is also present in repoconfig_invalidLocation_test.csv
and ug/repo-config.csv.

Let's remove an extra 's' present in the Ignore Commit List header,
previously Ignore Commits List, in the three aforementioned files.
```

PR #1554, opened 2021-06-25
[#382, #1568] Make annotation tagged author appear in report

aidoxe-123:
Fixes #382.
Fixes #1568.

Created along with [reposense/testrepo-Alpha#9](https://github.com/reposense/testrepo-Alpha/pull/9)

Updated AnnotatorAnalyzer behavior:

- Only the lines with the format `(START OF LINE) COMMENT_SYMBOL @@author ONE_STRING_WITH_NO_SPACE (END OF LINE)` will be analyzed. Otherwise, the line will be ignored and treated as normal lines.
- If the line is analyzed, and the string following the author tag is a valid git id, and there is no author config file, then the code will be attributed to the author with that git id. Otherwise, the code will be attributed to the unknown author

Commit message:

```
Make annotation tagged author appear in report

The current implementation will mark tagged authors as unknown if
their names do not appear in the commit history of the repo or in the
author config file.

Let's show all author introduced via the @@author tags if there is no
author config file. If there is an author config file, @@author tags
should not result in the author appearing in the report unless the
author is listed in the author config file.
```

PR #1068, opened 2020-02-10
[#1044] Update date hashes on reset date range

ccyccyccy:
Fixes #1044
```
Resetting the date range would occasionally not change the date hashes
in the url, causing a refresh of the page to load the outdated filter
date range. The bug occurs when until or since date is not specified on
generating the report.

Let's make sure that the date hashes gets updated all the time.
```

PR #1565, opened 2021-07-13
[#569] Allow finding previous authors for ignored commits

### LINKED ISSUE #569: IgnoreCommitList: enhance to produce better results
IgnoreCommitList removes the authorship of given commits. 
https://github.com/reposense/RepoSense/blob/6dd74a04a3d1da6aa3b1636d0b0a7fdd2f873591/src/main/java/reposense/authorship/FileInfoAnalyzer.java#L94-L97

A better implementation would be to find the authors prior to commits in case of reversion.
Once such use case is as described in https://github.com/nus-cs3281/2019-dashboard/pull/28#issuecomment-466428270 where mass deletion and restoration caused a major shift in the authorship.

To test whether if its possible to achieve this, I've tried 2 solutions.
1. Using [git blame's -c options](https://git-scm.com/docs/git-blame#git-blame--Cltnumgt), in hope to be able to looks for copies from other files in any commit.
2. Using [git replace](https://git-scm.com/docs/git-replace) 

The first option has been tested to be ineffective in the above use case. 

In the second option, I used the command `git replace --graft 9a134979327a17255547ecc0b2f446174da04169 9a134979327a17255547ecc0b2f446174da04169~~` and manage to restore the author prior to the changes.
```
$ git blame "src/lib/markbind/src/handlers/cyclicReferenceError.js"             
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  1) class CyclicReferenceError extends Error {
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  2)   constructor(callStack) {
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  3)     super();
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  4)     const fileStack = callStack.slice(Math.max(callStack.length - 5, 0));
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  5)     this.message = 'Cyclic reference detected.\n'
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  6)         + `Last 5 files processed:${'\n\t'}${fileStack.join('\n\t')}`;
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  7)   }
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  8) }
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800  9) CyclicReferenceError.MAX_RECURSIVE_DEPTH = 200;
a1e9c9ed (Si Jie 2019-02-07 12:13:53 +0800 10) module.exports = CyclicReferenceError;
```

As of now, it is unsure whether this is effective in only this scenario. Do explore the possibility of using the second option to enhance IgnoreCommitList.

FH-30:
Resolves #569.

Enable RepoSense to find previous authors based on IgnoreCommitList

The current implementation of RepoSense assigns lines last modified by a commit in the IgnoreCommitList to an unknown author. This means that when a commit involving a bulk change occurs (ex: a formatting change) and that commit is added to the ignore commit list, the ```authorship.json``` file generated by RepoSense will not contain much meaningful information as most of the lines are assigned to an unknown author.

This PR fixes this issue by utilizing git's ```--ignore-revs-file``` command. RepoSense will create a ```.git-blame-ignore-revs``` file (containing the commit hashes from the ```IgnoreCommitList``` in the final config object passed to the ```analyzeRepo``` method in ```ReportGenerator.java```) in the directory of each cloned repository that has the option to find previous authors enabled. This can be enabled either by adding ```yes``` to the ```Find Previous Authors``` column in ```repo-config.csv``` or by adding ```--find-previous-authors``` / ```-F``` through the CLI. The proposed implementation in this PR has the default behavior to still assign lines due to ignored commits to an unknown author in the event that there is no previous commit that touched that line before the ignored commit. Note that the blamed lines still remain the same (can be seen in ```authorship.json``` file generated) but only the author to who it is blamed to is changed when Finding Previous Authors is enabled.

Commit Message:
```
[#569] Allow finding previous authors for ignored commits

The current implementation of RepoSense assigns lines last modified by
a commit in the IgnoreCommitList to an unknown author. This means that
when a commit involving a bulk change occurs (ex: a formatting change)
and that commit is added to the ignore commit list, the
```authorship.json``` file generated by RepoSense will not contain
much meaningful information as most of the lines are assigned to an
unknown author.

Let's add a --find-previous-authors flag or -F to tell RepoSense to
use Git blame's --ignore-revs-file argument and generate a
.git-blame-ignore-revs file in the repositories analyzed to allow
RepoSense users to have a more meaningful authorship report from
RepoSense's analysis in the event that a bulk change occurs in the
repositories analyzed.
```


damithc:
@reposense/21-22-active-team any feedback for @FH-30 ?

dcshzj:
I think the idea is great! However, I have some questions about this feature:
1. Any reason for having both a CLI argument and a repo-config.csv column for specifying whether to find previous authors? What happens when they conflict?
2. When this setting is enabled and a particular ignored commit adds a new file, will subsequent commits to that new file have the entire file attributed to that author, or will it retain the previous behavior of attributing to an unknown author?
3. This Git blame ignore revs feature in Git was introduced in 2019 which is rather recent (version 2.23). Luckily GitHub Actions can support it as it is running >=2.31, but some users may still be running an older version of Git on their computers. Will there be a fail-safe mechanism implemented in the event that the user is running RepoSense locally that uses an older version of Git?

FH-30:
Thanks for the feedback @dcshzj , these are my answers to your questions:

**1.**
The reason why I have decided to add both a CLI argument and a repo-config.csv column for specifying whether to find previous authors is to allow for flexibility.  There might be scenarios where we want to just ignore commits (without finding previous authors) for some repositories but we want to find the previous authors of certain commits for some repositories. Placing it in repo-config.csv allows for users to utilize this feature in RepoSense for such scenarios . The CLI argument on the other hand makes it easy if the user wants to set the finding previous authors mode for all the repositories without having to add the yes flag to all rows in the FInd Previous Authors column in repo-config.csv.

If the CLI argument happens to conflict with the repo-config.csv argument, there are four possible scenarios as shown in the table below:
|repo-config.csv for Find Previous Authors Column | CLI  for -F flag | Find Previous Authors|
|--------------------------------------------------|----------------|-----------------------|
|False|False|False|
|True|False|True|
|False|True|True|
|True|True|True|

This is similar to how current conflicting arguments in RepoSense are handled.

**2.**
I'm not sure if I got your question right, do correct me if I got it wrong (I'm not sure which author is "that author" referring to). Assuming "that author" is the author that modified the file, when this setting is enabled and a particular ignored commit adds a new file then subsequent commits to that file will not necessarily attribute that whole file to the author. Different lines of the file will remain attributed to the individual authors that modified those lines. An example of how this feature works in RepoSense:

Commits in chronological order:
Commit ```0dfba``` by author Alice adds a new txt file to a repository called Test_Repo.
Commit ```5dedf``` by author Bob modifies the txt file by adding something to line 1.
Commit ```abde5``` by author Dave modifies the txt file by modifying line 1.
Commit ```1ab41``` by author Eve modifies the txt file by adding something to line 2.

Some different possibilities:
| Content of Ignore Commit List | Line 1 blamed to | Line 2 blamed to |
|--------------------------------------------------|----------------|-----------------------|
||Dave|Eve|
|0dfba|Dave|Eve|
|5dedf;1ab41|Dave||
|abde5|Bob|Eve|
|5dedf;abde5||Eve|
|0dfba;5dedf;abde5||Eve|
|0dfba;5dedf;abde5;1ab41|||

The creating of a file is never taken note of in ```authorship.json``` by RepoSense (this is not due to the PR change but it is in fact RepoSense's default behaviour) therefore Alice is never blamed for anything. The only evidence Alice created the file is probably in ```commits.json```.

**3.**

I was also contemplating about this, is it preferable to throw an error when Find Previous Authors is enabled but the current git version of the user is not >= 2.23 or is it better to try to resort to using the ```git hyper-blame``` tool suggested in the [issue discussion](https://github.com/reposense/RepoSense/issues/569) even though it seems to have been discontinued (no longer maintained) and is probably not as accurate as git's own implementation.

For this last question, maybe prof @damithc can give his opinion too, thanks!

damithc:
@FH-30 Good to see you are doing a deep investigation on this issue. Keep it up!

> I was also contemplating about this, is it preferable to throw an error (asking the user to upgrade his version of git if he wants to use this feature) when Find Previous Authors is enabled but the current git version of the user is not >= 2.23 or is it better to try to resort to using the `git hyper-blame` tool suggested in the [issue discussion](https://github.com/reposense/RepoSense/issues/569) even though it seems to have been discontinued (no longer maintained) and is probably not as accurate as git's own implementation.

> For this last question, maybe prof @damithc can give his opinion too, thanks!

I think no need to use an additional tool if the user's git version is not matching, in the interest of keeping the code simple. 

In general, it is OK to expect the user to have a recent version of Git, although we may not want to go for a bleeding edge version as the CI platforms we piggy back on e.g., GitHub Actions, Travis, etc. might not have support for very new Git versions.

It is ideal (but not compulsory) if the rest of the features work with older versions of Git if only the feature in concern requires a new version of Git.

FH-30:
Hi @dcshzj, may I have your feedback based on my previous response? Thanks in advance!

FH-30:
I have decided to throw an error instead, can inform me if it is advisable not to do so

dcshzj:
@FH-30 Apologies for the delay, here are my replies to your comment.

> The reason why I have decided to add both a CLI argument and a repo-config.csv column for specifying whether to find previous authors is to allow for flexibility.

Sounds good, let's go ahead with your implementation of having the configuration on both repo-config.csv and as a CLI argument. Do make sure that the documentation is clear about how the CLI argument has a higher priority (i.e. even if a user explicitly disables finding previous authors for a particular repository, it will still be done when the CLI argument is provided).

> Different lines of the file will remain attributed to the individual authors that modified those lines.

Yep, this is the ideal behaviour. I was thinking of the following scenario:
- Commit A by person A creates a new file of 500 lines.
- Commit B by person B modifies the same file by changing only the last line

If commit A is added to the ignore commit list and Git cannot find any previous authors to this file, I am concerned whether Git will assign the rest of the 499 lines to person B. RepoSense's default behaviour is to assign to an unknown author, but I am not sure if Git blame ignore revs will do the same as RepoSense, or does it perform differently.

> I was also contemplating about this, is it preferable to throw an error (asking the user to upgrade his version of git if he wants to use this feature) when Find Previous Authors is enabled but the current git version of the user is not >= 2.23 or is it better to try to resort to using the `git hyper-blame` tool suggested in the [issue discussion](https://github.com/reposense/RepoSense/issues/569) even though it seems to have been discontinued (no longer maintained) and is probably not as accurate as git's own implementation.

My personal preference is to notify the user that this feature is not supported after checking the current Git version that the user has, and perform the default behaviour of not finding previous authors. I think let's not introduce hyper-blame into our project.

Finally, do mark this PR as ready to review, when you are ready to have it reviewed.

FH-30:
Thanks for the review @dcshzj , no worries regarding the delay! I will proceed with implementing the tests and documentation shortly!

FH-30:
Frontend tests are failing because #1068 has yet to be merged in.

src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -18,24 +20,34 @@
...
  * Contains templates for git testing.
  */
+
 public class GitTestTemplate {
-    protected static final String TEST_REPO_GIT_LOCATION = "https://github.com/reposense/testrepo-Alpha.git";
+    protected static final String TEST_REPO_GIT_LOCATION = "https://github.com/FH-30/testrepo-Alpha.git";
     protected static final String DISK_REPO_DISPLAY_NAME = "testrepo-Alpha_master";
+    // repos/reposense_testrepo-Alpha/testrepo-Alpha/git-blame-ignore-revs
+    protected static final String IGNORE_REVS_FILE_LOCATION =
+            "repos/FH-30_testrepo-Alpha/testrepo-Alpha/.git-blame-ignore-revs";
```
FH-30, line 41: These two variables need to be changed when merging in. It is important that the commit hashes in the branch don't change when somehow adding the branch to testrepo-Alpha (might have to make quite a lot of changes if that's the case)

FH-30:
Relevant comments to take note of:
https://github.com/reposense/RepoSense/pull/1565#issuecomment-895334200
https://github.com/reposense/RepoSense/pull/1565#discussion_r685350513

FH-30:
The tests are failing after I merged in the master branch, how do you advice me to proceed from here onwards?

gerhean (dismissed):
Please gitignore the .DS_Store files.

Furthermore I was wondering if it is possible to make repo-config.csv columns be optional so that it can be backwards compatible with existing repo-config (also to reduce clutter).

Other than that I am not very familiar with the backend code so I can't really review this properly.

FH-30:
If possible, I hope that this PR can be wrapped by the end of this week. Thanks!

FH-30:
@dcshzj I have resolved the merge conflicts arising from #1561, however the system tests here are failing. Is changing the files in resources here necessary?

dcshzj:
@FH-30 It seems like an error arising from testrepo-Charlie, which was not touched recently. Can you double check if it is due to your code, as it should keep the original behaviour even with the additional column?

FH-30:
ok that's odd, because the system tests were passing around 3-4 weeks ago and it started failing about 1 week ago when other pull requests were merged in, I'll try to look into it.

FH-30:
Based on the other pull requests merged in, do you have any possible guess on which ones are likely to affect files in this PR?

dcshzj:
I would make a wild guess that it is #1561? It is the latest one that touched the test repositories. The other one would be #1554 that you can look at.

FH-30:
This is what I got from diffing both commits.json (left is expected and right is output). I have not touched any code relating to changes in commits.json (my code only affects authorship.json). Apparently the author "charlesgoh-reused" disappears.
Screenshot 2021-08-30 at 1 33 51 AM

I'll try to find out why, however your insights on this is also greatly appreciated

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -343,8 +347,12 @@ private static AnalyzeJobOutput analyzeRepos(String outputPath, List<RepoConfigu
         updateAuthorList(config);
         updateIgnoreCommitList(config);
 
-        AuthorshipSummary authorshipSummary = AuthorshipReporter.generateAuthorshipSummary(config);
+        if (config.isFindingPreviousAuthorsPerformed()) {
+            generateIgnoreRevsFile(config);
+        }
+
         CommitContributionSummary commitSummary = CommitsReporter.generateCommitSummary(config);
+        AuthorshipSummary authorshipSummary = AuthorshipReporter.generateAuthorshipSummary(config);
```
FH-30, line 355: I have found the bug, it appears that I have accidentally switched the positions of generating the commitContributionSummary and the authorship summary when merging PRs in.

FH-30:
@dcshzj All the checks have passed, you may proceed to review this PR! Thanks in advance!

FH-30:
Hi @dcshzj , sorry for bothering (I know you must be busy too). I just hope that you can give a review over the weekend as the deadline for this PR is next Friday, thanks in advance!

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -478,10 +486,35 @@ private static void handleFailedConfigs(
...
+     * in the config's repo root directory.
+     */
+    private static void generateIgnoreRevsFile(RepoConfiguration config) {
+        List<CommitHash> expandedIgnoreCommitList = config.getIgnoreCommitList().stream()
+                .map(CommitHash::toString)
+                .map(commitHash -> {
+                    try {
+                        return GitShow.getExpandedCommitHash(config.getRepoRoot(), commitHash);
+                    } catch (CommitNotFoundException e) {
+                        logger.warning(String.format("Cannot expand %s, it shall remain unexpanded", commitHash));
```
dcshzj, line 500: This string can become its own constant, similar to line 89 above.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -110,6 +111,26 @@ public static void zipFoldersAndFiles(List<Path> pathsToZip,
...
     }
 
+    /**
+     * Writes the ignore revs file containing the {@code ignoreCommitList} at the given {@code path}.
+     * @return An Optional containing the Path to the ignore revs file, or an empty Optional
+     *         if there was an error while writing the ignore revs file.
+     */
+    public static Optional<Path> writeIgnoreRevsFile(String path, List<CommitHash> ignoreCommitList) {
+        String contentOfIgnoreRevsFile = ignoreCommitList
+                                            .stream()
```
dcshzj, line 121: This line can be put together with line 120.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -110,6 +111,26 @@ public static void zipFoldersAndFiles(List<Path> pathsToZip,
...
+                                            .map(CommitHash::toString)
+                                            .reduce("", (hashes, newHash) -> hashes + newHash + "\n");
```
dcshzj, line 123: The indentation looks weird, I think it should be 8 spaces after the previous line?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -110,6 +111,26 @@ public static void zipFoldersAndFiles(List<Path> pathsToZip,
...
+
+        try (PrintWriter out = new PrintWriter(path)) {
```
dcshzj, line 125: I don't think use this type of try-catch statement in this project. Can the expression be placed within the try block?

src/main/java/reposense/util/FileUtil.java
```diff
@@ -299,4 +320,5 @@ public static boolean isEmptyFile(String directoryPath, String relativeFilePath)
     private static boolean isFileTypeInPath(Path path, String... fileTypes) {
         return Arrays.stream(fileTypes).anyMatch(path.toString()::endsWith);
     }
+
```
dcshzj, line 323: Extra line?

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -116,7 +137,8 @@ private String getInputWithDates(String sinceDate, String untilDate) {
     /**
      * Generates the testing report to be compared with expected report.
      */
-    private void generateReport(String inputDates, boolean shouldIncludeModifiedDateInLines, boolean shallowCloning)
+    private void generateReport(String inputDates, boolean shouldIncludeModifiedDateInLines, boolean shallowCloning,
+                                boolean findPreviousAuthors)
```
dcshzj, line 141: Line 142 can be merged together with this line.

src/test/java/reposense/authorship/FileAnalyzerTest.java
```diff
@@ -119,6 +158,38 @@ public void analyzeFile_blameTestFileIgnoreFakeAuthorCommitFullHash_success() {
...
+
+        FileInfo fileInfoFull = generateTestFileInfo("blameTest.java");
+        config.setIgnoreCommitList(Collections.singletonList(MAIN_AUTHOR_BLAME_TEST_FILE_COMMIT_06022018));
+        createTestIgnoreRevsFile(config.getIgnoreCommitList());
+        FileInfoAnalyzer.analyzeTextFile(config, fileInfoFull);
+        removeTestIgnoreRevsFile();
+
+        FileInfo fileInfoShort = generateTestFileInfo("blameTest.java");
+        config.setIgnoreCommitList(
+                Collections.singletonList(
```
dcshzj, line 177: These two lines can be merged together.

src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -47,6 +47,12 @@
     private static final Path SHALLOW_CLONING_FLAG_OVERRIDE_TEST_CONFIG_FILES =
             loadResource(RepoConfigurationTest.class,
             "RepoConfigurationTest/repoconfig_shallowCloningOverrideCsv_test");
+    private static final Path FIND_PREVIOUS_AUTHORS_TEST_CONFIG_FILES =
+            loadResource(RepoConfigurationTest.class,
+                    "RepoConfigurationTest/repoconfig_findPreviousAuthors_test");
```
dcshzj, line 52: The extra spaces are not necessary.

src/test/java/reposense/model/RepoConfigurationTest.java
```diff
@@ -47,6 +47,12 @@
...
+    private static final Path FIND_PREVIOUS_AUTHORS_FLAG_OVERRIDE_TEST_CONFIG_FILES =
+            loadResource(RepoConfigurationTest.class,
+                    "RepoConfigurationTest/repoconfig_findPreviousAuthorsOverrideCsv_test");
```
dcshzj, line 55: The extra spaces are not necessary.

src/test/java/reposense/parser/RepoConfigParserTest.java
```diff
@@ -35,6 +35,9 @@
             "RepoConfigParserTest/require_trailing_whitespaces/repoconfig_redundantLines_test.csv");
     private static final Path REPO_CONFIG_INVALID_LOCATION_FILE = loadResource(RepoConfigParserTest.class,
             "RepoConfigParserTest/repoconfig_invalidLocation_test.csv");
+    private static final Path REPO_CONFIG_UNRECOGNIZED_VALUES_FOR_YES_KEYWORD_HEADERS_FILE =
+            loadResource(RepoConfigParserTest.class,
+                    "RepoConfigParserTest/repoconfig_unrecognizedValuesForYesKeywordHeaders_test.csv");
```
dcshzj, line 40: The extra spaces are not necessary.

src/test/java/reposense/template/GitTestTemplate.java
dcshzj: Do you have a PR for this branch?

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -225,6 +238,10 @@ public static void setStandaloneConfigIgnoredToRepoConfigs(
         }
     }
 
+    public static boolean isAnyRepoFindingPreviousAuthors(List<RepoConfiguration> configs) {
+        return configs.parallelStream().anyMatch(RepoConfiguration::isFindingPreviousAuthorsPerformed);
```
dcshzj, line 242: Let's use `stream()` instead of `parallelStream()` to standardize with the rest of the codebase.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -84,6 +87,7 @@
     private static final String MESSAGE_BRANCH_DOES_NOT_EXIST = "Branch %s does not exist in %s! Analysis terminated.";
 
     private static final String LOG_ERROR_CLONING = "Failed to clone from %s";
+    private static final String LOG_ERROR_FILE_DELETION = "Failed to delete file at %s";
```
dcshzj, line 90: This constant does not seem to be used.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -478,10 +486,35 @@ private static void handleFailedConfigs(
         return generatedFiles;
     }
 
+    /**
+     * Creates the .git-blame-ignore-revs file containing the contents of {@code IgnoreCommitList}
+     * in the config's repo root directory.
+     */
+    private static void generateIgnoreRevsFile(RepoConfiguration config) {
```
dcshzj, line 493: Two questions:
1. Will this function override the existing `.git-blame-ignore-revs` in the repository if it already exists?
2. Will this generated file be deleted after the report generation is completed?

dcshzj (changes requested):

src/main/java/reposense/util/FileUtil.java
FH-30: It's actually used right above it in another function
Screenshot 2021-09-05 at 11 47 22 PM

src/main/java/reposense/git/GitVersion.java
```diff
@@ -0,0 +1,38 @@
...
+ * Contains git version related functionalities.
+ * Git version is responsible for finding out the version of git the user of RepoSense is running.
+ */
+public class GitVersion {
+    public static final String FINDING_PREVIOUS_AUTHORS_INVALID_VERSION_WARNING_MESSAGE =
+            "--find-previous-authors/-F requires git version 2.23 and above. Feature will be disabled for this run";
+
+    /** Regex for matching Git version 2.23 and above */
+    public static final Pattern FINDING_PREVIOUS_AUTHORS_VALID_GIT_VERSION_PATTERN =
+            Pattern.compile("(((2\\d*\\.(2[3-9]\\d*|[3-9]\\d+|\\d{3,}))|((([3-9]\\d*)|(1\\d+))\\.\\d+))\\.\\d*)");
```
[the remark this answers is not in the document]
FH-30: Nope, didn't take this from any source. Thanks for the compliment!

src/main/java/reposense/util/FileUtil.java
dcshzj: Ok! Then let's keep it this way.

src/test/java/reposense/model/RepoConfigurationTest.java
FH-30: Screenshot 2021-09-06 at 12 14 52 AM
For spaces are you referring to this?

src/test/java/reposense/parser/RepoConfigParserTest.java
FH-30: I have yet to make a PR for this branch, I'm not sure which branch to direct the PR to since the branch is non existent on the remote repository. Aside from that I need the commit hashes in my the branch above to be preserved (as it is used in the tests), is there any way you can ensure that?

src/main/java/reposense/model/RepoConfiguration.java
FH-30: I have yet to make a PR for this branch, I'm not sure which branch to direct the PR to since the branch is non existent on the remote repository. Aside from that I need the commit hashes in my the branch above to be preserved (as it is used in the tests), is there any way you can ensure that?

src/main/java/reposense/report/ReportGenerator.java
FH-30: 1. Yes, it will override the existing file if it already exists.
2. Yes, the whole repo will be deleted therefore the file also gets deleted.

src/test/java/reposense/model/RepoConfigurationTest.java
dcshzj: No sorry, I meant that the lines 54 and 55 should be aligned, so total 12 spaces in front.

FH-30: ah ok, thanks for the clarification

src/test/java/reposense/template/GitTestTemplate.java
dcshzj: @FH-30 I think you meant to reply to this thread.

> the branch is non existent on the remote repository

I just pushed a new branch `1565-find-previous-authors` (changed the name to be more informative). You can update line 42 to use this name, and submit a PR to that branch.

> I need the commit hashes in my the branch above to be preserved

I will use rebase merging on the command line to preserve the commit hash, so it should not be an issue.

FH-30:
Just adding this here, just in case you didn't notice:

I have yet to make a PR for this branch, I'm not sure which branch to direct the PR to since the branch is non existent on the remote repository. Aside from that I need the commit hashes in my the branch above to be preserved (as it is used in the tests), is there any way you can ensure that?

src/test/java/reposense/template/GitTestTemplate.java
FH-30: I have submitted a PR here:
https://github.com/reposense/testrepo-Alpha/pull/13

FH-30: I have created the commit to change the branch(line 42) and path (line 41) to be consistent with the newly merged PR in testrepo-alpha. However, I will refrain from pushing until it is merged to prevent triggering failing tests on this PR.

FH-30: > I will use rebase merging on the command line to preserve the commit hash, so it should not be an issue.

That's good to hear!

dcshzj: @FH-30 For the PR, can you rebase your branch? There were some updates to master, which will cause your commit to not become the latest one in the final branch.

FH-30: @dcshzj I have just rebased it and have done a force push, however it appears that in the process of rebasing the git commit hash gets altered, I'll make the necessary changes in the PR, in the mean time I hope you can merge the branch as soon as possible, thanks!

dcshzj: @FH-30 Done! The PR has been merged.

FH-30: @dcshzj the tests are now passing locally, however they aren't passing on the github CI for ubuntu 18.04. I do not have an ubuntu machine so I'm undure about how to fix this, what do you propose?

FH-30: Upon merging with the master branch of RepoSense, it passes. Not entirely sure why.

dcshzj: Looks like all good now, probably some hiccup somewhere.

FH-30: Yeah, thanks for the quick review over the past two days. I guess now you can pass it on to the next stage reviewer!

dcshzj (approved):
LGTM!

FH-30:
Hi, @fzdy1914. I'm guessing that you must be the stage 2 reviewer for this PR. I do hope you can review this PR soon as the deadline is this Friday, thanks!

damithc:
> I do hope you can review this PR soon as the deadline is this Friday, thanks!

It's a soft deadline @FH-30 Can take a few more days if needed.

FH-30:
Ah that's good to hear, thanks prof!

fzdy1914:
I am sorry that this PR is too long for me to review in time, @dcshzj Can just merge before Friday.

dcshzj:
@FH-30 Looks like the system tests are failing after #1586 was merged. Can I trouble you to look into it?

FH-30:
Since I added new system tests for this PR and I noticed that the PR you just merged into the master branch had all the expected authorship.json and commits.json changed for the existing system tests, I believe it is the case that I need to change the expected authorship.json and commits.json for the new system tests in this PR. @dcshzj is it alright if I do so?

dcshzj:
@FH-30 Yup please do, thanks!

FH-30:
@dcshzj All the tests are now passing!

PR #1566, opened 2021-07-15
[#937] Make all tests throw generic Exception

### LINKED ISSUE #937: Test: use generic throw Exception in all tests

FH-30:
Resolves #937.

Commit Message:
```
Make all tests throw generic Exception

The current test code for RepoSense contains throws clauses that do
not throw a generic Exception. This can result in a series of code
changes that have to be made when the exception(s) thrown by an
existing method used in the test code is changed, to prevent a
compiler error.

Let's make all methods in the test (and systemtest) folder(s) of
RepoSense throw a generic Exception.
```


HCY123902:
LGTM. Just 1 minor suggestion. Although there are not many lines affected, you can actually split this into 2 separate pull requests because it seems that this pull request is handling 2 separate issues. The reason why pull request needs to be atmoic is that this can help to keep a clear record of what work is done for each individual issue and enable future contributors to know the previous progress

FH-30:
Thanks @HCY123902 for the suggestion, I have undone the removal of redundant throws clauses in this PR and have made another PR for it [here](https://github.com/reposense/RepoSense/pull/1567).

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -544,7 +544,7 @@ public void parse_repoAliases_sameResult() throws Exception {
 
     @Test (expected = ParseException.class)
     public void parse_noValidRepoLocation_throwsParseException()
-            throws ParseException, HelpScreenException {
+            throws Exception {
```
dcshzj, line 547: I think this can fit into one line.

dcshzj (changes requested):

src/test/java/reposense/parser/ArgsParserTest.java
FH-30: Right, thanks for pointing it out!

FH-30:
Hi @HCY123902 and @dcshzj, any updates in this PR?

HCY123902 (approved):
LGTM

HCY123902:
@dcshzj

dcshzj (approved):

fzdy1914 (approved):

PR #1567, opened 2021-07-15
Test: remove redundant throws Exception clauses

FH-30:
Commit Message:
```
The current test code for RepoSense contains two methods that include
a redundant throws clause.

Let's remove both redundant throws clauses in GitCloneTest.java and
TimeUtilTest.java.
```

### DIFF
--- a/src/test/java/reposense/util/TimeUtilTest.java
+++ b/src/test/java/reposense/util/TimeUtilTest.java
@@ -9,7 +9,7 @@ import org.junit.Test;
 
 public class TimeUtilTest {
     @Test
-    public void extractDate_validDate_success() throws Exception {
+    public void extractDate_validDate_success() {
         String expectedDate = "20/05/2019";
         String actualDate = TimeUtil.extractDate(expectedDate);
         Assert.assertEquals(expectedDate, actualDate);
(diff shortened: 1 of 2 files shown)


src/test/java/reposense/util/TimeUtilTest.java
```diff
@@ -9,7 +9,7 @@
 
 public class TimeUtilTest {
     @Test
-    public void extractDate_validDate_success() throws Exception {
+    public void extractDate_validDate_success() {
```
dcshzj, line 12: I think you missed out on `extractDate_validDateAndTime_success()`.

dcshzj (changes requested):

src/test/java/reposense/util/TimeUtilTest.java
FH-30: For some reason IntelliJ does not indicate that is is redundant as shown in the image below:
Screen Shot 2021-07-17 at 2 21 20 PM

However, it can be removed without incurring a compilation error either. It makes sense as both test cases use the same set of methods internally. Thanks for pointing this out!

FH-30:
Hi @dcshzj , any updates on this PR?

Tejas2805 (approved):
LGTM!

PR #1570, opened 2021-07-20
[#1563, #1579] Tests: allow a fresh clone for some test cases

### LINKED ISSUE #1563: Allow forcing of a fresh repository close for some test cases
As identified in #1562, #1478 introduced a regression that caused certain test cases that test the cloning mechanism to skip cloning entirely. One such example is `GitClone.java` and the two system tests, `testSinceBeginningDateRangeWithShallowCloning` and `test30DaysFromUntilDateWithShallowCloning`, which handles shadow cloning.

There should be a mechanism for individual tests to specify if they wish to obtain a fresh clone of a repository, and to also ensure that it does not affect subsequent tests on the same repository.

### LINKED ISSUE #1579: ConfigSystemTest run shallow-cloning test as normal cloning
**Steps to reproduce:**

1. Run ConfigSystemTest.testSinceBeginningDateRangeWithShallowCloning()

**Expected:**
The repoConfigs created in ConfigSystemTest.generateReport will have isShallowCloningPerformed = true

**Actually:**
The repoConfigs created in ConfigSystemTest.generateReport will have isShallowCloningPerformed = false

image

aidoxe-123:
Fixes #1563.
Fixes #1579.

Commit Message:

```
Tests: allow a fresh clone for some test cases

The test repositories are cloned only once during testing, and a full
clone is performed. This causes certain test cases that require a
shallow clone to be tested incorrectly on a full clone.

Let's allow individual tests to specify if they wish to obtain a
fresh clone of a repository.
```


aidoxe-123:
Request review from @dcshzj

dcshzj (changes requested):
It looks like you are specifying whether or not to request for a fresh clone when it comes to shallow cloning. However, it may be possible that a test requires a normal clone, but would not have requested for a fresh clone. Will this change be able to handle that?

Also, the proposed commit message should have lines no longer than 70 characters.

aidoxe-123:
> It looks like you are specifying whether or not to request for a fresh clone when it comes to shallow cloning. However, it may be possible that a test requires a normal clone, but would not have requested for a fresh clone. Will this change be able to handle that?

If nothing is specified, the program will clone the repo if the repo hasn't been cloned before, and use the existing files if they are cloned already. Is this what you mean?

dcshzj:
I was thinking of a possible scenario (all tests are using testrepo-Alpha):
1. Test case 1 requests for a fresh clone, as the repository has not been cloned before.
2. Test case 2 is programmed to request for a fresh clone, as it requires the repository to be shallow cloned.
3. Test case 3 requires a normal cloned repository, but is not programmed to request for a fresh clone.

In this case, test case 3 will be working with a shallow cloned repository, instead of a normal cloned repository. We would want test case 3 to request for a fresh clone.

However, the following scenario may occur:
1. Test case 1 requests for a fresh clone, as the repository has not been cloned before.
2. Test case 3 requires a normal cloned repository and uses the same repository that was cloned earlier.
3. Test case 2 is programmed to request for a fresh clone, as it requires the repository to be shallow cloned. A fresh clone is performed.

In this case, we want test case 3 to not request for a fresh clone, as the repository is a normal cloned repository and it has been cloned before.

We need to account for both cases as it is possible that the tests run in random order, which changes the context in which each test case is run.

aidoxe-123:
Instead of implementing a new feature, I will make a test case that shallow-clones a repo to re-clone that repo normally after finishing checking. Is it ok to do it this way?

dcshzj:
> Instead of implementing a new feature, I will make a test case that shallow-clones a repo to re-clone that repo normally after finishing checking. Is it ok to do it this way?

I think it might also be possible where we get two tests that require the same shallow-cloned repository consecutively, which results in the repository being shallow-cloned twice. Is it possible to work out a solution that can handle this case as well?

aidoxe-123:
>  Is it possible to work out a solution that can handle this case as well?

I am thinking of 2 possible solutions to handle this: 

1. Use a static variable to keep track of whether the repo is shallow cloned. I'm not sure if this is a good idea since it might cause problems by tests sharing the same variable
2. Create an extra file in the cloned repos if they are shallow cloned. I have implemented this locally but it seems to cause trouble when those repos are analyzed

However, currently, the only 5 tests that need a fresh clone are the one in ConfigSystemTest. The test for GitClone.java only checks if the clone has been executed successfully by GitTestTemplate, and the GitTestTemplate always clone the repo without checking SystemUtil.isTestEnvironment(). So is it really necessary to be overcomplicated and implement this instead of making all the test cases in ConfigSystemTest require a fresh clone?

@dcshzj Can I have your take on this?

aidoxe-123:
> Is it possible to work out a solution that can handle this case as well?

Done, @dcshzj can you help me review it?

dcshzj:
I'll let someone else review as I created the original issue, so we can see others' opinions on this as well.

dcshzj (dismissed):

damithc:
@reposense/21-22-active-team can we try to wrap this up over this weekend?

HCY123902 (approved):

dcshzj:
@Tejas2805 or @fzdy1914 Any comments?

fzdy1914 (approved):
For all variable name, `isFreshClone` or `shouldFreshClone` is enough though. The original name is too long. Functional-wise looks good to me. @dcshzj Your though?

dcshzj (approved):
LGTM! Ideally, we should not mix code related to tests inside the functional code, but it will require a major overhaul which I think should be done in a separate PR in the future.

FH-30:
Hi, I believe that since [#1586](https://github.com/reposense/RepoSense/pull/1586) was merged into the master branch and that this PR also changes the expected files in resources for system tests, you might need to replace them again @aidoxe-123

aidoxe-123:
@FH-30 I got this error message dumped into the summary.json file when running `testSinceBeginningDateRangeWithFindPreviousAuthors` on Github CI. Do you have any ideas why?
image

All the test cases pass locally, all test cases pass on Github CI except for that one. You can look at the detail of the error in the Github CI's log.

FH-30:
@aidoxe-123 I got a similar issue the other day but it disappeared when the tests rerun, maybe you can try rerunning the tests by pushing a dummy commit. @dcshzj should an issue be made about this? I do not possess an ubuntu device so I guess it might be good if someone with one can investigate into this.

aidoxe-123:
@FH-30 Thanks, the test passes after I rerun it

dcshzj:
> @aidoxe-123 I got a similar issue the other day but it disappeared when the tests rerun, maybe you can try rerunning the tests by pushing a dummy commit. @dcshzj should an issue be made about this? I do not possess an ubuntu device so I guess it might be good if someone with one can investigate into this. Particularly it always bugs with ubuntu 18.04 resulting in the other ubuntu tests to be cancelled. In my case it happened with testrepo-charlie, now it happened with testrepo-delta.

@FH-30 I think the issue is with our use of concurrent cloning of repositories, as it makes the behavior a little unpredictable. I agree that an issue should be created for this, so we can investigate and make the behavior more deterministic.

FH-30:
I have made an issue for this [here](https://github.com/reposense/RepoSense/issues/1588)

PR #1574, opened 2021-07-27
[#1575] Fix misidentification of deleted files as edited files

### LINKED ISSUE #1575: Analysis of edited files misidentifies deleted files

1. Add the following 2 println statements to the `FileInfoExtractor` beginning at line 116. (Or any means of viewing the value of `filePath` at these lines).
image

1. Run RepoSense on intellij with the following arguments: `--repos https://github.com/reposense/RepoSense.git --since 31/12/2018`

Output contains lines with `DELETED-FILE: /dev/null`

Output contains lines with `INVALID-TEXT-FILE: dev/null`
image

chan-j-d:
Fixes #1575 

A normal file path looks like `+++ b/file/path/` while a deleted file shows up as `+++ /dev/null/`

The current regex will instead parse out the string `dev/null` without the initial slash for deleted files and attempt to match it to `/dev/null/`.

This change has no direct impact on the output of the program at the moment as the relevant lines of code (linked below) has an additional `if` statement where the `dev/null` filepaths are discarded unintentionally. However, if there is a change in the program flow then it might become an issue in the future.
https://github.com/reposense/RepoSense/blob/6b94f82545bbd84852f7bd110af72e495ad23652/src/main/java/reposense/authorship/FileInfoExtractor.java#L114-L121

Commit message:
```
Current regex expression matches dev/null/ while the variable is
/dev/null/. This causes files to be wrongly treated as being edited
instead of being deleted.

Let's fix the misidentification of deleted files.
```

### DIFF
--- a/src/main/java/reposense/authorship/FileInfoExtractor.java
+++ b/src/main/java/reposense/authorship/FileInfoExtractor.java
@@ -38,7 +38,7 @@ public class FileInfoExtractor {
     private static final String LINE_INSERTED_SYMBOL = "+";
     private static final String STARTING_LINE_NUMBER_GROUP_NAME = "startingLineNumber";
     private static final String FILE_CHANGED_GROUP_NAME = "filePath";
-    private static final String FILE_DELETED_SYMBOL = "/dev/null";
+    private static final String FILE_DELETED_SYMBOL = "dev/null";
     private static final String MATCH_GROUP_FAIL_MESSAGE_FORMAT = "Failed to match the %s group for:\n%s";
     private static final String BINARY_FILE_LINE_DIFF_RESULT = "-\t-\t";
 


chan-j-d:
Ready for review.

HCY123902:
Hello, thanks for the contribution. Can you link it to a corresponding issue? In particular, can you reproduce the problems caused by the misidentification of deleted files in the issue

chan-j-d:
Hi @HCY123902, I have created issue #1575 to demonstrate the faulty logic with the current `if` statements.

HCY123902 (approved):

dcshzj (approved):
LGTM! I was initially confused by your PR description, so I wanted to add on to it (feel free to correct me if I interpret it incorrectly).

The file path regex matcher that we currently use is given as follows;
https://github.com/reposense/RepoSense/blob/6b94f82545bbd84852f7bd110af72e495ad23652/src/main/java/reposense/authorship/FileInfoExtractor.java#L49

However, the part of the regex `b?/(?<filePath>.*)\n` causes the initial forward slash to be stripped when the pattern matching is performed here:
https://github.com/reposense/RepoSense/blob/6b94f82545bbd84852f7bd110af72e495ad23652/src/main/java/reposense/authorship/FileInfoExtractor.java#L105

This meant that by the following line, the `filePath` would be `dev/null`, as the original match line will say `+++ /dev/null`.
https://github.com/reposense/RepoSense/blob/6b94f82545bbd84852f7bd110af72e495ad23652/src/main/java/reposense/authorship/FileInfoExtractor.java#L112

Hence, there is a need to change the `FILE_DELETED_SYMBOL` to remove the leading forward slash, so that it can correctly match deleted files.

fzdy1914 (approved):

PR #1576, opened 2021-07-29
[#1571] Fix incorrect invalidation of file path with spaces

### LINKED ISSUE #1571: Analysis incorrectly identifies valid file path as invalid

* **RepoSense Version:** master

1. Run program on intellij with the following CLI arguments `--repos https://github.com/reposense/RepoSense.git --since 31/12/2018`

Analysis completed without issue and report is generated.

Analysis identifies two files as invalid file paths ignores them and proceeds to generate the report.
image

chan-j-d:
Fixes #1571 

As explained and demonstrated in #1571, for `git diff` commands where the filename contains a space, the resulting `git diff` output generates an additional `\t` character at the end of the filename. This causes Windows to throw an exception (and results in a warning in the log) and skips it while Linux and Mac skips over the file without a warning. 

This pull request adjusts the regex to account for this additional tab character.

Commit message:
```
Fix incorrect invalidation of file path with spaces

The "git diff" command on files where the filename contains a space
causes the resulting output to generate an additional \t character at
the end of the filename. This becomes an invalid file path in Windows
while being incorrectly ignored in Macs and Linux.

Let's fix the incorrect identification of valid file paths with
spaces.
```


chan-j-d:
Below is the pull request that will add the relevant file required for this test case: 
https://github.com/reposense/testrepo-Alpha/pull/8

dcshzj:
@chan-j-d You can follow https://github.com/reposense/RepoSense/pull/1554#issuecomment-888405130 to demonstrate the test, and we will change it back just before merging.

chan-j-d:
@dcshzj I have changed the test-repo location to the one on my account and it passes all the tests. 

PR is ready to review.

The issue with the codecov seems to be a block of code where I am unsure if it is feasible to test for since the exception is likely only thrown when the git output is something unexpected that does not match well with the regex pattern. It is added to hopefully spot future issues similar to the one in #1571. Any suggestions are welcome.

dcshzj:
> @dcshzj I have changed the test-repo location to the one on my account and it passes all the tests.

Sorry @chan-j-d, is it possible to change your `master` branch in your `testrepo-Alpha` fork, instead of using a separate branch? Our tests works only on the `master` branch.

You can just make the same change to the `master` branch directly in your fork.

chan-j-d:
Hi @dcshzj, I am slightly confused by the suggestion. I have made the change to my own master branch but it causes multiple other tests to fail when trying to run the tests. Would like to clarify if the intention is to run the test on the `master` branch itself and if I am supposed to update values like the `LATEST_COMMIT_HASH` in `GitTestTemplate` and the `assertEquals` value of 4 to 5 in `FileInfoExtractorTest::extractFileInfos_sinceDateFebrauaryNineToLatestCommit_success` etc.

Would like to get a clarification before I push a change to this branch.

dcshzj:
@chan-j-d Yep, you will need to update some other parts of the test suite like you mentioned.

chan-j-d:
Hi @dcshzj, I have updated the test cases as requested. I have changed only test cases that failed and adjusted them to my best ability to match their original intention. Would like to request a proper review. Thanks!

dcshzj (approved):
LGTM!

chan-j-d:
Hi @fzdy1914, requesting the second-level of review. Thanks!

src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -28,7 +28,7 @@
  * Contains templates for git testing.
  */
 public class GitTestTemplate {
-    protected static final String TEST_REPO_GIT_LOCATION = "https://github.com/reposense/testrepo-Alpha.git";
+    protected static final String TEST_REPO_GIT_LOCATION = "https://github.com/chan-j-d/testrepo-Alpha.git";
```
fzdy1914, line 31: Should be changed back?

fzdy1914 (approved):
Other looks good to me

src/test/java/reposense/template/GitTestTemplate.java
chan-j-d: Yep! Thanks for the reminder. Was leaving it in to show the test cases passing.

chan-j-d:
Hi @dcshzj, this PR has received both approvals and I have reverted the testrepo location. Now it just requires the `space test.txt` file to be merged to the test repo to work. Thanks for your hard work!

PR #1577, opened 2021-08-01
[#1463] Fix case insensitive bug in @@author feature

### LINKED ISSUE #1463: Warnings in the log
Refer to the following warnings:

image

https://github.com/nus-cs2103-AY2021S2/tp-dashboard/runs/2048437170?check_suite_focus=true

Are they necessary? At least some of them don't seem necessary e.g., ones caused by case difference

FH-30:
Resolves #1463.

We should remove ```String.CASE_INSENSITIVE_ORDER``` from the instance of ```authorDetailsToAuthorMap``` in ```AuthorConfiguration.java``` as shown below as it results in a bug with the @@author feature of RepoSense:

You can refer to this peculiar TreeMap's (the one used in the current implementation of RepoSense) behavior here:
Screen Shot 2021-07-29 at 2 03 06 AM

In addition, we should replace the TreeMap with a HashMap because we no longer need to use a comparator for the Map and a HashMap is more efficient than a TreeMap in terms of searching (O(1) for HashMap vs O(logn) for TreeMap).

Before reading my justification, do take note of this:
 ```authorDetailsToAuthorMap``` in ```AuthorConfiguration.java``` is used only in ```AnnotatorAnalyzer.java``` (This file contains methods for implementing the @@author annotation feature of RepoSense).

Justification is as follows

Current behavior of RepoSense:
|```Author's GitHub ID``` column in ```author-config.csv``` | ```Author's Git Author Name``` column in ```author-config.csv``` | Content of ```authorDetailsToAuthorMap``` (In chronological order excluding email keys)|
|--------------------------------------------------|----------------|-----------------------|
|FH-30|Francis Hodianto|{FH-30 -> FH-30 author object, Francis Hodianto -> FH-30 author object}|
|Anonymous|fh-30|{FH-30 -> Anonymous author object, Francis Hodianto -> FH-30 author object, Anonymous -> Anonymous author object}|

The warning logged:
```Duplicate alias fh-30 found. The alias will belong to the last author - Anonymous```.

Now when this happens, whenever the repository consisting the two authors above are analyzed and the actual author FH-30 uses ```@@author FH-30``` wherever he writes code to mark it as his, it will be assigned to Anonymous instead.

Reading the warning logged also does not explain anything to the user because from the user's perspective, ```FH-30``` is different from ```fh-30``` so by reading the warning the user will only assume that ```@@author fh-30``` will be assigned to Anonymous but will not expect ```@@author FH-30``` to also be assigned to Anonymous.

Behavior of RepoSense if  ```String.CASE_INSENSITIVE_ORDER``` is removed:
|```Author's GitHub ID``` column in ```author-config.csv``` | ```Author's Git Author Name``` column in ```author-config.csv``` | Content of ```authorDetailsToAuthorMap``` (In chronological order)|
|--------------------------------------------------|----------------|-----------------------|
|FH-30|Francis Hodianto|{FH-30 -> FH-30 author object, Francis Hodianto -> FH-30 author object}|
|Anonymous|fh-30|{FH-30 -> FH-30 author object, Francis Hodianto -> FH-30 author object, fh-30 -> Anonymous author object, Anonymous -> Anonymous author object}|

If we change the ```authorDetailsToAuthorMap``` from a ```TreeMap``` to a ```HashMap```, ```@@author FH-30``` will be assigned to FH-30 and ```@@author fh-30``` will be assigned to Anonymous. Not to mention there will be no warning logged in this scenario.

This also resolves issue #1463 because the warnings in the log are reduced to the necessary ones only as they are mostly generated due to the fact that githubIds and gitIds are considered duplicates even if they have case differences.

Now the warning will only be logged when gitIds and githubAliases are exactly the same in terms of characters and cases.

Commit Message:
```
Fix case insensitive bug of Map object used in @@author feature 
implementation

In AuthorConfiguration.java, the authorDetailsToAuthorMap object
utilized in AnnotatorAnalyzer.java for implementing the @@author
feature marks githubIds and gitIds with the same character sequence
but different casing as duplicates resulting in unnecessary warnings
in the log. When duplicates are encountered previous mappings might
also be altered resulting in potential unexpected assignings when
using the @@author feature.

Let's change the type of Map associated with the
authorDetailsToAuthorMap object from a TreeMap with
String.CASE_INSENSITIVE_ORDER as its Comparator to a HashMap.
```


dcshzj (approved):
LGTM! Neat work and really great analysis, thank you!

fzdy1914 (approved):

fzdy1914 (approved):

PR #1580, opened 2021-08-08
Cypress: replace nonexistent input element with pug id in selector in codeView_reload.js

FH-30:
Fixes failing Cypress test in ```codeView_reload.js``` which occurs after merging in [PR#1412](https://github.com/reposense/RepoSense/pull/1412)

Commit Message:
```
Fix failing Cypress test in codeView_reload.js

Cypress test in codeView.js is failing because the test analyzes
RepoSense's repository itself to run tests on. When PR#1412
was merged in, RepoSense no longer has any pug files whereas
the test case assumes pug files are still there thus leading to an
error. The selector utilized in the test refers to an input element
with id pug that no longer exists.

Let's change the selector in the test to refer to an input element
with id java that still exists in RepoSense's repository.
```

### DIFF
--- a/frontend/cypress/tests/codeView/codeView_reload.js
+++ b/frontend/cypress/tests/codeView/codeView_reload.js
@@ -29,7 +29,7 @@ describe('reload page', () => {
         .uncheck()
         .should('not.be.checked');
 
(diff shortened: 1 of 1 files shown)


gerhean (approved):
LGTM

PR #1581, opened 2021-08-10
[#1536] Frontend: show short commit hash in commits view

### LINKED ISSUE #1536: Commits view: show commit hash
We can show the last part of the commit hash here so that it is easy to uniquely identify a commit. The commit subject may not be unique.

yhtMinceraft1010X:
Resolves #1536 

```
As commit subject lines can be identical, this may cause confusion to
the user when browsing the list of commits. Having 4 characters of 
the commit hash may be insufficient for differentiating between
commits.

Let's display the first 7 characters of the commit hash in the commits
view.
```

Current UI:
image


frontend/src/views/v-zoom.vue
```diff
@@ -94,6 +94,9 @@
       v-bind:key="slice.hash",
       v-bind:class="{ 'message-body active': slice.messageBody !== '' }"
     )
+      .tooltip
+        | Commit hash: {{ slice.hash.substr(0, 4) }} &nbsp;
```
HCY123902, line 98: I am actually not sure whether 4 digits is enough for identifying a commit in a repository, especially for project with a large number of commits. Is it possible to change it to 7 digits instead?

HCY123902:
@damithc Hello professor, what do you think of this UI layout?

image

frontend/src/views/v-zoom.vue
damithc: Yes, I too think 7 is better.

damithc:
> @damithc Hello professor, what do you think of this UI layout with the commit hash?

I think we can omit `Commit hash:` part. Perhaps show the commit hash with a very light grey background? i.e., similar to tags but lighter grey.

gerhean:
I rather the commit hash be added near the back of the title since it's just extra information which is meaningless 99% of the time but takes up too much space in front. Plus having the title link in the middle only is ugly.

damithc:
> I rather the commit hash be added near the back of the title since it's just extra information which is meaningless 99% of the time but takes up too much space in front. Plus having the title link starting from the middle is ugly.
> 
> Something like:
> [xxx] Add Readme ... (51 lines) (34t5h78...) `tests` ...

Yes, I agree. In fact, it can even be an icon that reveals the hash upon hover/click?

yhtMinceraft1010X:
An update on what the hash currently looks like on the screen:
image

damithc:
> An update on what the hash currently looks like on the screen:
> image

Looks good. Might not even need the tooltip as git users would know that this is. The report is heavy in content already -- best not to add elements that are not that useful.  Not sure we need the brackets either. But do consider styling it like this: `ds8732k`

yhtMinceraft1010X:
Taking into account the changes suggested by Prof @damithc, this is what the ui looks like now:
image

damithc:
> Taking into account the changes suggested by Prof @damithc, this is what the ui looks like now:
> image

OK for me 👍

HCY123902 (approved):

dcshzj (approved):

Issue #1560, opened 2021-07-06
https://reposense.org outdated


damithc:
It looks like https://reposense.org/ was updated last in Jun 2020. That means it's not matching the current release version, right?

HCY123902:
Yes, it has not been updated since release V1.12

HCY123902:
Yes, it has not been updated since V1.12. I will try to make a documentation release this week

damithc:
Thanks @HCY123902

FH-30:
![Screen Shot 2021-08-10 at 12 00 29 AM](https://user-images.githubusercontent.com/61911161/128744416-d5e9113b-fa3e-42e3-9bac-6251695a0004.png)

Hi, just wondering if the lack of documentation for the git folder here is intended or is it because the UG is outdated as mentioned in this issue?

HCY123902:
> ![Screen Shot 2021-08-10 at 12 00 29 AM](https://user-images.githubusercontent.com/61911161/128744416-d5e9113b-fa3e-42e3-9bac-6251695a0004.png)
> 
> Hi, just wondering if the lack of documentation for the git folder (ex: GitCatFile is not documented) here is intended or is it because the UG is outdated as mentioned in this issue?

Yes, this is created in #1449. I will adjust the documentation accordingly

dcshzj:
The website still seems to be outdated even with the v2.0 release. @HCY123902 are you able to update the website?

HCY123902:
> The website still seems to be outdated even with the v2.0 release. @HCY123902 are you able to update the website?

I have attempted to deploy the documentation, It seems that the changes cannot be pushed to /reposense with empty `baseUrl` in `site.json`. Let me create a corresponding issue on the Markbind side

damithc:
@HCY123902 Are you following the instructions in https://reposense.org/dg/projectManagement.html#deploying-the-production-website ?

In particular, are you on the release branch? It has a slightly different site.json file

HCY123902:
Yes, there is the `CNAME` pointing at `reposense.org` and the `filterTags` has been set to `production` in `site.json`, as is indicated in the deploy guide. However, the site is still deployed to `reposense.org/RepoSense`

HCY123902:
Can I check who published the site in June last year?

damithc:
> Can I check who published the site in June last year?

Looks like it was me :-p

@HCY123902 I think the site.json should be updated as follows (on the `release` branch):

```html
"deploy": {
[...]

PR #1593, opened 2021-09-30
Release branch update

HCY123902:
Related to #1560 

The current release of RepoSense 2.0 contains the latest implementation and documentation corresponding to the master branch. The corresponding documentation at `reposense.org/RepoSense` also matches the master branch, generated by MarkBind 3.0.6. However, the release branch is outdated and it needs to be merged with the master branch.

Note that `site.json` is the same in the 2 branches. The filter tags include both `-production` and `dev`. This may need to be adjusted, as the release branch should only have the `production` tag.

Commit message:
```
Merge the release branch with the current master branch

The current release branch was outdated compared with the
current master branch, The release branch documentation is written
in MarkBind 2.0 while the master branch documentation has been
migrated to MarkBind 3.0.

It is necessary to merge the release branch with the current
master branch as a prerequisite of publishing the documentation
at `reposense.org`.

Let's merge the release branch with the current master branch
```


dcshzj:
I think it might not be a good idea to have 2 PRs every time we make a release. Instead, I propose to use rebasing to update the `release` branch, which will allow us to have both the `master` and the `release` branches to have the same commit history, with the `release` branch having additional commits that are specific to it (currently to get the documentation site working).

To outline the proposed process of making a release:
1. Make a release on GitHub as already outlined in our developer documentation. There is no change to this step in the current workflow.
2. Temporarily configure the branch protection rules on the `release` branch to allow for force pushes (see screenshot below):

image

3. Rebase the `release` branch to the `master` branch. This step will put aside any additional commits on the `release` branch, apply all the additional commits from the `master` branch that is not yet present in the `release` branch, and then apply the commits put aside onto the new `release` branch. The following picture illustrates it (main branch is the `master` branch and your branch is the `release` branch, [source](https://www.mohammedasker.com/explaining-how-git-rebase-works-in-laymans-terms-%F0%9F%8D%95-%F0%9F%A5%91-%F0%9F%8D%96)):

image

4. Force push the new `release` branch to GitHub.
5. Revert the temporary configuration of the branch protection rules on the `release` branch, to disable force pushes.

If additional commits need to be made to the `release` branch, the usual pull request workflow to the `release` branch is sufficient. If there are any conflicts during step 3, they can be resolved locally by the person making the release. This also means that only administrators have the ability to make new releases (since the branch protection rules need to be changed in step 2).

I think this way allows us to maintain the integrity of the `release` branch and no additional pull requests need to be made for every new release. It should also be less error prone as the custom changes on the `release` branch do not have to be made every time a release happens. @HCY123902 and Prof @damithc what do you think?

HCY123902:
> I think it might not be a good idea to have 2 PRs every time we make a release. Instead, I propose to use rebasing to update the `release` branch, which will allow us to have both the `master` and the `release` branches to have the same commit history, with the `release` branch having additional commits that are specific to it (currently to get the documentation site working).
> 
> To outline the proposed process of making a release:
> 
> 1. Make a release on GitHub as already outlined in our developer documentation. There is no change to this step in the current workflow.
> 2. Temporarily configure the branch protection rules on the `release` branch to allow for force pushes (see screenshot below):
> 
> image
> 
> 1. Rebase the `release` branch to the `master` branch. This step will put aside any additional commits on the `release` branch, apply all the additional commits from the `master` branch that is not yet present in the `release` branch, and then apply the commits put aside onto the new `release` branch. The following picture illustrates it (main branch is the `master` branch and your branch is the `release` branch, [source](https://www.mohammedasker.com/explaining-how-git-rebase-works-in-laymans-terms-%F0%9F%8D%95-%F0%9F%A5%91-%F0%9F%8D%96)):
> 
> image
> 
> 1. Force push the new `release` branch to GitHub.
> 2. Revert the temporary configuration of the branch protection rules on the `release` branch, to disable force pushes.
> 
> If additional commits need to be made to the `release` branch, the usual pull request workflow to the `release` branch is sufficient. If there are any conflicts during step 3, they can be resolved locally by the person making the release. This also means that only administrators have the ability to make new releases (since the branch protection rules need to be changed in step 2).
> 
> I think this way allows us to maintain the integrity of the `release` branch and no additional pull requests need to be made for every new release. It should also be less error prone as the custom changes on the `release` branch do not have to be made every time a release happens. @HCY123902 and Prof @damithc what do you think?

Actually, it is possible to just release with the master branch? Currently, the purpose of the release branch is to make it more convenient to publish documentation at `reposense.org`. However, the fact that `reposense.org` is an organization site means it may only be possible to modify `reposense.org` in https://github.com/reposense/reposense.github.io, which makes the `release` branch less meaningful than intended.

It may be more convenient just to release with the master branch, and then deploy the documentation to the `reposense.github.io` repository. There is no `release` branch involved in this case

damithc:
> Actually, is it possible to just release with the master branch?

A release branch is useful when we need to hot-patch a previous release to fix a critical bug. In that case, the PR can come against the release branch rather than the master branch.

The release is usually done by someone with admin permissions, which means merging master to release doesn't need a PR.

While rebasing the release branch is possible, I  think it is more common to maintain two parallel branches, which shows the release history more visually, including hot-patches.

@fzdy1914 can weigh in too.

HCY123902:
I agree that direct merging will make release more convenient. However, it seems that I currently do not have access to merge and then directly push the changes to the release branch, or rebase it

damithc:
> I agree that direct merging will make release more convenient. However, it seems that I currently do not have access to merge and then directly push the changes to the release branch, or rebase it, without creating this pull request. Can I propose to merge this pull request first? After merging and publishing `reposense.org`, I will modify the `site.json` in the release branch and the release guide

@HCY123902 I've removed protection from the `release` branch. You should be able to push to it now.

damithc:
@HCY123902 Next time, do a normal merge locally and push (not via a PR). No need for a lengthy commit message as merging master to release is simply a default step of releasing a new version.

HCY123902:
Noted. Many thanks

damithc:
BTW, something is not right with the revision tree. I don't see a merge commit showing where master branch was merged to the release branch.

HCY123902:
Oh the merge is fast forwarded. There is no additional commit created by the merge. Should I reset and merge again? Really sorry for not noticing this

damithc:
> Oh the merge is fast forwarded. There is no additional commit created by the merge. Should I reset and merge again? Really sorry for not noticing this

No worries. We can leave it be. Next time, use a merge commit.

PR #1406, opened 2021-01-17
[#1401] Replace images as puml diagrams

rolandyuwy:
Resolves #1401

```
The Developer Guide uses screenshots of diagrams for 
documentation.

This makes it more tedious to update diagrams.

Let's use MarkBind's native support for PlantUML by tagging diagrams to
.puml files of diagrams in the documentation.

This allows for easier updating of the Developer Guide in the future 
through updating the .puml files directly.

Refer to MarkBind's User Guide on for PlantUML usage:
https://markbind.org/userGuide/usingComponents.html

Refer to PlantUML's Reference Guide for PlantUML syntax:
https://deepu.js.org/svg-seq-diagram/Reference_Guide.pdf

PR #1598, opened 2021-10-11
Fix non-rendering puml diagrams

yhtMinceraft1010X:
I noticed that some puml diagrams are not rendering properly on chrome:
image

2 approaches I thought of were to expand the puml tag or to convert the puml to images. I went with the latter to standardise with the conversion of ReportArchitecture.puml to a png.

Commit message:
```
Fix broken puml

The puml diagrams for html report may not render on Chrome.

Let's convert them into png images.
```


dcshzj:
The code you changed was introduced in #1406, and I think we should stick with using MarkBind's native support for PlantUML diagrams. @yhtMinceraft1010X is this issue still present, and if it is, can you file an upstream task on the MarkBind repository, letting them know that the diagram cannot be rendered in Chrome? Thank you.

yhtMinceraft1010X:
Sorry for the late reply. The issue has been fixed.

PR #1604, opened 2021-12-03
[#1564] Lint script: add files under cypress folder

### LINKED ISSUE #1564: Cypress tests folder is not linted
Lint `frontend/cypress/tests`

In `./package.json` line 10:
https://github.com/reposense/RepoSense/blob/c2a25380a3bdd678483223ccf3aca139f33ee76b/package.json#L10

We see that `frontend/cypress/**.js` is being linted. However this only lints `frontend/cypress/support.js`

The correct expression to lint is `frontend/cypress/tests/**/*.js frontend/cypress/support.js`, to avoid linting the `node_modules` folder in `frontend/cypress/`.

Fix lint expression and the subsequent linting errors (most can be solved by calling lintfix).

ganhongyao:
Closes #1564 

Proposed commit message:
```
For the frontend, the current "lint" script does not pick up files
under the "cypress" folder.

This should be fixed so that any programmatic and stylistic errors in
these files can be detected and corrected.

Let's add the files under the "cypress" folder to the lint script
and fix violations of the existing code.
```

### DIFF
--- a/frontend/cypress/tests/codeView/codeView_load_benchmark.js
+++ b/frontend/cypress/tests/codeView/codeView_load_benchmark.js
@@ -1,4 +1,4 @@
-describe('load code view benchmark', function() {
+describe('load code view benchmark', () => {
   const NUM_TRIALS = 5;
   const THRESHOLD_LOADING_TIME = 8000;
   const THRESHOLD_LOADING_TIME_SECONDS = THRESHOLD_LOADING_TIME / 1000;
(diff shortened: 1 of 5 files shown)


gerhean:

Please read http://se-education.org/guides/conventions/github.html for PR naming conventions, we use option 2.

frontend/cypress/tests/codeView/codeView_load_benchmark.js
```diff
@@ -1,61 +1,65 @@
...
           }
         });
   };
 
-
-  for (let i = 0; i < NUM_TRIALS; i++) {
-    it(`time taken to load code view (trial ${i+1})`, function() {
-      if (isATrialWithinMaxTime) {
+  for (let i = 0; i < NUM_TRIALS; i += 1) {
+    const shouldSkipCurrentTrial = isATrialWithinMaxTime;
```
gerhean, line 53: Is there a reason for making this a new variable?

frontend/package.json
```diff
@@ -9,8 +9,8 @@
   "scripts": {
     "serve": "vue-cli-service serve",
     "serveOpen": "vue-cli-service serve --open",
-    "lint": "eslint src/**/*.{js,vue} && stylelint ./src/**/*.{vue,scss} && npm run puglint",
-    "lintfix": "eslint --fix src/**/*.{js,vue} cypress/tests/**/*.js && stylelint --fix ./src/**/*.{vue,scss}",
+    "lint": "eslint src/**/*.{js,vue} cypress/tests/**/*.js cypress/support.js && stylelint ./src/**/*.{vue,scss} && npm run puglint",
```
gerhean, line 12: Would it be better to combine `cypress/tests/**/*.js` `cypress/support.js` into `cypress/**/*.js`?

gerhean (commented):
Otherwise LGTM

frontend/package.json
ganhongyao: That would mean that files under `node_modules` will be linted as well. Also because I referred to 

> The correct expression to lint is frontend/cypress/tests/**/*.js frontend/cypress/support.js, to avoid linting the node_modules folder in frontend/cypress/.

in the [issue description](https://github.com/reposense/RepoSense/issues/1564).

frontend/cypress/tests/codeView/codeView_load_benchmark.js
ganhongyao: This was done because of the linting error `no-loop-func` - see [here](https://eslint.org/docs/rules/no-loop-func#disallow-functions-in-loops-no-loop-func) for more information.

To get around that, I made a copy of the variable in the scope of the loop iteration.

gerhean: I think this was done intentionally so that the inside of the function can reference `isATrialWithinMaxTime` as a global variable.

frontend/package.json
gerhean: Ok

gerhean (approved):
LGTM

gerhean:
Actually please fix the Cypress frontend tests issues first.

gerhean:
Can I confirm that there is really no more elegant way to handle the eslint problems than the `eslint-disable-next-line` comments?

ganhongyao:
> Can I confirm that there is really no more elegant way to handle the eslint problems than the `eslint-disable-next-line` comments?

I think disabling the lint rule for the unnamed function is fine since that's the norm for mocha test functions. 

For the other lint rule, I can try to work out an alternative implementation.

gerhean (approved):
LGTM, checks all passed.

dcshzj (approved):

fzdy1914 (approved):

Issue #1597, opened 2021-10-10
Duplicate error message for missing repos


damithc:
Each error is shown twice:
![image](https://user-images.githubusercontent.com/1673303/136682644-daab28b3-0db8-4619-a3ef-778daf1c9f39.png)

https://nus-tic2002-2021.github.io/ip-dashboard/?search=&sort=groupTitle&sortWithin=title&timeframe=commit&mergegroup=&groupSelect=groupByRepos&breakdown=true&checkedFileTypes=java~md~fxml~sh~bat~gradle~txt&since=2021-08-09

Settings:

https://github.com/nus-tic2002-2021/ip-dashboard

gerhean:
Multiple error as the problem repo is listed in both `author-config.csv` and `repo-config.csv` and Reposense does not really record down problematic repos to skip.

Solution is probably to remove duplicates in error message list before writing to json

PR #1605, opened 2021-12-11
[#1597] Remove duplicate error messages in error summary

Zhou-Jiahao-1998:
Fixes https://github.com/reposense/RepoSense/issues/1597

Proposed commit message:
```
Duplicate error messages appear in the error summary shown in the
final report when the same repository has errors in multiple
configuration files.

Let's change the error summary data structure from List to Set to
remove duplicate error messages without affecting performance. The
order of error messages may change, but should not be a problem.
```


gerhean:
Can you verify that your implementation works by creating a test case?

A set of mutable objects (map) sometimes have unexpected behavior (as the hash function is not properly defined).

Zhou-Jiahao-1998:
@gerhean I hope I am addressing your question. According to [Java 11 Map](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/AbstractMap.html#equals(java.lang.Object)), two `Map` objects are equal if the mappings are identical. Since we only add into the set and do not modify the `Map` in any ways, we can treat each `Map` as immutable. I have added a test for a quick check.

gerhean:
@Zhou-Jiahao-1998 By test I mean that you should try to recreate the situation where the duplicated error messages occurred, and ensure that the error messages indeed do not duplicate anymore.

Zhou-Jiahao-1998:
@gerhean Alright I get what you mean now. Give me some time to figure out how to test this out without manually testing.

gerhean:
Seems to be fine but I prefer that you verify that this solution works even if the problem repo is listed in the separate files of `author-config.csv` and `repo-config.csv`

dcshzj (changes requested):
Per above

Zhou-Jiahao-1998:
@dcshzj Took me some time to figure out how to do system test. I added an incorrect repo in both `author-config.csv` and `repo-config.csv` on the last line. This helps to check if the duplicated error is shown.

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -46,6 +47,55 @@ public void repoLocationParser_parseEmptyString_success() throws Exception {
         RepoLocation repoLocation = new RepoLocation("");
     }
 
+    @Test
+    public void duplicate_error_removed() {
```
dcshzj, line 51: The function name should follow our coding conventions: `featureUnderTest_testScenario_expectedBehavior()`

dcshzj (changes requested):

dcshzj (approved):
LGTM!

fzdy1914 (approved):

Issue #1603, opened 2021-11-25
Warn if a config file has an unknown column


damithc:
RepoSense requires config file column headers to follow the exact column name as specified.

Suggestion: If a column name doesn't match an expected column name, give a warning (even aborting might be appropriate) as it is likely to be a user mistake that could affect the output.

On a side note, perhaps column name matching can be case-insensitive?

PR #1609, opened 2021-12-12
[#1603] Terminate if a CSV config file has an unknown column

Zhou-Jiahao-1998:
Fixes https://github.com/reposense/RepoSense/issues/1603

Proposed commit message:

```
RepoSense does not show any error if a config file has an unknown
column as it is likely to be a user mistake that could affect the
output.

Let's add a new exception (InvalidHeaderException) to handle this, as
using InvalidCsvException will not terminate the program. The check is
done within CsvParser::validateHeader.

The original code already handles case insensitive matching of column
names.
```


gerhean:

Please read http://se-education.org/guides/conventions/github.html for commit message subject formatting, please do check out how they are formatted from past PRs as well.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -223,17 +226,34 @@ private String getRowContentAsRawString(final CSVRecord record) {
      * Generates map of column header to position number for input {@code possibleHeader}.
      * @throws InvalidCsvException if {@code possibleHeader} does not contain all the mandatory headers.
      */
-    private void validateHeader(String[] possibleHeader) throws InvalidCsvException {
+    private void validateHeader(String[] possibleHeader) throws InvalidCsvException, InvalidHeaderException {
         int headerSize = possibleHeader.length;
+        Set<String> knownColumns = new HashSet<>();
+        StringBuilder unknownColumns = new StringBuilder();
```
dcshzj, line 232: I understand you are trying to build the string for the final output here, but I think it's better practice to store as a string array first, then use `String.join()` at the end when throwing the InvalidHeaderException.

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -223,17 +226,34 @@ private String getRowContentAsRawString(final CSVRecord record) {
...
+                    knownColumns.add(possible);
                     break;
                 }
             }
+            if (!knownColumns.contains(possible)) {
+                if (unknownColumns.toString().equals("")) {
+                    unknownColumns.append(possible);
+                } else {
+                    unknownColumns.append(", ");
+                    unknownColumns.append(possible);
```
dcshzj, line 247: Per the above comment, this should be able to be simplified into something that does not need an if-else statement.

src/test/resources/AuthorConfigParserTest/authorconfig_unknownHeaders_test.csv
```diff
@@ -0,0 +1,3 @@
+Repository Location,Branch,Author's GitHub ID,Author's Emails,Author's Display Name,Author's Git Author Name,Ignore Glob
```
dcshzj, line 1: Can we change the unknown headers to be something more obvious? Something like "Totally Not A Real Header" or something better. I can't really tell directly which header should not be there.

src/test/resources/GroupConfigParserTest/groupconfig_unknownHeader_test.csv
```diff
@@ -0,0 +1,5 @@
+Group,Glob,Repository Location
```
dcshzj, line 1: Same for this as well, to make the unknown header more obvious.

src/test/resources/RepoConfigParserTest/repoconfig_unknownHeaders_test.csv
```diff
@@ -0,0 +1,2 @@
+Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config,Ignore Commits List,Ignore Authors Lists,Shallow Cloning,Find Previous Author
```
dcshzj, line 1: Same for this as well, to make the unknown header more obvious.

dcshzj (changes requested):

dcshzj (approved):
LGTM!

fzdy1914 (approved):
feels like InvalidHeaderException should be a child class of InvalidCSVException, others LGTM

Zhou-Jiahao-1998:
> feels like InvalidHeaderException should be a child class of InvalidCSVException, others LGTM

I agree and changes are made

dcshzj:
@Zhou-Jiahao-1998 There are still some `throws InvalidHeaderException`, can you remove them too?

dcshzj:
Looks great!

Zhou-Jiahao-1998:
> 

Ah yes that was the whole reason for making another exception. I revert to the previous commit after realising the problem. Thanks for pointing it out!

dcshzj:
Sorry @Zhou-Jiahao-1998 I think you can revert back to the original approved commit. `InvalidHeaderException` should be a subclass of `InvalidCsvException`, and the program correctly terminates when an `InvalidCsvException` is encountered.

Zhou-Jiahao-1998:
@dcshzj In `PepoSense::main`, the only time the program terminates because of `InvalidCsvException` is when `getRepoConfigurations` (line 61) throws it. In the same file we can find `getRepoConfigurations` (line 116). `InvalidCsvException` can only be thrown at line 118 which is when processing `repo-config`. Therefore, if the problem occurs in other config files, the program will not be terminated because the two try catch blocks below both handle `InvalidCsvException`, no exceptions will be thrown. 

Making `InvalidHeaderException` a subclass will mean that we need to have a special case for the subclass as we want it to behave differently from its parent exception. This is a violation of LSP in my opinion. Let me know what do you think.

dcshzj:
Ah I see what you mean. I think it might be a bit confusing in terms of the behavior of `InvalidCsvException`. I understand from #872 that repo-config.csv is considered a mandatory file and the other two are considered optional. This was the reasoning why any errors in repo-config.csv will terminate the program whereas errors in the other two files will only result in a warning.

I think we should retain the original behavior of terminating only for errors in repo-config.csv. Having `InvalidHeaderException` a subclass of `InvalidCsvException` should now be fine, since they now share the same behavior.

Zhou-Jiahao-1998:
@dcshzj Do you think we should ask prof for his inputs as he opened this issue?

damithc:
> @dcshzj Do you think we should ask prof for his inputs as he opened this issue?

Guys, what exactly do you want my input on? i.e., what's the question, what are the options, pros and cons of each, your recommendation (if any)?

Zhou-Jiahao-1998:
@damithc The issue here is that I added a new `InvalidHeaderException` for the case where any config file contains unknown columns. Throwing this exception will terminate the program gracefully and output the error message.

The other option we have is to follow the current format of `InvalidCsvException` and make `InvalidHeaderException` a subclass of `InvalidCsvException`. The program is only terminated if the error is thrown from parsing `repo-config`. This is how the program handles parsing errors in config files now.

Using the first approach, we will terminate the program as long as there is an unknown column. I personally prefer this option as mentioned in the issue. It is likely a mistake from the user side and the mistake might affect the actual output.

If we use the second approach, we will have a smoother behaviour as all exceptions behave the same way in different config files. It is also logical to think that `InvalidHeaderException` can be a subclass of `InvalidCsvException`.

damithc:
> @damithc The issue here is that I added a new `InvalidHeaderException` for the case where any config file contains unknown columns. Throwing this exception will terminate the program gracefully and output the error message.
> 
> The other option we have is to follow the current format of `InvalidCsvException` and make `InvalidHeaderException` a subclass of `InvalidCsvException`. The program is only terminated if the error is thrown from parsing `repo-config`. This is how the program handles parsing errors in config files now.
> 
> Using the first approach, we will terminate the program as long as there is an unknown column. I personally prefer this option as mentioned in the issue. It is likely a mistake from the user side and the mistake might affect the actual output.
> 
> If we use the second approach, we will have a smoother behaviour as all exceptions behave the same way in different config files. It is also logical to think that `InvalidHeaderException` can be a subclass of `InvalidCsvException`.

The actual implementation details, I'll leave you to decide.
But I can give inputs from external behavior PoV. The question seems to be: should we (a) terminate or (b) warn only, if there is an unknown column.

Pros of terminating: the user will have no choice but to fix.
Cons: the user will not be able to keep extra columns in the file, e.g., for reference purposes.

I'm leaning slightly towards terminating, as it avoids the situation of a user missing the warning and a mistake in the column name going unnoticed for a while.

dcshzj (approved):
Yup I think let's keep it this way where `InvalidHeaderException` is a separate class from `InvalidCsvException`.

dcshzj:
@fzdy1914 Can you confirm if this is a good design?

fzdy1914 (approved):

Issue #1591, opened 2021-09-26
Omit percentile if the list is sorted alphabetically


damithc:
This percentile is meaningless and potentially confusing to the viewer when the list is sorted alphabetically.

![image](https://user-images.githubusercontent.com/1673303/134794571-f0d68178-0dd7-4d5a-b7e0-92f2756505da.png)

Suggested: omit it if the list is sorted by title

gerhean:
Actually I'm not very sure of why the percentile is even there for in the first place.

From what I have seen it is just the index of the repo in the list divided by number of repos, which is mostly useless since you can already easily gauge which index you are at by looking at scroll bar.

dcshzj:
It was originally introduced in #922 which I understand is to help in the grading.

gerhean:
In that case I think it's better to just leave it for consistency, since it would be even more confusing to have a number which only appears there sometimes for a seemingly arbitrary reason (It's not obvious to anyone who isn't grading what this number is for).

damithc:
It is not a major feature; just to easily find things such as _what is the 20th percentile of LoC contribution_. It only matters when the list is sorted by a quantitative criterion. In fact the only time I use it is when the sorting is done by the contribution. Perhaps we show it only when sorted by contribution?

PR #1610, opened 2021-12-13
[#1591] Omit 'ignore globs' files from processing

Zhou-Jiahao-1998:
closes https://github.com/reposense/RepoSense/issues/1591

### DIFF
--- a/frontend/src/components/v-summary-charts.vue
+++ b/frontend/src/components/v-summary-charts.vue
@@ -72,7 +72,9 @@
               v-bind:class="{ 'active-icon': isSelectedTab(repo[0].name, repo[0].repoName, 'zoom', true) }"
             )
             span.tooltip-text Click to view breakdown of commits
-      .summary-charts__title--percentile {{ getPercentile(i) }} %
+      .summary-chart__title--percentile(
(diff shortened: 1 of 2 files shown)


gerhean:

Please read http://se-education.org/guides/conventions/github.html for commit message subject formatting, please do check out how they are formatted from past PRs as well.

By the way, percentage within the group is based on the `sort group by` option instead of the `sort within groups by` option, might need to check if this behavior is desired.

gerhean (approved):
LGTM

gerhean (approved):
LGTM

dcshzj:
@gerhean Are we able to add a cypress test case here?

gerhean:
> @gerhean Are we able to add a cypress test case here?

Yes we can actually.

dcshzj:
@Zhou-Jiahao-1998 Yup I think we should add a cypress test case here, so that we ensure the percentile only appears when it should

gerhean:
Maybe add tests to show that `group by` and `sort within groups` doesn't affect this functionality.

gerhean:
The name for the new tests are a bit confusing, I would prefer something like `percentage is shown even after changing other fields`

gerhean (approved):
LGTM

fzdy1914 (approved):

PR #1620, opened 2022-01-18
Support frontend hot reload on Linux

gok99:
**Proposed commit message:**

```
The hotReloadFrontend gradle task does not support Linux. Apache
ant's `Os` class does not have an OS family specifically for Linux.
The Mac hot reload feature also does not appear to work.

Let's:
- add hot reload for Linux using gradle's OperatingSystem::isLinux
- use `npm run serveOpen` command for both Mac and Linux
```

Tested on Windows, Mac and Fedora Linux

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,4 @@
-import org.apache.tools.ant.taskdefs.condition.Os
+import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform
 
 plugins {
     id 'application'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -228,22 +230,30 @@ task syncFrontendPublic(type: Sync) {
...
+    onlyIf {os.isWindows()}
     workingDir 'frontend/'
     commandLine 'cmd','/c', 'START', '"hotreload RepoSense frontend"', 'npm', 'run', 'serveOpen'
 }
 
+task linuxHotReloadFrontend(type: Exec) {
+    dependsOn(installFrontend)
+    onlyIf {os.isLinux()}
+    workingDir 'frontend/'
+    commandLine 'npm', 'run', 'serveOpen'
```
gok99, line 249: This works for me, but I'm not sure if there's a canonically correct way to write this.

gerhean:
For Windows OS, the reason why I spawn a new window for hotreload is due to Gradle causing the process to be orphaned (I think).

In other words, if I `npm run serveOpen` without openning new window, doing  `ctrl-c` to kill the application  will cause the Vue process to run in the background and not terminate properly. 

Can you check if such is also the case for linux?

build.gradle
gerhean: There is no canonically correct way, there is only consistency, and I find it consistent other than the fact that this one doesn't spawn the `serveOpen` in a new window which may or may not be needed.

gerhean (approved):
LGTM. No regressions on Windows.

fzdy1914 (approved):

gok99:
> Can you check if such is also the case for linux?

The `vue-cli-service serve` process seems to stop after a `ctrl-c` and the locally hosted site is no longer accessible.

fzdy1914 (changes requested):
See our discussion in slack for more information.

fzdy1914 (approved):

dcshzj:
Confirmed to work on ARM-based Macs as well.

PR #1625, opened 2022-01-22
[#1601] Convert datetime functionalities to java.time package

### LINKED ISSUE #1601: Intermittent NumberFormatException
See #1600 #1599 

Perhaps we should add a retry mechanism as a temporary workaround, until we are able to pin down the exact cause?

yhtMinceraft1010X:
Fixes #1601 

Proposed Commit Message:
```
SimpleDateFormat, Date and Calendar are not thread-safe, resulting in
tests occasionally failing when multi-threading functionality is used.

Let's use DateTimeFormatter and the java.time package APIs since they
are thread-safe.
```


gerhean:
Tests are failing likely as timezone used is changed. It should not be using the local timezone since I have no idea what timezone github actions runs on.

Timezone logic should be based on config (#1360), else I believe timezone should be ignored.

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -208,11 +210,27 @@ public static String extractDate(String date) {
...
+                processDate[0] = '0' + processDate[0];
+            }
+            if (processDate[1].length() == 1) {
+                // Likewise for month
+                processDate[1] = '0' + processDate[1];
+            }
+            date = processDate[0] + "/" + processDate[1] + "/" + processDate[2];
+            return Date.from(ZonedDateTime.parse(date,
+                    CLI_ARGS_DATE_FORMAT.withResolverStyle(ResolverStyle.STRICT)
+                            .withZone(ZoneId.systemDefault())).toInstant());
```
yhtMinceraft1010X, line 229: I'm not sure if we should revert back to `SimpleDateTime` here. I noticed that I cannot pass in a `zoneId` from the arguments without putting the timezone together with the date.

src/main/java/reposense/system/CustomLogFormatter.java
```diff
@@ -19,7 +20,8 @@
  */
 public class CustomLogFormatter extends SimpleFormatter {
 
-    private static final DateFormat dateFormat = new SimpleDateFormat("hh:mm:ss");
+    private static final DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("hh:mm:ss")
+            .withZone(ZoneId.systemDefault());
```
yhtMinceraft1010X, line 24: I am also not sure about what to do with this one.

src/main/java/reposense/util/TimeUtil.java
gerhean: I feel this code is a little too complicated, the given java libraries really should have a way to resolve this.

gerhean:
Is it possible to not change the tests since tests do not run in parallel?

src/main/java/reposense/system/CustomLogFormatter.java
gerhean: What if you just remove the `.withZone`?

gerhean:
Actually what's the reason why `ZonedDateTime` is being used here.

dcshzj:
> Is it possible to revert the changes to tests since tests do not run in parallel?

No, we have to test in parallel as well, otherwise we would not be able to catch concurrency-related issues that occur when users use the software.

> Actually what's the reason why `ZonedDateTime` is being used here.

The timezone uses the user's local timezone on the computer, so that it is something expected when the user runs the software on their computer. GitHub Actions uses UTC+0, unless otherwise specified in a particular test.

This is not supposed to be an easy bug to fix, especially on the timezone issues. I'll do a proper review in a couple of days and provide some guidance on how to address the issue.

src/main/java/reposense/util/TimeUtil.java
gerhean: Might want to try reading [this](https://stackoverflow.com/questions/42903391/convert-simpledateformat-to-datetimeformatter)

dcshzj: For this part of the code, we only need to store the date (the time will be discarded later on). We don't have to worry about the time zone here, so using system default is fine. The time zone information will be added later in `ArgsParser.java`.

src/test/java/reposense/git/GitShowTest.java
```diff
@@ -26,7 +27,8 @@ public void getExpandedCommitHash_nonExistentCommit_throwsEmptyCommitException()
     @Test
     public void getCommitDate_normalCommit_success() throws Exception {
         Date commitDate = GitShow.getCommitDate(config.getRepoRoot(), TEST_COMMIT_HASH);
-        Date expectedDate = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss Z").parse("2018-02-09 22:17:39 +0800");
+        Date expectedDate = Date.from(ZonedDateTime.parse("2018-02-09 22:17:39 +0800",
+                DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss Z")).toInstant());
```
dcshzj, line 31: `uuuu` to `yyyy`.

src/test/java/reposense/git/GitShowTest.java
```diff
@@ -39,15 +41,17 @@ public void getParentCommits_nonExistentCommit_throwsEmptyCommitException() thro
     public void getEarliestCommitDate_singleCommit_success() throws Exception {
         Date earliestDate = GitShow.getEarliestCommitDate(
                 config.getRepoRoot(), Arrays.asList(TEST_COMMIT_HASH));
-        Date expectedDate = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss Z").parse("2018-02-09 22:17:39 +0800");
+        Date expectedDate = Date.from(ZonedDateTime.parse("2018-02-09 22:17:39 +0800",
+                DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss Z")).toInstant());
```
dcshzj, line 45: `uuuu` to `yyyy`.

src/test/java/reposense/git/GitShowTest.java
```diff
@@ -39,15 +41,17 @@ public void getParentCommits_nonExistentCommit_throwsEmptyCommitException() thro
...
         Assert.assertEquals(expectedDate, earliestDate);
     }
 
     @Test
     public void getEarliestCommitDate_multipleCommits_success() throws Exception {
         Date earliestDate = GitShow.getEarliestCommitDate(
                 config.getRepoRoot(), Arrays.asList(TEST_COMMIT_HASH, ROOT_COMMIT_HASH, LATEST_COMMIT_HASH));
-        Date expectedDate = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss Z").parse("2018-02-05 16:00:39 +0800");
+        Date expectedDate = Date.from(ZonedDateTime.parse("2018-02-05 16:00:39 +0800",
+                DateTimeFormatter.ofPattern("uuuu-MM-dd HH:mm:ss Z")).toInstant());
```
dcshzj, line 54: `uuuu` to `yyyy`.

src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -101,6 +101,7 @@ public void before() throws Exception {
         config = new RepoConfiguration(new RepoLocation(TEST_REPO_GIT_LOCATION), "master");
         config.setAuthorList(Collections.singletonList(getAlphaAllAliasAuthor()));
         config.setFormats(FileTypeTest.DEFAULT_TEST_FORMATS);
+        config.setZoneId(TIME_ZONE_ID_STRING);
```
dcshzj, line 104: I don't think this is needed. It is supposed to use the user's time zone, and the rest of the code should work for different time zones.

dcshzj (changes requested):

dcshzj:
The code that added the time zone parameter are in #1346. We use the user's computer time zone (`ZoneId.systemDefault()`) unless otherwise provided by the user. This time zone will determine the exact instant for the since and until dates (if midnight in Singapore, then will be 4pm in UTC, etc).

Git commit date and time will have the time zone attached to them. We want to convert them into an instant, so that we can compare them with the since and until date instants.

Another tip, to test the time zone stuff, you can change the time zone on your local computer and run the program again to see if it works. GitHub Actions uses UTC+0 time zone.

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -21,7 +23,7 @@
     private static Long startTime;
     private static final String DATE_FORMAT_REGEX =
             "^((0[1-9]|[12][0-9]|3[01])\\/(0[1-9]|1[012])\\/(19|2[0-9])[0-9]{2})";
-    private static final DateFormat CLI_ARGS_DATE_FORMAT = new SimpleDateFormat("dd/MM/yyyy HH:mm:ss");
+    private static final DateTimeFormatter CLI_ARGS_DATE_FORMAT = DateTimeFormatter.ofPattern("d/M/uuuu HH:mm:ss");
```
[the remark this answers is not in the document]
yhtMinceraft1010X: I can change all the other `uuuu` but changing this one back seems to break some tests.

yhtMinceraft1010X:
I'm converting this to draft since this may need more time to fix.

yhtMinceraft1010X:
I have made some huge changes to the code. Currently, `systemtest` isn't working due to some changes to how each author's contribution variance is calculated.

src/main/java/reposense/util/TimeUtil.java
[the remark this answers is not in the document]
yhtMinceraft1010X: For this one, `yyyy` doesn't work with `ResolverStyle.STRICT` as described [here](https://stackoverflow.com/questions/27454025/unable-to-obtain-localdatetime-from-temporalaccessor-when-parsing-localdatetime). 

It is possible to use `yyyy` with the default `ResolverStyle` but that would break the invalid date test in which "31 February" is used as the input date.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -140,13 +139,13 @@
...
-        Date reportSinceDate = (cliSinceDate.equals(SinceDateArgumentType.ARBITRARY_FIRST_COMMIT_DATE))
+        LocalDateTime reportSinceDate = (cliSinceDate.equals(SinceDateArgumentType.ARBITRARY_FIRST_COMMIT_DATE))
                 ? earliestSinceDate : cliSinceDate;
 
         Optional<Path> summaryPath = FileUtil.writeJsonFile(
                 new SummaryJson(configs, reportConfig, generationDate,
-                        TimeUtil.getZonedDateFromSystemDate(reportSinceDate, zoneId),
-                        TimeUtil.getZonedDateFromSystemDate(untilDate, zoneId), isSinceDateProvided,
+                        reportSinceDate,
+                        untilDate, isSinceDateProvided,
```
dcshzj, line 148: I think this should be using the time zone given in zoneId? We are generating the report based on the user's given time zone.

src/main/java/reposense/util/TimeUtil.java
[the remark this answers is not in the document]
dcshzj: I see, thanks for the linked StackOverflow answer! Let's use `uuuu` as you suggested.

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -66,118 +63,61 @@ public static String getElapsedTimeMessage() {
...
-        int zoneRawOffset = getZoneRawOffset(zoneId);
-        int systemRawOffset = getZoneRawOffset(ZoneId.systemDefault());
-        return new Date(current.getTime() + zoneRawOffset - systemRawOffset);
-    }
-
-    /**
-     * Returns a {@code Date} that is set to midnight for the given {@code zoneId}.
-     */
-    public static Date getZonedSinceDate(Date sinceDate, ZoneId zoneId) {
+    public static LocalDateTime getZonedSinceDate(LocalDateTime sinceDate) {
```
dcshzj, line 68: It is no longer timezone-aware, so the method name should be changed to avoid confusion.

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -66,118 +63,61 @@ public static String getElapsedTimeMessage() {
...
-                .Builder()
-                .setInstant(untilDate.getTime())
-                .build();
-        cal.set(Calendar.HOUR_OF_DAY, 23);
-        cal.set(Calendar.MINUTE, 59);
-        cal.set(Calendar.SECOND, 59);
-        cal.set(Calendar.MILLISECOND, 0);
-        cal.add(Calendar.MILLISECOND, systemRawOffset - zoneRawOffset);
-        return cal.getTime();
+    public static LocalDateTime getZonedUntilDate(LocalDateTime untilDate) {
```
dcshzj, line 79: It is no longer timezone-aware, so the method name should be changed to avoid confusion.

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -699,10 +698,12 @@ public void parse_withDatesAndTimezone_success() throws Exception {
...
-        int[] expectedSinceTime = {21, 0, 0};
-        Date expectedSinceDate = TestUtil.getDate(2017, Calendar.JUNE, 30, expectedSinceTime);
-        int[] expectedUntilTime = {20, 59, 59};
-        Date expectedUntilDate = TestUtil.getDate(2017, Calendar.NOVEMBER, 30, expectedUntilTime);
+        int[] expectedSinceTime = {0, 0, 0};
+        LocalDateTime expectedSinceDate = TestUtil.getDate(2017,
+                Month.JULY.getValue(), 1, expectedSinceTime);
+        int[] expectedUntilTime = {23, 59, 59};
+        LocalDateTime expectedUntilDate = TestUtil.getDate(2017,
+                Month.NOVEMBER.getValue(), 30, expectedUntilTime);
```
dcshzj, line 706: This changes the original intent of the test. Is it possible to test such that we know the `--timezone` argument is correctly parsed?

src/test/java/reposense/util/TestUtil.java
```diff
@@ -105,65 +103,42 @@ public static boolean compareDirectories(Path expected, Path actual) throws Exce
...
-        Calendar cal = Calendar.getInstance();
-        cal.set(Calendar.YEAR, year);
-        cal.set(Calendar.MONTH, month);
-        cal.set(Calendar.DAY_OF_MONTH, date);
-        cal.set(Calendar.HOUR_OF_DAY, time[0]);
-        cal.set(Calendar.MINUTE, time[1]);
-        cal.set(Calendar.SECOND, time[2]);
-        cal.set(Calendar.MILLISECOND, 0);
-        return cal.getTime();
+    public static LocalDateTime getLocalDate(int year, int month, int date, int[] time) {
```
dcshzj, line 131: This should be merged into the `getDate` function above.

src/test/java/reposense/util/TestUtil.java
```diff
@@ -105,65 +103,42 @@ public static boolean compareDirectories(Path expected, Path actual) throws Exce
...
+        return LocalDateTime.of(year, month, date, time[0], time[1], time[2], 0);
     }
 
     /**
      * Wrapper for {@code getLocalDate} method to get since date with time 00:00:00
      */
-    public static Date getLocalSinceDate(int year, int month, int date) {
+    public static LocalDateTime getLocalSinceDate(int year, int month, int date) {
```
dcshzj, line 138: This should be merged into the `getSinceDate` function above.

src/test/java/reposense/template/GitTestTemplate.java
dcshzj: I see. Let's keep just one line and we can remove line 106, which is a duplicate of this line.

dcshzj (changes requested):
Great job with the huge changes, left some comments

src/main/java/reposense/report/ReportGenerator.java
yhtMinceraft1010X: I'm not sure what you mean here since `reportSinceDate` and `untilDate` are `LocalDateTime` objects. The timezone is not fixed in the code but there is an implicit assumption that `reportSinceDate` and `untilDate` are with reference to the timezone of `zoneId`, which is also passed into the constructor.

src/test/java/reposense/parser/ArgsParserTest.java
yhtMinceraft1010X: For this one, I'm not sure about the UTC times. Example for `expectedSinceTime = {21, 0, 0}`: 

Since the time for a `since date` is set to midnight for the timezone which it is in (in this case UTC+11), the `expectedSinceTime` should be `13:00:00` in UTC.

yhtMinceraft1010X: Screenshot (67)

In the original implementation of `getZonedSinceDate`, it appears that the midnight time was subtracted by 3 hours (`systemRawOffset - zoneRawOffset`). Likewise for `getZonedUntilDate`.

src/main/java/reposense/report/ReportGenerator.java
dcshzj: The `reportSinceDate`, if it uses `cliSinceDate`, then it is correct as it is implicitly assumed as the user's input, which will be in the time zone provided. However, if the user provides `d1` as the input, what is the time zone used?

src/test/java/reposense/parser/ArgsParserTest.java
dcshzj: Yup, the subtraction was to account for the time zone, as previously the since and until dates would be saved in the given time zone. However, as now the since and until dates are stored as `LocalDateTime` with no zone information, I was wondering if we can test if the `--timezone` flag is correctly used, which was the original intent of the test.

It does not have to be right here in this test function, especially since the implementation has changed quite a bit, but we still need to test to see if the `--timezone` flag is still working.

yhtMinceraft1010X: I see. It makes sense if the original expected times were written in terms of UTC+8. 

Because I'm currently writing the expected times in terms of UTC to follow Github Actions.

src/main/java/reposense/report/ReportGenerator.java
yhtMinceraft1010X: I just realised that the timezone in each `CommitInfo` may be different from what is given in `config`. Not only could the earliest commit date be in the wrong timezone if `d1` were used, the `LocalDateTime` stored in each `CommitResult` would also be off.

Thanks for pointing it out!

src/main/java/reposense/git/GitClone.java
```diff
@@ -207,7 +207,7 @@ private static String getCloneBareCommand(RepoConfiguration config, String outpu
      * into the folder {@code outputFolderName}.
      */
     private static String getCloneShallowBareCommand(RepoConfiguration config,
-                                                    String outputFolderName, Date shallowSinceDate) {
```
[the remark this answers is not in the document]
yhtMinceraft1010X: Done

src/test/java/reposense/template/GitTestTemplate.java
yhtMinceraft1010X: Done

src/main/java/reposense/report/ReportGenerator.java
yhtMinceraft1010X: I'm changing some lines in `CommitInfoAnalyzer` and `CommitInfoAnalyzerTest` to account for the difference in timezone.

dcshzj: Can I confirm if the date obtained from `d1` is generated based on the user's given time zone given in `config.zoneId` (basically does it now correctly use the user's time zone)?

dcshzj (changes requested):
2 more issues left, the issue of `d1` and test cases for checking if time zone is respected by RepoSense.

src/test/java/reposense/commits/CommitInfoAnalyzerTest.java
```diff
@@ -267,6 +269,32 @@ public void analyzeCommits_emptyCommits_success() throws Exception {
         Assert.assertEquals(expectedCommitResults, actualCommitResults);
     }
 
+    @Test
+    public void analyzeCommits_emptyCommitsWithDifferentTimeZone_success() throws Exception {
```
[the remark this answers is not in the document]
yhtMinceraft1010X: I have added 2 new test cases for the timezone: one with a timezone that would result in the commit date being one day ahead in the config timezone, another with a timezone for the commit date being one day behind in config timezone.

src/main/java/reposense/report/ReportGenerator.java
yhtMinceraft1010X: I also noticed that for `commitInfos.isEmpty()`, `d1` would result in `getStartDate` of `CommitResultAggregator` returning a date in UTC time rather than the config timezone. I've changed it to return the `ARBITRARY_FIRST_COMMIT_DATE` in the config timezone instead.

Other than that, when `d1` is used, the date of the earliest commit date would be used for `reportSinceDate` when creating the `SummaryJson` object in `ReportGenerator`. The commit date is translated into the config timezone in `CommitInfoAnalyzer`

src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -98,19 +103,19 @@ private static float getContributionVariance(List<AuthorDailyContribution> contr
...
+            commitStartDate = getStartOfDate(commitResult.getTime(), zoneId);
             Author commitAuthor = commitResult.getAuthor();
 
             List<AuthorDailyContribution> authorDailyContributions = authorDailyContributionsMap.get(commitAuthor);
 
-            if (authorDailyContributions.isEmpty()
-                    || !authorDailyContributions.get(authorDailyContributions.size() - 1).getDate()
+            if (authorDailyContributions.isEmpty() || !getStartOfDate(authorDailyContributions
+                    .get(authorDailyContributions.size() - 1).getDate(), zoneId)
                             .equals(commitStartDate)) {
```
dcshzj, line 119: Let's create a new boolean variable that describes the second part of this condition, because I think this is too long and not easily readable.

src/main/java/reposense/report/ReportGenerator.java
dcshzj: Okay nice. These two lines can be merged into a single line.

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -88,7 +88,7 @@ public void extractFileInfos_branchWithValidWhitelistedFileName_success() {
 
     @Test
     public void extractFileInfos_sinceDateAfterLatestCommit_emptyResult() {
-        Date date = TestUtil.getSinceDate(2050, 12, 31);
+        LocalDateTime date = TestUtil.getSinceDate(2050, 12, 31);
```
dcshzj, line 91: Change `12` to `Month.DECEMBER.getValue()`.

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -97,7 +97,7 @@ public void extractFileInfos_sinceDateAfterLatestCommit_emptyResult() {
 
     @Test
     public void extractFileInfos_untilDateBeforeFirstCommit_emptyResult() {
-        Date date = TestUtil.getUntilDate(2015, 12, 31);
+        LocalDateTime date = TestUtil.getUntilDate(2015, 12, 31);
```
dcshzj, line 100: Change `12` to `Month.DECEMBER.getValue()`.

dcshzj (changes requested):

src/main/java/reposense/commits/CommitResultAggregator.java
yhtMinceraft1010X: This is what I currently have for the refactoring:
Screenshot (82)

Personally, I think that splitting the `addDailyContributionForNewDate` into nested if-else blocks isn't ideal. I can't put the boolean in front since there would be an invalid `authorDailyContributions.get(-1)` for empty `authorDailyContributions` before the `isEmpty()` check is done.

What I suggest instead is to put comments for the boolean check, like:
```
// Checks whether there are no contribution dates present 
// or the most recently inserted contribution date does not match commit date
```

dcshzj (approved):
LGTM! Great job on this PR!

dcshzj:
In the interest of moving this PR along, I will merge this directly. @fzdy1914 Do let me know if you have any additional comments on this PR.

PR #564, opened 2019-02-19
GitLsTree: validate filepaths before cloning #391

yong24s:
Fixes #391 

```
Unix and Windows use a different set of invalid characters, with the
latter being more restrictive.

A valid file path in Unix may be invalid in Windows, which causes git
clone to fail. We classify this error generally as GitCloneException,
which may not be useful in conveying the exact cause of error.

Let's validate file paths and use a more specialized exception to log
invalid file paths, when analysis is done on Windows.
```

PR #1628, opened 2022-01-23
[#1616] Fix issue generating repos with special chars on Windows

### LINKED ISSUE #1616: Unable to generate report on Windows for repos containing files with special characters

1. Run the command `java -jar RepoSense.jar --repos https://github.com/AY2122S1-CS2113T-W12-2/tp.git --since 12/10/2021 --until 12/10/2021 --view`

Repository analyzed displaying visible commits. 
image
(This was analyzed on Ubuntu 18.04.)

image

```Jan 15, 2022 3:22:20 PM reposense.report.ReportGenerator analyzeRepos
INFO: [1/1] Analyzing https://github.com/AY2122S1-CS2113T-W12-2/tp.git (master)...
Jan 15, 2022 3:22:20 PM reposense.git.GitLsTree isValidWindowsFilename
SEVERE: Invalid filepath: 'text-ui-test/EXPECTED-UNIX.TXT\357\200\215' contains '\'
Jan 15, 2022 3:22:21 PM reposense.report.ReportGenerator generateReposReport
INFO: The report is generated at C:\Users\Chan Jun Da\Desktop\testing\reposense-report
Jan 15, 2022 3:22:21 PM reposense.RepoSense main
INFO: Elapsed processing time: 2.68 second(s)
```

Relevant lines of the log file. It encounters an error parsing one specific file and skips the repository entirely.

The repo being analyzed has a file containing special characters which is not allowed for Windows but valid for Linux and Mac. I believe that the following lines of code 
https://github.com/reposense/RepoSense/blob/1fcd96e3d523cc0562b1ce8eec8ef8ca981b5e24/src/main/java/reposense/git/GitLsTree.java#L50-L60
are the reason for this issue. The `ReportGenerator::analyzeRepos` method uses the above method `GitLsTree::validateFilePaths` to check all the file paths at once. When a single file path is invalid due to containing special characters, the entire repository is skipped.

chan-j-d:
Fixes #1616 

At the moment, a blanket check on all filenames in the repository is performed to see if they are valid file paths with the `GitLsTree::validateFilePaths` method in `ReportGenerator::analyzeRepos`  before analyzing it. Relevant lines of code here: https://github.com/reposense/RepoSense/blob/065a7fa51d501414bb997ee84178816fd33ab7af/src/main/java/reposense/report/ReportGenerator.java#L321-L327. 

This causes RepoSense run from Windows OS to skip the repository entirely even though there are no issues with running RepoSense against the repository itself. The invalid file is automatically skipped with a warning similar to this 
image

I believe this implementation on Windows will agree with a similar run on Linux as RepoSense is already unable to properly attribute the authorship for these files with special characters.

Commit message:
```
Running RepoSense on Windows on repositories containing filenames with
special characters will be skipped entirely. Instead, it should continue
analyzing the repository and skip the problem files.

Let's prevent RepoSense on Windows skipping repositories with filenames
containing special characters.
```


src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -320,7 +318,6 @@ private static AnalyzeJobOutput analyzeRepos(String outputPath, List<RepoConfigu
                             + MESSAGE_START_ANALYSIS, configToAnalyze.getLocation(), configToAnalyze.getBranch()));
             try {
                 GitRevParse.assertBranchExists(configToAnalyze, FileUtil.getBareRepoPath(configToAnalyze));
-                GitLsTree.validateFilePaths(configToAnalyze, FileUtil.getBareRepoPath(configToAnalyze));
```
dcshzj, line 323: This code is specifically run for Windows, I don't think it should be removed?

dcshzj (changes requested):
I'm quite confused, do you mind explaining again what you are trying to achieve in this PR?

src/main/java/reposense/report/ReportGenerator.java
chan-j-d: At the moment, running RepoSense on Windows against a repo with a single file that contains special characters will cause the entire repo to be rejected by this line (#1616) as it first views all filenames in the repo and throws an error. 

I think a better approach will be to skip the file which is what will be done after this line of code is removed as the rest of the repository is analyzed as per normal. Unless the intention is to disallow RepoSense to work on any such repositories. I think the fix is relatively simple as it can still be cloned but the specific file cannot be worked on directly.

dcshzj: This was introduced in #564 but I could not find a reason why it was decided to just skip the repository entirely. @fzdy1914 Do you have any opinion about this?

fzdy1914: @chan-j-d Your suggestion makes sense. Just wondering, will the analysis of other files be normal given that this file is wrong?

chan-j-d: @fzdy1914 the other files are not affected. In general, files with filenames containing special characters are not processed properly for authorship and skipped regardless of OS and I am still investigating. So if someone narrows the time window to just a single commit and compares the number of lines to the authorship lines, there will be a mismatch. But with regards to the analysis for the remainder files, they are still accurate at least from my experience. 

For the run of RepoSense highlighted in #1616 that highlights this problem, I ran the same command with the master branch RepoSense on WSL and compared the report generated to one from this pull request branch ran on Windows, The summary and authorship json files are exactly the same (tested with `cmp`) so it is at least consistent.

chan-j-d: Hi @fzdy1914, @dcshzj, would like to ask if there is any further action suggested for this PR?

dcshzj: @chan-j-d I think there are no other comments, feel free to just skip the file and maybe add a warning in the report to show the user that the file was skipped due to invalid characters.

chan-j-d: Thanks for the update @dcshzj. Would like to clarify what you mean by a warning in the report. The current analysis already handles it appropriately and outputs a warning as per the image in the issue for Windows systems. Or are you referring to a warning on the report when opened in a web browser similar to that of when a repo fails?

dcshzj: > The current analysis already handles it appropriately and outputs a warning as per the image in the issue for Windows systems.

Yup, I was referring to this. If it is already handled then it's all good.

dcshzj (approved):
LGTM!

fzdy1914 (approved):

Issue #1622, opened 2022-01-19
Upgrade highlight.js to version 11


dcshzj:
The highlight.js package used by the frontend is currently using version 10.7.3, which is [no longer supported](https://github.com/highlightjs/highlight.js/blob/main/SECURITY.md). The package should be upgraded to version 11 by following [the upgrading guide](https://github.com/highlightjs/highlight.js/blob/main/VERSION_11_UPGRADE.md).

Zhou-Jiahao-1998:
@dcshzj I am not too sure if I am doing it correctly. After upgrading to version 11, everything seems to work fine. All the test passed and there are no errors in the project. Is that normal?

gerhean:
Usually nothing breaks when package is upgraded to be honest.

gerhean:
Want to close this in favor of #1633.

dcshzj:
> Want to close this in favor of #1633.

Yup, let's do that.

PR #1629, opened 2022-01-24
[#1622] Upgrade highlight.js to version 11

Zhou-Jiahao-1998:
Fixes https://github.com/reposense/RepoSense/issues/1622

Proposed commit message:

```
The highlight.js package used by the frontend is currently
version 10.7.3, which is no longer supported. 

Upgraded highlight.js to 11.4.0 for better compatibility.
```

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -31,7 +31,7 @@
     "@fortawesome/free-solid-svg-icons": "^5.15.2",
     "@fortawesome/vue-fontawesome": "^2.0.2",
     "core-js": "^3.6.5",
-    "highlight.js": "^10.5.0",
+    "highlight.js": "^11.4.0",
     "jszip": "^3.5.0",
     "minimatch": "^3.0.4",
     "pug-lint-vue": "^0.4.0",


gerhean (approved):
Code-wise LGTM, can you improve your [commit message](http://se-education.org/guides/conventions/github.html) a bit?

Zhou-Jiahao-1998:
@gerhean Alright noted! Is this better now?

gerhean:
> @gerhean Alright noted! Is this better now? On a side note, is it possible to test on your system to double check as I am on MacOS.

Computer OS should have no impact on frontend changes, since it is the browser which runs the JavaScript code.

That being said, I do not see anything wrong when running on Chrome.

gerhean (approved):
LGTM

dcshzj (changes requested):
We still have a reference to an older version of highlight.js here:

https://github.com/reposense/RepoSense/blob/c4d7690e7b2c3b2626187a2c26ee447d99adf15a/frontend/src/app.vue#L321

Do we need to upgrade this as well?

Also, looking at the dashboard deployment for this PR, it doesn't seem like any of the code is highlighted.

gerhean:
> We still have a reference to an older version of highlight.js here:
> 
> https://github.com/reposense/RepoSense/blob/c4d7690e7b2c3b2626187a2c26ee447d99adf15a/frontend/src/app.vue#L321
> 
> Do we need to upgrade this as well?
> 
> Also, looking at the dashboard deployment for this PR, it doesn't seem like any of the code is highlighted.

Thanks for catching this, by right nothing should be imported `cdn.jsdelivr.net`, and should instead be installed and imported from `node_modules` folder. (This should be fixed in another PR).

In fact, I just realized that highlightjs is broken in the [netlify deployment](https://reposense.netlify.app/) as well, seems like a regression of #1412 which nobody caught unfortunately.

Perhaps this PR should be closed, and a new PR to be opened to fix the regression.

Zhou-Jiahao-1998:
@gerhean In that case, I will close the PR.

PR #1631, opened 2022-01-25
[#1630] Rearrange Gradle tasks and their dependencies

### LINKED ISSUE #1630: Remove building of frontend when running backend tests

Currently, when running `gradlew test` or `gradlew test systemtest`, the frontend is built even though running the backend tests do not appear to require frontend in any way. 

I propose restructuring the dependencies in `build.gradle` such the frontend is not built when running `test` or `systemtest` alone. (I.e. not part of build)

yhtMinceraft1010X:
Fixes #1630 

Proposed commit message:
```
When running pure backend tests alone using gradlew test systemtest, the
frontend is built before these tests. This is due to zipReport task
being needed for compileJava. zipReport in turn relies on buildFrontend.

The frontend is not used for these tests. It is wasteful to run 
buildFrontend when only the backend tests are being run.

In addition, some other tasks and properties are scattered throughout
the build.gradle file or have dependencies that could be pushed forward.
This complicates maintenance.

Let's remove the `test` and `systemtest` dependency on the frontend as
well as re-order and consolidate some task properties and dependencies.
This should make running either of them faster.
```


gerhean:
Is it possible to just remove `zipReport` since it is just zipping an empty folder if build report is not run?

dcshzj:
> Actually the entire process is a little weird and should probably be overhauled. Maybe introduce a build everything command, and maybe add checks that the required components exist for each command else build them.

I think let's discuss this in #1630 to see what solutions we should adopt. Let's not rush into implementation yet. @gerhean Do you mind copying your comment over and to continue the discussion on #1630?

yhtMinceraft1010X:
Now that #1691 has been merged, I'm re-opening this pull request.

The `zipReport` task is still needed for putting the frontend assets in `src/main/resources`. This is especially important for generating a jar file for releases such that the frontend assets don't need to be downloaded separately.

However, pure backend tests and systemtests do not need any of the frontend assets.

build.gradle
```diff
@@ -132,11 +137,7 @@ task zipReport(dependsOn: 'buildFrontend', type: Zip) {
     destinationDir = file('src/main/resources')
 }
 
-tasks.shadowJar.dependsOn('zipReport');
-tasks.compileJava.dependsOn('zipReport');
-tasks.run.dependsOn('compileJava');
-
-task systemtest(dependsOn: 'zipReport', type: Test) {
+task systemtest(type: Test) {
```
gok99, line 140: This might be a separate discussion but our systemtests currently are more of an integrated pure backend test. In addition to bypassing CLI arguments (as is an issue raised in #1645), we also don't seem to have any tests for the integration of front and backend - specifically the correct replication of `template.zip` files.

Should we also be including doing frontend-backend integration tests in systemtests as well?

build.gradle
```diff
@@ -155,7 +156,8 @@ task systemtest(dependsOn: 'zipReport', type: Test) {
     }
 }
 
-task startServerInBackground(dependsOn: 'classes', type: com.github.psxpaul.task.JavaExecFork) {
+task startServerInBackground(dependsOn: ['zipReport', 'compileJava', 'classes'], type: com.github.psxpaul.task.JavaExecFork) {
```
gok99, line 159: The name of this gradle task was actually a little misleading to me before. It seems like a general server start command for any report (that was already generated). It seems however that it's only really used for the Cypress tests.

Would something like `serveTestReportInBackground` be a more accurate description?

.github/workflows/integration.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       run: ./gradlew lintFrontend
 
     - name: Build with Gradle
-      run: time ./gradlew clean checkstyleMain checkstyleTest test systemTest coverage
+      run: time ./gradlew clean checkstyleMain checkstyleTest zipReport test systemTest coverage
```
gok99, line 70: Would not including `zipReport` actually affect anything?

build.gradle
chan-j-d: > specifically the correct replication of `template.zip` files.

I think this can be settled by changing `ConfigSystemTest` to compare the `index.html` file since right now it only compares the `.json` files. Although this would require someone to 'validate' the html file each time something is changed to changes the expected html

.github/workflows/integration.yml
yhtMinceraft1010X: I put this here because I thought `Build preview website` may need it. Initially, zipReport would have been done as compileJava was entirely dependent on it. But in this case, with the general dependency removed, none of the tasks here would have called zipReport.

build.gradle
yhtMinceraft1010X: For any changes to `ConfigSystemTest`, I think they are best done in a separate PR. However, if `index.html` is going to be compared, the dependency on `zipReport` should be added back in that PR

.github/workflows/integration.yml
gok99: Yea, it seems like a separate gradle run is done for `Build preview website`. Perhaps @dcshzj could confirm that there is no dependency here between the steps.

build.gradle
gok99: > I think they are best done in a separate PR

Yes, I agree. I think it's also okay to leave it as-is and add back the dependency if we decide to check frontend build files.

gerhean: I just want to say that I prefer for `zipReport` to be removed completely since it is basically a convoluted way of copying files around, instead of just copying it directly to the output folder.

yhtMinceraft1010X: From what I know, `zipReport` would still be needed to ensure that there is a `templateZip.zip` file in `src/main/resources` so that all the frontend assets in the zip file can be packaged in the jar file when running `shadowJar`.

Once a release jar file is downloaded by someone, the backend code is needed to copy the frontend assets inside to the output directory. This enables viewing the report in browser. Gradle is only relevant for users who fork the entire project.

yhtMinceraft1010X:
Closing and reopening due to `Expected - Waiting for status to be reported`.

yhtMinceraft1010X:
My original intention was to remove the need to run `zipReport` for pure backend tests but I also realised that there were also some other areas where refactoring could be done, such as putting `zipReport` as a dependency for `serveTestReportInBackground` since it is needed for the 2 cypress-related tasks anyway.

Another point of note is that for `build.gradle`, some tasks can only be referenced if declared before hand. I moved the Jacoco tasks to the back when consolidating the `executionData` since the build would fail as `systemtest` and `frontendTest` were not yet defined when `jacocoTestReport` was in front of both tests.

yhtMinceraft1010X:
> Correct me if my understanding is wrong, but most of the line changes here are reordering to group related tasks together? The key line changes here is removing the need to run `zipReport` in `systemTest` and `compileJava` which now allows us to run tests and system tests without building frontend.

Now that you've mentioned it, I've edited the commit message and PR title to reflect the new changes.

chan-j-d:
I see! thanks for the explanation.

dcshzj (approved):
LGTM! Looks like the CI tests run a little faster now too

PR #1639, opened 2022-01-30
[#1638] Replace parseInt with Math.floor for contribution bars

### LINKED ISSUE #1638: Wrongly scaled contribution bar when contribution is significantly lower than average

* **RepoSense Version:** c4d7690
* **OS and Version:** Fedora 34

1. Generate reposense report for https://github.com/cheeaun/sgbusdata.git since 01/01/2021

Correct report with appropriately scaled contribution bars

Committer with only a 1 line diff has the longest contribution bar when the average contribution is 1547905 lines.

image

https://sneaky-toes.surge.sh

gok99:
Fixes #1638 

**Proposed commit message:**
```
To calculate the number of contribution bars to display in 
getContributionBars, a parseInt of the ratio of a user's contribution
to the contribution limit is used. However, if the contribution is
much lower than average, the ratio is represented in scientific
notation and parseInt does not return the integer part of the number
(parseInt of 3.2301723943006836e-7 would return 3).

Let's use Math.floor to get the integer part of the contribution ratio 
instead of parseInt in getContributionBars, as is currently done in 
getFileTypeContributionBars.
```
Before:

image

After:

image

### DIFF
--- a/frontend/src/components/v-summary-charts.vue
+++ b/frontend/src/components/v-summary-charts.vue
@@ -288,7 +288,7 @@ export default {
       const res = [];
       const contributionLimit = (this.avgContributionSize * 2);
 
-      const cnt = parseInt(totalContribution / contributionLimit, 10);
+      const cnt = Math.floor(totalContribution / contributionLimit);
       for (let cntId = 0; cntId < cnt; cntId += 1) {
         res.push(100);
       }


gerhean:
Seems fine, is there any way to check the before and after? Like a screenshot for example?

gerhean (approved):
LGTM

HCY123902 (approved):
LGTM

PR #1642, opened 2022-02-01
Add wildcard type parameter

yhtMinceraft1010X:
Proposed Commit Message:
```
IntelliJ throws a warning that loadResource parameter classForLoading
uses raw type for Class. This warning is similar to
javac -Xlint:rawtypes.

Let's add a wildcard type parameter to Class to remove that warning.
```

Screenshot:
Screenshot (62)


dcshzj (approved):
LGTM!

Issue #1640, opened 2022-01-30
Unable to open github commits on a cloned repository


gok99:
**Tell us about your environment**

* **RepoSense Version:** c4d7690
* **OS and Version:** Fedora 34
* **Web Browser and Version (if applicable):**

**Please include the steps to reproduce the bug.**

1. Clone https://github.com/cheeaun/sgbusdata.git 
1. Generate a reposense report since 01/01/2021
1. Open the report and click on a commit ramp

**What was expected to happen?**

Opens the correct commit on the remote repository (github).

**What actually happened? Please include a screenshot of the output.**

Opens https://github.com/undefined/sgbusdata/commit/HASH (where HASH is the correct commit hash) which of course shows a 404. This could be especially confusing if a user does not read the URL, as they might think the commit no longer exists.

**If possible, include the URL to your RepoSense report or log files (if any).**

gok99:
I am able to access github commits using gitlens on vscode, so it seems remote user information might be accessible somehow.

gok99:
I noticed that there is already an open issue from 2018 #231 about this same issue. So I'll close this one.

Issue #713, opened 2019-05-28
No commits found for authors with name as display name


jamessspanggg:
When clicking on ramp slices that are from authors with their name as display name rather than the GitHub user name, it leads to no commits found for this author.

For example, for the author Eugene Peh, all ramp slices link to: No commits found for "Eugene Peh".
<img width="821" alt="Screen Shot 2019-05-28 at 12 07 13 PM" src="https://user-images.githubusercontent.com/32864116/58450653-6abb5f80-8142-11e9-9986-4765771905b8.png">
<img width="263" alt="Screen Shot 2019-05-28 at 12 14 22 PM" src="https://user-images.githubusercontent.com/32864116/58450610-39429400-8142-11e9-9d88-dffa75f51b64.png">

Furthermore, it is shown in the image above that the author Eugene Peh has a commit "Restore functional test". But in GitHub it shows that it was the author with name eugenepeh that committed instead.
<img width="432" alt="Screen Shot 2019-05-28 at 12 10 24 PM" src="https://user-images.githubusercontent.com/32864116/58450565-06989b80-8142-11e9-80dd-21433ee96d97.png">

[Link to the report](https://brave-mahavira-2133df.netlify.com/#search=eug&sort=searchPath&sortWithin=name&since=2017-05-28&until=2019-05-16&timeframe=day&reverse=false&groupSelect=groupByRepos&breakdown=false)

fzdy1914:
@jamessspanggg 
For further bugs in a self-generated report, please state the command you use to get the report so we are able to reproduce it when reviewing it.

fzdy1914:
The error comes from [here](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/git/GitShortlog.java#L34), we use `git author name` to be the `GitHub id` of `Author`.

PR #1644, opened 2022-02-02
[#1572] Accept different types of URLs besides https github

### LINKED ISSUE #1572: Remote repos: allow different types of URLs besides the `http`s `github` type

Allow for Reposense to accept Git URLs other than the `https` `github` one. Idea originates from the following comment https://github.com/reposense/RepoSense/issues/1546#issuecomment-838572658.

N.A.

We can remove the location verification performed. There are also several places where a regex match with the `https` `github` url is used to identify repo name and organisation which would also need to be updated. In particular, the differentiation of local and remote repo is done with the above regex matching. Instead, GIT URLs mostly have a standard form as described in https://git-scm.com/docs/git-clone#_git_urls which would allow us to differentiate them. 
> This syntax is only recognized if there are no slashes before the first colon. This helps differentiate a local path that contains a colon.

A sample [PR](https://github.com/chan-j-d/RepoSense/pull/3) with some code changes to demonstrate the feasibility of such a change. Passes all test cases except those directly tied to the existing location verification (which are removed in the PR).

**Additional context**

### LINKED ISSUE #1637: Multiple local paths with the same final directory causes one report to overwrite others

1. Run `git clone https://github.com/chan-j-d/ip.git parent1/same-name`
2. Run `git clone https://github.com/chan-j-d/tp.git parent2/same-name`
1. Run `java -jar RepoSense.jar --repos parent1/same-name parent2/same-name --since 01/06/2020 --view`

Both repositories are analyzed and displayed.

image

The repo name is defaulted to the parent directory name of the `.git` folder which both have the same name. As such, the reports are both generated in the same folder `same-name_master`. It could be that one report overwrites the other or the existing report prevents another report form being written. Regardless, only one report is generated. 

### LINKED ISSUE #1641: `~` in config file used for relative file pathing does not work for generating reports

1.  Run `git clone https://github.com/reposense/RepoSense ~/test` on a Linux terminal to clone a repo (any repo will do) into a relative directory.
1. Run `java -jar RepoSense.jar --repos ~/test --view` - this should work without issue
2. Add the following config file to the same directory or create one with location:`~/test` and branch: `master`
[repo-config.csv](https://github.com/reposense/RepoSense/files/7968135/repo-config.csv)
3. Run `java -jar RepoSense.jar --view`

Repository analyzed and report generated without issue for both cases.

For the run of step 3, an error occurs. Specific messages in the command line:
```
Chan Jun Da@LAPTOP-B8NFSECE MINGW64 ~/Desktop/testing
$ java -jar RepoSense.jar --view
12:42:16 - Parsed header of CSV file, repo-config.csv, and recognized columns: Repository's Location,  Branch,  Shallow Cloning,  File formats,  Ignore Glob List,  Ignore Authors List,  Find Previous Authors,  Ignore standalone config,  Ignore Commits List
12:42:16 - [WARNING] Error parsing line 2 in CSV file, repo-config.csv.
Content: ~/test,master,,,,,,,
Error: ~/test is an invalid location.
12:42:16 - [WARNING] No valid configurations in the repo-config.csv.
```
[log.txt](https://github.com/reposense/RepoSense/files/7968192/log.txt)

chan-j-d:
Resolves #1572 

Commit message:
```
The repo location verification performed in RepoLocation is too 
restrictive and rejects many otherwise valid git clone targets. 
External links in the generated webpage also hardcode parts of the
link to be GitHub specific.

This change would allow RepoSense to work with a greater variety of 
locations both remote and local while still providing an appropriate
guess for the organization and repo-name. 

The way external links are generated in the frontend has also been 
changed to support GitHub, GitLab and BitBucket. Support has also 
been implemented to easily include other domain names if needed.

Unsupported repository links (local or unsupported remote) will no
longer open an external webpage when clicked.

Let's allow RepoSense to work with a greater variety of locations.

```

_Backend Changes_

This change removes direct validation of location based on the URL provided as was done in the previous implementation. This might indirectly fix the issue #1641. This implementation now provides a best guess of the repo name and 'organization' via regex match based on https://git-scm.com/docs/git-clone#_git_urls. 

The method of guessing for remote repos is such: Given a URL `https://host.xz[:port]/path/to/repo.git/`, it drops the domain and keeps the `path/to/repo.git`. The last directory becomes the `repo-name` and the `path/to` has the `/` changed to a `-` and becomes the `organization`. 

Local repositories are sorted similarly. This allows differentiation between two local directories with the same final directory although this can result in some awkward organization names. E.g. the path `../reposense/RepoSense` will have `organization` `..-reposense`. The benefit is that two different paths provided will always provide two unique output directories. This should also fix the issue #1637 but I will investigate it separately. 

The class `SupportedDomainUrlMap` has also been added to be the 'single source of truth' for which types of remote repository domains we would like to support. Adding a new domain can be easily done by adding a new entry to the map. The map provides template mappings for how things like the file blame URL and file history URL would look like. The information is uploaded via `summary.json`. 

Related PRs (that are possibly under review but already merged to this PR):
- [x] https://github.com/reposense/RepoSense/pull/1662
- [x] https://github.com/reposense/RepoSense/pull/1683
- [x] https://github.com/reposense/RepoSense/pull/1685

_Frontend Changes_
The template links are uploaded via `summary.json` and stored in `window.DOMAIN_URL_MAP`. 

The way the various types of links are generated (file history, author profile, repo url) have been aggregated in `api.js`. All other files that generated their own links have been edited to use one of the methods from `api.js`. 

Currently, the buttons that leads to once-invalid links will not open anything. This is an improvement over the master branch behaviour of opening an invalid link. A follow-up change (not included in this PR) would be to add a message notification if trying to open an unsupported link. 

_Reviewing this PR_
The backend changes relating to the organization and repo-name guessing in `RepoLocation` (up to commit https://github.com/reposense/RepoSense/pull/1644/commits/60f57e34180784f8e1b43a2cd0afe14840deba7e) have mostly been approved. 

I would like reviews regarding the changes after which are dealing with supporting the frontend link generations. 
In particular, I would like the frontend code reviewed as I'm not too familiar with the layers of abstraction and design in the frontend code. 

_How to test_
I used the two repositories below for testing.
GitLab repository: https://gitlab.com/tortoisegit/tortoisegit.git
BitBucket repository: https://bitbucket.org/atlassian/atlassian-scheduler.git
Pull the branch https://github.com/chan-j-d/RepoSense/tree/1572-experimental-implementation, build and run something like
`gradlew run -Dargs="--repos https://gitlab.com/tortoisegit/tortoisegit.git https://bitbucket.org/atlassian/atlassian-scheduler.git https://github.com/reposense/RepoSense -i --view -s 10/02/2022"`


src/main/java/reposense/model/RepoLocation.java
```diff
@@ -1,47 +1,53 @@
...
-    private String organization;
+    private final String organization;
 
     /**
      * @throws InvalidLocationException if {@code location} cannot be represented by a {@code URL} or {@code Path}.
      */
     public RepoLocation(String location) throws InvalidLocationException {
-        if (SystemUtil.isWindows()) {
+        boolean isWindows = SystemUtil.isWindows();
+        if (isWindows) {
```
dcshzj, line 34: I think we can just directly use it without creating a new variable.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -1,47 +1,53 @@
...
+            repoNameAndOrg = new String[] {"", ""};
+        } else if (isLocalRepo(location)) {
+            repoNameAndOrg = getLocalRepoNameAndOrg(location);
         } else {
-            repoName = Paths.get(location).getFileName().toString().replace(GIT_LINK_SUFFIX, "");
+            repoNameAndOrg = getRemoteRepoNameAndOrg(location);
         }
+
+        repoName = repoNameAndOrg[0];
+        organization = repoNameAndOrg[1];
```
dcshzj, line 50: Should add a `this` to `repoName` and `organization`.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -57,29 +63,60 @@ public String getOrganization() {
...
-        boolean isValidGitUrl = false;
-
-        try {
-            Path pathLocation = Paths.get(location);
-            isValidPathLocation = Files.exists(pathLocation);
-        } catch (InvalidPathException ipe) {
-            // Ignore exception
+    public static boolean isLocalRepo(String repoArgument) {
+        boolean containsColon = repoArgument.contains(":");
+        if (!containsColon) {
```
dcshzj, line 72: I think we can just directly use it without creating a new variable.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -57,29 +63,60 @@ public String getOrganization() {
...
+        return new String[] {tempRepoName, tempOrganization};
     }
 
+    private String[] getRemoteRepoNameAndOrg(String location) throws InvalidLocationException {
+        Matcher remoteRepoMatcher = GIT_REPOSITORY_LOCATION_PATTERN.matcher(location);
+        Matcher sshRepoMatcher = SCP_LIKE_SSH_REPOSITORY_LOCATION_PATTERN.matcher(location);
+
+        boolean isValidRemoteRepoUrl = remoteRepoMatcher.matches() || sshRepoMatcher.matches();
+        if (!isValidRemoteRepoUrl) {
+            throw new InvalidLocationException(String.format(MESSAGE_INVALID_LOCATION, location));
```
dcshzj, line 111: I think we should be able to add a unit test for this.

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -1,49 +1,116 @@
...
+        assertValidLocation("https://github.com/reposense/RepoSense",
+                "RepoSense", "reposense");
+
+        // Test against other types of URL protocols that are valid for git clone
+        assertValidLocation("ssh://[user@]host.xz[:port]/path/to/repo.git/",
+                EXPECTED_REPO_NAME, EXPECTED_ORGANIZATION);
+        assertValidLocation("git://host.xz[:port]/path/to/repo.git",
+                EXPECTED_REPO_NAME, EXPECTED_ORGANIZATION);
+        assertValidLocation("https://host.xz[:port]/path/to/repo.git",
+                EXPECTED_REPO_NAME, EXPECTED_ORGANIZATION);
```
dcshzj, line 105: Are we able to change the `[user@]` and `[:port]` to a concrete example?

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -1,49 +1,116 @@
...
+
+        // Test against the conventional ssh protocol used for GitHub, e.g. git@github.com:reposense/RepoSense.git
+        assertValidLocation("[user@]host.xz:path/to/repo.git/",
+                EXPECTED_REPO_NAME, EXPECTED_ORGANIZATION);
```
dcshzj, line 109: Same for here as well, to change the `[user@]` part.

src/systemtest/resources/30daysFromUntilDate/expected/summary.json
```diff
@@ -1 +1 @@
...
...ster]","outputFolderName":"reposense_testrepo-Alpha_master"},{"location":{"location":"https://github.com/reposense/testrepo-Beta.git","repoName":"testrepo-Beta","organization":"reposense"},"branch":"master","displayName":"reposense/testrepo-Beta[master]","outputFolderName":"reposense_testrepo-Beta_master"},{"location":{"location":"https://github.com/reposense/testrepo-Charlie.git","repoName":"testrepo-Charlie","organization":"reposense"},"branch":"master","displayName":"reposense/testrepo-Charlie[master]","outputFolderName":"reposense_testrepo-Charlie_master"},{"location":{"location":"https://github.com/reposense/testrepo-Delta.git","repoName":"testrepo-Delta","organization":"reposense"},"branch":"master","displayName":"reposense/testrepo-Delta[master]","outputFolderName":"reposense_testrepo-Delta_master"}],"errorSet":[{"repoName":"--github.com-reposense/testrepo-Beta[HEAD]","errorMessage":"Failed to clone from ttps://github.com/reposense/testrepo-Beta.git"},{"repoName":"reposense/testrepo-Delta[nonExistentBranch]","errorMessage":"Branch \"nonExistentBranch\" does not exist."}],"sinceDate":"2017-10-01","untilDate":"2017-11-01","isSinceDateProvided":false,"isUntilDateProvided":true}
```
dcshzj, line 1: I think the error message is not correct. We should retain the fact that `ttps://github.com/reposense/testrepo-Beta.git` is an invalid URL (because Git only supports ssh, git, http, https, ftp and ftps protocols). Giving the message that it failed to clone is not specific enough, and the user will need to take an extra step to figure out why.

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -1,49 +1,116 @@
...
+        assertValidLocation(LOCAL_REPO_LOCATION_WINDOWS_VALID_WITH_DOT_GIT_ONE,
+                EXPECTED_REPO_NAME, EXPECTED_ORGANIZATION);
+
+        assertValidLocation(LOCAL_REPO_LOCATION_WINDOWS_VALID_MIXED,
+                EXPECTED_REPO_NAME, "..-" + EXPECTED_ORGANIZATION);
+    }
+
+    @Test
+    public void repoLocation_parseValidRemoteRepoUrl_success() throws Exception {
+        // valid url without specifying branch
```
dcshzj, line 74: Let's add BitBucket as well to support the top 3 websites, the rest I think we can just ignore

dcshzj (changes requested):
I think we should still come up with a list of invalid URLs and make sure they are correctly reported as invalid.

src/main/java/reposense/model/RepoLocation.java
chan-j-d: Thanks for the comment. However, after quite a bit of testing with the regex on my end, I realise that it is very difficult to construct an example of an invalid remote repo location that does not match either the https/ssh/git URLs and the scp-like protocol. This was added in case such a string does exist. More details on it in the comment replying to `ttps://...` invalid URL error message.

src/systemtest/resources/30daysFromUntilDate/expected/summary.json
chan-j-d: I believe the behaviour here is that this URL is instead detected as part of the scp-like protocol for cloning (similar to the `git@github.com:...`. In this case it is the host name `ttps` with path `//github.com/reposense/testrepo-Beta.git` which is a valid pathname as multiple slashes are just treated as a single slash.

src/test/java/reposense/model/RepoLocationTest.java
chan-j-d: The idea I had in mind was a kind of white-box testing where the domain is dropped regardless. But noted, I will try to change the examples here. Thanks!

src/test/java/reposense/model/RepoLocationTest.java
chan-j-d: Noted. Do we have a bitbucket reposense remote repo? Else I will add in a random bitbucket repository.

src/main/java/reposense/model/RepoLocation.java
chan-j-d: On some further consideration, I believe the regex pattern matching can be stricter for URL matches following this [stackoverflow post](https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid/1547940#1547940). So if it matches the format but contains illegal characters then it will throw an exception. Thanks!

chan-j-d:
Hi @dcshzj, requesting a follow-up review.

For the code coverage, I don't think it can go any higher. There is a test case for the Windows file path replacement on line 89 but it is only ran on Windows systems. And for the invalid local path location, I cannot think of an example that fails because of how creative valid Unix filenames can get.

src/test/java/reposense/model/RepoLocationTest.java
dcshzj: Nope, we do not. I think the URL you used is fine.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -16,14 +14,21 @@
  * Represents a repository location.
  */
 public class RepoLocation {
-    private static final String GIT_LINK_SUFFIX = ".git";
     private static final String MESSAGE_INVALID_LOCATION = "%s is an invalid location.";
+    private static final String MESSAGE_INVALID_REMOTE_URL = "%s is an invalid remote URL.";
+
     private static final Pattern GIT_REPOSITORY_LOCATION_PATTERN =
-            Pattern.compile("^https?://github.com/(?<org>.+?)/(?<repoName>.+?)\\.git$");
+            Pattern.compile("^(ssh|git|https?|ftps?)://[^/]*?/(?<path>.*?)/?(?<repoName>[^/]+?)(/?\\.git)?/?$");
```
dcshzj, line 21: Are we also able to add support for `file://` protocol? Git treats it as a local path when cloning.

dcshzj (changes requested):

src/main/java/reposense/model/RepoLocation.java
chan-j-d: Hi thanks for reminding me about this! I initially supported it then removed it because it caused quite a bit of issue being a URL and being local. However, I think the current implementation should support it easily. I will update it when I get back. Thanks!

chan-j-d: @dcshzj I've updated the implementation to support `file://` as well now. Would like to request a follow-up review. Thanks!

chan-j-d:
I just realised there is a serious problem with the logic for local absolute path on windows that starts with `C:\path\to\repo` as it is detected as a remote repo and some problems occur.

chan-j-d:
Hi @dcshzj, apologies about the delay. I have fixed the issue with the local disk drives and added a fix for an unexpected possible NullPointerException. Would like to request a follow-up review for this.

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -1,49 +1,149 @@
...
 import reposense.util.AssertUtil;
+import reposense.util.SystemUtil;
 
 public class RepoLocationTest {
+
+    private static final String LOCAL_REPO_LOCATION_VALID_WITHOUT_DOT_GIT_ONE = "repo";
+    private static final String LOCAL_REPO_LOCATION_VALID_WITHOUT_DOT_GIT_TWO = "../path/to/repo";
+    private static final String LOCAL_REPO_LOCATION_VALID_WITHOUT_DOT_GIT_THREE = "file://path/to/repo";
+    private static final String LOCAL_REPO_LOCATION_VALID_WITH_DOT_GIT_ONE = "path/to/repo/.git";
+    private static final String LOCAL_REPO_LOCATION_VALID_WITH_DOT_GIT_TWO = "file://path/to/repo.git";
```
dcshzj, line 19: I don't think this is a valid path. Using [RFC 8089](https://datatracker.ietf.org/doc/html/rfc8089) under Appendix B, there should be 3 slashes after `file:` (i.e. `file:///path/to/repo.git`). If there are only 2, then the hostname immediately follows (i.e. `file://localhost/path/to/repo.git`). This is also the same for Windows.

Is it also possible to add a test that uses a relative path? Something like `file:///path/another/../to/repo.git` or `https://github.com/reposense/wrong/path/../../RepoSense.git`. I'm not sure how the parsing will be affected, but it should still have the same expected repo and organization name.

dcshzj (changes requested):

src/test/java/reposense/model/RepoLocationTest.java
chan-j-d: > I don't think this is a valid path.

Admittedly, I'm not too well-versed in this.
image but these two both seem to work and that was why I tested it as such. My interpretation of it was that `/path/to/repo/` was a generic path and `file://` was prepended to it. 

However, 1 slash don't seem to work as shown below:
image

even though they should as seen from page 17 of the same [document](https://datatracker.ietf.org/doc/html/rfc8089#page-18), which seems to suggest that the single slash is valid. So maybe git clone does not strictly follow the convention?

chan-j-d: > Is it also possible to add a test that uses a relative path?

Noted on this. I did not think about such a scenario. Just to confirm, the goal is to ensure that whatever output folder is generated lines up with the generated name so there are no name mismatch after analysis. Because this `https://github.com/reposense/wrong/path/../../RepoSense.git` will produce an 'organization' like `reposense-wrong-path-..-..` even though it's basically the same as `reposense`. 

I'm not too sure if manually collapsing it would be easy. Would like to clarify which situation you are referring to.

dcshzj: > So maybe git clone does not strictly follow the convention?

I think probably we might be spending too much effort into a feature that might not be too commonly used. Probably let's just make sure `file://` and `file:///` works, and let GitClone throw an exception if any edge cases were to appear.

> It seems one way to do it might be to wrap the given `filepath` for organization into a Path object, use `normalize()`

Yup, I think this would be ideal? It's probably very rare someone does this but since we have `normalize()` I think it should be worth to use it.

chan-j-d: > Probably let's just make sure `file://` and `file:///` works, and let GitClone throw an exception if any edge cases were to appear.

I agree. I think it is sufficient to provide a working case for the above scenarios and for everything else, a 'good enough' guess would be sufficient if it the location provided actually works with git clone but is not one of the cases we specified.  I have chosen to use `file:/*` instead which always strips all `/` in front of `file:` and treats everything after as the path to make handling easier since none of this actually changes the argument used for the actual `git clone` that we do. 

> Yup, I think this would be ideal? It's probably very rare someone does this but since we have normalize() I think it should be worth to use it.

I have implemented it in the update and added test cases for it.

@dcshzj I have updated the PR and would like another review. Thanks!

dcshzj (approved):
LGTM! I think we should have achieved a "should work for most scenarios"  standard. Any additional nit will probably not be very beneficial, as the use cases will be extremely rare.

gok99:
I noticed the frontend also hard-codes a remote for generating links for viewing commits or the blame view. 

https://github.com/reposense/RepoSense/blob/1e4fa415c17eaeb9178728e07e39438e8555386f/frontend/src/utils/api.js#L4

https://github.com/reposense/RepoSense/blob/1e4fa415c17eaeb9178728e07e39438e8555386f/frontend/src/utils/api.js#L123-L127

So, clicking on a commit ramp or blame view in the code panel won't work correctly when the remote isn't GitHub. A solution will likely involve the backend extracting and storing the remote base url in `summary.json` that frontend can use.

To make things difficult, GitHub, BitBucket and GitLab all follow slightly different conventions for commit and blame view URLs.

| |`base_url`|Commit View|Blame View|
|:-:|:-:|:-:|:-:|
|GitHub|github.com|`{base_url}/{org}/{repo_name}/commit/{commit_hash}`|`{base_url}/{org}/{repo_name}/blame/{branch}/{file_path}`|
|GitLab|gitlab.com|`{base_url}/{org}/{repo_name}/-/commit/{commit_hash}`|`{base_url}/{org}/{repo_name}/-/blame/{branch}/{file_path}`|
|BitBucket|bitbucket.org|`{base_url}/{org}/{repo_name}/commits/{commit_hash}`|`{base_url}/{org}/{repo_name}/annotate/{branch}/{file_path}`|

chan-j-d:
> I noticed the frontend also hard-codes a remote for generating links for viewing commits or the blame view.

Wow! Thanks for pointing this out, must have been quite a bit of research. Admittedly, I'm not too familiar with the frontend so I didn't consider this at all. This does feel like it'll open a lot of new problems. I'm not too sure if there's any good solution to support the links.

When I initially suggested the improvement, I did not really think about opening the links in the report. To me, it was an overview of the activity in the repository. Additionally, without exiting the webpage, we can already view the authored lines. 

I think there are two 'quick'-fixes to approach this (please suggest other ideas if possible):
1. Hardcode the domains of the top few remote repository hosts and add the domain name to summary.json as you suggested. For other repo types (more obscure hosts or local repos), add some kind of warning (similar to when repos fail to clone) to state that the links might not work.
2. We go straight to supporting only GitHub and for other repo types, add some kind of warning to state that the links might not work.

In some sense, local repos already face the same problem as in #231 and in the issue that you raised in #1640. In terms of a long term solution, I'm not entirely sure if it is possible for local repos.

> To make things difficult, GitHub, Bitbucket and GitLab all follow slightly different conventions for commit and blame view URLs.

On the bright side at least user profiles have the same URL format.

gok99:
Option (1) seems like a more complete generalization. I think this could be done in a separate PR since everything that used to work in the frontend will continue to work after this PR. I really can't think of a way around hard-coding the special case for Bitbucket :cry:. For more obscure remotes, or for repos with no remotes, I think it makes sense to issue a warning and get rid of frontend links to the remotes.

> In some sense, local repos already face the same problem as in https://github.com/reposense/RepoSense/issues/231 and in the issue that you raised in https://github.com/reposense/RepoSense/issues/1640. In terms of a long term solution, I'm not entirely sure if it is possible for local repos.

If local repositories are cloned or have remote information, we can get this information with `git remote -v`. I've got a rough implementation to do this, and for disabling frontend links when no remote information is available (have opened a draft for it [#1679] that I'll need to update after this PR is merged).

Also, I'm having some trouble analyzing a local repo with this branch and I can't quite figure out what's going wrong -
Error: `SEVERE: Exception met while cloning or checking out -home-gok99-Local-NUS/RepoSense[1572-temp].Analysis terminated.`
Error log: [fedora-output.txt](https://github.com/reposense/RepoSense/files/8104973/fedora-output.txt)

dcshzj:
> I noticed the frontend also hard-codes a remote for generating links for viewing commits or the blame view.

I agree here, this is a really great observation that I missed out on myself. Excellent work!

I would prefer to go with option (1) as well. For cases where the blame/history views are not possible (i.e. local repos or some non-standard Git hosting website), then we can just disable the links and add a warning. Same goes for the author profile links.

chan-j-d:
>  I've got a rough implementation to do this

Interesting. Though I think you might be held up for quite awhile.

> Also, I'm having some trouble analyzing a local repo with this branch and I can't quite figure out what's going wrong

@gok99 Are you running reposense repo with gradlew against its current directory? I couldn't get it to work when I did it with the directory './' but the actual Java error after some investigation was something about permission creating a directory so I chalked it up to some behaviour relating to gradle. 

Though I think this does bring up a very big issue that the log file is really not helpful in this instance. I think I should create an issue about it.

gok99:
> Are you running reposense repo with gradlew against its current directory

Yes, I am. I ran the following command:

```
./gradlew run --args="--since 01/01/2017 --repos /home/gok99/test-repo --view"
```
which works on the master branch. I think I see why this causes issues. In `RepoLocation.java`, we replace all `/` with `-`, so the folder name becomes `-home-gok99-test-repo...`. Unfortunately, Unix names that start with `-` (even in quotes) in a command gets [mistaken for a flag](https://serverfault.com/questions/124659/how-can-i-open-a-file-whose-name-starts-with) (I assume macs work the same way).

Since the output of `pwd` generally starts from root (`/`) it might be likely that a path starting with `/` is provided. The solution here seems to me to be to remove `/` if it's the first character before replacing it with the path separator replacement.

chan-j-d:
Interesting. In trying to replicate your error on Windows, I got a completely different error which does not bode well.
```
java.lang.RuntimeException: reposense.system.CommandRunnerProcessException: Error returned from command git clone "repos\C--Users-Chan Jun Da-Documents-NUS-NUS Year 2 Summer-cp3108a_RepoSense\RepoSense_bare" --branch 1572-experimental-implementation "repos\C--Users-Chan Jun Da-Documents-NUS-NUS Year 2 Summer-cp3108a_RepoSense\RepoSense"on path . :
Cloning into 'repos\C--Users-Chan Jun Da-Documents-NUS-NUS Year 2 Summer-cp3108a_RepoSense\RepoSense'...
done.
warning: unable to access 'src/systemtest/resources/30daysFromUntilDate/expected/reposense_testrepo-Charlie_master/.gitattributes': Filename too long
error: unable to create file src/systemtest/resources/30daysFromUntilDate/expected/reposense_testrepo-Charlie_master/authorship.json: Filename too long
fatal: cannot create directory at 'src/systemtest/resources/30daysFromUntilDateFindPreviousAuthors/expected/reposense_testrepo-Alpha_master': Filename too long
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```
The source of this error seems to be that the clone is not 'bare' and somewhere down the analysis a checkout is performed which causes the 'filename too long' issue. 

@gok99 Thanks for the diagnosis. I think the one you raised should be an easier fix. I didn't know about the flag thing that you mentioned but it shouldn't be too difficult to strip the leading slash.

chan-j-d:
Apologies @gok99, but could you try the same command but run against the test version of the PR located in the branch here:
https://github.com/chan-j-d/RepoSense/tree/test-branch-with-1619-1657-1681-1682-merged

image

With this version, there should be no change in the logic for Linux (only one change was made between the current PR and this test branch which has to do with Windows disk drive). The rest of the changes are merges with other branches that are still in PR. And running it now does not seem to result in the same issue as what you get.

If I am not wrong, the issue you highlighted does not really occur because most paths specified are `repos/-some-repo/` as the repos are stored inside the `repos` directory during runtime.

gok99:
Yes this fixes it!

> most paths specified are `repos/-some-repo/` as the repos are stored inside the `repos` directory during runtime

Yea, removing the relativize seems to have fixed the problem. The repo name in the report however still has a leading `-`:

image

which might look a little strange. Do you think this are worth removing?

chan-j-d:
I think there is no issue with that. Because `/home/gok99/test-repo` is an absolute path starting from your home. But `home/gok99/test-repo` is a relative path from the existing directory which will become home-gok99-test-repo. I think for this simple case, the intention was to retain the information about the 'slash' which results in two very different repository locations.

chan-j-d:
@dcshzj I'm not sure how to proceed with this PR. I am still investigating the error found in https://github.com/reposense/RepoSense/pull/1644#issuecomment-1046658863. But besides that, I am not sure if I am supposed to include [the frontend changes](https://github.com/reposense/RepoSense/pull/1644#issuecomment-1046402661) in this PR as well. 

This PR has grown to rely on a lot of other parts of the codebase causing unexpected errors like the one gok99 brought up that can be fixed by merging other unrelated and ongoing PRs. I'm not too sure if I want to include even more changes to other parts of the codebase within this PR. I'm also not sure if I should merge the other 4 PRs mentioned above with the assumption that those will be merged before this.

dcshzj:
@chan-j-d I think let's implement the frontend changes as well, but we can keep this PR as ongoing. Feel free to create smaller PRs (like the 4 you mentioned) that builds towards resolving this issue, and we can merge this one in the future once all other changes that are needed have been completed.

chan-j-d:
Requesting a follow-up review from @reposense/active-developer to review the new changes to generating links in the frontend. As it is quite a large PR, the original post has been edited and updated on what I need reviewed. The commit diff is linked there for ease of viewing though you'll probably still need to review it on the original `files changed` which includes many other files.

frontend/src/utils/api.js
```diff
@@ -120,10 +120,42 @@ window.toggleNext = function toggleNext(ele) {
   parent.className = classes.join(' ');
 };
 
-window.getBaseLink = function getBaseLink(repoId) {
-  return `${window.BASE_URL}/${
-    window.REPOS[repoId].location.organization}/${
-    window.REPOS[repoId].location.repoName}`;
+window.getAuthorLink = function getAuthorLink(repoId, author) {
```
yhtMinceraft1010X, line 123: Nice use of abstraction

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -1,59 +1,182 @@
...
-        assertInvalidLocation("https://github.com/reposense");
-        // url without a .git suffix should be rejected
-        assertInvalidLocation("https://github.com/reposense/RepoSensegit");
-        assertInvalidLocation("https://github.com/reposense/RepoSense-git");
-        assertInvalidLocation("https://github.com/reposense/RepoSense.gi");
-        assertInvalidLocation("https://github.com/reposense/RepoSense");
+        // valid url from other domains
+        assertValidLocation("https://bitbucket.org/reposense/RepoSense.git",
+                "RepoSense", "reposense", "bitbucket");
+        assertValidLocation("https://opensource.ncsa.illinois.edu/bitbucket/scm/u3d/3dutilities.git",
```
yhtMinceraft1010X, line 120: I think this one should be labelled separately since the url is unexpected

src/test/java/reposense/model/RepoLocationTest.java
```diff
@@ -1,59 +1,182 @@
...
+    public void repoLocation_parseInvalidRemoteRepo_throwsInvalidLocationException() throws Exception {
+        // Invalid URL protocol
+        assertInvalidLocation("ttp://github.com/reposense.RepoSense.git");
+        assertInvalidLocation("not-valid-protocol://abc.com/reposense/RepoSense.git");
+        // URL contains illegal characters
+        assertInvalidLocation("https://github.com/contains-illegal-chars/^\\/");
     }
 
     /**
      * Compares the information parsed by the RepoLocation model with the expected information
```
yhtMinceraft1010X, line 172: I think `assertValidLocation` and `assertInvalidLocation` may need to be renamed because some unrecognised domains are considered "valid" although another developer may think that these unrecognised domains should be considered invalid instead.

yhtMinceraft1010X (changes requested):
Since we are also supporting `bitbucket` and `gitlab`, we should also mention it in either the doc homepage or the User Guide. `configFiles.md repo-config.csv section` and `cli.md --repos flag` should also include the new domains.

Zhou-Jiahao-1998 (approved):
LGTM. I am only looking at the frontend changes. The changes looks logical. The abstraction is good and easy for future changes and maintenance.

chan-j-d:
@yhtMinceraft1010X requesting follow-up review as I've updated the docs and changed the parts in your comments.

docs/ug/configFiles.md
```diff
@@ -59,6 +59,17 @@ When using [standalone config](#config-json-standalone-config-file) (if it is no
 
 Optionally, you can use an `author-config.csv` (which should be in the same directory as the `repo-config.csv` file) to provide more details about the authors to analyze ([example](author-config.csv)). It should contain the following columns:
 
+<<<<<<< HEAD
```
yhtMinceraft1010X, line 62: The merge conflict is still here

yhtMinceraft1010X (changes requested):
After looking through the backend code again, I noticed that some files in `model` still refer to GitHub (such as `Author#STANDARD_GITHUB_EMAIL_DOMAIN`, `RepoLocation#GIT_REPOSITORY_LOCATION_PATTERN` and `StandaloneAuthor#githubId`)

docs/ug/configFiles.md
chan-j-d: Well that's embarrassing. I could have sworn I added the changes before committing. 

> After looking through the backend code again, I noticed that some files in model still refer to GitHub (such as Author#STANDARD_GITHUB_EMAIL_DOMAIN, RepoLocation#GIT_REPOSITORY_LOCATION_PATTERN and StandaloneAuthor#githubId)

Regarding this, I do not think it is wise to do the renaming of variables here because there are in fact a lot of specific references to GitHub both in variable/method names, string constants and DG. It would add a sizeable amount of lines changed to this PR without really affecting any functionality. As such, I edited only the parts I felt were most immediately relevant parts in the UG. 

@yhtMinceraft1010X requesting follow-up review

yhtMinceraft1010X (approved):
LGTM

src/systemtest/resources/30daysFromUntilDate/expected/summary.json
```diff
@@ -1 +1 @@
...
...ub.com/reposense/testrepo-Beta.git is an invalid remote URL."},{"repoName":"reposense/testrepo-Delta[nonExistentBranch]","errorMessage":"Branch \"nonExistentBranch\" does not exist."}],"sinceDate":"2017-10-01","untilDate":"2017-11-01","isSinceDateProvided":false,"isUntilDateProvided":true,"supportedDomainUrlMap":{"github":{"BRANCH":"tree/BRANCH","REPO_URL":"https://github.com/ORGANIZATION/REPO_NAME/","BASE_URL":"https://github.com/","HISTORY_PATH":"commits/BRANCH/FILE_PATH","COMMIT_PATH":"commit/COMMIT_HASH","BLAME_PATH":"blame/BRANCH/FILE_PATH"},"NOT_RECOGNISED":{"BRANCH":"","REPO_URL":"UNSUPPORTED","BASE_URL":"UNSUPPORTED","HISTORY_PATH":"","COMMIT_PATH":"","BLAME_PATH":""},"gitlab":{"BRANCH":"-/tree/BRANCH","REPO_URL":"https://gitlab.com/ORGANIZATION/REPO_NAME/","BASE_URL":"https://gitlab.com/","HISTORY_PATH":"-/commits/BRANCH/FILE_PATH","COMMIT_PATH":"-/commit/COMMIT_HASH","BLAME_PATH":"-/blame/BRANCH/FILE_PATH"},"bitbucket":{"BRANCH":"src/BRANCH","REPO_URL":"https://bitbucket.org/ORGANIZATION/REPO_NAME/","BASE_URL":"https://bitbucket.org/","HISTORY_PATH":"history-node/HEAD/FILE_PATH?at\u003dBRANCH","COMMIT_PATH":"commits/COMMIT_HASH","BLAME_PATH":"annotate/BRANCH/FILE_PATH"}}}
```
gok99, line 1: This is a minor point, but in view of bloat in `summary.json`, do you think it would be a good idea to move supportedDomainUrlMap to a separate file, since it is a general mapping and not related to specific summary details for the repositories in question.

gok99 (commented):
Did some general testing on random GitLab and Bitbucket repositories:

Things worked nicely for https://USER@bitbucket.org/angelsl/ms-renx.git, except for some committers who didn't have bitbucket accounts who had broken profile links (this is mirror of a github repo). Not sure if it's even possible to check if accounts exists and hide links, but that would be nice to have.

The Gitlab repo I tested https://gitlab.com/asus-linux/asusctl-gex.git had broken profile links for every user. I'm not entirely sure why it was unable to find the correct usernames. All other links seem to work fine.

Also, I just realized bitbucket allows users to use custom domains like: https://opensource.ncsa.illinois.edu/bitbucket, which makes full bitbucket support total hell (not sure how common this is). I think we can just pretend that this is not a thing, although perhaps a note about it in the UG could be helpful. From what I can tell, the behaviour is that we can analyze such reports just fine, but the clicking on links just don't take you anywhere.

chan-j-d:
> The Gitlab repo I tested https://gitlab.com/asus-linux/asusctl-gex.git had broken profile links for every user. I'm not entirely sure why it was unable to find the correct usernames. All other links seem to work fine.

I believe this is related to https://github.com/reposense/RepoSense/issues/713 and https://github.com/reposense/RepoSense/issues/1185. Even on GitHub, our links don't work if a `config` file containing the author's github id is not provided. I believe the conclusion was that there is just not enough information provided by git to guess the exact BitBucket id in this case. 

In a run against https://gitlab.com/asus-linux/asusctl-gex.git, there appears to be commits attributed to `Marco Laux` and `marco` but the actual author's bitbucket id is `marcaux`. I'm not sure if this can ever be fixed. It should be sufficient to link to the correctly formatted URL which I think it does. 

> which makes full bitbucket support total hell

I believe bitbucket is actually a remote repository hosting tool more than it is the actual host. They just so happen to have an actual remote repo host at `bitbucket.org`. Anyway, I don't think we need to fully support all the clickable links. The original goal of this was actually to allow `RepoSense` to analyze any repository it is capable of cloning and not arbitrarily restrict it to GitHub. I think it is sufficient to support just the main domains while others can still analyze their remote repositories but are unable to go to the actual links. 

Perhaps down the line, if there is really a need, we can add implement a way for users to add a json file specifying the same mappings we have in `supportedDomainUrlMap` and read the json file to add it to the map of supported links. 

>  I think we can just pretend that this is not a thing, although perhaps a note about it in the UG could be helpful.
I did add a comment about this in the UG in `generatingReports.md` although I left it somewhat vague. I wasn't sure exactly where to place it or how much should be specified. Suggestions are welcome though!

src/systemtest/resources/30daysFromUntilDate/expected/summary.json
chan-j-d: Hmm, this is a good point. My concern here was actually if I should be adding a new file at all 😅 since I don't think many people have actually added an entirely new file to every generated report. My thought process here was just to have a proof of concept that we can selectively allow domains to support links with the information in a json file. Moving it to another json file shouldn't be too difficult.

damithc:
> In a run against https://gitlab.com/asus-linux/asusctl-gex.git, there appears to be commits attributed to `Marco Laux` and `marco` but the actual author's bitbucket id is `marcaux`. I'm not sure if this can ever be fixed. It should be sufficient to link to the correctly formatted URL which I think it does.

BTW, our author-config file specifically mentions `Author's GitHub ID`,  right?

chan-j-d:
@damithc Oops, seems like I missed that out! Yes, the term in the UG is the same as the column header in the author config file itself. Though now that I've thought about it for awhile, I think this might break all previous author config files if I were just to change the header like this.

gok99:
> BTW, our author-config file specifically mentions Author's GitHub ID, right?

I completely forgot about this. The Author class has some GitHub specific names and constants as well.

src/systemtest/resources/30daysFromUntilDate/expected/summary.json
gok99: > add a json file specifying the same mappings we have in supportedDomainUrlMap and read the json file to add it to the map of supported links

Having it as a separate file would likely make it easier for users to write their own mappings as well if they wanted.

chan-j-d: I was thinking more along the lines of adding it to `report-config.json` for example, rather than having the user manually edit a file in the final report.

gok99: Yea I figured :laughing:, if there is some gap before we implement that though and users **really** wanted support, this will probably make it slightly easier to do.

docs/ug/configFiles.md
```diff
@@ -63,7 +63,7 @@ Optionally, you can use an `author-config.csv` (which should be in the same dire
 |-------------|-------------|
 | Repository's Location | Same as `repo-config.csv`. Default: all the repos in `repo-config.csv` |
 | Branch | The branch to analyze for this author, e.g., `master`. Default: the author will be bound to all the repos in `repo-config.csv` that has the same repo's location, regardless of branch. |
-| Author's Git Host ID {{ mandatory }} | Username of the target author's profile on GitHub, GitLab or Bitbucket, e.g.`JohnDoe`. |  
+| Author's GitHub ID {{ mandatory }} | Username of the target author's profile on GitHubHub, GitHubLab or Bitbucket, e.g.`JohnDoe`. |
```
yhtMinceraft1010X, line 66: The `Hub` is repeated

yhtMinceraft1010X (changes requested):

docs/ug/configFiles.md
chan-j-d: Oh crap, good spot. I was trying to use line replacement to avoid intellij auto expanding the tables

frontend/src/components/v-summary-charts.vue
```diff
@@ -45,7 +45,7 @@
           span.tooltip-text Click to view group's repo
```
dcshzj, line 45: The icon for this still shows a GitHub logo, might have to find another icon that is more suitable?

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -16,14 +19,35 @@
  * Represents a repository location.
  */
 public class RepoLocation {
-    private static final String GIT_LINK_SUFFIX = ".git";
+
+    protected static final String UNSUPPORTED_DOMAIN_NAME = "NOT_RECOGNISED";
```
dcshzj, line 23:
```suggestion
    protected static final String UNSUPPORTED_DOMAIN_NAME = "NOT_RECOGNIZED";
```

src/main/java/reposense/model/SupportedDomainUrlMap.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final SupportedDomainUrlMap DEFAULT_DOMAIN_URL_MAP = new SupportedDomainUrlMap();
+
+    private final Map<String, Map<String, String>> domainUrlMap;
+
+    private SupportedDomainUrlMap() {
+        domainUrlMap = new HashMap<>();
+        domainUrlMap.put("github", GITHUB_MAP);
+        domainUrlMap.put("gitlab", GITLAB_MAP);
+        domainUrlMap.put("bitbucket", BITBUCKET_MAP);
+        domainUrlMap.put(RepoLocation.UNSUPPORTED_DOMAIN_NAME, NOT_SUPPORTED_MAP);
```
dcshzj, line 74: Do you think it is possible if we only include the mappings if that particular host is used? For instance, if all my repos are on GitHub, the mappings provided to the frontend only contains `GITHUB_MAP`.

dcshzj (changes requested):

docs/ug/configFiles.md
```diff
@@ -63,9 +63,9 @@ Optionally, you can use an `author-config.csv` (which should be in the same dire
 |-------------|-------------|
 | Repository's Location | Same as `repo-config.csv`. Default: all the repos in `repo-config.csv` |
 | Branch | The branch to analyze for this author, e.g., `master`. Default: the author will be bound to all the repos in `repo-config.csv` that has the same repo's location, regardless of branch. |
-| Author's GitHub ID {{ mandatory }} | GitHub username of the target author, e.g., `JohnDoe` |
-| Author's Emails<sup>*</sup> | Associated Github emails of the author. This can be found in your [GitHub settings](https://github.com/settings/emails). |
-| Author's Display Name | The name to display for the author. Default: author's GitHub username. |
+| Author's GitHub ID {{ mandatory }} | Username of the target author's profile on GitHub, GitLab or Bitbucket, e.g.`JohnDoe`. |
```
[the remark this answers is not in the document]
chan-j-d: Regarding this, I think it's related to this https://github.com/reposense/RepoSense/pull/1644#issuecomment-1073298074 where I think there's are quite a lot more considerations regarding changing the column name header than just changing it in the UG. 

These values as well as the ones in the `config.json` component of the UG are directly tied to the backend parser class that reads the csv/json files and specifically look for this column header, it's quite hard to change it without considering backward compatibility of older config files. 

I discussed this in an earlier 3281 meeting and I intend to create a follow-up PR to change these column references together with the backend code to maintain some kind of backward compatibility. I was mainly thinking of explicitly accepting `Author's Git Host ID` and `Author's GitHub ID`. However, I think there is more to discuss on the issue that I did not want to directly include in this PR.

docs/ug/generatingReports.md
```diff
@@ -18,6 +18,12 @@ Let's look at different ways to generate RepoSense reports.
 
 * **If you don't have Java on your computer or do not wish to run the executable on your computer**, some alternatives are provided in the [_Generating reports remotely_](#generating-reports-remotely) section below.
 
+<box type="info" seamless>
+
+RepoSense is built to analyze any type of git repo, remote or local. It works best when analyzing remote repositories hosted on GitHub, GitLab or BitBucket.
+For other types of repositories, features such as external links might not work properly.
```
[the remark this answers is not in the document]
chan-j-d: Currently, the links for local repositories and non GitHub, GitLab and BitBucket repos do not open at all which I think was the desired outcome we agreed on. Regarding the wording used in the UG here, it was wrote in this way in view of https://github.com/reposense/RepoSense/pull/1679 where local repositories might be able to support links. This might also change if we include a way for users to specify their own remote repository URL templates. 

However, I am fine to change it to `For other types of repositories, external links are disabled.` as these other features might not be coming so soon, if ever.

docs/ug/usingReports.md
```diff
@@ -39,6 +38,11 @@ Here is an example of how a typical report looks like:
 
 The report is divided into two sections: [_**Chart panel**_](#chart-panel) and the [_**Code panel**_](#code-panel). In some situations, the [_**Commits panel**_](#commits-panel) will appear in place of the _code panel_. All three are explained in the sections below.
 
+<box type="info" seamless>
+
+The links to external webpages is only properly supported when RepoSense analyzes a remote repository belonging to GitHub, GitLab or BitBucket.
```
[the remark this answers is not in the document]
chan-j-d: good point, will add this

frontend/src/utils/api.js
```diff
@@ -120,10 +121,52 @@ window.toggleNext = function toggleNext(ele) {
...
+  const domainName = window.REPOS[repoId].location.domainName;
+  return window.filterUnsupported(window.DOMAIN_URL_MAP[domainName].BASE_URL + author);
+};
+
+window.getRepoLinkUnfiltered = function getRepoLink(repoId) {
+  // abstraction for repo link construction. Not supposed to be used by other files
+  const domainName = window.REPOS[repoId].location.domainName;
+  return window.DOMAIN_URL_MAP[domainName].REPO_URL
+      .replace('ORGANIZATION', window.REPOS[repoId].location.organization)
+      .replace('REPO_NAME', window.REPOS[repoId].location.repoName);
```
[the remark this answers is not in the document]
chan-j-d: good idea, will add this

frontend/src/utils/api.js
```diff
@@ -1,8 +1,9 @@
 // utility functions //
 window.$ = (id) => document.getElementById(id);
 window.enquery = (key, val) => `${key}=${encodeURIComponent(val)}`;
-window.BASE_URL = 'https://github.com';
+window.REPOSENSE_REPO_URL = 'https://github.com/reposense/RepoSense';
 window.HOME_PAGE_URL = 'https://reposense.org';
+window.UNSUPPORTED_INDICATOR = 'UNSUPPORTED';
```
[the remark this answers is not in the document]
chan-j-d: The links are only generated in the frontend so that each repository only needs reponame, domain and org, together with the URL templates. This constant was just a way for the backend to signal to the frontend that this specific link is not supported, in which case it returns an `undefined`. The `undefined` is then set to the `href` which causes the link to be disabled. I'm not sure if there's a better way to go about this though suggestions are greatly appreciated.

src/main/java/reposense/model/SupportedDomainUrlMap.java
chan-j-d: Oh this is a great point. Yes, it's probably possible to implement. Will add this.

frontend/src/components/v-summary-charts.vue
chan-j-d: 😅 I can't think of anything too good. I think the `git` logo might be fitting for this though I'm unclear on anything related to licensing on this matter.

chan-j-d:
@dcshzj Requesting follow-up review. Thanks!

docs/ug/configFiles.md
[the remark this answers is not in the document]
dcshzj: > Regarding this, I think it's related to this [#1644 (comment)](https://github.com/reposense/RepoSense/pull/1644#issuecomment-1073298074) where I think there's are quite a lot more considerations regarding changing the column name header than just changing it in the UG.

My apologies, I missed out that comment. Thanks for the clarification!

> I discussed this in an earlier 3281 meeting and I intend to create a follow-up PR to change these column references together with the backend code to maintain some kind of backward compatibility. I was mainly thinking of explicitly accepting `Author's Git Host ID` and `Author's GitHub ID`. However, I think there is more to discuss on the issue that I did not want to directly include in this PR.

I think this is a good idea. This PR may be getting a little too big, and trying to maintain compatibility with previous author configs may be difficult to implement. Please do follow up on another PR, thanks!

docs/ug/generatingReports.md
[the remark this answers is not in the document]
dcshzj: Yup I think the wording is better now. #1679 seems to be more aligned with what I was thinking about, so I think we can leave this change to that PR.

frontend/src/utils/api.js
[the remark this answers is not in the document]
dcshzj: I see, I think I got confused about this part. Seems like the code you have already achieves the desired result.

frontend/src/components/v-summary-charts.vue
dcshzj: Hmm the git logo seems weird here, what about `fa-file-code` or `fa-file-lines`? Gitlab seems to use the file-lines icon:

Screenshot 2022-03-25 at 18 31 16

dcshzj (changes requested):
Replied to some of the comments, mainly now is about the icon used for the repository.

frontend/src/components/v-summary-charts.vue
chan-j-d: Between the two, I prefer file-code. I'll put forward this suggestion though 
image

Since I think it's a pretty common image for remote repositories, especially in images illustrating the forking workflow

chan-j-d: I've updated it to `fa-file-code`, do take a look. Thanks!

dcshzj: @damithc Prof, do you have any thoughts about the icon to use to represent the link to the actual repository? Previously we were using the GitHub logo and it looks like this:
Current

We are currently evaluating the following options to replace this logo so that it is more general and not GitHub-specific:
Database
File code
File lines

damithc: For generic icon, perhaps the `database` one? Also, fontawsome has this icon for git which may be suitable.
image

It would be cool if we can use respective github, gitlab, and bitbucket icons (fontawesome has all three) when we know the repo is one of them, but that can be a separate enhancement.

chan-j-d: I've chosen to settle on the `database` icon for now. I've tried the `git-alt` icon and it appears really small on the actual report with the branch lines being barely visible. 

This suggestion about the respective icons is really interesting! I think I'll create a new issue for it as I'm not too familiar with how to implement that right now.

dcshzj (approved):
LGTM!

chan-j-d:
Follow-up actions are being tracked in this issue https://github.com/reposense/RepoSense/issues/1726

PR #1647, opened 2022-02-06
[#1504] Add file size limit to prevent large reports crashing

### LINKED ISSUE #1504: Fail gracefully for large commits
In case a commit is huge (e.g., someone committed a huge library by accident), ensure RepoSense fails gracefully. Might need to check the following areas:
* during analysis
* chart view
* code view

### LINKED ISSUE #1716: File-size limit for backend processing

Part of #1504. 

The PR #1647 currently processes all files, and ignores sufficiently large files (in file size) that could potentially crash the frontend. It might be a good idea to introduce a similar limit for files that the backend processes as well (i.e. if a file exceeds a limit, the backend will skip the file instead of analyzing it), to prevent excessively long report generation times.

The limit probably should be high enough that this situation will be **very** rare, and should only really applicable in cases of deliberate ill-intent or extreme accidents.

gok99:
Part of #1504
Fixes #1716

**This PR implements the following:**

1. Frontend authorship view overall character limit. Only files within the total character limit are expanded. The remaining files are hidden to prevent expensive renders. This protects against views with a large number of small files (potentially solved with lazy loads in the future) or single large files (will still require a character limit in the future). This prevents most reports from crashing altogether on opening the authorship view.
1. File size limit is enabled by default with a default value. This value can be modified either in a standalone config, or in the repo config csv (can override standalone config).
1. File size limits are of two kinds which can be specified on a per-repo basis in the repo config csv:
    - [DEFAULT] Files above limit are analyzed but not included in the final report. Line diff values are displayed but files are tagged ignored. (Reduces report bloat, improving initial load-up)
  image
    - Files above limit are not analyzed, and diff values are not included in the aggregated diffs. (Can speed up analysis substantially)
1. A flag --ignore-filesize-limit or -I to ignore the size limit altogether. Can be set on a per-repo basis in the repo config csv.

**Proposed Commit Message**
```
Reports with files that are large tend to crash the report when rendered
in the authorship view. Furthermore, large files that users may have 
intended to ignore are analyzed and included in the report, causing
bloated reports and slow report generation.

Let's address these by:
- Adding a frontend overall character limit to render only files that in
total are under the character limit.
- Implement an override-able and ignore-able file size limit that can 
be used either to skip analysis entirely or to analyze but not include 
line diffs in the final report
```

**Additional Notes**
I've been using https://github.com/cheeaun/sgbusdata for testing features in the PR. It has plenty of large auto-generated files.


gerhean:
Correct me if I am wrong, but the original issue did not mention anything about large files, but rather large libraries being committed. 
Using nodejs as an example, this would be equivalent to the `node_modules` folder being committed, where each file maybe has an average of 200 lines, but the folder has over 30k files which is extremely problematic.  

image

gok99:
> Correct me if I am wrong, but the original issue did not mention anything about large files, but rather large libraries being committed.

I went with the approach of setting a diff limit for individual files (which seemed to be worth catching in any case) because: 
1) Choosing what set of potentially small files should be ignored didn't seem to me to be straight-forward to arbitrate. Ignoring all files in a directory (like a node_modules library) or all files touched by a certain commit didn't seem like great solutions (there could be legitimate reasons for having large folders or there could have been a large refactor). 
2) Reports generally struggled when the code view was opened. It seemed to me to be because of needing to render more content and was correlated with line diffs.

> Instead of lines of code, memory size should be considered instead

Thanks for this suggestion! I'll look into using memory usage instead.

frontend/src/views/v-authorship.vue
```diff
@@ -119,21 +127,31 @@
...
           ) {{ file.fileType }}&nbsp;{{ file.lineCount }}
             |&nbsp;({{ file.lineCount - file.blankLineCount }})
           span.fileTypeLabel.binary(v-if='file.isBinary') binary&nbsp;
-        pre.hljs.file-content(v-if="file.wasCodeLoaded && !file.isBinary", v-show="file.active")
+          span.fileTypeLabel(
+            v-if='file.isIgnored',
+            v-bind:style="{\
+              'background-color': '#000000',\
+              'color': '#ffffff'\
+              }"
```
gerhean, line 143: v-bind style is for dynamic styles, this isn't dynamic so it's better to use a class and place the css in the css section.

frontend/src/views/v-authorship.vue
```diff
@@ -119,21 +127,31 @@
...
+          ) ignored ({{ file.lineCount }})&nbsp;
+        pre.hljs.file-content(v-if="file.wasCodeLoaded && !file.isBinary && !file.isIgnored", v-show="file.active")
```
gerhean, line 145: Would [v-else-if](https://v2.vuejs.org/v2/guide/conditional.html) be better?

frontend/src/utils/api.js
```diff
@@ -209,6 +209,7 @@ window.api = {
     const res = [];
     const repo = window.REPOS[repoName];
 
+
```
gerhean, line 212: Try not to add extra newlines for non-format related PR.

frontend/src/views/v-authorship.vue
```diff
@@ -422,6 +446,12 @@ export default {
         out.wasCodeLoaded = lineCnt <= COLLAPSED_VIEW_LINE_COUNT_THRESHOLD;
         out.fileType = file.fileType;
 
+        if (file.isIgnored) {
+          out.isIgnored = true;
+        } else {
+          out.isIgnored = false;
+        }
+
```
gerhean, line 454: `out.isIgnored = file.isIgnored` ?

frontend/src/views/v-authorship.vue
gok99: Fixed

frontend/src/utils/api.js
gok99: Fixed

frontend/src/views/v-authorship.vue
gok99: file.isIgnored and file.isBinary can take values `true | undefined`

`out.isIgnored = !!file.isIgnored` should do the trick!

frontend/src/views/v-authorship.vue
gok99: @gerhean does the fix look better?

frontend/src/views/v-authorship.vue
gerhean: Seems good

gerhean:
Actually does it make sense to include say a `load anyway` button?

gok99:
I just came to a somewhat troubling (or reassuring depending on your philosophical inclinations) realization. The front-end currently already hides files above a certain line count threshold. For single files that cause loading issues, this largely fixes the problem and does not crash the report. The reason I didn't notice it before was because the repository I was using to test the limits of the report had a large number of files with line count smaller than the line count threshold and so the report was still crashing. Many files were single line jsons, which if pretty printed would definitely exceed the line count threshold. 

To specifically address this, I am currently including file sizes with authorship data, and using a total file size threshold. Now, for files that wouldn't be hidden by the single file line threshold, only an amount of files that are within the total file size threshold are displayed - all other files are collapsed. Users can choose to open a hidden file at their own peril. This seems to largely fix most instances of frontend crashes due to large files in most use cases. Do let me know if there might be a better solution to this.

I still think it does make sense to have a separate limit to ignore a file from a report entirely, to address files that were accidentally committed or that are unlikely to be of interest to human readers. Including diffs for these files seems like would cause unnecessary bloat. Ignoring files above a file size threshold entirely might help users automatically exclude files that they likely intended to exclude anyway. This limit can be overridden in a config file if a user wishes (although viewing these files in the report will almost certainly crash the browser). 

The following may help mitigate the intrusiveness/increase usefulness of this feature:
- We could allow users to disable the limit entirely (perhaps as a flag)
- We could let users choose if files above the threshold should be processed and accounted for and simply marked as ignored on the report, or ignored even from processing (related to #1716). This could be useful for users who care particularly about report generation times.

gerhean:
> To specifically address this, I am currently including file sizes with authorship data, and using a total file size threshold. Now, for files that wouldn't be hidden by the single file line threshold, only an amount of files that are within the total file size threshold are displayed - all other files are collapsed. 

I think you can calculate file size directly in the frontend by just checking the length of the string. No need to add extra field to json.

gok99:
> by just checking the length of the string

I just learned getting string length is a constant time operation in js, so this seems like a good solution too and involves minimal changes to the report. However, as discussed, selectively rendering code and de-loading them off screen seems like the ideal solution here. This could take some time to do (if possible), so perhaps a character limit could be a satisfactory temporary solution. Related PR: #1477

gok99:
The following strategies are currently used for robustness against large repositories:
1. Frontend authorship view overall character limit. Only files within the total character limit are expanded. The remaining files are hidden to prevent expensive renders. This protects against views with a large number of small files (potentially solved with lazy loads in the future) or single large files (will still require a character limit in the future). This prevents most reports from crashing altogether on opening the authorship view.
2. File size limit is enabled by default with a default value. This value can be modified either in a standalone config, or in the repo config csv (can override standalone config). 
3. File size limits are of two kinds which can be specified on a per-repo basis in the repo config csv: 
   - [DEFAULT] Files above limit are analyzed but not included in the final report. Line diff values are displayed but files are tagged `ignored`. (Reduces report bloat, improving initial bloat) 
   - Files above limit are not analyzed, and diff values are not included in the aggregated diffs. (Can speed up analysis substantially)
4. A flag `--ignore-filesize-limit` or `-I` to ignore the size limit altogether. Can be set on a per-repo basis in the repo config csv.

docs/ug/configFiles.md
```diff
@@ -40,11 +40,18 @@ Given below are the details of the various config files used by RepoSense.
 | Ignore Commits List<sup>*+</sup> | The list of commits to ignore during analysis. For accurate results, the commits should be provided with their full hash. Additionally, a range of commits can be specified using the `..` notation e.g. `abc123..def456` (both inclusive). |
 | Ignore Authors List<sup>*+</sup> | The list of authors to ignore during analysis. Authors should be specified by their [Git Author Name](#a-note-about-git-author-name). |
 | Shallow Cloning | Enter **`yes`** to clone the repository using Git's shallow cloning functionality. This option can significantly reduce the time taken to clone large repositories. However, the option should ideally be disabled for smaller repositories where the `.git` file is smaller than 500 MB, as it would create overhead. |
+| File Size Limit<sup>+</sup>  | Enter a file size limit for the repository that will override the default file size limit. |
```
yhtMinceraft1010X, line 43: Should we implement a custom file size limit flag? Most of the columns here have cli flag equivalents.

yhtMinceraft1010X (changes requested):
I also noticed that there are no changes to the config files for `systemtest` despite the addition of the new columns.

docs/ug/configFiles.md
```diff
@@ -40,11 +40,18 @@ Given below are the details of the various config files used by RepoSense.
 | Ignore Commits List<sup>*+</sup> | The list of commits to ignore during analysis. For accurate results, the commits should be provided with their full hash. Additionally, a range of commits can be specified using the `..` notation e.g. `abc123..def456` (both inclusive). |
```
yhtMinceraft1010X, line 40: The `ug` folder also has its own copy of `repo-config.csv`. Remember to add the new column headers and commas

docs/ug/configFiles.md
gok99: This could make sense! The reason I didn't also add a cli flag for this was because this felt like a fine knob for control over the innards of analysis that would rarely need to be meddled with. With a properly calibrated size limit, users would typically only need to increase the file limit (instead of decrease). This would mean that they'd need to enter very large numbers of at least 6 digits without error in a command, which seemed unwieldy. If they wanted to quickly lift the size limit, they could use the `-I` flag over cli.

gok99:
> I also noticed that there are no changes to the config files for systemtest

Thanks for the note! I hadn't touched the systemtests since none of the new columns are mandatory. It's probably a good idea to add to the systemtests as well.

chan-j-d (changes requested):
I only reviewed the docs and backend changes. PR looks really good! Just some very minor things I spotted, hope I didn't miss out too much. Unit test coverage is extremely comprehensive as well.

I was wondering if you might want to include a system test case which could for example, set a comically low file size limit such that everything is ignored. I'm not sure if the current setup would allow for easy implementation of this though. Just a thought as I think the unit tests might be good enough.

src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -87,6 +94,49 @@ protected void processLine(List<RepoConfiguration> results, CSVRecord record) th
...
+            if (!StringsUtil.isNumeric(fileSizeLimitString)
+                    || (parseValue = Integer.parseInt(fileSizeLimitString)) <= 0) {
+                logger.warning(String.format("Values in \"%s\" column should be positive integers.",
+                        FILESIZE_LIMIT_HEADER));
+                isFileSizeLimitOverriding = false;
+            } else {
+                fileSizeLimit = parseValue;
+            }
+        }
+
```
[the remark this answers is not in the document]
gok99: That's true. Not sure how to fix this in a way that adds more utility. Perhaps we could create a refactor issue to discuss it.

src/main/java/reposense/authorship/model/FileResult.java
```diff
@@ -14,34 +14,42 @@
...
         this.path = path;
         this.fileType = fileType;
         this.lines = lines;
         this.authorContributionMap = authorContributionMap;
+        // isBinary and isIgnored are mutually exclusive
         if (isBinary) {
             this.isBinary = true;
+        } else if (isIgnored) {
+            this.isIgnored = true;
         }
```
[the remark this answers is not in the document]
gok99: The else case here would be valid because most files would be neither binary not ignored. I've added an assertion that `isBinary` and `isIgnored` are not both true, since that would indicate programmer error that should be fixed.

docs/ug/configFiles.md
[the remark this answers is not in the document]
gok99: Good point! I've added this the units (bytes)

docs/ug/cli.md
```diff
@@ -110,6 +110,27 @@ This flag overrides the `Ignore standalone config` field in the CSV config file.
 
 <!-- ------------------------------------------------------------------------------------------------------ -->
 
+### `--ignore-filesize-limit`, `-I`
+
+**`--ignore-filesize-limit`**: Specifies that the file size limit for .
```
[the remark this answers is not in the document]
gok99: Wow good catch 😓 In this case `-I` would ignore both default and user-set size limits in config files.

yhtMinceraft1010X (approved):
LGTM

chan-j-d (approved):
LGTM! I reviewed only the docs and backend code.

frontend/src/views/v-authorship.vue
```diff
@@ -119,21 +127,27 @@
...
+          ) ignored ({{ file.lineCount }})&nbsp;
+        pre.hljs.file-content(v-if="file.wasCodeLoaded && !file.isBinary && !file.isIgnored", v-show="file.active")
           template(v-for="segment in file.segments")
             v-segment(v-bind:segment="segment", v-bind:path="file.path")
-        pre.file-content(v-if="file.isBinary", v-show="file.active")
+        pre.file-content(v-else-if="file.isBinary", v-show="file.active")
           .binary-segment
             .indicator BIN
             .bin-text Binary file not shown.
+        pre.file-content(v-else="file.isIgnored", v-show="file.active")
```
gerhean, line 148: I dont think `v-else` has arguments

gok99: That is true, thanks for catching this!

frontend/src/views/v-authorship.vue
```diff
@@ -119,21 +127,27 @@
...
               'color': getFontColor(fileTypeColors[file.fileType])\
               }"
           ) {{ file.fileType }}&nbsp;{{ file.lineCount }}
             |&nbsp;({{ file.lineCount - file.blankLineCount }})
           span.fileTypeLabel.binary(v-if='file.isBinary') binary&nbsp;
-        pre.hljs.file-content(v-if="file.wasCodeLoaded && !file.isBinary", v-show="file.active")
+          span.ignored-tag.fileTypeLabel(
+            v-if='file.isIgnored'
+          ) ignored ({{ file.lineCount }})&nbsp;
+        pre.hljs.file-content(v-if="file.wasCodeLoaded && !file.isBinary && !file.isIgnored", v-show="file.active")
```
gerhean, line 141: A little nitpicky, but would it be better to reorder the `v-if`s as:
`v-if="file.isBinary"`
`v-else-if="file.isIgnored"`
`v-else-if="file.wasCodeLoaded"`
So that you don't have to repeat `!file.isBinary && !file.isIgnored`?

frontend/src/views/v-authorship.vue
```diff
@@ -293,6 +312,7 @@ export default {
       if (!files) {
         files = await window.api.loadAuthorship(this.info.repo);
       }
+      this.getRepoProps(repo);
```
gerhean, line 315: Can I ask about why this was moved?

frontend/src/views/v-authorship.vue
```diff
@@ -306,27 +326,26 @@ export default {
...
+        let files = [];
         if (this.info.isMergeGroup) {
-          // sum of all users' file type contribution
-          repo.users.forEach((author) => {
-            this.updateTotalFileTypeContribution(author.fileTypeContribution);
-          });
+          files = repo.files;
         } else {
           const author = repo.users.find((user) => user.name === this.info.author);
-          if (author) {
```
gerhean, line 316: Why can the `if (author) {` be removed? Is it guaranteed that there will be author?

gok99: I can't quite remember now why I removed it (it may have been an oversight). In any case, a defensive author check seems like a good idea so I've added it back in.

frontend/src/views/v-authorship.vue
gok99: `getRepoProps` now calculates file type contributions with the actual files (instead of getting them from author) to filter out ignored files. Since files are only loaded in line 313, `getRepoProps` needs to be done after it.

frontend/src/views/v-authorship.vue
gok99: Implemented this suggestion.

gerhean (approved):
Frontend LGTM, will merge as 2 active developers already approved backend.

PR #1650, opened 2022-02-09
[#1202] Missing javadoc for functions

yhtMinceraft1010X:
Fixes #1202.

Proposed Commit Message:
```
Some non-trivial private methods have no Javadoc. At the same time, 
Javadoc styles vary across the files. This is also due to some Javadoc
checks not being used in Checkstyle.

Let's add the missing Javadoc, standardize the style and implement
additional checks in Checkstyle.
```

Proposed Javadoc standard for RepoSense (in addition to what is already covered in [SE-EDU](https://se-education.org/guides/conventions/java/index.html#comments) and [Google](https://google.github.io/styleguide/javaguide.html#s7-javadoc)) can be found in:
* Raw `styleGuides.md` file: https://github.com/reposense/RepoSense/pull/1650/files#diff-bd63b2861c8a46fbd8a67f313e555901c8d6d06762d1c0551c3254467f45b484
* Style Guides on `surge.sh`: https://docs-1650-pr-reposense-reposense.surge.sh/dg/styleGuides.html

PR #1652, opened 2022-02-09
[#1651] Enhance Checkstyle

### LINKED ISSUE #1651: Enhance Checkstyle

Checkstyle 8.1 does not have checks for certain Javadoc standards like empty line between description and block tags. At the same time, it is tedious to type out `gradlew checkstyleMain checkstyleTest checkstyleSystemtest` when doing these 3 tasks.

As a result of Checkstyle 8.1 limitations, some Javadoc comments in the files deviate from the standards given by [SE-EDU](https://se-education.org/guides/conventions/java/index.html#comments) and [Google](https://google.github.io/styleguide/javaguide.html#s7-javadoc), which were not picked up. I noticed the varying Javadoc styles while looking at #1202.

I am currently working on a migration to Checkstyle 9.3, which incorporates some new checks for Javadoc, such as the empty line between description and block tags. I also intend to add a gradle task called `checkstyleAll` that does `checkstyleMain`, `checkstyleTest` and `checkstyleSystemtest` together.

yhtMinceraft1010X:
Fixes #1651.

Proposed Commit Message:
```
Checkstyle 8.1 does not have checks for certain Javadoc standards like
whitespace after asterisks and an empty line between description and
block tag group. At the same time, it is tedious to type out all 3
checkstyle tasks for gradlew.

Let's migrate to Checkstyle 9.3 to add the missing checks and create a
consolidated checkstyleAll task to manage all 3 checkstyle tasks.
```


damithc:
@yhtMinceraft1010X are you wrapping commit message body at 50 chars? It should be 70 chars.

yhtMinceraft1010X:
> @yhtMinceraft1010X are you wrapping commit message body at 50 chars? It should be 70 chars.

Oops. I have fixed the commit message.

yhtMinceraft1010X:
CI is currently failing because the Javadoc has not been updated to pass the new `@throws` tag check.

Rather than do the update in this PR, I intend to wait for PR #1650 to be merged since I made most of the Javadoc updates there.

dcshzj:
I'll mark this PR as on hold first, since it depends on #1650 as you mentioned.

I am not exactly familiar with checkstyle, so two questions:
1. Is it possible to add the check for "each line should not start with punctuation/special characters"?
2. What's the difference between checkstyle 8.38 and 9.3? Can we just directly use 9.3 while retaining compatibility?

yhtMinceraft1010X:
> 1. Is it possible to add the check for "each line should not start with punctuation/special characters"?
> 2. What's the difference between checkstyle 8.38 and 9.3? Can we just directly use 9.3 while retaining compatibility?

1. There is a check called `SummaryJavadocCheck` which can do this. However, it is limited in that it only checks for the presence of a fullstop, not for every sentence.

Example of a Javadoc comment that has only one fullstop, but still passes.
```
/**
 * Analyzes the authorship of a {@code FileInfo} using the given annotations on the file
 * Only the lines with the format (START OF LINE) COMMENT_SYMBOL @@author ONE_STRING_WITH_NO_SPACE (END OF LINE)
 * will be analyzed. Otherwise, the line will be ignored and treated as normal lines
 * If the line is analyzed, and the string following the author tag is a valid git id, and there is no author config
 * file, then the code will be attributed to the author with that git id. Otherwise, the code will be attributed to
 * unknown author
 */
```
2. I stopped at Checkstyle 8.38 because that was what I considered the minimum version for implementing some new checks. However, it appears that Checkstyle 9.3 still works for Java 8 so I'm moving forward with it.

yhtMinceraft1010X:
Re-requesting review from @reposense/active-developer as CI is now passing.

Some checks have been removed since it is better to implement them in PR #1650, which is expected to take longer.

config/checkstyle/checkstyle.xml
```diff
@@ -341,46 +340,62 @@
...
     -->
+    <!-- Checks that all block-tags are ordered correctly. -->
+    <module name="AtclauseOrder"/>
+
+    <!-- Checks that Javadoc block tags appear only at the beginning of the line. -->
+    <module name="JavadocBlockTagLocation"/>
+
+    <!-- Checks that all Javadoc comments start from the second line. -->
+    <module name="JavadocContentLocationCheck">
+      <property name="location" value="second_line"/>
```
dcshzj, line 351: I think we can remove this line, from [the documentation](https://checkstyle.sourceforge.io/config_javadoc.html#JavadocContentLocation) the default value is `second_line`.

docs/dg/workflow.md
```diff
@@ -43,7 +43,10 @@ The issues for first timers usually have guidance provided in the comment or hav
...
+* **This project uses Checkstyle** to check the compliance of Java code. You can use [this document](https://se-education.org/guides/tutorials/checkstyle.html) to find how to use it. 
+  * In particular, run `gradlew checkstyleAll` to check the style of all the relevant Java code. This will check the Java code in the following order: main code -> test code -> systemtest code.
+  * If you only want to check the main code, run `gradlew checkstyleMain`. Likewise, for test code, run `gradlew checkstyleTest` and for systemtest code, run `gradlew checkstyleSystemtest`.
+  * You can also run any of the combination of the above, such as `gradlew checkstyleMain checkstyleTest` or `gradlew checkstyleTest checkstyleSystemtest`.
 * **To check Pug files for style errors**, run `npm run lint` from the project root directory. You can use the `npm run lintfix` to automatically fix some of the JavaScript and CSS lint errors.
```
dcshzj, line 50: @gerhean I think this part is outdated?

dcshzj (changes requested):
I think overall is okay, we can add new checks in subsequent PRs.

docs/dg/workflow.md
gerhean: I don't touch anything regarding the 'compliance of Java code', so I cannot really give a good assessment of whether this part is outdated. As far as I know, the highlighted green parts are still correct.
However style error checks for the Pug portion of the frontend are currently broken, but that isn't the concern of this PR. `npm run lint` also no longer works on the project root directory, but isn't the concern of this PR either.

dcshzj: Yup I was referring to the Pug files. Do we have a current working way to lint the frontend?

gerhean: Yes, running `npm run lint` in the `frontend` folder lints the JavaScript.

dcshzj: Okay, @yhtMinceraft1010X is it possible to update this line as well? We can change instead of running "from the project root directory", specify that it should be run in the `frontend` folder. We can fix the linting in another PR.

gerhean: You can run `gradlew lintFrontend` too from root, but there is currently no lintfix gradle command.

yhtMinceraft1010X: Should `Pug` be changed to `frontend` or `Vue`? We don't have anymore `.pug` files since all the `pug` code is now in `.vue` files.

gerhean: I guess you can say Vue files.

dcshzj (approved):
LGTM!

PR #1653, opened 2022-02-10
[#1619] Remove unused GitLsTree class and methods

### LINKED ISSUE #1619: Refactor file validation method in GitLsTree
**Currently**: `GitLsTree` contains only one public method, `validateFilePaths` and two private methods. Of the two private methods, only one of them uses a Git command directly.

**Suggestion**: Move the `validateFilePaths` and `isValidWindowsFilename` methods to `FileUtil` and change the method `getFilePaths` (that uses the `git ls-tree` command) to `public`. 

This would put the `GitLsTree` class more in-line with the other git classes which contain methods mainly to run a specific git command. At the moment, the current method `validateFilePaths` performs verification on all the files in the repository and throws an exception if any of the files contains an invalid filename for Windows computers. I believe that the logic on how to verify and what actions to perform if a file fails verification should not be located in the Git class itself.

chan-j-d:
Fixes #1619 

After merging #1628, the `GitLsTree` class and methods are no longer used. Based on the discussion in #1628, it would be more appropriate to handle exceptional file names on a case-by-case basis instead of skipping the entire repository for a single invalid file name. As such, I do not think there are any use cases for a `validateFilePaths` method.

Additionally, there are 5 test cases used for this method which need to be maintained if we are to leave this in. If necessary, we can always recover the code from this commit. 

The mention of `GitLsTree` has also been removed from the DG.

Commit Message:
```
GitLsTree class and methods are no longer used. They should be removed
to reduce the maintenance burden.

Let’s remove the unused code in GitLsTree and its test class.
```


yhtMinceraft1010X (commented):
Since `GitLsTree` is to be removed, let's also delete `GitLsTree` from `docs/dg/architecture.md`.

Other than that, the code looks okay.

chan-j-d:
> Since `GitLsTree` is to be removed, let's also delete `GitLsTree` from `docs/dg/architecture.md`.
Oh thanks! I overlooked that. Will remove that as well.

dcshzj:
@chan-j-d #1628 has been merged, you can proceed with this PR.

yhtMinceraft1010X (commented):
Looks good, would like to request for @reposense/stage-1-active-reviewer.

dcshzj:
Issue #1619 mentioned that you want to refactor to use `FileUtil::isValidPath`, is that still accurate?

chan-j-d:
No. Usage of `FileUtil::isValidPath` was a suggestion as replacement to a similar method in `GitLsTree` `isValidWindowsFilename` as it already checks if a file path is valid for Windows if RepoSense was running on Windows. However, since the entire class is being removed, that part is not relevant any more.

dcshzj (approved):
LGTM!

dcshzj:
I will move this PR along, since the changes do not affect the logic of the software (just cleanup of unused code).

PR #1654, opened 2022-02-11
[#1615] Clear deployments for closed pull requests

### LINKED ISSUE #1615: GitHub Environments should be cleaned up after a pull request is merged
The current setup of our GitHub Actions workflow is that 2 deployments are made for every pull request. However, these deployments are not deleted once the pull request is merged, which can cause lots of stale environments to be present:

Screenshot 2022-01-15 at 14 55 29

There should be a workflow that cleans up the environments after a pull request is merged, so that the number of environments present is 2x + 3, where x is the number of open pull requests.

gok99:
Fixes #1615 

Proposed Commit Message:
```
Deployment previews generated for open PRs are not cleared
automatically when a PR is closed or merged. As a result, environments
need to be cleared manually periodically, which is less than ideal.

Let's add a GitHub workflow for closed (which includes merged) PRs
that deletes generated preview environments automatically, and posts 
a comment on the PR with the preview links for later reference.
```


gerhean:
Will it be good to destroy the surge websites as well? Or should they just continue being hosted?

gok99:
> Will it be good to destroy the surge websites as well? Or should they just continue being hosted?

I don't see downsides to leaving them up, if it doesn't cost us anything (it could be helpful to have a snapshot of a particular version at least for some time after the PR is merged). If we're using a paid subscription that prices by deployments, perhaps it'd be a good idea to take them down.

chan-j-d:
Now that you mention it, I went to look around and it seems that the following still work: 
- https://dashboard-1562-pr-reposense-reposense.surge.sh/ 
with the deployment environment url 
https://github.com/reposense/RepoSense/deployments/activity_log?environment=dashboard-1562
- https://dashboard-1400-pr-reposense-reposense.surge.sh/
with the deployment environment url
https://github.com/reposense/RepoSense/deployments/activity_log?environment=dashboard-1400

I'm not sure if the environment in 1562 still exists but the environment page for 1400's dashboard is empty while the dashboard itself works fine.

dcshzj:
> I don't see downsides to leaving them up

I agree here. It is better to keep them up so that we can potentially trace back to the exact PR that a bug arises in the future. We can revisit deleting actual Surge.sh deployments if cost is an issue in the future.

> I'm not sure if the environment in 1562 still exists but the environment page for 1400's dashboard is empty while the dashboard itself works fine.

Some of the deployments might be in an inconsistent state, as I was manually deleting some of them about 2 weeks ago. I think as long as it does not appear under Environments in the main page of the repository, it should be fine.

Also, @gok99 I think the script's name should not be `delete.sh`, as it may be confusing. Perhaps something like `delete-deployments.sh` would be better?

gok99:
> think the script's name should not be delete.sh, as it may be confusing. Perhaps something like delete-deployments.sh would be better?

Thanks for the suggestion! I've renamed it as suggested.

gerhean:
> I agree here. It is better to keep them up so that we can potentially trace back to the exact PR that a bug arises in the future. > We can revisit deleting actual Surge.sh deployments if cost is an issue in the future.

In that case would it be possible to maybe maintain a link to the deployment? Maybe when a PR is closed we can automatically comment the link to the surge deployment in the PR? So that there is an easy way to access the site other than changing url manually. @dcshzj

gok99:
> Maybe when a PR is closed we can automatically comment the link to the surge deployment in the PR?

This seems like a good idea! I can add this in here if it is within scope of this PR, or open a separate one if it's not.

gok99:
> Yup I think it's a good idea to add the link as a comment when the PR is closed. Let's do that in this same PR.

I've added this, and it seems to work on my fork. Github actions leaves the following comment on the closed PR:
```
Dashboard Preview: DASHBOARD_URL
Docs Preview: DOCS_URL
```

> Looks good. Let's also change docs/dg/devOpsGuide.md to reflect the new functionality, especially the Surge.sh section.

I've also updated the dev ops guide. The section on Surge.sh already mentions that deployments are kept forever, and I'm not interacting with Surge in this workflow, so I didn't make any changes to that section.

yhtMinceraft1010X:
> > Yup I think it's a good idea to add the link as a comment when the PR is closed. Let's do that in this same PR.
> 
> I've added this, and it seems to work on my fork. Github actions leaves the following comment on the closed PR:
> 
> ```
> Dashboard Preview: DASHBOARD_URL
> Docs Preview: DOCS_URL
> ```
> 
> > Looks good. Let's also change docs/dg/devOpsGuide.md to reflect the new functionality, especially the Surge.sh section.
> 
> I've also updated the dev ops guide. The section on Surge.sh already mentions that deployments are kept forever, and I'm not interacting with Surge in this workflow, so I didn't make any changes to that section.

Nice. Currently the links to the files in the guide do not work, but this is because they are not yet on the `master` branch. Other than that, I have no other comments and would like to request review from @reposense/stage-1-active-reviewer.

gerhean (changes requested):
Please update commit message to include the commenting thanks.

gerhean (approved):
LGTM

config/gh-actions/delete-deployments.sh
```diff
@@ -0,0 +1,105 @@
...
+  -H "Authorization: token ${GITHUB_TOKEN}"
+}
+
+# Function to post surge deployment links as a comment on PR via a cURL command
+post_preview_links_comment() {
+  curl "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/issues/${ACTIONS_PULL_REQUEST_NUMBER}/comments" \
+  -X POST \
+  -H "Accept: application/vnd.github.v3+json" \
+  -H "Authorization: token ${GITHUB_TOKEN}" \
+  -d "{\"body\": \"Dashboard Preview: ${DASHBOARD_DEPLOY_DOMAIN}\nDocs Preview: ${DOCS_DEPLOY_DOMAIN}\"}"
```
dcshzj, line 78: I would prefer to provide a little more context in the comment, something along these lines:

```
The following links are for previewing this pull request:
- Dashboard preview: <link>
- Docs preview: <link>
```

I feel that the original phrasing may be a little confusing for people who are not used to our workflow. Also, the original comment uses a single newline, but Markdown requires a double newline to be recognized as a new paragraph.

.github/workflows/clear-deploy.yml
```diff
@@ -0,0 +1,24 @@
+name: Clear deployments for closed pull requests
```
dcshzj, line 1: I think we should standardize whether or not we mean "clear deployments" or "delete deployments". I am okay with using either, but would be best if file names and terminology used are all consistent with one another.

dcshzj (changes requested):

config/gh-actions/delete-deployments.sh
gok99: The comment looks something like this currently with a single newline:
image

.github/workflows/clear-deploy.yml
gok99: I've changed everything to 'delete' and renamed `delete-deployments.sh` to `delete-deploy.sh` for brevity and for similarity with `delete-deploy.yml`

docs/dg/devOpsGuide.md
```diff
@@ -53,6 +53,8 @@ Both the "Surge.sh pending build" and "Continuous Integration" workflows produce
 
 This task is not performed on commits to the repository, as there is no need to do so.
 
+After the pull request is closed or merged, the Surge.sh deployments are retained, while the GitHub environments for them are cleared by another workflow "Clear deployments for closed pull requests" (defined in [`clear-deploy.yml`](https://github.com/reposense/RepoSense/blob/master/.github/workflows/clear-deploy.yml)). The workflow uses [`delete-deployments.sh`](https://github.com/reposense/RepoSense/blob/master/config/gh-actions/delete-deployments.sh) (found in the `config/gh-actions/` folder) to remove the environments on GitHub, and to leave a comment on the pull request with links to the Surge.sh previews for later reference.
```
dcshzj, line 56: The link to the file should be updated.

dcshzj (changes requested):

docs/dg/devOpsGuide.md
gok99: Thanks for catching that!

dcshzj (approved):
LGTM! Will merge this PR first before the others to see if it works, in about 2 days.

dcshzj:
@gok99 Looks like it is not working as intended. The error message gives `Resource not accessible by integration`. Can you have a look?

PR #1656, opened 2022-02-11
[#1648] Frontend: display reportTitle as browser tab title

### LINKED ISSUE #1648: Report does not display the value of reportTitle in <title>

1. Visit the RepoSense preview website for the latest master branch: https://reposense.netlify.app
1. Observe that the tab title (obtained from &lt;title>) says "reposense" instead of "RepoSense Report" configured in summary.json.

The page title should say "RepoSense Report", or whatever the value of `title` is configured to in `report-config.json`.

The page title says "reposense", as shown in the screenshot below:
Screenshot 2022-02-07 at 10 09 01

https://reposense.netlify.app

gok99:
Fixes #1648 

Before:

image

After:

image

Proposed Commit Message:
```
The title of the browser window displays 'reposense' instead of the 
`reportTitle` as specified in report-config.json. This is becuase the 
default index page title in the backend is wrongly specified. 

Let's correctly specify the default index page title, so that it gets 
correctly replaced by the specified reportTitle and shown in the browser
tab.
```

### DIFF
--- a/frontend/src/utils/api.js
+++ b/frontend/src/utils/api.js
@@ -177,7 +177,7 @@ window.api = {
       }
       throw error;
     }
-    const { reportGeneratedTime, reportGenerationTime } = data;
+    const { reportTitle, reportGeneratedTime, reportGenerationTime } = data;
     window.sinceDate = data.sinceDate;
     window.untilDate = data.untilDate;
     window.repoSenseVersion = data.repoSenseVersion;
@@ -196,6 +196,7 @@ window.api = {
       names.push(repoName);
     });
     return {
+      reportTitle,
       creationDate: reportGeneratedTime,
       reportGenerationTime,
       errorMessages,
--- a/frontend/src/utils/api.js
+++ b/frontend/src/utils/api.js
@@ -183,7 +183,7 @@ window.api = {
     window.repoSenseVersion = data.repoSenseVersion;
     window.isSinceDateProvided = data.isSinceDateProvided;
     window.isUntilDateProvided = data.isUntilDateProvided;
-    document.title = data.reportTitle || 'RepoSense Report';
+    document.title = data.reportTitle || document.title;
 
     const errorMessages = {};
     Object.entries(data.errorSet).forEach(([repoName, message]) => {
--- a/frontend/src/utils/api.js
+++ b/frontend/src/utils/api.js
@@ -177,12 +177,13 @@ window.api = {
       }
       throw error;
     }
-    const { reportTitle, reportGeneratedTime, reportGenerationTime } = data;
+    const { reportGeneratedTime, reportGenerationTime } = data;
     window.sinceDate = data.sinceDate;
     window.untilDate = data.untilDate;
     window.repoSenseVersion = data.repoSenseVersion;
     window.isSinceDateProvided = data.isSinceDateProvided;
     window.isUntilDateProvided = data.isUntilDateProvided;
+    document.title = data.reportTitle || 'RepoSense Report';
 
     const errorMessages = {};
     Object.entries(data.errorSet).forEach(([repoName, message]) => {
@@ -196,7 +197,6 @@ window.api = {
       names.push(repoName);
     });
     return {
-      reportTitle,
       creationDate: reportGeneratedTime,
       reportGenerationTime,
       errorMessages,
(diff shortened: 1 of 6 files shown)


gerhean:
Any reason why `App.vue` needs to know about the report title? Why not set the `document.title` directly in `api.js`?

gok99:
> Any reason why App.vue needs to know about the report title? Why not set the document.title directly in api.js?

@gerhean The intention was to store the title somewhere if we needed it (even though App.vue doesn't currently use it). I realize I could have put it into window.reportTitle. Would that be better to do, or should we not retain the title at all until we have reason to?

gerhean:
It's generally better to not leave unneeded things lying around. Unused code are being cleaned up semi-regularly, so it's best not to have them in the first place.
I rather have clean code and not wonder about whether a certain variable is actually useless a few months down the road.

gok99:
> I rather have clean code and not wonder about whether a certain variable is actually useless a few months down the road 

That makes sense! I'll keep this within `api.js` then.

chan-j-d (changes requested):
The part where it updates the title to use the `reportTitle` in `data` seems to be working. 

However, running it with the report-config with title below:
image
seems to produce no difference. 

image
The resulting title is still the same.

chan-j-d:
It seems like the `ReportConfiguration` class does not actually have an option of a title to begin with. 

https://github.com/reposense/RepoSense/blob/728a9f9c14f706ace6873cae8696647b91ff3ffb/src/main/java/reposense/model/ReportConfiguration.java#L6-L13

If I did not make a mistake, `SummaryJson` class gets its report title from here which is then used in the report title itself it seems.

gok99:
> If I did not make a mistake, SummaryJson class gets its report title from here which is then used in the report title itself.

It seems the report config is ignored if the report is generated with `--repos`

https://github.com/reposense/RepoSense/blob/728a9f9c14f706ace6873cae8696647b91ff3ffb/src/main/java/reposense/parser/ArgsParser.java#L269-L273

The report config title works for me if i put `report-config.json` in a config folder with a `repo-config.csv` and generate the report with a `--config` flag instead.

chan-j-d (approved):

chan-j-d:
My bad, I tested it wrongly with the --repos argument.

frontend/src/utils/api.js
```diff
@@ -183,6 +183,7 @@ window.api = {
     window.repoSenseVersion = data.repoSenseVersion;
     window.isSinceDateProvided = data.isSinceDateProvided;
     window.isUntilDateProvided = data.isUntilDateProvided;
+    document.title = data.reportTitle || 'RepoSense Report';
```
gerhean, line 186: I believe `document.title = data.reportTitle || document.title;` would be better, as I prefer only a single source for the tab default name. I think the default name can be changed in `package.json`.

gerhean (changes requested):

frontend/src/utils/api.js
gerhean: I mean `index.html`, my bad.

gok99: Thanks for the suggestion! I was looking for a way to standardize the default name with backend but couldn't think of any.

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -69,7 +69,7 @@
     // zip file which contains all the report template files
     private static final String TEMPLATE_FILE = "/templateZip.zip";
     private static final String INDEX_PAGE_TEMPLATE = "index.html";
-    private static final String INDEX_PAGE_DEFAULT_TITLE = "<title>RepoSense Report</title>";
+    private static final String INDEX_PAGE_DEFAULT_TITLE = "<title>reposense</title>";
```
gok99, line 72: I feel like I should explain this. The default title in the template index.js reflects the project name in package.json which currently is `reposense`. So, it should replace `"<title>reposense</title>"` and not `"<title>RepoSense Report</title>"` with the report config title. The report config title defaults to `RepoSense Report` so the the title in index.html after replacement will be `RepoSense Report`

gerhean:
Actually would it be good to write a quick test to verify that it will work?

src/main/java/reposense/report/ReportGenerator.java
dcshzj: Can I confirm if the replacement is done in line 180 of this file?

gok99: Yes, line 180 is the only reference to `INDEX_PAGE_DEFAULT_TITLE`. The replacement is done along with `prepareTemplateFile`.

gerhean:
Ok then let's merge this as a hotfix first.

Issue #1627, opened 2022-01-22
Upgrade frontend to Vue3


gerhean:
**What feature(s) would you like to see in RepoSense?**

Upgrade to Vue3

**Is the feature request related to a problem?**

Vue3 has better support for TypeScript, and TypeScript works well with most text editors to easily check what are in those 3 to 4 layer nested data objects passed to various functions.

Also Vue3 might be faster, current frontend is quite slow for large reports.

**If applicable, describe alternatives you've considered**

Can consider integrate typescript into Vue2, although less recommended.

PR #1658, opened 2022-02-12
Remove deprecated slot syntax

gerhean:
Needed for #1627

```
The current syntax used to provide content to slots has been officially
deprecated, and cannot be used in Vue 3.

Lets use the new slot syntax.
```

Refer [here](https://v2.vuejs.org/v2/guide/components-slots.html?redirect=true#Deprecated-Syntax) for deprecation notice.

Declaring a named slot like in `v-resizer` still uses the same syntax as mentioned [here](https://v2.vuejs.org/v2/guide/components-slots.html?redirect=true#Named-Slots).


dcshzj (changes requested):
I think `v-resizer.vue` needs to be updated too?

gerhean:
> I think v-resizer.vue needs to be updated too?

Declaring a named slot like in v-resizer still uses the same syntax as mentioned [here](https://v2.vuejs.org/v2/guide/components-slots.html?redirect=true#Named-Slots).

dcshzj (approved):

frontend/src/app.vue
```diff
@@ -6,65 +6,69 @@
     v-bind:opacity='loadingOverlayOpacity',
     v-bind:is-full-page="true"
   )
-    i.overlay-loading-icon.fa.fa-spinner.fa-spin(slot='default')
-    h3(slot='after') {{ loadingOverlayMessage }}
+    template(slot:default)
+      i.overlay-loading-icon.fa.fa-spinner.fa-spin()
+    template(slot:after)
```
dcshzj, line 11: @gerhean Can I double check if `slot` should be changed to `v-slot` here?

gerhean: It's the same, but I suppose  `v-slot` is safer.

dcshzj:
Moving this PR ahead.

PR #1662, opened 2022-02-13
[#1657] Escape special chars for bash commands

### LINKED ISSUE #1657: Special characters such as $ not properly escaped in BASH commands

* **OS and Version:** Ubuntu 18.04

1. Run `java -jar RepoSense.jar --repos https://github.com/AY2021S1-CS2103-F09-1/tp.git --since 01/06/2020 --view`

Report generated and displayed without issue.

**What actually happened?**
Report was not generated. 

Reason for the problem can be found in the discussion in https://github.com/reposense/RepoSense/issues/1636#issuecomment-1036172244. 

[ubuntu-log.txt](https://github.com/reposense/RepoSense/files/8052739/ubuntu-log.txt)

chan-j-d:
Fixes #1657 

Currently, the file path arguments are added to `git` commands using normal double quotes. However, for Bash terminals, double quotes still allow for special symbols like `$` to perform a special function like referencing a variable name. There is a need to use single quotes for Bash terminals. Single quotes `'` themselves are escaped as such `'"'"'` by breaking up the string into various quoted segments where a single quote is left in a double quoted segment as per this [stackoverflow post](https://stackoverflow.com/questions/1250079/how-to-escape-single-quotes-within-single-quoted-strings). 

The test case specifically tests against special bash characters according to this [unix stackexchange post](https://unix.stackexchange.com/questions/347332/what-characters-need-to-be-escaped-in-files-without-quotes). 

Unfortunately, Windows CMD does not work with single quotes and only accepts double quotes. As such, there needs to be a distinction in the way this is done for different OS.

A small note was added to the DG of the importance in using this additional method. 

Commit message:
```
Special characters in file names are not properly translated into git
commands on bash for non-Windows OS. Usage of single quotes on bash is
only necessary to escape such special characters.

Let's fix the escaping of special characters for bash commands.
```


src/main/java/reposense/util/StringsUtil.java
```diff
@@ -31,10 +31,33 @@ public static String replaceSpecialSymbols(String regexString, String replacemen
...
+    public static String addQuotes(String original) {
         return "\"" + original + "\"";
     }
 
+    /**
+     * Adds the appropriate quotation marks for a file path depending on the OS.
+     */
+    public static String addQuotesForFilePath(String filePath) {
+        if (SystemUtil.isWindows()) {
+            return "\"" + filePath + "\"";
```
Zhou-Jiahao-1998, line 46: Is it possible to cover this line in the test as well?

Zhou-Jiahao-1998 (commented):
I think it is trivial but maybe we can cover the case on that line. Otherwise LGTM.

src/main/java/reposense/util/StringsUtil.java
```diff
@@ -31,10 +31,33 @@ public static String replaceSpecialSymbols(String regexString, String replacemen
...
+     * Adds quotes to the string.
+     */
+    public static String addQuotes(String original) {
         return "\"" + original + "\"";
     }
 
+    /**
+     * Adds the appropriate quotation marks for a file path depending on the OS.
+     */
+    public static String addQuotesForFilePath(String filePath) {
```
gok99, line 44: It might be possible that developers forget to use these if they weren't sure how or when to use them. Do you think it would be a good idea now to create a CommandBuilder to take care of these things.

This could be tricky though, given that the construction of commands can be non-trivial so this might be out of scope for this PR.

src/test/java/reposense/util/StringsUtilTest.java
```diff
@@ -70,4 +73,17 @@ public void removeTrailingBackslash_emptyString_noChange() {
 
         Assert.assertEquals(emptyString, actualString);
     }
+
+    @Test
+    public void addQuotationMarksForPath_specialBashCharacters_success() {
+        Assume.assumeTrue(!SystemUtil.isWindows());
+        String specialBashSymbols = "!\"#$&'()*,;<=>?\\[]^`{| }";
```
gok99, line 80: Perhaps we could link the relevant SO posts here and in `addQuotesForFilePath` for easy reference.

src/main/java/reposense/util/StringsUtil.java
chan-j-d: Unfortunately, I don't think there is a good way to test the escaping specifically. We can add a test case just to check if quotation marks are added which does not seem very useful. For this case, it was to avoid mishandling of spaces in the repo directory and the test cases are located [here](https://github.com/reposense/RepoSense/commit/2c05624a7c59e3d47c7a5ddf97f800b8655b7b99#diff-e940f5fe270077934d6b1a6ecb957d627681a5a38abaea308070d46b27e130ebR36). 

For testing the escaping specifically, I don't think there's a good way to test this for Windows filesystems. The CMD's `echo` command echoes the quotation marks compared to bash's `echo` command which strips the quotation marks and does symbol substitution where appropriate.

src/main/java/reposense/util/StringsUtil.java
chan-j-d: Interesting. I actually voiced it [here](https://github.com/reposense/RepoSense/pull/1562#issuecomment-877820941) and you can read the response from dcshzj but besides adding paths, most other argument types are quite tailored to the specific git command. So this is a specific standard for file paths only.

src/test/java/reposense/util/StringsUtilTest.java
chan-j-d: Noted, will update this.

src/main/java/reposense/util/StringsUtil.java
gok99: Thanks for the link! It's interesting seeing similar discussions and considerations springing up independently at different points in this project :)

src/main/java/reposense/util/StringsUtil.java
chan-j-d: I think it's more that there isn't a specific test case that runs this line of code at all since the test case I added only runs on non-Windows systems.

src/test/java/reposense/util/StringsUtilTest.java
chan-j-d: I've updated it.

src/main/java/reposense/util/StringsUtil.java
```diff
@@ -31,10 +31,33 @@ public static String replaceSpecialSymbols(String regexString, String replacemen
...
+        if (SystemUtil.isWindows()) {
+            return "\"" + filePath + "\"";
+        } else {
+            StringBuilder sb = new StringBuilder();
```
gok99, line 48: Should we link https://stackoverflow.com/a/1250279 here as well?

src/test/java/reposense/util/StringsUtilTest.java
gok99: Let's link the answer specifically: https://unix.stackexchange.com/a/357932

gok99 (changes requested):

src/main/java/reposense/util/StringsUtil.java
chan-j-d: I think this is far more straightforward. It is a full single quote wrap while specially escaping single quotes. There are various ways to do it. Replacing `'` with `'\''` instead of `'"'"'` also works if I am not wrong. The test case is also written to catch this specific issue if someone tries to change it in the future.

src/test/java/reposense/util/StringsUtilTest.java
chan-j-d: Resolved.

gok99 (approved):
LGTM!

docs/dg/architecture.md
```diff
@@ -27,6 +27,7 @@
 ## Git
 
 [`Git`](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/git) package contains the wrapper classes for respective *git* commands.
+Note that when constructing new commands containing path arguments, use `StringsUtil::addQuotesForFilePath` method to safely convert a Java string into an equivalent Bash/CMD argument.
```
dcshzj, line 30: Should move to the end of this section (Git) and use `<box type="info" seamless>` so that it looks nicer on the DG.

dcshzj (changes requested):

docs/dg/architecture.md
chan-j-d: Have changed it to use markbind's box. Seems like there needs to be an empty line after the first `<box>` tag to properly parse code wrapped in " ` ".

chan-j-d:
@dcshzj @fzdy1914 would it be possible to get a final review sooner or merge it earlier? I think the `StringsUtil::addQuotesForFilePath` method is necessary for issue #1682 which is further exacerbated in PR https://github.com/reposense/RepoSense/pull/1644 when the output directory now retains information of the entire path that might have more spaces.

dcshzj:
I will move this PR along since it is necessary for some of the other PRs.

Issue #1661, opened 2022-02-13
Upgrade Gradle to at least 7.3


yhtMinceraft1010X:
**What feature(s) would you like to see in RepoSense?**

As mentioned in #1635, Gradle 5.2.1 doesn't support Java 17. There is thus a need to move to a minimum of Gradle 7.3 since that is the first Gradle version to support Java 17. 

**If applicable, describe alternatives you've considered**

~~Gradle has also released 7.4 very recently. This version also incorporates the [latest mitigation for the log4j vulnerability](https://github.com/gradle/gradle/issues/19526). I suggest going straight for 7.4 if possible.~~
Gradle has released 7.4.1 recently. I suggest going straight for it.

**Update**
There are 3 issues described below that may not be easy to solve if we are to jump straight to `7.4` or even `7.3`. ~~In the meantime, I will try to upgrade Gradle to `6.9.2`.~~

gerhean:
Seems fine, you don't need to create an issue for simple tasks like this, if it involves only a simple version upgrade (with no refactoring), and does not break backwards compatibility (which you should mention).

yhtMinceraft1010X:
I have encountered some issues with third-party plugins when attempting to move to Gradle 7.4:

**1. Property without annotation error in `com.liferay.gradle.plugins.node.tasks.ExecutePackageManagerTask`**

When running `gradlew`, I noticed this error:
![Screenshot (91)](https://user-images.githubusercontent.com/69678785/153754619-d5112a08-7b7b-41b8-b5d1-20171dce05c6.png)

After some googling, I discovered that Gradle 7 implemented stricted validation checks as described in this [link](https://dev.to/runningcode/a-problem-was-found-with-the-configuration-of-the-gradle-7-task-3ej1)

Possible remedy: There's another Gradle plugin that might work, [gradle-node-plugin](https://github.com/node-gradle/gradle-node-plugin). It is described as being tested for Gradle 7.4, although I wish to open this for discussion before deciding whether to change plugins.

**2. No compatibility with Java 8 for latest available version of `com.github.psxpaul.execfork`**

The latest available version for the aforementioned plugin is `0.1.15`. However, trying to use this version results in the below error:
![Screenshot (93)](https://user-images.githubusercontent.com/69678785/153755105-1f57167e-6ff0-46bb-adb8-763eaf4ec702.png)

Possible remedy: Using version `0.1.13` does not throw the same error. 

[...]

PR #1664, opened 2022-02-13
Upgrade Gradle to 6.9.2

yhtMinceraft1010X:
Part of #1661. Upgrading to version `6.9.2` is meant to be a stopgap measure while resolving errors from third-party plugins.

Propsed Commit Message:
```
Gradle has been stuck at 5.2.1 for quite some time. This version is not
compatible with a future migration to Java 17. However, upgrading to
Gradle 7.4 cannot be done yet due to a need to resolve errors from
third-party plugins.

Let's upgrade Gradle to 6.9.2 in the meantime to close the gap between
the version we are using and the latest available version.
```


chan-j-d:
I've requested a second run of the deployment. It seems the previous build failed on this 
https://github.com/reposense/RepoSense/runs/5174576252?check_suite_focus=true#step:11:1465.
I've encountered it myself a few times and it's always somewhat random, perhaps we should raise a separate issue about it.

yhtMinceraft1010X:
> I've requested a second run of the checks. It seems the previous build failed on this https://github.com/reposense/RepoSense/runs/5174576252?check_suite_focus=true#step:11:1465. I've encountered it myself a few times and it's always somewhat random, perhaps we should raise a separate issue about it.

Might be a concurrency issue similar to #1601. I think we should wait for #1625 to be merged.

build.gradle
```diff
@@ -103,40 +103,44 @@ test {
...
 
-task installFrontend(type: com.liferay.gradle.plugins.node.tasks.ExecuteNpmTask) {
+tasks.register('installFrontend', com.liferay.gradle.plugins.node.tasks.ExecutePackageManagerTask) {
     workingDir 'frontend/'
     args = ['install']
 }
 
-task buildFrontend(dependsOn: ['installFrontend'], type: com.liferay.gradle.plugins.node.tasks.ExecuteNpmTask) {
+tasks.register('buildFrontend', com.liferay.gradle.plugins.node.tasks.ExecutePackageManagerTask) {
+    dependsOn 'installFrontend'
```
chan-j-d, line 120: Would it be possible to do an assignment like `def installFrontend = tasks.register(...)`? This way all further references to `installFrontend` like in line 120 can reference `installFrontend` without quotations directly. At least this seems possible when I tried it locally.

chan-j-d (approved):
LGTM! The current implementation seems to work well. Just some minor comments/clarifications.

Would like to confirm if rewriting the task definitions as `tasks.register(...)` is to support a future gradle 7 upgrade [described here](https://docs.gradle.org/7.4.1/userguide/more_about_tasks.html#sec:defining_tasks) since it seems the gradle 6.8.2 documentation uses quite a different syntax.
Additionally, would it be possible to assign the tasks to variables to avoid referencing them by string? More details are described in the specific comment.

build.gradle
yhtMinceraft1010X: I managed to change some of the assignments

yhtMinceraft1010X:
For `tasks.register(...)`, it's used for [task configuration avoidance](https://docs.gradle.org/6.9.2/userguide/task_configuration_avoidance.html). This prevents unnecessary tasks from being configured, which is best paired with #1631 - when only backend tests or backend checkstyle are run, configuration of frontend-related tasks like `installFrontend` is not carried out.

gerhean:
May I ask about the significance of including the `Apache License` block of text? It's a little weird that it `Copyright 2015 the original author or authors.` but nowhere states who these author or authors are.

dcshzj:
> May I ask about the significance of including the `Apache License` block of text? It's a little weird that it `Copyright 2015 the original author or authors.` but nowhere states who these author or authors are.

@gerhean I think the Apache License terms [requires that the copyright notice is retained](https://choosealicense.com/licenses/apache-2.0/). I think for the "the original author or authors", we should leave it as it is, in case of licensing issues that may occur.

gerhean:
Just want to check that this has been tested to work on WIndows?

chan-j-d:
> Just want to confirm that this has been tested to work on Windows? I assume it works on MacOs and Linux since the CI passed.

@gerhean Yep, I pulled the branch, `clean build`, `frontendTest` and ran it against a few repositories when I tested it.

gerhean (approved):
LGTM

fzdy1914 (approved):

PR #1666, opened 2022-02-14
Docs: fix table format in configFiles.md

yhtMinceraft1010X:
Proposed commit message:
```
IntelliJ shows a compatibility warning for tables in configFiles.md
saying that all rows should have pipe borders at the start and end.

Let's add the borders to resolve this.
```

This is similar to a [pull request for the CS3281 2022 repo](https://github.com/nus-cs3281/2022/pull/13).


docs/ug/configFiles.md
```diff
@@ -29,17 +29,17 @@ Given below are the details of the various config files used by RepoSense.
...
-Shallow Cloning | Enter **`yes`** to clone the repository using Git's shallow cloning functionality. This option can significantly reduce the time taken to clone large repositories. However, the option should ideally be disabled for smaller repositories where the `.git` file is smaller than 500 MB, as it would create overhead.
+| Column Name                           | Explanation                                                                                                                                                                                                                                                                                                            |
+|---------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
```
chan-j-d, line 33: Is there a difference if the padding is left out? I think it might be quite a lot of extra work to maintain the alignment if these tables were ever updated.

chan-j-d (commented):
Would like to know your thoughts about this.

docs/ug/configFiles.md
yhtMinceraft1010X: This padding separates the headers from the entries. I don't think it should be left out.

chan-j-d: My bad I didn't make myself clear. I believe that line itself is necessary for markdown tables but I was referring to the padding that results in the bars aligning. 

My Intellij's markdown addon does not actually show any warnings on the current master branch with single `|` separator per line. Perhaps someone else should comment on this.

chan-j-d: Ah I see. I thought I had set it to automatically notify me of updates. It seems Intellij can auto perform the alignment for us. In that case, the only downside I can see is if a new contributor attempts to realign it manually after making a change in the table.

chan-j-d (approved):
LGTM!

gerhean:
According to the [markdown guide](https://www.markdownguide.org/extended-syntax/#tables), the cell widths can vary, while the rendered output will look the same. So why pad the cells with space?

chan-j-d:
> According to the [markdown guide](https://www.markdownguide.org/extended-syntax/#tables), the cell widths can vary, while the rendered output will look the same. So why pad the cells with space?

It's a context action that Intellij supports which immediately does the padding and border `|` for us which was faster to do than adding the border `|` myself when I tried.

gerhean:
Is there a way to enforce these style guides without intellij? Cause intellij doesn't have free support for frontend development and I don't want to switch between multiple IDE to edit some docs.

chan-j-d:
> Is there a way to enforce these style guides without intellij? Cause intellij doesn't have free support for frontend development and I don't want to switch between multiple IDE to edit some docs.

I think we can choose one of two options. 
1. Align it as such for now but not necessary to enforce the alignment for future PRs
2. Add in only the bars at the start and end for this PR and not perform any padding.

I personally lean towards the latter but I don't really see much of a difference.

yhtMinceraft1010X:
> 1. Align it as such for now but not necessary to enforce the alignment for future PRs
> 2. Add in only the bars at the start and end for this PR and not perform any padding.
> 
> I personally lean towards the latter but I don't really see much of a difference.

Personally, I think for the auto-updating of tables in markdown, it's tedious to undo the padding. I lean towards option 1 in this case.

gerhean:
I would lean towards option 2 unless there is a way to enforce the padding using linters and such.

dcshzj (changes requested):
I personally would prefer option 2 as well. Also, does this also apply to the number of dashes used after the column header?

dcshzj:
@yhtMinceraft1010X What about the dashes after the column header? Should we standardize to use a certain number of dashes or follow some alignment?

yhtMinceraft1010X:
> @yhtMinceraft1010X What about the dashes after the column header? Should we standardize to use a certain number of dashes or follow some alignment?

For this one, I intend to add enough dashes such that they are aligned with the column headers.

dcshzj (approved):
LGTM. Will merge directly since it's just documentation changes.

PR #1670, opened 2022-02-15
Frontend: add deep watchers to arrays

Zhou-Jiahao-1998:
Needed for #1627 

```
watch is shallow by default in Vue 3: the callback will only trigger
when the watched property has been assigned a new value.

Let's use deep watcher to enable callback to fire on all nested
mutations.
```
Refer [here](https://vuejs.org/guide/essentials/watchers.html#deep-watchers) for more details.


gerhean:
Can you verify that these are all the watchers which need to be deep watched? What about this one?
https://github.com/reposense/RepoSense/blob/0db807d1c402c310b8b7589894aa7bd2f52b6daf/frontend/src/views/v-summary.vue#L191

Zhou-Jiahao-1998:
@gerhean `store.state.summaryDates` is an object, not an array. It can be found in `src/store/store.js`. I think all the necessary watchers are updated because in the compatible build, all the array watchers throws a warning and I updated all to remove all the warnings.

gerhean (approved):
LGTM

dcshzj (approved):
LGTM. Moving this PR ahead.

PR #1672, opened 2022-02-15
Frontend: combine v-for and v-if

Zhou-Jiahao-1998:
Part of #1627.

```
In Vue 2, when they exist on the same node, v-if has a lower priority
than v-for. However, this is the opposite in Vue 3.

Let's use the recommended filtered array to avoid ambiguity.
```

Details can be found [here](https://vuejs.org/guide/essentials/list.html#v-for-with-v-if).

### DIFF
--- a/frontend/src/components/v-ramp.vue
+++ b/frontend/src/components/v-ramp.vue
@@ -5,7 +5,7 @@
       a.ramp__slice(
         draggable="false",
         v-on:click="rampClick",
-        v-for="(commit, k) in slice.commitResults.filter(commitResult => commitResult.insertions>0)",
+        v-for="(commit, k) in slice.commitResults.filter(commitResult => commitResult.insertions > 0)",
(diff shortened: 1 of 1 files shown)


frontend/src/components/v-ramp.vue
```diff
@@ -21,8 +20,7 @@
   template(v-else)
     a.ramp__slice(
       draggable="false",
-      v-for="(slice, j) in user.commits",
-      v-if="slice.insertions > 0",
+      v-for="(slice, j) in user.commits.filter(commit => commit.insertions>0)",
```
gerhean, line 23: Maybe add a space around the `>`?

yhtMinceraft1010X (approved):
LGTM

gerhean (approved):
LGTM

fzdy1914 (approved):

PR #1675, opened 2022-02-16
Actions: use pull_request_target for delete deploy workflow

gok99:
**Proposed Commit Message:**
```
The delete deployments workflow merged in #1654 doesn't work for PRs 
from forked repos due to insufficient GitHub token permissions.

Additionally, the pending and surge workflows use fetch-depth: 0 when
checking out the repository, which unnecessarily fetches the full
history of the repository and cause the workflow to be slightly slower.

Let's use the pull_request_target workflow (instead of pull_request) 
which provides necessary permissions for the delete task, and give the
task a name. Let's also remove the unnecessary fetch-depth: 0 from the
pending and surge workflows.
```


dcshzj (approved):
LGTM. The `pull_request_target` event is required here because the `delete-deploy.sh` script requires write permissions to modify the deployment statuses (similar to `pending.yml` workflow).

This set up is safe, because only the `delete-deploy.sh` is run, and the version in the `master` branch is used. Code from external users should not be able to run in this context.

Will be merging directly as it breaks the delete deployments workflow.

PR #1677, opened 2022-02-16
Actions: pull deployment info specific to the PR for deletion

gok99:
**Proposed Commit Message**:
```
For the delete deployment workflow, we are pulling all deployments in
the repo and filtering by PR ID. Hence, it is likely that older 
deployments will not be included in the 30 that is returned by the
GitHub API by default.

Let's fix this by pulling only deployments specific to the environments
to be deleted, and increasing the limit from 30 to 100 (which is the
maximum), to ensure that all deployment IDs for a PR are correctly
pulled and deleted. 
```


config/gh-actions/delete-deploy.sh
```diff
@@ -62,8 +59,9 @@ mark_deployment_inactive() {
 }
 
 # Function to get deployment data about repo via a cURL command
+# $1: Deployment environment name
 get_deployment_data() {
-  curl "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/deployments" \
+  curl "https://api.github.com/repos/${REPO_OWNER}/${REPO_NAME}/deployments?per_page=100&environment=${1}" \
```
chan-j-d, line 64: Do you think there's a possibility 100 is not enough?

chan-j-d (approved):
The workflow looks correct and makes more sense. LGTM

gerhean:
Why fetch all the deployments? Isn't the PR id already in the deployment name?

chan-j-d:
> Why fetch all the deployments? Isn't the PR id already in the deployment name? I'm not very familiar with github api, but is it possible to filter by name with this?

I think most of the deployments have the same 'task' which is just 'deploy'. For example running the command 
`curl   -H "Accept: application/vnd.github.v3+json"   https://api.github.com/repos/reposense/RepoSense/deployments?environment=dashboard-1650` which essentially asks for a list of deployments for dashboard-1650 comes up with a pretty long list, each of which have different ids but all have the same task "deploy" as seen in this image:
image

So I think what @gok99 suggests is the way to go about it where the API only returns deployments relating to `dashboard-1650` for example.

However, while looking up on this, I actually found out about this section in GitHub's API under deployments where you can [delete an environment](https://docs.github.com/en/rest/reference/deployments#delete-an-environment). I wonder if it does the same thing as deleting it from the settings since many deployments are performed on the same environment and this would be far more convenient.

It also seems to be a [very new feature](https://github.blog/changelog/2020-12-15-github-actions-environments-environment-protection-rules-and-environment-secrets-beta/) since I could not find much information about how to use it.

gok99:
>  the API only returns deployments relating to dashboard-1650. However, the list is still really long which I think is why the suggestion of increasing the limit to 100 came up.

Yes, that's right! An environment could have multiple deployments. It is unlikely that it would have a 100, but I decided to pull the maximum possible amount since I didn't see a reason not to.

> you can [delete an environment](https://docs.github.com/en/rest/reference/deployments#delete-an-environment).

Interesting! I hadn't come across this endpoint. It would be great if this worked off the bat! Let me play around with this and update this PR if necessary.

chan-j-d:
> The zombie deployments remain in the PR and in the deployment history (2).

What happens if you try to access the deployment in the history which already had its environment deleted?
I'm wondering if it is even necessary to delete the deployments themselves since the initial goal was to clear out the excess environments.

gok99:
> What happens if you try to access the deployment in the history which already had its environment deleted?

There is no way to access older deployments directly from the entries in (2) as far as I can tell, so their only purpose is to act as a history of when deployments were made, which would roughly coincide with when pushes are made to a branch anyway - so I'm not sure how much value there is in keeping them. At the same time, there may not be much harm in keeping them either, other than clutter in (2).

I think we'd need to decide on whether there is value in keeping a history of deployments after a PR is closed (if historically there has been a need to do this).

chan-j-d:
I think at this point we should probably consult @dcshzj

dcshzj:
Personally I would prefer if we cleaned up as much as possible, as these deployment information are no longer useful once the PR has been merged/closed.

I am not too concerned with the deployments history accessed from the main page of this repository. I am more concerned with the deployments that happen during the journey of the pull request, such as this:

image

These deployment notifications add a lot of clutter, especially for PRs where the author made a lot of pushes. I understand that the only way to remove these is when the deployments itself are removed, not just the environment. If we can achieve a way to remove these (together with removing environments from the main page), then I think the intent of the original task is achieved.

chan-j-d:
Then I think @gok99 has already achieved that in the original PR! The only exception is if there are more than 100 separate deployment for a single PR's deployment.

dcshzj:
> Then I think @gok99 has already achieved that in the original PR! The only exception is if there are more than 100 separate deployment for a single PR's deployment.

Yup, he managed to remove the environments earlier. However, the earlier deployments still exists, and if you see some of the recently merged PRs, they do not show the latest deployment, but the earlier ones (abandoned) still do. So I think we still need to delete the older deployments.

I think regarding the 100 separate deployments, if it is too complicated to implement a way to get the next n number of deployments, then we probably can just stick to 100 for now. It is only once in a long while we get PRs that have more than 100 deployments.

gok99:
> if there are more than 100 separate deployment for a single PR's deployment.

Unfortunately this is a limitation of API. We can increase it further if GitHub increases this limit and it becomes a concern that > 100 PR updates is realistically possible before a merge (not sure if a PR has ever been in gestation for that long before).

dcshzj (approved):
LGTM! I think we can stick with 100 for now, should be able to cover most cases.

PR #1678, opened 2022-02-19
Frontend: add Object.assign() to hash assignment

Zhou-Jiahao-1998:
Needed for #1627 

```
The current implementation will result in some race condition where some
data are read and written at the same time. This is an issue in Vue 3 as
it does not wait for current function to finish before running another
async function.

Let's avoid this race condition by using Object.assign().
```


gok99:
@Zhou-Jiahao-1998 I'm not too familiar with this section of the frontend unfortunately. Are you able to provide more details about how the race condition arises and how  `Object.assign` fixes it?

Zhou-Jiahao-1998:
@gok99 In Javascript, Objects are reference type. Therefore, if assign using `=`, when the original object mutates, the value of `hash` in this cases also changes. While I am not entirely sure of the exact cause of the issue, I can only say that Vue 3 does a lot of optimisation internally. To improve speed, Vue may run certain parts concurrently, resulting in mismatch read and write value here. When using `Object.assign()`, a new copy is returned. Therefore, we do not have to worry about the source mutating in this case. I hope this explains.

gerhean:
Pinging @gok99 for review.

gerhean:
> hash seems to be a local variable in this method. Wouldn't parallel calls to renderFilterHash just create new variables which won't interfere with one other?

The problem is not parallel calls to `renderFilterHash`. Rather, it is because reassigning one of the `this.something` caused some watcher to fire, which caused `setSummaryHash` to be called while `renderFilterHash` is executing.

The current solution is actually not very good, but it is sufficient to get Vue 3 to work. The ideal solution is to prevent 
https://github.com/reposense/RepoSense/blob/a1c25f8506fd086ed450c8abd49326b32b8c9a06/frontend/src/views/v-summary.vue#L408
from being called multiple times while `renderFilterHash` is executing. I am personally not very familiar which watchers are being called.

gok99 (approved):
> The current solution is actually not very good, but it is sufficient to get Vue 3 to work.

I suppose we could clean this up in a separate issue after the Vue 3 migration.

gerhean (approved):
LGTM, better fixes can and should be done after upgrade to Vue 3.

gerhean:
Will be moving this PR ahead as it is a minor change.

PR #1679, opened 2022-02-20
[#231] Modify local repository URL to the correct remote repository

### LINKED ISSUE #231: Clicking on ramps in ramp chart generated using local repos redirects to non-existent link

gok99:
Implements the suggestion in https://github.com/reposense/RepoSense/pull/717#issuecomment-497537302:
- For local repositories, attempt to find remote URL with `git remote`
- If no remote is found, disable links to the remote repository for commit view and blame view

Fixes #231 

Proposed Commit Message: 
```
Local repositories cloned from a remote or with remotes added manually
do not have working links to a remote repository, even though remote
repository information can be obtained with `git remote -v`.

Let's create a GitRemote class to obtain available remote repository 
information for local repositories, and use it for generating valid
links in a report.
```


chan-j-d:
Just to record and follow up on what was discussed earlier today:

I believe a way to support this on top of PR #1644 would be to use the `RepoLocation::getRemoteRepoNameAndOrg` method but it would require changing the access modifier. 

Beside that, you might need to add separate fields to `RepoLocation`, something like `remoteRepoName` and `remoteOrg`. For remote repositories, the `remoteRepoName` and `repoName` will be the same and similarly for `remoteOrg`. For local repositories, they might be different as the output directory of the cloned local repo might have been changed. 

This is mostly related to #1637 where if the `repoName` and `organization` are the same for two directories, then their output directories generated from those two fields will be the same and the final report will override each other.

gok99:
> Just to record and follow up on what was discussed earlier today:
> 
> I believe a way to support this on top of PR https://github.com/reposense/RepoSense/pull/1644 would be to use the RepoLocation::getRemoteRepoNameAndOrg method but it would require changing the access modifier.
> 
> Beside that, you might need to add separate fields to RepoLocation, something like remoteRepoName and remoteOrg. For remote repositories, the remoteRepoName and repoName will be the same and similarly for remoteOrg. For local repositories, they might be different as the output directory of the cloned local repo might have been changed. Then you can edit the getRepoLink in api.js to use remoteRepoName and remoteOrg instead.
> 
> This is mostly related to https://github.com/reposense/RepoSense/issues/1637 where if the repoName and organization are the same for two directories, then their output directories generated from those two fields will be the same and the final report will override each other.

I am currently working around this by adding transient fields specifically for local repositories, which store what repoName and repoOrg used to and get used for generating output directories. If a remote repository exists, then repoName and repoOrg (which get serialized and used for generating links) are obtained from the remote repository from GitRemote.

src/main/java/reposense/git/GitRemote.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import reposense.system.CommandRunner;
+
+/**
+ * Contains git remote related functionality.
+ * Git remote is responsible for getting remote repository information.
+ */
+public class GitRemote {
```
chan-j-d, line 14: Not directly related to this PR but this seems like an extremely strange line for codecov to flag as not covered.

src/main/java/reposense/git/GitRemote.java
```diff
@@ -0,0 +1,43 @@
...
+     * Extracts remote repository information. Returns a map
+     * with keys of the form REMOTE_NAME(fetch) or REMOTE_NAME(push).
+     */
+    public static Map<String, String> getRemotes(String repoRoot) {
+        Map<String, String> remotes = new HashMap<>();
+        String result;
+        try {
+            result = CommandRunner.runCommand(Paths.get(repoRoot), "git remote -v");
+        } catch (RuntimeException re) {
+            return remotes;
```
chan-j-d, line 26: Do you think some kind of logged message is necessary here? Something like 'no remote repository was found'

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -61,11 +68,27 @@ public RepoLocation(String location) throws InvalidLocationException {
...
+            String[] localRepoNameAndOrg = getLocalRepoNameAndOrg(location);
+            this.localRepoName = localRepoNameAndOrg[0];
+            this.localOrganization = localRepoNameAndOrg[1];
+
+            Map<String, String> remotes = GitRemote.getRemotes(location);
+            String newLocation = remotes.size() == 0
+                    ? location
+                    : remotes.containsKey("origin(fetch)")
+                    // Get fetch remote named 'origin' if possible
+                    ? remotes.get("origin(fetch)")
```
chan-j-d, line 85: I think for lines 83 and 85, the constant `origin(fetch)` should be stored in the `GitRemote` class. Either that or create some kind of static accessor method in `GitRemote` that uses this key instead.

src/main/java/reposense/git/GitRemote.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Contains git remote related functionality.
+ * Git remote is responsible for getting remote repository information.
+ */
+public class GitRemote {
+
+    /**
+     * Extracts remote repository information. Returns a map
+     * with keys of the form REMOTE_NAME(fetch) or REMOTE_NAME(push).
```
chan-j-d, line 18: This comment is chronologically after the comment about lines 83 and 85. Are these key names meant to be stored as constants? Because the one in the javadoc does not seem to match with the one used in lines 83 and 85.

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -80,10 +103,16 @@ public boolean isEmpty() {
...
+        if (this.isLocal && !isEmpty()) {
+            return localRepoName;
+        }
         return repoName;
     }
 
     public String getOrganization() {
+        if (this.isLocal && !isEmpty()) {
+            return localOrganization;
+        }
```
chan-j-d, line 115: I was thinking that instead of storing `transient` fields `localRepoName` and `localOrganization`, we instead store `repoName` and `organization` as transient fields. Then change `localRepoName` and `localOrganization` to `remoteRepoName` and `remoteOrganization` which will be the ones uploaded via the JSON file. This way, these 2 methods do not need the `if-else` and `isLocal` field since we can always just return `repoName` and `organization`.

chan-j-d (changes requested):
I've tested the enhancement a bit and it seems to work well. Some minor comments about variable usage and javadocs.

src/main/java/reposense/model/RepoLocation.java
gok99: That's a good idea! Thanks for the suggestion.

src/main/java/reposense/git/GitRemote.java
gok99: I initially did have a log message here actually. I removed it because this catch will generally only hit when the directory isn't even a git repository. It seemed a bit strange to be logging that here, since the report generation will have to fail eventually anyway.

src/main/java/reposense/git/GitRemote.java
gok99: Oh I assumed that adding the tests would have gotten rid of this. I'm actually not sure why this is flagged.

src/main/java/reposense/git/GitRemote.java
gok99: REMOTE_NAME is meant to be substituted with the name of the remote. Lines 83 and 85 check for the remote 'origin' first if it exists. It could be possible that the repository has other remotes, Am open to suggestions for making the substitution more obvious.

src/main/java/reposense/model/RepoLocation.java
gok99: That's probably a good idea. I was trying my best not to have to change anything outside of this class, but your suggestion seems cleaner overall.

src/main/java/reposense/git/GitRemote.java
chan-j-d: Because right now, if the command is run against a purely local repository as opposed to a non-git repository, the outcomes are the same and there is no way to distinguish. I thought it might be helpful to add in the event that down the line, some relative pathing issue results in the command being run in the wrong directory altogether. Or a fully remote repository being mistaken for a non-git directory.

src/main/java/reposense/git/GitRemote.java
chan-j-d: oh I understand it now. I think I understand the PR a lot better now. Then I believe there might be a second point of failure for repos where the user renamed the `origin` reference though I guess that is quite a fringe case and the behaviour of reposense would still be acceptable.

src/main/java/reposense/model/RepoLocation.java
chan-j-d: Oops yea. I believe you might need to change the key references in `api.js` as well.

src/main/java/reposense/git/GitRemote.java
gok99: Yes, `git remote -v` will still succeed in a purely local git repository. The closest log message to the truth that i can think of is something like: `Unable to run git remote command in directory: %s`. Do you have any suggestions for improving this?

src/main/java/reposense/model/RepoLocation.java
gok99: What do you think of the current solution?

src/main/java/reposense/git/GitRemote.java
gok99: Yes, if `origin` doesn't exist, we currently get whatever is available.

src/main/java/reposense/git/GitRemote.java
chan-j-d: I think that is good enough. I think this should be mostly for defensive programming since I don't expect that line to ever be reached given proper execution. But we never know 🤷

src/main/java/reposense/model/RepoLocation.java
chan-j-d: I think it looks good! I'll approve the PR once you've added the logger warning.

chan-j-d (approved):
LGTM!

src/main/java/reposense/git/GitRemote.java
```diff
@@ -0,0 +1,70 @@
...
+     */
+    public static Optional<String> getAvailableRemoteLocation(Map<String, String> remoteMap) {
+        String remoteLocation = remoteMap.size() == 0
+                ? null
+                : remoteMap.containsKey(DEFAULT_FETCH_REMOTE)
+                // Get default fetch remote if possible
+                ? remoteMap.get(DEFAULT_FETCH_REMOTE)
+                // Get any remote otherwise
+                : remoteMap.values().iterator().next();
+        return Optional.ofNullable(remoteLocation);
```
dcshzj, line 67: This part feels a little hard to read. May I suggest first checking if `remoteMap.size() == 0`, then return `Optional.empty()`? After that, we can continue with the `remoteLocation` part with the `remoteMap.containsKey(DEFAULT_FETCH_REMOTE)` check.

Something along these lines:
```java
if (remoteMap.size() == 0) {
    return Optional.empty();
}

String remoteLocation = remoteMap.containsKey(DEFAULT_FETCH_REMOTE)
...
```

src/main/java/reposense/model/RepoLocation.java
```diff
@@ -61,30 +67,43 @@ public RepoLocation(String location) throws InvalidLocationException {
...
-        String[] repoNameAndOrg;
+        String[] remoteRepoNameAndOrg;
+        String[] outputFolderRepoNameAndOrg;
         if (location.isEmpty()) {
-            repoNameAndOrg = new String[] {"", "", UNSUPPORTED_DOMAIN_NAME};
+            remoteRepoNameAndOrg = new String[] {"", "", UNSUPPORTED_DOMAIN_NAME};
+            outputFolderRepoNameAndOrg = remoteRepoNameAndOrg;
         } else if (isLocalRepo(location)) {
-            repoNameAndOrg = getLocalRepoNameAndOrg(location);
+            outputFolderRepoNameAndOrg = getLocalRepoNameAndOrg(location);
```
dcshzj, line 76: Nothing to do with this line in particular, but are you also able to update the Javadoc for this `getLocalRepoNameAndOrg()` function? It is still referring to the array returned as length 2. Same for the `getRemoteRepoNameAndOrg()` function.

src/test/java/reposense/git/GitRemoteTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import reposense.template.GitTestTemplate;
+
+class GitRemoteTest extends GitTestTemplate {
+
+    @Test
+    void getRemotesTest() {
```
dcshzj, line 13: The name of the function should follow something like the format `featureUnderTest_testScenario_expectedBehavior()`. I think if possible, it would be nice to test `getRemotes()` and `getAvailableRemoteLocation()` separately.

dcshzj (changes requested):

gok99:
@dcshzj Updated with your suggestions changes

src/test/java/reposense/git/GitRemoteTest.java
```diff
@@ -0,0 +1,66 @@
...
+        Map<String, String> remotes = new HashMap<>();
+        remotes.put(NON_ORIGIN_FETCH_REMOTE_NAME_1, REMOTE_URL_NON_ORIGIN);
+        remotes.put(NON_ORIGIN_PUSH_REMOTE_NAME_1, REMOTE_URL_NON_ORIGIN);
+        remotes.put(NON_ORIGIN_FETCH_REMOTE_NAME_2, REMOTE_URL_ORIGIN);
+        remotes.put(NON_ORIGIN_PUSH_REMOTE_NAME_2, REMOTE_URL_ORIGIN);
+
+        String remoteName = GitRemote.getAvailableRemoteLocation(remotes).orElse("");
+        Assertions.assertTrue(remotes.containsValue(remoteName));
+    }
+
```
dcshzj, line 65:
```suggestion
```

Extra line here.

dcshzj (changes requested):

dcshzj (approved):
LGTM!

dcshzj:
Putting the merge on hold first, as there seems to be some issues with the building of the preview dashboards. @yhtMinceraft1010X Is there a chance that it is due to the Codecov action?

yhtMinceraft1010X:
This is most likely the issue causing our surge deployments to fail: https://github.com/dawidd6/action-download-artifact/issues/147

dcshzj:
> This is most likely the issue causing our surge deployments to fail: [dawidd6/action-download-artifact#147](https://github.com/dawidd6/action-download-artifact/issues/147)

I see, thanks for investigating! I will probably hold the merge train until tomorrow, hopefully the GitHub API issue resolves by then.

yhtMinceraft1010X:
An update on the issue: https://github.com/dawidd6/action-download-artifact/issues/147#issuecomment-1096685055

PR #1685, opened 2022-02-21
[#1682] Fix space in local repo causing cloning exception

### LINKED ISSUE #1682: Space in local repo final directory causes cloning exception

* **OS and Version:**

1. Clone any repository to a local directory containing a space. E.g. `git clone https://github.com/reposense/RepoSense "output directory"
1. Run reposense on this directory. `java -jar RepoSense.jar --repos "output directory"`

Analysis is done without issue.

Analysis fails at cloning.
image

The repo directory name used here is slightly different but serves the same purpose to include a space.

The issue is likely due to the clone directory being quoted when running git clone commands but the output directory is not.

[log.txt](https://github.com/reposense/RepoSense/files/8106241/log.txt)

chan-j-d:
Fixes #1682 

Currently, clone output paths are not quoted. The issue is rare as it can only occur against local repositories where the final directory contains a space. 

This PR also includes changes to the existing `GitClone::cloneFromBareAndUpdateBranch` method. It used to be the only method that constructs its own git command without the use of a helper private method which is likely why it was missed and none of its directories are quoted. This change aims to standardize the format between different git clone methods with
1. The method handling the logic of setting up and running the command has to create its own 'command constructor' method to create the String. This way, it is easy to standardize changes to the command formats by looking at these methods without having to see what they are used for.
2. Defaulting the directory in which the commands are run to the current directory (location the user invokes `java -jar RepoSense.jar`). As of this PR, the `cloneFromBareAndUpdateBranch` is the only one that does not run the command from the current directory. This is the result of https://github.com/reposense/RepoSense/pull/1562. Before that, commands were run from the cloned repo directory but the commands themselves were constructed as if they were in the working directory. This caused issues when relative pathing was used. The fix changed all but one of the use of `git clone` to start from the working directory. Only the `cloneFromBareAndUpdateBranch` method was not changed as it seemed to break other things. 

Commit message:
```
Output directories are not quoted which can lead to errors. The 
GitClone::cloneFromBareAndUpdateBranch method is also not standardized
to match other clone methods which always run from the working
directory. This can be difficult to maintain.

Let’s fix the issue of spaces in repo directory causing clone failure
and standardize the cloneFromBareAndUpdateBranch method.
```


src/main/java/reposense/git/GitClone.java
```diff
@@ -55,7 +55,7 @@ public static CommandRunnerProcess cloneShallowBareAsync(RepoConfiguration confi
             String outputFolderName, LocalDateTime sinceDate) throws GitCloneException {
         try {
             return CommandRunner.runCommandAsync(rootPath,
-                                                    getCloneShallowBareCommand(config, outputFolderName, sinceDate));
+                    getCloneShallowBareCommand(config, outputFolderName, sinceDate));
```
gok99, line 58: The tests for GitClone seems lacking compared to the other Git classes. Perhaps we could figure out a way to test the various clone variations.

chan-j-d: In a strange coincidence, I did voice the same thought some time back and the response is here https://github.com/reposense/RepoSense/pull/1562#issuecomment-877829751 under point 2.

I think in general, cloning is quite expensive so they can be done together with the system tests. If the system tests work, then the cloning likely works with it. There is an argument to be made for `cloneFromBareAndUpdateBranch` which is a local-to-local clone done while analyzing but I'm not sure how we would go about testing it.

src/main/java/reposense/git/GitClone.java
```diff
@@ -160,7 +160,7 @@ public static void cloneBare(RepoConfiguration config, Path rootPath,
      */
     public static void cloneFromBareAndUpdateBranch(Path rootPath, RepoConfiguration config)
             throws GitCloneException, IOException {
-        Path relativePath = rootPath.relativize(FileUtil.getBareRepoPath(config));
+        Path relativePath = FileUtil.getBareRepoPath(config);
```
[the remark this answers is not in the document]
chan-j-d: The way it works at the moment is somewhat strange. Suppose our working directory is `/directory/`, then the `rootPath` here (in the only instance where this method is used) is in `/directory/repos/` where `repos` is a hard-coded constant designated for where the cloned repos go. 

`FileUtil::getBareRepoPath` would construct something like `repos/reposense/reposense_bare`. Code below:
https://github.com/reposense/RepoSense/blob/c6564331d2888a20ca1276cc71ab1cb897c2a4ae/src/main/java/reposense/util/FileUtil.java#L254-L257

So the `relativize` ends up trying to relativize the paths `/directory/repos/` and `repos/reposense/reposense_bare` which settles on `/directory/repos/reposense/reposense_bare`. 

The `ReportGenerator` class was changed together with the above code and now runs from current directory, `/directory/` and then adds on the entire bare repo path to the back. Functionally, it is the same.

src/main/java/reposense/git/GitClone.java
gok99: I think it makes to test these with the system tests, since the issues seem to be with system-specific quirks (although I'm not sure how much of GitClone is covered by the systemtests). I think the linked comment is probably right about tests in that we can only go to the extent of ensuring that the methods return the correct Strings and throw the right exceptions.

gok99 (approved):
LGTM!

src/main/java/reposense/git/GitClone.java
```diff
@@ -199,10 +198,27 @@ private static String getCloneCommand(RepoConfiguration config, String outputFol
...
+     * branch to the designated branch in {@code config}.
+     *
+     * @param repoPath Location of repo.
+     * @param config Config of the repo to be analyzed.
+     * @param outputFolderName Output directory for the cloned repo.
+     * @return Command to be used.
+     */
+    private static String getCloneBareAndBranchCommand(Path repoPath, RepoConfiguration config,
+            String outputFolderName) {
+        return "git clone "
```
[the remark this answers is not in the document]
chan-j-d: I'm not too sure if this is necessary at this point unless we're anticipating that `git` changes the `clone` command to a different word. As it is currently standardized with other `git` command classes, I think it'll be somewhat strange if only this class has `git clone` itself abstracted out. And if we're intending to change it for all git command classes, I personally think it is more work than would be worth it.

dcshzj (approved):
LGTM!

Issue #1427, opened 2021-01-23
Some PlantUML diagrams will not render in deployed RepoSense website 


rolandyuwy:
**Tell us about your environment**

* **RepoSense Version:** master
* **OS and Version:** windows10
* **Web Browser and Version (if applicable):** Chrome

**Please include the steps to reproduce the bug.**

1. Visit https://docs-1406-pr-reposense-reposense.surge.sh/dg/report.html for the surge preview of #1406, which replaces images in the Developer Guide with PlantUML diagrams, to incorporate MarkBind's native support for PlantUML and improve scalability. 
2. Scroll to `Report architecture` section to view the PlantUML diagram.
3. Scroll to `Summary view` and `Authorship view` to view the other PlantUML diagrams.

**What was expected to happen?**
The PlantUML component diagram in the `Report architecture` section should be rendered as such:
![image](https://user-images.githubusercontent.com/59356158/105571137-7da78c80-5d88-11eb-85d1-c7f22889988a.png)

**What actually happened? Please include a screenshot of the output.**
The PlantUML component diagram in the `Report architecture` section is not rendered, with an error output shown. Other PlantUML diagrams are rendered.
![image](https://user-images.githubusercontent.com/59356158/105571127-5355cf00-5d88-11eb-8699-6b67cd705b61.png)

**If possible, include the URL to your RepoSense report or log files (if any).**

**If possible, describe the solution.**
- The error output indicated that Graphvis is uninstalled in the deployed website 
- The root issue of PlantUML diagrams not rendering is not solved in this issue: https://github.com/MarkBind/markbind/issues/916
- A possible solution is to include the following lines in the build configuration file (.travis.yml) to install Graphviz on Linux for in the deployment repository:

```
before_install:
  - sudo apt-get -y install graphviz

damithc:
I'm not even sure PlantUML diagrams will work with surge/netlify previews. You may want to ask the MarkBind guys (e.g., file a bug report or query in their issue tracker)

yhtMinceraft1010X:
Might be relevant for us: https://github.com/MarkBind/markbind/issues/1903#issuecomment-1098778682

PR #1686, opened 2022-02-21
Standardise PUML usage

yhtMinceraft1010X:
Proposed Commit Message:
```
Overall report architecture image is a static image while other
PUML diagrams are rendered using MarkBind.

Let's change the report architecture image to use MarkBind
PUML rendering.
```


yhtMinceraft1010X:
PR closed bcos the bug from #1427 is still present

PR #1688, opened 2022-02-24
Docs: update ternary operator in the style guide

yhtMinceraft1010X:
Proposed Commit Message:
```
The ternary style guide in the documentation refers to the Date class,
which has been replaced with LocalDateTime in the code.

Let's update the style guide example accordingly.
```


damithc:
Guys, a side note: 
Try not to create lengthy style guides as this becomes an additional maintenance burden. We don't want every NUS-OSS project to have their own lengthy style guide as it's going to be a waste of effort and will make it harder for students to work in multiple projects.

It's best to follow existing style guides as a base (e.g., https://se-education.org/guides/) and point to other more comprehensive guide e.g., those provided by Google) as a fallback when the base doesn't cover something.

chan-j-d (commented):
I'm not entirely sure if it is necessary to follow the similar line of code in the codebase itself. I personally think the original example is sufficient to illustrate the idea of ternary operators and I would like to know your thoughts on this.

docs/dg/styleGuides.md
```diff
@@ -21,15 +21,18 @@ Our coding standards are mostly based on those at [se-education.org/guides](http
...
 }
+return min;
 ```
 
 The result would look something like this:
 ```
-Date min = commitInfos.isEmpty() ? new Date(Long.MIN_VALUE) : commitInfos.get(0).getTime();
+return (commitInfos.isEmpty())
+       ? ARBITRARY_FIRST_COMMIT_DATE_UTC.withZoneSameInstant(zoneId).toLocalDateTime()
+       : commitInfos.get(0).getTime();
```
dcshzj, line 35: I think this should be 8 spaces?

dcshzj (changes requested):
I think it should be okay to update the Date to LocalDateTime to match the codebase, in case of curious people who want to find this line of code in the codebase. It is not a hard requirement however, so it should not add to the maintenance burden.

chan-j-d (approved):
LGTM!

dcshzj (approved):
LGTM! Will merge since it's just documentation changes.

PR #1690, opened 2022-02-27
[#1655] Remove duplicated method in GitRevList

### LINKED ISSUE #1655: Duplicated method in GitRevList.java
**Currently**

`getCommitHashBeforeDate` and `getCommitHashUntilDate` in `GitRevList` have identical logic. This can be confusing for readers who are deciding between the methods to use. It is also easy to forget to update one of the methods because it's not immediately obvious that they do the same thing.

https://github.com/reposense/RepoSense/blob/728a9f9c14f706ace6873cae8696647b91ff3ffb/src/main/java/reposense/git/GitRevList.java#L19-L47

https://github.com/reposense/RepoSense/blob/728a9f9c14f706ace6873cae8696647b91ff3ffb/src/main/java/reposense/git/GitUtil.java#L27-L28

**Proposed Suggestion**

If there is no reason to keep the duplicated methods, perhaps we could remove one of them. `getCommitHashBeforeDate` seems to be a better description of what the methods are doing.

gok99:
Fixed #1655 

**Proposed Commit Message**:
```
The methods `getCommitHashBeforeDate` and `getCommitHashUntilDate` in
GitRevList.java have identical logic.

Let's remove `getCommitHashBeforeDate` and replace the `--before` flag 
in `getCommitHashUntilDate` with `--until` to match with the method 
name.
```


chan-j-d (approved):
LGTM!

dcshzj (approved):
LGTM!

dcshzj:
Will move this PR ahead since it is just code cleanup.

PR #1691, opened 2022-02-27
[#1659] Separate backend and frontend responsibilites

### LINKED ISSUE #1659: Backend tampering with frontend build files
**Currently:**

In ReportGenerator.java, the backend makes edits to the content of index.html generated by the frontend (for changing the title of the generated report). This breaks the separation of responsibility of the files generated by the frontend and backend. Even if only the title is being modified, there can arise hard-to-catch bugs like the one fixed in #1656. `index.html` is also not included in systemtests despite being modified by the backend, which is likely why the issue slipped through.

**Possible Solution:**

The value of `reportTitle` is currently stored in `summary.json` generated by the backend. The report title can be dynamically modified by the frontend after loading `summary.json`, so that the backend will not have to touch `index.html`. This is likely to cause a negligible but noticeable delay before the title of the report is corrected (since the title in `index.html` is currently manually corrected by the backend, this delay doesn't exist). An ideal solution will maintain separation of responsibility while also loading the title instantly.

gok99:
Fixed #1659 

**Proposed Commit Message**:
```
The backend modifies files generated by the frontend. This breaks the
separation of responsibilities for generated report files, which makes 
testing more challenging.

Let's stop the backend manually modifying report titles in frontend
generated files, and have the frontend update the report title itself. 
The default report title generated by the frontend is now also 
'RepoSense Report' instead of 'reposense' so that there is no delay in
updating the report title in the default case. The backend will proceed
to generate report files in the absence of `templateZip.zip`. Let's
also add a test to check that report titles are correctly updated.
```


src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -153,26 +151,15 @@
...
-     */
-    private static void setLandingPageTitle(String filePath, String pageTitle) throws IOException {
-        Path indexPagePath = Paths.get(filePath, INDEX_PAGE_TEMPLATE);
-        String line = new String(Files.readAllBytes(indexPagePath));
-        String newLine = line.replaceAll(INDEX_PAGE_DEFAULT_TITLE, "<title>" + escapeHtml4(pageTitle) + "</title>");
-        Files.write(indexPagePath, newLine.getBytes());
+        if (is != null) {
+            FileUtil.copyTemplate(is, outputPath);
+        } else {
+            logger.info(MESSAGE_MISSING_TEMPLATE);
```
yhtMinceraft1010X, line 161: Since `templateZip.zip` is normally needed for the frontend to work, `MESSAGE_MISSING_TEMPLATE` should be indicative of a problem. I recommend changing `logger.info` to `logger.warning`.

gok99: Thanks for the suggestion!

yhtMinceraft1010X (approved):
LGTM

gerhean:
Is it possible to change the default title to `RepoSense Report` as well? I think in `package.json` but I am not very sure. Also this does not completely fix #1659 as `build.gradle` still copies `template.zip` into the `src/main/resources` folder.

gok99:
> Is it possible to change the default title to RepoSense Report as well? I think in package.json but I am not very sure.

I think that's a good idea! That would completely eradicate the (already small) delay in updating the report title in the default case.

> Also this does not completely fix https://github.com/reposense/RepoSense/issues/1659 as build.gradle still copies template.zip into the src/main/resources folder.

Is there a good way around this? In this case, the backend continues to be responsible for putting the report together, and will use template.zip (without tampering with the contents therein) if it exists, but is able to proceed with report generation without it.

gerhean:
> Is there a good way around this? In this case, the backend continues to be responsible for putting the report together, and will use template.zip (without tampering with the contents therein) if it exists, but is able to proceed with report generation without it.

Get `build.gradle` to dump the frontend build directly into the output folder perhaps?

gerhean:
Additionally, do you think that the title for the testing repo should be different for testing? So that we can test that the title can actually be changed properly.

yhtMinceraft1010X:
> Is it possible to change the default title to `RepoSense Report` as well? I think in `package.json` but I am not very sure.
> 
> Also this does not completely fix #1659 as `build.gradle` still copies `template.zip` into the `src/main/resources` folder.

For this one, I intend to fix this in #1631

gok99:
> Is it possible to change the default title to RepoSense Report as well? I think in package.json but I am not very sure.

It seems spaces are not allowed in `package.json` since it is a non-URL safe character.

> Get build.gradle to dump the frontend build directly into the output folder perhaps?

I'll leave @yhtMinceraft1010X to do this as part of #1631, although I suspect this is not possible since the Jar file will ultimately need template.zip to create the report.

gerhean:
> It seems spaces are not allowed in the package.json name since it is a non-URL safe character.

Maybe choose one of the ways to change the default title as described [here](https://stackoverflow.com/a/67846199). I don't mind which one but giving a small justification would be good.

gok99:
Thanks for the suggestions! I chose option 3:

```
module.exports = {
  pages: {
    index: {
      // entry for the page
      entry: 'src/main.js',
      title: 'RepoSense Report',
    },
  }
}
```
since RepoSense is a single page application and we already have `vue.config.js`. Option 1 of hard-coding the title in index.js didn't seem like an ideal solution.

gerhean:
Pinging @yhtMinceraft1010X for review.

yhtMinceraft1010X (approved):
LGTM

gerhean (approved):
LGTM

fzdy1914 (approved):

PR #1692, opened 2022-02-28
Housekeeping - Fix indentations

### LINKED ISSUE #1717: Question on indentation for throws statements
While working on #1692, I noticed in the [Java coding standard](https://se-education.org/guides/conventions/java/index.html#comments) Javadoc example that the `throws` statement in the method header has an indentation of 4. However, I could not find a specific rule for `throws` line-wrapping. The closest rule I can think of is the [general 8 spaces rule](https://se-education.org/guides/conventions/java/index.html#layout). 

Question: Should we follow 8 spaces or 4 spaces for line-wrapping `throws` statements?

yhtMinceraft1010X:
Proposed commit message:
```
Some indentations are unnecessary or misaligned. There are also
additional whitespaces in some parts of the code.

This is due to single whitespace check in Checkstyle not being
implemented as well as our current Checkstyle config allowing for
more whitespaces than the prescribed value for line-wrapping
indentations.

Let's correct the coding standard errors and enhance the Checkstyle
config such that additional whitespaces and longer indents are
properly flagged.
```

Fixes #1717


gerhean:
Instead of doing this manually, can it be enforced using a linter? I am personally not a fan of style changes which is not enforced since different people have different opinions for what is the 'right style'.

yhtMinceraft1010X:
For the indentations, its difficult to use Checkstyle. The current check is designed such that for line wrapping, more than 8 whitespaces is okay. That's why the indentations for the method parameters could pass Checkstyle.

Using [`forceStrictCondition`](https://checkstyle.sourceforge.io/config_misc.html#Indentation) is not an option because this would cause code like the one below to fail even though the additional indentations enhance readability:
```
Gson gson = new GsonBuilder()
                .registerTypeAdapter(LocalDateTime.class, (JsonSerializer<LocalDateTime>) (date, typeOfSrc, context)
                        -> new JsonPrimitive(date.format(DateTimeFormatter.ofPattern(GITHUB_API_DATE_FORMAT))))
                // The line above is the failing line.
                .registerTypeAdapter(FileType.class, new FileType.FileTypeSerializer())
                .create();
```

However, I did find the [`SingleSpaceSeparator`](https://checkstyle.sourceforge.io/config_whitespace.html#SingleSpaceSeparator) check to be useful for finding unwanted whitespaces. It works with Checkstyle 8.1 and should still work for Checkstyle 9.3 in PR #1652

gerhean:
> Using [forceStrictCondition](https://checkstyle.sourceforge.io/config_misc.html#Indentation) is not an option because this would cause code like the one below to fail even though the additional indentations enhance readability:

Is it possible to move the lambda functions out and turn them into proper functions then?

Alternatively would it be okay to explicitly suppress warnings only for a few lines of code? Something like shown [here](https://stackoverflow.com/questions/4023185/disable-a-particular-checkstyle-rule-for-a-particular-line-of-code).

gerhean:
Actually I'm a little confused, so let's start by going through what exactly you mean by `Some indentations are unnecessary or misaligned`, and lets try to find some checkstyle rules which fit.

src/main/java/reposense/RepoSense.java
```diff
@@ -114,8 +114,8 @@ public static void main(String[] args) {
      * @throws IOException if user-supplied csv file does not exists or is not readable.
      * @throws InvalidCsvException if user-supplied repo-config csv is malformed.
      */
-    public static List<RepoConfiguration> getRepoConfigurations(ConfigCliArguments cliArguments)
-            throws IOException, InvalidCsvException, InvalidHeaderException {
+    public static List<RepoConfiguration> getRepoConfigurations(ConfigCliArguments cliArguments) throws IOException,
+            InvalidCsvException, InvalidHeaderException {
```
gerhean, line 118: For example, I am not sure how this change makes the code clearer.

dcshzj:
Looks like we don't have any options other than to force strict condition. @reposense/active-developer any reviews/thoughts on this?

gerhean:
Is it possible to include in the commit message some of the new rules which are enforced?

yhtMinceraft1010X:
Commit message has been edited

yhtMinceraft1010X:
Temporarily closing PR since GitHub Actions is stuck at `Expected - Waiting for status to be reported`.

yhtMinceraft1010X:
Re-opening PR

chan-j-d:
image

Personally, I'm not a fan of the indentation that is being enforced here. I don't think the burden of maintenance outweighs enforcing quite a different indentation standard from what we're used to. The tradeoff is that I have to now scan each line to see if they belong to the 'overall parent line' (the `addArgument`) or belong to a 'sub parent line' `.help(String...)`. Especially since I think that has been how most new contributors would have learnt to indent their code before coming here. Like [Google's checks](https://github.com/search?q=path%3Asrc%2Fmain%2Fresources+filename%3Agoogle_checks.xml+repo%3Acheckstyle%2Fcheckstyle+Indentation) taken from the example from [checkstyle's documentation](https://checkstyle.sourceforge.io/config_misc.html#Indentation_Examples) doesn't use force strict condition either. 

Side note:
The issue is apparently a pretty old bug with `forceStrictCondition`. I found this https://github.com/checkstyle/checkstyle/issues/5448 from some time back along with some more recent similar issues and PR.

yhtMinceraft1010X:
For some context, the reason why I put `String.format` is because of a weird bug where the + symbol line gets flagged as `method call indentation level is 16, expected should be 20`, but if I were to move the line by 4 spaces, it then becomes `+ indentation level is 20, expected should be 16`, so either way, the check will fail. `String.format` seems to allow the + symbol to be put at the same indentation level as `.help`.

Also, I worked on this PR with Checkstyle 8.1 initially. I think that after upgrading to Checkstyle 9.3, some checks got changed.

yhtMinceraft1010X:
Closing and re-opening PR due to `Expected - Waiting for status to be reported`.

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -162,34 +162,34 @@ private static ArgumentParser getArgumentParser() {
...
 
         parser.addArgument(FIND_PREVIOUS_AUTHORS_FLAGS)
                 .dest(FIND_PREVIOUS_AUTHORS_FLAGS[0])
                 .action(Arguments.storeTrue())
-                .help("A flag to make RepoSense utilize Git blame's ignore revisions functionality, RepoSense "
-                        + "will attempt to blame the line changes caused by commits in the ignore commit list to the "
-                        + "previous authors who altered those lines (if available)");
+                .help(String.format("A flag to make RepoSense utilize Git blame's ignore revisions functionality, "
+                + "RepoSense will attempt to blame the line changes caused by commits in the ignore commit list to "
+                + "the previous authors who altered those lines (if available)"));
```
chan-j-d, line 183: https://github.com/checkstyle/checkstyle/issues/6020

src/test/java/reposense/parser/AuthorConfigParserTest.java
```diff
@@ -69,14 +69,15 @@
...
             Stream.of(new SimpleEntry<>(FIRST_COMMAS_AND_DOUBLEQUOTES_AUTHOR,
-                            FIRST_COMMAS_AND_DOUBLEQUOTES_DISPLAY_NAME),
+            FIRST_COMMAS_AND_DOUBLEQUOTES_DISPLAY_NAME),
                     new SimpleEntry<>(SECOND_COMMAS_AND_DOUBLEQUOTES_AUTHOR,
-                            SECOND_COMMAS_AND_DOUBLEQUOTES_DISPLAY_NAME),
-                    new SimpleEntry<>(THIRD_COMMAS_AND_DOUBLEQUOTES_AUTHOR, THIRD_COMMAS_AND_DOUBLEQUOTES_DISPLAY_NAME))
-                    .collect(Collectors.toMap(entry -> entry.getKey(), entry -> entry.getValue()));
+            SECOND_COMMAS_AND_DOUBLEQUOTES_DISPLAY_NAME),
+                    new SimpleEntry<>(THIRD_COMMAS_AND_DOUBLEQUOTES_AUTHOR,
+            THIRD_COMMAS_AND_DOUBLEQUOTES_DISPLAY_NAME))
```
chan-j-d, line 79: https://github.com/checkstyle/checkstyle/issues/6024

chan-j-d (commented):
I spent a bit of time digging through the checkstyle repository. I think some of the behaviour being 'fixed' here as a result of `forceStrictCondition` are actually bugs they intend to fix. I have pasted the relevant issues in a comment to the relevant line of code. I believe the issues they have with the `approved` tags are those properly recognised as bugs unlike the issue I linked previously. 

So I'm really not sure if enabling `forceStrictCondition` is the right thing to do since it seems a sizeable number of changes within this PR itself is to change generally agreeable style in the codebase to the one that checkstyle enforces. And we might end up having to change it back anyway if and when checkstyle fixes the issues listed. 

With regards to my own opinion on which is better, I'll use the following excerpt from `AuthorConfigParserTest` 
**old**:
https://github.com/reposense/RepoSense/blob/2b864022bbf5db55ae9d4f96eaf0a70e961646f5/src/test/java/reposense/parser/AuthorConfigParserTest.java#L73-L79

**new**:
https://github.com/reposense/RepoSense/blob/e4a3f7043d0e758cf723bbe3cc5944c2224655c8/src/test/java/reposense/parser/AuthorConfigParserTest.java#L73-L80)

I think the old one is significantly easier to understand which lines belong to which parent lines. I'm not sure if it is worth the tradeoff for ease of enforcement since such auto-indentations are easily supported by IDEs.

Would like the thoughts of other @reposense/active-developer on this

yhtMinceraft1010X:
I agree that `forceStrictCondition` is a little weird in this regard and would work better if all the mentioned issues were fixed. However, I used `forceStrictCondition` to ensure that there are no additional indentations in the method signature line wrap, which was the original intention of this PR. 

For now, there are 2 possible options:
1) Set `forceStrictCondition` to true and accept that the line wrapping will not look good.
2) Set `forceStrictCondition` to false and accept that excess indentations in the method signature may pop up again.

As a side note,  checkstyle 10 and above requires Java 11. Even if the checkstyle issues were fixed, they would most likely be in a release that is inaccessible to us unless we migrate to Java 11 or use a backport (which I am personally not keen on).

gerhean:
My opinion is that I will only accept style change PR if it is enforceable. Unenforced styles will definitely be violated by new contributors (next year's students).  If it needs Java 11, then is it possible for us to upgrade to Java 11 first?

yhtMinceraft1010X:
In which case, I think it's better to leave the indentations and linewraps as-is since `forceStrictCondition` makes the whole thing weird and I can't find any checkstyle checks to maximise number of parameters on each line.

However, I still wish to implement the `SingleSpaceSeparator` check and `throwsIndent value = 8` property for `Indentation` check. I will close this PR and open a new one for those.

PR #1702, opened 2022-03-05
Frontend: fix vue-fontawesome Vue 3 compatibility

Zhou-Jiahao-1998:
Needed for #1627

## Proposed commit message

```
vue-fontawesome cannot be used directly in another component in Vue 3.
Cypress test is updated to ensure the correct element is fetched during
testing.

Let's move the condition check one level above to avoid this problem.
```

## Other information

You can find the `vue-fontawesome` issue [here](https://github.com/FortAwesome/vue-fontawesome/issues/313).

### DIFF
--- a/frontend/src/views/v-authorship.vue
+++ b/frontend/src/views/v-authorship.vue
@@ -97,14 +97,12 @@
       .file(v-bind:key="file.path")
         .title
           span.path(v-on:click="toggleFileActiveProperty(file)")
-            span(v-show="file.active")
-              .tooltip
-                font-awesome-icon(icon="caret-down", fixed-width)
-                span.tooltip-text Click to hide file details
-            span(v-show="!file.active")
-              .tooltip
-                font-awesome-icon(icon="caret-right", fixed-width)
-                span.tooltip-text Click to show file details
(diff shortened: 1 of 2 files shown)


frontend/src/views/v-authorship.vue
```diff
@@ -97,11 +97,14 @@
       .file(v-bind:key="file.path")
         .title
           span.path(v-on:click="toggleFileActiveProperty(file)")
-            .tooltip
-              font-awesome-icon(icon="caret-down", fixed-width, v-show="file.active")
-              span.tooltip-text(v-show="file.active") Click to hide file details
-              font-awesome-icon(icon="caret-right", fixed-width, v-show="!file.active")
-              span.tooltip-text(v-show="!file.active") Click to show file details
+            span(v-show="file.active")
+              .tooltip
```
gerhean, line 101: How about `.tooltip(v-show="file.active")`

Zhou-Jiahao-1998: Yes I think that is better.

gerhean (approved):
LGTM, minor change so will move forward.

Issue #1634, opened 2022-01-28
Replace cdn links to import from `node_modules` instead


gerhean:
**What feature(s) would you like to see in RepoSense?**
Replace [this](https://github.com/reposense/RepoSense/blob/c4d7690e7b2c3b2626187a2c26ee447d99adf15a/frontend/src/app.vue#L317-L322) with `node_module` imports

**Is the feature request related to a problem?**

Dependabot can track outdated packages in `package-lock.json` files, but not if they are imported directly in the code. Furthermore, it would be good if the frontend only had a 'single source of truth' for all externally imported packages, making the project easier to manage.

Zhou-Jiahao-1998:
@gerhean Looking at [normalize.css@8.0](https://classic.yarnpkg.com/en/package/normalize.css), it seems like it is not reviving any updates. After removing the import, the visual changes are insignificant. Do you think we can simply remove this import?

gerhean:
@Zhou-Jiahao-1998 may I know what insignificant visual changes they are?

Zhou-Jiahao-1998:
![Screenshot 2022-03-07 at 11 38 42](https://user-images.githubusercontent.com/54730603/156964139-66557725-1550-482f-b0cb-f4fa0591df77.png)
![Screenshot 2022-03-07 at 11 40 10](https://user-images.githubusercontent.com/54730603/156964152-15595c71-6b27-4bcd-82ce-46e185e20898.png)
The middle line is slightly thicker and the + icon in the code is more centred.

gerhean:
@Zhou-Jiahao-1998 actually let's leave it in so that web pages styles behave and look consistent across web browsers. As described [here](https://jeffry.in/normalize-css-or-not/).

Zhou-Jiahao-1998:
@gerhean Alright! I am looking into this issue now while Vue 3 upgrade is in progress.

PR #1709, opened 2022-03-07
Import vue-loading-overlay from node_module

Zhou-Jiahao-1998:
Needed for #1634.

## Proposed commit message

```
vue-loading-overlay.css is currently imported directly in the code, making it
difficult to manage.

Let us import from node_module for easy management.
```


gerhean:
I think it's better to just put them all into the same PR as they are all related and minor changes.

PR #1710, opened 2022-03-10
[#1634] Replace cdn links to import from node_modules instead

### LINKED ISSUE #1634: Replace cdn links to import from `node_modules` instead

Zhou-Jiahao-1998:
Fixes #1634 

## Proposed commit message

```
Some css files are currently imported directly in the code, making it
difficult to manage.

Let us import from node_module for easy management.
```


yhtMinceraft1010X (approved):
LGTM

chan-j-d (commented):
Just wanted to check if I understood it correctly. The dependencies needed are stored in `frontend/package.json` which are downloaded when the frontend is being build. And the issue here is that there are direct imports in `app.vue` from the websites themselves bypassing these dependencies. For example `highlight.js 10.5.0` is currently the dependency used but `app.vue` directly imports from `highlight.js 9.14`. Same for `vue-loading-overlay`. 

`muicss` and `normalize` were both originally not included in the dependency so they are included now. Lastly, the `font.css` used is stored in its own local file and does the linking to the google font from there. 

Is that correct?

Zhou-Jiahao-1998:
@chan-j-d Yes that is correct. `package.json` allows us to easily track which dependency can be upgraded. With this PR, we can make sure that there is only one source of truth here. Therefore, avoiding the problem of `highlight.js` here in future.

gerhean:
Instead of loading fonts from `fonts.gstatic.com`, would it be better to just bundle it into the project, such as by using [this](https://github.com/KyleAMathews/typefaces/) or other similar packages?

Zhou-Jiahao-1998:
@gerhean I have decided to use [fontsource](https://github.com/fontsource/fontsource) because the one you suggested is deprecated. This is the alternative suggested by that package. I am not too sure why `codecov/project` is failing.

Zhou-Jiahao-1998:
@gerhean @damithc I propose that we use the GitHub theme in highlight.js for syntax highlighting.
Screenshot 2022-03-14 at 13 03 28
This is what we are currently using.

Screenshot 2022-03-14 at 13 03 09
This is how the new one looks like.

Screenshot 2022-03-14 at 13 06 58

This is how it looks in GitHub.

damithc:
> @gerhean @damithc I propose that we use the GitHub theme in highlight.js for syntax highlighting.

Good idea @Zhou-Jiahao-1998 . Not sure if this should be a separate PR though, given it is a stand-alone tweak.

gerhean (approved):
LGTM, minor change so will merge.

PR #1712, opened 2022-03-11
Ignore cypress package-lock

yhtMinceraft1010X:
## Proposed commit message

```
Cypress package-lock.json is not ignored unlike frontend
package-lock.json.

Let's set the cypress package-lock.json to be ignored.
```


yhtMinceraft1010X:
Closing pull request since removing the cypress package-lock.json causes the frontend tests CI to fail.

PR #1719, opened 2022-03-15
Use liferay node for hotreload

gerhean:
## Proposed commit message

```
Currently, frontend hot reloading is done using the system node.
This is inconsistent with how the other npm tasks are being done.

Let's use the node installed by the liferay node gradle package for
hot reloading.

Let's also move the installFrontend requirement to the base 
hotreload task.
```

## Additional Stuff

I need someone to check that this runs properly on linux.


HCY123902 (approved):

gerhean:
I realized I should probably wait for #1664 to be merged first.

build.gradle
```diff
@@ -228,29 +228,27 @@ task syncFrontendPublic(type: Sync) {
...
-task macHotReloadFrontend(type: Exec) {
-    dependsOn(installFrontend)
+task macHotReloadFrontend(type: com.liferay.gradle.plugins.node.tasks.ExecuteNpmTask) {
     onlyIf {os.isMacOsX()}
     workingDir 'frontend/'
-    commandLine 'npm', 'run', 'serveOpen'
+    args = ['run', 'serveOpen']
 }
 
 task windowsHotReloadFrontend(type: Exec) {
```
dcshzj, line 237: Does this need to be updated to `com.liferay.gradle.plugins.node.tasks.ExecuteNpmTask` too?

gerhean: yes, which is why it is on hold.

yhtMinceraft1010X:
Instead of using liferay for this, should we switch all our other npm-related tasks to use system node? 

My concern is that my proposed [annotations fix](https://github.com/liferay/liferay-portal/pull/4642) has been stalling for a month (and may continue to stall). This is hampering the attempt at upgrading to Gradle 7.

I have sent a follow-up email on this, but I wish to discuss this alternative while waiting for their reply.

dcshzj:
> Instead of using liferay for this, should we switch all our other npm-related tasks to use system node?
> 
> My concern is that my proposed [annotations fix](https://github.com/liferay/liferay-portal/pull/4642) has been stalling for a month (and may continue to stall). This is hampering the attempt at upgrading to Gradle 7.
> 
> I have sent a follow-up email on this, but I wish to discuss this alternative while waiting for their reply.

My experience with open source projects provided by commercial companies is that they would not look at your PRs until much later, because they do not have the resources to review the code.

@gerhean Do you think it is possible to switch to the system's nodejs as proposed by @yhtMinceraft1010X? Would there be a difference?

build.gradle
```diff
@@ -256,29 +256,27 @@ def syncFrontendPublic = tasks.register('syncFrontendPublic', Sync) {
...
-def macHotReloadFrontend = tasks.register('macHotReloadFrontend', Exec) {
-    dependsOn installFrontend
+def macHotReloadFrontend = tasks.register('macHotReloadFrontend', com.liferay.gradle.plugins.node.tasks.ExecutePackageManagerTask) {
     onlyIf {os.isMacOsX()}
     workingDir 'frontend/'
-    commandLine 'npm', 'run', 'serveOpen'
+    args = ['run', 'serveOpen']
 }
 
 def windowsHotReloadFrontend = tasks.register('windowsHotReloadFrontend', Exec) {
```
dcshzj, line 265: Can I double check again on this, do we need to change the `Exec` to `com.liferay.gradle.plugins.node.tasks.ExecutePackageManagerTask`?

dcshzj (commented):
> @dcshzj I don't mind switching to system node, only downside is that user will need to install node by themselves, which may be troublesome depending on their experience with node, especially if they have an outdated version installed.

I see, I think personally I would want to make it easier for people to use the software without having to worry about installing dependencies.

yhtMinceraft1010X:
I'm not sure who user is referring to since from what I recall, only us developers need to have node for the gradle tasks.

gerhean:
@yhtMinceraft1010X There are a few developers who are not really touching the frontend code, so making it easier for them might be good?

yhtMinceraft1010X:
> @yhtMinceraft1010X There are a few developers who are not really touching the frontend code, so making it easier for them might be good? I think I phrased it badly.

In which case I think we can make the switch to liferay. The Gradle 7 upgrade can wait for a while more.

build.gradle
gerhean: I think it has already been changed? It was changed to be consistent with the rest of the file. Only windows is different because of some difficulties with gradle not handling kill signal properly causing the hotreload to be unable to be killed by `ctrl-c`.

PR #1722, opened 2022-03-17
Frontend: change highlight.js theme to follow GitHub's style

Zhou-Jiahao-1998:
## Proposed commit message

```
The current theme used is `colour-brewer`. This theme is very pale and
the syntax highlighting is not very obvious.

Let's use the GitHub theme for a more intuitive syntax highlighting by
adding a new custom CSS file within highlight.js.
```


chan-j-d (commented):
I took a look at the generated report and compared this file: https://github.com/reposense/RepoSense/blob/master/src/test/java/reposense/util/AssertUtil.java

Master branch:
image

New report:
image

How it looks on GitHub:
image

GitHub seems to have an advantage of highlighting blocks within comments. But besides that, my eyesight is not the greatest but I do think that the pale grey on green in the comments is quite hard to read compared to the master branch. Everything else is much better though. Would like to know your thoughts on this.

damithc:
> GitHub seems to have an advantage of highlighting tags within comments. However, besides that, my eyesight is not the greatest but I do think that the pale grey on green in the comments is quite hard to read compared to the master branch. Everything else is much better though. Would like to know if anyone feels the same.

Yes, there are some downsides but seems there is an overall gain over the current master branch.

yhtMinceraft1010X (approved):
LGTM. With regards to the comments being difficult to read, I think people are more interested in looking at the main code so it shouldn't matter too much.

gerhean:
Is there an option to tweak the comment color darker, or make the green lighter perhaps?

Zhou-Jiahao-1998:
@chan-j-d @gerhean I manually override the comment colour. This is how to looks before and after the changes.

Before:
Screenshot 2022-03-27 at 13 48 34

After:
Screenshot 2022-03-27 at 13 48 15

I think it is necessary to make comments a little lighter than the actual code to differentiate them. If you have a better colour in mind, feel free to suggest or put the Hex Code RGB.

chan-j-d (approved):
LGTM! The new comment color looks exactly the same (?) as the old one. I think it looks great as there is now the added highlighting of other keywords.

frontend/src/app.vue
```diff
@@ -321,4 +321,9 @@ export default app;
 <style lang="scss">
 @import './styles/style.scss';
 @import './styles/panels.scss';
+
+.hljs-comment,
+.hljs-meta {
+  color: #5d635a;
+}
```
gerhean, line 328: Will it be better if this css is moved closer to the element like in `v-authorship.vue`?

Zhou-Jiahao-1998: Actually do you think it is a good idea to just create our own css file and import it instead of doing it partially over here? In case in future updates this theme is removed or we want more configurations?

gerhean: Sure

frontend/src/styles/hightlight-js-style.css
```diff
@@ -0,0 +1,73 @@
...
+
+.hljs-comment,
+.hljs-meta {
+    color: #5d635a;
+}
+
+.hljs-variable,
+.hljs-template-variable,
+.hljs-strong,
+.hljs-emphasis,
```
gerhean, line 17: Why declare classes with no css?

gerhean: Actually I just realized this is applying css to multiple classes at once, but it seems like a lot compared to the previous one?

gerhean: Ok I am fine with this as it makes the css clearer.

gerhean (approved):
LGTM

PR #1727, opened 2022-03-22
[#1641] Fix tilde pathing in repo config not working

### LINKED ISSUE #1641: `~` in config file used for relative file pathing does not work for generating reports

chan-j-d:
Fixes #1641

## Proposed commit message

```
Currently, if a local repo file path starting with a tilde is specified
in repo-config.csv file, the entire file path is quoted and tilde
expansion will not be performed.

This change performs bash expansion via an echo command for repo-config
locations. However, the responsibility still falls onto the user to
properly specify their location should they decide to use variable/tilde
expansion.

Let's allow use of variable/tilde expansion in repo-config.csv.
```

## Other information

The solution relies on the fact that `echo` -ing a command does variable expansion for us on both Bash and CMD. There are some additional characters at the end (CRLF for Bash and \r\n for Windows) which need to be stripped.


yhtMinceraft1010X (approved):
LGTM

Zhou-Jiahao-1998 (approved):
LGTM

src/main/java/reposense/util/StringsUtil.java
```diff
@@ -45,6 +45,13 @@ public static String addQuotesForFilePath(String filePath) {
         if (SystemUtil.isWindows()) {
             return "\"" + filePath + "\"";
         } else {
+            // Adds support for tilde expansion by not quoting a filepath starting with a '~' symbol
+            // https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html#Tilde-Expansion
+            if (filePath.charAt(0) == '~') {
+                // if the user used tilde expansion correctly, the filepath itself can be directly reused
+                return filePath;
```
gok99, line 52: Why not replace `~` with `$HOME`?

chan-j-d: I don't think this quite works. If it's just the tilde only, like `~/path/to/repo` then it does get replaced by $HOME. But if it's something like `~chan-j-d/path/to/repo` then it's the home directory of the user `chan-j-d`. And I'm not sure what shortcut that is since the tilde expansion manual doesn't show either

gok99: Ooh I didn't think of this case. If this is the only other case though, we could maybe just split at the first `/` and prepend an unquoted PREFIX at the end: https://askubuntu.com/a/1032374

chan-j-d: I believe there are quite a lot of issues with this to consider. For example, if a user specifies ~/"{someVariable}/path/" in the config file then if we perform our standard manipulation, we would requote the double quotes above into single quotes which removes the variable substitution. This would not be an issue for commands direct from command line since all the relevant bash substitutions are done for us before it reaches us.

As such, I think my rationale was actually that if a user correctly specifies a command that would have worked in bash, we can just copy-paste it in its entirety as they should have done what is linked in your post. Otherwise, that same command would not work if they specified in the CLI instead of the config file. 

I am somewhat uncomfortable with doing too much string manipulation of CLI command inputs because I feel like before we know it, we would have implemented a complete bash mini inside reposense.

gok99: > I am somewhat uncomfortable with doing too much string manipulation of CLI command inputs

This makes a lot of sense, and I agree! I was probably misunderstanding quoting in reposense, but I was originally concerned about inconsistencies in behaviour of reposense for tilde paths and absolute paths, even if both paths worked in shell (would files with special characters as in #1657 cause issues?). If no such inconsistencies exist, then I think this solution is good!

chan-j-d: I think I started with the assumption that the user knows how to properly quote the file in Bash, in that whatever the user specifies in the context of Bash is what they are referring to. When that file path is specified in CLI and reaches reposense, it has already gone through variable substitution and tilde expansion and have its quotation marks removed. So RepoSense only sees the actual file path. Thus, any special chars RepoSense sees should in theory be special characters that were quoted out and were not meant to perform any special functionality. So for all other types of file paths, I think a generic single quote works because I assume that the remaining special characters in the file path are not meant to perform anything special. 

The tilde pathing case is quite specific which is that this only occurs if tilde pathing is used in a config file since if it was specified in CLI, it would have been expanded. Then I use the assumption that whatever else the user specifies there is correct for CLI and we can reuse it directly. I believe specified quotation marks are kept as well. 

There is a case to be made directly for not quoting config file paths at all since for example users cannot now specify bash variables (when not also using tilde paths) in their file path as they will be quoted out. Although I am not sure how useful implementing something like this is as I think tilde path references are quite common, not so much implementing variable references in the file path in a config file.

gok99: Thanks for the clarification!  :+1: 

I think what I was worried about was not so much that the feature of using special character functionality is not available for non tilde paths but that there is an inconsistency in behavior for tilde and non-tilde paths that refer to the same directory. In fact I think it would be okay even to disallow special character functionality for all paths as long as it is consistent (using special bash characters seems like a niche use case anyway). 

I think this generally principle of consistency is probably important because we might otherwise quickly end up with a bunch of gotchas and special cases that make quick changes scary to do without first combing through documentation. Why something would work for tilde paths but not absolute might seem a little magical from the outside.

That being said, this inconsistency only seems to apply with special characters (correct me if I'm wrong) that is unlikely to be a common use case. So, if there is really no way around it, I think it would be worth documenting this difference in behavior (if it isn't already).

chan-j-d: I'm somewhat confused by what you suggested. I believe absolute paths and their tilde counterparts will always work. It's the directories that are like `$HOME/$REPO_PATH/repo.git` specified in the config file or the more magical bash characters that won't work. 

I do think tilde pathing is worth supporting since pathing on server with different accounts often use it.

gok99: The inconsistency I surmised was this:
`~/${VAR}/folder`
`/home/user/${VAR}/folder` 
in a config file.

Apologies if I misunderstood something! :sweat_smile:

chan-j-d: Oh yeah that makes sense then. You would be right that the first would work and the second would not. But I think intentionally disabling or supporting it would require more resources than I think is worth implementing. One way is implementing some kind of interface for path handling and 3 different path handlers for windows CLI, linux CLI and repo configs. Which seems way too much effort for pretty little payoff 🤣

gok99: Yea, this particular inconsistency may not be very significant. I think in general, consistency should be something we try not to break though, and where we are aware of them to document.

By way of a solution, a command builder that has some semantic understanding of commands (preferably with the help of a library) instead of layers of String manipulation will probably help with code quality in general. But something to that effect seems quite far off at this point :(

chan-j-d: I definitely agree with the overall point. Perhaps we can bring it up on a separate issue and discuss exactly what functionality we want to cover and what packages might fit the bill.

chan-j-d:
@gok99 I have had an epiphany in realising that the previous method is kind of... terrible. But that the solution might actually be quite simple. I added an echo command which expands all variables for repo configs only, be it in bash or windows (assuming the user used it correctly). Then this essentially does everything for us. I've tested it on `~`, `~user-id` and `$HOME` type variables and I'm assuming it might work for the stranger cases you've described because I don't see why it wouldnt.

src/main/java/reposense/util/FileUtil.java
```diff
@@ -333,6 +334,15 @@ public static boolean isEmptyFile(String directoryPath, String relativeFilePath)
         return (new File(directoryPath, relativeFilePath).length() == 0);
     }
 
+    /**
+     * Returns the Bash expanded version of the {@code filePath}.
+     */
+    public static String getVariableExpandedFilePath(String filePath) {
+        String echoOutput = CommandRunner.runCommand(Paths.get("."), "echo " + filePath);
+        // CommandRunner returns some white space characters at the end
+        return StringsUtil.stripTrailingWhitespaces(echoOutput);
```
dcshzj, line 343: Any reason why we need to keep the leading whitespaces? My thoughts about this is why are we not able to use `String.trim()` and we have to implement a custom function.

dcshzj (changes requested):

src/main/java/reposense/util/FileUtil.java
chan-j-d: Oops! I only knew about `String::strip` which is not available in Java 8. But it seems `trim()` works for this functionality. I have updated accordingly. Thanks for pointing out!

src/main/java/reposense/util/FileUtil.java
```diff
@@ -350,4 +360,5 @@ public static boolean isEmptyFile(String directoryPath, String relativeFilePath)
     private static boolean isFileTypeInPath(Path path, String... fileTypes) {
         return Arrays.stream(fileTypes).anyMatch(path.toString()::endsWith);
     }
+
```
dcshzj, line 363:
```suggestion
```

dcshzj (changes requested):

dcshzj (approved):
LGTM!

PR #1729, opened 2022-03-22
[#1728] GitVersion: Fix insufficient Git version build failure

### LINKED ISSUE #1728: `git` requirement of 2.14 in DG not sufficient for passing test cases

* **OS and Version:** Ubuntu 18.04
* Git version: 2.17

Precondition: Git version lower than 2.23.0 (2.22.5 or earlier)
1. `./gradlew test`

All test cases passed without issue.

**What actually happened?**
About 5 test cases fail with the error message being that `--ignore-revs-file` is not a recognised flag for `git blame`. 

**Relevant information**:
In the changelog for [version 2.23.0](https://github.com/git/git/blob/master/Documentation/RelNotes/2.23.0.txt) of git, there is the line `"git blame" learned to "ignore" commits in the history, whose
   effects (as well as their presence) get ignored.` which I think is how `git` documents new flags. 

This version is where the `--ignore-revs-file` and `--ignore-rev` flags were added. 
(Version comparison can be found [here](https://github.com/git/git/compare/v2.22.5...v2.23.0) and the commit https://github.com/git/git/commit/ae3f36dea16e51041c56ba9ed6b38380c8421816 shows the addition of the new flags.)

**Proposed Fix**:
We can either change the DG to require 2.23.0 or later or do something like what is done in `configSystemTest` currently:
https://github.com/reposense/RepoSense/blob/a4ee1877a9fc47abde0592201ae489911a4bb604/src/systemtest/java/reposense/ConfigSystemTest.java#L198-L200

I am more inclined to updating the DG as I think the developers working on the codebase should have sufficient git version to interact with all the currently used `git` features.

chan-j-d:
Fixes #1728

## Proposed commit message

```
Machines with Git version less than 2.23 would experience build failure
from failing several test cases that use git blame --ignore-revs-file. 

Additionally, the current implementation of GitVersion does not 
correctly assess if the machine’s Git version is sufficient to support 
the use of --ignore-revs-file flag.

This change fixes and generalizes the implementation of GitVersion.
It also skips the test cases that would require Git version 2.23 
or higher for a machine that does not have the required Git version.

Let’s fix GitVersion and allow developers without Git version 2.23 or 
earlier to properly build RepoSense.

```

## Other information

The current implementation of `GitVersion` does not properly assess that `git version 2.17.0.windows.1` is an earlier version than `2.23.0`. In general, I believe the implementation does not correctly compare the machine's git version to the required git version. 

This implementation change is aimed at correctly comparing the version numbers and to be more general should there be other version checks needed. 

With the `GitVersion` fix, we will be able to properly use `GitVersion::isGitVersionSufficientForFindingPreviousAuthors` to skip test cases for developers whose git version is not recent enough.


src/main/java/reposense/git/GitVersion.java
```diff
@@ -4,19 +4,20 @@
...
 import java.nio.file.Paths;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
  * Contains git version related functionalities.
  * Git version is responsible for finding out the version of git the user of RepoSense is running.
  */
 public class GitVersion {
-    public static final String FINDING_PREVIOUS_AUTHORS_INVALID_VERSION_WARNING_MESSAGE =
```
yhtMinceraft1010X, line 14: I'm not sure why we are moving this.

yhtMinceraft1010X (approved):
LGTM

src/main/java/reposense/git/GitVersion.java
chan-j-d: My reasoning was that we should pair the message constant with the location where the logging is actually done. So either we log and store the constant in `GitVersion`, or we log and store in `RepoSense`. 

I moved it to `RepoSense` because I wanted to the method to be without side effects so that whenever we call the `isGitVersionSufficient...` method can be called without filling up the log.

yhtMinceraft1010X: Nice

Zhou-Jiahao-1998: Thanks for the clarification. I had the same question.

src/main/java/reposense/git/GitVersion.java
```diff
@@ -4,19 +4,20 @@
...
-            "--find-previous-authors/-F requires git version 2.23 and above. Feature will be disabled for this run";
 
-    /** Regex for matching Git version 2.23 and above */
-    public static final Pattern FINDING_PREVIOUS_AUTHORS_VALID_GIT_VERSION_PATTERN =
-            Pattern.compile("(((2\\d*\\.(2[3-9]\\d*|[3-9]\\d+|\\d{3,}))|((([3-9]\\d*)|(1\\d+))\\.\\d+))\\.\\d*)");
+    /** Regex for finding Git Version */
```
Zhou-Jiahao-1998, line 16: The original one simply checks if the version is `2.23` and above. Am I right to say that new one is trying to have a more general method that checks if the version is above this input version (in this case `2.23`) that we choose?

chan-j-d: The new one roughly implements general version comparison. I'm actually not too sure what the original pattern finds, which is my general issue with using extensive regex patterns. I believe even the first part `2\\d*\\.`  is looking for git versions of the form `2.`, `20.` `211.` and so on but I believe fails for even something like git version `1.5.0` or `3.0.0`. 

There is a `Version` class introduced in Java 9 but we're still on Java 8.

dcshzj (approved):
LGTM!

src/main/java/reposense/git/GitVersion.java
Zhou-Jiahao-1998: Thanks for the answer. Agree that this will be helpful in future if we need another version checks!

PR #1730, opened 2022-03-23
Remove unused test resource for `ConfigSystemTest`

chan-j-d:
Removes an unused expected test case file for `ConfigSystemTest`.

## Proposed commit message

```
The dateRange resource stopped being used in commit 81ee3e4 when it was
changed to using the dateRangeWithModifiedDateTimeInLines resource.
However, the test case still remains and is still occasionally updated
even though it is no longer used.

Let's remove this unused system test resource to avoid any further
confusion.
```

## Other information

The `dateRange` resource stopped being use in this commit https://github.com/reposense/RepoSense/commit/81ee3e4bc37e13f7978141d2a6bdbd177273e283 where the same test case was changed to use `dateRangeWithModifiedDateTimeInLines/expected` instead. However, the test case still remains and is still occasionally updated even though it is no longer used. Let's remove it to avoid any further confusion.


yhtMinceraft1010X (approved):
LGTM. Since the test resource is a directory containing the `summary.json` file, will this also delete the directory?

chan-j-d:
Yep. In general, git does not track empty directories. You can view the branch's `resources` folder here where the `dateRange` folder is not present

dcshzj (approved):
Nice catch!

Issue #1703, opened 2022-03-05
Upgrade to Node.js 16


yhtMinceraft1010X:
**What feature(s) would you like to see in RepoSense?**

Currently, the version of Node.js used in GitHub workflows is `12.x`. On the other hand, `build.gradle` uses Node.js version `10.16.0`. I recommend switching both to version `16.14.0` since this is the latest LTS version.

yhtMinceraft1010X:
Closing this issue so that it can be rephrased

PR #1731, opened 2022-03-23
[#1723] Run Windows CI on GitHub Actions instead of AppVeyor

### LINKED ISSUE #1723: Windows CI on GitHub actions

Ubuntu and Mac CI is currently run on GitHub actions, while Windows CI is run on Appveyor. This is done to limit our reliance on a single CI provider. However, GitHub CI does make maintaining and monitoring CI much more convenient. Furthermore, when updates need to be made to CI scripts, it is easy to forget to also update Appveyor scripts.

As a compromise we might want to do Windows CI on GitHub actions as well. Changes made to GitHub actions scripts can be moved to Appveyor periodically/when necessary (this step can be non-trivial).

gok99:
Fixes #1723

## Proposed commit message

```
Due to historical reasons, Linux and Mac CI are run on GitHub Actions
while Windows CI is run on AppVeyor. This makes Windows CI hard to
monitor and maintain.

Let's run CI on Windows (windows-2022, windows-2019) on GitHub Actions
as well. Let's also stop using AppVeyor and remove AppVeyor CI scripts.
```


yhtMinceraft1010X:
In addition, I'm also changing Node.js version to `14.x` and the setup action to `actions/setup-node@v2` for the Linux and Mac CI in #1703. Although it's not necessary to change them now for Windows CI, do rmb to change them if #1703 merges first

yhtMinceraft1010X (approved):
LGTM without factoring in #1703

dcshzj (changes requested):
I think we can completely remove AppVoyer in this PR, similar to how we migrated away from Travis in #1400.

.github/workflows/integration.yml
```diff
@@ -155,6 +155,53 @@ jobs:
...
+    - name: Checkout repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Set up JDK 1.8
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
```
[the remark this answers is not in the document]
gok99: Oops I forgot to put this back in thanks for catching it!

.github/workflows/integration.yml
```diff
@@ -199,8 +202,7 @@ jobs:
       run: ./gradlew lintFrontend
 
     - name: Build with Gradle
-      run: ./gradlew clean checkstyleMain checkstyleTest test systemTest coverage
-
+      run: Measure-Command { Start-Process -NoNewWindow -Wait ./gradlew "clean checkstyleAll test systemTest coverage" }
```
dcshzj, line 205: I think it's okay we do not need to measure how long the process takes? Gradle already has a built-in timer and will report how long the build takes at the end.

Also, I realised you are using `gradlew`. Does this automatically use the `gradlew.bat` file?

dcshzj (changes requested):

.github/workflows/integration.yml
gok99: > Gradle already has a built-in timer

Ah I see, I put that in for consistency with the other tasks, but it does seem somewhat contrived.

> Does this automatically use the gradlew.bat file?

This is a great question. I was following some of [these](https://github.com/gradle/gradle-build-action/tree/main/.github/workflows) workflows which use `./gradlew` for all OSs - I'm not actually sure why this works. It's possible that powershell on these server images have some kind of bash support.

.github/workflows/integration.yml
```diff
@@ -155,6 +155,55 @@ jobs:
...
+        path: $HOME/.npm
+        key: ${{ matrix.os }}-node-${{ hashFiles('**/package-lock.json') }}
+        restore-keys: |
+          ${{ matrix.os }}-node-
+
+    - name: Run linter
+      run: ./gradlew lintFrontend
+
+    - name: Build with Gradle
+      run: Measure-Command { Start-Process -NoNewWindow -Wait ./gradlew "clean checkstyleAll test systemTest coverage" }
```
gok99, line 205:
```suggestion
      run: ./gradlew clean checkstyleAll test systemTest coverage
```

.github/workflows/integration.yml
dcshzj: I see, I think it might be because PowerShell supports running bash scripts.

Regarding the timing of the build commands, is it possible to remove it from the other workflows as well? So that we can make it more consistent for all 3 OSs

.github/workflows/integration.yml
```diff
@@ -153,7 +153,56 @@ jobs:
...
+      with:
+        java-version: 1.8
+
+    - name: Verify Java version
+      run: Write-Output "Actual JDK in use -- "; java -version
+
+    - name: Set up Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
```
dcshzj, line 181: This needs to be updated too now that #1714 has been merged.

dcshzj (changes requested):

.github/workflows/integration.yml
gok99: Updated!

.github/workflows/integration.yml
```diff
@@ -178,7 +178,7 @@ jobs:
     - name: Set up Node.js
       uses: actions/setup-node@v1
```
yhtMinceraft1010X, line 179: There's also one more change to make here. I also updated `actions/setup-node` to use `v2` instead of `v1` for the other OS CI in #1714

yhtMinceraft1010X (changes requested):

.github/workflows/integration.yml
gok99: Thanks for catching that!

dcshzj (approved):
LGTM!

yhtMinceraft1010X (approved):
LGTM

PR #1735, opened 2022-03-28
[#1734] DevOps: update uploading of coverage reports to Codecov

### LINKED ISSUE #1734: Update how Codecov is invoked in pull requests
We have always been using [Codecov Bash uploader](https://docs.codecov.com/docs/about-the-codecov-bash-uploader) to perform code coverage checks for our pull requests, and to annotate on the pull request files on the specific lines that are not covered by tests. However, this practice [has been deprecated since 1 February 2022](https://about.codecov.io/blog/introducing-codecovs-new-uploader/).

We should update the way we invoke Codecov as part of our pull request checks, particularly in 2 aspects:
1. Make use of new/updated ways to invoke Codecov
2. Make Codecov aware of Windows-specific code, so that code are not flagged unnecessarily when they are actually covered in our Windows-based tests

yhtMinceraft1010X:
Fixes #1734 

## Proposed commit message

```
The Codecov bash uploader has been deprecated since February 2022 and
may stop working at any time, causing the coverage reports used for pull
requests to not be generated once that happens.

In addition, the code coverage has always been specific to the Ubuntu
environment, and Codecov often incorrectly report missing coverage for
Windows-specific code.

Let's switch to the new uploader using the codecov/codecov-action GitHub
Action, and expand the coverage testing to include Mac and Windows.
```


.github/workflows/integration.yml
```diff
@@ -71,7 +71,10 @@ jobs:
 
     - name: Run code coverage
       if: ${{ success() && matrix.os == 'ubuntu-20.04' }}
-      run: bash <(curl -s https://codecov.io/bash)
+      run: |
+        curl -Os https://uploader.codecov.io/latest/linux/codecov
+        chmod +x codecov
+        ./codecov
```
gok99, line 77: Just to check, is a codecov token not needed here?

dcshzj: Is it possible to switch to https://github.com/marketplace/actions/codecov instead?

yhtMinceraft1010X: From what I know, codecov tokens are [not needed for public repositories](https://docs.codecov.com/docs/frequently-asked-questions#where-is-the-repository-upload-token-found).

I think the reason why we never use `codecov-action` is to avoid comments like [this one](https://github.com/se-edu/addressbook-level3/pull/128#issuecomment-1080217405) appearing everytime we push new commits and run the action.

dcshzj: > I think the reason why we never use `codecov-action` is to avoid comments like [this one](https://github.com/se-edu/addressbook-level3/pull/128#issuecomment-1080217405) appearing everytime we push new commits and run the action. Hence, I used the direct replacement for the original bash upload line.

I think the comments are currently [disabled in codecov.yml](https://github.com/reposense/RepoSense/blob/28ab050fa80e9e130ca339e20a452688658f577b/codecov.yml) and not because we are using the bash uploader, or will this action not consider the existence of this file?

yhtMinceraft1010X: I tried codecov-action just now. It seems to work - no comments + the `codecov/patch` and `codecov/project` checks are still run and shown.

dcshzj:
Seems ok, looks like it works as intended. However, are you also able to expand the coverage report to the Mac and Windows runs? We have a couple of Windows-specific code and the code coverage erroneously mark them as not being covered by tests, when they are covered when run on Windows machines.

yhtMinceraft1010X:
I can try it out after #1731 is merged

yhtMinceraft1010X:
Closing and reopening PR due to `Expected - Waiting for status to be reported`.

yhtMinceraft1010X:
For some reason, I couldn't get the `coverage.xml` filepath to work for Windows. I have thus reverted back to Linux-only for Codecov Action.

I think we should merge this first and then work on the Windows/Mac coverage as a separate issue.

gok99 (approved):
This PR addresses point 1 in #1734 (doesn't resolve the issue completely), with point 2 being addressed in a separate PR. 

Changes for point 1 looks good to me!

gok99:
Has this been tested this on a test PR with un-covered code, to make sure that the annotations are actually done?

yhtMinceraft1010X:
> Has this been tested on a test PR with un-covered code, to make sure that the annotations are actually done?

I just tested it with changes from #1747. The line annotation still shows up. Will revert now.

yhtMinceraft1010X:
Update: I managed to get this to work on Windows by splitting the `coverage.xml` filepath into the `directory` flag and `files` flag.

dcshzj (approved):
LGTM!

Issue #1736, opened 2022-03-30
Cypress test failing


gerhean:
**Tell us about your environment**

* **RepoSense Version:**
Master
* **OS and Version:**
* **Web Browser and Version (if applicable):**

```
 1 passing (46s)
  1 failing

https://github.com/reposense/RepoSense/runs/5747462686?check_suite_focus=true

  1) switch authorship
       switch authorship view should restore all default controls:
     AssertionError: Timed out retrying after 30000ms: Expected to find element: `#tab-authorship > .title > .contribution > .fileTypes input[id="yml"]`, but never found it.
      at Context.eval (http://localhost:9000/__cypress/tests?p=tests/codeView/codeView_switchAuthorship.js:118:85)
```

**Please include the steps to reproduce the bug.**

Just run it.

**What was expected to happen?**

Pass

**What actually happened? Please include a screenshot of the output.**

Fail

**Extra Information**

Caused by #1731 deleting a yml file. When files are deleted, the attribution to the original author is deleted as well.
Unfortunately, we are using Reposense master branch for frontend testing, which is highly unstable. This issue should be solved in two parts:

1. Short term: Hotfix the test failure quickly by disabling the test or changing it. (Done)
2. Long term: Migrate all tests to not rely on unstable sources (Use a fork or test branch?).

PR #1737, opened 2022-03-30
[#1736] Temporary adjust cypress test codeView_switchAuthorship

dcshzj:
Part of #1736.

## Proposed commit message

```
Due to the removal of the appveyor.yml file in #1731, the
codeView_switchAuthorship cypress test no longer works, as it expects
yml files to be present in the RepoSense repository.

Let's temporarily adjust the codeView_switchAuthorship cypress test so
that the frontend CI tests can pass, by removing the check on yml files.
In the longer term, the frontend tests' dependency on the RepoSense
repository should be removed, as it can make the CI tests unstable.
```

## Other information

The yml file removed is the appveyor.yml file, which was removed in #1731. In the long term, CI tests should not be affected by changes to the RepoSense repository.


dcshzj:
Self-merging since CI tests pass now, and was separately verified by @gerhean.

PR #1740, opened 2022-03-30
[#1739] Add schema for json data files

### LINKED ISSUE #1739: Add schema for json data files

I want to see json schema added for data files generated.

It is troublesome troublesome to manually check what went wrong with the data file. Having a schema will allow easy debugging and see which specific field is missing/incorrect.

Zhou-Jiahao-1998:
Fixes #1739 

## Proposed commit message

```
There is no json schema for the data files generated. This can
be difficult to debug when the file is corrupted or when backend
generates a wrong json file.

Let us add schema file to our json files to make future debugging
easier.
```

## Other information

Refer to [this](https://www.jetbrains.com/help/idea/json.html#ws_json_using_schemas) for the guide on how to use the schema.


yhtMinceraft1010X (changes requested):
This sounds like a great idea. I think the instructions on how to use this should be included in the DG, such as `workflow.md`.

yhtMinceraft1010X:
One concern I have though is that the plugin described in the given guide is only available on IntelliJ IDEA Ultimate Edition, which is a paid product. Not every developer may have the paid version and consequently won't be able to make use of this feature.

Zhou-Jiahao-1998:
> One concern I have though is that the plugin described in the given guide is only available on IntelliJ IDEA Ultimate Edition, which is a paid product. Not every developer may have the paid version and consequently won't be able to make use of this feature.

Yes that is a valid concern. I have added the instruction for VS Code in `workflow.md`.

yhtMinceraft1010X (approved):
LGTM

docs/dg/workflow.md
```diff
@@ -89,6 +89,11 @@ The issues for first timers usually have guidance provided in the comment or hav
 See [vue-devtools project home](https://github.com/vuejs/vue-devtools) page for more details.
 </box>
 
+**You can use the json schemas provided in `/frontend/json-schema/` to check if the json data generated in `/reposense-report/` has the correct format.**
+* If you are using **IntelliJ**, please follow the guide [here](https://www.jetbrains.com/help/idea/json.html#ws_json_using_schemas).
+* If you are using **Visual Studio Code**, open the `.json` data file. Using `summary.json` as an example, add `"$schema": "../frontend/json-schema/summary-schema.json"` to the start of json. If you need to check using other schemas, replace `summary-schema.json` with other schema files provided.
```
yhtMinceraft1010X, line 94: Will adding the `$schema` line affect the way the frontend processes the json?

gerhean:
Will this schema be used for typescript as well? I rather not have multiple schemas.

frontend/json-schema/authorship-schema.json
```diff
@@ -0,0 +1,45 @@
+{
```
gok99, line 1: I'm not sure if non-required fields need to be included here, but there is also a field `isBinary` (and `isIgnored` added in #1647) that are only serialized if true.

See: [FileResult.java](https://github.com/reposense/RepoSense/blob/2b864022bbf5db55ae9d4f96eaf0a70e961646f5/src/main/java/reposense/authorship/model/FileResult.java)

gok99 (commented):
Just had a thought (this could probably be explored separately if at all): would it be a good idea to generate these automatically from the Java classes which can act as a single truth source? Haven't looked into this too much but maybe this is useful: http://wilddiary.com/generate-json-schema-from-java-class/

gerhean:
I agree with @gok99 that I would rather have a single file which both the frontend and java can verify json against.

However my priority for now is Typescript integration.

PR #1744, opened 2022-04-01
Gradle: update node and execfork plugins to newer versions

yhtMinceraft1010X:
Part of #1661 

## Proposed commit message

```
The com.github.psxpaul.execfork and com.liferay.node have newer versions
available (0.2.0 and 7.2.16 respectively), and they are compatible with
Java 8 and Gradle 6.9.2 that we are currently using.

Let's upgrade them to move closer towards Gradle 7.
```

## Other information

The 2 updated plugins are:

* [com.github.psxpaul.execfork](https://plugins.gradle.org/plugin/com.github.psxpaul.execfork) Version `0.2.0` has now been released. This version re-implements Java 8 compatibility and is also compatible with Gradle 7.
* [com.liferay.node](https://plugins.gradle.org/plugin/com.liferay.node) Version `7.2.16` has been released. However, there are no significant changes relating to Gradle 7 compatibility. We have to continue using Gradle 6.9.2 for now.

There is one other plugin that I tried to upgrade: `com.palantir.git-version`, which has the latest version of `0.14.0`. However, it is not compatible with Java 8, hence we have to stick to `0.13.0` for now.

Screenshot (172)


build.gradle
```diff
@@ -185,6 +185,7 @@ def startServerInBackground = tasks.register('startServerInBackground', com.gith
     args = ['--config', './frontend/cypress/config', '--since', 'd1', '--view']
     String versionJvmArgs = '-Dversion=' + getRepoSenseVersion()
     jvmArgs = [ versionJvmArgs ]
+    killDescendants = false // Needed because Java 8 does not support killDescendants
```
gok99, line 188: Is this change related to the `com.github.psxpaul.execfork` upgrade?

yhtMinceraft1010X: This flag had been implemented in version `0.1.13`, but it won't work with Java 8. Considering that `0.1.13` was not compatible with Gradle 7 anyway, I decided to use `0.1.12` before `0.2.0` came out.

As a side note, `0.1.14` was the first version to have Gradle 7 compatibility but it still could not work with Java 8. `0.2.0` was the first version to re-add Java 8 compatibility.

gok99: Thanks for the clarification. This looks super confusing, nice work keeping track of everything! :sweat_smile:

gok99 (approved):
LGTM!

build.gradle
```diff
@@ -185,6 +185,8 @@ def startServerInBackground = tasks.register('startServerInBackground', com.gith
     args = ['--config', './frontend/cypress/config', '--since', 'd1', '--view']
     String versionJvmArgs = '-Dversion=' + getRepoSenseVersion()
     jvmArgs = [ versionJvmArgs ]
+    killDescendants = false // Needed because Java 8 does not support killDescendants.
+    // This flag should be removed if we fully migrate to Java 11.
```
gerhean, line 189: First is it possible to specify what this flag does? Next can `This flag` be changed to `The above flag`?

gerhean (commented):
LGTM

Minor change so will just merge.

PR #1745, opened 2022-04-02
Refactor Json comparison methods and resources in systemtest

chan-j-d:
Related to adding new system tests for local repos for #1637 

## Proposed commit message

```
The current structure of the system tests is such that it is meant to
only support the ConfigSystemTest. As such, the Json comparison methods
in ConfigSystemTest are not reusable. The resources used are also not
properly grouped.

This change aims to refactor some of the more useful Json comparison
methods so that they can be reused for future system tests.
Additionally, the resources are now grouped by the class file that is 
using them. 

Let's refactor the methods and resources in systemTest to make it easier
to implement new system test cases.
```

## Other information
The main motivation for making this change is to pave the way for the possibility of adding new system test cases. In particular, I would like to add local test cases specifically for issues like #1637. Regardless, even if we decide to not add such system test cases, I do believe that the changes made in this PR help to better categorize the files and make the Json comparison methods more reusable.


yhtMinceraft1010X (approved):
LGTM

src/systemtest/java/reposense/util/SystemTestUtil.java
```diff
@@ -0,0 +1,48 @@
...
+        try (Stream<Path> pathStream = Files.list(expectedDirectory)) {
+            for (Path file : pathStream.collect(Collectors.toList())) {
+                Path expectedFilePath = expectedDirectory.resolve(file);
+                Path actualFilePath = actualDirectory.resolve(file);
+                if (Files.isDirectory(file)) {
+                    verifyAllJson(expectedFilePath, actualFilePath);
+                }
+                if (file.toString().endsWith(".json")) {
+                    assertJson(expectedFilePath, actualFilePath);
+                }
```
dcshzj, line 29: Are these two conditions mutually exclusive? Is there a chance where we have a directory that ends with ".json"?

dcshzj (changes requested):

src/systemtest/java/reposense/util/SystemTestUtil.java
chan-j-d: image

Seems like you were right! I've adjusted it to be `else if`.

dcshzj (approved):
LGTM!

PR #1746, opened 2022-04-02
[#1637] Add system test cases for local repositories 

### LINKED ISSUE #1637: Multiple local paths with the same final directory causes one report to overwrite others

chan-j-d:
Fixes #1637

## Proposed commit message
```
This will add new system test cases for testing two core issues
previously pertaining to 1. relative pathing and 2. two local
repositories sharing the same final directory.

This also serves as a template for possible future system tests.

Let's add some system test cases for local repositories.
```

## Other information
The two issues that these test cases covered are #1637 and #1546. 
With regards to the concern of significant increase in build time
  - At the moment, I have chosen to only clone one remote repository, `testrepo-Alpha` which is also quite small. The remaining clones will be local clones which are extremely fast. 
  - Additionally, unlike the `ConfigSystemTest`s, once the repositories are cloned, RepoSense will not perform any further remote cloning, just local cloning. 

This PR also implements a possible way `summary.json` files can be tested, by reading the file into memory then excluding the fields that change from run to run.


src/systemtest/java/reposense/LocalRepoSystemTest.java
```diff
@@ -0,0 +1,71 @@
...
+    @AfterAll
+    public static void deleteClonedLocalRepos() throws Exception {
+        FileUtil.deleteDirectory(LOCAL_DIRECTORY_ONE_PARENT);
+        FileUtil.deleteDirectory(LOCAL_DIRECTORY_TWO_PARENT);
+    }
+
+    @Test
+    public void testSameFinalDirectory() {
+        String cliInput = String.format("-r %s %s -s d1 -u 01/04/2022 -o local-test",
+                LOCAL_DIRECTORY_ONE, LOCAL_DIRECTORY_TWO);
```
yhtMinceraft1010X, line 54: I'm not sure why the style for the args is different from that of lines 33-34 and 36-37.

Zhou-Jiahao-1998:
@chan-j-d Can I check that if you mean that there will be a significant **reduction** in build time?

Also, I feel that the three types of json file should either be all included or excluded. It is true that `Summary.json` is the one that is most likely to change, I think that they all serve the same purpose of keeping some data for the frontend to consume. It will be a little confusing to just make an exception here? Maybe some form of guide can be made so that it is easier to make changes?

src/systemtest/java/reposense/LocalRepoSystemTest.java
chan-j-d: I'm not too sure what you mean. The above you have quoted for lines 53 and 54 is the actual system test simulating if we were running it from CLI. Lines 33-34 and 36-37 is setup for the test as we have to clone local repositories before we can use them for testing.

chan-j-d:
@Zhou-Jiahao-1998 It is to say that there will not be a significant increase in build time. These local system tests are new tests meant to facilitate local repo testing so it can only increase build time. But my point was just to address the main concern with regards to adding new system test cases which was that it greatly increases build time (which is true for the `ConfigSystemTest`s which are all remote repo tests). 

With regards to the `Summary.json`, I do agree with it. But I've been struggling to find a way to compare equality between the file contents while bypassing the transient fields like report generation time. It is part of the reason why the `ConfigSystemTest` methods are so convoluted as they manually set those fields during report generation. And actually there is some precedent where `dateRangeWithModifiedDateTimeInLines` does not have the `summary.json` file present.

src/systemtest/java/reposense/LocalRepoSystemTest.java
yhtMinceraft1010X: What I meant to say is I was curious why it's
```
       Paths.get("."),
       LOCAL_DIRECTORY_ONE
```
in lines 33-34 and not `Paths.get("."), LOCAL_DIRECTORY_ONE` on the same line.

Or the other way round with respect to line 54

chan-j-d: Oh I see! I think it was more that they were different types of arguments, since `Paths.get(".")` is the working directory and `LOCAL_DIRECTORY_ONE` is the output directory. Whereas to me, for lines 53 and 54, the two local directories were just string substitutions being performed so I did not see a need to separate them. It's just the way I decided to code it and I have no strong preference one way or the other. Do you want me to change it?

chan-j-d:
@Zhou-Jiahao-1998 I hacked together a way to get the summary checks done. It achieves it by treating only the summary.json file differently. It reads the file into memory as a `JsonObject`, removes the transient fields and then does their `equals` method which apparently supports deep comparison according to their API. I've checked it over a few cases and it seems to correctly fail when I fiddle with the fixed fields in the `summary.json` file. 

Besides that, I've put the PR on hold as I can't quite seem to figure out why the build is failing while it is passing on both windows and ubuntu 18.04 on my machine. Especially since the test case failure seems unrelated to the changes I've made for summary.json.

src/systemtest/java/reposense/LocalRepoSystemTest.java
yhtMinceraft1010X: I personally prefer having as many variables as possible on one line.

chan-j-d:
This PR is currently on hold at least till the end of the semester. The current issue is failing CI which seems to be caused by the commit https://github.com/reposense/testrepo-Alpha/commit/7d7584fc204922cc5ff3bd5ca073cad6bed2c46a being detected as being on 02/05/2018 in CI while it is detected as 02/06/2018 on my own computer. The build passes locally on my Windows 10 and Ubuntu 18.04 machines but fails 100% of the time on GitHub CI.

dcshzj:
> This PR is currently on hold at least till the end of the semester. The current issue is failing CI which seems to be caused by the commit [reposense/testrepo-Alpha@7d7584f](https://github.com/reposense/testrepo-Alpha/commit/7d7584fc204922cc5ff3bd5ca073cad6bed2c46a) being detected as being on 02/05/2018 in CI while it is detected as 02/06/2018 on my own computer. The build passes locally on my Windows 10 and Ubuntu 18.04 machines but fails 100% of the time on GitHub CI.

The timezone used on GitHub Actions is UTC+0, your computer is probably using UTC+8 timezone. Are your test cases able to account for the time zone differences? It should be possible if you are using the internal time utils.

src/systemtest/java/reposense/LocalRepoSystemTest.java
```diff
@@ -0,0 +1,81 @@
...
+        String[] args = cliInput.split(" ");
+        RepoSense.main(args);
+        Path expectedFilePath = loadResource(getClass(), "LocalRepoSystemTest/testSameFinalDirectory");
+        SystemTestUtil.verifyReportJsonFiles(expectedFilePath, REPORT_DIRECTORY_PATH);
+    }
+
+    @Test
+    public void testRelativePathing() {
+        String relativePathForTesting = "parent1/../parent1/./test-repo";
+        String cliInput = String.format("-r %s -s d1 -u 01/04/2022 -o local-test",
```
yhtMinceraft1010X, line 74: Try adding the `-t` flag here and in line 63 using SG timezone.

chan-j-d:
Thanks so much @dcshzj and @yhtMinceraft1010X! It seems I wasted many hours barking up the wrong tree. I think if we fix the time zone used for analysis, it reliably passes the system test cases. It has passed about 5 consecutive times since I made the fix. The rest are minor changes to the code. 

This also implements the `summary.json` checks as suggested by @Zhou-Jiahao-1998. From my own tests, it seems to reliably fail when I change the 'fixed' `summary.json` fields.

yhtMinceraft1010X (approved):
LGTM

chan-j-d:
Requesting review from @reposense/stage-1-active-reviewer

src/systemtest/java/reposense/LocalRepoSystemTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @AfterAll
+    public static void deleteClonedLocalRepos() throws Exception {
+        FileUtil.deleteDirectory(LOCAL_DIRECTORY_ONE_PARENT);
+        FileUtil.deleteDirectory(LOCAL_DIRECTORY_TWO_PARENT);
+    }
+
+    @Test
+    public void testSameFinalDirectory() {
+        String cliInput = String.format("-r %s %s -s d1 -u 01/04/2022 -o local-test -t UTC+08",
```
dcshzj, line 61: A minor nit: I think the `01/04/2022` should be set as a constant in this class, and the `local-test` is already declared as `OUTPUT_DIRECTORY`, that should be used instead.

src/systemtest/java/reposense/LocalRepoSystemTest.java
```diff
@@ -0,0 +1,84 @@
...
+                LOCAL_DIRECTORY_ONE, LOCAL_DIRECTORY_TWO);
+        runTest(cliInput, "LocalRepoSystemTest/testSameFinalDirectory");
+    }
+
+    @Test
+    public void testRelativePathing() {
+        String relativePathForTesting = "parent1/../parent1/./test-repo";
+        String cliInput = String.format("-r %s -s d1 -u 01/04/2022 -o local-test -t UTC+08",
```
dcshzj, line 69: Same comment as above.

dcshzj (changes requested):

src/systemtest/java/reposense/LocalRepoSystemTest.java
chan-j-d: Good catch, I've fixed it

dcshzj (approved):
LGTM!

PR #1747, opened 2022-04-02
Release Logger resources once RepoSense finishes processing

chan-j-d:
Releases the `reposense.log.0` resource for writing logs.

## Proposed commit message

```
Log resources are currently only fully released when the process exits.
This leads to the persistence of the reposense.log.0.lck lock file in
some scenarios.

Let's release the log file resource at the end of RepoSense execution.
```

## Other information
Necessary for #1746 as the log file is not properly released during each test case and thus cannot be deleted while running the system tests. I'm not too sure how `ConfigSystemTest` gets around it but it does not seem to have the same problem deleting the report directories. The implementation between `ConfigSystemTest` is quite different as it does not make use of `RepoSense::main`.

Right now, I believe the issue can be seen when we run `gradlew run -Dargs="-r <some repo> -v"` and then Ctrl+C the server. The lock file `.lck` can still be seen even after the execution ends. 
image

Solution taken from here:
https://stackoverflow.com/a/22957009

### DIFF
--- a/src/main/java/reposense/RepoSense.java
+++ b/src/main/java/reposense/RepoSense.java
@@ -8,6 +8,7 @@ import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.logging.Level;
+import java.util.logging.LogManager;
 import java.util.logging.Logger;
 
 import net.sourceforge.argparse4j.helper.HelpScreenException;
@@ -109,6 +110,8 @@ public class RepoSense {
         } catch (HelpScreenException e) {
             // help message was printed by the ArgumentParser; it is safe to exit.
         }
+
+        LogManager.getLogManager().reset();
     }
 
     /**


gok99 (commented):
Just to clarify, the current behavior is that the `.lck` file is only removed after a Ctrl+C of the server but we want it to be released after report generation instead. Is this right?

chan-j-d:
Roughly from what I tested, if we run via `java -jar ...`, the lock is released whenever the JVM is shut down. So this issue does not occur whether or not the server is opened or not. 

But when running on `gradlew run ...`, if no server is opened (no `-v` flag) then the lock is also released. But if `-v` flag is used, then for some reason my intellij is unable to release the log file lock after Ctrl+C. This becomes very problematic when running the new system tests I wrote in #1746 as I am unable to delete the report directory for cleanup as something is holding on to the log file lock even at the end of the `RepoSense::main` method. 

I'm not too clear on the exact reasoning behind the above behaviour 😅. A separate diagnosis or different solution is very welcome!

gok99:
> But if -v flag is used, then for some reason my intellij is unable to release the log file lock after Ctrl+C.

Yea I see the same behavior here. It seems we're starting the server in ReportServer,java so perhaps JVM cannot be shut down until the server process stops? In any case, it seems like explicitly releasing the lock is a sensible thing to do!

gok99 (approved):
LGTM!

chan-j-d:
@gok99 What kind of platform are you running the gradlew commands on? I'm testing on IntelliJ and regardless of whether I use the run configurations of IntelliJ or use their command lines, Ctrl+C never seems to release the log file lock. 

With the one line change in this PR, the log file seems to be released even while the server is up as the `RepoSense::main` wraps up. I've tried it on both intellij and a newly generated Jar file.

gok99:
> What kind of platform are you running the gradlew commands on? 

Am running them in a bash shell in IntelliJ fedora.

chan-j-d:
Now I'm very confused 😅 I think it's not actually the JVM but the process ending that causes the release of the locks. That is why the jar file never has the issue for either of us. But now the strange behaviour is with Windows IntelliJ having issues releasing the lock after the execution ends whereas it seems just fine for you in fedora. 

So I guess there are 2 follow-up questions:
1. Is it just an IntelliJ Windows issue where the lock isn't being released when run with gradlew? Need someone else to test for this.
2. Is it necessary to manually release the log file lock at the end of the Java execution? I think my answer for this is still yes. (Although my answer is obviously biased as I want to implement #1746 with this 😂)

dcshzj:
Can I check if there are any updates from the discussion? Is it only a Windows-specific issue? Do you need more people to help test if the bug exists?

chan-j-d:
Haven't managed to get anyone else to test. From my own testing, it seems to be a windows specific issue. 

On Ubuntu 18.04, when running either the jar file or `./gradlew`, the `.lck` file disappears after terminating the process via ctrl+C.
On Windows, running the jar file and then ctrl+C removes the `.lck` file. However, on both IntelliJ and from CMD, using `gradlew run` and then ctrl+C leaves the `.lck` file behind.

chan-j-d:
@dcshzj, an update for the testing. @yhtMinceraft1010X helped me test on his own windows system and it seems that running `gradlew run ...` with `-v` and then Ctrl+C to stop the process results in the lingering `.lck` file. So I think it might be something worth cleaning up by including this line. The added benefit is that this releases the file lock the moment the java program ends (even while the report is still working).

dcshzj (approved):
LGTM! I see, it makes sense to clean up ourselves even if it is done automatically.

PR #1749, opened 2022-04-02
Refactor GitClone::clone method to increase flexibility in usage

chan-j-d:
Part of PR #1746 

## Proposed commit message

```
There are various methods in GitClone that are used only in testing.
However, they are still part of the main code and frequently updated.

This change would refactor the clone to cloneAndBranch, and move these
methods to a testing utility class to reduce the maintenance burden. A
more generic GitClone::clone method is also added in its place for
flexible usage during testing.

Let's move the GitClone::clone and cloneBare method to a testing class
to improve the ease of maintenance and implement a more generic clone
method for more flexible usage.
```

## Other information
Necessary for #1746 as there is currently no easy way to clone a repository with no extra actions.


yhtMinceraft1010X:
Tbh, I'm not sure why we have this `clone` method (referring to the method before refactoring) as well as `cloneBare` in `GitClone.java` since both only appear to be used in test cases. I feel it makes more sense to remove them and redesign the test cases to use the other cloning methods.

chan-j-d:
Yeah I noticed that some time back too. I believe `cloneBare` stopped being used because the main codebase started to use the asynchronous `cloneBareAsync`. However, I think there is a benefit to creating these methods that simply clone a repository for setting up test cases for various scenarios a normal user might encounter. Which was mainly why I wanted to implement a generic `clone` method for #1746 to use. The main distinction of the methods used in the RepoSense main codebase is that RepoSense generally only does bare clones which are not really the same as the normal clone we use.

src/main/java/reposense/git/GitClone.java
```diff
@@ -100,21 +100,39 @@ public static void cloneShallowPartialBare(RepoConfiguration config, Path rootPa
...
-            Path rootPath = FileUtil.getRepoParentFolder(config);
-            Path repoPath = Paths.get(rootPath.toString(), config.getRepoName());
+            config.updateBranch();
+            GitCheckout.checkout(config.getRepoRoot(), config.getBranch());
+        } catch (GitBranchException gbe) {
+            logger.log(Level.SEVERE,
+                    "Exception met while trying to get current branch of repo. Analysis terminated.", gbe);
+            throw new GitCloneException(gbe);
+        } catch (RuntimeException rte) {
+            logger.log(Level.SEVERE, "Branch does not exist! Analysis terminated.", rte);
```
yhtMinceraft1010X, line 115: Is there a way to test either `catch` branches?

chan-j-d: I believe the 'branch does not exist' scenario is already covered in `systemTest` as the `repo-config.csv` that they use has a `nonExistentBranch` for its second last repo.

For the `exception met while trying to get current branch` case, I'm actually not too sure if we can force this exception to occur. It sounds like something that only happens when command runner outright fails.

chan-j-d: After doing a bit of digging, I remembered that you pointed out that this method isn't used in the main codebase. In that case, I think it actually makes more sense to get rid of all these logging messages and retain the barebones functionality of the method itself to simplify it. 

Branch checking is actually done with `GitRevParse` in the main code.

chan-j-d:
@yhtMinceraft1010X I thought about your suggestions for some time. I do agree that the original `clone` method is extremely out of place, especially since it has code behaving differently depending on whether it is in the test environment or not but is used only in test cases. However, I don't think we can do without these methods as we need them to setup other test cases.

As such, I think it would be beneficial to transfer these methods to a `test` directory where we can vastly simplify the code to perform the same functionality. This also resolves the issue of testing the try-catch brackets. 

Requesting another review. Thanks!

src/main/java/reposense/git/GitClone.java
yhtMinceraft1010X: Should we also move the `cloneBare` method to the new `TestRepoCloner` class?

chan-j-d: I agree! I didn't realise `cloneBare` was also entirely not used in the main code. All instances were replaced with `cloneBareAsync`. I also don't think we'll be changing any of the cloning settings in the main code any time soon, so I think it might be better to remove it for now.

I have checked the remainder methods and they seem to be used in the main code in some form, so I have left them untouched.

yhtMinceraft1010X (approved):
LGTM

dcshzj (changes requested):
Your commit message mentions "A more generic GitClone::clone method is also added in its place for flexible usage during testing.". Is this still true?

chan-j-d:
Yep. The new `clone` only clones the repo to the designated output directory. The new `cloneAndBranch` does what the old `clone` does, which is to clone to a designated directory and then branch depending on the config file.

dcshzj (approved):
I see, it makes more sense now. LGTM!

PR #1752, opened 2022-04-06
Add zod for json validation

### LINKED ISSUE #1756: Add support for json validation

Add support for json validation during runtime for frontend.

Having json validation can avoid unexplained runtime errors due to incorrect json data.

[zod](https://github.com/colinhacks/zod#comparison) can be used. It is a TypeScript-first schema declaration and validation library. It has no dependencies and smaller than the alternatives.

- [Joi](https://github.com/hapijs/joi) is similar but doesn't support static type inference. 
- [Yup](https://github.com/jquense/yup) makes all field optional by default. Since most of the fields in our json are mandatory, it will be rather troublesome to implement and maintain. The syntax can be a little confusing as well.

**Additional context**

Zhou-Jiahao-1998:
Fixes #1756 

## Proposed commit message

```
api.js is heavily coupled with any frontend files. As a result, it
is very difficult to convert the whole file. The json parsing part
has been isolated into a new file called json-parser.ts. Schemas
are added for 3 types of json files we have.

Let us add support for json validation so that we can identify the
issue clearly during runtime.
```


Zhou-Jiahao-1998:
@gerhean To reduce the work, I have separated the json parsing part out from `api.js`. I have added `zod` schema check for  `summary.json` and `commits.json`. I personally think that the delay is too significant. This is why I have yet to implement the schema for `authorship.json`. 

If you agree that the delay is too significant, I can stop this pr and make this PR a partial conversion of typescript as it took a lot of time and effort to sort the types out. We can also just stick with the json schema check in #1740.

gerhean:
One of the reasons why I suggested zod is as the schema can be directly turned into typescript interfaces as follows:
```
const A = z.string();
type A = z.infer<typeof A>; // string

const u: A = 12; // TypeError
const u: A = "asdf"; // compiles
```

Please try to follow DRY principle thanks!

frontend/src/utils/json-parser.ts
```diff
@@ -0,0 +1,465 @@
...
+      BRANCH: z.string(),
+      COMMIT_PATH: z.string(),
+      HISTORY_PATH: z.string(),
+      REPO_URL: z.string(),
+    }),
+  }),
+});
+
+const commitsSchema = z.object({
+  authorFileTypeContributionMap: z.object({}).catchall(z.object({}).catchall(z.number())),
```
gerhean, line 165: A little confusing, is it possible to use Records or Maps?

Zhou-Jiahao-1998: Thanks for pointing this out. This is possible.

Zhou-Jiahao-1998:
> One of the reasons why I suggested zod is as the schema can be directly turned into typescript interfaces as follows:
> 
> ```
> const A = z.string();
> type A = z.infer<typeof A>; // string
> 
> const u: A = 12; // TypeError
> const u: A = "asdf"; // compiles
> ```
> 
> Please try to follow DRY principle.

The structure is very complex and I do not have a clean solution here. Some parts have been extracted out. Removing all duplicates may not be feasible and logical because the structure of  `window` and the json files are different. Although they share many fields, I cannot leverage on the schema generated from zod.

gerhean:
Is it possible to make all dictionaries of objects into its own interface, as I have a feeling that they will need to be used in other parts of the code.

Also is it possible to put the interfaces into their own file? Maybe the zod objects should be in their own file as well. You can maybe use `-types` as a suffix for the filename, like `json-types.ts` or `api-types.ts` or something, which is similar to how angular works.
Alternatively you can try making a folder named `types` and drop all the type definitions inside.

Zhou-Jiahao-1998:
@gerhean I have attempted to separate the types out. Do let me know if this is the right direction. I had to disable `eslint` in `json-parser.ts` as it kept complaining about the import. I saw a solution [here](https://stackoverflow.com/questions/59265981/typescript-eslint-missing-file-extension-ts-import-extensions) but it did not work for me. I am still looking into this problem.

gerhean:
Can you post the relevant eslint error messages?

Also it would be good to rename the `types.ts` file to something that describes the types which are defined inside.

Zhou-Jiahao-1998:
@gerhean The issue is solved. I needed to update two items in `.eslintrc.json`. I have split up the types as well

frontend/src/utils/json-parser.ts
```diff
@@ -0,0 +1,166 @@
...
+    const folderName = window.REPOS[repoName].outputFolderName;
+    return this.loadJSON(`${folderName}/authorship.json`, 'authorship')
+        .then((files: AuthorshipSchema) => {
+          window.REPOS[repoName].files = files;
+          return files;
+        });
+  },
+
+  // calculate and set the contribution of each commitResult and insert repoId into commitResult, since not provided in json file
+  setContributionOfCommitResultsAndInsertRepoId(dailyCommits: { date: string; commitResults: { deletions: number;
```
gerhean, line 153: Can you make `commitResults` an interface since it is hard to read.

Zhou-Jiahao-1998: Thanks for catching this. There is already an interface for it but I forgot to update here.

frontend/src/utils/json-parser.ts
```diff
@@ -0,0 +1,166 @@
...
+    const folderName = window.REPOS[repoName].outputFolderName;
+    return this.loadJSON(`${folderName}/authorship.json`, 'authorship')
+        .then((files: AuthorshipSchema) => {
+          window.REPOS[repoName].files = files;
+          return files;
+        });
+  },
+
+  // calculate and set the contribution of each commitResult and insert repoId into commitResult, since not provided in json file
+  setContributionOfCommitResultsAndInsertRepoId(dailyCommits: { date: string; commitResults: CommitResult[] }[], repoId: string) {
```
gerhean, line 155: I would still prefer that the dailyCommits parameter have its own type, maybe `dailyCommitType` or something

gerhean:
Can I request to add return types for the functions in `json-parser.ts` as well?

Zhou-Jiahao-1998:
> Can I request to add return types for the functions in `json-parser.ts` as well?

@gerhean I think adding return types will make the code hard to read. Typescript auto infer the return types of the function. The function here all return a Promise of some rather complex type. I do not think that we should state the return type explicitly because once the other files are converted to typescript, it will be very clear. Adding return type will complicate future code modifications.

gerhean:
> I think adding return types will make the code hard to read. Typescript auto infer the return types of the function. The function here all return a Promise of some rather complex type. I do not think that we should state the return type explicitly because once the other files are converted to typescript, it will be very clear. Adding return type will complicate future code modifications.

I think it would help make code easier to review though, as reviewers can instantly see whether the return type of function changes, which can sometimes lead to bugs.

gerhean:
Also can you move the files in `json-parser.ts` back into `api.js` for now? Or separate them in another PR first. So that I can actually see the difference in code. Sorry!

Zhou-Jiahao-1998:
@gerhean Ok I have added the return type here. Im not too sure if it really helps in reviewing the code. Creating a special interface here seems unnecessary because it is only for `app.vue` and `v-authorship.vue` to consume. 

Putting `json-parser.ts` back into `api.js` is going to be very hard. I decided to split them up because it was almost impossible to tackle everything in such as big util class at once.

I think if it is too hard for you to review, we can leave it later. I am ok with putting it on hold first.

gerhean:
> Putting json-parser.ts back into api.js is going to be very hard. I decided to split them up because it was almost impossible to tackle everything in such as big util class at once.

Doesn't typescript allow for a mix of typescript and Javascript? You can just add type to half the file and ignore the rest right?

PR #1755, opened 2022-04-10
GitHub Actions: Change node version semantic for Cypress

yhtMinceraft1010X:
## Proposed commit message

```
The node-version semantic used for cypress does not match those in other
jobs.

Let's change the node-version semantic for cypress.
```


gok99 (approved):
LGTM!

gerhean (approved):
LGTM, minor change will merge.

PR #1761, opened 2022-04-22
Frontend Lint: add ts file extension to eslint

yhtMinceraft1010X:
## Proposed commit message

```
The eslint file types do not include ts. This means that lint errors in
ts files such as main.ts will go undetected.

Let's update the lint and lintfix scripts to include the ts file type
for eslint.
```

## Other information

I removed some semicolons in `api.js` and `main.ts` for testing and noticed that with `eslint src/**/*.{js,vue}`, the missing semicolon does not get flagged in `main.ts`:
Screenshot (266)

Using `eslint src/**/*.{ts,js,vue}` works:
Screenshot (267)

Considering that it is reading week, this PR isn't urgent and can be merged after the final exams.


yhtMinceraft1010X:
As a side note, I did not add the `ts` file extension for the `cypress` directory since we currently do not have any `ts` files there.

gok99:
Nice catch!

> As a side note, I did not add the ts file extension for the cypress directory since we currently do not have any ts files there.

Are there downsides to preemptively checking .ts tests?

yhtMinceraft1010X:
> Are there downsides to preemptively checking .ts tests?

I'm not sure about whether we will be adding `.ts` files to cypress in the future. Perhaps it's best to add it when we are sure about adding/migrating cypress files to TypeScript.

gok99:
> I'm not sure about whether we will be adding .ts files to cypress in the future.

I see, I was concerned with forgetting to add it in if/when we do and thought we might want to cover possible future cases with the addition of Typescript at once. Even if we don't end up migrating the tests, checking for the extension doesn't seem to incur significant cost, and in principle it might make sense for eslint to check any file that can possibly be checked rather than only the ones we have currently.

yhtMinceraft1010X:
> Hi, We are going to mark this PR as stale because it has been inactive for the past 30 days. If no further activity occurs within the following 7 days, it will be automatically closed so that others can take up the issue. If you are still working on this PR, please make a follow-up commit within 7 days and leave a comment to remove the stale label. Do let us know if you are stuck so that we can help you!'

I am still working on this

dcshzj:
@yhtMinceraft1010X Is this PR ready? Seems like the test cases pass even when TypeScript files are included in the lint.

yhtMinceraft1010X:
> @yhtMinceraft1010X Is this PR ready? Seems like the test cases pass even when TypeScript files are included in the lint.

I've added the `.ts` extension for cypress lint.

yhtMinceraft1010X:
> I've added the `.ts` extension for cypress lint.

Reverting this because MacOS CI expects TypeScript files to be present for `cypress` directory.

dcshzj (approved):
LGTM!

PR #1763, opened 2022-04-23
GitHub Actions: update deps to latest versions

yhtMinceraft1010X:
## Proposed commit message

```
Many actions that we use have been updated. Most of the updates are
related to the underlying actions upgrading from Node 12 to 16.

Let's use the updated versions in our workflows.
```

### Node version in underlying actions
Although the updated actions use Node 16 in their underlying implementation, it does not appear to affect CI even though we are using Node 14.

### `distribution` for `actions/setup-java`:
The `distribution` used here is `zulu` since that is the [originally supported distribution in `v1`](https://github.com/actions/setup-java#v2-vs-v1). On another note, legacy Java syntax such as `1.8` was also dropped in `v2`.

### DIFF
--- a/.github/workflows/delete-deploy.yml
+++ b/.github/workflows/delete-deploy.yml
@@ -17,7 +17,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - name: Delete deployments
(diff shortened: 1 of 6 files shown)


dcshzj (approved):

yhtMinceraft1010X:
Re-requesting review since I updated the version for `cypress-io/github-action`

dcshzj (approved):

PR #1766, opened 2022-04-28
Frontend: change JavaScript line length limit for ESLint

yhtMinceraft1010X:
## Proposed commit message

```
There is a mismatch between the JavaScript line length limit allowed
in .eslintrc.json and the hard limit in the coding standard.

Let's change the allowed limit in .eslintrc.json and update the files
that fail the check upon this change.
```

## Other information

Reference: https://se-education.org/guides/conventions/javascript.html


dcshzj (approved):
LGTM!

PR #1767, opened 2022-04-29
[#1764] Actions: refactor Cypress CI build step

### LINKED ISSUE #1764: Cypress CI: Install packages step gives warning

1. Go to `Cypress frontend tests` in a recent integration workflow run
1. Select `Install packages`

No warning is thrown.

A warning is thrown which states that there is no file or directory for `/home/runner/work/RepoSense/RepoSense/package.json`:
Screenshot (271)

This warning should be addressed since it will be raised as an error (causing CI to fail) if Node 16 is used. Since Node 14 will reach end-of-life next year, we will eventually have to upgrade to Node 16. We should address this before that day.

An example is shown in the [integration workflow for the PR that updates package dependencies](https://github.com/reposense/RepoSense/runs/6140207197?check_suite_focus=true), which is supposed to be a step towards the eventual upgrade to Node 16.

yhtMinceraft1010X:
Fixes #1764 

## Proposed commit message

```
The Install packages step looks for a non-existent package.json file.
The Set up Cypress step is basically a repeat of installCypress within
build.gradle, which runs before frontendTest task.

Let's remove both steps from Cypress CI.
```

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -268,18 +268,6 @@ jobs:
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
 
-    - name: Install packages
-      run: npm install
-
-    - name: Set up Cypress
-      uses: cypress-io/github-action@v2
-      with:
-        build: npm i -D cypress
-        working-directory: ./frontend/cypress
-        browser: chrome
-        headless: true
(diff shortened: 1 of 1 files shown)


dcshzj (approved):
LGTM! Great catch. Are you also able to update the PR to resolve the conflicts?

PR #1776, opened 2022-05-07
[#1774] Fix unintended behavior related to --since d1

### LINKED ISSUE #1774: GitHub Actions: Failing CI for system tests and frontend tests

* **RepoSense Version:** varies
* **OS and Version:** Any OS

1. Re-run CI for `ubuntu-18.04 JDK 8` (can substitute with any other OS) and `Cypress frontend tests` (Example given [here](https://github.com/reposense/RepoSense/runs/6322483468?check_suite_focus=true))

All workflow runs should pass.

* The system tests fail like [this](https://github.com/reposense/RepoSense/runs/6322483468?check_suite_focus=true#step:11:1455) whereas the frontend tests fail like [this](https://github.com/reposense/RepoSense/runs/6322483061?check_suite_focus=true#step:10:412)
* In some cases, the `authorship.json` file is checked before the `commits.json` file (which reveals a similar error)
* All these tests passed on local systems and only fail in GitHub Actions workflows
* I suspect it has something to do with how `--since d1` works in GitHub Actions since all the failing tests happen to use that. This may have resulted in certain commit and file info not being included.

### LINKED ISSUE #1778: Combining --since d1 with --period gives invalid date ranges

* **RepoSense Version:** Latest commit in `master` branch as well as a branch for fixing CI fails related to `--since d1`

1. (For developers) Run `gradlew run -Dargs="--config frontend/cypress/config --since d1 --period 30d --view" (Equivalent to `java -jar RepoSense.jar --config frontend/cypress/config --since d1 --period 30d --view`)
1. Wait for report to come up

The report has a valid date range (from 2018-05-03 to 2018-06-02) and data for commits within that date range.

For `master` branch: there are no dates. It is just `yyyy-mm-dd`:
Screenshot (283)

For the branch which I am using to fix some CI fails, the date range is 1970-01-01 to 1970-01-31:
Screenshot (284)

yhtMinceraft1010X:
Fixes #1774.
Fixes #1778.

## Proposed commit message

```
The use of Instant.ofEpochMilli(Long.MIN_VALUE) as the arbitrary first
commit date now fails shallow cloning for
testSinceBeginningDateRangeWithShallowCloning system test.

In addition, using this arbitrary date in convertToGitDateRangeArgs
when converting timezone results in an overflow causing sinceDate to
be greater than untilDate. Thus, no commit results are retrieved,
leading to failing system tests.

Also, the arbitrary first commit date is in UTC. However, in the
convertToGitDateRangeArgs method, the zoneId used for the conversion
is from the config, which may be different. This leads to an erroneous
timezone conversion.

Finally, using --since d1 with --period results in an unintended date
range because the --period and the arbitrary commit date from 
--since d1 are used to calculate the until date as the actual 
earliest commit date is not immediately available.

Let's change Instant.ofEpochMilli() to use 0 so that the arbitrary date
is 1970-01-01 and the overflow of the sinceDate can be avoided.

At the same time, let's convert the arbitrary date to the correct
timezone within the ArgsParser class and add a component test for
parsing --since d1 to check that the correct timezone is used for each
RepoConfiguration before cloning any repos.

Finally, let's raise a warning when --since d1 is being used with 
--period.
```

## Other information

Screenshot showing the overflow of the sinceDate when `Instant.ofEpochMilli(Long.MIN_VALUE)` is used:
Screenshot (277)

Notice how the sinceDate has `+` sign instead of `-` sign.

## Timezone Trivia
There also seems to be something strange with the timezone here:
Screenshot (278)

Although the code for this method has been changed, the timezone conversion logic in lines 60-61 is still preserved. The time appears to be converted to +07:30 instead of +08:00. However, tests still work correctly. This is likely following the [old SG timezone being used before 1 January 1982](https://en.wikipedia.org/wiki/Singapore_Standard_Time).

### DIFF
--- a/docs/ug/cli.md
+++ b/docs/ug/cli.md
@@ -172,7 +172,7 @@ This flag overrides the `Ignore file size limit` field in the CSV config file.
 <box type="info" seamless>
 
 * If both start date and end date are not specified, the date of generating the report will be taken as the end date.
-* May analyse the incorrect date range if used with `--since d1`. The program will throw a warning.
+* May analyze the incorrect date range if used with `--since d1`. The program will throw a warning.
 * Cannot be used with both `--since` and `--until`. The program will throw an exception.
 </box>
 <!-- ------------------------------------------------------------------------------------------------------ -->
@@ -218,7 +218,7 @@ Cannot be used with `--last-modified-date`. This may result in an incorrect last
 
 * If the start date is not specified, only commits made one month before the end date (if specified) or the date of generating the report, will be captured and analyzed.
 * If `d1` is specified as the start date (`--since d1` or `-s d1`), then the program will search for the earliest commit date of all repositories and use that as the start date.
-* If `d1` is specified together with `--period`, then the program will warn that the date range being analysed may be incorrect.
+* If `d1` is specified together with `--period`, then the program will warn that the date range being analyzed may be incorrect.
 </box>
 <!-- ------------------------------------------------------------------------------------------------------ -->
 
--- a/docs/ug/cli.md
+++ b/docs/ug/cli.md
@@ -172,6 +172,7 @@ This flag overrides the `Ignore file size limit` field in the CSV config file.
 <box type="info" seamless>
 
 * If both start date and end date are not specified, the date of generating the report will be taken as the end date.
+* May analyse the incorrect date range if used with `--since d1`. The program will throw a warning.
 * Cannot be used with both `--since` and `--until`. The program will throw an exception.
 </box>
 <!-- ------------------------------------------------------------------------------------------------------ -->
@@ -216,7 +217,8 @@ Cannot be used with `--last-modified-date`. This may result in an incorrect last
 <box type="info" seamless>
 
 * If the start date is not specified, only commits made one month before the end date (if specified) or the date of generating the report, will be captured and analyzed.
-* If `d1` is specified as the start date (`--since d1` or `-s d1`), then the earliest commit date of all repositories will be taken as the start date.
+* If `d1` is specified as the start date (`--since d1` or `-s d1`), then the program will search for the earliest commit date of all repositories and use that as the start date.
+* If `d1` is specified together with `--period`, then the program will warn that the date range being analysed may be incorrect.
 </box>
 <!-- ------------------------------------------------------------------------------------------------------ -->
 
--- a/src/main/java/reposense/commits/CommitResultAggregator.java
+++ b/src/main/java/reposense/commits/CommitResultAggregator.java
@@ -14,8 +14,8 @@ import reposense.commits.model.CommitContributionSummary;
 import reposense.commits.model.CommitResult;
 import reposense.model.Author;
 import reposense.model.RepoConfiguration;
-import reposense.parser.SinceDateArgumentType;
 import reposense.report.ReportGenerator;
+import reposense.util.TimeUtil;
 
 /**
  * Uses the commit analysis results to generate the summary information of a repository.
@@ -31,7 +31,7 @@ public class CommitResultAggregator {
             RepoConfiguration config, List<CommitResult> commitResults) {
         LocalDateTime startDate;
         ZoneId zoneId = ZoneId.of(config.getZoneId());
-        startDate = (config.getSinceDate().equals(SinceDateArgumentType.getArbitraryFirstCommitDateConverted(zoneId)))
+        startDate = (TimeUtil.isEqualToArbitraryFirstDateConverted(config.getSinceDate(), zoneId))
                 ? getStartOfDate(getStartDate(commitResults, zoneId), zoneId)
                 : config.getSinceDate();
         ReportGenerator.setEarliestSinceDate(startDate);
@@ -159,7 +159,7 @@ public class CommitResultAggregator {
      * timezone given by {@code zoneId}. Otherwise, return a {@link LocalDateTime} adjusted to have a time of 00:00:00.
      */
     private static LocalDateTime getStartOfDate(LocalDateTime current, ZoneId zoneId) {
-        if (current.equals(SinceDateArgumentType.getArbitraryFirstCommitDateConverted(zoneId))) {
+        if (TimeUtil.isEqualToArbitraryFirstDateConverted(current, zoneId)) {
             return current;
         }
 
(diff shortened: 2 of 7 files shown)


src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -157,11 +155,11 @@ private static void addDailyContributionForNewDate(
     /**
      * Gets the starting point of the {@code current} date.
      *
-     * @return the {@code current} date if it is equal to the {@code ARBITRARY_FIRST_COMMIT_DATE_UTC} adjusted to the
+     * @return the {@code current} date if it is equal to the {@code ARBITRARY_FIRST_COMMIT_DATE} adjusted to the
      * timezone given by {@code zoneId}. Otherwise, return a {@link LocalDateTime} adjusted to have a time of 00:00:00.
      */
     private static LocalDateTime getStartOfDate(LocalDateTime current, ZoneId zoneId) {
-        if (current.equals(ARBITRARY_FIRST_COMMIT_DATE_UTC.withZoneSameInstant(zoneId).toLocalDateTime())) {
+        if (current.equals(SinceDateArgumentType.getArbitraryFirstCommitDateConverted(zoneId))) {
```
gok99, line 162: This check seems to be performed a lot, would it make sense to create TimeUtil::isArbitratyFirstCommitDate[UtcLocal] or something to that effect?

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -290,6 +290,10 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
 
             if (isSinceDateProvided) {
                 sinceDate = TimeUtil.getSinceDate(cliSinceDate.get());
+                // For --since d1, need to adjust the arbitrary date based on timezone
+                if (sinceDate.equals(SinceDateArgumentType.getArbitraryFirstCommitDateUtcLocal())) {
```
gok99, line 294: Could this logic go into TimeUtil::getSinceDate?

gok99 (commented):
Thanks for identifying the issue! Looks good to me, except for some minor questions.

src/main/java/reposense/parser/ArgsParser.java
gok99: Would just like to get this clarified too - there seems to be an identical clause in `getSinceDate`. If it can't/shouldn't be moved into `getSinceDate`, I'll approve the PR.

yhtMinceraft1010X: Now that you've mentioned it, I think the clause in `TimeUtil.getSinceDate` can be safely removed. Since the new arbitrary date is `1970-01-01 00:00:00` in UTC time, adding `withHour(0).withMinute(0).withSecond(0)` does nothing to it so it is effectively returning the same date. The equality check in `ArgsParser` should still pass.

That being said, we still have to keep `withHour(0).withMinute(0).withSecond(0)` in `TimeUtil.getSinceDate` since the `getDateMinusAMonth` and `getDateMinusNDays` methods still need to convert the since date to midnight.

yhtMinceraft1010X:
For #1778, I decided to use a warning instead of throwing an exception and terminating the program before cloning. This is due to the possibility that someone might put `--since 1/1/1970 --period 30d`. In this case, an exception would be thrown since the given date would be equal to the arbitrary date that is used to check for whether --since d1 might be used and no cloning would be done.

Although this is an unlikely scenario, I believe no exception should be thrown since setting 31/1/1970 as the until date is expected behaviour.

docs/ug/cli.md
```diff
@@ -172,6 +172,7 @@ This flag overrides the `Ignore file size limit` field in the CSV config file.
 <box type="info" seamless>
 
 * If both start date and end date are not specified, the date of generating the report will be taken as the end date.
+* May analyse the incorrect date range if used with `--since d1`. The program will throw a warning.
```
dcshzj, line 175:
```suggestion
* May analyze the incorrect date range if used with `--since d1`. The program will throw a warning.
```

docs/ug/cli.md
```diff
@@ -216,7 +217,8 @@ Cannot be used with `--last-modified-date`. This may result in an incorrect last
 <box type="info" seamless>
 
 * If the start date is not specified, only commits made one month before the end date (if specified) or the date of generating the report, will be captured and analyzed.
-* If `d1` is specified as the start date (`--since d1` or `-s d1`), then the earliest commit date of all repositories will be taken as the start date.
+* If `d1` is specified as the start date (`--since d1` or `-s d1`), then the program will search for the earliest commit date of all repositories and use that as the start date.
+* If `d1` is specified together with `--period`, then the program will warn that the date range being analysed may be incorrect.
```
dcshzj, line 221:
```suggestion
* If `d1` is specified together with `--period`, then the program will warn that the date range being analyzed may be incorrect.
```

src/main/java/reposense/parser/SinceDateArgumentType.java
```diff
@@ -16,14 +17,17 @@
...
      * When user specifies "d1", arbitrary first commit date will be returned.
+     * This date is equivalent to 1970-01-01 00:00:00 in UTC time.
      * Then, ReportGenerator will replace the arbitrary since date with the earliest commit date.
      */
-    public static final LocalDateTime ARBITRARY_FIRST_COMMIT_DATE = LocalDateTime.ofInstant(
-            Instant.ofEpochMilli(Long.MIN_VALUE), ZoneId.of("Z"));
     public static final String FIRST_COMMIT_DATE_SHORTHAND = "d1";
+    private static final ZonedDateTime ARBITRARY_FIRST_COMMIT_DATE = ZonedDateTime.ofInstant(
+            Instant.ofEpochMilli(0), ZoneId.of("Z"));
+    private static final LocalDateTime ARBITRARY_FIRST_COMMIT_DATE_UTC_LOCAL = ARBITRARY_FIRST_COMMIT_DATE
```
dcshzj, line 26: The variable naming seems a little confusing. May I suggest using `ARBITRARY_FIRST_COMMIT_DATE_UTC` and `ARBITRARY_FIRST_COMMIT_DATE_LOCAL` instead?

dcshzj (changes requested):

dcshzj (approved):
LGTM!

Issue #1769, opened 2022-04-29
Implement more Checkstyle checks


yhtMinceraft1010X:
**What feature(s) would you like to see in RepoSense?**

While working on https://github.com/se-edu/addressbook-level3/pull/133, I found additional checks that could be useful for us:
* [JavadocTagContinuationIndentation](https://checkstyle.sourceforge.io/config_javadoc.html#JavadocTagContinuationIndentation) - Check for Google Java Style Guide requirement of [4 whitespaces for continuation lines in block tags](https://google.github.io/styleguide/javaguide.html#s7.1.3-javadoc-block-tags)
* [DefaultComesLast](https://checkstyle.sourceforge.io/config_coding.html#DefaultComesLast) - Ensures that all cases in `switch` statement are before `default`
* [NoWhitespaceBeforeCaseDefaultColon](https://checkstyle.sourceforge.io/config_whitespace.html#NoWhitespaceBeforeCaseDefaultColon) - E.g. `default :` would not be allowed

**If possible, describe the solution**

Before implementing any new checks, we should look through [other available checks](https://checkstyle.sourceforge.io/checks.html) and see which checks are suitable. By having as many checks as possible in a single PR, this will allow us to avoid revisiting this issue with multiple PRs in the future.

damithc:
Looks like the bot auto-closed this. Should it be closed?

yhtMinceraft1010X:
I plan to keep this open since there may be more checks that we need to implement.

PR #1781, opened 2022-05-29
Checkstyle: add MissingJavadocType check and update comments

yhtMinceraft1010X:
Part of #1769 

## Proposed commit message

```
The oss-generic link for the coding standard is obsolete. At the same
time, the description vaguely refers to a module. The checkstyle file 
also references our own standards, which should be mentioned explicitly.

The JavadocType check actually checks that the Javadoc for classes,
enums and interfaces are of the correct format. The MissingJavadocType
check should be used instead to check that the public classes, enums and
interfaces have Javadoc.

Finally, the phrase "format's Javadoc" for the comment on the
JavadocMethod check does not make sense.

Let's change the coding standard link to the one used by SE-EDU, remove
the reference to the module and add the link to our style guide.

Let's also update the comments for the checks and add the
MissingJavadocType check.
```

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -402,6 +402,9 @@
       <property name="tokens" value="METHOD_DEF, ANNOTATION_FIELD_DEF"/>
     </module>
 
+    <!-- Checks that every public class, enumeration and interface has a header comment. -->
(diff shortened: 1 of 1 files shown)


dcshzj (approved):
LGTM!

PR #1762, opened 2022-04-23
Frontend: update Node packages and fix code for checks

yhtMinceraft1010X:
Part of #1703 

## Proposed commit message

```
Many dependencies are not up to date with the latest versions. This
gives rise to vulnerabilities as well as dependency conflicts when
attempting to run npm update.

In addition, there are ESLint checks for Airbnb JS standards which fail
upon updating these dependencies.

Let's update the dependencies and refactor the Javascript code
accordingly.
```

### New argument format for `CopyWebpackPlugin` in `vue.config.js`
With the upgrading of the dependencies, Webpack has been bumped to version 5. In particular, `ignore` is now under `globOptions` as illustrated in [this format](https://webpack.js.org/plugins/copy-webpack-plugin/#ignoring-files):
```
new CopyPlugin({
      patterns: [
        {
          from: "public/**/*",
          globOptions: {
            dot: true,
            gitignore: true,
            ignore: ["**/file.*", "**/ignored-directory/**"],
          },
        },
      ],
    }),
```
The change in `vue.config.js` is to allow `'*.json'` to be passed in through the new method. Otherwise, the original method would result in an error.

### Removal of automatic syntax inference in Stylelint from `v14` onwards
As mentioned in the Stylelint [migration guide for 14](https://stylelint.io/migration-guide/to-14/), Stylelint no longer has support for SCSS and Vue syntaxes. The `.stylelintrc.json` file has to be extended with `stylelint-config-recommended-scss` and `stylelint-config-recommended-vue`, without which errors like these can happen (Some other files that have this error are not shown):

Screenshot (269)

### Additional ESLint checks:
The Javascript code has been updated for the following checks:
* [`no-mixed-operators`](https://eslint.org/docs/rules/no-mixed-operators)
* [`function-paren-newline`](https://eslint.org/docs/rules/function-paren-newline)
* [`no-multiple-empty-lines`](https://eslint.org/docs/rules/no-multiple-empty-lines)

The above 3 checks can be found in the [Airbnb style guide](https://github.com/airbnb/javascript/blob/master/README.md). There is also one more check ([`prefer-object-spread`](https://eslint.org/docs/rules/prefer-object-spread)) that the guide refers to but I am not sure about whether to enable it since the eslint website mentions not to if our code does not support ES2018. An example of a line that fails this check is given below:
https://github.com/reposense/RepoSense/blob/b12d3485d4b7a3aca5f2545404b5545a5277ff5e/frontend/src/views/v-zoom.vue#L177

[...]

PR #1786, opened 2022-06-07
[#1765] Change npm version requirement to Node.js

### LINKED ISSUE #1765: Developer Guide: Change npm requirement to Node
**What would you like added or modified in the documentation?**

In the [Setting Up page](https://reposense.org/RepoSense/dg/settingUp.html) of the Developer Guide, the requirements are for `git`, `npm` and `JDK`. I am of the opinion that using `Node.js` in place of `npm` is better since `npm` is usually installed as part of `Node.js`. Other NUS-OSS projects also describe `Node.js` as a requirement, rather than `npm` by itself.

By using `Node.js`, it should also be easier to maintain the Developer Guide since the version can simply be updated whenever an older version reaches EOL, unlike `npm` where cross-referencing of the version against `Node.js` version is required.

yhtMinceraft1010X:
Fixes #1765 

## Proposed commit message

```
Node.js is often installed together with npm. With regards to tracking
version LTS and EOL, Node.js is often used instead of npm. Thus, it is
easier to update the Node.js version in the DG whenever the existing
version goes EOL and we upgrade to a newer version.

On the other hand, when updating the corresponding npm version, there
is a need to look for the Node.js version and then find the
corresponding npm version.

Let's change the DG pre-requisites to reference the Node.js version and
specify that it is for npm. This improves maintainability of the DG.
```


dcshzj:
@yhtMinceraft1010X Were you able to verify if newer versions of Node.js are able to work, at least for Node.js 16?

yhtMinceraft1010X:
I tested Node 16 with the existing CI. It did not work according to #1764 but merging #1767 should solve this.

yhtMinceraft1010X:
> I tested Node 16 with the existing CI. It did not work according to #1764 but merging #1767 should solve this.

To add on, #1762 should also be merged before Node 16 can be used. I remember trying to use Node 16 directly but failing due to dependency conflicts that are addressed by #1762

yhtMinceraft1010X:
I have tested it out using Node 16. It works so I have updated the Node requirement accordingly.

dcshzj (approved):
LGTM!

PR #1790, opened 2022-06-17
[#1674] Convert zoneId String to ZoneId Object

### LINKED ISSUE #1674: Convert zoneId String in RepoConfiguration and SummaryJson to ZoneId object

I noticed that although `zoneId` is stored as a `ZoneId` object in `CliArguments.java`, it is then converted to a `String` when stored in `RepoConfiguration.java`, subsequently, when `zoneId` needs to be used in places like `FileInfoAnalyzer.java`, it has to do the conversion back to `ZoneId`. 

This conversion from `ZoneId` to `String` to `ZoneId` is redundant. In my opinion, it is also quite strange to have `zoneId` be a `String`, when the variables that `zoneId` is commonly used with, `sinceDate` and `untilDate`, are `LocalDateTime` objects.

Since `SummaryJson` objects for repos are written into Json files during tests, changing `zoneId` to be `ZoneId` instead of `String` may cause the Json output to be different from expected Json files, potentially breaking the tests.
* If the above scenario happens, it is okay to keep the `zoneId` variable as a `String` in `SummaryJson`.
* However, the conversion from `String` to `ZoneId` in `SummaryJson` should be attempted and tested first before reverting back if tests are broken.
* Recommended approach:
  1. Follow step 2 of the [**Learning the Basics** guide for backend](https://reposense.org/RepoSense/dg/learningBasics.html#backend). Throughout this step, take note of how the `zoneId` variable is used in the code.
  1. Convert `zoneId` variable from `String` to `ZoneId` in `RepoConfiguration.java` and propagate the change to any classes and methods (inclusive of test code) that break after this change.
  1. Repeat the above steps for `SummaryJson.java`.
  1. Run tests and checkstyle. Commit these changes if successful.

**Additional context**

This should be a good first-timer issue for future contributors such as CP3108A/B students. The tracing of the `zoneId` variable and the subsequent refactoring should provide enough exposure to the backend code.

SkyBlaise99:
Fixes #1674

## Proposed commit message

```
zoneId is stored as ZoneId object in CliArguments.java. It is converted 
into String when stored in RepoConfiguration.java. Subsequently, it is 
then converted back to ZoneId when required in FileInfoAnalyzer.java or 
other places.

Such repeated conversion is redundant.

Let's standardize and save zoneId as ZoneId Object across all files.
```

## Other information

1. `zoneId` in `SummaryJson.java` remains as `String` as changing to `ZoneId` will break `systemtest`. Instead, the constructor will now accept a `ZoneId` input which will subsequent be converted and stored as `String`.

### DIFF
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -563,6 +563,10 @@ public class RepoConfiguration {
         this.zoneId = zoneId;
     }
 
+    public void setZoneId(String zoneId) {
+        this.zoneId = ZoneId.of(zoneId);
+    }
+
     public void setAuthorDisplayName(Author author, String displayName) {
         authorConfig.setAuthorDisplayName(author, displayName);
     }
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -563,10 +563,6 @@ public class RepoConfiguration {
         this.zoneId = zoneId;
     }
 
-    public void setZoneId(String zoneId) {
-        this.zoneId = ZoneId.of(zoneId);
-    }
-
     public void setAuthorDisplayName(Author author, String displayName) {
         authorConfig.setAuthorDisplayName(author, displayName);
     }
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -2,6 +2,7 @@ package reposense.model;
 
 import java.io.File;
 import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -26,7 +27,7 @@ public class RepoConfiguration {
     private String branch;
     private String displayName;
     private String outputFolderName;
-    private transient String zoneId;
+    private transient ZoneId zoneId;
     private transient LocalDateTime sinceDate;
     private transient LocalDateTime untilDate;
     private transient String repoFolderName;
@@ -102,7 +103,7 @@ public class RepoConfiguration {
         }
     }
 
-    public static void setZoneIdToRepoConfigs(List<RepoConfiguration> configs, String zoneId) {
+    public static void setZoneIdToRepoConfigs(List<RepoConfiguration> configs, ZoneId zoneId) {
         for (RepoConfiguration config : configs) {
             config.setZoneId(zoneId);
         }
@@ -554,11 +555,11 @@ public class RepoConfiguration {
         this.untilDate = untilDate;
     }
 
-    public String getZoneId() {
+    public ZoneId getZoneId() {
         return zoneId;
     }
 
-    public void setZoneId(String zoneId) {
+    public void setZoneId(ZoneId zoneId) {
         this.zoneId = zoneId;
     }
 
--- a/src/main/java/reposense/report/SummaryJson.java
+++ b/src/main/java/reposense/report/SummaryJson.java
@@ -1,7 +1,6 @@
 package reposense.report;
 
 import java.time.LocalDateTime;
-import java.time.ZoneId;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -31,7 +30,7 @@ public class SummaryJson {
 
     public SummaryJson(List<RepoConfiguration> repos, ReportConfiguration reportConfig, String reportGeneratedTime,
             LocalDateTime sinceDate, LocalDateTime untilDate, boolean isSinceDateProvided, boolean isUntilDateProvided,
-            String repoSenseVersion, Set<Map<String, String>> errorSet, String reportGenerationTime, ZoneId zoneId) {
+            String repoSenseVersion, Set<Map<String, String>> errorSet, String reportGenerationTime, String zoneId) {
         this.repos = repos;
         this.reportGeneratedTime = reportGeneratedTime;
         this.reportGenerationTime = reportGenerationTime;
@@ -42,7 +41,7 @@ public class SummaryJson {
         this.isUntilDateProvided = isUntilDateProvided;
         this.repoSenseVersion = repoSenseVersion;
         this.errorSet = errorSet;
-        this.zoneId = zoneId.toString();
+        this.zoneId = zoneId;
         this.supportedDomainUrlMap = SupportedDomainUrlMap.getDefaultDomainUrlMap();
     }
 }
--- a/src/main/java/reposense/report/SummaryJson.java
+++ b/src/main/java/reposense/report/SummaryJson.java
@@ -1,6 +1,7 @@
 package reposense.report;
 
 import java.time.LocalDateTime;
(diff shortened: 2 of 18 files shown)


SkyBlaise99:
I should have kept it as a Draft is it? Sorry in advance if this is not the intended workflow.

src/main/java/reposense/report/SummaryJson.java
```diff
@@ -30,7 +31,7 @@ public class SummaryJson {
 
     public SummaryJson(List<RepoConfiguration> repos, ReportConfiguration reportConfig, String reportGeneratedTime,
             LocalDateTime sinceDate, LocalDateTime untilDate, boolean isSinceDateProvided, boolean isUntilDateProvided,
-            String repoSenseVersion, Set<Map<String, String>> errorSet, String reportGenerationTime, String zoneId) {
+            String repoSenseVersion, Set<Map<String, String>> errorSet, String reportGenerationTime, ZoneId zoneId) {
```
yhtMinceraft1010X, line 34: Since you mentioned that the system tests would break if `ZoneId` type were used in `SummaryJson`, it's okay to leave the input type as a `String` since the `zoneId` is unlikely to be converted to another type once it has been passed into `SummaryJson`.

src/main/java/reposense/report/SummaryJson.java
```diff
@@ -41,7 +42,7 @@ public SummaryJson(List<RepoConfiguration> repos, ReportConfiguration reportConf
         this.isUntilDateProvided = isUntilDateProvided;
         this.repoSenseVersion = repoSenseVersion;
         this.errorSet = errorSet;
-        this.zoneId = zoneId;
+        this.zoneId = zoneId.toString();
```
yhtMinceraft1010X, line 45: Related to the above comment, this `toString()` conversion looks out of place relative to the other variable initializations.

yhtMinceraft1010X (changes requested):
Excellent work! There is one requested change with respect to `SummaryJson`.

yhtMinceraft1010X:
> I should have kept it as a Draft is it? Sorry in advance if this is not the intended workflow.

The `Draft` feature is optional. If you are confident that the PR is ready for review, you can leave it as `Open`.

src/main/java/reposense/report/SummaryJson.java
SkyBlaise99: My idea is that if we allow `String` input, then we should validate it in the constructor? I believe you also mentioned this in the issue.

My proposed solution is therefore to directly accept a `ZoneId` object as input and hence the validation part is somewhat handled within `ZoneId`. The downside is probably what you mentioned below, that it kinda out of place compared to others.

Would you still like me to change it back to the original version? If so, what about the validation part?

SkyBlaise99: @yhtMinceraft1010X What do you think about this?

src/main/java/reposense/report/SummaryJson.java
SkyBlaise99: Actually I took a look, there are also other variables similar to this

``` java
this.reportTitle = reportConfig.getTitle();
// and
this.supportedDomainUrlMap = SupportedDomainUrlMap.getDefaultDomainUrlMap();
```

SkyBlaise99:
Btw the Codecov isn't passing.

image

But I feel like `-0.03%` isn't a lot. What would be the usual way to deal with this?

yhtMinceraft1010X:
https://github.com/reposense/RepoSense/issues/1674#issuecomment-1157444841

With regards to your comment on the setters, you can keep the original `String` setter.

yhtMinceraft1010X:
> Btw the Codecov isn't passing.

This is usually ignored unless the difference is significant.

src/main/java/reposense/report/SummaryJson.java
SkyBlaise99: Switched back to `this.zoneId = zoneId;`

src/main/java/reposense/report/SummaryJson.java
SkyBlaise99: Switched back to `String zoneId`

yhtMinceraft1010X (approved):
LGTM from my POV. I'll request for the next reviewer.

SkyBlaise99:
Hi @dcshzj, can you help to review this PR? Thank you!

src/main/java/reposense/report/SummaryJson.java
dcshzj: I would like to revisit this, I think I agree with @SkyBlaise99 here, it would be a better design to ensure that the ZoneId object is passed into the constructor here, rather than the String value. The `zoneId` variable in the `SummaryJson` object should also change type to ZoneId.

Can I check what errors you are getting with the systemtests? Is it potentially because we are lacking a serializer to convert the ZoneId object to a string and vice versa?

dcshzj (changes requested):
Left a comment above, the rest LGTM.

src/main/java/reposense/report/SummaryJson.java
SkyBlaise99: The error in `systemtest` should be due to the change in json file output.

Previously the json file is as such:
``` json
{
    "reportGeneratedTime":"Mon Jul 11 11:31:42 2022 SGT",
    "reportGenerationTime":" 23.02 second(s)",
    "zoneId":"Asia/Singapore",
    "rest of the data":"..."
}
```

After using `ZondId` object type:
``` json
    {"reportGeneratedTime":"Mon Jul 11 11:35:16 2022 SGT",
    "reportGenerationTime":" 22.39 second(s)",
    "zoneId":{"id":"Asia/Singapore"},
    "rest of the data":"..."
}
```

There is now an extra bracket there which I'm not sure if it will affect frontend side.

dcshzj: I see, I think it is still preferable that we take in a ZoneId object, as well as store it as a ZoneId. The json file is printed slightly differently as we have not defined a custom serializer. We currently do that for `LocalDateTime` and `FileType`:

https://github.com/reposense/RepoSense/blob/40d87be077f56fdfdc7b1ca40b56449b3a6bbff0/src/main/java/reposense/util/FileUtil.java#L104-L108

You can refer to the way the serializer is implemented in the `FileType` class, and I believe it should be possible to resolve the system test failures.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -554,14 +555,18 @@ public void setUntilDate(LocalDateTime untilDate) {
...
 
-    public void setZoneId(String zoneId) {
+    public void setZoneId(ZoneId zoneId) {
         this.zoneId = zoneId;
     }
 
+    public void setZoneId(String zoneId) {
+        this.zoneId = ZoneId.of(zoneId);
+    }
+
```
dcshzj, line 569: I think this function can be removed after removing the usages in `CommitInfoAnalyzerTest.java`.

dcshzj (changes requested):

dcshzj (approved):
LGTM! Great work!

PR #1796, opened 2022-06-24
[#1794] Allow more date formats

### LINKED ISSUE #1794: Allow other date formats

Currently the only supported date format is `d/M/yyyy`. We can allow other different formats such as `d-M-yyyy`, `d.M.yyyy`, or even reverse the order (e.g. `yyyy/M/d`).

[`DATE_FORMAT_REGEX`](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/util/TimeUtil.java#L19-L20) and [`extractDate()`](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/util/TimeUtil.java#L165-L175) can be modified to allow more formats.

**Additional context**

You can refer to [this](https://github.com/reposense/RepoSense/issues/1789) discussion and this [PR](https://github.com/SkyBlaise99/RepoSense/pull/2).

SkyBlaise99:
Fixes #1794 

## Proposed commit message

```
Different users have different preferences when it comes to the format 
of date. Currently, the accepted format for date is 'dd/mm/yyyy', with 
'd/m/yyyy' also accepted as an un-advertised feature.

It will be nice to loosen this constraint and allow more date formats. 
For example, 'dd-mm-yyyy', 'dd.mm.yyyy', or even 'yyyy/mm/dd'.

Let's allow more date formats.
```

### DIFF
--- a/src/main/java/reposense/util/TimeUtil.java
+++ b/src/main/java/reposense/util/TimeUtil.java
@@ -5,6 +5,9 @@ import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
 import java.time.format.ResolverStyle;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


damithc:
> It will be nice to loosen this constraint and allow more date formats. 
> For example, 'dd-mm-yyyy', 'dd.mm.yyyy', or even 'yyyy/mm/dd'.

@SkyBlaise99 Personally, I don't think we need to support these. The added code/documentation maintenance burden is not worth the incremental benefit to the user. In fact, it may not be a benefit to the user at all; instead of learning and following one date format, the user will now have to know about multiple date formats we support which is extra reading for the user. 

In general, to avoid feature bloat, we should be very careful in adding new features. Each features increase our maintenance burden forever. e.g., adding a new test will delay the CI result for _every_ CI run in the future. 

What do you (and others) think?

SkyBlaise99:
Hi prof,

I suggest we can add this in as an un-advertised feature, like what you mentioned about the single-digit dates. In this way, the burden on documentation is lightened. Users also do not need to learn so many formats and will only learn `dd/mm/yyyy` if need to.

In fact, the additional formats are, in my opinion, targeted toward users who do not read the user guide and jump straight into using the software. In that case, they aren't really sure about the date formats and will use what they prefer. Accepting more date formats will allow them to use a format that they are more familiar with, and also accepted by the program, instead of forcing them to stick to `dd/mm/yyyy`.

For users who read through the user guide, they will only see `dd/mm/yyyy` and learn that, hence no extra burden or confusion in the acceptable formats.

Regarding testing, the part I modified is only `extractDate()`, the ultimate decision is still done in `parseDate()`. I would say this will not increase the load on testing by a lot as compared to the rest.

damithc:
@SkyBlaise99 Un-advertised certainly means less documentation burden. Nevertheless, let's not add support for other date formats as the value of doing so is very low. We need to be brutal in keeping low value features out so that we can keep the product (and the codebase) lean. But OK to support single digits for day/month because it's a legit mistake a user can make and hard to spot later. In contrast, incorrectly using `-` when `/` was expected is easy to spot and correct. Hopefully, we give an error message for such cases informing the user the date format is not as expected (do we?).

SkyBlaise99:
Alright then. Do I still keep this here, in case it may be needed in the future?

PR #1798, opened 2022-06-24
[#452] RepoConfiguration: Restrict sinceDate and untilDate

### LINKED ISSUE #452: GitLogTest fails due to system time >= 2019
The gitLog_sinceDateInFuture_noContent methoed fails at the latest master commit.
`[#316] Split RepoConfiguration into more classes (#425)`
image
image

bingsen0806:
Fixes #452 

```
The flags sinceDate and untilDate are not checked to ensure they are
between 1970/01/01 00:00:00UTC and 2099/12/31 23:59:59UTC (inclusive)

Git is unable to handle timestamps before the Unix epoch and after 2099.
This had caused GitLogTest unit tests to fail unexpected, for example showing
logs of all commits when untilDate is set to 1950. Upon investigation, this
problem applies across git commands, and not only for git log.

Therefore, sinceDate and untilDate should always range between 1970UTC
and 2099UTC for git commands to function expectedly.

Let's create a new method in TimeUtils to get a validated time. If the
input time does not lie in the range, we'll log a warning and then reset
the date to the latest or earliest permitted time. This function will
be used when first parsing CLI arguments, and also in the getSinceDate()
and getUntilDate() in RepoConfiguration to ensure the validity of the two
dates everywhere.
```

### DIFF
--- a/src/main/java/reposense/model/RepoConfiguration.java
+++ b/src/main/java/reposense/model/RepoConfiguration.java
@@ -2,6 +2,7 @@ package reposense.model;
 
 import java.io.File;
 import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


damithc:
@bingsen0806 closing this as you already have a PR for a first-timer issue. Feel free to close the other one and reopen this if you want.

Issue #1726, opened 2022-03-21
Tracking tasks related to supporting other types of remote repo URLs in PR#1644


chan-j-d:
The following are a list of necessary tasks related to but not implemented in PR#1644. This issue is meant to document and track the necessary changes.  

- [ ] [Docs] Update UG, DG and codebase for direct references to GitHub to use more general language
- [x] [Docs & Backend] Change `author-config.json` column header `Author's GitHub ID` to `Author's Git Host ID`  while maintaining backward compatibility with older `author-config.json` files that still use `Author's GitHub ID` column header
    - **Assigned to @SkyBlaise99** (https://github.com/reposense/RepoSense/pull/1799) 
- [x] ~[Frontend] Add message or warning when unsupported links are clicked to notify user that the link is not supported~
- [x] [Frontend] Allow the repo icon to use respective github, gitlab, and bitbucket icons (fontawesome has all three) when we know the repo is one of them, but that can be a separate enhancement (taken from [this suggestion]( https://github.com/reposense/RepoSense/pull/1644#discussion_r835710517))
    - **Assigned to @SkyBlaise99** (https://github.com/reposense/RepoSense/pull/1800)
- [x] [Tests] Expand list of repositories used for PR dashboard previews to include other git hosting platforms
    - **Assigned to @SkyBlaise99** (https://github.com/reposense/RepoSense/pull/1805)

SkyBlaise99:
Hi, I wonder if this is still ongoing and if I can attempt it? I have an idea for the 2nd point regarding `author-config.json`.

SkyBlaise99:
Hi @chan-j-d, any update on this?

chan-j-d:
Hi @SkyBlaise99, apologies I missed your original message. You're welcome to attempt any of them!

SkyBlaise99:
@chan-j-d I have a question regarding my part about the backward compatibility.

Would you like the program to just be able to read from older `author-config.json`, or to also update the header from `Author's GitHub ID` to `Author's Git Host ID` also?

SkyBlaise99:
The part on the backward compatibility might involve changing quite a lot of the codes. Currently the parsers recognize the header using the `XXX_HEADER` strings that we defined.

What I have in mind is to change this into an array that stores both the new and old headers. So something like

``` java
private static final String[] GIT_ID_HEADERS = {"Author's Git Host ID", "Author's GitHub ID"};
```

[...]

PR #1799, opened 2022-06-27
[#1726] Update author config

SkyBlaise99:
Part of #1726 

## Proposed commit message

```
Previously, Reposense is only used with GitHub, hence 
"Author's GitHub ID" was used as the header in author-config.csv.

This header should be updated as we expand Reposense to be compatible 
with other platforms. Meanwhile, we would also like to still support 
the older version of author-config.csv that uses "Author's GitHub ID".

Let's update the header of author-config.csv to "Author's Git Host ID" 
while maintaining backward compatibility to the older version that 
still uses the "Author's GitHub ID" header.
```

## Other information

This resolves Part 2 of #1726

The 2 abstract methods, `mandatoryHeaders()` and `optionalHeaders()`, will now return `String[][]` objects. This is to allow multiple equivalent headers, as each equivalent header is of type `String[]`.

`Author's Git Host ID` and `Author's GitHub ID` will be recognized as the same header, hence the name `equivalent header`.

Example of `equivalent header`:

``` java
private static final String[] GIT_ID_HEADERS = {"Author's Git Host ID", "Author's GitHub ID"};
```

For headers that are not `equivalent header`, they are still converted into `String[]` type.

``` java
private static final String[] LOCATION_HEADER = {"Repository's Location"};
```


src/main/java/reposense/parser/RepoConfigCsvParser.java
```diff
@@ -48,23 +48,35 @@ public RepoConfigCsvParser(Path csvFilePath) throws FileNotFoundException {
...
+        };
+
+        String[] singleHeaders = new String[] {
                 BRANCH_HEADER, FILE_FORMATS_HEADER, IGNORE_GLOB_LIST_HEADER, IGNORE_STANDALONE_CONFIG_HEADER,
                 IGNORE_FILESIZE_LIMIT_HEADER, IGNORE_COMMIT_LIST_CONFIG_HEADER, IGNORE_AUTHOR_LIST_CONFIG_HEADER,
                 SHALLOW_CLONING_CONFIG_HEADER, FIND_PREVIOUS_AUTHORS_CONFIG_HEADER, FILESIZE_LIMIT_HEADER,
                 SKIP_IGNORED_FILE_ANALYSIS_HEADER
         };
+
+        return combineHeaders(equivalentHeaders, singleHeaders);
```
chan-j-d, line 79: Do you think it might be better to place all the single headers into a length 1 array to simplify the process? I.e. something like placing `BRANCH_HEADER` into `{BRANCH_HEADER}` to initialise the `String[][]` directly. 

It avoids the need to treat equivalent and single headers separate and coupling with the `combineHeaders` method. I am aware that that is what `combineHeaders` does but I don't think it should be a necessity to call it like this.

config/author-config-old.csv
```diff
@@ -0,0 +1,9 @@
+Repository's Location,Branch,Author's GitHub ID,Author's Emails,Author's Display Name,Author's Git Author Name,Ignore Glob List
```
chan-j-d, line 1: I think this file can be removed altogether

chan-j-d (changes requested):
Not quite a full review. Wanted to know your thoughts about my suggestion.

config/author-config-old.csv
SkyBlaise99: Ok, will update after I'm back from reservists.

src/main/java/reposense/parser/RepoConfigCsvParser.java
SkyBlaise99: I'm not too sure what's the style in reposense. For me, I prefer to keep to minimal changes. Single arrays seems a bit weird to me and that's why I did it this way. But if you think single arrays are ok then sure, I'll make the changes.

chan-j-d: I personally don't think that there's anything strange with single element arrays or even a need to carefully distinguish the `String[]` groups and the single `String` headers. I would just change the method name to something like `optionalHeaderGroups` and `mandatoryHeaderGroups` instead of `optionalHeaders` and `mandatoryHeaders`. 

I guess my main issue was more of the unused local variables lying around together with the need to always combine them with `combineHeaders`

SkyBlaise99: The unused local variables are just there as a template for future dev to know how to use. But sure, I'll just make everything into arrays then.

docs/ug/configFiles.md
```diff
@@ -70,13 +70,15 @@ Optionally, you can use an `author-config.csv` (which should be in the same dire
...
-| Author's GitHub ID {{ mandatory }} | Username of the target author's profile on GitHub, GitLab or Bitbucket, e.g.`JohnDoe`. |  
+| Author's Git Host ID<sup>**</sup> {{ mandatory }} | Username of the target author's profile on GitHub, GitLab or Bitbucket, e.g.`JohnDoe`. |  
 | Author's Emails<sup>*</sup> | Associated emails of the author. For GitHub users, this can be found in your [GitHub settings](https://github.com/settings/emails). |
 | Author's Display Name | The name to display for the author. Default: author's username. |
 | Author's Git Author Name<sup>*</sup> | The meaning of _Git Author Name_ is explained in [_A note about git author name_](#a-note-about-git-author-name). |
 | Ignore Glob List<sup>*</sup> | Files to ignore for this author, in addition to files ignored by the patterns specified in `repo-config.csv`. The path glob syntax is the same as that of Ignore Glob List in `repo-config.csv`. |
 
 <sup>* **Multi-value column**: multiple values can be entered in this column using a semicolon `;` as the separator.</sup>
+</br>
+<sup>** **Backward compatibility**: `Author's Git Host ID` can be replaced with `Author's GitHub ID`.</sup>
```
chan-j-d, line 81: I would write it as something like 
> For backward compatibility, `Author's GitHub ID` is also accepted as the header in place of `Author's Git Host ID`

and remove the bolded label

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -150,13 +150,23 @@ private boolean isLineMalformed(CSVRecord record) {
...
-            if (get(record, header).isEmpty()) {
+
+        for (String[] equivalentHeaders : mandatoryHeaders()) {
+            boolean isLineFormatMalformed = true;
+            for (String header : equivalentHeaders) {
+                if (!get(record, header).isEmpty()) {
+                    isLineFormatMalformed = false;
+                    break;
+                }
+            }
```
chan-j-d, line 161: I think streams might be appropriate here. If I'm understanding this correctly, the inner for loop is equivalent to
`boolean isLineFormatMalformed = Arrays.stream(equivalentHeaders).anyMatch(...);`

src/main/java/reposense/parser/CsvParser.java
```diff
@@ -257,31 +289,42 @@ private void validateHeader(String[] possibleHeader) throws InvalidCsvException,
...
-        for (String mandatory : mandatoryHeaders()) {
-            if (!headerMap.containsKey(mandatory)) {
+        for (String[] equivalentHeaders : mandatoryHeaders()) {
+            boolean isAnyEquivalentHeaderPresent = false;
+            for (String mandatory : equivalentHeaders) {
+                if (headerMap.containsKey(mandatory)) {
+                    isAnyEquivalentHeaderPresent = true;
+                    break;
+                }
+            }
```
chan-j-d, line 299: This can probably also be changed to use `Stream::anyMatch`

chan-j-d (changes requested):
PR looks good! Some minor points to change

docs/ug/configFiles.md
SkyBlaise99: I see, I had it in bold just to follow the previous line XD

SkyBlaise99: btw just to double-check, `**` is not used right? Or will some other symbols be preferred to make it less ambiguous with the single `*`?

src/main/java/reposense/parser/CsvParser.java
SkyBlaise99: Yup, definitely better and shorter. I will use `noneMatch()` instead to remove the `!` infront as well.

docs/ug/configFiles.md
SkyBlaise99: Maybe a `1` subscript? Or I can use `+` like the other table above?

chan-j-d: I think we should reserve `+` for a single meaning (overridable column in this case). Maybe something like `#` or `&`?

SkyBlaise99: Okay, I'll change it to `#` then.

dcshzj (approved):
LGTM!

PR #1800, opened 2022-06-30
[#1726] Use respective icons for known repository types

SkyBlaise99:
Part of #1726 

## Proposed commit message

```
Currently, 'database' is used as the generic icon for all repository 
types.

As we allow more repository types, it will be clearer to the user if 
the icon reflects the repository used. Currently we are supporting 
'GitHub', 'GitLab', and 'BitBucket'. 'database' icon is kept as the 
default icon for repositories that does not belong to any of the above 
mentioned.

Let's use the respective icons for known repository types.
```

## Other information

Part 4 of #1726 (and maybe part 5 as well)


config/repo-config.csv
```diff
@@ -8,3 +8,5 @@ https://github.com/reposense/testrepo-Delta.git,add-binary-file,,,,,,,
 https://github.com/reposense/RepoSense.git,master,,,,,,,
 https://github.com/reposense/testrepo-Empty.git,master,,,,,,,
 ftp://github.com/reposense/RepoSense.git,master,,,,,,,
+https://bitbucket.org/skyblaise/testrepo-bitbucket.git,master,,,,,,,
+https://gitlab.com/reposense/testrepo-gitlab.git,main,,,,,,,
```
chan-j-d, line 12: Does this repo exist?

chan-j-d (commented):
What's the reason behind including these repos inside this pull request?

config/repo-config.csv
SkyBlaise99: I created it and made it public. You cannot access or view it?

SkyBlaise99:
Included just to test out if the icons are working.

config/repo-config.csv
chan-j-d: The bitbucket one is accessible. The gitlab one is not accessible even while I'm logged in. I can't tell if the repo exists though your profile does not have any visible repos either

chan-j-d:
If it's just a demonstration, it doesn't have to be included. I'm not too comfortable to include it in the `repo-config.csv` itself as part of the PR especially since the repos are not in the control of the reposense org

frontend/src/components/v-summary-charts.vue
```diff
@@ -316,6 +330,24 @@ export default {
...
+        return 'gitlab';
+      }
+
+      if (repoLink.includes('bitbucket')) {
+        return 'bitbucket';
+      }
+
+      return 'default';
+    },
+
```
chan-j-d, line 350: I don't agree with this method of detection. You could have a repo hosted on gitlab with github in its name and this will identify it wrongly. 

The identification is already done by the backend code and can be found in `window.REPOS[repo].location.domainName` which gives the domain type. You can use that instead.

frontend/src/components/v-summary-charts.vue
```diff
@@ -316,6 +330,24 @@ export default {
...
+      if (repoLink.includes('gitlab')) {
+        return 'gitlab';
+      }
+
+      if (repoLink.includes('bitbucket')) {
+        return 'bitbucket';
+      }
+
+      return 'default';
```
chan-j-d, line 348: Is it possible to declare `'default'` as some kind of constant variable that can be referenced here and in lines 45 and 115.

chan-j-d (changes requested):
Thanks for doing it up. It seems to work well! Do take a look at some of the comments about the code.

config/repo-config.csv
SkyBlaise99: After some checking, the visibility is somehow set to `private` and I'm unable to change it to `public`, which is weird. Seems to be some problem from GitLab's side. Anyways, I will remove the 2 repos as you mentioned so it should be a problem.

frontend/src/components/v-summary-charts.vue
SkyBlaise99: Guess I missed that part, will patch it

frontend/src/components/v-summary-charts.vue
SkyBlaise99: Sure, let me google and update it

SkyBlaise99: Regarding this, I now have a better idea. Instead of returning the type, I can straight away return the icon. Trying to work towards this goal now.

SkyBlaise99: Is it fine to leave the 'github', 'gitlab', and 'bitbucket' as they are?

chan-j-d: yep, those are fine

chan-j-d (approved):
LGTM!

Zhou-Jiahao-1998 (approved):
LGTM! Thanks @chan-j-d for following the PR.

SkyBlaise99:
Hi @gerhean, can you help to review this PR? Thank you!

gerhean (approved):
LGTM

PR #1804, opened 2022-07-09
[#1498] Frontend: maintain file type label after filtering

### LINKED ISSUE #1498: Zoom inconsistent labels
image

image

When I untick `tests`, the commit containing both `tests` and `backend` remain, but the label for `tests` disappears.

Even if this is intended behavior, I find it rather unintuitive. What do you think?

https://dashboard-1462-pr-reposense-reposense.surge.sh/

jbkim1999:
Fixes #1498

## Proposed commit message

```
When a commit contains two or more file types and one of them is
filtered out, the commit could still be visible if it has other file
types that are still filtered in.

However, in such cases, the filtered-out file type label is no longer
visible for that commit. This could confuse the users by indicating that
the commit no longer consists of that filtered-out file type.

Let's retain all the file types of a commit by returning all the file
types if at least one of them is still filtered in and hence viewable to
the users.
```

### When All is checked (See Add edit action... commit)
image

### After html file is filtered out (previous behaviour)
image

Same commit is visible, but without the html label (which should still be existent)

### After html file is filtered out (fixed behaviour)
image

html label is now visible.


frontend/src/views/v-zoom.vue
```diff
@@ -101,8 +101,9 @@
       .hash
         span {{ slice.hash.substr(0, 7) }}
       span.fileTypeLabel(
+        v-if="containsAtLeastOneSelected(Object.keys(slice.fileTypesAndContributionMap))"
```
yhtMinceraft1010X, line 104: Try adding a comma here to address the failing CI.

jbkim1999: Thank you, addressed it for my next commit!

yhtMinceraft1010X (approved):
LGTM

frontend/src/views/v-zoom.vue
```diff
@@ -383,8 +384,14 @@ export default {
       window.encodeHash();
     },
 
-    filterSelectedFileTypes(fileTypes) {
-      return fileTypes.filter((fileType) => this.selectedFileTypes.includes(fileType));
+    containsAtLeastOneSelected(fileTypes) {
+      let includesSelectedFileTypes = false;
+      fileTypes.forEach((fileType) => {
+        if (this.selectedFileTypes.includes(fileType)) {
+          includesSelectedFileTypes = true;
```
Zhou-Jiahao-1998, line 391: This is just my personal preference, but I think it is more neat to simple return `true` here and return `false` on **line 394**. That way we can terminate the function earlier and declare one less variable. However, the difference is minimal, so it's really just my personal view. Let me know what you think :)

Zhou-Jiahao-1998 (changes requested):
@jbkim1999 Please refer to my comments. Thank you!

frontend/src/views/v-zoom.vue
jbkim1999: I think that's a very reasonable point and I'll make the fix soon. Thank you for the comment!

frontend/cypress/tests/zoomView/zoomView_selectFileTypes.js
```diff
@@ -44,10 +44,7 @@ describe('check file types ', () => {
         .uncheck()
         .should('not.be.checked');
 
-    cy.get('.zoom__day > .commit-message > .fileTypeLabel')
-        .should('not.contain.text', 'java');
-
-    cy.get('.zoom__day > .commit-message > .fileTypeLabel')
-        .should('contain.text', 'gradle');
+    cy.get('#tab-zoom .fileTypes input[value="all"]')
+        .should('not.be.checked');
```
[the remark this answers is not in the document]
jbkim1999: I see, I was a little afraid of hard coding but I do agree that some more test cases can be generated from this PR. Will make this change also. Thanks for the suggestion!

frontend/cypress/tests/zoomView/zoomView_selectFileTypes.js
```diff
@@ -44,10 +44,67 @@ describe('check file types ', () => {
         .uncheck()
         .should('not.be.checked');
 
-    cy.get('.zoom__day > .commit-message > .fileTypeLabel')
-        .should('not.contain.text', 'java');
+    cy.get('#tab-zoom .fileTypes input[value="all"]')
+        .should('not.be.checked');
+  });
+
+  it('a commit of a specific file type should not be seen when that type is unchecked', () => {
```
Zhou-Jiahao-1998, line 51: Maybe putting as `a commit of only one file type...` will be clearer for people who look at this for the first time?

Zhou-Jiahao-1998 (approved):
@jbkim1999 LGTM! Thanks for writing the cypress test cases. Let me know what you think about the mentioned point.

frontend/cypress/tests/zoomView/zoomView_selectFileTypes.js
jbkim1999: What do you think about `a commit should not be seen when all of its file types are unchecked`?

In this way the condition is not restricted to having only one file type.

dcshzj (approved):
LGTM!

Issue #1770, opened 2022-04-29
Parallelize backend tests


yhtMinceraft1010X:
**What feature(s) would you like to see in RepoSense?**

The `Test` class in Gradle has a [`maxParallelForks` property](https://docs.gradle.org/6.9.2/dsl/org.gradle.api.tasks.testing.Test.html#org.gradle.api.tasks.testing.Test:maxParallelForks) that would allow multiple JUnit test classes to run at the same time. Running test classes in parallel can reduce the build time.

**Potential Problem**

A potential problem lies in the `GitTestTemplate` class and test classes that extend it:
https://github.com/reposense/RepoSense/blob/b12d3485d4b7a3aca5f2545404b5545a5277ff5e/src/test/java/reposense/template/GitTestTemplate.java#L107-L111

The `@BeforeAll` annotation indicates that the method is called once when a test class such as `AnnotatorAnalyzerTest` is started. Within this method, the repo is cloned (in short, the `testrepo-Alpha.git` remote is cloned to the same output folder every time a test class that extends `GitTestTemplate` starts):
https://github.com/reposense/RepoSense/blob/b12d3485d4b7a3aca5f2545404b5545a5277ff5e/src/test/java/reposense/util/TestRepoCloner.java#L28-L33

If 2 test classes that extend `GitTestTemplate` run in parallel, it is possible that while 1 test class is still running through some test methods, the other class is re-cloning the repo, which could interfere with the first test class.

**Limitation**

According to [this website](https://kousenit.org/2020/11/22/simple-demo-of-gradle-parallel-tests/), the methods within a single test class do not run in parallel. This means that there won't be time savings for `ConfigSystemTest`.

SkyBlaise99:
For this, I think maybe we can try to change `outputFolderName`. So even if they are the same repo, the `outputFolderName` will be different and hence the 2 tests that relies on the same repo can be executed concurrently.

SkyBlaise99:
According to this [website](https://www.baeldung.com/junit-5-parallel-tests), the following can be added to `junit-platform.properties` file in `src/test/resources folder`.

Test Parallelization Within a Class: `junit.jupiter.execution.parallel.mode.default=concurrent`
Test Parallelization Within a Module: `junit.jupiter.execution.parallel.mode.classes.default=concurrent`

SkyBlaise99:
[...]

PR #1806, opened 2022-07-14
[#1770] Parallelize backend tests

SkyBlaise99:
Part of #1770 (only `test`, no `systemtest`)

## Proposed commit message

```
Currently, JUnit tests in the test folder are running sequentially, 
which is slow when we have more than 300 test cases now.

If we allow parallel execution, multiple tests can be executed 
simultaneously, speeding up the build time.

Let's run JUnit tests in parallel.
```

## Other information

| Class | Method | Time Taken for `test` |
| --- | --- | --- |
| Sequential | Sequential | ~40s |
| Parallel | Sequential | ~17s |

There are only 2 test classes (`RepoConfigurationTest` which takes ~15s and `FileAnalyzerTest` which takes ~10s) that take relatively longer than the rest (<5s), so I did not see the need to implement parallel method execution.

Now they run together 😄 

image

`systemtest` is a bit harder to switch to parallel execution due to the way how Reposense is made. There are too many `static` variables and methods within. Pretty much the whole `cloneAndAnalyzeRepos()` method needs to run `synchronized`, and this is also the method that contributes largely to the execution time.

The best achieved is ~5s improvement, with consistently ~2s improvement. Hence I did not include the part on `systemtest` as I feel my solution is not good enough. If anyone has some new ideas on how to make `cloneAndAnalyzeRepos()` run in parallel, feel free to try it out and see if it is possible to improve `systemtest`.


yhtMinceraft1010X (changes requested):
Remember to merge the changes in the `master` branch into this branch and resolve the merge conflicts.

yhtMinceraft1010X (approved):
LGTM

dcshzj:
Hi @SkyBlaise99, thanks for your contribution! I noticed that you added `EXTRA_OUTPUT_FOLDER_NAME_SUPPLIER`, which allows different threads to clone to different folders so that they don't conflict with one another. However, I got two questions about this implementation:

1. The `extraOutputFolderName` in `RepoConfiguration.java` seems to append the thread ID to the path, so it is cloned into a subdirectory. Can I check if this would cause issues for tests which may be using the parent repo directory?
    For example, if we are using `testrepo-Alpha`, the repo was originally cloned into `path/to/repos/testrepo-Alpha` but with your implementation, it now becomes `path/to/repos/testrepo-Alpha/12345`. Is it possible that another test case assumes the repo is cloned at `path/to/repos/testrepo-Alpha` and treats the `12345` folder as part of the original repo?
2. Now that the repos are cloned to a thread-specific path, does this still retain the original functionality of cloning only when needed (was introduced in #1478 and #1570)? My concern would be whether this implementation would cause the same repo to be cloned twice by different threads, which does extra work and may cause GitHub to rate limit.

SkyBlaise99:
For 1, I believe that all the tests uses subdirectory. None of them should be using the parent directory, and the alpha repo is not cloned to the parent directory either. So this shouldn't be an issue for now, but I do see that this may cause an issue one day.

I can try to update it such that the thread id comes before the folder name. E.g. `12335/test repo...`. 

For 2, I think this PR will switch back to cloning once for each test, such that they run in parallel. Let me try to make it back to cloning once, and then copy the content over with a util function etc.

dcshzj:
> For 1, I believe that all the tests uses subdirectory. None of them should be using the parent directory, and the alpha repo is not cloned to the parent directory either. So this shouldn't be an issue for now, but I do see that this may cause an issue one day.
> 
> I can try to update it such that the thread id comes before the folder name. E.g. `12335/test repo...`.

I think this is possible, but question 2 would be a bigger issue.

> For 2, I think this PR will switch back to cloning once for each test, such that they run in parallel. Let me try to make it back to cloning once, and then copy the content over with a util function etc.

I think another concern would be that the threads would have to be blocked if one of the threads is still cloning the repo, so possibly some mutex implementation may be needed.

SkyBlaise99:
> I think another concern would be that the threads would have to be blocked if one of the threads is still cloning the repo, so possibly some mutex implementation may be needed.

I need to check my code again for this. Iirc currently the all clone together and there seems to be no issue. Originally they cannot clone together because the dest folder is used, hence I introduced the `extraOutputFolderName` to solve this issue. Technically all the test cases can clone in parallel.

SkyBlaise99:
@dcshzj just did a count for number of times `cloneAndBranch()` is called using `AtomicInteger`. Both my PR and master is at 50.

I realised that all the repos are cloned once as specified in `@BeforeAll` of `GitTestTemplate`. What I'm doing is just making this done in parallel. So part 2 is not a concern i think. But I will still fix part 1.

Now there's a bigger issue with the `systemtest` 😢 . Will fix it soon.

SkyBlaise99:
A drawback to this is that each new class in `systemtest` will need to add this annotation `@Execution(ExecutionMode.SAME_THREAD)` infront of `public class XXXTest`. Lucky we only have 2 for now but this might be an issue in the future.

SkyBlaise99:
@dcshzj perhaps you can review this again?

SkyBlaise99:
Hi @dcshzj any updates?

dcshzj (approved):
LGTM!

PR #1807, opened 2022-07-19
[#1753] Backend: refactor processLine()

### LINKED ISSUE #1753: Refactoring `processLine` in `RepoConfigCsvParser` class
This is a follow-up issue from discussions in https://github.com/reposense/RepoSense/pull/1647#discussion_r841910967

**What ideas would you like to see in RepoSense?**

The `processLine` method is now at more than a 100 lines, and getting hard to maintain. It might be a good idea to break them apart. 

Perhaps the config parser could be broken down into a collection of parser subclasses that are logically related to each other (e.g for parsing columns related to size limit), so that parsing is easier to maintain while avoiding unnecessary proliferation of parsing classes.

SkyBlaise99:
Fixes #1753 

## Proposed commit message

```
The processLine() method is now at more than 100 lines, which is hard to
maintain.

There are quite a few lines in the method that are very similar to one
another, which can be extracted into a new generic helper method to be
reused, hence shortening the code.

Let's extract the common method and reuse them to shorten the method.
```


SkyBlaise99:
Hi @dcshzj can you review this?

dcshzj (approved):
LGTM!

dcshzj:
@SkyBlaise99 are you able to resolve the conflicts here?

SkyBlaise99:
@dcshzj done, local test all good, waiting for CI to pass.

PR #1811, opened 2022-07-23
[#1817] Actions: update list of job runners environments for CI

### LINKED ISSUE #1817: Continuous Integration for macos-10.15 is always cancelled

* **RepoSense Version:** linked the PRs

1. push my branch to trigger github checks

The checks run and either pass or fail.

`Continuous Integration / macos-10.15 JDK 8 (pull_request)` is always cancelled for all 3 PRs.

#1806 
image

#1807 
image

#1816 
image

A bit too much of a coincidence when all 3 PR's mac is cancelled. I tried to push new commits but same keep happening.

dcshzj:
Fixes #1817.

## Proposed commit message

```
The virtual environments for the GitHub Actions runners have been
updated, with the introduction of macos-12 and ubuntu-22.04, and the
deprecation of macos-10.15.

Let's update the list of environments used for running the CI jobs.
```

## Other information

List of available environments are given here: https://github.com/actions/virtual-environments#available-environments


.github/workflows/integration.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     name: ${{ matrix.os }} JDK 8
     strategy:
       matrix:
-        os: [ubuntu-18.04, ubuntu-20.04]
+        os: [ubuntu-18.04, ubuntu-20.04, ubuntu-22.04]
```
yhtMinceraft1010X, line 18: I'm not sure if we should adopt `ubuntu-22.04` since it's still in beta according to the [actions/virtual-environments page](https://github.com/actions/virtual-environments#available-environments).

.github/workflows/integration.yml
```diff
@@ -173,7 +173,7 @@ jobs:
     name: ${{ matrix.os }} JDK 8
     strategy:
       matrix:
-        os: [windows-2022, windows-2019]
+        os: [windows-2016, windows-2019, windows-2022]
```
yhtMinceraft1010X, line 176: I'm not sure if we should add another Windows os for CI since this may impose additional load. Particularly because Windows CI is slower than that of Ubuntu.

.github/workflows/integration.yml
dcshzj: Yup, it is in beta currently. I added this such that we can run on this version, but I will not be marking the run as a required check (i.e. it is allowed to fail but will not block the PR from being merged).

.github/workflows/integration.yml
dcshzj: The jobs will run in parallel, so I think it does not cause the CI builds to take longer. Nonetheless, I removed it because I realized it was [supposed to be deprecated by now](https://github.com/actions/virtual-environments/issues/4312).

yhtMinceraft1010X (approved):
LGTM

PR #1812, opened 2022-07-24
[#946] Add visual indication for sorting under authorship panel

### LINKED ISSUE #946: CodeView: improve visualization on different sort parameters

jbkim1999:
Fixes #946 

## Proposed commit message

```
In the authorship panel, when the files are sorted by different
parameters, there is no visual indication that files are sorted using
that parameter.

Let's improve the indication by placing the sort parameter in front of
the title so that the users know the sorting parameter they are using to
view the files.
```

### Sorted by `lineOfCode`
image

### Sorted by `fileType`
image

### Sorted by `fileName`
image

### Sorted by `path`
image


jbkim1999:
Hi @Zhou-Jiahao-1998 

I will try the sorting by filename like you've suggested, and I do agree for file path, it is a bit redundant to wrap it around with square bracket.

However, I did try bolding respective parameters, as can be seen from my latest comment [this issue](https://github.com/reposense/RepoSense/issues/946), but it seems (to me, at least) that the effect of bolding is not that prominent in terms of the visuals, so I am a little skeptical about bolding them (and it is even more so inside the file type label like you've mentioned)

jbkim1999:
Hi @damithc @Zhou-Jiahao-1998 

I have accommodated some of your suggestions and made an improvement, please comment how it looks like!

Overall change: the tile is **not bold** by default. Only parameters used for sorting will be bolded.

### `lineOfCode`
image

### `fileType`
image

### `fileName` (only `fileName` is bolded)
image

### `path` (the whole file path is bolded)
image

Zhou-Jiahao-1998 (approved):
LGTM! Thanks for the changes

dcshzj:
I think the code is okay, @damithc any comments on the visual changes? The above comment has a summary of how the different sorting options will change the look of the frontend and previously, the titles are similar to the "`path` (the whole file path is bolded)" sorting, without the bolding, similar to this:

Screenshot 2022-08-07 at 11 45 19

damithc:
@dcshzj @jbkim1999 I'm OK with the change but for the below, I think the word `from` makes things harder for the viewer. We can either use a symbol e.g., `@` or use a dim style on that word (i.e., grey color) so that the file name and path are visually separated better. What you do you guys think? Perhaps try it out and post screenshots?

> ### `fileName` (only `fileName` is bolded)
> image

jbkim1999:
Hi prof @damithc,

Will try that and update on it!

jbkim1999:
Hi @dcshzj @damithc @Zhou-Jiahao-1998 

I have tried these two approaches, sharing the screenshots below:

### Use `@` instead of `from`
Screenshot 2022-08-08 at 5 55 08 PM
I think the @ symbol is a bit intruding

### Use dim style on the word `from`
Screenshot 2022-08-08 at 6 40 58 PM

Personally I feel the dimming is better, but what do you think?

damithc:
> Personally I feel the dimming is better, but what do you think?

Sure. You can also use `in` instead of `from` to save space.

jbkim1999:
Yep, I think using `in` helps with shortening the text too!

image

dcshzj (approved):
LGTM!

PR #1816, opened 2022-07-27
[#1645] Use RepoSense::main in ConfigSystemTest

### LINKED ISSUE #1645: Change how `ConfigSystemTest` is done by using `RepoSense::main`
**What change(s) would you like to see in RepoSense?**

Change how system tests are run by using the `RepoSense::main` method to provide full end-to-end validation through supplying CLI arguments and comparing the output JSON files. 

Adding new system tests can be difficult as the existing methods use the config files in the `resources` folder by default. This makes it difficult to perform tests on more specific scenarios such as using a new config file to test against a specific branch of a specific repo only. 

The current setup with the [`ConfigSystemTest::generateReport` method](https://github.com/reposense/RepoSense/blob/731a268e7eeee66b8894fecccea4fc0c27b50465/src/systemtest/java/reposense/ConfigSystemTest.java#L152) also bypasses the parsing of the original CLI arguments and setting up of `RepoConfiguration`s. If the `main` method is changed, then this method would also have to be changed with it. I believe it would be beneficial to use `RepoSense::main` instead for the full end-to-end validation. 

Setting up test cases would then be similar to specifying the exact CLI arguments that we would normally run against the RepoSense.jar file. 

Rewrite the `runTest` method similar to something in this experimental PR [here](https://github.com/chan-j-d/RepoSense/blob/testing-systemtest-refactoring/src/systemtest/java/reposense/ConfigSystemTest.java#L90). Only one test case has been rewritten here to be as similar to the original as possible. The output for the authorship files are not exactly the same as the one used currently and I have not been able to replicate it properly using only the CLI arguments. 

**Additional context**

_Immediate drawbacks_
- The current implementation of RepoSense with this change will be unable to avoid multiple repeated cloning of the same few repositories for system test. One possible way to overcome this is to leave one test that performs clone and delete. For other tests, a copy of the remote repo is cloned locally and the tests are performed against the local copies instead.
- There are some parameters such as `shallowCloning` and `findPreviousAuthors` that still have to be passed to the `runTest` method as they directly affect the behaviour of the test case. 
- The time generated for the report summary would be different from the expected JSON files. Some sort of bypass is needed to compare the summary JSON files while ignoring the time the report is generated.

SkyBlaise99:
Fixes #1645

## Proposed commit message

```
The current setup with the ConfigSystemTest::generateReport method 
bypasses the parsing of the original CLI arguments and setting up of 
RepoConfigurations.

If the main method is changed, then this method would also have to be 
updated, which can be troublesome. It is more beneficial to use 
RepoSense::main instead for the full end-to-end validation.

Let's rewrite the runTest method to use RepoSense::main method.
```


SkyBlaise99:
Hi @chan-j-d can you review this?

src/systemtest/java/reposense/ConfigSystemTest.java
```diff
@@ -106,111 +107,54 @@ public void test30DaysFromUntilDateWithShallowCloning() throws Exception {
...
-        String formats = String.join(" ", TESTING_FILE_FORMATS);
-
-        InputBuilder inputBuilder = new InputBuilder().addConfig(configFolder)
-                .addFormats(formats)
-                .addTimezone(TEST_TIME_ZONE)
-                .add(inputDates);
-
-        if (shallowCloning) {
-            inputBuilder = inputBuilder.addShallowCloning();
+    private void runTest(InputBuilder inputBuilder, boolean shouldFreshClone, String pathToResource) {
```
chan-j-d, line 148: This method does not include javadocs for the new arguments

chan-j-d (changes requested):
The work looks really good. I was initially hesitant with the amount of code that might need to be added to production code but this seems acceptable. 

Do update the Javadocs as besides the method that I highlighted, I think there are number of other methods that also need to update their Javadocs, especially the method arguments.

SkyBlaise99:
@chan-j-d thx for the feedback. 

> Do update the Javadocs... 

Done, pls review it again and update me if I still missed out any of them.

SkyBlaise99:
On the other hand, this PR introduces a new flag for `fresh clone`. From what I understood, this is because some of the test cases use shallow cloning but some do not. While all of them clone to the same directory, this results in the repo being cloned over and over again. 

Perhaps we can split the test cases into 2, 1 with shadow cloning and 1 without. Then following similar structure like `LocalRepoTest`, maybe we can achieve cloning once and use it for all test cases as well?

I can make a new followup issue for this if you agree.

src/main/java/reposense/RepoSense.java
```diff
@@ -85,26 +85,37 @@ public static void main(String[] args) {
...
+
+            if (!isTestMode) {
+                FileUtil.zipFoldersAndFiles(reportFoldersAndFiles, cliArguments.getOutputFilePath().toAbsolutePath(),
+                        ".json");
+            }
 
             logger.info(TimeUtil.getElapsedTimeMessage());
 
-            if (cliArguments.isAutomaticallyLaunching()) {
+            if (!isTestMode && cliArguments.isAutomaticallyLaunching()) {
```
chan-j-d, line 118: Can I know what's the reason for adding this? I don't think it's necessary to enforce this test code-related behavior in production code

chan-j-d (changes requested):
Everything else looks good. There's only one comment which I'd like to clarify

chan-j-d:
> On the other hand, this PR introduces a new flag for `fresh clone`. From what I understood, this is because some of the test cases use shallow cloning but some do not. While all of them clone to the same directory, this results in the need to clone the repo over and over again.
> 
> Perhaps we can split the test cases into 2 files, 1 with shadow cloning and 1 without. Then following similar structure like `LocalRepoSystemTest`, use 2 different foldera for each.
> 
> There are 3 test cases that does not use `shadow clone` nor `fresh clone`. They should be able to share the same cloned files for running the test cases?
> 
> I can make a new followup issue for this if you agree.

You can just create the issue. Personally, I don't think the improved speed in running the test cases is worth the effort of maintaining the somewhat odd behavior that the test cases would be actually performing

src/main/java/reposense/RepoSense.java
SkyBlaise99: I think I added it to all the codes that previously are not in the old `runTest()`. This also applies to [`FileUtil.zipFoldersAndFiles(...)`](https://github.com/reposense/RepoSense/pull/1816/files/f58870a6305b843c8f896af16b142ff450bc25a4#diff-c782c38fa12442cbc9d82ce735beb26380383527acd4255637c2905ff012ea7bR111).

chan-j-d: does it still work if those two `isTestCode` segments are removed? This and the zipFoldersAndFiles. My original intention was to actually get rid of `generateReposReport`'s own way of running the report generation so I don't think it's important to retain this behavior unless it does not work

SkyBlaise99: Just tested locally, can still pass all test cases when they are removed. Negligible difference in run time also. I shall remove them then.

SkyBlaise99:
> You can just create the issue. Personally, I don't think the improved speed in running the test cases is worth the effort of maintaining the somewhat odd behavior that the test cases would be actually performing

What I propose it to separate the test cases so that there will not be the need for this part.
``` java
didNotNormallyClonedRepo = inputBuilder.isShallowCloning();
```

So in 1 file, all the test cases will not fresh clone while in the other, all will fresh clone. Then the `private static boolean didNotNormallyClonedRepo` variable can be removed?

chan-j-d:
> So in 1 file, all the test cases will not fresh clone while in the other, all will fresh clone. Then the private static boolean didNotNormallyClonedRepo variable can be removed? Which I believe is the odd behavior you refering to? Speeding up the test cases will be another issue that can be worked on.

Oh it seems I misunderstood what you were intending. Regardless, I think you should write this up in a separate issue from this as it does not directly pertain to this PR and the idea sounds promising

src/main/java/reposense/RepoSense.java
```diff
@@ -85,20 +85,29 @@ public static void main(String[] args) {
...
                 logger.warning(FINDING_PREVIOUS_AUTHORS_INVALID_VERSION_WARNING_MESSAGE);
                 RepoConfiguration.setToFalseIsFindingPreviousAuthorsPerformedToRepoConfigs(configs);
             }
 
+            boolean isTestMode = cliArguments.isTestMode();
+
+            if (isTestMode) {
+                assert !isGitVersionInsufficient;
+                AuthorConfiguration.setHasAuthorConfigFile(false);
+            }
```
chan-j-d, line 100: Sorry I think I missed this the last time round. For insufficient git version even on test mode, please log instead. 

I'm not too familiar with exactly how the original system test cases were run. Could you explain the need for `AuthorConfiguration.setHasAuthorConfigFile(false);` and why it needs the git version check

SkyBlaise99: The assert is because the original behavior is to exit the test straight away when git is insufficient. I used assert as a temporary measure. Do you think I should make it throw an exception or just remove it?

As for the latter line, it is the only difference between the original version and the one in main. Removing this will cause some test cases to fail. As I'm not familiar with how the old test cases works, I have kept this line for now.

chan-j-d: Noted on the second point. Would it be possible to set it directly in the `runTest`method in ConfigSystemTest?

Also i misremembered the original behaviour but to clarify, it aborts the test instead of outright failing which affects whether someone who has a lower git version is able to properly build. 

I think it might be possible to do this in runTest by checking if the inputbuilder has the finding previous author flag and then doing `Assumptions.assumeFalse` similar to the original

SkyBlaise99: For the first, I doubt it can be achieved. It involves setting some parameters that is within main, which is against what this PR is achieving.

Another way I can think of is probably to add a static boolean variable in main and let test cases manipulate that. Then use the variable to set this line. Some what similar to what was currently done but without the need to use a flag.

SkyBlaise99:
@chan-j-d I have made quite a bit of changes. `main` is now almost the same as before. Most of the logic parts are shifted to `ConfigSystemTest`. Please review it again when you are free.

chan-j-d:
> @chan-j-d I have made quite a bit of changes. `main` is now almost the same as before. Most of the logic parts are shifted to `ConfigSystemTest`. Please review it again when you are free.

I'm so sorry about this as I was busy with work stuff. I actually preferred the test mode flag to this global variable in reposense. Some other potential benefits might be to execute code just for tests later on. 

I actually only just wanted the git version insufficient thing changed.

On top of that, after reverting to checking the test mode and setting author config flag, do add a comment explaining why its there. Something simple like the configsystemtests require it to pass

SkyBlaise99:
> I'm so sorry about this as I was busy with work stuff. I actually preferred the test mode flag to this global variable in reposense. Some other potential benefits might be to execute code just for tests later on.

It's ok, reverted back the changes.

chan-j-d (approved):
LGTM! Good work!

SkyBlaise99:
@dcshzj can you review this when you are free? Last PR for my IWM. Thank you.

dcshzj:
@SkyBlaise99 apologies for the late review, I think so far the code is great! One small nit, would it be possible to add the extra flags into the docs as well? It can be in the CLI syntax reference page but in a separate section from the other flags.

SkyBlaise99:
> One small nit, would it be possible to add the extra flags into the docs as well? It can be in the CLI syntax reference page but in a separate section from the other flags.

I have updated `cli.md`. Not sure what you mean by a separate section so I just put them at the very bottom. Do you want a separate header for it like this?

``` markdown
# Flags for system test

### `--test-mode`
...
```

dcshzj:
Yup I was referring to a separate header, wanted to make it clear to users that the flag is only for testing and not meant to be used. I'm not sure if it would also be good to place it inside a collapsed section, but I leave that up to you.

SkyBlaise99:
@dcshzj added a header for it with a line of brief intro. Do you think this is enough?

dcshzj (approved):
LGTM!

SkyBlaise99:
@dcshzj can you merge this as soon as possible so I can wrap up my IWM? Thank you.

dcshzj:
@SkyBlaise99 Usually we wait for at least 2 days after the ToMerge label is applied for any last-minute comments from anyone before we merge, but I will merge this now nonetheless.

SkyBlaise99:
@dcshzj thank you so much 😄

damithc:
@SkyBlaise99 should this behavior be in the UG? If the behavior is only meant for developers, it should not be mentioned in the UG.

dcshzj:
I think @damithc made a good point, @SkyBlaise99 would it be possible to create a new page similar to the CLI syntax reference but in the developer guide? We can also put it in the appendix.

SkyBlaise99:
My bad, I forgot about UG and DG diff. I will do the changes quickly.

The new `cli.md` in DG, do you want it to contain the original flags as well? Or just the flags for testing.

dcshzj:
> The new `cli.md` in DG, do you want it to contain the original flags as well? Or just the flags for testing.

I think just do the flags related to testing would be sufficient. It would also be good to elaborate more on the purpose of the flags, as if you are explaining to another developer.

SkyBlaise99:
Made another PR #1822 as this is closed already.

PR #1819, opened 2022-08-07
Hotfix: revert highlight.js from version 11 to 10

yhtMinceraft1010X:
## Proposed commit message

```
After upgrading highlight.js to version 11, the formatting of code in
the code panel breaks.

Let's revert highlight.js back to version 10.
```

### Code panel with highlight.js version 10:

Screenshot (5)

### Code panel with highlight.js version 11:

Screenshot (6)

### Description

It appears that the following lines in `v-segment.vue` do not work with version 11:
https://github.com/reposense/RepoSense/blob/1caadd3cacb3aa293f563053aa18b05512062565/frontend/src/components/v-segment.vue#L15-L18

After reading through the [changelog for `highlight.js` version 11](https://highlightjs.readthedocs.io/en/latest/upgrade-11.html#changes-to-result-data), I noticed that the overall class for each segment will have `language-[language name]` appended to it, but I'm not sure how this affects the way the code is parsed.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -20,7 +20,7 @@
     "@fortawesome/free-solid-svg-icons": "^6.0.0",
     "@fortawesome/vue-fontawesome": "^3.0.0-5",
     "core-js": "^3.6.5",
-    "highlight.js": "^11.5.0",
+    "highlight.js": "^10.5.0",
     "jszip": "^3.5.0",
     "minimatch": "^5.0.1",
     "muicss": "^0.10.3",


dcshzj:
@yhtMinceraft1010X Thanks for the quick patch, however I'm not able to load the dashboard. Does it load on your end? The console gives me a `Uncaught SyntaxError: Invalid or unexpected token` error message.

yhtMinceraft1010X:
> @yhtMinceraft1010X Thanks for the quick patch, however I'm not able to load the dashboard. Does it load on your end? The console gives me a `Uncaught SyntaxError: Invalid or unexpected token` error message.

It works on my end. I'm using Chrome version `104.0.5112.81`.

dcshzj (approved):
LGTM! I think it was a transient error and it works now, thanks!

PR #1823, opened 2022-08-09
Actions: remove Ubuntu 18.04 runner

dcshzj:
## Proposed commit message

```
The Ubuntu 18.04 runner is deprecated and will be fully removed on
2022-12-01, after which the jobs will fail.

Let's remove the Ubuntu 18.04 runner.

Reference: https://github.com/actions/runner-images/issues/6002
```

## Other information

- The deprecation was [announced on 2 August 2022](https://github.com/actions/runner-images/issues/6002).
- The Ubuntu 22.04 image is now GA, so it has now been made a required check.
- The Ubuntu 18.04 image is now made an optional check.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -15,7 +15,7 @@ jobs:
     name: ${{ matrix.os }} JDK 8
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


yhtMinceraft1010X (approved):
LGTM

PR #1826, opened 2022-08-17
[#1813] Fix git clone exception when cloning lfs files

### LINKED ISSUE #1813: Bug: RepoSense cannot clone a repository

* **RepoSense Version: 07d6ce9e9c**
* **OS and Version: Ubuntu 20.04.4 LTS**
* **git version: 2.25.1**

1. Run `java -jar RepoSense.jar --repos https://github.com/fyusuf-a/ft_transcendence.git --view`

Cloning of the repository should work. A `git clone https://github.com/fyusuf-a/ft_transcendence.git` runs without a problem.

The following message is displayed:
```[ERROR] Exception met while cloning or checking out fyusuf-a/ft_transcendence[master].Analysis terminated.```

### PR DESCRIPTION
**Pull request by lawwm**
Fixes #1813 

## Proposed commit message

```
When cloning repositories that utilize git-lfs, an exception will be
raised when the user's Git client is not properly configured.

Let's fix the exception that occurs while generating reports for
repositories that are using git-lfs. This is done by setting the Git
configuration to skip retrieving lfs files, and reset to user's original
configuration after the clone is complete.
```

## Other information

Explanation of why exception occurs is in #1813.

Full `git clone` will now always be configured to skip retrieving lfs files.
However, since that will overwrite the user's original git-lfs configuration, the user's original configuration is saved 
before configuration and reapplied after repos have been analyzed.

Since git-lfs files are not retrieved, only pointer files left behind.
image

### CODE REVIEW

**Comment by lawwm** [comment]
CI tests failed because `fatal: unable to read config file '/home/runner/.gitconfig': No such file or directory`. Will be trying to solve it.

**Comment by lawwm** [comment]
I've fixed the failing CI tests.

File: src/main/java/reposense/RepoSense.java [hunk 1]
```diff
@@ -110,6 +113,8 @@ public static void main(String[] args) {
             FileUtil.zipFoldersAndFiles(reportFoldersAndFiles, cliArguments.getOutputFilePath().toAbsolutePath(),
                     ".json");
 
+            GitConfig.setGlobalGitLfsConfig(globalGitConfig);
```
yhtMinceraft1010X, line 116: Not sure why this is here since the git cloning would have been done before this line.

yhtMinceraft1010X (changes requested):
Nice work on identifying the root cause of the lfs issue. That said, this new functionality needs some test cases, which can be set up on a new branch of [testrepo-Alpha](https://github.com/reposense/testrepo-Alpha/tree/master)

src/main/java/reposense/RepoSense.java
yhtMinceraft1010X: I take this back.

yhtMinceraft1010X:
To add on, the PR can be renamed as `[#1813] Fix git clone exception when cloning lfs files`

yhtMinceraft1010X (approved):
LGTM

src/main/java/reposense/git/GitConfig.java
```diff
@@ -0,0 +1,84 @@
...
+        CommandRunner.runCommand(Paths.get("."), command);
+    }
+
+    private static String setGitLfsConfigCommand(List<String[]> lfsConfigs) {
+        String command = "";
+        for (String[] config : lfsConfigs) {
+            command += "git config --global " + config[0] + " " + "\"" + config[1] + "\"" + " && ";
+        }
+        return command.substring(0, Math.max(0, command.length() - 4));
+    }
```
dcshzj, line 79: I'm not really a fan of how this function is written. A couple of points:
1. The string concatenation should be replaced with `String.format()` so that it is easier to read. Here's an example of how it's done elsewhere in the code:
https://github.com/reposense/RepoSense/blob/37425888df381680c7118a095ad3e5c5b0793cff/src/main/java/reposense/git/GitDiff.java#L32-L33
2. The command chaining is done in a rather hacky way. Can I suggest instead to append each command obtained from each iteration of `lfsConfigs` into an ArrayList, then use `String.join()` to chain the various commands together. We can then omit the use of `Math.max()` and `command.substring()` here.
3. Are we able to guarantee that the size of `lfsConfigs` would always be 2? What happens if the user has a malformed Git config where only the key exists? For example, you are expecting something like `filter.lfs.smudge=blah`, but what happens if the file only has `filter.lfs.smudge` without any `=`? I would suggest having an additional check here to ensure that we truly have a size of 2, otherwise we would get an IndexOutOfBoundsException.

For the last point, an example file could be something like this (open using `git config --global -e`):
```
[filter "lfs"]
        smudge
```
The output would just be `filter.lfs.smudge` on its own line.

src/main/java/reposense/git/GitConfig.java
```diff
@@ -0,0 +1,84 @@
...
+            CommandRunner.runCommand(Paths.get("."), command);
+        }
+    }
+
+    /**
+     * Delete the global git lfs configuration values.
+     */
+    public static void deleteGlobalGitLfsConfig() {
+        String command = "git config --global --unset " + FILTER_LFS_SMUDGE_KEY
+                + " && git config --global --unset " + FILTER_LFS_PROCESS_KEY;
```
dcshzj, line 69: This should be replaced with a `String.format()` (elaborated in the next comment).

src/main/java/reposense/RepoSense.java
```diff
@@ -110,6 +115,8 @@ public static void main(String[] args) {
...
```
dcshzj, line 118: Let's add a comment before this line to explain that we are setting the git config back to the original value, so that it is clearer for future developers.

dcshzj (changes requested):

dcshzj (approved):
LGTM!

PR #1830, opened 2022-08-21
[#1689] Disable links when they are broken

### LINKED ISSUE #1689: Broken links to remote repositories
This issue consolidates issues with links in the report to remote repositories. These links should be fixed for when (a) no remote repositories exist or (b) remote repository is not yet supported by RepoSense (is not Github, GitLab or Bitbucket). The following report links redirect to broken links if either of the above conditions are satisfied.

1. Commit ramp link (also mentioned in #231):
image
1. Link to author profile
image
1. Link to group repo tree
image
1. Link to blame view
image
1. Link to history view
image

**To discuss:**

What should be the expected behaviour in the above scenarios if either (a) or (b) is satisfied? Some options:
- Hide the links entirely
- Popover that explains that link is not available

LuoZhijie-tom:
Fixes #1689.

# Proposed commit message
```
There are some links in the report to the report repos such as view
group repo. However, when the link is missing or not supported by
RepoSense (i.e. is not Github, GitLab or Bitbucket), it points to
a non-existent link.

Let's disable the link when it is broken and change the message when the
users hover over the link to "This remote link is unsupported", so that
the users will know that the link is broken.
```

# Other information
This feature is added to commit ramp link, link to author profile, link to group repo tree, link to blame view and link to history view according to [#1689]

This is a sample:
sample


frontend/src/views/v-authorship.vue
```diff
@@ -579,6 +595,11 @@ export default {
       return window.getBlameLink(this.info.repo, repo.branch, file.path);
     },
 
+    isBrokenLink(link) {
```
Zhou-Jiahao-1998, line 598: @LuoZhijie-tom This function appeared 3 times across 3 different files. I think it will be better to extract them out into a util file and import it. This way it is much easier to maintain and trace in the future if we want to change this function. I will suggest adding the string `This remote link is unsupported` to this new file as well.

You can probably add  a new file under `src/utils`.

Zhou-Jiahao-1998 (changes requested):

frontend/src/views/v-authorship.vue
LuoZhijie-tom: Hi @Zhou-Jiahao-1998 I have put this function into a file under src/utils, could you have a look?

frontend/src/utils/broken-link-disabler.js
```diff
@@ -0,0 +1,12 @@
+const mixin = {
+  methods: {
+    isBrokenLink(link) {
+      const linkFormat = /^(https:\/\/)(github.com|bitbucket.org|gitlab.com).*/;
```
gok99, line 4: This logic is already performed in the backend and stored in the report `summary.json` in `supportedDomainUrlMap` (see api.js for usage). You can probably just use that here too.

LuoZhijie-tom: OK, I have used the data in supportedDomainUrlMap instead

frontend/src/utils/broken-link-disabler.js
```diff
@@ -0,0 +1,18 @@
+window.$ = (id) => document.getElementById(id);
+const mixin = {
+  methods: {
+    isBrokenLink(link) {
```
gok99, line 4: Seems like the links provided to this function will already be filtered and are of type String | undefined (undefined only if the link is broken). Can't we just check that the links are not undefined to determine if they're broken?

LuoZhijie-tom: Oh I see, thanks!

Zhou-Jiahao-1998 (approved):
Thanks for the changes. LGTM!

frontend/src/components/v-ramp.vue
```diff
@@ -2,20 +2,36 @@
...
+          v-bind:title="getContributionMessage(slice, commit)",
+          v-bind:class="'ramp__slice--color' + getSliceColor(slice.date)",
+          v-bind:style="{\
+            zIndex: user.commits.length - j,\
+            borderLeftWidth: getWidth(commit) + 'em',\
+            right: ((getSlicePos(slice.date)\
+              + (getCommitPos(k, slice.commitResults.length))) * 100) + '%'\
+            }"
+        )
+        a.ramp__slice.broken-link(
```
dcshzj, line 20: I think this design can be improved further. Currently, the difference between the two code segments are that:
1. The unsupported link has an additional `.broken-link` class.
2. The cursor is set to `default` for the unsupported link.

I think this can be improved such that we can use a ternary operator to determine if the `.broken-link` class is needed, and the `cursor: default` definition can be moved into that class.

Alternatively, if this design were to remain, I would suggest to remove the attributes that are not relevant for the click action (such as `v-on:click`, `target`, etc). This can make it more obvious that the broken link is not meant to be clicked on.

frontend/src/utils/broken-link-disabler.js
```diff
@@ -0,0 +1,11 @@
...
+      return link === undefined;
+    },
+  },
+  data: {
+    disabledLinkMessage: 'This remote link is unsupported',
+  },
+};
+export default mixin;
```
dcshzj, line 11: We can just merge this with the first line, so just an `export default { ... }` instead of declaring a constant and exporting that constant.

dcshzj (changes requested):
In addition to the review comments, can you also provide a test report that showcases the functionality that you have added? The repositories that we are currently using are from GitHub, GitLab and Bitbucket. Feel free to include any small repository that are not hosted by the 3 hosting providers and provide us with a link to that report. Thanks!

LuoZhijie-tom:
> In addition to the review comments, can you also provide a test report that showcases the functionality that you have added? The repositories that we are currently using are from GitHub, GitLab and Bitbucket. Feel free to include any small repository that are not hosted by the 3 hosting providers and provide us with a link to that report. Thanks!

Hi @dcshzj I have added a new remote repository from sourceforge and modified the code according to your advice, could you have a look is this what you mean?

damithc:
@dcshzj @gerhean let's push this PR forward.

frontend/src/components/v-ramp.vue
```diff
@@ -2,20 +2,22 @@
...
-          right: ((getSlicePos(slice.date)\
-            + (getCommitPos(k, slice.commitResults.length))) * 100) + '%'\
-          }"
-      )
+      template(v-for="(commit, k) in slice.commitResults.filter(commitResult => commitResult.insertions > 0)")
+        a.ramp__slice(
+          draggable="false",
+          v-on:click="rampClick",
+          v-bind:href="getLink(commit)", target="_blank",
+          v-if="!isBrokenLink(getLink(commit))",
```
dcshzj, line 10: Can I check: Does this mean that if the repository is not supported, then there will not be any ramps shown on the dashboard for that repository?

LuoZhijie-tom: Oh yeah, if the repository is not supported, the remote link would be undefined. The link to blame view and link to history will also not shown. So I can remove the conditional statement for ramp, blame view and history component?

LuoZhijie-tom: @dcshzj Thank you for the review! I found a bug in my code that hide the broken link ramp, I have fixed it and the ramps with broken link are visible now. Could you have a look?

frontend/src/components/v-ramp.vue
```diff
@@ -2,20 +2,21 @@
...
-          right: ((getSlicePos(slice.date)\
-            + (getCommitPos(k, slice.commitResults.length))) * 100) + '%'\
-          }"
-      )
+      template(v-for="(commit, k) in slice.commitResults.filter(commitResult => commitResult.insertions > 0)")
+        a.ramp__slice(
+          draggable="false",
+          v-on:click="rampClick",
+          v-bind:href="getLink(commit)", target="_blank",
+          v-bind:title="getLinkMessage(getLink(commit), getContributionMessage(slice, commit))",
```
dcshzj, line 10: I think this one should still show the original contribution message. It is fine to show the commit title even when the ramp is not clickable.

dcshzj (changes requested):
Also, the `c-zoom.vue` commits panel still shows as a link even though it is not clickable. It should be changed to just a normal text. Screenshot:

Screenshot 2022-10-09 at 17 17 26

dcshzj:
Are you able to resolve the file conflicts?

LuoZhijie-tom:
Oh sorry, I just resolved the file conflicts

dcshzj (approved):
LGTM!

Issue #1665, opened 2022-02-13
Rename frontend component files to remove `v-`


gerhean:
**What feature(s) would you like to see in RepoSense?**

Rename frontend component files to remove the `v-` in front of every file.

I would also like some input on whether the names should be changed within the code, for example from `v-summary` to `summary` or some other standardized way to refer to components. Most reserved syntax used by Vue uses the `v-` prefix, like `v-bind`, so I would prefer to avoid it. Maybe use `c-` for component?

**Is the feature request related to a problem?**

Vue component files can easily be told apart from non-component files, as only component files use the `.vue` file extension. Hence the `v-` in front is redundant.

**Additional context**

Documentation needs to be changed as well.

LuoZhijie-tom:
Hi is this issue still open? It is OK to work on this issue?

dcshzj:
@LuoZhijie-tom Sure! I have assigned you this issue. Let us know if you have any questions!

LuoZhijie-tom:
Thanks! May I ask do I need to remove the "v-" prefix from all components in the front end? Or do I need to replace "-v" with "-c"?

dcshzj:
I think it would be good to just remove the "v-" prefix, not just for components but for all the files that contain it.

LuoZhijie-tom:
Hi can I add "-c" in front of each component? It turns out that "summary" is a build in html element, thus "v-summary" can't be renamed to "summary". But except for that, all other components can be renamed.

dcshzj:
Done in #1836.

PR #1834, opened 2022-09-04
[#1665] Rename frontend component files to remove v-

LuoZhijie-tom:
Add changes according to [#1665]

# Proposed commit message
The components are all named with a prefix `-v`; however, most reserved syntax used by Vue uses the `-v` .

Let's change the `-v` prefix to prefix `-c`, so that it will be less likely to cause confusion

# Other information
I tried to just remove the prefix previously, but the `summary` is a reserved word in html, thus the `v-summary` can't be renamed. So I add `-c` as a prefix.

### DIFF
--- a/frontend/src/views/v-authorship.vue
+++ b/frontend/src/views/v-authorship.vue
@@ -40,7 +40,7 @@
           type="radio",
           value="search",
           v-model="filterType",
-          v-on:change="indicateSearchBar(); updateSearchBarValue();"
+          v-on:change="indicateSearchBar"
         )
         .mui-form--inline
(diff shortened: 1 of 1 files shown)


Zhou-Jiahao-1998:
@LuoZhijie-tom Hi there. Could you resolve the conflict first?

LuoZhijie-tom:
Hi I'll resolve the conflict now, sorry I closed it by accident

LuoZhijie-tom:
Hi @Zhou-Jiahao-1998 , I have solved the conflict and created a new pull request

PR #1837, opened 2022-09-11
Changes reposense to always perform normal clone instead of bare clone

### LINKED ISSUE #1813: Bug: RepoSense cannot clone a repository

chan-j-d:
This is meant to be a prototype of a proposed change. This will indirectly fix #1813 if implemented.

### Analysis of bare cloning
For bare cloning, there are some minor memory space performance difference as bare clones do not need to set up the work tree. 

For reposense, it's about 76mb for bare and 110 for non bare. But for larger repos, the difference becomes insignificant. Like https://github.com/raspberrypi/firmware is a 20+GB repo and the difference is about 200mb with negligible cloning time difference.
Additionally, reposense bare clones every unique repo once before performing a local non-bare clone for each branch to be analyzed. The space advantage is only for the initial clone for checks. 
Seems like the bare cloning was implemented for the first time in https://github.com/reposense/RepoSense/commit/fed37581bbf17ae61a587256633be47924284b70 for checking repo data and was later expanded to check things like if a branch exists.

The primary advantage seems to be that the bare clone becomes almost an identical copy of the remote repo with the various remote branches available immediately. 
For non-bare clones, the remote branches are not easily accessible. For instance, you can do `git clone repo --branch some-remote-branch` for a local bare clone but not a local non-bare clone.
A workaround is to `git checkout some-remote-branch` which causes the local repo to create a copy of the remote branch called `some-remote-branch` after which the branched clone can be performed.

### Advantages of this change and implementing it
- Easier to understand `git clone` vs `git clone --bare`.
- Removes the `GitRevParse` class.
- Removes one test environment branch check which brings the test behavior closer to actual. 

### Disadvantages 
- The 'checkout' that creates the local copy of the branch might potentially cause concurrency issues if multiple checkouts occur and one fails. However, this is unlikely to occur as the checkout is quite fast.
- Some marginal increase in disc usage.
- A lot of refactoring due to many variable and method names and some parts of the UG/DG directly referencing the idea of a 'bare' clone.

Thoughts on whether we should move forward with this?


damithc:
> For reposense, it's about 76mb for bare and 110 for non bare.

This is a rather significant difference. How about speed difference?

chan-j-d:
It's about 6.5 sec for non bare to 6.0 for bare over 10 trials. 

Also it seems I did not test it properly and there are quite a lot of issues with this PR when running against multiple repos at once. Seems there might be hard naming references for bare clones and other branch checks that I was not aware of.

PR #1839, opened 2022-10-09
[#1594] Add guard in glob processing when generating authorship summary

### LINKED ISSUE #1594: Omit 'ignore globs' files from processing
This dashboard takes about 9 minutes, whereas it should take about 3-4 minutes only.
I suspect it is because one team has included a 9K data files https://github.com/AY2122S1-CS2113T-W12-2/tp/tree/master/data
However, even after I add `/data/**` to the 'Ignore globs' of that repo (in repo-config.csv), it still takes 9 minutes. So, I suspect even ignored files undergo heavy processing at first and are omitted only later in the processing pipeline. 
Is that the case?
If so, is it possible to omit them from processing earlier?

lawwm:
Fixes #1594 

## Proposed commit message

```
Ignored directories still undergo heavy processing despite being set to
ignored. 

Let's add guard clauses in authorship summary generation to prevent
ignored files from being processed.
```

## Other information

**Sample repository to test ignore globs:**
Repo with large files: https://github.com/lawwm/repoWithLargeFiles
Repo without large files: https://github.com/lawwm/repoWithoutLargeFiles
I stored the repo-config.csv file within the repository

**With large files, without globs:**
Elapsed processing time: 1 minute(s) 27.95 second(s)

**With large files, with globs:** `data/**;pdf/**`
Elapsed processing time: 22.03 second(s)

**Without large files:**
Elapsed processing time: 3.80 second(s)

**Comments:**
The timing required for a repository with large files even with globs is larger than that of a repository without those large files. 

Even with the skipping of processing within the generation of authorship and commits summary, there is probably still some processing in the git clone or checkout portion.

The pull request prevents the processing of large files that are ignored. However, it does not ignore processing for large amount of small files.


damithc:
@lawwm is this PR ready for review?

damithc:
@reposense/active-developer for your review ...

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -68,6 +68,23 @@ public void extractFileInfosTest() {
...
+        GitCheckout.checkout(config.getRepoRoot(), TEST_COMMIT_HASH);
+        List<FileInfo> files = FileInfoExtractor.extractTextFileInfos(config);
+
+        Assertions.assertEquals(4, files.size());
+
+        Assertions.assertTrue(isFileExistence(Paths.get("annotationTest.java"), files));
+        Assertions.assertTrue(isFileExistence(Paths.get("blameTest.java"), files));
+        Assertions.assertTrue(isFileExistence(Paths.get("inMasterBranch.java"), files));
+        Assertions.assertTrue(isFileExistence(Paths.get("newFile.java"), files));
+    }
```
chan-j-d, line 86: I think this is missing the check for files that are correctly ignored, such as some `.md` files or `newPos/*.java` files.

src/test/java/reposense/authorship/FileInfoExtractorTest.java
```diff
@@ -136,6 +153,22 @@ public void getEditedFileInfos_editFileInfoBranchSinceFebrauryEight_success() {
...
+
+        List<FileInfo> files = FileInfoExtractor.getEditedFileInfos(config, FEBRUARY_EIGHT_COMMIT_HASH);
+
+        Assertions.assertEquals(2, files.size());
+
+        Assertions.assertTrue(isFileExistence(Paths.get("annotationTest.java"), files));
+        Assertions.assertTrue(isFileExistence(Paths.get("newPos/movedFile.java"), files));
+
+        // file renamed without changing content, not included
+        Assertions.assertFalse(isFileExistence(Paths.get("renamedFile.java"), files));
```
chan-j-d, line 169: This is also missing the check that the `README.md` file is not present.

chan-j-d (changes requested):
Thanks for doing up! Some minor suggested changes.

Would also like to check the exact parameters that you used to demonstrate the performance improvement as I have been unable to replicate it. Thanks!

lawwm:
> Thanks for doing up! Some minor suggested changes.
> 
> Would also like to check the exact parameters that you used to demonstrate the performance improvement as I have been unable to replicate it. Thanks!

Thnak you for the review. I added the link to the repositories that I used to measure performance. I stored the `repo-config.csv` that I ran Reposense on within the repo.

chan-j-d (approved):
LGTM for large file sizes. I've tested it and it seems to do well.

However, I think the PR actually does not directly address #1594 which seems to be more to do with large number of small files.

 Try analyzing https://github.com/AY2122S1-CS2113T-W12-2/tp/ with the arguments `-s d1` (since day 1). The `data/**` files are no longer in the repo but you can still see them in the commit history and so they are still analyzed. Including `data/**` in the config removes them from analysis but the analysis time I had is about the same with this implementation and the master branch. Can consider looking into that if you have the time.

Consider changing the PR description as well to reflect this as well.

lawwm:
> However, I think the PR actually does not directly address #1594 which seems to be more to do with large number of small files.

Thank you for catching that, I failed to notice it and will be looking into it.

chan-j-d:
@lawwm thanks! Do create a separate PR for that though as I think this change is a pretty good standalone change

lawwm:
Hi, I'm not sure if I'm right but it seems to me like the analysis time for https://github.com/AY2122S1-CS2113T-W12-2/tp/ is the same because the time required for processing the commits is negligible.

I tested using two factors, using the ignore globs and using a different date. 

`--since d1` includes the `data/**` files in the commit history, while `--since 2/10/2021` excludes the `data/**` files in the commit history (both the addition and deletion commits).
```
--since d1 (with ignore globs)
11:07:33 - Elapsed processing time: 24.72 second(s)
11:08:31 - Elapsed processing time: 22.87 second(s)

--since d1
10:56:22 - Elapsed processing time: 21.53 second(s)
11:18:43 - Elapsed processing time: 24.44 second(s)

--since 2/10/2021 (with ignore globs)
11:09:14 - Elapsed processing time: 19.34 second(s)
11:09:49 - Elapsed processing time: 20.99 second(s)

--since 2/10/2021
11:05:40 - Elapsed processing time: 21.46 second(s)
11:17:14 - Elapsed processing time: 22.82 second(s)
```

Codewise, the ignore globs for generation of commit summary are used during a `GitLog.getWithFiles(config, author)` command which occurs quite early in the process within the `generateCommitSummary`, so the ignore globs does seem to ignore processing of commits.

Please correct me if I'm wrong or if I've missed out on something, thank you.

chan-j-d:
I'm not too sure what the exact issue and time disparity is supposed to be as well as I only checked the files in the repo in question. You might be able to get more insight by raising the question on the issue directly and tagging.

One possibility might be that the poor performance is specifically while the files themselves are present and not deleted yet since the date of the issue was raised just 1 day after the files were actually removed.

dcshzj (approved):
LGTM!

PR #1840, opened 2022-10-12
[#1131] Refactor GitTestTemplate to remove unnecessary methods

### LINKED ISSUE #1131: GitTestTemplate: `generateTestFileInfo` method is not reusable due to side effects
I wish to clarify why the `generateTestFileInfo` method (a utility method in the test code) also updates the RepoConfiguration object used in the tests.

https://github.com/reposense/RepoSense/blob/145a1b4dda02ac9fdc42554835e7be98a534b0a0/src/test/java/reposense/template/GitTestTemplate.java#L90-L100

Shouldn't each test case update the RepoConfiguration on its own, if needed?
I.e., if a test case needs to track `MAIN_AUTHOR` and `FAKE_AUTHOR`, the config can be updated separately in the test case?
Alternatively, perhaps we should document why `generateTestFileInfo` does this?

lawwm:
Fixes #1131

## Proposed commit message

```
The generateTestFileInfo method in the test suite mutates the
RepoConfiguration object when running tests.

Let's remove the helper methods as they were unnecessary and refactor
to avoid side effects.
```

## Other information

Decided to remove the utility methods in `GitTestTemplate.java` as they were unnecessary after removing the lines that mutate the `RepoConfiguration` object as recommended in the thread for the issue.


chan-j-d (commented):
Thanks for doing this up! 

I'm not sure if I agree with this implementation. I think the issue is not that `generateTestFileInfo` mutates the config but that it always mutates the config (usually for `FileAnalyzerTest`) even when other classes extending `GitTestTemplate` might not need this mutation. 

I think a much more straightforward fix would be to move the lines in the `generateTestFileInfo` method 
```
config.addAuthorNamesToAuthorMapEntry(new Author(MAIN_AUTHOR_NAME), MAIN_AUTHOR_NAME);
config.addAuthorNamesToAuthorMapEntry(new Author(FAKE_AUTHOR_NAME), FAKE_AUTHOR_NAME);
config.addAuthorNamesToAuthorMapEntry(new Author(IGNORED_AUTHOR_NAME), IGNORED_AUTHOR_NAME);
```

to the `@BeforeEach` tagged method in `FileAnalyzerTest` and retain the uses of `generateTestFileInfo`. What do you think?

lawwm:
Thank you for the review!
I shifted the config mutation to `@BeforeEach`, but I removed the `generateTestFileInfo` method since it only had one line left.

chan-j-d (approved):
LGTM!

Wanted to point out that the method being one line isn't a good enough reason to remove it. I think the method is at least useful in defaulting the behavior of generating `FileInfo` with the current test case's `config` and not every class extending `GitTestTemplate` has the `config` attribute by default. 

But I do think there is some benefit in explicitly showing the direct dependence on the `config` when generating the file info in test cases. At the very least, it shows that the order in which the lines of code are executed somewhat matters between each call of `generateFileInfo`.

dcshzj (approved):
LGTM!

PR #1841, opened 2022-10-21
[#1833] Add frontend tests for filter glob

### LINKED ISSUE #1833: Add Test Cases for Filter Glob User Input

#1810 was fixed by #1827. However, no frontend test cases were added for this PR due to the high priority.
Please add Cypress test cases to ensure this feature works as expected and facilitate future testing.

**Additional context**

Punpun1643:
Fixes #1833 

## Proposed commit message

```
There is currently no testing done for the new front-end change 
in filter glob related functionalities.

This may result in regression bugs in the future.

Let's add tests for these new frontend changes made for the filter glob functionalities
to ensure this feature works as expected and facilitate future testing.
```

### DIFF
--- a/frontend/cypress/tests/codeView/codeView_filterGlob.js
+++ b/frontend/cypress/tests/codeView/codeView_filterGlob.js
@@ -1,0 +1,106 @@
+describe('filter glob', () => {
+  it('check filter glob radio button is clickable', () => {
+    // open the code panel
+    cy.get('.icon-button.fa-code')
+      .should('be.visible')
+      .first()
+      .click();
+
+    // click on filter glob radio button
+    cy.get('.radio-button--search')
+      .should('be.visible')
+      .click();
+
+    // filter by file type should be unchecked
+    cy.get('.radio-button--checkbox')
+      .should('not.be.checked')
+  });
+
+  it('check no filter glob input should show all file types by default', () => {
+    cy.get('.icon-button.fa-code')
+      .should('be.visible')
+      .first()
+      .click();
+
+    cy.get('.radio-button--search')
+      .should('be.visible')
+      .click();
+
+    // the contents of all the files should be visible
+    cy.get('#tab-authorship .files', { timeout: 90000 })
+      .should('be.visible');
+  });
+
+  it('check filter glob input should be in focus after click', () => {
+    cy.get('.icon-button.fa-code')
+      .should('be.visible')
+      .first()
+      .click();
+
+    cy.get('.radio-button--search')
+      .should('be.visible')
+      .click();
+
+    // click on input box
+    cy.get('#search')
+      .click()
+      .should('have.focus');
+  });
+
+  it('check filter glob input should have default empty value', () => {
+    cy.get('.icon-button.fa-code')
+      .should('be.visible')
+      .first()
+      .click();
+
+    cy.get('.radio-button--search')
+      .should('be.visible')
+      .click();
+
+    cy.get('#search')
+      .should('have.value', '');
+  });
+
+  it('check filter nothing should show all file types', () => {
+    cy.get('.icon-button.fa-code')
+      .should('be.visible')
+      .first()
+      .click();
+
+    cy.get('.radio-button--search')
+      .should('be.visible')
+      .click();
+
+    // enter on empty input
+    cy.get('#search')
+      .type('{enter}');
+
+    cy.get('#tab-authorship .files', { timeout: 90000 })
+      .should('be.visible');
+  });
+
+  it('check invalid glob should not show any files', () => {
+    cy.get('.icon-button.fa-code')
+      .should('be.visible')
+      .first()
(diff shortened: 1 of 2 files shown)


frontend/cypress/tests/codeView/codeView_filterGlob.js
```diff
@@ -0,0 +1,177 @@
...
+
+    cy.get('.radio-button--search')
+        .should('be.visible')
+        .click();
+
+    cy.get('#search')
+        .should('have.value', '');
+  });
+
+  it('check request to filter nothing by enter should show all file types', () => {
```
Zhou-Jiahao-1998, line 64: I think you can extend this test to change its input from something to nothing. That way we can be sure that the filter empty input will return all files rather than not doing any changes to the filter result.

frontend/cypress/tests/codeView/codeView_filterGlob.js
```diff
@@ -0,0 +1,177 @@
...
+
+    cy.get('#submit-button')
+        .type('{enter}');
+
+    // no file should be shown
+    cy.get('#tab-authorship .files')
+        .should('not.be.visible');
+  });
+
+  it('check request to filter invalid glob by clicking should not show any files', () => {
```
Zhou-Jiahao-1998, line 124: I feel that one test to check if click or enter works will do. After that, we can just test one of it. But this way is fine for me as well. Just that it felt a little redundant to do so.

Zhou-Jiahao-1998 (changes requested):
Sorry for the late review! Please look at my comments and let me know if you have any comments/clarification. Drop me a dm on slack if I do not reply within 2 days.

frontend/cypress/tests/codeView/codeView_filterGlob.js
Punpun1643: Yup I agree, will work on this!

frontend/cypress/tests/codeView/codeView_filterGlob.js
```diff
@@ -0,0 +1,177 @@
...
+        .should('be.visible')
+        .first()
+        .click();
+
+    cy.get('.radio-button--search')
+        .should('be.visible')
+        .click();
+
+    // the contents of all the files should be visible
+    cy.get('#tab-authorship .files', { timeout: 90000 })
```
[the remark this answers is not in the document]
Punpun1643: For this, it is time to wait before the `get` timeout. I think it is ok to remove it as well. But I also thought that we could include the `timeout` option in case the loading of content takes some time to get resolved. `90000` is consistent with other previous tests that try to load files at `#tab-authorship` as well. Feel free to let me know what you think :)

frontend/cypress/tests/codeView/codeView_filterGlob.js
Punpun1643: Hmm the reason I include test for click and enter both times for empty inputs and invalid inputs was that I think invalid inputs, we could check whether both click and enter work. 

For empty input, these tests were added because I remember the filter glob wasn't even clickable and enterable in the first place and this means that all files were present. Entering and clicking for empty inputs is to check whether the button and input work as expected. 

I lean a little towards including these tests, but feel free to let me know as well :)

Zhou-Jiahao-1998 (approved):
LGTM!

frontend/cypress/tests/codeView/codeView_filterGlob.js
```diff
@@ -0,0 +1,198 @@
...
+
+    cy.get('.radio-button--search')
+        .should('be.visible')
+        .click();
+
+    cy.get('#search')
+        .should('have.value', '');
+  });
+
+  it('check request to filter something to nothing by enter should show all file types', () => {
```
dcshzj, line 64: Can this test name be reworded? I'm not sure what it does by reading this description.

frontend/cypress/tests/codeView/codeView_filterGlob.js
```diff
@@ -0,0 +1,198 @@
...
+
+    // enter on empty input
+    cy.get('#search')
+        .type('{enter}');
+
+    cy.get('#tab-authorship .files', { timeout: 90000 })
+        .should('be.visible');
+  });
+
+  it('check request to filter something to nothing by clicking should show all file types', () => {
```
dcshzj, line 94: Same for this as well, I think you can be more explicit about it originally having something, then later resetting to empty value or something like that.

dcshzj (changes requested):

frontend/cypress/tests/codeView/codeView_filterGlob.js
Punpun1643: Hi, the descriptions are updated in the new commit

Punpun1643:
@dcshzj thank you for the review! I have made the changes to the test case descriptions.

PR #1846, opened 2022-12-15
[#1118] Modify regex and respective tests

### LINKED ISSUE #1118: Use less restrictive regex for author name in @@author tags
Currently, the regex used to validate names in `@@author some-randomeuser` tags is:

 This regex ensures the author name has at least 2 consecutive alphanumeric characters.

As discussed in this [comment](https://github.com/reposense/RepoSense/pull/1067#discussion_r382583692), we can consider using a less restrictive regex, so that tags such as `@@author --myname--` are accepted.

RyanCheungJF:
Fixes #1118

## Proposed commit message

```
Currently, the regex restriction for author names is too strict.
It does not allow unconventional names for authors.

Let us remove this restriction to allow any valid author name
to be accepted.
```

## Other information

When I was looking through the tests, I realized some of them depended on `annotationTest.java`. I tried changing the values declared inside of the file, but am unable to do so as it will reset everytime I run the tests.

As `annotationTest.java` uses an author name of `-invalidGitUsername_TreatedAsUnknownUser`, I've thus created a new Author object for this unconventional name and used it in the test cases.

The changes in the json files are with respect to `gradlew systemtest`. 

Previously, since it was `UNKNOWN_AUTHOR_GIT_ID = "-";`, and now it is `UNCONVENTIONAL_AUTHOR_NAME = "-invalidGitUsername_TreatedAsUnknownUser";`.


src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -46,6 +46,7 @@ public class GitTestTemplate {
     protected static final String TEST_COMMIT_HASH_PARENT = "c5a6dc774e22099cd9ddeb0faff1e75f9cf4f151";
     protected static final String MAIN_AUTHOR_NAME = "harryggg";
     protected static final String FAKE_AUTHOR_NAME = "fakeAuthor";
+    protected static final String UNCONVENTIONAL_AUTHOR_NAME = "-invalidGitUsername_TreatedAsUnknownUser";
```
HCY123902, line 49: Thanks for the contribution. May I propose changing -invalidGitUsername_TreatedAsUnknownUser to another author name value such as -unconventionalAthorName since the former is no longer invalid or treated as an unknown author with the new regex.

HCY123902, line 49: Thanks for the contribution. May I propose changing -invalidGitUsername_TreatedAsUnknownUser to another author name value such as -unconventionalAuthorName since the former is no longer invalid or treated as an unknown author with the new regex. Otherwise, this may cause confusion

HCY123902 (changes requested):

src/test/java/reposense/template/GitTestTemplate.java
RyanCheungJF: Hi, I believe I didn't change it yet because the original value is used in `annotationTest.java` from the test repository. I can clone the test repository and make a PR there, but could I ask how I could run `gradlew test` using my cloned local version of the test repository?

HCY123902: I think it is sufficient to run gradlew in the main repository. The test repositories are for CI dashboard preview

src/systemtest/resources/LocalRepoSystemTest/testRelativePathing/parent1_test-repo_master/authorship.json
```diff
@@ -1 +1 @@
...
...nMap":{"fakeAuthor":9,"-invalidGitUsername_TreatedAsUnknownUser":5,"harryggg":5,"-":3}},{"path":"blameTest.java","fileType":"java","lines":[{"lineNumber":1,"author":{"gitId":"harryggg"},"content":"line 1"},{"lineNumber":2,"author":{"gitId":"harryggg"},"content":"line 2"},{"lineNumber":3,"author":{"gitId":"fakeAuthor"},"content":"fake"},{"lineNumber":4,"author":{"gitId":"harryggg"},"content":"line 3"}],"authorContributionMap":{"fakeAuthor":1,"harryggg":3}},{"path":"newFile.java","fileType":"java","lines":[{"lineNumber":1,"author":{"gitId":"harryggg"},"content":"line 1"},{"lineNumber":2,"author":{"gitId":"harryggg"},"content":"line 2"}],"authorContributionMap":{"harryggg":2}},{"path":"newPos/movedFile.java","fileType":"java","lines":[{"lineNumber":1,"author":{"gitId":"harryggg"},"content":"line 1"},{"lineNumber":2,"author":{"gitId":"harryggg"},"content":"line 2"},{"lineNumber":3,"author":{"gitId":"harryggg"},"content":"line 3"},{"lineNumber":4,"author":{"gitId":"harryggg"},"content":"line 4"}],"authorContributionMap":{"harryggg":4}},{"path":"space test.txt","fileType":"txt","lines":[{"lineNumber":1,"author":{"gitId":"chan-j-d"},"content":"1"}],"authorContributionMap":{"chan-j-d":1}}]
```
HCY123902, line 1: I think you may need to have a new line for this. Otherwise, style check in CI will throw an error. The same applies to other JSON and scripts here

HCY123902 (changes requested):

RyanCheungJF:
Hi! For future reference, Si Kai has offered to take over the issue, so I'll be closing this PR. The new PR can be found [here](https://github.com/reposense/RepoSense/pull/1857).

Issue #1484, opened 2021-03-18
Author profile not found in deployed dashboard


Tejas2805:
The deployed dashboard doesn't show the profile for the author Eugene Peh. It works for all other authors. 

The link that should open should be [this link](https://github.com/EugenePeh) instead of [the wrong link](https://github.com/Eugene%20Peh).

Notice the addition of `%20` in the latter URL.

HCY123902:
Maybe this particular author can be ignored using the `repo-config.json` in this case? Not sure about the exact cause but it seems that the Github Id for this particular author generated in the json file is incorrect.

gerhean:
%20 is hexadecimal for ' ' or whitespace, seems like someone did not enter the correct information in `author-config.csv`

gerhean:
Referring to: https://api.github.com/repos/reposense/testrepo-Alpha/commits

Seems like backend uses `commit.author.name` to determine the gitId.
This is consistent with the fakeauthor's username which leads to a person who never contributed to reposense.

However it should be using `author.login` instead.

clyveycui:
After some investigation, I think with the current way RepoSense works, it is impossible to extract out github id if it is different from author name.

Currently the gitId/Author name is sourced from `git log` command, which provides the author name used when committing. In the the case that the github id is different from the author name (in the case of 'eugenepeh' and 'Eugene Peh' for example) there is simply no information provided by git that allows us to get the correct github id.

A possible solution is to request more information about the commit via the Github API as seen in gerhean's comment:
> Referring to: https://api.github.com/repos/reposense/testrepo-Alpha/commits
This will require the need to work with not just Github, but also potentially GitLab's and BitBucket's API.
(Another minor caveat is the need to differentiate between committer and author, but this shouldn't be too important.)

Another possibility is to add some sort of configuration file that maps the author names to the correct github API. This configuration file would then need to be set up for each repository that needs to be analyzed as long as there is a difference between the author name and the Github id.

damithc:
My understanding is, `author-config.csv` already provides a way to specify the mapping between a GitHub username and author names in the commits. This specific problem might be due to an error/omission in the `author-config.csv`.

clyveycui:
It seems that eugene peh is missing from the author-config.csv, so that's probably the issue.

What about using github api? Is there any form of consideration to move towards that solution?

damithc:
> What about using github api?

[...]

PR #1849, opened 2022-12-15
Add functionality to take in mapping of Author name to git id

clyveycui:
Part of #1484

## Proposed commit message

```
Corrects git id associated with the authors in the report

The git id of the authors are incorrect in the generated report in certain cases

This can lead to broken links on the report, reducing the user experience

Let's correct the associated git id for the authors
```

## Other information

This is a proof of concept using the config file solution as mentioned in [this comment](https://github.com/reposense/RepoSense/issues/1484#issuecomment-1353696437). 

Before the fix:
The two authors in red boxes have incorrect github id
image

After the fix:
The two authors now have the correct github id
image

### DIFF
--- a/config/author-config.csv
+++ b/config/author-config.csv
@@ -1,12 +1,2 @@
 Repository's Location,Branch,Author's Git Host ID,Author's Emails,Author's Display Name,Author's Git Author Name,Ignore Glob List
-https://github.com/reposense/testrepo-Beta.git,master,nbriannl,,"Ashwin Mukadaan Singhania, Rathod",,
-https://github.com/reposense/testrepo-Beta.git,master,zacharytang,zacharytang.tc@gmail.com,Yagami Rukei Polland,,src/main**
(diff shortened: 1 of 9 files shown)


clyveycui:
Still need to write tests and redesign the architecture. Submitted as a proof of concept.

chan-j-d:
As mentioned here https://github.com/reposense/RepoSense/issues/1484#issuecomment-1354114804, the author can declare a git ID and various git names under it which will all be mapped to that author. Under `AuthorConfiguration`, you can see the parsing of the `author-config.csv` which allows for this mapping. 

There are couple of additional complications with finding the git name vs git host id. If you merge a PR on GitHub, your GitHub name becomes the author of the commit in `git log` while if you commit something locally, you'll use your local git ID which is likely different. I believe our general stance is to use whatever `git` tells us without using any external APIs since we have recently added support for other remote repositories sites like GitLab which makes it difficult to support it. 

Really appreciate the effort though! Thanks!

dcshzj:
Hi @clyveycui, are you able to resolve the issues with the file formatting?

clyveycui:
Hi, unfortunately I did not fix the issue with file formatting

PR #1852, opened 2022-12-17
[#1499] Declare vue prop types explicitly

### LINKED ISSUE #1499: Declare props explicitly and give proptypes

For props to be explicitly declared for every component, and for the proptype to be added.

Declare at the top like this:
image

https://vuejs.org/v2/guide/components-props.html

ckcherry23:
Fixes #1499 

## Proposed commit message

```
In the current version of RepoSense, props are declared explicitly for
Vue components as an array of strings. 

Let's declare props using the object syntax so that we can document the
prop types.

This will also show a warning in the browser console if props of the
wrong type are passed to the component. 

Let's also define classes for `Segment` and `User` for stricter type 
validation.
```

## Other information

This is a simple enhancement of the code quality. 

Perhaps one doubt I have is that the `c-ramp` component is used with different data types (`Number`, `String`) for `avgsize` in different cases according to the console. Currently, I have allowed `avgsize` to take both data types, but please let me know if a change is required.


Zhou-Jiahao-1998 (changes requested):
@ckcherry23 hi! can you resolve all the warnings shown by GitHub Actions? you need to either mark the prop as `required: true` or add in a default value.

frontend/src/components/c-ramp.vue
```diff
@@ -39,7 +39,44 @@ import brokenLinkDisabler from '../mixin/brokenLinkMixin.ts';
...
   mixins: [brokenLinkDisabler],
   name: 'c-ramp',
-  props: ['groupby', 'user', 'tframe', 'avgsize', 'sdate', 'udate', 'mergegroup', 'fromramp', 'filtersearch'],
+  props: {
+    groupby: {
+      type: String,
+      default: 'groupByRepos',
+    },
+    user: {
+      type: Object,
```
Zhou-Jiahao-1998, line 48: I think it will be more helpful to define a type for the places that you need to put `Object`. This will help a lot with the typescript migration in future as well. You can perhaps define the types in a separate util file in `frontend`.

Zhou-Jiahao-1998 (changes requested):
@ckcherry23 Do let me what you think. Thank you!

frontend/src/components/c-ramp.vue
ckcherry23: Yes, that sounds better. I have added classes for those objects with pre-defined attributes to have better type validation.

frontend/src/utils/segment.js
```diff
@@ -0,0 +1,13 @@
+export default class Segment {
```
Zhou-Jiahao-1998, line 1: Not sure if this will be too much work but I think it might be better to just declare this file in `.ts` so that we don't need to rewrite this file again in future. @gerhean @dcshzj any thoughts?

ckcherry23: I have written the classes in TypeScript now. Please take a look!

frontend/src/utils/user.ts
```diff
@@ -0,0 +1,37 @@
+export default class User {
+  checkedFileTypeContribution : number;
+
+  commits: Array<object>;
+
+  dailyCommits: Array<object>;
+
+  displayName: string;
+
+  fileTypeContribution: object;
```
Zhou-Jiahao-1998, line 10: I was thinking about not having any `object` type here but I think it can be done another PR.

Zhou-Jiahao-1998 (approved):

Zhou-Jiahao-1998 (approved):
LGTM!

frontend/src/utils/user.ts
ckcherry23: I believe `fileTypeContribution` had nested objects so I did not implement them yet. Will work on it later.

yhtMinceraft1010X:
I merged the changes from `master` branch in order to test if CI passes following the fix for #1853.

CI now appears to be failing for Codecov uploads. Might be [an issue on their end](https://github.com/codecov/codecov-action/issues/897).

HCY123902:
> I merged the changes from `master` branch in order to test if CI passes following the fix for #1853.
> 
> CI now appears to be failing for Codecov uploads. Might be [an issue on their end](https://github.com/codecov/codecov-action/issues/897).



> I merged the changes from `master` branch in order to test if CI passes following the fix for #1853.
> 
> CI now appears to be failing for Codecov uploads. Might be [an issue on their end](https://github.com/codecov/codecov-action/issues/897).

Let's wait for their reply. At the same time, I think this pull request is ready for merging

yhtMinceraft1010X:
> Thanks for the check. I think this pull request is ready for merging while waiting for the reply regarding the upload issue

All CI is now passing. It appears there were some intermittent issues which did not recur after re-running CI.

PR #1854, opened 2023-01-04
[#1843] Fix authorship label LoC

### LINKED ISSUE #1843: Code view LoC numbers don't match the ramp view LoC numbers
Example https://nus-cs2113-ay2223s1.github.io/tp-dashboard

image

zhoukerrr:
Fixes #1843 

## Proposed commit message

```
The labels on top of the files in `c-authorship.vue` display the LoC of
all the files that the author has written to. This is not the intended
action.

Let us update the logic for this calculation to ensure that the labels
correctly display the LoC.
```

### DIFF
--- a/frontend/src/views/c-authorship.vue
+++ b/frontend/src/views/c-authorship.vue
@@ -328,6 +328,11 @@ export default {
         files = await window.api.loadAuthorship(this.info.repo);
       }
 
+      const author = repo.users.find((user) => user.name === this.info.author);
+      if (author) {
+        this.authorDisplayName = author.displayName;
+      }
+
       this.processFiles(files);
 
       if (this.info.isRefresh) {
--- a/frontend/src/views/c-authorship.vue
+++ b/frontend/src/views/c-authorship.vue
@@ -327,7 +327,7 @@ export default {
       if (!files) {
         files = await window.api.loadAuthorship(this.info.repo);
       }
-      this.getRepoProps(repo);
+
       this.processFiles(files);
 
       if (this.info.isRefresh) {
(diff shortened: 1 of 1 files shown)


frontend/src/views/c-authorship.vue
```diff
@@ -339,34 +339,6 @@ export default {
       this.setInfoHash();
     },
 
-    getRepoProps(repo) {
```
HCY123902, line 342: Can I ask whether it is safe to remove the getRepoProps method? Where will the logic handling authorDisplayName and filtering logic of author files go in this case

Zhou-Jiahao-1998: @HCY123902 Thanks for catching the mistake here! I forgot about `authorDisplayName`. The logic is added back now.

As for the filtering logic, the `processFiles` function already does the filtering.

HCY123902 (approved):

frontend/src/views/c-authorship.vue
HCY123902: Regarding the filtering logic for MergeGroup, the original practice is to directly use `repo.files`, while `isValidFile` only includes files with contributions from at least 1 known author. Can you verify that they are consistent? Alternatively, will this adjustment cause behavioral changes on the dashboard

Zhou-Jiahao-1998: @HCY123902 They are consistent. `repo.files` contains all the files of the repo. This includes all the files that someone within the team has written to. It should work the same as before. If a file is not touched by anyone from the team, it should not show up in the first place.

HCY123902: Thanks for the clarification

dcshzj:
@zhoukerrr Sorry, I don't have a lot of context about the frontend. Are you able to explain the differences between the old and new behavior? Ideally with screenshots or steps to reproduce. Thanks!

Also, it looks like there are issues with cloning the reposense/RepoSense repo, are you able to look into it as well?

zhoukerrr:
> @zhoukerrr Sorry, I don't have a lot of context about the frontend. Are you able to explain the differences between the old and new behavior? Ideally with screenshots or steps to reproduce. Thanks!
> 
> Also, it looks like there are issues with cloning the reposense/RepoSense repo, are you able to look into it as well? Seems to be this error but needs more investigation:
> 
> ```
> fatal: fetch-pack: invalid index-pack output
> ```

@dcshzj you can refer to my reply in #1843 for more details on the difference in behaviour. You can verify that the changes are indeed correct in the deployment as well. 

As for the issue with `reposense/RepoSense`, may I know how do I reproduce it? The deployment does not include that repo from what I can see. This PR does not involve any changes to the backend logic so I am not too sure what you are referring to here.

dcshzj (approved):
Ah the issue description provides the needed context, thanks! Also, I think the RepoSense repo was an intermittent issue which seems to be fixed.

PR #1855, opened 2023-01-08
[#1853] Change git log author regex type

### LINKED ISSUE #1853: Failing CI for MacOS only

* **RepoSense Version:** `master` branch, on GitHub Actions
* **OS and Version:** MacOS 12
* **Web Browser and Version (if applicable):** Google Chrome Version 108.0.5359.125

1. Run continuous integration on GitHub Actions

No errors

CI fails but only for MacOS. Here is a sample [CI report](https://github.com/reposense/RepoSense/actions/runs/3720179430) with the error.

The screenshot below shows that there is an AssertionFailedError for certain tests in `CommitInfoAnalyzerTest.java`:
image

The screenshot below shows that certain tests in `GitLogTest.java` also have AssertionFailedError:
image

The first instance of this CI failure occurred on December 17 2022 at 9.08 pm. The last successful CI run was on the same day, but at 5.50 am:
image

This cannot be seen from looking at the picture alone. It is necessary to go to the [workflows page](https://github.com/reposense/RepoSense/actions/workflows/integration.yml), find the workflow runs for #1844 and #1499 and hover the cursor over the "2 weeks ago" line.

yhtMinceraft1010X:
Fixes #1853

## Proposed commit message

```
In Git 2.39, there was a change to the regex library used for MacOS.
This may have caused the git log command with the --author=<pattern>
option, which interprets <pattern> as basic regex, to return nothing on
MacOS.

Let's switch the pattern to extended regex to restore the broken
functionality.
```

## Description of probable cause of MacOS CI problem.

One [pull request on Git](https://github.com/git/git/pull/1319) changed the regex library used on MacOS to the native MacOS regex library, from the original Git regex library. This change was part of [Git release 2.39](https://github.com/git/git/blob/v2.39.0/Documentation/RelNotes/2.39.0.txt).

As a result, the way basic regex patterns was interpreted in MacOS `git log` may have changed. This could result in the `git log` command not giving back any results when the `--author` option was used.

I suspect that this was what caused our MacOS CI in #1853 to fail since the MacOS CI uses Git 2.39. Given that the regex library used for Windows and Linux was not changed, this may have been the reason why the failure was exclusive to MacOS CI.

## Special Credit
I would like to thank @zhoukerrr for helping to diagnose and solve the issue. Personally, I do not have a Mac at home so trying to identify the root cause of the problem was inefficient for me as I would have to push any debugging code onto my remote fork and observe the MacOS CI.

@zhoukerrr helped to manually run the `git log` command and identify that the `--author` option was causing `git log` on Git 2.39 MacOS to fail. 

I would also like to thank @dcshzj for identifying Git 2.39 as the version which causes MacOS CI to fail.

## A primer on basic vs extended regex
Below was the site used to compare basic and extended regex:
* https://www.gnu.org/software/sed/manual/html_node/BRE-vs-ERE.html

### DIFF
--- a/src/main/java/reposense/git/GitLog.java
+++ b/src/main/java/reposense/git/GitLog.java
@@ -30,7 +30,7 @@ public class GitLog {
     public static String get(RepoConfiguration config, Author author) {
         Path rootPath = Paths.get(config.getRepoRoot());
 
-        String command = "git log --no-merges -i ";
+        String command = "git log --no-merges -i --extended-regexp ";
         command += GitUtil.convertToGitDateRangeArgs(config.getSinceDate(), config.getUntilDate(), config.getZoneId());
         command += " --pretty=format:" + PRETTY_FORMAT_STRING + " --shortstat";
         command += GitUtil.convertToFilterAuthorArgs(author);
@@ -47,7 +47,7 @@ public class GitLog {
     public static String getWithFiles(RepoConfiguration config, Author author) {
         Path rootPath = Paths.get(config.getRepoRoot());
 
-        String command = "git log --no-merges -i ";
+        String command = "git log --no-merges -i --extended-regexp ";
         command += GitUtil.convertToGitDateRangeArgs(config.getSinceDate(), config.getUntilDate(), config.getZoneId());
         command += " --pretty=format:" + PRETTY_FORMAT_STRING + " --numstat --shortstat";
         command += GitUtil.convertToFilterAuthorArgs(author);
--- a/src/main/java/reposense/git/GitUtil.java
+++ b/src/main/java/reposense/git/GitUtil.java
@@ -58,6 +58,7 @@ class GitUtil {
 
     /**
      * Returns the {@code String} command to specify the {@code author} to analyze for `git log` command.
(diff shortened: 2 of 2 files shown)


chan-j-d (commented):
I think it might be worth adding a comment in `GitUtil` indicating that the regex type is extended regex since the flag is added elsewhere

chan-j-d (approved):
LGTM! Great work identifying and fixing the bug

dcshzj:
@yhtMinceraft1010X Thanks for this fix! Can I check whether this breaks backwards compatibility? Will using `--extended-regexp` throw an error for older Git versions?

yhtMinceraft1010X:
> @yhtMinceraft1010X Thanks for this fix! Can I check whether this breaks backwards compatibility? Will using `--extended-regexp` throw an error for older Git versions?

Based on what I observed for the [`git log` documentation for version `2.19.0`](https://git-scm.com/docs/git-log/2.19.0#Documentation/git-log.txt---extended-regexp), which is the oldest version i could find, the `--extended-regexp` flag was already a thing back then.

I don't think it should be a problem

dcshzj (approved):
Nice, LGTM!

PR #1857, opened 2023-01-13
[#1118] Remove authorship tag validation regex

### LINKED ISSUE #1118: Use less restrictive regex for author name in @@author tags

sikai00:
Fixes #1118

```
GitHub's IDs have certain requirements to be valid, but Git does not
have such requirements. This is evident in .gitconfig files, where 
the username can be set to anything the user may like.

The current validation regex is based off GitHub's ID requirement. 
However, we are currently relying on Git author names found in 
.gitconfig files to attribute code authorship. This may result in 
missing code authorship attribution, when the Git author names do 
not match GitHub's ID requirements.

Let's remove this restriction to allow any valid author names to be
accepted and properly attributed.
```

#### Accepting usernames with whitespaces
This is the main reason for this PR: Previous PRs did not allow usernames with whitespaces. However, I find those to be relatively common.
image
Given that the issue was to allow any usernames as RepoSense relies on Git author names rather than GitHub IDs, I believe usernames with whitespaces should be allowed.

However, it is done within reason: Usernames with the respective closing comment syntax may lead to unexpected results. For example, `<!--  @@author   fake-4u-th0r-->  -->`  will be extracted as `fake-4u-th0r` , though some may have expected `fake-4u-th0r-->`. This is because in a such a scenario, there should only be a single close comment, even in the source file itself. 

~~#### Slightly messier code~~
~~`array[1].trim().split("", 1))` found in `AnnotatorAnalyzer.java`, line 109: I am unsure of a cleaner way to create this String[]. It will require a messier syntax involving return statements and braces within the lambda expression to do new String[] { array[1] }.~~

#### Using `null` as an identifier for end of comment for comment formats without closing comment syntax
I used `null` here as a special value to indicate the end of a comment for certain comment formats without closing comment syntax. While this leads to slightly messier code, I believe it is more accurate to the comment's actual syntax, as a comment do not necessarily end on any whitespace (the old end of comment identifer was a `\s`).

I explored other means of doing this, such as overloading `AnnotatorAnalyzer#generateCommentRegex` with a version with only `commentStart`. But I found that those were less than satisfactory, as it also further involved changing the size of `COMMENT_FORMAT`, resulting in arrays of different shapes.

#### System tests
As mentioned in a [previous PR for the same issue](https://github.com/reposense/RepoSense/pull/1846), there were some reliance on a separate repo to pass the system test. For that reason, I have yet to update the author names with a better name as noted in the [code review for the previous PR](https://github.com/reposense/RepoSense/pull/1846/files#r1063225218) as I am unsure if how I should proceed with the updating the separate repos.

`commits.json` and `authorship.json` for the (previously) failing 2 system tests, `LocalRepoSystemTest` were updated by generating with the same parameters as found in `LocalRepoSystemTest`. This is due to the validation changes, as the system test relied on a separate repo (testrepo-Alpha), which are slightly outdated in terms of its intentions, as they also checked for invalid usernames which are now valid.


src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
```diff
@@ -189,6 +232,10 @@ public void extractAuthorName_matchCommentPattern4_returnAuthorName() {
         line = "  %  @@author   fake-4u-th0r  ";
```
yhtMinceraft1010X, line 232: You may want to include a test case for ` line = " % @@author fake-4u-th0r"` (without the ending whitespace) in order to check the new comment format of
```
{"%", null}
```

yhtMinceraft1010X (changes requested):
Generally excellent, although there is one test case which I think needs to be added.

src/test/java/reposense/authorship/AnnotatorAnalyzerTest.java
yhtMinceraft1010X: In addition, please update https://reposense.org/ug/usingAuthorTags.html to reflect the use of Git author names rather than GitHub IDs.

sikai00: Hi, sorry for the late response, but can I check if this case is covered by line 224 (`line = "%@@author fakeAuthor";`)? Because the case without the ending whitespace I think is sufficiently covered by that.

`line = " % @@author fake-4u-th0r"` Or did you mean to have a case where there is a whitespace at the beginning, and no whitespace at the end?

sikai00: Will update the necessary md documents, thanks!

yhtMinceraft1010X: > Or did you mean to have a case where there is a whitespace at the beginning, and no whitespace at the end?

Yup.

yhtMinceraft1010X (approved):
LGTM. Note that #1851, which may be merged anytime, may result in merge conflicts with your PR.

src/test/java/reposense/template/GitTestTemplate.java
```diff
@@ -46,6 +46,8 @@ public class GitTestTemplate {
     protected static final String TEST_COMMIT_HASH_PARENT = "c5a6dc774e22099cd9ddeb0faff1e75f9cf4f151";
     protected static final String MAIN_AUTHOR_NAME = "harryggg";
     protected static final String FAKE_AUTHOR_NAME = "fakeAuthor";
+    protected static final String UNCONVENTIONAL_AUTHOR_NAME = "-invalidGitUsername_TreatedAsUnknownUser";
```
HCY123902, line 49: I personally still suggest changing this name since it does not reflect the new logic. You will need to submit a pull request to modify https://github.com/reposense/testrepo-Alpha/blob/master/annotationTest.java to address that

HCY123902 (changes requested):

src/test/java/reposense/template/GitTestTemplate.java
sikai00: I have changed the names to better reflect the new authors and their respective names. As a result there should be issues with the automated checks on GitHub. I have made a [separate PR](https://github.com/reposense/testrepo-Alpha/pull/14) to the testrepo. Thanks for your suggestion!

sikai00: The 2 failing tests are found in `AnnotatorAnalyzerTest.java`, as the names now do not match with what is found in `annotationTest.java`.

src/systemtest/resources/LocalRepoSystemTest/testRelativePathing/parent1_test-repo_master/authorship.json
```diff
@@ -1 +1 @@
...
...GitUsername_TreatedAsUnknownUser":5,"harryggg":5,"harryggg invalidAuthorLineFormat":3}},{"path":"blameTest.java","fileType":"java","lines":[{"lineNumber":1,"author":{"gitId":"harryggg"},"content":"line 1"},{"lineNumber":2,"author":{"gitId":"harryggg"},"content":"line 2"},{"lineNumber":3,"author":{"gitId":"fakeAuthor"},"content":"fake"},{"lineNumber":4,"author":{"gitId":"harryggg"},"content":"line 3"}],"authorContributionMap":{"fakeAuthor":1,"harryggg":3}},{"path":"newFile.java","fileType":"java","lines":[{"lineNumber":1,"author":{"gitId":"harryggg"},"content":"line 1"},{"lineNumber":2,"author":{"gitId":"harryggg"},"content":"line 2"}],"authorContributionMap":{"harryggg":2}},{"path":"newPos/movedFile.java","fileType":"java","lines":[{"lineNumber":1,"author":{"gitId":"harryggg"},"content":"line 1"},{"lineNumber":2,"author":{"gitId":"harryggg"},"content":"line 2"},{"lineNumber":3,"author":{"gitId":"harryggg"},"content":"line 3"},{"lineNumber":4,"author":{"gitId":"harryggg"},"content":"line 4"}],"authorContributionMap":{"harryggg":4}},{"path":"space test.txt","fileType":"txt","lines":[{"lineNumber":1,"author":{"gitId":"chan-j-d"},"content":"1"}],"authorContributionMap":{"chan-j-d":1}}]
```
HCY123902, line 1: Will you need to modify these JSON files as well after merging the changes in https://github.com/reposense/testrepo-Alpha/pull/14

sikai00: Yes my apologies for missing that! Will update it

HCY123902: Actually, after a second thought, I think this may become too tedious for a single change of author name ...... You may revert back to the original -invalidGitUsername_TreatedAsUnknownUser and close https://github.com/reposense/testrepo-Alpha/pull/14 accordingly. I will approve the pull request after that. Sorry for the inconvenience

sikai00: Ah, no worries! Can I check then, should I still continue with making the names better in a separate PR?

HCY123902 (approved):

src/systemtest/resources/LocalRepoSystemTest/testRelativePathing/parent1_test-repo_master/authorship.json
HCY123902: I do not think that is necessary

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
```diff
@@ -102,18 +99,23 @@ public static Optional<String> extractAuthorName(String line) {
                 .map(l -> l.split(AUTHOR_TAG))
                 .filter(array -> array.length >= 2)
                 // separates by end-comment format to obtain the author's name at the zeroth index
-                .map(array -> array[1].trim().split(COMMENT_FORMATS[getCommentTypeIndex(line)][1]))
+                .map(array -> COMMENT_FORMATS[getCommentTypeIndex(line)][1] != null
+                        ? array[1].trim().split(COMMENT_FORMATS[getCommentTypeIndex(line)][1])
+                        : array[1].trim().split("", 1))
```
dcshzj, line 104: Sorry, can I check what is this line for? When you do a `.split("", 1)` you will get back an array of length 1, and the only member of the array is the original string. Is this intended?

If that was intentional, I think we should instead create the String array using the proper syntax instead of using this.

dcshzj (changes requested):

src/main/java/reposense/authorship/analyzer/AnnotatorAnalyzer.java
dcshzj: Sorry, I missed out your comment in the PR description. Is it possible to change to use `Arrays.copyOfRange()` instead?

sikai00: Apologies, I just tried again with the `new` keyword to create a new String array and it turns out it is possible. I must have missed out a parentheses when trying that out. Will update to use the explicit String[] constructor!

dcshzj (approved):
LGTM! Thanks for updating!

PR #1859, opened 2023-01-20
Codecov: specify threshold to be 0.3%

dcshzj:
## Proposed commit message

```
By default, the threshold for code coverage is 0%, which means that any
decrease in the code coverage will cause the status checks to fail. This
can be confusing to contributors and are often ignored as false
positives.

Let's set a threshold of 0.3% based on the recent commits made to the
repository, to reduce the rate of false positives.
```

## Other information

This change [references the project documentation on commit statuses](https://docs.codecov.com/docs/commit-status#basic-configuration), and the value of 0.3% was a judgement call based on the past commit history of this repository.


yhtMinceraft1010X (approved):
LGTM

ckcherry23 (approved):
LGTM

PR #1845, opened 2022-12-08
[#1073] Use ES6+ way for string interpolation

calvintanwj:
Fixes #1073

## Proposed commit message

```
There are several .vue files that use string concatenation instead.

This is rather messy for lines with a lot of variables concatenated
together. Opting for string interpolation improves code readability.

Let's use ES6's way of string interpolation for such files.
```

PR #1860, opened 2023-01-21
[#1865] Frontend: pin file title on scroll in the Code Panel

### LINKED ISSUE #1865: Cannot collapse a file after scrolling through it
I would like to be able to collapse a file after reading through it. This is a common feature in code review platforms like the 'Files changed' tab for GitHub PRs.

Users cannot collapse a file after scrolling through it without scrolling back up. This increases the unnecessary time spent on scrolling lengthy files.

ckcherry23:
Fixes #1865 

## Proposed commit message
```
Currently, users cannot hide/collapse a file after they finish reading
through it, without scrolling back to the top again.

Let us pin the title of the file (which includes the collapse button) so
that the user can collapse a file when/after reading through it instead
of only on the top. Let us also position the tooltips to show/hide a
file dynamically based on whether they are out of the viewport.
```

## Alternative UI Designs
There are 2 possible designs I considered to allow users to collapse files after reading through it:
1) Add another collapse button at the end of the file
[+] User can collapse file both before and after reading through a file
[-] User cannot collapse file when they are in the middle of the file
[-] Will increase the length of each file div, requiring more scrolling 

2) Pin the title of the file so that the collapse button is pinned to the top when scrolling
[+] User can collapse a file at any time of reading the file

I went with the 2nd option considering its advantages.


zhoukerrr (commented):
@ckcherry23 The idea is great and it works as described. However, it will be better to look into the possibility of adding a cypress test case for this new feature. I am not sure how this could be done for this case but do give it a try first.

Also, I think you can open an issue for this PR and assign it to yourself.

ckcherry23:
@zhoukerrr I have added a cypress test case for this feature, please take a look!

frontend/src/views/c-authorship.vue
```diff
@@ -359,10 +359,15 @@ export default {
     },
 
     toggleFileActiveProperty(file) {
+      this.scrollFileIntoView(file);
```
vvidday, line 362: UX wise, would it be better to only scroll when the element is at the top (like `getBoundingClientRect().y <= 0` or something similar)?
I think github does it this way and it'll also be similar to the original version when the user is not currently in the middle of a file.

vvidday (commented):
Really great change that will definitely improve the UX! Tested locally and the functionality & test cases work well.
Just one small detail about the scrolling behaviour

frontend/src/views/c-authorship.vue
ckcherry23: Yes, that would be better! Have made the necessary modifications

vvidday (approved):
LGTM!

HCY123902 (approved):

frontend/cypress/tests/codeView/codeView_hideFileDetails.js
```diff
@@ -66,6 +66,42 @@ describe('hide all file details', () => {
...
+    // contents of the first file should not be visible
+    cy.get('#tab-authorship .file-content ')
+        .first()
+        .should('not.be.visible');
+
+    // title of the first file should still be visible
+    cy.get('#tab-authorship .file ')
+        .first()
+        .should('be.visible');
+  });
```
zhoukerrr, line 103: @ckcherry23 Thanks for the work! Just a small suggestion. Is it possible to add the case when you scroll a bit without collapsing the file and check that the title is still visible. Another case will be scrolling back to the previous file (without reaching the top of the file) should display the title of the file.

ckcherry23: Hi @zhoukerrr, I've merged your suggested test cases in the same Cypress test. Hope this is okay?

zhoukerrr (approved):
LGTM! The feature works as expected and the test cases are robust.

yhtMinceraft1010X:
> I think this is a great idea but I think the tooltips should appear below the file header if it goes beyond the viewport. Here's an example:
> 
>  Screen.Recording.2023-02-02.at.12.08.53.mov 
> This problem already existed before, but this PR would have made this issue more prominent. Hence, I am proposing that we should fix it in this PR.
> 
> There is also a slight issue where the text is broken when the file header is collapsed ("file" becomes "fil<newline>e"), which should be fixed as well.

If I may add in my views, I think that the tooltip beyond viewport bug (as well as the broken text bug, provided its not caused by this PR) should be fixed in separate PRs so that the bugfixes don't get coupled with a new feature during the merger.

dcshzj:
For the first issue on the viewport, I proposed to fix in this PR because it was not really a bug before this change was introduced. So in that sense, this change "introduced" this bug, which I felt should be fixed in this PR.

The broken text bug can be fixed separately, or fixed within this PR, I am okay with both.

ckcherry23:
Hi @reposense/active-reviewers and @reposense/active-developers, I have made changes to make the 'show/hide file' tooltip appear below the file header if it is not in the viewport. Right now it checks if the tooltip text is not visible using the `v-observe-visibility` custom directive, and adds a CSS class based on whether the tooltip should appear below the header or not with a local method.

Other alternatives I considered was to compute the position of the tooltip only when clicked, and adjust the position accordingly using CSS classes. But I did not go with this method as I wanted to avoid the use of template refs.

However I have some questions regarding the best way to design it. Since there are tooltips on many parts of the page and not just the 'show/hide file' tooltip, I think this feature should be extended to all of them. In the current implementation, it would require adding a directive to all the tooltips (~10 types) on the dashboard, and creating a global function to change CSS styles. 

If you think this design is good enough, I can change all the tooltips to follow this dynamic positioning. If there is a better way for me to design this, please let me know. I would also propose that we make the changes for tooltip positioning in another PR.

HCY123902:
@ckcherry23 I personally agree that other tooltips need to adopt dynamic positioning as well, which can be done in a separate pull request.

There is currently an issue with the rendering though. When a file is collapsed, the tooltip will be occluded by the next file title

image

ckcherry23:
> When a file is collapsed, the tooltip will be occluded by the next file title

Yes, because the file title uses `position: sticky` to be pinned on top, it creates a new stacking context, which can be fixed by setting `position: sticky` dynamically only when needed. Will make the necessary changes.

> The tooltips of the items are placed above the title during initial rendering, which is the original way how tooltips are positioned before this pull request. However, for any file, if it is collapsed or its title is scrolled past, the tooltip will then move below the title, and it will remain that way until the next page refresh

If the tooltips should always be above by default unless obstructed by the viewport, then I will change the implementation to change CSS class only on hover, using template refs.

HCY123902:
> If the tooltips should always be above by default unless obstructed by the viewport, then I will change the implementation to change CSS class only on hover, using template refs.

I personally think this is a suitable design. What do the rest think of this

dcshzj:
Thanks for investigating @ckcherry23. Some questions and thoughts.

> Another alternative I considered was to compute the position of the tooltip only when hovered, and adjust the position accordingly using CSS classes. But I did not go with this method as I wanted to avoid the use of template refs.

Are you able to elaborate a little bit on why you want to avoid using template refs in your initial design? Just wanted to be clearer on the trade-offs we are making, because I also agree that the tooltips should appear above by default unless there is no space above.

> I would also propose that we make the changes for tooltip positioning in another PR.

Yup, let's apply this tooltip positioning logic to the other tooltips in a separate PR, as it would have deviated significantly from the original intent of this PR.

One final point: I think this turns out to be more difficult than I thought (I had the impression that there was some easy-to-use API to detect if an element is/is not in the viewport). If it is possible to implement the functionality using the template refs design you suggested, let's do that and aim to get this PR merged next week. Otherwise, let's file it as an issue and tackle it in a separate PR.

ckcherry23:
> Are you able to elaborate a little bit on why you want to avoid using template refs in your initial design? Just wanted to be clearer on the trade-offs we are making.

I believe the trade-off is just no reusability, as we need to have separate functions for each Vue component that contains a tooltip to detect whether it is within the viewport on hover using the ref value. This is because refs can only be accessed within the same component (afaik).

Whereas the `vue-observe-visibility` directive can be reused for all tooltips and can trigger a Vue mixin to change CSS. I can will try creating a new directive just for the tooltips. 

> because I also agree that the tooltips should appear above by default unless there is no space above.

Will make changes to follow this expected behavior!

HCY123902 (approved):
Great work. The changes for the other tooltips can be put in a separate pull request

frontend/src/views/c-authorship.vue
```diff
@@ -103,24 +103,20 @@
       .file(v-bind:ref="file.path")
         .title
           span.caret(v-on:click="toggleFileActiveProperty(file)")
-            .tooltip(v-show="file.active")
+            .tooltip(
+              v-show="file.active"
+              v-on:mouseover="onTooltipHover(file.path + 'hide-file-tooltip')",
```
HCY123902, line 108: Just a minor suggestion. This needs to adopt ES6+ way of string interpolation as is shown [here](https://github.com/reposense/RepoSense/pull/1845/files)

dcshzj (approved):
Approving this for now. There's still an issue with overlapping as shown:

https://user-images.githubusercontent.com/27919917/221465545-375b0ad8-7202-4ce2-8181-07be21bb8df7.mov

We can tackle this in another PR.

PR #1866, opened 2023-01-23
[#1340] Deprecate use of window.toggleNext in zoom panel

### LINKED ISSUE #1340: Deprecate use of window.toggleNext in zoom panel

vvidday:
Fixes #1340

## Proposed commit message

```
Currently, the zoom panel uses the window.toggleNext function to toggle
the display of commit message bodies. The window.toggleNext function
directly modifies the DOM instead of using Vue directives, which may
interfere with the functionality provided by Vue.

Let's deprecate the use of window.toggleNext by refactoring c-zoom.vue
to use the v-show directive to show/hide the body of commit messages.
This will make the code more consistent, and eliminates the need to
directly mutate the DOM.
```

### Design decision (Vuex vs local state)
The implementation in this PR involves adding a new (optional) field `isOpen` in the `commitResult` object within the `tabZoomInfo` object of the Vuex store, and using it together with `v-show` to control the toggling of commit message bodies.

Functionally, it is the same as the previous version, namely:
- Toggle persists after sorting
- "Show all commit messages" and "Hide all commit messages" are only applied to the displayed commits
-  Toggles are reset when switching between users

This implementation differs slightly from the one used in `c-authorship`, which stores the `isOpen/active` field in the local state instead of in Vuex, so it doesn't mutate the Vuex state on toggle. However, the difference is that in `c-authorship`, the state for the files is originally local, whereas for `c-zoom`, the state for the user/commit info originally comes from Vuex. Also, `c-authorship` updates the view a bit differently with the async method `updateSelectedFiles` (with the loading screen).

I made an alternative implementation [here](https://github.com/vvidday/RepoSense/tree/1340-local-state-alternative) that is more similar to the implementation in `c-authorship` (uses local state instead of Vuex). Although this approach will make the code more consistent, it requires making a deep copy of the data in the Vuex store to local state, and it has to be done after the Vuex store loads, which triggers a re-render.

I'm opening the PR as a draft for now to discuss which will be the best approach.


frontend/src/views/c-zoom.vue
```diff
@@ -241,21 +244,12 @@ export default {
...
     info() {
       const newData = {
-        expandedCommitMessagesCount: this.totalCommitMessageBodyCount,
         ...zoomInitialState(),
       };
       Object.assign(this.$data, newData);
       this.initiate();
       this.setInfoHash();
     },
-    selectedFileTypes: {
```
HCY123902, line 251: Is it safe to remove this watcher here

frontend/src/components/c-summary-charts.vue
```diff
@@ -144,7 +146,7 @@
           span.tooltip-text.right-aligned {{ getPercentileExplanation(j) }}
 
       .summary-chart__ramp(
-        v-on:click="openTabZoomSubrange(user, $event, isGroupMerged(getGroupName(repo)))"
+        v-on:click="openTabZoomSubrange(JSON.parse(JSON.stringify(user))\, $event, isGroupMerged(getGroupName(repo)))"
```
HCY123902, line 149: I think this line contains a redundant back slash

HCY123902 (commented):
I left several initial comments for the implementation. Regarding the design alternative, is it possible to move the toggle states of `c-authorship` to Vuex store as well to make state management more centralized

frontend/src/components/c-summary-charts.vue
vvidday: Apologies for missing this, have rectified the issue

vvidday:
Sorry for the late response. I've rectified the issues mentioned above and will look into moving the toggle states of `c-authorship` into Vuex - will update again.

vvidday:
> I left several initial comments for the implementation. Regarding the design alternative, is it possible to move the toggle states of `c-authorship` to Vuex store as well to make state management more centralized

The latest commits moved the `files` object from `c-authorship`'s local state to vuex, similar to how the `zUser.commits` object for `c-zoom` is stored in vuex. Other components can now access these objects from the vuex store to see the current toggle states.

I've also added regression test cases for both authorship and zoom views to test that the toggles persist after sorting/ordering to ensure that this functionality remains in any future changes as well.

frontend/cypress/tests/zoomView/zoomView_hideCommitMessages.js
```diff
@@ -53,4 +53,160 @@ describe('hide all commit messages ', () => {
     cy.get('#tab-zoom .commit-message .body')
         .should('be.visible');
   });
+
+  it('check correct display of show all and hide all commit messages', () => {
```
HCY123902, line 57: May I suggest that you split the hide/hide into 2 separate cases for consistency reason

HCY123902:
> > I left several initial comments for the implementation. Regarding the design alternative, is it possible to move the toggle states of `c-authorship` to Vuex store as well to make state management more centralized
> 
> The latest commits moved the `files` object from `c-authorship`'s local state to vuex, similar to how the `zUser.commits` object for `c-zoom` is stored in vuex. Other components can now access these objects from the vuex store to see the current toggle states.
> 
> I've also added regression test cases for both authorship and zoom views to test that the toggles persist after sorting/ordering to ensure that this functionality remains in any future changes as well.

You can replace the Draft label when the changes are ready

frontend/cypress/tests/zoomView/zoomView_hideCommitMessages.js
vvidday: Got it, I've split them into 3 cases (only 'hide all', only 'show all' and both 'show all' & 'hide all') - I hope that's okay?

frontend/cypress/tests/zoomView/zoomView_hideCommitMessages.js
```diff
@@ -53,4 +53,174 @@ describe('hide all commit messages ', () => {
...
+        .click();
+
+    cy.get('#tab-zoom .commit-message', { timeout: 90000 })
+        .should('be.visible');
+
+    // should only display 'hide all commit messages'
+    cy.get('#tab-zoom .toolbar--multiline')
+        .children()
+        .should('have.length', 1);
+
```
HCY123902, line 71: I think you can also try clicking it and verify that it indeed hides the commits as intended. Similar for the show case

vvidday: Oh I thought it was already covered in the first test case in line 2? Or would it be better to combine the test cases and check the text & functionality together?

HCY123902: Thanks for noting that. You can ignore the previous comment

HCY123902 (approved):

zhoukerrr (approved):
LGTM. I do agree that using vuex is a better approach here.

vvidday:
Resolved the conflict that resulted from #1867, CI is currently failing because of https://github.com/reposense/testrepo-Alpha/pull/15, should pass after #1875 is merged into this branch

frontend/src/views/c-authorship.vue
```diff
@@ -5,9 +5,9 @@
   .toolbar--multiline
     a(
       v-if="activeFilesCount < selectedFiles.length",
-      v-on:click="expandAll()"
+      v-on:click="toggleAllFileActiveProperty(true)"
     ) show all file details
-    a(v-if="activeFilesCount > 0", v-on:click="collapseAll()") hide all file details
+    a(v-if="activeFilesCount > 0", v-on:click="toggleAllFileActiveProperty(False)") hide all file details
```
dcshzj, line 10: Should this be a lowercase `false`?

frontend/src/components/c-summary-charts.vue
```diff
@@ -66,7 +66,8 @@
             span.tooltip-text Click to view group's code
         a(
           onclick="deactivateAllOverlays()",
-          v-on:click="openTabZoom(repo[0], filterSinceDate, filterUntilDate, isGroupMerged(getGroupName(repo)))"
+          v-on:click="openTabZoom(JSON.parse(JSON.stringify(repo[0])), filterSinceDate, filterUntilDate,\
```
dcshzj, line 69: I think it was not addressed earlier, but any reason why you decided to stringify and parse again?

frontend/cypress/tests/zoomView/zoomView_hideCommitMessages.js
```diff
@@ -53,4 +53,174 @@ describe('hide all commit messages ', () => {
...
+    cy.get('#tab-zoom .toolbar--multiline > a')
+        .should('have.text', 'hide all commit messages');
```
dcshzj, line 73: I think we should also check that the "show all commit messages" text is hidden. I think we might be able to do that by checking that only 1 item matches the selector.

frontend/cypress/tests/zoomView/zoomView_hideCommitMessages.js
```diff
@@ -53,4 +53,174 @@ describe('hide all commit messages ', () => {
...
+    cy.get('#tab-zoom .toolbar--multiline > a')
+        .click();
+
+    // should only display 'show all commit messages'
+    cy.get('#tab-zoom .toolbar--multiline')
+        .children()
+        .should('have.length', 1);
+
+    cy.get('#tab-zoom .toolbar--multiline > a')
+        .should('have.text', 'show all commit messages');
```
dcshzj, line 96: Same for this as well but for "hide all commit messages".

dcshzj (changes requested):

frontend/src/components/c-summary-charts.vue
vvidday: This was to be consistent with the previous functionality of toggles being reset when switching between users (if we pass without deep copying, toggling the visibility in the zoom panel will change the commits in the underlying `user` object, and so will persist after switching)

frontend/cypress/tests/zoomView/zoomView_hideCommitMessages.js
vvidday: I think this should be addressed in the previous line(s)? 
```
    cy.get('#tab-zoom .toolbar--multiline')
        .children()
        .should('have.length', 1);
```
is checking that there's only one element, and the above is checking that the element has the correct text

frontend/src/views/c-authorship.vue
vvidday: Yes it should, thanks for pointing this out! Will fix this

frontend/cypress/tests/zoomView/zoomView_hideCommitMessages.js
dcshzj: Right, yes. I was originally thinking of checking the text itself, then edited my comment to suggest about counting, which I didn't realize had already been done. Thanks!

frontend/src/components/c-summary-charts.vue
dcshzj: I see, would it be possible to move the stringify and parse into the `openTabZoom` function and add a comment stating its purpose? I think it was not immediately clear from reading the code.

vvidday: Just pushed a commit for this!

dcshzj (approved):
LGTM!

PR #1867, opened 2023-01-23
[#968] Enforce Vue option order

### LINKED ISSUE #968: Enforce Vue option order
As pointed out by @jamessspanggg , it is better to follow the option order style recommended by Vue in our JS file, which may increase the readability of the code a lot.

See https://vuejs.org/v2/style-guide/index.html#Component-instance-options-order-recommended for more details.

vvidday:
Fixes #968

## Proposed commit message

```
Currently, the order of component options is not enforced, resulting in
inconsistent ordering across files. Although there is a rule in the
ESLint Vue plugin that enforces this, it is currently being omitted.

Let's enable the order-in-components rule to enforce a consistent order
of component options according to the recommended order in the Vue style
guide.

This will improve the consistency and readability of the code.
```

## Other information

In addition to the option order rule, there's the [`require-prop-types`](https://eslint.vuejs.org/rules/require-prop-types.html) rule that is still being ignored. It was however resolved in #1852 and can now be removed to enforce prop types in the future - please advise if I should include it here or open another PR to remove it. Thanks!


ckcherry23 (approved):
LGTM!

HCY123902 (approved):
Thanks for the contribution. I think you should remove `require-prop-types` in a separate pull request

yhtMinceraft1010X (approved):
LGTM

PR #1863, opened 2023-01-22
[#1791] Switch to package-lock.json and npm ci

MarcusTXK:
Fixes #1791 
## Proposed commit message

```
Issue 1784 demonstrates how changes to upstream technologies and
dependencies can break CI.  Although it was more likely to be due to
Chrome updating to version 102, there is also a possibility that
updating a dependency during a CI run might cause a breaking change. 

Let's consider switching to using package-lock.json and npm ci to ensure
that dependency versions are fixed when running CI. This allows for CI
builds to be repeatable and consistent.
```

## Other information

Currently, there are significant overlaps between the 2 `.gitignore` files in the root folder as well as the frontend. I suggest we consolidate it into a single .gitignore file for easier maintainability. This is done in PR #1869.

PR #1869, opened 2023-01-24
[#1868] Merge .gitignore files into single root file

### LINKED ISSUE #1868: Merge .gitignore files for easier maintainability

A centralised .gitignore file will allow for easier maintainability as it is easy to find exclusions and reduces duplicated files in gitignore.
 
When working on Issue #1791, multiple gitignore files had to be updated in order to make a change due to significant overlaps in  2 .gitignore files in the root folder as well as the frontend.

Merge both .gitignore files into a single .gitignore file and have a single source of truth.

MarcusTXK:
Fixes #1868 

## Proposed commit message

```
A centralised .gitignore file will allow for easier maintainability as
it is easy to find exclusions and reduces duplicated files in
gitignore.

Let's merge both .gitignore files into a single .gitignore file and
have a single source of truth.
```


chan-j-d (approved):
LGTM!

.gitignore
```diff
@@ -1,22 +1,23 @@
...
 repos/
 logs/
 out/
-ft_temp/
-*.zip
 
-build/
-frontend/build
 node_modules
-package-lock.json
```
yhtMinceraft1010X, line 18: You may want to leave the `package-lock.json` line in for now since removing the frontend package-lock.json should not be in this PR's scope.

yhtMinceraft1010X (changes requested):
There is one more .gitignore file under the `frontend/cypress` directory

.gitignore
MarcusTXK: Alright, I will re-add this and add an exception to exclude `frontend/cypress/package-lock.json`

MarcusTXK: Ideally, this is merged before #1863, so that in #1863 we can remove the exception of `!frontend/cypress/package-lock.json`

yhtMinceraft1010X (approved):
LGTM

PR #1870, opened 2023-01-26
[#1848] Add instruction for debugging argument

### LINKED ISSUE #1848: Debugging with Arguments
**What problem(s) did you run into that caused you to request additional documentation? What, if any, existing documentation relates to this proposal?**

This issue is specific to debugging with Intellij.

Relevant sections of the Documentation:
[Learning the Basics, Backend, Step 2](https://reposense.org/dg/learningBasics.html)

While learning the architecture and playing around with the code base, I realized that debugging without any arguments is fairly useless. There is then a need to provide additional arguments in the run configuration for Intellij. However, due to the way Gradle run works, it is not possible to supply the argument flags directly as one would normally when debugging with args:
image
This leads to the following error:
image

The correct way to supply args as stated [here](https://stackoverflow.com/questions/52267489/how-do-i-use-gradle-bootrun-with-args-in-intellij) is this:
image

I feel this is rather counterintuitive and should really be mentioned.

However, the current documentation makes no mention of this, and simply redirects the developer to an introduction to debugging article for Intellij.

**What would you like added or modified in the documentation?**

Just a quick tooltip with a screenshot telling the user how to provide arguments would help a lot with figuring out how to perform debugging. 

Another possible solution is to add a debug entry point inside the code base so that the developer can provide the command line arguments there:

```
DebugEntry{
        String cliArg = "--since 01/01/2017";
        String[] args = cliArg.split();
        RepoSense.main(args);
    }
}
```

HCY123902:
Fixes #1848 

## Proposed commit message
```
New contributors may want to use IDEA debugger to walk
through the backend. Currently, the learning the basics section 
does not provide instructions to supply debugging arguments.

Let's add an instruction to supply debugging argument.
```


ckcherry23 (approved):
LGTM

zhoukerrr (approved):
LGTM

dcshzj (approved):

Issue #1042, opened 2020-02-02
Include deletion in contribution


eugenepeh:
RepoSense should include all form of work done (including deletion). This is because deletion can be removing redundant codes, logic or refactoring etc. It is still a form of contribution as it may be introducing better code quality, project management and/or making product more efficient.

However, at the same time, we can't simply add the deletion into the the teams' contribution because it would be difficult to aggregate the deleted codes in a readable form to the users/reviewers for them to validate the results. 

Hence I propose:
* include deletion into team's contribution
* add a selection in tool bar to show just either `Addition`, `Deletion`, `Both`
* making the default to show / sort the results in `Addition`

Issue #389, opened 2018-10-29
Should we include merge commit in the ramp chart?


eugenepeh:
Some merge commits may contain merge changes (lines change), perhaps we should include them in the ramp chart?

damithc:
What does the current RS do with merge+squash commits? treat as normal commit?

For merge commits, we can show them but in a different way? It will be misleading to show the commit ramp as usual as it makes the same change appear twice in the ramp chart.

eugenepeh:
> What does the current RS do with merge+squash commits? treat as normal commit?

Yes.

>For merge commits, we can show them but in a different way? It will be misleading to show the commit ramp as usual as it makes the same change appear twice in the ramp chart.

The actual problem is for example we do this

![image](https://user-images.githubusercontent.com/19277206/47637658-64a6d580-db97-11e8-98f1-6c3bfe86b515.png)

![image](https://user-images.githubusercontent.com/19277206/47637681-77b9a580-db97-11e8-8d0e-5e39947986f3.png)

There's a commit on the `merge-conflict` branch which has helped to resolve the conflict (make some changes and etc) but wasn't reflected in the git log when we set the flag `--no-merges`.

![image](https://user-images.githubusercontent.com/19277206/47637804-ce26e400-db97-11e8-9881-d7e72a152f4a.png)

![image](https://user-images.githubusercontent.com/19277206/47637921-10502580-db98-11e8-96a6-4663eb16b0d3.png)

So what we get will be all 3 commits happily contributed 3 lines each. But in the actual fact, the merge commit has removed one line, left 8 lines, in the process and this was not reflected in our commit analysis.

I think we need someways to track the changes made in the merge commit.

eugenepeh:
More info:

![image](https://user-images.githubusercontent.com/19277206/47639096-6ffc0000-db9b-11e8-9e60-21fc6d8d8076.png)

The final analysis is still correct though, just that no ramp shows up for my GitHub account even though it made the effort to resolve the merge conflict.

damithc:
> The final analysis is still correct though, just that no ramp shows up for my GitHub account even though it made the effort to resolve the merge conflict.

It is fair to show a ramp for the merge commit, but what about the size of the ramp? Can we make it reflect the size of the conflict resolution?

eugenepeh:
>It is fair to show a ramp for the merge commit, but what about the size of the ramp? Can we make it reflect the size of the conflict resolution?

[...]

PR #1882, opened 2023-02-02
[#1877] Show merge commits in the report

### LINKED ISSUE #1877: Show merge commits in the report
Current: merge commits are totally ignored in the report visualizations

Suggested: include merge commits in the report

* As most of changes in the merge commits are already reflected in other commits, perhaps we show it as a very thin ramp in the ramp chart?

Please see these for more context: #1876 #389

ckcherry23:
Fixes #1877

## Proposed commit message

```
Currently, merge commits are completely ignored during report generation
and analysis. This misrepresents the number of commits the user has made
and misses out on important information included in the merge commits,
such as tags.

Let's include merge commits in the report so that important information
in such commits is not missed out.
```

## Implementation details
* To include merge commits in the reports generated by the backend, I have modified the `git log --no-merges` to be `git log --full-history`.
* The git log command now also returns the parent hashes of the commit, and I determine whether it is an `isMergeCommit` based on whether it has more than 1 parent hash. 
* The **Ramp Chart** shows merge commits now as very thin ramps. 
* The **Commits Panel** shows all commits including merge commits when all file types are selected. When only some file types are selected, *empty* merge commits are not shown (since they do not come under any file category). We could extend this to explicitly show/hide merge commits from the commits panel.

## Other changes made:
- [x] Backend unit tests: Created test branch `1882-CommitInfoAnalyzerTest-analyzeCommits_mergeCommits_success` in [testrepo-Alpha](https://github.com/reposense/testrepo-Alpha)
- [x] Backend system tests: Regenerated all `commits.json` files
- [x] Frontend cypress tests: Made 2 commits with Eugene's credentials in the `cypress` test branch


damithc:
@ckcherry23 thanks for taking this up. When ready, also post a screenshot.

yhtMinceraft1010X:
> * The **Ramp Chart** shows merge commits now as very thin ramps. Ramp sizes are now calculated based on `insertions + deletions` instead of just `insertions`.
> * File changes are now shown as `+60 -12 lines` instead of `60 lines`.

With regards to these points, I'm of the opinion that changes to the ramp size calculation and file changes display should be done in a separate PR. What does everyone else @reposense/active-reviewers @reposense/devs think?

ckcherry23:
> @ckcherry23 thanks for taking this up. When ready, also post a screenshot.

Here's a screenshot of the behavior now. Will update again based on my peer reviews.
Screenshot 2023-02-02 at 5 04 38 PM

This is the repo that was used to report the bug. The light blue ramps are for `2023-01-29`, and have 3 merge commit ramps between 2 normal ramps. The Commits Panel have the merge commits and the previously missing tags as well. 

> With regards to these points, I'm of the opinion that changes to the ramp size calculation and file changes display should be done in a separate PR (which I think can solve #1042), and once that is merged, continue working on this PR.

Okay, I think I will make the changes to ramp size calculation and file changes separately to limit the scope of this PR.

damithc:
> With regards to these points, I'm of the opinion that changes to the ramp size calculation and file changes display should be done in a separate PR (which I think can solve #1042), and once that is merged, continue working on this PR.

Yes, this requires more deliberations.

frontend/src/components/c-ramp.vue
```diff
@@ -2,7 +2,7 @@
 .ramp
   template(v-if="tframe === 'commit'")
     template(v-for="(slice, j) in user.commits")
-      template(v-for="(commit, k) in slice.commitResults.filter(commitResult => commitResult.insertions > 0)")
+      template(v-for="(commit, k) in slice.commitResults")
```
HCY123902, line 5: Can I ask whether it is safe to remove this check and include empty commits? Another question here is whether merge commits can be designed to be empty commits. The reason for this question is, certain merge commits may introduce additional changes when resolving conflicts, which I think is reasonable to be reflected on the dashboard. In this case, merge commits should still have the actual number of additions and deletions and these should get reflected in the tooltip and zoom panel.

frontend/src/components/c-ramp.vue
```diff
@@ -21,7 +21,7 @@
   template(v-else)
     a.ramp__slice(
       draggable="false",
-      v-for="(slice, j) in user.commits.filter(commit => commit.insertions > 0)",
```
HCY123902, line 24: Similar question for here

frontend/src/views/c-summary.vue
```diff
@@ -703,7 +703,7 @@ export default {
         const filteredFileTypes = this.getFilteredFileTypes(result);
         this.updateCommitResultWithFileTypes(result, filteredFileTypes);
         return result;
-      }).filter((result) => Object.values(result.fileTypesAndContributionMap).length > 0);
```
HCY123902, line 706: Similar question for the logic here

frontend/src/components/c-ramp.vue
HCY123902, line 5: Regarding this, I understand that the change here is intended to include merge commits. However, I am thinking about whether it is suitable to design merge commits to be empty. As the example given in #389 suggests, there can be certain merge requests that introduce additional changes when resolving conflicts. This means it may be necessary to show the number of additions and deletions for merge commits on the dashboard as well for consistency reasons.

ckcherry23: Sorry, I'm not sure what you mean by "designing merge commits to be empty". 

Since even empty merge commits can include information like tags, I have included them in the chart. I believe I am displaying the number of additions and deletions, irrespective of the type of commit, for both the ramp chart and the commits panel. 

For the highlighted line in the code, are you suggesting we filter using: 
`.filter(commitResult => this.getContributions(commitResult) > 0 || commitResult.isMergeCommit)`?

HCY123902: Oh I was under the impression that merge commits will be empty cos you mentioned above that these commits will only not be shown if any file type is excluded. If fileTypeAndContributionMap and number of additions, deletions work the same way as other commits, then you can just ignore my comment. Thanks

ckcherry23: Oh, what I meant to say is that *empty* merge commits will not be shown when any file type is excluded, since they do not come under any file type category. 

When all file types are selected, I bypass the filtering of commits by the selected file types and display all commits, so that empty merge commits (that do not come under any file type) are included.

HCY123902: Thanks for the clarification. Actually, what caused my confusion is that while the merge commit in https://github.com/Jnjy/ip/commit/d847cd09cfe4c4591e960d77e5b9b6bc70d97fce has a positive number of additions and deletions, the screenshot above shows otherwise ...... If that is intended, please go ahead

HCY123902: According to the discussion in the lecture, let's add the following changes to the commits panel

1. Add a Merge label next to each commit title in the commits panel
2. In addition to the numerical line counts, add a contribution bar for each commit in the commits panel, similar to that in the chart panel, to indicate the number of additions and deletions. Optionally, color the addition bar to be green and the deletion bar to be red.

chan-j-d (commented):
The backend changes look pretty good. I think it might be necessary to add some new unit tests to check if merge commits are correctly identified

frontend/src/components/c-ramp.vue
ckcherry23: I've added the `code-merge` icon beside the titles of merge commits.

frontend/src/views/c-summary.vue
```diff
@@ -708,7 +708,7 @@ export default {
...
```
HCY123902, line 711: Great work. Can I ask why this filter is removed? According to the behavior of the commits panel, the merge commits without any addition or deletion should only be shown when every file type is checked. If certain file types are unchecked, then these need to be hidden.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -30,7 +30,7 @@ Call-to-action statement describing what you are doing in this PR.
 
 ### Other information
 <!--
-    Are there other relevant information, such as special testing instructions, 
+    Are there other relevant information, such as special testing instructions,
```
ckcherry23, line 33: **To reviewers:** 
Changes made to files under `.github` and `docs` are only formatting changes and can be ignored.

frontend/src/views/c-summary.vue
ckcherry23: Thanks! I have changed the behavior of the Ramp Chart to be similar to the Commits Panel.

src/main/java/reposense/commits/model/CommitResult.java
```diff
@@ -63,6 +66,10 @@ public String getHash() {
         return hash;
     }
 
+    public Boolean getIsMergeCommit() {
+        return isMergeCommit;
+    }
+
```
ckcherry23, line 72: The `codecov/patch` test is failing because of a new getter method that is unused. This might require changing the target number in `codecov/patch` CI or I could get rid of the method.

HCY123902: yeah sure. You can remove this method

HCY123902:
The current changes look great to me. You can request a separate review from me as soon as you complete the cypress changes.

I only have 1 more doubt. This is again related to the case where a merge commit containing changes from conflict resolution.

According to this [post](https://stackoverflow.com/questions/57333971/git-log-numstat-has-weird-data) and this [post](https://stackoverflow.com/questions/37036277/git-show-vs-git-log-numstat-output-difference), `git log` does not actually diff a merge commit against either of its parent by default, because it does not know which parent to compare with. To actually get the number of additions and deletions, there has to be an additional [option](https://git-scm.com/docs/git-log#Documentation/git-log.txt--c) such as `-c`, `--cc`, or `-m` to either combine or pick the difference. 

According to the posts, `-c` and `--cc` will only compute differences for a merge commit that is different from both of its parents, meaning that the merge commit contains additional changes for conflict resolution, and other merge commits will again be skipped. The `-m` option compares the commit with each of its parents separately.

I am not sure about the current behavior of the wrapper class in `GitLog.java`. I do not think what I mentioned above is significant either. This is just to open the room for potential future discussion

HCY123902 (changes requested):
It seems that there are 2 cypress cases regarding zoom view that did not pass. Can you take a look

HCY123902 (approved):

zhoukerrr (approved):
I am only reviewing the frontend side of this PR. LGTM! The cypress tests looks good to me as well.

chan-j-d (approved):
LGTM! Unit tests look good and the frontend changes seem to work well. Great work!

PR #1883, opened 2023-02-02
[#1042] Include deletions in contributions

### LINKED ISSUE #1042: Include deletion in contribution

ckcherry23:
Fixes #1042

Part of #1877
To be merged before #1882 

## Proposed commit message
```
Currently, the ramp chart only reflects number of insertions. Hence, 
commits that only delete code are omitted from the ramp chart.

Let us modify the ramp chart to include thin ramps to represent commits
that only delete code. Let us also include the number of lines deleted
when displaying lines changed on the Ramp Chart tooltip and on the
Commits Panel.
```

## Other information
- Based on the discussions here, the ramp sizes are calculated based on the number of `insertions` (unchanged). In addition to this, if there are commits with 0 insertions but some number of deletions, a thin ramp is added to the ramp chart to represent such commits.  
- For the lines changed, which are displayed in a tooltip on hovering a ramp, or on the Commits Panel along with each commit, the negative lines changed have been also added: `+60 -12 lines` instead of `60 lines`.
- The Code Panel remains the same, as it shows authorship, rather than the contributions of users.


damithc:
What happens when a user edits a line? Is it counted as a  2-line contribution?

damithc:
> > What happens when a user edits a line? Is it counted as a  2-line contribution?
> 
> Yes, since we are not checking which lines are edited to avoid this, edits are counted as 2-line contributions.

Hmm... feels like a misrepresentation to give double weight to a line edit, compared to a new line. I think that's why we initially limited to showing line additions only. Besides our code view only shows remaining lines, not deleted lines.

damithc:
An alternative:
Currently, any commit that edits or adds lines will show up as a ramp and the ramp size reflects the added/edited line count. 
In addition, we can show commits that only deletes code as a thin ramp (similar to merge commits) so that they are not omitted altogether. Plus we can show LoC for for additions and deletions when hovering over (in the tool tip).
What do you guys think?

HCY123902:
> An alternative: Currently, any commit that edits or adds lines will show up as a ramp and the ramp size reflects the added/edited line count. In addition, we can show commits that only deletes code as a thin ramp (similar to merge commits) so that they are not omitted altogether. Plus we can show LoC for for additions and deletions when hovering over (in the tool tip). What do you guys think?

Currently, `commits.json` only includes added and deleted lines from the output of `git log --numstat`. I think git does not seem to provide native support to calculate edited lines, which means we need to either adopt certain heuristics for word matching with the help of `git --word-diff` or seek help from external plugins such as `diffstat`. Here is a [discussion thread](https://stackoverflow.com/questions/9933325/is-there-a-way-of-having-git-show-lines-added-lines-changed-and-lines-removed) that I found to be relevant to this

damithc:
> Another alternative we could consider is calculating sizes using `max(insertions, deletions)` to reflect the higher number of lines changed.

In some cases this can be misrepresentative as well, A commit that deletes a buggy method and adds a smaller and correct method should get full credit for the added method.

> > An alternative: Currently, any commit that edits or adds lines will show up as a ramp and the ramp size reflects the added/edited line count. In addition, we can show commits that only deletes code as a thin ramp (similar to merge commits) so that they are not omitted altogether. Plus we can show LoC for for additions and deletions when hovering over (in the tool tip). What do you guys think?
> 
> Currently, `commits.json` only includes added and deleted lines from the output of `git log --numstat`. I think git does not seem to provide native support to calculate edited lines, which means we need to either adopt certain heuristics for word matching with the help of `git --word-diff` or seek help from external plugins such as [diffstat](https://invisible-island.net/diffstat/manpage/diffstat.html). Here is a [discussion thread](https://stackoverflow.com/questions/9933325/is-there-a-way-of-having-git-show-lines-added-lines-changed-and-lines-removed) that I found to be relevant to this

Yes, there is no native Git support for detecting edited lines. Git gives one edited line as one added line + one deleted line -- so, simply by counting added lines only (as we do now) we account for edited lines as well. The only case we are missing  in the ramp chart now is commits that delete lines only. Hence the proposal to add them as thin ramps.

damithc:
> Another alternative we could consider is calculating sizes using `max(insertions, deletions)` to reflect the higher number of lines changed. Since the actual weight that we use for displaying the ramp sizes is not made obvious to the user, I think this could be a good way of estimating commit contributions, without confusing the user on how they are calculated.

I'm more inclined to let the ramp chart focus on lines that remain, so that the ramp chart and the code view don't give drastically differing views. While deleting code is also a contribution, simply deleting a chunk of code is not a significant contribution. Usually such deletions come with a replacement code that could be slightly smaller than the deleted chunk, and that addition will still show up as a significant ramp in the ramp chart.

ckcherry23:
I've updated the ramp chart to include thin ramps for commits that only delete code. Here's a screenshot:

(The thin light blue ramp is a deletes commit)
Screenshot 2023-02-03 at 4 08 25 PM

zhoukerrr:
I think that the ramp could remain the same as to not confuse the users. A LoC deleted could be included next to the file titles. Somewhat similar to the LoC we see here.
Screenshot 2023-02-04 at 17 13 51

We could even include all the lines that were removed by the author and highlight in red, similar to how Github does it. That way we can also visually see the exact lines that are removed.

However, all these might require quite a bit of work to achieve. What do the rest think about this?

ckcherry23:
> We could even include all the lines that were removed by the author and highlight in red, similar to how Github does it. That way we can also visually see the exact lines that are removed.

I am not sure if it is possible to show the lines deleted (and the deleted LoC), since there isn't exactly a previous version of the file for us to compare with, which is not the case for commits on GitHub. 

Would you suggest that we compare the versions from the `since` and `until` dates? This may also change the purpose of the code panel from showing 'authorship' to showing 'changes made'.

> I think that the ramp could remain the same as to not confuse the users. 

Since commits with only deletions are rare, I do not think this version would confuse the users, as the rest of the behavior is the same. We could also show the deleted code ramps in red to visually identify them.

HCY123902:
> I think that the ramp could remain the same as to not confuse the users. A LoC deleted could be included next to the file titles. Somewhat similar to the LoC we see here. Screenshot 2023-02-04 at 17 13 51
> 
> We could even include all the lines that were removed by the author and highlight in red, similar to how Github does it. That way we can also visually see the exact lines that are removed.
> 
> However, all these might require quite a bit of work to achieve. What do the rest think about this?

Regarding the content of deleted lines, I think it is already easily accessible for most commits that have links to GitHub pages showing the deleted lines

zhoukerrr:
> > We could even include all the lines that were removed by the author and highlight in red, similar to how Github does it. That way we can also visually see the exact lines that are removed.
> 
> I am not sure if it is possible to show the lines deleted (and the deleted LoC), since there isn't exactly a previous version of the file for us to compare with, which is not the case for commits on GitHub.
> 
> Would you suggest that we compare the versions from the `since` and `until` dates? This may also change the purpose of the code panel from showing 'authorship' to showing 'changes made'.
> 
> > I think that the ramp could remain the same as to not confuse the users.
> 
> Since commits with only deletions are rare, I do not think this version would confuse the users, as the rest of the behavior is the same. We could also show the deleted code ramps in red to visually identify them.

Yes. I think reserving a colour (red) is a good idea.

ckcherry23:
I've updated the commits with only deleted lines to be indicated in red on the Ramp Chart.

HCY123902:
According to the discussion in the lecture today, is it possible to add a contribution bar for each commit in the commits panel, similar to that in the chart panel, to indicate the number of additions and deletions? Optionally, the addition bar can be colored green and the deletion bar can be colored red. This can go alongside the numerical line counts. What do the rest think

damithc:
> According to the discussion in the lecture today, is it possible to add a contribution bar for each commit in the commits panel, similar to that in the chart panel, to indicate the number of additions and deletions? Optionally, the addition bar can be colored green and the deletion bar can be colored red. This can go alongside the numerical line counts. What do the rest think

Yup, but this can also be a separate PR.

frontend/src/components/c-ramp.vue
```diff
@@ -211,6 +223,10 @@ export default {
     &--color4 {
       border-bottom: $height rgba(mui-color('pink'), .5) solid;
     }
+
+    &--color-deletes {
+      border-bottom: $height rgba(mui-color('red'), .5) solid;
```
HCY123902, line 228: I think the colors of certain pure deletion commit ramps are still generated from date. The ramp of commit `96d61c9` in the screenshot is colored blue here

image

ckcherry23: Yes, thanks for pointing that out!
I have fixed the issue, and now the thin red ramps appear for both `commit` granularity, and `day`/`week` granularity.

### `commit` granularity:
Screenshot 2023-02-07 at 12 12 00 AM

### `day` granularity:
Screenshot 2023-02-07 at 12 17 58 AM

HCY123902 (approved):

chan-j-d (approved):
LGTM!

zhoukerrr (approved):
LGTM!

PR #1885, opened 2023-02-04
[#1758] Consolidate OS jobs and tool versions in integration.yml

### LINKED ISSUE #1758: Refactor integration.yml

Currently, the CI for each OS is defined separately (1 for ubuntu, 1 for macos, 1 for windows) while the steps for each of them tend to be similar.

Such a setup violates the DRY principle and is difficult to maintain. In particular, when updating the `node-version`, it had to be done 4 times (1 for each OS and 1 more for cypress). The `node-version` update for cypress was actually missed and had to be added in a later PR.

There are 2 possible approaches I can think of:

**1) Merge the OS matrix, then make them share the same steps:**
```
name: ${{ matrix.os }} JDK 8
    strategy:
      matrix:
        os: [ubuntu-18.04, ubuntu-20.04, macos-10.15, macos-11, windows-2022, windows-2019]
    runs-on: ${{ matrix.os }}
    env:
      NODE_VERSION: "lts/*"

    steps:
    - name: Checkout repository
      uses: actions/checkout@v2
      with:
        fetch-depth: 0
```
The code for macos and ubuntu are similar, but the code for windows may differ. A possible solution with Gradle caching as an example would be:
```
- name: Set up Gradle caching (Ubuntu and MacOS)
      uses: actions/cache@v2
      if: ${{ !(matrix.os == 'windows-2022' || matrix.os == 'windows-2019') }}
      with:
        path: |
          ~/.gradle/caches
          ~/.gradle/wrapper
        key: ${{ matrix.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
        restore-keys: |
          ${{ matrix.os }}-gradle-

- name: Set up Gradle caching (Windows)
      uses: actions/cache@v2
      if: ${{ matrix.os == 'windows-2022' || matrix.os == 'windows-2019' }}
      with:
        path: |
          $HOME/.gradle/caches
          $HOME/.gradle/wrapper
        key: ${{ matrix.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
        restore-keys: |
          ${{ matrix.os }}-gradle-
```
This is similar to what is being done for [AB3](https://github.com/se-edu/addressbook-level3/blob/master/.github/workflows/gradle.yml)

One potential problem is that there are unique environment variables for each OS, such as `HOMEBREW_NO_AUTO_UPDATE: 1` for MacOS.

**2) Use composite actions**

As described [here](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action). This may allow us to keep the OS name definitions (and environment variables) separate while sharing the same steps. It may even be extended to the Cypress CI.

yhtMinceraft1010X:
Fixes #1758

## Proposed commit message

```
The current implementation of integration.yml is difficult to update
since steps and version numbers are duplicated across operating systems
and the cypress job.

At the same time, between Windows and Mac/Ubuntu jobs, the $HOME
directory is indicated differently.

Consolidating the OS jobs means that if a Windows build fails, all Mac
and Ubuntu builds will fail too, which makes diagnosing OS-specific 
problems difficult as the success of the build on other OSes cannot be
verified.

Let's consolidate the OS jobs and use environment variables to
keep track of tool versions across the steps, standardise the $HOME 
directory nomenclature and disable fail-fast. This makes changing the
versions easier, improves readability and makes diagnosing OS-specific
problems easier.

```

### Removal of `-e` flag for `echo` in `Verify Java version`

The `-e` flag is used to enable interpretation of backslash escapes for Ubuntu and Mac, which is not necessary even in our current implementation. However, it is not supported in Windows.

Removing the `-e` flag allows the `echo` command to be used by all 3 OSes. This also means that we do not need to define 2 separate steps for `Verify Java version` and use conditionals to check the OS.

### Operating System Environment Variables

I initially wanted to use environment variables for the operating systems. It would have been something like this:
```
  env:
    UBUNTU_SECOND_LATEST: ubuntu-20.04
    UBUNTU_LATEST: ubuntu-22.04 # Windows and Mac omitted for brevity

  mainbuild:
    name: ${{ matrix.os }} JDK 8
    strategy:
      fail-fast: false
      matrix:
        os: [ ${{ env.UBUNTU_SECOND_LATEST }}, ${{ env.UBUNTU_LATEST }} ]
    runs-on: ${{ matrix.os }}
    env:
      NODE_VERSION: "lts/*"
      HOMEBREW_NO_AUTO_UPDATE: 1 # Prevent time-consuming brew update
```

However, I noticed that `matrix.os` does not detect the use of the context, so I resorted to keeping the hardcoding of the OS. 
This is far from perfect, but I feel the refactored `integration.yml` is still easier to maintain compared to the existing setup (where steps are duplicated for Ubuntu, Mac and Windows) where there are nearly 300 lines.

### Cypress job still separate

The Cypress job is still defined separately, even though some steps are the same as that of what is now `mainbuild`.

Again, not ideal, but I feel it's still easier to maintain since the Java and Node versions (which are the most likely to be changed) are now environment variables that are shared between both jobs and placed at the top of the workflow file.

Additionally, while the OS for Cypress `runs-on` is still hardcoded, it is closer to the top of the workflow file. 

### Presence of `NODE_VERSION` environment variable

This environment variable was already present in the ubuntu version of the `mainbuild` job (see above code). I am not sure what this is for but I can remove it if needed. Not to be confused with `NODE_VER`, which is for the `Setup Node.js` step and named as such to differentiate it from `NODE_VERSION`.

### Is `Setup Node.js` necessary?

This is addressed in #1886. I intend to keep `Setup Node.js` (equivalent to developer-installed) and `RepoSense/build/node` versions of Node in this PR, to give enough flexibility to decide on the Node version source of truth later (and make the PR work for #1886 easier once that decision is made)


yhtMinceraft1010X:
Note for reviewers: Please check that tool and OS versions for conditionals, variables etc are copied over correctly.

.github/workflows/integration.yml
```diff
@@ -10,15 +10,22 @@ on:
...
+  mainbuild:
     name: ${{ matrix.os }} JDK 8
     strategy:
+      fail-fast: false
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
+        os: [ubuntu-20.04, ubuntu-22.04, macos-11, macos-12, windows-2019, windows-2022]
     runs-on: ${{ matrix.os }}
     env:
       NODE_VERSION: "lts/*"
```
chan-j-d, line 27: I'm also not too sure what it's for and I looked through the other .yml files they either set up this environment variable themselves or never reference it so I'm not too sure what it's for as well. It's present in the first version of this .yml file. Let's hope it's not some magic reference since I believe 'lts/*' is actually a way of stating 'use latest supported version'

docs/dg/devOpsGuide.md
```diff
@@ -28,9 +28,9 @@ This page documents the various components that form part of the DevOps infrastr
 
 All three types of tests in the test suite are run in a single GitHub Actions workflow called "Continuous Integration". The steps are defined in [`integration.yml`](https://github.com/reposense/RepoSense/blob/master/.github/workflows/integration.yml) and are split into four types of jobs:
 
-1. Ubuntu JDK 8 (`ubuntu`): Runs both unit tests and system tests on JDK 1.8 running on supported Ubuntu versions. This job also produces a RepoSense report and the MarkBind documentation website for previewing.
-2. macOS JDK 8 (`macos`): Runs both unit tests and system tests on JDK 1.8 running on supported macOS versions.
-3. Windows JDK 8 (`windows`): Runs both unit tests and system tests on JDK 1.8 running on supported Windows versions.
+1. Ubuntu JDK 8 (`mainbuild`): Runs both unit tests and system tests on JDK 1.8 running on supported Ubuntu versions. This job also produces a RepoSense report and the MarkBind documentation website for previewing.
+2. macOS JDK 8 (`mainbuild`): Runs both unit tests and system tests on JDK 1.8 running on supported macOS versions.
+3. Windows JDK 8 (`mainbuild`): Runs both unit tests and system tests on JDK 1.8 running on supported Windows versions.
```
chan-j-d, line 33: I wonder if these can simply be compressed to state that `mainbuild` runs unit and system tests on 6 different OS, with codecov being run on which 3 OS and additional stuff like report and documentation specifically on 20.04. Or perhaps we can even leave out the specific OS mentioned since those might get outdated and give a more general statement.

docs/dg/devOpsGuide.md
```diff
@@ -28,9 +28,9 @@ This page documents the various components that form part of the DevOps infrastr
 
 All three types of tests in the test suite are run in a single GitHub Actions workflow called "Continuous Integration". The steps are defined in [`integration.yml`](https://github.com/reposense/RepoSense/blob/master/.github/workflows/integration.yml) and are split into four types of jobs:
```
chan-j-d, line 29: Given the comment below, I think the 'three' might need to change if you change below. It should probably be 4 at this point anyway.

chan-j-d (changes requested):
Refactoring looks pretty good, just some minor comments

.github/workflows/integration.yml
yhtMinceraft1010X: I think this NODE_VERSION doesn't mean anything because the `setup-node` action takes in another env variable for the Node version instead.

docs/dg/devOpsGuide.md
yhtMinceraft1010X: True, maybe just say:
- Runs unit and system tests on supported versions of Ubuntu, MacOS and Windows. Ubuntu is also used for generating the RepoSense report and MarkBind documentation website for previewing.

.github/workflows/integration.yml
chan-j-d: yeah I noticed every where there is this lts line, there is separately a specific line that sets `node_version = (something)`.

docs/dg/devOpsGuide.md
chan-j-d: sounds good, maybe remove the brackets around Ubuntu/MacOS/Windows. Besides that, I wonder if it's worth mentioning about running the codecov here as well, unless it is documented elsewhere.

docs/dg/devOpsGuide.md
yhtMinceraft1010X: The three here refers to the unit tests, system tests and frontend tests, rather than the OS

docs/dg/devOpsGuide.md
yhtMinceraft1010X: Codecov is somewhere below in the same doc

chan-j-d (approved):
LGTM! 

Wanted to also add that I agree with the issue of the hard coded OS but I'm also not sure what an appropriate solution for it would be. I believe you can reference `ubuntu-latest` directly which would refer to 22.04 here but there isn't a way to reference second latest.

HCY123902 (approved):

PR #1890, opened 2023-02-05
[#1871] Update minor versions of Node dependencies

### LINKED ISSUE #1871: Frontend DevOps: Update Node.js dependencies

Some of our frontend dependencies have been flagged as out of date:
Screenshot_20230128_104119

How to interpret:

- Red: For current major release, the minor release and patch release are not at the latest versions. 
- Yellow: The minor release and patch release are at the latest versions, but there is a new major release available

The reason for the differentiation is that major releases tend to have breaking changes. See the [npm semver guide](https://docs.npmjs.com/about-semantic-versioning) for more information.
<br/>

**Solution for updating minor and patch releases**

Packages are updated on the computer using [`npm update`](https://docs.npmjs.com/cli/v9/commands/npm-update). Adding a `--save` flag will also update the `package.json` which can then be committed. 

However, some dependencies may conflict with each other. Care must be taken to resolve them accordingly.
<br/>

**Solution for updating major releases**

For updating the major release, [`npm install <package name>@<version>`](https://docs.npmjs.com/cli/v8/commands/npm-install) is needed. Breaking changes as a result of the major release need to be resolved.

Tip: Increment the major release number by 1 for each update. Example: 

1. Suppose you are at version 9.0.0
2. The latest major release is version 12.0.0
3. Upgrade to version 10.0.0
4. Then upgrade to version 11.0.0
5. Finally, upgrade to version 12.0.0

Rationale: Breaking changes from 1 major release tend to be easier to solve then breaking changes across 3 major releases. This also means that if version 12 has a problem that we cannot solve yet, we can rollback to version 11 rather than version 9.

yhtMinceraft1010X:
Fixes #1871. I intend to close this issue since other dependencies are only out-of-date for major versions, which require more work to update due to breaking changes (with new issues opened - one for each major dependency)

## Proposed commit message

```
Some dependencies are not at their latest minor or patch releases.

Let's update these dependencies.
```


chan-j-d (approved):
LGTM!

HCY123902 (approved):

yhtMinceraft1010X:
Due to some complications with resolving conflicts for package-lock.json (as a result of #1888), I will be closing and re-doing this PR.

PR #1893, opened 2023-02-05
Change header style for GitHub Issue and PR Templates

yhtMinceraft1010X:
## Proposed commit message

```
The current header style of bolding for issue templates is difficult to 
notice compared to the body text. The level 2 header for pull request
template is too large.

Let's update both to use level 3 headings.
```

## For PRs:

#1896 provides a preview of what level 3 headers would look like.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -8,24 +8,24 @@ labels: "c.Bug"
   Before opening a new issue, please search existing issues:  https://github.com/reposense/RepoSense/issues
 -->
 
-## Tell us about your environment
+### Tell us about your environment
 
 * **RepoSense Version:**
 * **OS and Version:**
 * **Web Browser and Version (if applicable):**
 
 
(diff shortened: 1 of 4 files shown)


dcshzj:
My personal preference is to use the 3rd level heading instead. I think the 2nd level might be too large, especially since the header is a little wordy. What do you/the others think?

HCY123902 (commented):
I agree with the third level heading as well

yhtMinceraft1010X:
> I agree with the third level heading as well

In this case, should we change to third level heading for the pull request template? I took reference from that when changing the header for the issue templates.

HCY123902:
> > I agree with the third level heading as well
> 
> In this case, should we change to third level heading for the pull request template? I took reference from that when changing the header for the issue templates.

@dcshzj Can I ask for your opinion on this

dcshzj:
No strong opinions from me, I think 3rd-level is fine too.

HCY123902 (approved):

chan-j-d (approved):
LGTM!

PR #1903, opened 2023-02-10
[#1756] Add JSON validation and migrate api.js to TypeScript

### LINKED ISSUE #1756: Add support for json validation

vvidday:
Fixes #1756
Should not be merged before #1905 

### Proposed commit message

```
Currently, the objects produced by reading from summary.json,
authorship.json and commits.json are not validated against any schema.
This results in these objects having an implicit any type, which might
lead to type errors during runtime.

Let's add support for JSON validation with the zod library and migrate
the file where the JSON files are parsed, api.js, to TypeScript. This
will increase type safety and enable the migration of other files.
```

### Organization
The files containing types are in the `src/types` folder, with the following layout:
```
└── types/
    ├── zod/
    │   ├── authorship-type.ts
    │   ├── commits-type.ts
    │   └── summary-type.ts
    ├── types.ts
    └── window.ts
```
The files in the `zod` subdirectory are [zod](https://zod.dev/) schemas that are used to validate the parsed objects from `authorship.json`, `commits.json` and `summary.json`, whereas the other two files contain regular typescript interfaces, hence the distinction. `window.ts` contains type definitions for all the global variables stored in the `window` object (`window.XXX`), and all the other 'normal' types are in `types.ts`. Please let me know if you have any suggestions for better organization!

### `api.ts`
Changes to this file were mostly related to adding types/typecasts to make it work with the typescript compiler. The only functional change is the three `xxxSchema.parse(json)` under `api.loadSummary`, `api.loadCommits` and `api.loadAuthorship`. This change will make it so that if the object doesn't match the schema (which can be either due to a modified `.json` file or a change to the schema in the backend of any of the three `.json` files), the app will throw an error instead of passing on the data.  This will make it so the error is surfaced immediately instead of being passed on to other components and silently breaking elsewhere.

The reason why I'm migrating the whole file in this PR is because adding JSON validation requires the migration of most of the file (everything in `window.api`), and the only 'extra' aspects are just the type definitions of the other global variables (`window.XXX`). If we want to separate this into 2 PRs, it will be a bit weird since we will be migrating half the file, and as far as I can tell there isn't an easy way to get typescript to ignore a section of the file (see https://github.com/Microsoft/TypeScript/issues/19573)

### Changes to eslint
The plugin added in line 9 allows proper resolving of typescript file imports, the change of rule from line 51 onwards makes the linter allow us to import typescript files like `import ... from ../file` instead of `import ... from ../file.ts` (which throws an error). The side effect is that the typescript imports in the `.vue` files must also be changed to remove the `.ts`, which is why this PR also modifies vue files. Although, if it is better, I can temporarily ignore the rule here and open a separate PR to change this.


frontend/src/utils/api.ts
```diff
@@ -30,7 +34,10 @@ window.getDateStr = function getDateStr(date) {
 window.getHexToRGB = function getHexToRGB(color) {
   // to convert color from hex code to rgb format
   const arr = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(color);
-  return arr.slice(1).map((val) => parseInt(val, 16));
+  if (arr) {
+    return arr.slice(1).map((val) => parseInt(val, 16));
+  }
+  return [];
```
ckcherry23, line 40:
```suggestion
return arr ? arr.slice(1).map((val) => parseInt(val, 16)) : [];
```

frontend/src/utils/api.ts
```diff
@@ -64,7 +71,7 @@ window.encodeHash = function encodeHash() {
 };
 
 window.decodeHash = function decodeHash() {
-  const hashParams = {};
+  const hashParams: { [key:string]: string } = {};
```
ckcherry23, line 74:
```suggestion
  const hashParams: { [key: string]: string } = {};
```

frontend/src/utils/api.ts
```diff
@@ -290,12 +303,14 @@ window.api = {
...
-        result.insertions = Object.values(result.fileTypesAndContributionMap)
+        (result as CommitResult).repoId = repoId;
+        (result as CommitResult).insertions = Object.values(result.fileTypesAndContributionMap)
             .reduce((acc, fileType) => acc + fileType.insertions, 0);
-        result.deletions = Object.values(result.fileTypesAndContributionMap)
+        (result as CommitResult).deletions = Object.values(result.fileTypesAndContributionMap)
             .reduce((acc, fileType) => acc + fileType.deletions, 0);
       });
+      // typecast to DailyCommit since we added repoId/insertions/deletions
+      commit as any as DailyCommit;
```
ckcherry23, line 313:
```suggestion
      commit as unknown as DailyCommit;
```
Is it better to typecast to `unknown` instead of `any` here?

ckcherry23 (commented):
Learned a lot from your code as I don't have much prior experience using TypeScript! Regarding file structuring, I think we could move the files `user.ts` and `segment.ts` from the `utils` directory to the `types` directory for better organization.

sikai00 (approved):
The JSON validation addition looks good to me! Although, I am of the opinion that the changes to eslint should be in a separate PR following the [good practice of limiting each PR to tackling a single issue](https://developers.google.com/blockly/guides/contribute/get-started/write_a_good_pr#keep-it-small)

frontend/src/types/types.ts
```diff
@@ -0,0 +1,47 @@
...
+export interface CommitResult extends CommitResultRaw {
+  repoId: string;
+  insertions: number;
+  deletions: number;
+}
+
+// Similar to AuthorDailyContributions, but uses the updated CommitResult with the three new fields
+export interface DailyCommit {
+  commitResults: CommitResult[];
+  date: String;
```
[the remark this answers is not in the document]
vvidday: Thanks for catching this! I realised that our current eslint configuration doesn't enforce any typescript rules for `.ts` files. I'll make a separate PR for this, since it'll require a decent amount of refactoring to remove the `any` types that eslint complains about, and I added some `any` definitions in this PR to minimize changes to the current code.

frontend/src/utils/api.ts
vvidday: I agree, just made the change. Thanks!

frontend/src/utils/api.ts
vvidday: Thanks for catching this, made the change

frontend/src/utils/api.ts
vvidday: Actually I realised that this typecast wasn't doing anything 😅 since it doesn't change the underlying type of the object, but creates a new object with the new type.
I originally wanted to typecast this to have the correct type when creating the `User` object. For now the `obj` in `new User(obj)` has the `any` type - changing this will require some refactoring which I'll do in a separate PR as mentioned above.

vvidday:
> Learned a lot from your code as I don't have much prior experience using TypeScript! Regarding file structuring, I think we could move the files `user.ts` and `segment.ts` from the `utils` directory to the `types` directory for better organization.

I agree, that would definitely help the organization/structure of the files! Since it will also require changing some `.vue` files due to imports, I think it will be better to do the move in a separate PR to limit the scope of this one.

vvidday:
> The JSON validation addition looks good to me! Although, I am of the opinion that the changes to eslint should be in a separate PR following the [good practice of limiting each PR to tackling a single issue](https://developers.google.com/blockly/guides/contribute/get-started/write_a_good_pr#keep-it-small)

Thanks, I'll open a separate PR for the eslint changes. For now, I added two lines in `.eslintrc.json` to ignore the import errors, to let CI continue with the cypress tests/deployments. I'll remove those lines once that PR is merged and I can merge the changes into this branch.

frontend/src/utils/api.ts
```diff
@@ -239,13 +244,15 @@ window.api = {
...
+    const json = await this.loadJSON(`${folderName}/commits.json`);
+    const commits = commitsSchema.parse(json);
+
+    const res: User[] = [];
     const repo = window.REPOS[repoName];
 
     Object.keys(commits.authorDisplayNameMap).forEach((author) => {
       if (author) {
-        const obj = {
+        const obj: any = {
```
zhoukerrr, line 255: Is it possible to just declare a `user` and fill up the fields and push it to `res`? That way we don't need to type it to `any` then covert to `User` again

frontend/src/utils/api.ts
```diff
@@ -290,10 +298,10 @@ window.api = {
   setContributionOfCommitResultsAndInsertRepoId(dailyCommits, repoId) {
```
zhoukerrr, line 298: I think we can type the parameters here so that we don't need to cast to the specific types that we need within the function.

zhoukerrr (changes requested):
Really impressive work given how fast you did it. I like the organisation of the files that you added. Codes are also arranged in an easy-to-follow manner.

One small request will be to add the function parameter types wherever appropriate. That way when the functions are called, we can avoid any type error directly. Do let me know your thoughts as well.

frontend/src/utils/api.ts
vvidday: Yup thanks, just pushed a commit for this! I originally used `any` as I wanted to keep the code as similar as possible and work on a bigger refactor to remove the `any`s (+ in other .ts files as well), but I think it makes sense to do a tiny refactor here to avoid introducing a redundant typecast that will be removed again in the future.

frontend/src/utils/api.ts
vvidday: The functions are all typed in `types/window.ts` - for this function we're changing the type of each commit object since we're adding three fields (`repoId`, `insertions`, `deletions`), so the typecast is necessary (I think? not sure how else to do it in Typescript).

I decided to define the two types separately (`CommitResultRaw` = the raw form obtained from the json file without the above 3 fields, `CommitResult` = after adding the three fields) instead of using just one type with the three fields as optional fields. This was to avoid having to check `if (commitResult.repoId)` etc in the rest of the app.

zhoukerrr: Ah yes. Sorry I missed the declaration in `types/window.ts`. I get what you mean now. Yes this looks good to me.

zhoukerrr (approved):
Thanks for the work. LGTM!

HCY123902 (approved):

dcshzj:
@vvidday Sorry, there are some file conflicts in this branch. Are you able to resolve them?

vvidday:
@dcshzj Sorry, have just resolved the conflicts!

PR #1905, opened 2023-02-11
[#1904] Add ESLint support for TypeScript imports

### LINKED ISSUE #1904: ESLint support for TypeScript imports

Within a TypeScript file, the default way to import from another local TypeScript file is to omit the file extension, e.g. `import x from y` over `import x from y.ts`. ([Docs](https://www.typescriptlang.org/docs/handbook/modules.html#import)). Attempting to include the `.ts` file extension results in a TypeScript error: `An import path cannot end with a '.ts' extension. Consider importing './y.js' instead.`

[Additional context](https://github.com/microsoft/TypeScript/issues/27481) - since `.ts` files are compiled into javascript files at runtime, the imports ending in `.ts` might fail.

Currently, each `import x from y` within a TypeScript file where `y` is a local TypeScript file causes ESLint to throw 2 errors: 
- Unable to resolve path to module '../..'
- Missing file extension for '../..'

The first error is because ESLint is not configured properly for TypeScript imports, and the second is because ESLint is configured to throw an error if the file extension is missing from a local `.ts` import.

ESLint throws errors when we import from TypeScript files. The current codebase doesn't face this problem because there aren't any `.ts` imports inside any TypeScript file, but this will change when we start migrating over from `.js` files (e.g. #1903)

Configure ESLint to extend the [`import/typescript`](https://github.com/import-js/eslint-plugin-import#typescript) plugin, which will solve the first error.

For the second, I suggest that we configure the [import extensions](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md) rule to disallow file extensions for `.js` and `.ts` files. This will allow us to enforce consistency throughout the codebase, while keeping the default method of importing TypeScript files as mentioned above. 

The side effect is that we will have to remove the file extension from existing `.ts` file imports **inside `.vue` files**, since ESLint will enforce this rule across the codebase. This however will have no functional change.

1) Ignore the ESLint import extension rule. This will get rid of the error and eliminate the need to change the imports in existing `.vue` files, but will make the codebase inconsistent, i.e. Typescript imports in `.ts` files will not have a file extension but imports in `.vue` will.
2) Use `.ts` file extensions in imports, and [configure webpack](https://webpack.js.org/configuration/resolve/#resolveextensions) to resolve the imports at compile time. This will allow the code to run, but 1) requires us to modify webpack and 2) VSCode/IDEs will still show the `An import path cannot end with a '.ts' extension. Consider importing './y.js' instead.` error, since they do not know webpack is resolving the files at compile time.

I feel that the proposed solution is the most straightforward, as it will only result in a one-time change of imports in `.vue` files, following which the imports will be consistent throughout the codebase.

vvidday:
Fixes #1904 

### Proposed commit message

```
ESLint is currently not configured to properly support TypeScript
imports. This results in an error when importing local TypeScript files
with the file extension omitted, which is the default way to import
TypeScript files.

Let's configure ESLint to support TypeScript imports and enforce the
omission of file extensions for .ts and .js files. This will allow us to
pass the lint check while using the default import method, and enforce
consistency throughout the codebase.
```

### Changes to `.eslintrc.json`
Adding `plugin:import/typescript` allows ESLint to properly resolve TypeScript imports that don't include the `.ts` file extension, which solves the "Unable to resolve path to module '../..'" error.

```
   "import/extensions": [
      "error",
      {
        "js": "never",
        "ts": "never"
      }
    ]
```
The above lines configure the `import/extensions` rule to make sure imports should never end with `.js` or `.ts`. If the rule is broken, it will be considered an `error` by ESLint. This solves the "Missing file extension for '../..'" error.

### Others
The other changes are removing the `.ts` extensions from existing imports.

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -5,7 +5,8 @@
   "extends": [
     "airbnb-base",
     "plugin:vue/recommended",
-    "@vue/typescript"
+    "@vue/typescript",
+    "plugin:import/typescript"
   ],
   "rules": {
     "vue/component-definition-name-casing": [
@@ -47,7 +48,14 @@
     ],
     "prefer-object-spread": 0,
     "function-call-argument-newline": 0,
-    "vue/no-computed-properties-in-data": 0
+    "vue/no-computed-properties-in-data": 0,
+    "import/extensions": [
+      "error",
+      {
+        "js": "never",
+        "ts": "never"
+      }
+    ]
   },
   "parserOptions": {
     "parser": "@typescript-eslint/parser"
--- a/frontend/src/components/c-ramp.vue
+++ b/frontend/src/components/c-ramp.vue
@@ -34,8 +34,8 @@
 </template>
 
 <script>
-import brokenLinkDisabler from '../mixin/brokenLinkMixin.ts';
-import User from '../utils/user.ts';
+import brokenLinkDisabler from '../mixin/brokenLinkMixin';
+import User from '../utils/user';
 
 export default {
   name: 'c-ramp',
--- a/frontend/src/components/c-segment.vue
+++ b/frontend/src/components/c-segment.vue
@@ -25,7 +25,7 @@
 </template>
 
 <script>
-import Segment from '../utils/segment.ts';
+import Segment from '../utils/segment';
 
 export default {
   name: 'c-segment',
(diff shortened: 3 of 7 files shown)


ckcherry23 (approved):
LGTM!

HCY123902 (approved):

zhoukerrr (approved):
LGTM!

dcshzj (approved):

PR #1906, opened 2023-02-13
[#1886] Consolidate Node.js installation source of truth

### LINKED ISSUE #1886: Node.js Consolidation: To use the developer-installed version or isolated `RepoSense/build/node` version?
**How the developer-installed version works**

1.  Developer installs Node/npm, as per the [Developer Guide setup page](https://reposense.org/RepoSense/dg/settingUp.html)
2. The `PATH` variable is set for this developer-installed Node, so `npm` can be run manually on the command line.

In `integration.yml`, the step `Setup Node.js` is akin to using the Developer-installed version
<br/>

**How `Reposense/build/node` version works**

1. The [Liferay plugin](https://github.com/liferay/liferay-portal/tree/master/modules/sdk/gradle-plugins-node) in `build.gradle` will download a specific `nodeVersion` into `RepoSense/build/node`, a behaviour which can be disabled. This version of Node/npm will not be registered in the `PATH` variable, so `npm` cannot be run manually by itself on the command line.
2. Any Node-related tasks in `build.gradle` will use this version. Such tasks do not require the developer to install Node.js beforehand.
<br/>

**Problem**
We have 2 sources of truth for the version of Node to use. Since `integration.yml` also runs Gradle tasks (that use the Liferay downloaded version of Node), the `Setup Node.js` step is completely redundant. Updating Node versions also becomes complicated.

We should stick to only one source of truth for the Node.js version. I propose sticking to the developer-installed version for the following reasons:

* The `Setup Node.js` step in `integration.yml` supports semver syntax like `14.x` or `14.20.x`, which allows for minor or patch releases to be automatically updated in the step. The exact Node version like `14.20.1` can also be specified, meaning that there is flexibility depending on how we want to manage Node updates.
  * Based on my experience, Liferay's `nodeVersion` requires it to be very specific like `14.20.1`, meaning that version updates have to be made manually.
* For managing updates to dependencies in `package.json`, we may still need to use `npm` command outside of Gradle anyway, so it should be good practice for developers to have Node.js and npm installed.
<br/>

**Solution for Approach 1: consolidating to Developer-installed Node**

1. Disable Liferay `node.Download` in `build.gradle` (see the Liferay link for details)
2. Check that builds still work (both manual and CI)
<br/>

**Solution for Approach 2: consolidating to `RepoSense/build/node`**

If a majority of us think that we should go with `RepoSense/build/node` instead, these are the steps to follow:

1. Delete `Setup Node.js` steps from `integration.yml`
2. Update the [Developer Guide setup page](https://reposense.org/RepoSense/dg/settingUp.html) pre-requisites to say that Node.js is only needed if the developer intends to manage frontend dependencies given in `package.json` and `package-lock.json`.
3. Check that builds still work (both manual and CI)
<br/>

**Additional context**

Relevant issues for this discussion are #1720 and #1286.

It is best to wait for #1885 to merge before creating a PR for this since #1885 lays the groundwork for consolidating the `Setup Node.js` steps and Node version environment variable in `integration.yml` for easier Node version updates (approach 1) or removal (approach 2).

yhtMinceraft1010X:
Fixes #1886

### Proposed commit message

```
Two independent installations of Node.js are carried out. One by the
developer as well as the Setup Node.js step in integration.yml and
another by the Liferay plugin in build.gradle.

All Node-related tasks run by gradle use the gradle-installed version
while all Node tasks run manually use the developer or integration.yml
version.

This is confusing and difficult to maintain. There could also be
conflicts for package.json and package-lock.json if the
developer-installed and gradle-installed Node versions are different.

Let's disable downloading and installing Node.js through Gradle.
```

### Other information

Most of the discussion for this is in #1886


chan-j-d (approved):
LGTM!

dcshzj (approved):

PR #1907, opened 2023-02-14
[#1887] Convert run-checks.sh into a Gradle task

### LINKED ISSUE #1887: Convert `run-checks.sh` into a Gradle task

In `integration.yml`, running environmental checks is done by calling the script directly:
https://github.com/reposense/RepoSense/blob/25f4ce29e45c0365c0627c835812fe5e29090eb0/.github/workflows/integration.yml#L155-L156

However, other linters are run by Gradle.

In my opinion, developers are more likely to forget about the existence of environmental checks (even though they are important for passing CI) due to the different abstraction level compared to the other linters.

From my experience, I only notice problems in changes I make to `.md` files after the environmental checks fail during a CI run after pushing to my GitHub fork rather than locally.

As mentioned in the title, use Gradle as an abstraction for `run-checks.sh`.

**Additional context**
Inspired by #1215.

MarcusTXK:
Fixes #1887 

### Proposed commit message

```
Developers are more likely to forget about the existence of 
environmental checks (even though they are important for passing CI) due
to the different abstraction levels compared to the other linters. They
are also unable to test these checks locally easily and have to rely on
GitHub Actions. 

Let's use Gradle as an abstraction for run-checks.sh. Bash scripts can 
be run for Linux and macOS, while batch scripts can be run for Windows.
```

### Other information

The batch scripts have all been tested and are expected to function exactly as the bash scripts.


yhtMinceraft1010X:
I will let one member of @reposense/active-developers review this first.

config/checks/check-eof-newline.bat
```diff
@@ -0,0 +1,30 @@
+@echo off
+rem Checks that all text files end with a newline.
+
+set temp=check-eof-newline~%RANDOM%.tmp
```
sikai00, line 4: is there a chance where the fact that RANDOM generates a random integer from 0 to 32767 only, causes issues? e.g., if the batch files are run concurrently/files are overwritten/reading from wrong file due to the same random integer?

sikai00 (approved):
Looks good for the most part, though I have one small concern. Unfortunately I'm not very familiar with batch scripting, so I sort of learnt along the way. But looks okay!

config/checks/check-eof-newline.bat
MarcusTXK: Currently, the files are run sequentially, so there should be no issue. The random variable at the back is just for extra safety. Each script outputs its output into a temp file with a different name (e.g. `check-eof-newline~%RANDOM%.tmp` and `check-line-endings~%RANDOM%.tmp`, so if in the future, even if its refactored such that the three checks run concurrently, there will be no collision as they all have different temp file names.

yhtMinceraft1010X (approved):
LGTM

config/checks/run-checks.bat
```diff
@@ -0,0 +1,12 @@
+@echo off
+rem Runs all check-*.bat scripts, and returns a non-zero exit code if any of them fail.
+
+setlocal enabledelayedexpansion
+set ret=0
+
+for /R %%X in (check-*.bat) do (
+ call %%X
```
chan-j-d, line 8: I think it might be better to place `%%X` in double quotes for repo locations that happen to have a space like mine. Seems to work after

MarcusTXK: Good catch! I did not realize this as I do not have any spaces.

chan-j-d: hmm can't say im the most experienced with batch scripts but it seems the removal of the `/R` actually fixed it so maybe there's no need to add the quotation marks

config/checks/run-checks.bat
```diff
@@ -0,0 +1,12 @@
...
+rem Runs all check-*.bat scripts, and returns a non-zero exit code if any of them fail.
+
+setlocal enabledelayedexpansion
+set ret=0
+
+for %%X in (check-*.bat) do (
+ call "%%X"
+ if ERRORLEVEL 1 set ret=1
+)
+echo bye
```
chan-j-d, line 11: Is this intentional?

MarcusTXK: That was not meant to be committed, but was only intended for testing purposes, sorry about that

chan-j-d: No problem, was just a little shocked when the script said 'bye' after I ran it 😂

chan-j-d (approved):
LGTM!

Issue #1879, opened 2023-02-01
Mismatch between system test case and actual RepoSense behavior


chan-j-d:
**Tell us about your environment**

* **RepoSense Version:** master branch
* **OS and Version:** Windows 10
* **Web Browser and Version (if applicable):** N.A.

**Please include the steps to reproduce the bug.**

1. run `gradlew run -Dargs="-c src/systemTest/resources/ConfigSystemTest -f java adoc -u 1/11/2017 -F"`

The above is to simulate the `test30DaysFromUntilDateWithFindPreviousAuthors` config test case but without the test mode flag.

**What was expected to happen?**
Both commit and authorship json files produced are identical

**What actually happened? Please include a screenshot of the output.**
The 'expected' json file contains commits from the author 'nbriannl' which are not present in the actual normal reposense run. 

This might be related to the test mode flag causing the following flag to be set `AuthorConfiguration.hasAuthorConfigFile = false;` in test mode which causes new authors to be added to the analysis when the intended behavior would be to not include any new authors for that repo. 

**If possible, include the URL to your RepoSense report or log files (if any).**

Issue #1873, opened 2023-01-28
Backend: Parallelize System Tests


yhtMinceraft1010X:
**What feature(s) would you like to see in RepoSense?**

The running time for system tests is currently quite long. Although unit tests were parallelized in #1806, methods in system test classes still run sequentially.

**If possible, describe the solution**

Although I previously mentioned in #1770 that methods in `ConfigSystemTest` could not be parallelized using [`maxParallelForks` in `build.gradle`](https://kousenit.org/2020/11/22/simple-demo-of-gradle-parallel-tests/), it turns out that #1806 did the [parallelization using `junit-platform.properties`](https://www.baeldung.com/junit-5-parallel-tests). This allows methods within a test class to be parallelized. 

A `junit-platform.properties` file can be added to `systemtest/resources`. Since we currently have very few `systemtest` methods, a fixed approach using `junit.jupiter.execution.parallel.config.fixed.parallelism` should suffice.

One problem related to concurrency is that all methods within the same `systemtest` class will clone to an output directory with the same name. 

**Additional context**

Please read through the discussion in #1806 before attempting this.

sikai00:
After an investigation of `ConfigSystemTest.java`, I believe a new flag may be necessary in order to allow each repository to be cloned into different folders.

Currently, `test` is able to get down to the lower levels of being able to specify the directory of each cloned repository via `EXTRA_OUTPUT_FOLDER_NAME` and using the thread number as a folder name. This is not possible in `ConfigSystemTest.java` to do so, as I believe the main purpose of `ConfigSystemTest.java` is to run the RepoSense process via a single entry point of `RepoSense.main`.

<img width="790" alt="image" src="https://user-images.githubusercontent.com/63795894/218047675-d05fdb66-677b-48c8-a991-66041244ba7b.png">

We are also not able to circumvent this via the technique used by `LocalRepoSystemTest.java`. In `LocalRepoSystemTest.java`'s case, the idea was to test local repos (unlike in `ConfigSystemTest.java`). As such, it is able to define the parameters for cloning from a remote repo to a local repo and name each repos accordingly.

What do you guys think? Should I proceed with this via specifying a new flag, or does anyone have any other solution for this? @reposense/active-developers @reposense/active-reviewers

sikai00:
- Add a flag to RepoSense to allow the customization of the report’s output folder name
    - Pros:
      - Ease of use inside `LocalRepoSystemTest`
    - Cons:
[...]

PR #1910, opened 2023-02-19
[#1880] Change hasAuthorConfigFile to become a local variable

### LINKED ISSUE #1880: Changing `AuthorConfiguration.hasAuthorConfigFile` into a local variable
Possibly related to #1879, `AuthorConfiguration.hasAuthorConfigFile` is currently a global variable that determines whether during the analysis of repo files, a new author should be added and tracked. If an author config file is provided, an unrecognized author is ignored, otherwise it is added and tracked. 

This might cause parallel runs of test cases at the method level, especially system test cases, to have issues. A suggestion would be that in the `RepoConfiguration::merge` method, we can individually set the `AuthorConfig` inside each `RepoConfig` file to have the local variable flag `hasAuthorConfigFile` as true since the method is only ever called if an author config csv is specified.

sikai00:
Fixes #1880 
Part of #1879
Part of #1873 

### Proposed commit message

```
AuthorConfiguration.hasAuthorConfigFile is currently a static variable 
that determines whether during the analysis of repo files, a new author
should be added and tracked. If an author config file is provided, an
unrecognized author is ignored, otherwise it is added and tracked.

This might cause parallel runs of test cases at the method level, 
especially system test cases, to have issues.

Let's make AuthorConfiguration.hasAuthorConfigFile a local variable to
the AuthorConfiguration found within each RepoConfiguration.
```

### Other information

As mentioned by @chan-j-d, this issue is indeed related to #1879. This PR has to be merged in before the PR relevant to #1879. This is the case because in that PR, `AuthorConfiguration.hasAuthorConfigFile` is no longer set to false for all systemtest. LocalConfigSystemTest will fail as a result, as it does not have author-config.csv, however due to the global variable RepoSense believes it does.

This is also highly relevant to parallelizing of the ConfigSystemTest, as these static variables are shared across the threads used to run each test in ConfigSystemTest.

### DIFF
--- a/src/main/java/reposense/model/AuthorConfiguration.java
+++ b/src/main/java/reposense/model/AuthorConfiguration.java
@@ -334,6 +334,6 @@ public class AuthorConfiguration {
     }
 
     public boolean hasAuthorConfigFile() {
-        return this.hasAuthorConfigFile;
+        return hasAuthorConfigFile;
     }
 }
--- a/src/main/java/reposense/model/AuthorConfiguration.java
+++ b/src/main/java/reposense/model/AuthorConfiguration.java
@@ -20,7 +20,7 @@ public class AuthorConfiguration {
     private static final Pattern EMAIL_PLUS_OPERATOR_PATTERN =
             Pattern.compile("^(?<prefix>.+)\\+(?<suffix>.*)(?<domain>@.+)$");
 
-    private static boolean hasAuthorConfigFile = DEFAULT_HAS_AUTHOR_CONFIG_FILE;
+    private boolean hasAuthorConfigFile = DEFAULT_HAS_AUTHOR_CONFIG_FILE;
 
     private RepoLocation location;
     private String branch;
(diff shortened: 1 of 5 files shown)


ckcherry23 (approved):
LGTM!

src/main/java/reposense/model/AuthorConfiguration.java
```diff
@@ -329,11 +329,11 @@ public boolean containsName(String name) {
...
-    public static void setHasAuthorConfigFile(boolean hasAuthorConfigFile) {
-        AuthorConfiguration.hasAuthorConfigFile = hasAuthorConfigFile;
+    public void setHasAuthorConfigFile(boolean hasAuthorConfigFile) {
+        this.hasAuthorConfigFile = hasAuthorConfigFile;
     }
 
-    public static boolean hasAuthorConfigFile() {
-        return hasAuthorConfigFile;
+    public boolean hasAuthorConfigFile() {
+        return this.hasAuthorConfigFile;
```
chan-j-d, line 337: Minor point but iirc we don't use 'this' if the reference is not ambiguous, so just `return hasAuthorConfigFile`.

sikai00: Okay, I've updated it, thanks for noticing that!

chan-j-d (approved):
LGTM!

yhtMinceraft1010X (approved):
LGTM

PR #1911, opened 2023-02-19
[#1879] Remove unnecessary test mode logic in ConfigSystemTest

### LINKED ISSUE #1879: Mismatch between system test case and actual RepoSense behavior

sikai00:
Fixes #1879 
Possibly part of #1873 

### Proposed commit message

```
Due to legacy reasons, the test mode flag was added in order to allow 
systemtest to pass. However, systemtest was previously failing because 
the json file were incorrect and should not have included the addition 
of authors not listed in author config files.

In reality, RepoSense's behavior is correct and thus do not require 
hasAuthorConfiguration to be set to false for all the systemtest.

Let's update the json files to more accurately match the actual 
behavior of RepoSense. Let's also remove the setting of 
hasAuthorConfiguration to false if test mode is enabled.
```

### Other information

Should only be merged after #1910.

This is currently failing since LocalConfigSystemTest does not have author-config.csv, however due to the global variable RepoSense believes it does.


dcshzj:
Are you able to adjust the PR title to be more of an action? The title should describe what the PR will do/change instead of what the issue was. Thanks!

sikai00:
> Are you able to adjust the PR title to be more of an action? The title should describe what the PR will do/change instead of what the issue was. Thanks!

I have updated the title as per your request, I believe this should more accurately reflect what this PR is doing!

ckcherry23 (changes requested):
Hi @sikai00, I think there are some conflicts that need to be resolved as the structure of `commits.json` files had to be modified in #1882.

ckcherry23 (approved):
LGTM!

docs/dg/cli.md
```diff
@@ -18,9 +18,6 @@ syntax reference under Appendix section of User Guide.
 <box type="info" seamless>
 
 * Used in `ConfigSystemTest`.
-* Some test cases requires `AuthorConfiguration.hasAuthorConfigFile` to be set to `false` to pass. The exact
-reason is unknown as the test cases are pretty old to track back. Tracing this shows that there is something to do
-with `AnnotatorAnalyzer.findAuthorInLine()` where `AuthorConfiguration.hasAuthorConfigFile()` is called.
```
chan-j-d, line 23: Great catch for this, I actually forgot this statement was there.

chan-j-d (approved):
LGTM!

chan-j-d:
On a tangentially related note, I think we should create an issue regarding the test mode flag, to see if there's any other uses for it and otherwise we should consider removing it

yhtMinceraft1010X (approved):
LGTM

dcshzj:
@sikai00 Are you able to resolve the file conflicts? Thanks!

sikai00:
> @sikai00 Are you able to resolve the file conflicts? Thanks!

I have merged and resolved the conflict. Thanks for letting me know!

PR #1913, opened 2023-02-21
[#1912] Add TypeScript rules to ESLint

vvidday:
Fixes #1912
Should only be merged after #1914 fixes the lint errors

### Proposed commit message

```
Currently, ESLint is not configured to lint TypeScript specific code.
This may result in errors not being detected by the linter, which could
lead to bugs and a decrease in code quality and consistency.

Let's add the airbnb-typescript config to enforce style rules, and add
the typescript-eslint recommended rules to enforce greater type safety
within TypeScript code.
```

### Other information

We're specifying the TypeScript rules under `overrides` -> `files: *.ts` to separate the linting of `.vue` and `.ts` files. (Otherwise, the linter will try to lint the `.vue` files according to the TypeScript rules, which causes errors.)

The two rulesets that are extended are [airbnb typescript config](https://github.com/iamturns/eslint-config-airbnb-typescript), which is just the style rules from `airbnb-base` configured for TypeScript code, and the [recommended rules from typescript-eslint](https://typescript-eslint.io/rules/). As mentioned in #1912, these rules will help enforce type safety in our TypeScript code (e.g. by warning when we use `any`, which should be minimized since it defeats the purpose of using TypeScript).

The rule override is due to the base airbnb config configuring [SwitchCase to 1](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L124), which [enforces the indent in case statements](https://eslint.org/docs/latest/rules/indent#switchcase).

I'll be fixing the existing errors (e.g. use of `any`) in #1914, which should resolve the errors produced by this PR.

Issue #1912, opened 2023-02-21
Add TypeScript rules to ESLint


vvidday:
### What feature(s) would you like to see in RepoSense

Currently, our frontend's ESLint is linting [based on](https://github.com/reposense/RepoSense/blob/master/frontend/.eslintrc.json#L5) rules from `airbnb-base`, which is only configured for JavaScript code. The `@vue/typescript` ruleset only enables rules for TypeScript code within `.vue` files. 

This means that ESLint does not currently lint any TypeScript-specific code (It still lints regular JavaScript code within `.ts` files). This can be seen by introducing a style error in TypeScript code, such as removing the space before the curly brace [here](https://github.com/reposense/RepoSense/blob/master/frontend/src/types/types.ts#L10), and noticing that it still passes the linter. On the other hand, introducing a similar style error in JavaScript code, such as removing the space before the curly brace [here](https://github.com/reposense/RepoSense/blob/master/frontend/src/utils/api.ts#L23) will rightly be caught by the linter.

### Is the feature request related to a problem?

TypeScript code should be linted similar to JavaScript code, to enforce consistency and avoid errors.

### If possible, describe the solution

We can add the [airbnb-typescript](https://github.com/iamturns/eslint-config-airbnb-typescript) ruleset, which is similar to the existing airbnb-base ruleset that we are following, but configured for TypeScript code.

Additionally, I propose adding the [rules recommended by typescript-eslint](https://typescript-eslint.io/rules/) that includes rules that enforces more strict type safety (e.g. disallowing explicit `any` types). This will increase type safety across the codebase, which is one of the main reasons for using TypeScript in the first place.

PR #1914, opened 2023-02-21
[#1912] Refactor frontend code to minimize explicit any types

vvidday:
Part of #1912 

### Proposed commit message

```
Currently, there exists some explicit any types within the frontend
code. This functions as a dangerous escape hatch from the type system,
and it could lead to type errors not being caught at compile time.

Let's refactor the existing code to remove as many explicit any
declarations as possible, which will increase type safety of the
codebase.
```

### Other information

The modifications are separated by each individual commit.

The only `any` declaration that I didn't remove is in the declaration of the custom directive in `main.ts` - as far as I can tell it's a specific workaround for using that directive in Vue 3, and there's no easy way to remove it (nor would there be much benefit to removing it). I think for this specific case it is ok to add an exception for the linter to ignore this one line.

This PR will fix the errors that are introduced by the new linting rules in #1913.


frontend/src/utils/api.ts
```diff
@@ -77,12 +77,7 @@ window.decodeHash = function decodeHash() {
     .forEach((param) => {
       const [key, val] = param.split('=');
       if (key) {
-        try {
-          hashParams[key] = decodeURIComponent(val);
-        } catch (error) {
-          (this as any).userUpdated = false;
-          (this as any).isLoading = false;
-        }
+        hashParams[key] = decodeURIComponent(val);
```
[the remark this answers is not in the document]
vvidday: Oh I moved it to the app component itself (https://github.com/reposense/RepoSense/pull/1914/commits/eaf1b7ada45b84a71e05620ee03d3bdeb9c22fe5) since the `this.userUpdated` was updating the data of that vue component, so I thought it would make sense to change it within the component itself vs in the global function (since the error will bubble up anyway).

I also removed the `.isLoading` as I don't think it is used anywhere.

zhoukerrr (approved):
LGTM!

frontend/src/main.ts
```diff
@@ -21,7 +21,7 @@ dom.watch();
 const app = createApp(App);
 
 app.directive('hljs', {
-  mounted: (ele: any, binding: DirectiveBinding) => {
+  mounted: (ele, binding: DirectiveBinding) => {
```
ckcherry23, line 24:
```suggestion
  mounted: (ele: HTMLElement, binding: DirectiveBinding) => {
```
Is there any reason we do not define the type here?

frontend/src/main.ts
```diff
@@ -31,6 +31,8 @@ app.directive('hljs', {
 
 app.directive('observe-visibility', {
   beforeMount: (el, binding, vnode) => {
```
ckcherry23, line 33:
```suggestion
  beforeMount: (el: HTMLElement, binding: DirectiveBinding, vnode) => {
```
same for here

ckcherry23 (approved):
The changes LGTM! I just have a few questions on why we cannot enforce the type of parameters for the custom directives in `main.ts`.

frontend/src/main.ts
vvidday: Thanks for catching this! I think it would be good to add the type declaration here as well.

HCY123902 (approved):

PR #1916, opened 2023-02-23
[#1915] Make ReportGenerator and its Reporters non-static

### LINKED ISSUE #1915: Make ReportGenerator and its corresponding Reporters non-static
Currently, `ReportGenerator`, as well as its corresponding Reporters (`CommitsReporter` and `AuthorshipReporter`) are static utility classes. From a simple lookthrough of git blame, I believe it is made static only for legacy reasons (when RepoSense was a smaller project).

These classes have to be made non-static if we want to parallelize/have concurrent runs of RepoSense. For example, in `ConfigSystemTest`, each RepoSense instance is run through the `RepoSense.main` method. In order to parallelize `ConfigSystemTest`, we would have to make these classes non-static so that there will not be any thread interference.

Given that the `earliestSinceDate` is currently static in ReportGenerator (and `sinceDate` and `untilDate` are report-level parameter), I believe that this is possibly part of #634. If we are able to set this for each individual analyze of `RepoConfiguration`, we can possibly showcase two projects with different start and end dates at the same time.

It appears that most of the code can be easily made into non-static as it was used previously as a utility class. Tests will have to be changed also.

N/A.

### Additional context

I am currently working on parallelizing ConfigSystemTest as well, and I believe the only way to parallelize it is to make these classes non-static, due to thread interference. Related to #1873.

sikai00:
Fixes #1915 
Part of #1873 

### Proposed commit message

```
Currently, ReportGenerator, as well as its corresponding Reporters 
(CommitsReporter and AuthorshipReporter) are static utility classes. 
From a simple lookthrough of git blame, I believe it is made static 
only for legacy reasons (when RepoSense was a smaller project).

These classes have to be made non-static if we want to parallelize/have
concurrent runs of RepoSense. For example, in ConfigSystemTest, each
RepoSense instance is run through the RepoSense.main method. In order 
to parallelize ConfigSystemTest, we would have to make these classes 
non-static so that there will not be any thread interference.

Let's make ReportGenerator and its corresponding Reporters non-static.
```

### Other information

Within `CommitResultAggregator#aggregateCommitResults`, I passed in `ReportGenerator` as an argument, where it used `ReportGenerator`'s static method previously. This feels slightly unnatural, but it is only necessary as CommitResultAggregator has to set the `earliestSinceDate` in `ReportGenerator`. I believe there may be ways to possibly not include `ReportGenerator` as an argument, or even uncoupling this entirely, but I am unsure of means to do so.


src/main/java/reposense/commits/CommitResultAggregator.java
```diff
@@ -27,14 +27,14 @@ public class CommitResultAggregator {
      * Returns the {@link CommitContributionSummary} generated from aggregating the {@code commitResults}.
      * Uses {@code config} to obtain details like author name, since date and timezone.
      */
-    public static CommitContributionSummary aggregateCommitResults(
-            RepoConfiguration config, List<CommitResult> commitResults) {
+    public CommitContributionSummary aggregateCommitResults(
+            RepoConfiguration config, List<CommitResult> commitResults, ReportGenerator reportGenerator) {
```
MarcusTXK, line 31: Passing in `ReportGenerator` as an argument here as well as to `generateCommitSummary` is in my opinion an acceptable way as its the most direct way with the least hassle since it can no longer use `ReportGenerator`'s static method.

MarcusTXK (approved):
LGTM, nice job

yhtMinceraft1010X (approved):
LGTM

chan-j-d (commented):
I was wondering if there's a way to not pass `reportGenerator` object around just to set the `earliestSinceDate`. In particular, I think we can consider adding it in a field in `CommitContributionSummary` that `CommitResultAggregator` can add when constructing the summary. Then `ReportGenerator` reads it off of the `CommitContributionSummary`. Wondering what you think about it

sikai00:
Regarding this, my concern with this solution is that `CommitContributionSummary`'s role in this is to simply store the consolidated results of the commit analysis, and `earliestSinceDate` shouldn't be a part of this result. Furthermore, I believe `earliestSinceDate` has to be used in ReportGenerator to generate the summary file as well. It doesn't seem like `CommitContributionSummary` is ever returned to ReportGenerator for use as well.

chan-j-d:
My bad I should probably have been more specific. Currently `CommitResultAggregator` sets the `earliestSinceDate`. `ReportGenerator` calls `CommitsReporter::generateCommitSummary` which calls `CommitResultAggregator::aggregateCommitResults` which has a reference to `ReportGenerator` and sets the earliest since date there. Even with the changes you propose, this strange cyclic dependency still occurs.

My actual suggestion was that instead of `CommitResultAggregator` doing that, it can pass `earliestSinceDate` as an attribute to `CommitContributionSummary`. `reportGenerator` can then simply read off the `earliestSinceDate` from the `CommitContributionSummary`. All of this is done in the `ReportGenerator::analyzeRepo` method. Since right now, the `earliestSinceDate` is only being edited when this method is used anyway, just not by `ReportGenerator` itself so it will be 'in time' for the summary report generation. I don't feel quite as strongly that the `CommitsContributionSummary` should not have that result as the value is derived from the same analysis along the way, just currently not needed. 

For the last point, it's actually passed in the `analyzeRepo` method.
https://github.com/reposense/RepoSense/blob/dd34c383c9a541d7b3d6d5d9013356396001d372/src/main/java/reposense/report/ReportGenerator.java#L342-L357

I guess my main belief is that reducing unnecessary dependencies is in general a good thing as it makes the design of each class slightly more modular and testable. Which as I've only just come to realize, there's actually no `CommitResultAggregatorTest` while the equivalent one for file analysis does actually have one.

sikai00:
I have made the necessary changes as you have mentioned. Looking at it, I do believe this to be a better solution than the cyclic dependency. Thanks for bringing up this idea!

As for the `CommitResultAggregatorTest`, I believe it might be worthwhile to do so in another issue, as it will involve the creation of a branch in testRepo-Alpha if the test is to be similar to that of `FileResultAggregatorTest`.

chan-j-d (approved):
LGTM! Note the failing cypress tests that seem to be failing as a result of something related to the merge commits PR. Might want to test it after that PR is merged again

chan-j-d:
Also I agree regarding the `CommitResultAggregatorTest`, do create an issue for it and it would be great if you could address that too

sikai00:
> @sikai00 could you merge the current master and resolve the merge conflicts? Thanks!

Okay, I've resolved the conflicts. Thanks for letting me know!

PR #1918, opened 2023-02-25
Gradle: remove unneeded commons-text dependency

yhtMinceraft1010X:
### Proposed commit message

```
The commons-text dependency has been flagged as vulnerable by OWASP
dependency check. This dependency does not appear to be used anywhere
in the source code.

Let's remove this dependency to reduce the vulnerabilities in the
project.
```

### Other information

Attached is the relevant OWASP dependency check report
[RepoSense Dependency-Check Report.pdf](https://github.com/reposense/RepoSense/files/10830467/RepoSense.Dependency-Check.Report.pdf)


sikai00 (approved):
The first use of commons-text dependency appears to have been for a method `escapeHtml4` in `ReportGenerator`, which has since been removed from use after the separation of frontend and backend responsibility (#1691). LGTM!

chan-j-d (approved):
LGTM!

PR #1923, opened 2023-02-27
[#1891] Update Cypress to v11

vvidday:
Part of #1891 

### Proposed commit message

```
Currently, we are using Cypress 9, which is 3 major versions behind the
latest release. Older versions are less likely to be supported and
maintained.

Let’s update Cypress to version 11.
```

### Other information

Should be merged after #1922
Our tests aren't affected by the [breaking changes](https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-110) of Cypress 11, so there are no changes besides `package.json` and `package-lock.json`.

PR #1922, opened 2023-02-27
[#1891] Update Cypress to v10

vvidday:
Part of #1891

### Proposed commit message

```
Currently, we are using Cypress 9, which is 3 major versions behind 
the latest release. Older versions are less likely to be supported and
maintained.

Let’s update Cypress to version 10.
```

### Other information

Test files are renamed according to the [new recommended spec pattern](https://docs.cypress.io/guides/references/migration-guide#testFiles--specPattern) in Cypress 10.

The old configuration file has been migrated according to the [new changes](https://docs.cypress.io/guides/references/migration-guide#Configuration-File-Changes).

PR #1924, opened 2023-02-27
[#1891] Update Cypress to v12

### LINKED ISSUE #1891: Cypress: Update Major Release
## What feature(s) would you like to see in RepoSense

We are currently at Cypress 9. However, the latest major release is Cypress 12, which means that we are 3 major versions behind.

## Is the feature request related to a problem

As time goes on, an older version of Cypress is less likely to be maintained. With older APIs, there is a chance that we may be using code that is no longer industry-standard etc.

## If possible, describe the solution

Since major releases tend to having breaking changes, upgrading the major version is best done incrementally so that breaking changes are better managed. In this case, 3 sequential PRs may be opened for this:

1. Cypress 9 to 10
2. Cypress 10 to 11
3. Cypress 11 to 12

vvidday:
Fixes #1891 

### Proposed commit message

```
Currently, we are using Cypress 9, which is 3 major versions behind the
latest release. Older versions are less likely to be supported and
maintained.

Let's update Cypress to the latest version.
```

### Other information

Should be merged after #1922 and #1923.
Similar to #1923, our code isn't affected by the [changes introduced in Cypress 12](https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-120).


ckcherry23 (approved):
LGTM!

HCY123902 (approved):

yhtMinceraft1010X (changes requested):
There appears to be a merge conflict. You may need to regenerate the package-lock.json file as follows:
1) Accept the changes from `master` branch for `package-lock.json`
2) Incorporate your package.json changes
3) Regenerate `package-lock.json`.

PR #1925, opened 2023-03-01
[#1892] Gradle: change serveTestReportInBackground workingDir

### LINKED ISSUE #1892: Gradle Deprecations: workingDir in serveTestReportInBackground

* **RepoSense Version:** latest `master`
* **Web Browser and Version (if applicable):** N/A

1. Run `gradlew clean build systemtest cypress`
2. This will run checkstyle, test, systemtest and build the app

There are some warnings during `serveTestReportInBackground`, which is needed for `cypress`, shown in the below file:
[Stacktrace.txt](https://github.com/reposense/RepoSense/files/10610431/Stacktrace.txt)

All warnings fall under these 2 categories:
* Input directory, in the case, `workingDir` has [unreadable content](https://docs.gradle.org/7.4.2/userguide/upgrading_version_7.html#declare_unreadable_input_output) because there is a lock on it
* [Implicit dependencies between tasks](https://docs.gradle.org/7.4.2/userguide/validation_problems.html#implicit_dependency)

While the Gradle docs suggests solutions for each of the above warnings, I think that the actual problem is that the `workingDir` for `serveTestReportInBackground` is too broad since [by default it is the project root directory](https://github.com/psxpaul/gradle-execfork-plugin#javaexecfork) `RepoSense`. 

Output directories for tasks like checkstyle, test etc are all subdirectories of this root directory, thereby leading to the assumption of an implicit dependency even when there should not be one (example, running checkstyle, while recommended, should not be needed to run `serveTestReportInBackground` or its dependents `cypress` and `frontendTest`).

**Context**
This is the current code for `serveTestReportInBackground`:
https://github.com/reposense/RepoSense/blob/f760115ac7bb72a492bec1f0b5de96bdb9bd9b4f/build.gradle#L174-L186

Of note is that one argument `./frontend/cypress/config` is with reference to the root project directory, which needs to be changed if the `workingDir` property is changed. I'm of the opinion that any argument files or subdirectories should be copied somewhere into `build` such that `serveTestReportInBackground` can be run within a subdirectory that is separate from those used by unrelated tasks like checkstyle and test.

yhtMinceraft1010X:
Fixes #1892

### Proposed commit message

```
The working directory of serveTestReportInBackground is quite broad and
includes the entire build subdirectory. This results in tasks such as
checkstyle tasks being considered undeclared dependencies even when they
should not be.

Such unwanted dependencies will be flagged as errors in Gradle 8.

Let's move the working directory of serveTestReportInBackground into a
separate subdirectory and copy all the necessary files into this new
subdirectory.
```

### Other information

ExecFork plugin [GitHub page](https://github.com/psxpaul/gradle-execfork-plugin)


build.gradle
```diff
@@ -181,12 +191,13 @@ tasks.register('systemtest', Test) {
 }
 
 def serveTestReportInBackground = tasks.register('serveTestReportInBackground', com.github.psxpaul.task.JavaExecFork) {
-    dependsOn zipReport, compileJava, processResources, classes
+    dependsOn zipReport, compileJava, processResources, copyCypressConfig, copyMainClasses
     tasks.compileJava.mustRunAfter(zipReport)
     tasks.processResources.mustRunAfter(zipReport)
+    workingDir = 'build/serveTestReport'
     main = mainClassName
     classpath = sourceSets.main.runtimeClasspath
```
sikai00, line 199: I'm not 100% on this since my Gradle knowledge mainly comes from the past 10-20 minutes of googling, but is it possible that by using `sourceSets.main.runtimeClasspath` as a classpath, the copying of main classes is unnecessary? or rather i.e., does workingDir take precedence over the classpath stated?

sikai00 (approved):
LGTM! Just have one uncertainty regarding due to my inexperience in Gradle

sikai00:
Also as a sidenote do you think it would be good to include a description for each of the tasks added in `build.gradle`?

image

Right now I find it a little difficult to understand certain aspects of the `build.gradle` file due to my lack of experience in this, so I believe that adding a description to each of the task would be helpful for future developers in understanding the build file. This also serves as a form of documentation since I don't believe there's any on the gradle tasks right now. What do you think?

chan-j-d (approved):
LGTM!

build.gradle
yhtMinceraft1010X: The reason why I copied the classes was to ensure that the necessary stuff for running the task was self-contained within the workingDir. Because I initially put `build/classes` and Gradle flagged it with a deprecation warning.

yhtMinceraft1010X:
> Also as a sidenote do you think it would be good to include a description for each of the tasks added in `build.gradle`?
> 
> image
> 
> Right now I find it a little difficult to understand certain aspects of the `build.gradle` file due to my lack of experience in this, so I believe that adding a description to each of the task would be helpful for future developers in understanding the build file. This also serves as a form of documentation since I don't believe there's any on the gradle tasks right now. What do you think?

I can add documentation for the tasks that I'm introducing in this PR, but for the other tasks, it's best to raise it as a separate issue since it is out of scope of this PR.

build.gradle
```diff
@@ -181,12 +193,14 @@ tasks.register('systemtest', Test) {
 }
 
 def serveTestReportInBackground = tasks.register('serveTestReportInBackground', com.github.psxpaul.task.JavaExecFork) {
-    dependsOn zipReport, compileJava, processResources, classes
+    description 'Creates a background server process for the test report that is to be used by Cypress'
+    dependsOn zipReport, compileJava, processResources, copyCypressConfig, copyMainClasses
     tasks.compileJava.mustRunAfter(zipReport)
     tasks.processResources.mustRunAfter(zipReport)
```
dcshzj, line 199: Can I check if we need to ensure that the `copyCypressConfig` and `copyMainClasses` tasks need to run in a particular order?

yhtMinceraft1010X: Between these 2 tasks, there is no order in which they must be run, since the cypress config is only used at runtime.

dcshzj: Sorry, I was referring to these tasks in relation to the other tasks when `serveTestReportInBackground` is executed.

yhtMinceraft1010X: For `copyCypressConfig`, no particular order needed.

For `copyMainClasses`, it already depends on `classes`, which in turn depends on `compileJava` and `processResources`.

dcshzj: I see. This is ok for now but I think we should document the Gradle tasks in the developer guide. It is getting more complicated and the task dependencies are not immediately obvious when looking at the code.

dcshzj (approved):

PR #1930, opened 2023-03-05
[#1220] Cypress: test ramp chart in zoom view 

### LINKED ISSUE #1220: More robust front end tests
Our front end testing coverage is rather low, hence UI regressions can happen very often.

Let's add more front end tests, in particular:
- summary view's functionalities: 
  - [x]  `ctrl-clicking and zooming into specific range`
  - [x]  `ctrl-clicking and zooming into specific range when merge group`
  - [x]  `clicking on view all commits icon`
  - [x]  `merge group features`
  - [x] `filter breakdown features`
  - [x] `restoring of summary controls after page reloads`
- zoom view functionalities: 
  - [ ]  `show ramp chart for this period`, 
  - [x]  `hide/show all commit message bodies`, 
  - [x]  `reloading the page should restore zoom view`
- code view functionalities: 
  - [x]  `filter by glob`, 
  - [x]  `check file types`,
  - [x]  `hide/show all file details`, 
  - [x]  `reloading the page should restore code view`

These are some of the important tests that I can come up now.

ckcherry23:
Fixes #1220 

### Proposed commit message

```
Currently, there are very few tests for the ramp chart component of 
RepoSense. One aspect that is not tested is whether the ramps shown in 
the ramp chart are correct for the timeframe provided.

Let us make our front-end tests more robust by adding cypress tests to 
check whether the ramps shown in the ramp chart of the zoom view 
correspond with the commits from the selected time period. Let us also
test other aspects of the ramp chart like ramp properties, relative 
distances between ramps and links.
```

### Other information

For this PR, I check whether the ramp chart corresponds with the selected timeframe by checking if the number of ramps is equal to the number of commit messages. Therefore the confidence level of these tests depends on the correctness of the filtering of commit messages. 

For the other tests, one downside is that a lot of them are hardcoded or skipped as the values cannot be correctly estimated using conventional Cypress test methods. I will be exploring the use of jQuery methods to make these test cases more robust.


ckcherry23:
The frontend tests are failing due to the new merge commits made and can be fixed by merging #1882.

vvidday (approved):
LGTM! I think that it would be important to add tests to test for the correctness of the filtering of commit messages, as these tests rely on the correctness of those (as you mentioned) e.g. setting date/zoom range in summary panel correctly filters commits in zoom panel.

I'll open a separate issue/PR for that!

HCY123902 (approved):

zhoukerrr (commented):
@ckcherry23 More test cases are always welcomed and the test cases added are certainly useful. However, I think other test cases could be added to make this test file more powerful. I am not too sure if it is possible to cover cases regarding the width, start/end date, link to the commit, relative distance between each ramp. Would you mind exploring if any of those are possible?

If there are other areas of testing that you think we can add, feel free to add on. If the ones that I mentioned above are not possible to do/not worth the effort, do share below and feel free to discuss!

ckcherry23:
> I am not too sure if it is possible to cover cases regarding the width, start/end date, link to the commit, relative distance between each ramp. Would you mind exploring if any of those are possible?

Though the initial purpose of this PR was just to test if the ramp chart matches the selected timeframe, I agree that more tests need to be added for the ramp charts. However, many of these properties are difficult to test and some values like ramp width and distance may even change between platforms, even if the viewport size is the same. 

While I have added a few test cases with hardcoded expected values, I will explore if there are ways to test these in a more robust manner.

zhoukerrr (approved):
LGTM! I do agree with your point. But I think these hardcoded test cases do add value as a sanity check for future updates.

Issue #1936, opened 2023-03-11
Migrate entire frontend to TypeScript


vvidday:
### What feature(s) would you like to see in RepoSense

A completely TypeScript frontend.

### Is the feature request related to a problem?

As described in #1632, converting our frontend from JavaScript to TypeScript has many advantages, including type safety, better developer experience (intellisense etc), as well as making the codebase less confusing for new contributors (currently, there's many objects being passed around in the frontend that are of the `any` type, see #1632 for an example).

Even though TypeScript has been added in #1721, the frontend is still mostly JavaScript, including all our `.vue` files.

### If possible, describe the solution

We can start slowly converting the existing `.js` and `.vue` files to TypeScript. For `.vue` files, this can be done by adding `lang='ts'` to the `<template>` tag, wrapping the component with `defineComponent()`, and fixing all the errors that pop up.

While converting, we should be mindful of best practices, such as trying to avoid the use of `any` or `// @ts-ignore` except in cases where it is unavoidable. (Since using `any` defeats the purpose of using TypeScript in the first place.)

As this is a big task, it should be split up into multiple PRs.

### Additional context

#1903 has provided a decent starting point by typing the data that the frontend reads in from the `.json` files generated by the backend. This data is read/processed by the frontend components & passed between them. We can try to 'follow the path' of this data (e.g. `app.vue` -> `c-summary.vue` -> `c-summary-charts.vue` -> `c-zoom.vue`) and convert the files to TypeScript in order. i.e. we should focus on converting the parent components first.

### Main priority
- [x] `api.js` (all the global functions) - #1903 
- [x] `store.js` (Vuex store) - #1937 
- [x] `app.vue` (Root component) - #1938 
- [x] `c-summary.vue` - #1953 
- [x] `c-summary-charts.vue` - #1976
- [x] #1963
- [x] `c-authorship.vue` - #1969 

### Smaller files/components
- [x] `c-ramp.vue` (moved here as it is a smaller component compared to zoom/summary/authorship) #2037 
- [x] `c-segment.vue` #2035 
- [x] `c-segment-collection.vue` #2036
- [x] `c-resizer.vue` #2038
- [x] JS files in `utils` folder (`repo-sorter`, `ramp-colour-generator` etc) 
- [ ] #1946

zhoukerrr:
[...]

PR #1937, opened 2023-03-11
[#1936] Migrate store.js to TypeScript

vvidday:
Part of #1936

### Proposed commit message

```

Let's migrate the store.js file to TypeScript, to get types for the all
the states that are stored within the Vuex store. This will provide a
good starting point to migrate the Vue files that access and mutate the
state of the store.
```

### Other information

The declaration of the module and `ComponentCustomProperties` in `vuex.d.ts` provides type definitions to the `this.$store` object that's used in the `.vue` files.

Also, although I have checked on my end, this PR should probably not be merged until after #1913 , so we can see the updated results of the CI that includes the new TypeScript rules.


frontend/src/types/vuex.d.ts
```diff
@@ -0,0 +1,64 @@
...
+  summaryDates: SummaryDates;
+  mergedGroups: string[]
+  fileTypeColors: { [key: string]: string }
+  loadingOverlayCount: number
+  loadingOverlayMessage: string
+  isTabActive: boolean
+}
+
+declare module '@vue/runtime-core' {
+  interface State {
```
ckcherry23, line 51: Would it be possible to reuse the `StoreState` interface here?

ckcherry23 (commented):
Thanks for your work! I just have a few questions about typing.

ckcherry23 (approved):
LGTM!

frontend/src/store/store.ts
```diff
@@ -1,55 +1,62 @@
 import { createStore } from 'vuex';
+import { AuthorshipFile, CommitResult, DailyCommit } from '../types/types';
+import {
+  AuthorshipInfo,
+  StoreState,
+  SummaryDates,
+  ZoomInfo,
+} from '../types/vuex.d';
 
 export default createStore({
```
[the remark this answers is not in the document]
vvidday: Thanks for catching this!

frontend/src/store/store.ts
```diff
@@ -1,55 +1,62 @@
...
     mergedGroups: [],
     fileTypeColors: {},
     loadingOverlayCount: 0,
     loadingOverlayMessage: '',
     isTabActive: true,
-  },
+  } as StoreState,
   mutations: {
-    updateTabZoomInfo(state, info) {
+    updateTabZoomInfo(state, info: ZoomInfo) {
```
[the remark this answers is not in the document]
vvidday: Great point, I've added in the types for the `state` variable - I think the `context` is already typed

frontend/src/types/vuex.d.ts
vvidday: Pushed a commit for the state to extend the `StoreState` interface, thanks for pointing this out!

frontend/src/types/vuex.d.ts
```diff
@@ -0,0 +1,56 @@
...
+interface SummaryDates {
+  since: string;
+  until: string;
+}
+
+interface StoreState {
+  tabAuthorshipInfo: AuthorshipInfo;
+  tabZoomInfo: ZoomInfo;
+  summaryDates: SummaryDates;
+  mergedGroups: string[]
```
HCY123902, line 43: Just a comment regarding format consistency. I think the semicolons are missing here. Actually, the current eslint rules do not seem to enforce the presence of semicolons, which can be discussed in a separate issue

HCY123902 (changes requested):

frontend/src/types/vuex.d.ts
vvidday: Thanks for pointing this out! Just fixed the semicolons in the latest commits.

Just checked the ESLint rules that we're adding in #1912 - the ruleset does enforce semicolons for regular statements (the [semi](https://typescript-eslint.io/rules/semi/) rule), but not for TypeScript `interface`s and `type`s specifically, as you mentioned. 

We can enforce this via the [member-delimiter-style](https://typescript-eslint.io/rules/member-delimiter-style) rule, i.e. just adding the following to `.eslintrc.json`.
```
"rules": {
  "@typescript-eslint/member-delimiter-style": "error"
}
```
Will open up a separate issue about this where we can investigate rules like the above one. Thanks!

HCY123902 (approved):

vvidday:
Just added a ESLint exception in https://github.com/reposense/RepoSense/pull/1937/commits/c010d83a038e6037bec1405182eb91d0b0b84f79 for the empty interface in preparation for #1912 to be merged - empty interfaces are usually not meaningful, but this case is an exception.

Because the empty/"repeated" interface is inside the `declare module` block, which is used to type the `this.$store` within Vue components. We can't export to other files from within the `declare module` block, hence the need to declare `StoreState` at the top level (for other TS files), and once more in the `declare module` block for the Vue-component-specific typings.

zhoukerrr (approved):
LGTM! @ckcherry23 caught the points that I want to bring up. Thanks for the help!

PR #1938, opened 2023-03-11
[#1936] Migrate app.vue to TypeScript

vvidday:
Part of #1936

### Proposed commit message

```

Let's migrate the root component, app.vue, to TypeScript. As this is
the entry point of the app, converting this file to TypeScript will
enable the conversion of other Vue components.
```

### Other information

The diff might look confusing because of the re-ordering of the options (apologies for missing this in #1867), please look at https://github.com/reposense/RepoSense/commit/4bfc3aae9212d7860d3af802d63e53b891dc7be9 to see the actual diff of the TypeScript conversion.

I had to use `any` to define the app property of `window.app` - not too sure if it's possible to type this.

Lastly, regarding the topic of combining/separating PRs, I was not too sure whether to separate this from #1937. Because of the nature of TypeScript migrations, the PRs would probably be mostly sequential (e.g. this PR relies on the previous PR for the `this.$store` to be typed), and hence also contains the changes of that PR. This might result in more work, such as requiring maintainers to merge a specific PR before another, as well as any changes to the preceding PR might have to be merged back into this one. I guess there's a tradeoff between the above and keeping the PR's to 'one small change'. Currently, I leaned towards separating them as each change is fairly substantial and they are two separate files. Would love to get any feedback/opinions on this.

### DIFF
--- a/frontend/src/app.vue
+++ b/frontend/src/app.vue
@@ -75,24 +75,26 @@
       input(type="file", accept=".zip", v-on:change="updateReportZip")
 </template>
 
-<script>
+<script lang='ts'>
 import JSZip from 'jszip';
 import LoadingOverlay from 'vue-loading-overlay';
 import { mapState } from 'vuex';
+import { defineComponent } from 'vue';
 
 import cResizer from './components/c-resizer.vue';
 import cZoom from './views/c-zoom.vue';
 import cSummary from './views/c-summary.vue';
 import cAuthorship from './views/c-authorship.vue';
+import { Repo } from './types/types';
 
 const loadingResourcesMessage = 'Loading resources...';
 
-const app = {
+const app = defineComponent({
   name: 'app',
   data() {
     return {
-      repos: {},
-      users: [],
+      repos: {} as { [key: string]: Repo },
+      users: [] as Repo[],
       userUpdated: false,
 
       loadingOverlayOpacity: 1,
@@ -116,10 +118,13 @@ const app = {
   },
   methods: {
     // model functions //
-    updateReportZip(evt) {
+    updateReportZip(evt: Event) {
       this.users = [];
-
-      JSZip.loadAsync(evt.target.files[0])
+      const target = evt.target as HTMLInputElement;
+      if (target.files === null) {
+        return;
+      }
+      JSZip.loadAsync(target.files[0])
         .then((zip) => {
           window.REPORT_ZIP = zip;
         }, () => {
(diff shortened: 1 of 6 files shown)


frontend/src/app.vue
```diff
@@ -141,15 +164,16 @@ const app = {
       this.userUpdated = false;
       await this.$store.dispatch('incrementLoadingOverlayCountForceReload', 1);
       try {
+        const summary = await window.api.loadSummary();
```
ckcherry23, line 167: Could it be useful to have an interface for `summary`?

frontend/src/app.vue
```diff
@@ -189,7 +213,7 @@ const app = {
       window.encodeHash();
     },
 
-    renderAuthorShipTabHash(minDate, maxDate) {
+    renderAuthorShipTabHash(minDate: string, maxDate: string) {
```
ckcherry23, line 216:
```suggestion
    renderAuthorshipTabHash(minDate: string, maxDate: string) {
```
For the 'render X tab hash' functions, maybe we could have interfaces for the `info`/`zoomInfo` we are going to store?

ckcherry23 (commented):
Great job on getting the TypeScript migration underway! I noticed that there are a few objects that could benefit from having explicit types as they are destructured later on. Left some comments to that effect.

frontend/src/app.vue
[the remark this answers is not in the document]
vvidday: Yup, thanks for noticing this! Just added in the type.

frontend/src/app.vue
vvidday: `loadSummary` is already typed in `api.ts`, so there should already be type inference for the `summary` object!

frontend/src/app.vue
vvidday: Good point! Just added in the types for `zoomInfo` and `authorshipInfo` - also noticed that certain fields are not added on refresh, but added to the objects later - made these fields optional in the interface declarations, and we can assert the existence of these fields in `zoomView` and `authorshipView` when the fields are accessed (when we migrate those files over). Thanks!

ckcherry23 (approved):
LGTM!

frontend/src/app.vue
```diff
@@ -298,27 +324,7 @@ const app = {
       return REPOS[hashParams.tabRepo].location.location;
```
HCY123902, line 324: Is it possible to add typing for line 318 as well

frontend/src/types/types.ts
```diff
@@ -11,6 +11,7 @@ export interface CommitResult extends CommitResultRaw {
   repoId: string;
```
HCY123902, line 11: These seem to be be a duplicate of #1937. I think for new pull requests linked to #1936, it is recommended to follow the sequential merge strategy so as to not include the duplicate upstream changes from previous pull requests. Otherwise, if implementation changes from the upstream pull request, the changes need to be made here as well. You can keep these changes for this pull request since the previous pull request is about to be merged

frontend/src/app.vue
vvidday: The `window` object in 318 is typed from `types/window.ts`, so the objects in this function should all be implicitly typed

<details>
<summary>gif</summary>

app-types
</details>

frontend/src/types/types.ts
vvidday: Yup will try to follow this in the future - I had to include the changes from #1937 in this case as I needed the types from `store`. Will try to not include duplicate changes in future PRs unless absolutely necessary

HCY123902 (approved):

zhoukerrr (approved):
LGTM

PR #1947, opened 2023-03-13
Make ErrorSummary non-static

### LINKED ISSUE #1943: Make ErrorSummary non-static by removing singleton design pattern
What feature(s) would you like to see in RepoSense
Currently, `ErrorSummary` implements the singleton design pattern (and has done so since the beginning). Looking through the PR that first implemented `ErrorSummary` #743, the justification was that the "feature could be extended to report other issues, such as invalid file name that couldn't be analyzed, etc". This was indeed done in #1509, where the faulty repo link is reported to the user in the frontend.

Is the feature request related to a problem?
`ErrorSummary` have to be made non-static if we want to parallelize/have concurrent runs of RepoSense. For example, for systemtest, there exist a `@BeforeEach` method that clears the static error set. This made it such that only sequential runs of the tests were 100% valid. Even now, the systemtest `LocalConfigSystemTest`'s report is actually missing an error message of an invalid repo link `"errorMessage":"ttps://github.com/reposense/testrepo-Beta.git is an invalid remote URL."`

If possible, describe the solution

If applicable, describe alternatives you've considered
N/A.

Additional context
I am currently working on parallelizing ConfigSystemTest as well, and I believe the only way to parallelize it is to make these classes non-static, due to thread interference. Related to https://github.com/reposense/RepoSense/issues/1873.

sikai00:
Fixes #1943 

### Proposed commit message

```
Currently, ErrorSummary implements the singleton design pattern (and 
has done so since the beginning).

ErrorSummary have to be made non-static if we want to parallelize/have
concurrent runs of RepoSense. For example, for systemtest, there exist 
a @BeforeEach method that clears the static error set. This made it 
such that only sequential runs of the tests result in the correct 
output.

Let's make ErrorSummary non-static.
```

### Other information

Currently, error messages are added into the `ErrorSummary` via the constructor of `RepoLocation`, as the checking and validating of repo locations is done there. Given that error messages were present as early as that, I elected for Reposense::main have a local `ErrorSummary` that would be passed into methods that invokes the constructor of `RepoLocation`.

`RepoLocation` will no longer add error messages to `ErrorSummary` through a static method, as that would involve passing `ErrorSummary` as a parameter to `RepoLocation`. Instead, methods that handle the `InvalidLocationException` will perform the adding of the error message. This is found in the subclasses of `CsvParser` for runs of RepoSense with configs, and the relevant `RepoSense::getRepoConfigurations` for `--repo` LocationCli based runs.

### DIFF
--- a/src/main/java/reposense/RepoSense.java
+++ b/src/main/java/reposense/RepoSense.java
@@ -31,6 +31,7 @@ import reposense.parser.InvalidHeaderException;
 import reposense.parser.InvalidLocationException;
 import reposense.parser.ParseException;
 import reposense.parser.RepoConfigCsvParser;
+import reposense.report.ErrorSummary;
 import reposense.report.ReportGenerator;
 import reposense.system.LogsManager;
 import reposense.system.ReportServer;
@@ -58,16 +59,17 @@ public class RepoSense {
             CliArguments cliArguments = ArgsParser.parse(args);
             List<RepoConfiguration> configs = null;
             ReportConfiguration reportConfig = new ReportConfiguration();
+            ErrorSummary errorSummary = new ErrorSummary();
 
             if (cliArguments instanceof ViewCliArguments) {
                 ReportServer.startServer(SERVER_PORT_NUMBER, ((
                         ViewCliArguments) cliArguments).getReportDirectoryPath().toAbsolutePath());
                 return;
             } else if (cliArguments instanceof ConfigCliArguments) {
-                configs = getRepoConfigurations((ConfigCliArguments) cliArguments);
+                configs = getRepoConfigurations((ConfigCliArguments) cliArguments, errorSummary);
                 reportConfig = ((ConfigCliArguments) cliArguments).getReportConfiguration();
             } else if (cliArguments instanceof LocationsCliArguments) {
-                configs = getRepoConfigurations((LocationsCliArguments) cliArguments);
+                configs = getRepoConfigurations((LocationsCliArguments) cliArguments, errorSummary);
             } else {
                 throw new AssertionError("CliArguments's subclass type is unhandled.");
             }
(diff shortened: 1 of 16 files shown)


chan-j-d:
I took a quick look at the PR and I think we should take a step back from converting everything that is static to non static in the interest of parallelizing the system tests. I think this is the first one which I don't really agree that the design is an improvement over the current one and doing this just for parallelizing system test cases might not be the best course of action, especially since there might not be a guaranteed pay out.

I think the new design introduces a lot of class dependencies onto ErrorSummary that were not there before and likely shouldn't be there. I was wondering if simply adding `synchronized` locks to the ErrorSummary methods would be sufficient for your case in the same way that the logging classes had to do when the unit tests were parallelized. 

If absolutely necessary, I think we can instead have RepoSense hold the ErrorSummary object and each detected ILE is added to it since the individual parser classes pass them on to RepoSense anyway. Then pass that to ReportGenerator via the generateReposReport method instead of via constructor since it's a bit odd that all other arguments are passed via the method arguments and not in the constructor.

sikai00:
> I think this is the first one which I don't really agree that the design is an improvement over the current one and doing this just for parallelizing system test cases might not be the best course of action, especially since there might not be a guaranteed pay out.

I understand the sentiment here and I actually agree with you, in that the design isn't actually an improvement over the current one. The only real improvement here is that `ErrorSummary` which I believe should not have been static becomes non-static. Otherwise, I agree that this does introduce more class dependency, mainly in the parsers.

The changes in this PR were done with the intention of mimicking the original code as closely as possible. As such, ErrorSummary, which was previously static and mainly invoked and added to through via a static method in `RepoLocation`, is now passed to the parser classes, where the exceptions raised by RepoLocation is handled. 

> I was wondering if simply adding `synchronized` locks to the ErrorSummary methods would be sufficient for your case in the same way that the logging classes had to do when the unit tests were parallelized.

I believe adding synchronized would be insufficient in this case. The ErrorSummary singleton contains error messages for each run of RepoSense. Since I have parallel runs of RepoSense, by making the methods synchronized or by adding synchronized lock to the instance itself wouldn't be good as the methods will not encounter thread inference only, but ultimately the error messages held by the singleton will still be incorrect for the different RepoSense runs. Thus a single instance per run of ErrorSummary is needed.

> If absolutely necessary, I think we can instead have RepoSense hold the ErrorSummary object instead of passing it to the individual parser classes.

I believe this is worth exploring as an option. I was initially hesitant on doing so, since RepoSense did not hold any instance level variable from before, but I will take some time to experiment with it. Thanks for the suggestion, it is very much appreciated!

PR #1950, opened 2023-03-15
[#1715] Add RepoSense CLI Wizard Walkthrough

### LINKED ISSUE #1715: RepoSense CLI set-up walkthrough

The RepoSense command line arguments may be hard to navigate and takes some effort to reference (from the user guide) or commit to memory. Ideally, running the jar file should walk the user through the most relevant settings to get a report up and running.

This convenience feature should probably work something like `yarn init` which prompts the following sequentially:
```
name (your-project):
version (1.0.0):
description:
entry point (index.js):
git repository:
author:
license (MIT):
```

sikai00:
Fixes #1715
This PR is co-authored with @MarcusTXK.

### Proposed commit message
```
For a new user trying to get into RepoSense, the command line arguments
may be hard and intimidating to navigate and takes some effort to 
reference from the user guide. If a new user wants to try out RepoSense
(instead of immediately putting it to actual use), he may be put off 
by the effort needed for entry.

Let's implement a RepoSense CLI wizard. This makes it easier for a 
new user to understand how a run of RepoSense may work, understand 
some of the flags we have, and eases the learning curve required 
to use RepoSense.

Co-authored-by: Marcus Tang <txk.marcus@gmail.com>
Co-authored-by: Si Kai <sikai4111@gmail.com>
```

#### How to test
Pull from the PR branch and run reposense with the flag `--init`.

#### Overview of the wizard implementation
The wizard is implemented using a discrete event simulator (DES) style. This allows us to avoid long if-else conditions for the prompting logic, as each prompt can invoke another prompt simply by creating it and adding it into the queue.

```
└── wizard/
    ├── prompts/
    │   ├── Prompt.java
    │   ├── OptionalPrompt.java
    │   └── ...
    ├── WizardRunner.java
    ├── Wizard.java
    ├── BasicWizard.java
    └── InputBuilder.java
```

The `WizardRunner.java` acts as the simulator and takes in `Wizard.java` (implemented under `BasicWizard.java`, so as to allow for different future Wizards). `WizardRunner.java` maintains a Deque of `Prompt`, which is initially supplemented by a concrete implementation of `Wizard.java`.

Each `Prompt.java` may be used to add a config flag and its argument via `InputBuilder.java` by calling `WizardRunner::buildInput(Scanner sc)`. At the end when no other prompts exist in the deque, the method ends. The complete input string is created when `WizardRunner::getBuiltInput()` and subsequently used when `WizardRunner::run()` is finally called. In a similar sense to that of the current `systemtest`, the string is directly passed into `RepoSense::main` to produce the expected output and behavior.

Each concrete implementation of `Prompt.java` can return even more prompts to the Deque, so a `Prompt` is able to define the next `Prompt` that should be run. This is particularly useful in this example case where we ask the user a yes-or-no question if they want to indicate the start date for analysis, we can follow that up with a Prompt to get the date or choose to simply move on.

#### OptionalPrompt
`OptionalPrompt` is an abstract subclass of `Prompt`, as there were many `Prompt` that required a yes-or-no-to-have question, then requiring a follow-up `Prompt` should the user answer yes. The follow-up is provided through the `optionallyRun` method defined only in `OptionalPrompt`, such that the `WizardRunner` sees no difference between an `OptionalPrompt` and a regular `Prompt`.

#### Possible future improvements
Below are some possible future improvements that were not included in the interest of time and to ensure a minimal viable product.

1. Currently, errors are only detected when we pass the completed input string to `RepoSense::main`. If a user were to mistype an input, they will have to go through the trouble of re-entering all the inputs. We can have better error prompts such that in the case of an invalid input, we can immediately correct and re-prompt, avoiding having to make the user go through the whole setup again just to correct an erroneous input.


src/main/java/reposense/wizard/prompt/UntilPrompt.java
```diff
@@ -0,0 +1,25 @@
...
+    private static final String DESCRIPTION = "Enter the end date for the period to be analyzed";
+    private static final String FORMAT = "DD/MM/YYYY";
+
+    public UntilPrompt() {
+        super(DESCRIPTION, FORMAT);
+    }
+
+    @Override
+    public InputBuilder addToInput(InputBuilder inputBuilder) {
+        return inputBuilder.addUntilDate(getResponse());
```
vvidday, line 18: With regard to error handling, since the DES implementation stores the current configuration in a `StringBuilder`, is it impossible to check for errors w.r.t earlier inputs? (e.g. until date earlier than start date)

I personally don't think this is super important but was just wondering if something like this is possible for the current design.

vvidday (commented):
Thanks for the work on this @sikai00 @MarcusTXK! Just tested a bit on my end and the wizard seems to work as described. I think that the design choice of DES is quite clever as it makes the wizard easily extendable and makes it easy to reorder prompts/using prompts to generate other prompts, as you described.

Just have a couple of questions/comments as I'm not too familiar with the backend code

src/main/java/reposense/wizard/prompt/OptionalPrompt.java
```diff
@@ -0,0 +1,38 @@
...
+        if (getResponse().compareToIgnoreCase(YES_FLAG) == 0) {
+            return optionallyRun();
+        }
+
+        if (getResponse().compareToIgnoreCase(NO_FLAG) == 0) {
+            return new Prompt[] {};
+        }
+
+        // Repeat OptionalPrompt until a valid input is provided
+        return new Prompt[] {this};
```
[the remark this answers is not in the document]
sikai00: Currently we are yet to perform any forms of input validation. At this point, only when the RepoSense run happens, then the errors will be thrown.

I believe validation will be a further milestone down the road, as we are currently only planning for the MVP to occur.

src/main/java/reposense/wizard/InputBuilder.java
```diff
@@ -0,0 +1,362 @@
...
+
+    /**
+     * Crack a command line.
+     *
+     * @param toProcess the command line to process.
+     * @return the command line broken into strings.
+     * @throws BuildException if there are unbalanced quotes.
+     * An empty or null toProcess parameter results in a zero sized array.
+     */
+    public static String[] translateCommandline(String toProcess) {
```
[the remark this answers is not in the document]
sikai00: Thanks for spotting that! Yes that is better

src/main/java/reposense/wizard/prompt/UntilPrompt.java
sikai00: That is an interesting point.

I suppose one possible solution would be to group the Since and Until prompt into a single prompt that stores the state for both sinceDate and untilDate, and as such we can check for correctness from there on.

It may also be possible for us to define the dates as attributes/states within the actual WizardRunner, though I believe this can make things very messy as we will have a lot of attributes within the WizardRunner itself, and is less extensible to different possible versions of Wizards such as possibly an AdvancedWizard in the future.

vvidday (commented):
Thanks for making the changes, great job overall! I think that the current state is sufficient as an MVP and works as described.

Not too sure w.r.t to the merging (e.g. how complete it needs to be before merging (need input validation etc), and test coverage (currently codecov is reporting some lines not being covered by tests).

chan-j-d:
Some quick comments before reviewing:
- Should probably add the command to test somewhere in the PR itself
- The commit message's final line should be a short call to action. Something like 'Let's implement a setup walkthrough wizard'. The additional reasoning can be moved into the explanation portion of the commit message

src/main/java/reposense/wizard/InputBuilder.java
```diff
@@ -0,0 +1,289 @@
...
+        return '"' + path + '"';
+    }
+
+    private static String addQuotationMarksToPath(Path path) {
+        return addQuotationMarksToPath(path.toString());
+    }
+
+    public boolean isShallowCloning() {
+        return shallowCloning;
+    }
```
chan-j-d, line 288: I think we should only add the ones we are currently using into this class and delete the remaining. I don't think this feature will grow to implement many of the other flags

src/main/java/reposense/wizard/prompt/ViewPrompt.java
```diff
@@ -0,0 +1,30 @@
...
+        if (getResponse().isEmpty()) {
+            return inputBuilder.addView();
+        }
+        return inputBuilder.addView(Paths.get(getResponse()));
+    }
+
+    @Override
+    public Prompt[] run() {
+        return new Prompt[] {};
+    }
```
chan-j-d, line 29: I don't think this is correct. It should be view or no view without specification. The `-v` with argument is for when the user does not want to analyze anything and just wants to view a pre-analyzed report

src/main/java/reposense/wizard/BasicWizard.java
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Represents a basic wizard run.
+ */
+public class BasicWizard extends Wizard {
+    private static final Prompt[] INITIAL_PROMPTS = new Prompt[] {
+            new OptionalSincePrompt(),
+            new OptionalUntilPrompt(),
+            new OptionalViewPrompt(),
+            new RepoPrompt()
+    };
```
chan-j-d, line 18: I tried it myself and I don't think it makes a lot of sense to put the repo last. I think it should be first since I'm usually thinking of what I want to analyze and then the time interval to analyze rather than the other way around

chan-j-d (changes requested):
Some quick comments. I think the code is overall pretty well written and don't have too much to say in that direction. And regarding testing I'm also not too sure how much automated testing can be done

src/main/java/reposense/wizard/prompt/ViewPrompt.java
MarcusTXK: Got it, we will clean up the unused `InputBuilder.java` functions.

src/main/java/reposense/wizard/BasicWizard.java
MarcusTXK: I agree, lets change it to be the first.

src/main/java/reposense/wizard/InputBuilder.java
MarcusTXK: Got it, we will clean up the unused `InputBuilder.java` functions.

src/main/java/reposense/wizard/prompt/ViewPrompt.java
MarcusTXK: The logic for this is implemented in `OptionalViewPrompt.java`, where the user is prompted with "Do you want to start a server to display the report?", allowing them to choose whether or not to use the `ViewPrompt`. `ViewPrompt` currently contains logic for configuration of the `--view` flag if it is present.

chan-j-d: I understand that. But the OptionalViewPrompt shouldn't spawn a ViewPrompt since they dont serve the same purpose. There is no need for the user to supply an argument for --view unless they simply want to view a report that has already been generated, so it doesn't really make sense at all in this wizard walkthrough

src/main/java/reposense/wizard/InputBuilder.java
```diff
@@ -0,0 +1,289 @@
...
+        return this;
+    }
+
+    /**
+     * Adds the repo flag with the {@code paths} as arguments to the input.
+     * This method should only be called once in one build.
+     *
+     * @param paths The repo paths.
+     */
+    public InputBuilder addRepos(String paths) {
```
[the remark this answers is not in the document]
MarcusTXK: Yup, it simply passes the input entirely to `RepoSense::main`, where it is parsed as if the user passed it via the cli, so all parsing functionality for file paths will be the same.

src/main/java/reposense/wizard/prompt/ViewPrompt.java
sikai00: Yes, for this I believe it will be better to drop the need for an argument as there is no way of changing the output directory immediately after generating it anyway.

src/main/java/reposense/wizard/prompt/OptionalPrompt.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+}
```
chan-j-d, line 38: Not a requested change but more a point of discussion. Was wondering if there is a point in creating the subclasses since you could get this class to have a constructor that takes in the description and an instance of the prompt to be shown if Yes is chosen. 

I wrote this comment more from the perspective of testing since the subclass behaviors are almost identical I feel. See other comment about testing

src/test/java/reposense/wizard/WizardRunnerTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+        String expectedInput = new InputBuilder()
+                .addRepos(REPO_LINK)
+                .addSinceDate(SINCE_DATE)
+                .addUntilDate(UNTIL_DATE)
+                .addView()
+                .build();
+        assertEquals(expectedInput, wizardRunner.getBuiltInput());
+    }
+}
```
chan-j-d, line 146: I was wondering if it might be better to test individual prompts instead and have a separate test suite for input builder which tests combining multiple repo flags. 
This way, you don't have to rewrite the inputs every time the ordering of prompts is changed.

Additionally, I think it's missing test cases for completely invalid inputs for the optional prompt queries since those are supposed to repeat on invalid inputs if I'm understanding it correctly.

src/main/java/reposense/wizard/InputBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private static String addQuotationMarksToPath(String path) {
+        return '"' + path + '"';
+    }
+
+    private static String addQuotationMarksToPath(Path path) {
+        return addQuotationMarksToPath(path.toString());
+    }
+
+}
```
chan-j-d, line 98: Actually, I recall that there is an `InputBuilder` class in a util package in `test`. Was wondering if it might be worthwhile to move that class over (apologies about this in relation to my previous comment about unused input flags)

chan-j-d (changes requested):
Documentation changes look good. I made quite a few comments and suggestions this time round after reviewing it again since I understand the PR better now. 

Also wanted to know if you thought it might be better to include some fancy text like `[RepoSense Wizard]:` to kind of denote that the program is 'talking to the user'.

src/main/java/reposense/wizard/prompt/Prompt.java
```diff
@@ -0,0 +1,58 @@
...
+    public String getResponse() {
+        return response.trim();
+    }
+
+    public abstract InputBuilder addToInput(InputBuilder inputBuilder);
+
+    // each prompt may have additional effects depending on the prompt.
+    // when we run a prompt, the prompt will ask for input.
+    // for example, repos prompt will launch a vim window.
+    // a prompt may also create other prompts () (for example a open vim prompt)
```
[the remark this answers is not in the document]
sikai00: Nope, this is leftover from during development.

src/main/java/reposense/wizard/prompt/OptionalPrompt.java
sikai00: In what you mentioned, do you mean to have a single class per flag (e.g., SincePrompt should handle both asking if it should take in a SincePrompt, as well as actually taking in a SincePrompt)?

MarcusTXK: In a previous implementation, it actually took in a `Prompt` in the constructor so we could avoid spawning optional subclasses. However, we decided that this design wass a little strange OOP-wise.

You can view the discussion me and @sikai00  had [here](https://github.com/sikai00/RepoSense/pull/2#discussion_r1133125714).

src/test/java/reposense/wizard/WizardRunnerTest.java
sikai00: We can implement a `TestWizard` (similar to `BasicWizard`) that takes in individual prompts. That way, we can test any individual prompts/combinations of prompts. In this way, we reduce the number of test cases for `BasicWizard` as such, so that the re-writing of input doesn't have to happen for any of the possible new wizards we write, while still being able to test for correctness overall.

Just to check, the test case you mentioned is the one where we repeated input garbage to the OptionalPrompt, and finally a "yes"?

src/main/java/reposense/wizard/prompt/OptionalPrompt.java
chan-j-d: I was referring to the version Marcus mentioned. I don't think cyclic dependency is correct for it, the dependency relations are almost the same in either case and is not cyclic since Prompt doesnt depend on OptionalPrompt. This just changes a dependency to an association. Regardless, I don't think it's a big issue

src/test/java/reposense/wizard/WizardRunnerTest.java
chan-j-d: Sorry I'm not sure if I'm misunderstanding this 
> Just to check, the test case you mentioned is the one where we repeated input garbage to the OptionalPrompt, and finally a "yes"?

But is that suggesting that such a test case exists already or that that is the test case I want. Because I don't think I see such a test case and yes that's kind of what I want.

In general, I'd like to see more individualized test cases and only a few combined ones that are preferably done in a way that is independent of the order of the actual prompts since those may be reordered arbitrarily and we wouldnt want to have to fix test cases every time we do that. 

Also I guess I have to ask at this point what the difference between TestWizard and Wizard would be

sikai00: > But is that suggesting that such a test case exists already or that that is the test case I want. Because I don't think I see such a test case and yes that's kind of what I want.

Sorry for the confusion, this is suggesting the test case that you want, and does not already exist.

> In general, I'd like to see more individualized test cases and only a few combined ones that are preferably done in a way that is independent of the order of the actual prompts since those may be reordered arbitrarily and we wouldnt want to have to fix test cases every time we do that.

This would be a beneficial change. Though it is as you mentioned, there is still going to be some (but minimal number) of test cases with the ordered prompts.

> Also I guess I have to ask at this point what the difference between TestWizard and Wizard would be

TestWizard would just be a concrete implementation of the abstract Wizard class. For example, we have a BasicWizard currently. TestWizard will be a new, test-specific implementation of the abstract Wizard class where we can input specific types of Prompts specifically to test the prompts within the TestWizard.

PR #1919, opened 2023-02-25
Upgrade Gradle to 7.6.1

yhtMinceraft1010X:
### Proposed commit message

```
The latest Gradle 7 version is 7.6.1. This release includes the ability
to re-run tasks individually.

Let's upgrade to Gradle 7.6.1.
```

### Other information

Release notes for [Gradle 7.6.1](https://docs.gradle.org/7.6.1/release-notes.html)

PR #1951, opened 2023-03-16
Remove Gradle wrapper version definition

yhtMinceraft1010X:
### Proposed commit message

```
The gradleVersion property of the wrapper fixes the Gradle version used
by the wrapper. This is confusing and difficult to maintain as Gradle
gets upgraded.

Let's remove the definition of the gradleVersion property in the
wrapper. This results in the current version of Gradle itself being used
for the wrapper too.
```

### Other information

I noticed that I missed out on updating the wrapper version when doing [previous upgrades of Gradle versions](https://github.com/reposense/RepoSense/pull/1919#pullrequestreview-1328993999).

Looking at the Gradle docs, it turns out that the `gradleVersion` property is for defining [what version of Gradle the wrapper itself is using](https://docs.gradle.org/7.6/dsl/org.gradle.api.tasks.wrapper.Wrapper.html#org.gradle.api.tasks.wrapper.Wrapper:gradleVersion). By default, it will match the version that Gradle itself uses. Hence, the removal of the definition.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,10 +52,6 @@ sourceSets {
     }
 }
 
-wrapper {
-    gradleVersion = '7.6.1'
-}
-
 def installFrontend = tasks.register('installFrontend', com.liferay.gradle.plugins.node.tasks.ExecutePackageManagerTask) {
     workingDir 'frontend/'
     args = ['ci']


sikai00 (approved):
LGTM!

dcshzj (approved):
LGTM!

HCY123902 (approved):

PR #1953, opened 2023-03-19
[#1936] Migrate c-summary.vue to TypeScript

vvidday:
Part of #1936

### Proposed commit message

```

Let's migrate one of the main components, c-summary.vue, to TypeScript.
In addition to providing greater type safety, the migration of this
component will also allow the migration of its child components, such
as c-summary-charts.vue.
```

### Forcing `hashParams` values to strings
One change that is included in this PR is [storing all the values of `window.hashParams` as strings](https://github.com/reposense/RepoSense/commit/5bd37bc45106edbed6a56c5ef0fd227e0af4b2d9) instead of strings or booleans.  This was done to prevent having to check/typecast between string/boolean constantly.

This should be safe because although the values of the hashParams object are sometimes set to booleans, they aren't ever referenced as booleans, and are decoded back as strings before being used.
- [In the root component `app.vue`](https://github.com/reposense/RepoSense/blob/master/frontend/src/app.vue#L313), on `created()`, the `window.decodeHash()` function is called, which [converts all `hashParams` values to strings](https://github.com/reposense/RepoSense/blob/master/frontend/src/utils/api.ts#L80).
- The rest of the codebase already treats the values of the `hashParams` values as strings:
  - https://github.com/reposense/RepoSense/blob/master/frontend/src/app.vue#L224
  - https://github.com/reposense/RepoSense/blob/master/frontend/src/app.vue#L240
  - https://github.com/reposense/RepoSense/blob/master/frontend/src/views/c-authorship.vue#L364
  - https://github.com/reposense/RepoSense/blob/master/frontend/src/views/c-zoom.vue#L324
### `any` type for comparator
When typing the `window.comparator` method, I mistakenly thought that the sorting function `fn` returned the same type - but this is not the case [e.g. takes in a Commit but returns a date string](https://github.com/reposense/RepoSense/blob/master/frontend/src/views/c-summary.vue#L530). I couldn't find anything like a comparable type for TypeScript, so I think we have to use any here. (Please correct me if I'm wrong!)

### DIFF
--- a/frontend/src/types/window.ts
+++ b/frontend/src/types/window.ts
@@ -46,7 +46,7 @@ declare global {
     getDateStr: (date: Date) => string;
     getHexToRGB: (color: string) => number[];
     getFontColor: (color: string) => string;
-    addHash: (newKey: string, newVal: string) => void;
+    addHash: (newKey: string, newVal: string | boolean) => void;
     removeHash: (key: string) => void;
     encodeHash: () => void;
     decodeHash: () => void;
--- a/frontend/src/types/window.ts
+++ b/frontend/src/types/window.ts
@@ -44,7 +44,7 @@ declare global {
     isMacintosh: boolean;
     REPORT_ZIP: JSZip | null;
     deactivateAllOverlays: () => void;
-    getDateStr: (date: Date) => string;
+    getDateStr: (date: number) => string;
     getHexToRGB: (color: string) => number[];
     getFontColor: (color: string) => string;
     addHash: (newKey: string, newVal: string | boolean) => void;
--- a/frontend/src/types/window.ts
+++ b/frontend/src/types/window.ts
@@ -13,7 +13,8 @@ interface ComparatorFunction<T> {
 }
 
 interface SortingFunction<T> {
-  (item: T, sortingOption?: string): T;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (item: T, sortingOption?: string): any;
 }
 
 interface Api {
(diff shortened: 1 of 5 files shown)


frontend/src/views/c-summary.vue
```diff
@@ -127,38 +127,50 @@
...
   data() {
     return {
-      checkedFileTypes: [],
-      fileTypes: [],
-      filtered: [],
+      checkedFileTypes: [] as string[],
+      fileTypes: [] as string[],
+      filtered: [] as User[][],
       filterSearch: '',
       filterGroupSelection: 'groupByRepos',
```
ckcherry23, line 175: Do you think it would be useful to add `enum`s for typing the different selection choices we have in the summary-picker dropdowns?

vvidday: Great point, I added in the enums in the latest commit(s)!

ckcherry23 (approved):
LGTM!

frontend/src/types/window.ts
```diff
@@ -13,7 +13,8 @@ interface ComparatorFunction<T> {
 }
 
 interface SortingFunction<T> {
-  (item: T, sortingOption?: string): T;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (item: T, sortingOption?: string): any;
```
HCY123902, line 17: Please consolidate this with the modification in #1965

HCY123902 (approved):

zhoukerrr (approved):
LGTM. The use of enum here is very appropriate in my opinion.

PR #1954, opened 2023-03-19
[#1894] Add embeddable ramp widget

### LINKED ISSUE #1894: Add embeddable ramp widget

The ability to copy a link for an embeddable ramp widget with the click of a button and embed the ramp widget into any existing website will be a useful feature. This allows users who intend to feature any contributions made to a repository/projects in their portfolio/any website, especially as I feel it is both eye-catching and provides alot of useful information.

Ideally, the embedded ramps should follow any filters configured by the users above so that the information displayed by the ramps can be customized.

image

This is supported by Vue.js as it allows the building of [standard Web Components](https://vuejs.org/guide/extras/web-components.html) that can be embedded in any HTML as stated in their docs [here](https://vuejs.org/guide/extras/ways-of-using-vue.html#embedded-web-components).

MarcusTXK:
Fixes #1894 

### Proposed commit message

```
The ability to copy a link for an embeddable ramp widget with the click 
of a button and embed the ramp widget into any existing website will be 
a useful feature. This allows users who intend to feature any 
contributions made to a repository/projects in their portfolio/any 
website, especially as I feel it is both eye-catching and provides a 
lot of useful information. The embedded ramps should follow any filters 
configured by the users above so that the information displayed by the 
ramps can be customized. 

Let's add a button to allow users to copy and embed widgets.
```

### Screenshots
RepoSense Report
image

Embedded Group Ramp (Unmerged)
image
* Note that elements such as the numbering and certain buttons that do not have usage in iframes are hidden. An additional button that redirects the user to the RepoSense report is added to each user.

Embedded Group Ramp (Merged)
image
image
Note: The reason why there is a lot of space below the chart is due to the contribution bars not rendering, this is fixed in #1959 

Embedded Single Ramps
image
* Similarly here, the numbering and unnecessary buttons are removed.

The height of the iframe are calculated based on `the number of charts in the iframe * chart height + title (if present)` such that no scrolling of the iframe is required when viewing it. The width is set to an arbitrary value of `800`. These dimensions can be edited by the user in the iframe should they prefer other values.

Tested the iframes on the following html file (green bg to see the widget margins better) :
```html
<!DOCTYPE html>
<html>
<head>
	<title>Example with iFrame:</title>
</head>

<style>
body {
  background-color: #7bc96f;
}
</style>
<body>
	<h1>Embedded Widget</h1>
        <iframe src="XXX" frameBorder="0" width="800px" height="XXXpx"></iframe>
</html>
```

#### Why iframes?
For embeddable widgets I initially considered two options, the first being [standard Web Components](https://vuejs.org/guide/extras/web-components.html) that can be embedded in any HTML or iframes.

I ultimately decided to use iframes as using standard Web Components required prebuilding of the components and for users to add two items, a link in the head for css and a script in the body for the widget source, which is troublesome for the user. Iframes on the otherhand can be directly embedded in the body of any html file with no hassle.

#### Why hash mode for history mode?
With the current design, `/` will lead to the usual home page, while `/widget` will lead to the embedded widget. This required the addition of routers and `Vue Router` was utilized for this

This was the main roadblock for me as I initally tried using HTML5 Mode with `createWebHistory()`. This allows urls to look as per usual, however since RepoSense is a single page client side app, without a proper server configuration, the users will get a 404 error if they access any url directly in their browser that was not the root path. I did not manage to configure webpack to work well with this and eventually settled on Hash Mode, which uses a hash character (#) before the actual URL that is internally passed (e.g `/#/widget` instead of `/widget`). This works well, but has a caveat of having poorer SEO, however since this routing is only used for widgets for now, and the reports are not likely built with the intent for optimal SEO, this is an acceptable alternative.

#### Reuse of `c-summary-charts.vue`
I initially intended to create a seperate widget component, however this has the drawback of being hard to maintain as if the summary chart is updated, the widget would have to be updated separately. Since `c-summary-charts.vue` renders the charts in a double for loop, I realized that we can easily reuse it but filter out non relevant elements and only render the specified charts. This can be thought of as a matrix, where we access the required row or required element directly by keeping track of indexes. This is done via the url parameters `chartGroupIndex` and `chartIndex`.

#### Overall
`app.vue` only renders `router-view` now, which renders `c-home` or `c-widget` depending on if the routing is `/` or `/#/widget`.  I have extracted out all common logic between `c-home` and c-widget` in `app.vue`, others are done in the respective child components. `c-summary` does not render the `summary-picker` if `isWidgetMode` prop is passed in, and now passes in `chartGroupIndex` and `chartIndex` to `c-summary-charts`  which only renders the specified indexes.

One other notable change is that the "scoped" css that was previously in `c-summary` has been shifted into `c-home` and c-widget`, since they have differing css styles.

To Do:
- [x] Write up designs considerations, roadblocks etc
- [x] Add functionality to embed merged group charts (currently assumes unmerged)
- [x] See if `frontend/package-lock.json` lines updated can be reduced (likely because I am using Node 16)
- [x] Update height calculation of widget to not use hardcoded for chart height and title height
- [ ] Check if documentation needs to be updated
- [ ] Update alot of the cypress tests (will update it all once the design for the widgets are approved)
- [x] Find a way to share functions between components (right now alot of logic is duplicated in both the widget and the original code
- [x] Clean up code for some unused functions and css in widget files + Refactor out common functions to avoid repeating

```
Previous before refactor:
app->c-home->c-home-summary->c-summary-charts
   ।->c-widget-home->c-widget-summary->c-summary-charts
     
Now:  (shares both c-summary and c-summary-charts + shared logic btw home and widget shifted to app)
app->c-home->c-summary->c-summary-charts
   ।->c-widget->c-summary->c-summary-charts
```


ckcherry23 (commented):
@MarcusTXK I took a look at your file changes and seems like a lot of CSS classes are repeated in `c-home.vue` and `c-summary.vue`.  Is there anything you can do to avoid this code duplication?

zhoukerrr:
@MarcusTXK I'm sorry but the team is having some difficulties understand what the PR is doing. Will it be possible to break it down to smaller pieces that we can inclemently review and merge?

MarcusTXK:
Alright sure, perhaps I can breakup the PR and start with merging the PR for the addition of the router first.

MarcusTXK:
As this PR is quite outdated from the master branch, (e.g. does not include some of the typescript refactorings), I am deciding to close this PR. This is also because the separated routing logic in #1974 was written from scratch with reference to this PR (and shares no common commits), it will be hard to rebase the changes onto the new code with the Vue Router.

This is used as a reference for the updated widget code PR #1988, where the code is completely rewritten from scratch since their branches share almost no common commits.

Issue #1872, opened 2023-01-28
Frontend DevOps: Upgrade Node to 16 then 18


yhtMinceraft1010X:
**What feature(s) would you like to see in RepoSense?**

Node 14 will reach [end-of-life](https://github.com/nodejs/release#release-schedule) on 30 April 2023. Although Node 16's EOL would usually be 1 year later, it has been pushed forward to 11 September 2023 since OpenSSL 1.1.1 is also [no longer supported](https://nodejs.org/en/blog/announcements/nodejs16-eol/) on the same day.

**If possible, describe the solution**

The Node version needs to be updated in `build.gradle` as well as our workflows under the `.github` directory.

Suggested upgrade step:

1. Upgrade to Node 16 before 30 April 2023. 
2. Upgrade to Node 18 before 11 September 2023

The rationale for doing this is so that it is easier to resolve any breaking changes.

**Additional context**

You may want to keep #1871 in mind while working on this.

ckcherry23:
For first-time contributors, we have already updated Node to version 16 in PR #1956. Please help update Node to version 18 using the same PR as a reference.

sopa301:
I tried working on this, but was unsuccessful. Here are some things I learnt:

- Updating Node to 18 requires [node-sass](https://www.npmjs.com/package/node-sass) to be updated to at least 8.0
- However, node-sass has been [depreciated](https://sass-lang.com/blog/libsass-is-deprecated/) in favor of dart-sass (sass), which is already used in the project.
- Additionally, node-sass on windows has a few more [requirements](https://github.com/sass/node-sass#install) that make it harder to set up.
- Therefore, it would be easier to remove node-sass and use sass exclusively.
- Attempting to do so resulted in a bug similar to #1784 , but the cause is unclear.

PR #1956, opened 2023-03-19
[#1872] Update Node version to 16

ckcherry23:
Part of #1872 

### Proposed commit message
```
Node.js 14 will be reaching its end-of-life on 30 April 2023.

Let us update Node.js to use Node 16 in the GitHub workflows.
```

### Other information
Since the Node version is not referenced in `build.gradle` anymore after #1906, it does not have to be updated there.


vvidday (commented):
The current changes LGTM, just wanted to check regarding the `package-lock.json`s for the frontend - if I'm not wrong running `npm install` with node 16 (compared to 14) will upgrade the lockfile version from 1 to 2. Should we regenerate the lockfile in this PR as well?

ckcherry23:
> The current changes LGTM, just wanted to check regarding the `package-lock.json`s for the frontend - if I'm not wrong running `npm install` with node 16 (compared to 14) will upgrade the lockfile version from 1 to 2. Should we regenerate the lockfile in this PR as well?

I am unsure if I should, as this adds 10000+ lines of code (dependencies) to the frontend `package-lock.json` file. @yhtMinceraft1010X Can you let me know whether I have to regenerate this file?

yhtMinceraft1010X:
> 
> I am unsure if I should, as this adds 10000+ lines of code (dependencies) to the frontend `package-lock.json` file. @yhtMinceraft1010X Can you let me know whether I have to regenerate this file?

I'd say go for it.

HCY123902 (approved):

yhtMinceraft1010X (changes requested):
The DG node requirement also needs to be updated.

HCY123902:
I think the Node version for developer setup here can be restricted to `16` as well, in order to comply with the CI workflow

image

docs/dg/settingUp.md
```diff
@@ -10,7 +10,7 @@
 
 **Prerequisites:**
 * **JDK `1.8.0_60`** up to **`17`** ([download :fas-download:](https://www.oracle.com/technetwork/java/javase/downloads/index.html)).
-* **Node.js** `14.19.0` to the latest minor version for `16` ([download :fas-download:](https://www.npmjs.com/get-npm)).
+* **Node.js** `16.19.1` to the latest minor version for `18` ([download :fas-download:](https://www.npmjs.com/get-npm)).
```
yhtMinceraft1010X, line 13: Has Node 18 been proven to work with RepoSense? Mentioning it here implies that we are sure it does.

ckcherry23: I have tried the Gradle tasks and CI with both Node 16 and 18, which work for me. There are some warnings about deprecated dependencies though:

Screenshot 2023-03-25 at 1 30 02 PM

yhtMinceraft1010X: These deprecated dependencies shouldn't be a problem for now. 

Some of them might be resolved by updating some of the dependency version numbers in `package.json`, but this can be done in a separate PR.

yhtMinceraft1010X (approved):
LGTM

PR #1957, opened 2023-03-20
[#1945] Enforce stricter formatting rules for TypeScript files

### LINKED ISSUE #1945: Enforce stricter formatting rules for TypeScript files in ESLint

As pointed out by @HCY123902, our current ESLint ruleset for TypeScript files does not enforce certain formatting rules, such as semicolons as delimiters in TypeScript interfaces.

We could add these formatting rules to ESLint.

We are [already following](https://se-education.org/guides/conventions/javascript.html) the [Airbnb coding standard](https://github.com/airbnb/javascript/blob/master/README.md), which contains style-specific rules (e.g. comma dangle, single quote for strings, semicolons at end of statement) that helps maintain consistency across the codebase, which improves readability.

As TypeScript introduces new code that isn't covered by the above JavaScript coding standards, it would be good to enforce consistency in terms of style.

#1913 adds support for the [typescript-eslint](https://typescript-eslint.io/) plugin, that allows us to enforce rules for TypeScript code. Although we add the Airbnb Typescript and Typescript-eslint recommended rulesets in that PR, there are still certain style rules that aren't included in those rulesets. 

Hence, the solution would be to simply add these missing rules to our `eslintrc.json` file.

**Additionally, it would be good to update the DG to specify these additional rules that we're adding, since they are not included in the Airbnb ruleset that is currently linked on the DG.**

### Additional context

It is not particularly easy/straightforward to get a list of rules that might be missing from the Airbnb ruleset.

All the rules included by the Airbnb typescript ruleset can be found in [this file](https://github.com/iamturns/eslint-config-airbnb-typescript/blob/master/lib/shared.js). The syntax might be confusing, so consider this example in [line 152](https://github.com/iamturns/eslint-config-airbnb-typescript/blob/master/lib/shared.js#L152) of the above file:
```
    'no-magic-numbers': 'off',
    '@typescript-eslint/no-magic-numbers': baseBestPracticesRules['no-magic-numbers'],
```
The first line is just turning the default ESLint option for that rule off, to not interfere with the TypeScript version. The second line is the actual value of that rule that is being set. However, the problem here is that since Airbnb imports the value from their base ruleset, we actually have to reference the [JavaScript repo](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base/rules) to find the actual value. In this case, we can see it's in `baseBestPracticesRules`, so we go to the [`best-practices.js`](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/best-practices.js) file in the JavaScript repo above to find the value that is being used:
```
    'no-magic-numbers': ['off', {
      ignore: [],
      ignoreArrayIndexes: true,
      enforceConst: true,
      detectObjects: false,
    }],
```
As you can tell, the process is quite tedious/cumbersome, hence why I mentioned that it is not really easy to get a list of rules that we might be missing.

### Rules to add

I've done a preliminary comb through the [TypeScript rules](https://typescript-eslint.io/rules/) and so far there are two rules that I think we should consider adding:
- [member-delimiter-style](https://typescript-eslint.io/rules/member-delimiter-style) - This rule enforces semicolons (or a chosen delimiter) in TypeScript interfaces and `type` aliases, which would be consistent with us enforcing semicolons at the end of statements. Thanks to @HCY123902 for spotting this. [Example of what this rule will prevent](https://github.com/reposense/RepoSense/pull/1937#discussion_r1133650426)
- [type-annotation-spacing](https://typescript-eslint.io/rules/type-annotation-spacing) - This rule enforces the style of TypeScript type annotations, e.g. `foo: string` vs `foo : string` vs `foo :string`. As our codebase currently uses the first style, it would be good to enforce this for consistency and readability. [Example of what this rule will prevent](https://github.com/reposense/RepoSense/pull/1903#discussion_r1103161312)

I've labelled this 'to discuss' since I might have missed some rules. If anyone has any suggestions/knows of any rules that should be enforced, please feel free to add them here! Otherwise, as mentioned above, since it's quite difficult to tell what's missing, this might be something that we add on to occasionally when we run into such issues (i.e. we realize the linter doesn't pick up on a style issue).

vvidday:
Fixes #1945 

### Proposed commit message

```
Currently, our ESLint ruleset does not enforce certain formatting rules
for TypeScript code, such as delimiters in TypeScript interfaces and
spacing in type annotations.

Let's update the ESLint ruleset to check for proper delimiters and
spacing in type annotations. This will enforce consistency throughout
the codebase and improve code readability.
```

### Other information

Most of the information is in #1945. The first rule enforces semicolons as delimiters in TypeScript interfaces, and the second rule enforces the following format for type annotations: `foo: string` over `foo : string` or `foo :string`.

### DIFF
--- a/docs/dg/styleGuides.md
+++ b/docs/dg/styleGuides.md
@@ -124,8 +124,9 @@ public LocalDateTime parseDate(String dateString) throws NullPointerException, P
 }
 ```
 ## TypeScript specific formatting
+
 For TypeScript specific code, such as within an `interface` or type annotations, we also stipulate the following standards:
-- Use semicolons as delimiters for TypeScript interfaces and types.
+* Use semicolons as delimiters for TypeScript interfaces and types.
 
 Negative Examples:
 ```typescript
@@ -149,7 +150,7 @@ interface Foo {
     greet(): string;
 }
 ```
-- For type annotations, use a space after but not before.
+* For type annotations, use a space after but not before.
 
 Negative Examples:
 ```typescript
@@ -183,4 +184,3 @@ class Foo {
     name: string;
 }
 ```
-
--- a/docs/dg/styleGuides.md
+++ b/docs/dg/styleGuides.md
@@ -16,6 +16,7 @@ Our coding standards are mostly based on those at [se-education.org/guides](http
 * [**Markdown/MarkBind** coding standard](https://se-education.org/guides/conventions/markdown.html)
 * [**Java** coding standard](https://se-education.org/guides/conventions/java/index.html)
 * [**JavaScript** coding standard](https://se-education.org/guides/conventions/javascript.html)
+* **TypeScript**: In addition to the JavaScript coding standard, follow the [**recommended ESLint rules**](https://typescript-eslint.io/rules/) and the formatting rules [described below](#typescript-specific-formatting).
 * **Vue Components**: Follow the [**Vue style guide**](https://vuejs.org/style-guide/), up to the **Recommended** section.
 * **Documentation**: Follow the [**Google developer documentation style guide**](https://developers.google.com/style).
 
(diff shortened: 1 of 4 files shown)


ckcherry23 (approved):
LGTM

HCY123902 (approved):

zhoukerrr (approved):
LGTM

PR #1960, opened 2023-03-20
[#1959] Missing contribution bar for merged groups after refresh

### LINKED ISSUE #1959: Missing contribution bar for merged groups after refresh

1. Merge any summary chart group
2. Observe that the contribution bar is present
3. Refresh the page / Copy the URL and revisit the URL
4. Contribution bar is no longer present.

The contribution bar should remain there when visited again.

missing contribution bar

To run on the exact repo in the example above, run `java -jar RepoSense.jar --repos https://github.com/AY2122S1-CS2103T-W13-4/tp.git --since 17/09/2021 --until 19/11/2021 --view`

### Observations
Even after refreshing, the number of lines calculated is correct.
image

However, if we console.log the input (`totalContribution`) and output returned (`res`) for `getContributionBars(totalContribution)` these are the logs:

Before refreshing:
image

After refreshing
image

The cause of this issue is due `NaN` being returned instead. The reason why `NaN` instead of the array being returned is not clear and will require a deeper dive into the code.

MarcusTXK:
Fixes #1959

### Proposed commit message

```
Currently, upon refreshing the page, the contribution bar disappears 
due to getContributionBars returning NaN instead of the expected array. 
This was caused by division by 0, due to the user's 
checkedFileTypeContribution not being initialized properly. 

Let's fix this checkedFileTypeContribution not being initialized 
properly and add guard clauses against division by 0.
```

### Other information

This was likely introduced in PR #1903 when the initial value of `checkedFileTypeContribution` was refactored to be set as `0`, when it was previously expected to be `undefined`. This resulted in `this.updateCheckedFileTypeContribution(user)` not being called when the page was refreshed, and resulted in both `totalLines` and `totalCount` to be 0. There were no guard clauses for division of 0 which resulted in `NaN` being returned for `avgContributionSize()`, causing `cnt` to be `NaN` as well when it was divided by `contributionLimit `.


ckcherry23:
Hi @MarcusTXK, this is a great catch, and thanks for taking up the issue! 
However, it seems like the cypress test for the contribution bar is falling, can you take a look?

frontend/src/utils/api.ts
```diff
@@ -269,7 +269,7 @@ window.api = {
           searchPath: searchParams.join('_').toLowerCase(),
           repoName: `${repo.displayName}`,
           location: `${repo.location.location}`,
-          checkedFileTypeContribution: 0,
+          checkedFileTypeContribution: -1,
```
zhoukerrr, line 272: I am just thinking if it will be simpler to just update the type and set the initial value to undefined?

vvidday (commented):
The changes LGTM, thanks for catching this! Do you think it will be helpful to add a cypress test case to make sure this functionality doesn't break in the future?

MarcusTXK:
> The changes LGTM, thanks for catching this! Do you think it will be helpful to add a cypress test case to make sure this functionality doesn't break in the future?

I think that is a great idea to help detect such issues. Would it be a good idea to add this as a separate issue?

vvidday:
> > The changes LGTM, thanks for catching this! Do you think it will be helpful to add a cypress test case to make sure this functionality doesn't break in the future?
> 
> I think that is a great suggestion to help detect such issues in advance in the future. Would it be a good idea to add this as a separate issue?

Yup I think it will be good to add it as a separate issue, especially since this is a bug fix and not a new feature

frontend/src/utils/api.ts
ckcherry23, line 272: Any reason we are using `-1` instead of `undefined` here as the default value? I feel that `undefined` or maybe `null` would be more intuitive for developers.

frontend/src/components/c-summary-charts.vue
```diff
@@ -318,7 +322,7 @@ export default {
               contributionBars.push(fullBarWidth);
             }
             const remainingBarWidth = barWidth % fullBarWidth;
-            if (remainingBarWidth !== 0) {
+            if (remainingBarWidth >= 0) {
```
HCY123902, line 325: Should it be a `>` or `>=` here

frontend/src/views/c-summary.vue
```diff
@@ -201,7 +201,7 @@ export default {
       let totalCount = 0;
       this.repos.forEach((repo) => {
         repo.users.forEach((user) => {
-          if (user.checkedFileTypeContribution === undefined) {
+          if (user.checkedFileTypeContribution === -1) {
```
HCY123902, line 204: I also think it might be more appropriate to replace `-1` with values such as `undefined` as default. Can you take a look again

HCY123902 (changes requested):

frontend/src/components/c-summary-charts.vue
MarcusTXK: Good catch, let me change this.

frontend/src/utils/api.ts
MarcusTXK: The initial reason was so I did not want to deal with `undefined`, however I do not have a strong opinion on this. Let me update it to `undefined`.

HCY123902 (approved):

zhoukerrr (approved):
LGTM. It was a good catch. Just remember to add a test on this to make sure we do not repeat the same bug in future.

dcshzj:
@MarcusTXK Are you able to resolve the issues with the `undefined` type?

dcshzj (approved):

PR #1550, opened 2021-06-18
[#1212] RepoLocation: Remove redundant backslashes in regex

aidoxe-123:
Resolves #1212 

Commit message:
```
RepoLocation: Remove redundant backslashes in regex

In RepoLocation.java, our current regex is 
"^https?:\\/\\/github.com\\/(?<org>.+?)\\/(?<repoName>.+?)\\.git$".

There is no need to escape the forward slash "/" character with the
"\\" characters, as the forward slash does not have a special meaning
when used on its own.

Let's remove the redundant backslashes and add more tests for
different types of repository URLs.
```

PR #1961, opened 2023-03-22
[#528] Allow multiple repos in one row for author-config.csv

### LINKED ISSUE #528: author-config.csv: allow multiple repos in one row

sikai00:
Fixes #528

### Proposed commit message
```
The same author may be working on several repositories to be analyzed.
In order to specify author information through author-config.csv, each
row has to be duplicated for each repository. This may be tedious for
the person generating the RepoSense report.

This change allows the repository location column in author-config.csv
to accept multiple repositories through the separator delimiter ";". It
also allows users to specify branches for each repository by using "~"
after the repository and separate multiple branches with "|", This
removes the need to duplicate the remaining columns' information in
multiple rows and reduce likelihood of errors in maintenance.

Let's allow each row in author_config.csv to contain multiple repos and
multiple branches for each repo to reduce row duplication.
```

#### Only GitHub branch URLs are accepted for now
It is however possible to include different variants of such remote source code repository in the future.

#### Explanation of implementation
In `AuthorConfigCsvParser::processLine`, the repository location column is now returned as a List. We use a for-loop on this list to process each of the repo location (possibly with their new delimiter syntax). Other columns are held the same per line. Branch column is now used as a default value if delimiter syntax/GitHub branch URL is not provided.

`AuthorConfigLocationParser::parseLocation` provides a mean of parsing the new syntax. It returns a 2 element array of {location, branch} which is then used in `processLine`. It returns branch as null if no branch is specified through the special syntax. It is supported by `LocalRepoLocationParser.java` and `RemoteRepoLocationParser.java` which parses the special syntaxes accordingly.

Erroneous location checking is still handled in `RepoLocation.java` rather than being handled in the parsers early as it made more sense to not increase class dependencies in the parsers, and also since it gets the job done properly.


damithc:
@sikai00 thanks for taking this up.
As we are aiming to reduce duplication, something like `https://github.com/reposense/RepoSense.git#release|master` to indicate multiple branches for the same repo? Would that be going too far?
Also, this needs to be explained as an advanced feature, separately from the documentation of basic usage.

sikai00:
> something like `https://github.com/reposense/RepoSense.git#release|master` to indicate multiple branches for the same repo? Would that be going too far?

Prof @damithc I believe that should be possible. I can use another delimiter to indicate the multiple branches, though I don't think I will be using "#" since it is a possible character in a URL which can make things complicated. Thanks for the suggestion!

> Also, this needs to be explained as an advanced feature, separately from the documentation of basic usage.

I will update the documentation accordingly then

sikai00:
Update: I have implemented the ability to list out multiple branches per repository, as well as update the documentation into its own page

src/main/java/reposense/util/SystemUtil.java
```diff
@@ -12,6 +15,18 @@ public static boolean isWindows() {
...
+     * Returns true if the given String is a valid URL.
+     */
+    public static boolean isValidUrl(String possibleUrl) {
+        try {
+            new URL(possibleUrl);
+        } catch (MalformedURLException mue) {
+            return false;
+        }
+        return true;
+    }
```
MarcusTXK, line 28: Interesting way to determine if URL is valid, was at first hesitant of using try-catch to determine that, but it seems to be a common way to do it.

docs/ug/authorConfigSyntax.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### Specifying multiple `Repository's Location` with multiple specific branches
+
+To list out multiple branches for a particular `Repository's Location`, we can use a pipe `|` separator to demarcate the different branches.
+
+Below is an example:
+
+| Repository's Location                                                | Branch   | Author's Git Host ID | ... Hidden columns    |
+|----------------------------------------------------------------------|----------|----------------------|-----------------------|
+| https://github.com/reposense/RepoSense.git~master\|release\|cypress; | master   | sikai00              | --                    |
```
MarcusTXK, line 81: The table here seems to be broken:

image

MarcusTXK (approved):
LGTM. Tested it with various examples and seems to work as documented, great job adding this feature! 

Overall this feature is definitely a useful addition to RepoSense to reduce the length of the csv config. I like the approach taken to implement this feature and I agree with your approach of letting Erroneous location checking is still be handled in `RepoLocation.java`. 

For the documentation, I am not sure if this should be a separate page as `Appendix: author-config.csv advanced syntax`, or if it would be better to include it within the `Config files format` page together with the documentation on the `author-config.csv`. If its a seperate page, I think it would be great to add a link to the `Appendix: author-config.csv advanced syntax` page from the `Config files format` page in the `author-config.csv`, perhaps with `{{ embed("Appendix: **author-config.csv advanced syntax**", "authorConfigSyntax.md") }}` or something similar.

src/main/java/reposense/parser/AuthorConfigLocationParser.java
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public static List<String> parseLocation(String location, String defaultSpecifiedBranch) {
+        if (isRemoteRepo(location)) {
+            return parseRemoteRepoLocation(location, defaultSpecifiedBranch);
+        } else {
+            return parseLocalRepoLocation(location, defaultSpecifiedBranch);
+        }
+    }
+
+    private static boolean isRemoteRepo(String location) {
```
chan-j-d, line 32: Does this actually work? I don't think `|` is a valid URL character so a remote repo with multiple branches specified fails this unless I'm understanding wrongly. It also ignores remote repos specified by the SSH protocol. I think we should try a more agnostic implementation that does not care whether it is remote or local

src/main/java/reposense/parser/LocalRepoLocationParser.java
```diff
@@ -0,0 +1,88 @@
...
+     * using the delimiter syntax.
+     */
+    private static boolean isUsingDelimiter(String locationAndBranch) {
+        if (pathExists(locationAndBranch)) {
+            return false;
+        }
+
+        return locationAndBranch.lastIndexOf(LOCATION_BRANCHES_DELIMITER) != -1;
+    }
+}
```
chan-j-d, line 88: I think a more clear separation between the two types of parser (remote vs local) would be 
1) domain specific branch URL (e.g. the github branch url)
2) multiple remote/local repo possibly with branch specification

And you can probably handle the checking like this: Check if the path exists (doesn't matter if local or remote), if so then its standalone (neither 1 or 2). Otherwise, if `|` or `;` detected, it's case 2. Otherwise try to fit the specified location into 1). If it fails, then treat the argument as a standalone repo location. 

This idea somewhat assumes that we don't need to use the `~` separator

src/main/java/reposense/parser/AuthorConfigLocationParser.java
chan-j-d: https://stackoverflow.com/a/3931701 Following this discussion and the comments below and Java 8's own URL API, I don't actually think the validation does all that much besides checking if a protocol is specified

chan-j-d (changes requested):
Thanks for doing this up. It's a really helpful feature to implement. I haven't quite reviewed the tests yet but wanted to discuss about the implementation regarding how we want to read repo locations and branches since I think the implementation can be simplified. I think we have to be very careful in trying to do any sort of remote vs local detection especially when they are parsed differently because it's really not an easy task

docs/ug/authorConfigSyntax.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### Specifying multiple `Repository's Location` with a single specific branch
+
+To override the Branch column with a specific branch for a particular `Repository's Location`, we can use a tilde `~` separator to denote our branch.
+
+Below is an example:
+
+| Repository's Location                                                                       | Branch  | Author's Git Host ID | ... Hidden columns    |
+|---------------------------------------------------------------------------------------------|---------|----------------------|-----------------------|
+| https://github.com/reposense/RepoSense.git~release;https://github.com/MarkBind/markbind.git | master  | sikai00              | --                    |
```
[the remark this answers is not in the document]
sikai00: That's a good suggestion, I hadn't thought of doing it that way

src/main/java/reposense/parser/LocalRepoLocationParser.java
sikai00: Yes I believe this will be far more ideal. I was initially too focused on separating the local and remote repos rather than whether it is necessary, but an agnostic implementation would work just fine and better in terms of justifying the code.

docs/ug/authorConfigSyntax.md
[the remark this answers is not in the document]
sikai00: Actually, I believe I had a concern while doing this: It becomes difficult to tell if `|` is part of the actual file path, or whether it is being used as a delimiter.

If the author happens to have `|` as part of their file path, but also wants to use the delimiter, it becomes impossible to tell which is the branch and which is the actual file path.

My current implementation does this by only getting the very last delimiter using the particular syntax which I believe is more ideal, since the author is likely to use the `~` and `|` delibrately for the location.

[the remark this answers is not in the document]
chan-j-d: Ok I actually read the original post examples and didn't verify if the code worked like that but for some reason I had the impression that `~` was used only for remote repos (see example 4). Which is also why I was suggesting standardizing the remote vs local since I thought they were parsed separately as well.

Since that is the case, I do think it makes some sense to look for the last `~` and then check if the entire thing is a valid repo filepath, otherwise split at the last `~` since branch names can't have `~`. There is still technically one scenario that is indistinguishable (in your example) both repos existing
```
/User/sikai00/actual|filepath/RepoSense~master|cypress 
/User/sikai00/actual|filepath/RepoSense
```
but I think it's improbable enough to not be of concern. Overall though, I think we should still look to merge remote and local parsing into one.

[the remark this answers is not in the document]
sikai00: I see, in that case I suppose the examples can be better specified then. I will most likely be merging the remote and local parsing into one, as I do believe that will be more ideal.

> There is still technically one scenario that is indistinguishable (in your example) both repos existing

I am not sure on what you mean with this. Do you mean having both as Repository's Location as separate rows in `author-config.csv`?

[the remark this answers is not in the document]
chan-j-d: The user could in theory have a repository located in the coincidentally-named directory `/User/sikai00/actual|filepath/RepoSense~master|cypress` or he could have one in `/User/sikai00/actual|filepath/RepoSense` with branches `master` and `cypress` he wishes to analyze. 
If they specify one line of `/User/sikai00/actual|filepath/RepoSense~master|cypress` in the csv, I don't think it's possible to detect which one the user 'intended' though I don't think this is likely to occur.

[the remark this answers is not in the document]
sikai00: I see, that is indeed a really obscure and improbable case. I currently don't have any good ideas on how this might be resolved actually. An possible solution might be to add a new column just to denote that the syntax is being used, but this affects usability since the user will need to indicate in a new column which goes against the purpose of the PR. Do you think it'll be good to add a small note, or should I disregard it entirely? I am fine with either cases, though I do believe that the note is a bit of an overkill, but at the same time it doesn't sit right to leave it as is.

sikai00:
> For the documentation, I am not sure if this should be a separate page as `Appendix: author-config.csv advanced syntax`, or if it would be better to include it within the `Config files format` page together with the documentation on the `author-config.csv`. If its a seperate page, I think it would be great to add a link to the `Appendix: author-config.csv advanced syntax` page from the `Config files format` page in the `author-config.csv`, perhaps with `{{ embed("Appendix: **author-config.csv advanced syntax**", "authorConfigSyntax.md") }}` or something similar.

Thank you for your review! I believe prof mentioned to add it as a separate page. I think I will use the embed in `configFile.md`.

src/main/java/reposense/parser/AuthorConfigLocationParser.java
sikai00: Thank you for spotting this! You are right, the code seems to be defaulting the remote repos into the local repos in this case. Given that both local and remote can be handled the same way, I have taken your advice to remove the separation in the most recent commit.

docs/ug/authorConfigSyntax.md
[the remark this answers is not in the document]
chan-j-d: I think it's not something we have to handle because it just seems too improbable. Perhaps you could add a comment near the parsing with the link to this thread detailing the possible loophole in parsing repo + branches in case it ever becomes an issue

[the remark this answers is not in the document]
sikai00: Completed in the most recent commit.

src/main/java/reposense/parser/AuthorConfigLocationParser.java
```diff
@@ -0,0 +1,130 @@
...
+                .substring(lastLocationBranchesDelimiterIndex + 1).split(SPLIT_SAFE_BRANCHES_DELIMITER);
+        for (String branch : branches) {
+            parsedLocationAndBranches.add(branch);
+        }
+
+        if (branches.length == 0) {
+            // by default, if no branches are specified, an empty branch is returned since
+            // the only case where branches.length is 0 is when delimiter is used but no
+            // branch is specified.
+            parsedLocationAndBranches.add("");
```
chan-j-d, line 98: I was wondering why codecov didnt detect this so I went to check. As it turns out, even if no branches are specified, after the split, the last entry will be an empty string so the previous for loop adds an empty string. So I don't actually think this part is necessary

chan-j-d (changes requested):
Some very small changes. The tests look really good and comprehensive.

src/main/java/reposense/parser/AuthorConfigLocationParser.java
sikai00: Strangely enough, I recall I had a particular edge case that I was trying to resolve here, but I am unable to recall the exact case here. I will remove it for the time being, at least until if this edge case ever shows up (which is very unlikely).

chan-j-d (changes requested):
PS one last small change

src/main/java/reposense/parser/AuthorConfigCsvParser.java
```diff
@@ -60,38 +60,46 @@ protected String[][] optionalHeaders() {
...
+                author.setDisplayName(!displayName.isEmpty() ? displayName : author.getGitId());
+                if (!aliases.isEmpty()) {
+                    author.setAuthorAliases(aliases);
+                }
+                if (!ignoreGlobList.isEmpty()) {
+                    author.setIgnoreGlobList(ignoreGlobList);
+                }
+
+                config.addAuthor(author);
+            }
```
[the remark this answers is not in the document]
sikai00: Apologies for misunderstanding that! I have fixed it in the most recent commit

chan-j-d (approved):
LGTM!

docs/ug/authorConfigSyntax.md
```diff
@@ -0,0 +1,80 @@
...
+
+#### Specifying `Repository's Location` with one or more specific branches
+
+To list out one or more branches for a particular `Repository's Location`, we can use a pipe `|` separator to demarcate the different branches.
+
+Below is an example:
+
+| Repository's Location                                                | Branch   | Author's Git Host ID | ... Hidden columns    |
+|----------------------------------------------------------------------|----------|----------------------|-----------------------|
+| https://github.com/reposense/RepoSense.git~master\|release\|cypress  | master   | sikai00              | --                    |
```
yhtMinceraft1010X, line 39: You may want to mention in the UG which column takes precedence if branch names are given in both the location and branch columns.

yhtMinceraft1010X (changes requested):

chan-j-d:
I forgot to add that we might want to simplify the commit message. I think it can be something like 
- first paragraph: problem (i.e. your first 2 paragraphs now)
- second paragraph: solution 3rd, 4th and 5th combined 
- Last line: short call to action, "let's ..." in like one sentence

Don't think it's necessary to include examples.

sikai00:
> I forgot to add that we might want to simplify the commit message. I think it can be something like
> 
> * first paragraph: problem (i.e. your first 2 paragraphs now)
> * second paragraph: solution 3rd, 4th and 5th combined
> * Last line: short call to action, "let's ..." in like one sentence
> 
> Don't think it's necessary to include examples.

I have simplified the commit message to fit the requirements. Do let me know if there is any further changes required!

chan-j-d:
@sikai00 I think I would do it something like this
```
The same author may be working on several repositories to be analyzed.
In order to specify author information through author-config.csv, each
row has to be duplicated for each repository. This may be tedious for the 
person generating the RepoSense report.

This change allows the repository location column in author-config.csv to
accept multiple repositories through the separator delimiter ";". It also 
allows users to specify branches for each repository by using "~" after
the repository and separate multiple branches with "|", This removes the 
need to duplicate the remaining columns' information in
multiple rows and reduce likelihood of errors in maintenance.

Let's allow each row in author_config.csv to contain multiple repos 
and multiple branches for each repo to reduce row duplication.
```

I realise that you're probably following the style guide from the se-education document that is linked in our developer workflow. Honestly I'm not sure if there's an authoritative source on this since I've been around for awhile and it's just what I've seen, but I actually followed this comment https://github.com/reposense/RepoSense/pull/1550#pullrequestreview-695636319 from a long time ago. Basically, the last call-to-action is generally quite short.

sikai00:
> @sikai00 I think I would do it something like this
> 
> ```
> The same author may be working on several repositories to be analyzed.
> In order to specify author information through author-config.csv, each
> row has to be duplicated for each repository. This may be tedious for the 
> person generating the RepoSense report.
> 
> This change allows the repository location column in author-config.csv to
> accept multiple repositories through the separator delimiter ";". It also 
> allows users to specify branches for each repository by using "~" after
> the repository and separate multiple branches with "|", This removes the 
> need to duplicate the remaining columns' information in
> multiple rows and reduce likelihood of errors in maintenance.
> 
> Let's allow each row in author_config.csv to contain multiple repos 
> and multiple branches for each repo to reduce row duplication.
> ```
> 
> I realise that you're probably following the style guide from the se-education document that is linked in our developer workflow. Honestly I'm not sure if there's an authoritative source on this since I've been around for awhile and it's just what I've seen, but I actually followed this comment [#1550 (review)](https://github.com/reposense/RepoSense/pull/1550#pullrequestreview-695636319) from a long time ago. Basically, the last call-to-action is generally quite short.

I see, I have updated the commit message as per your comment. Thanks for informing me on the commit message style!

yhtMinceraft1010X (approved):
LGTM

PR #1965, opened 2023-03-27
[#1963] Migrate c-zoom.vue to TypeScript

### LINKED ISSUE #1963: Migrate `c-zoom.vue` to TypeScript
Part of #1936 

Migrate `c-zoom.vue` to TypeScript

ckcherry23:
Fixes #1963 

### Proposed commit message
```

Let's migrate one of the main components, c-zoom.vue, to TypeScript.
In addition to providing greater type safety, the migration of this
component will also allow the migration of its child components, such
as c-ramp.vue.
```

### Other information
The `SortingFunction<T>` takes an object `T` and returns the value of `T.attribute` to be sorted on instead of `T` itself.

However, I think only `string` and `number` types are comparable (?), so I created a `ComparablePrimitive` type that is a union of both of these types. Hence, the sorting function will only return a value that can be compared by using `ComparablePrimitive` as its return type.

* [x] Enum for `commitsSortType`


MarcusTXK (approved):
LGTM, nice job for the migration

frontend/src/views/c-zoom.vue
```diff
@@ -317,8 +331,8 @@ export default {
 
     retrieveSortHash() {
```
HCY123902, line 332: Does the set on line 314 needs to be typed as well

vvidday (changes requested):
Thanks for your work on this! The existing changes look good.

Currently TypeScript can't infer the types of the computed properties from the store (`this.info` and `this.fileTypeColors` are both `any`) - I think changing the code for the `...mapState` computed property like so will let TypeScript infer the types properly:
```typescript
    ...mapState({
      fileTypeColors: (state: unknown) => (state as StoreState).fileTypeColors,
      info: (state: unknown) => (state as StoreState).tabZoomInfo,
    }),
```
This produces some errors (mainly having to do with `zUser` being possibly undefined) which can be solved by null checks/assertions.

Also, there will be some errors regarding the `addHash` calls in the `setInfoHash()` method due to it only accepting `string`s as the value, I think this part should be addressed in my changes in #1953.

ckcherry23:
@vvidday Thanks for your suggestions! I have made these changes.

I have a question about your implementation though. In `vuex.d.ts`, is there a reason you have used the interface `User` in `types` over the class `User` in `utils` for the `ZoomInfo` interface? 
I changed it to the class so that the commits attribute is not optional, but please let me know your reason!

vvidday:
> @vvidday Thanks for your suggestions! I have made these changes.
> 
> I have a question about your implementation though. In `vuex.d.ts`, is there a reason you have used the interface `User` in `types` over the class `User` in `utils` for the `ZoomInfo` interface? I changed it to the class so that the commits attribute is not optional, but please let me know your reason!

I think the main reason was that the existing code that generates `ZoomInfo` didn't use the User class https://github.com/reposense/RepoSense/blob/master/frontend/src/views/c-summary.vue#L842 (didn't call `new User(...)`), so I went with the interface instead.

Although now that you mentioned it, I'm not sure if the `commits` attribute should be optional - I will have to do more investigation about this. If it is safe to make the `commits` attribute of the `User` interface non-optional, then it might make sense to get rid of one of the two since they would be equivalent (except one is a TS interface and one is a class). Will do more investigation on this issue and update when I get the time - posting now just to update my plans.

Potentially useful: https://stackoverflow.com/questions/51716808/when-use-a-interface-or-class-in-typescript
https://stackoverflow.com/questions/40973074/difference-between-interfaces-and-classes-in-typescript

vvidday:
Just did some more investigation and it looks safe to make the `commits` attribute of the `User` interface (in `types`) non-optional. 

I originally set it to optional when first defining the types in #1903 because I was basing it off the `User` class, and the [original code created the `User` object with no commits field](https://github.com/vvidday/RepoSense/blob/04f33372caef16dfa3bfe6a2be82c2d5e9b61d68/frontend/src/utils/api.ts#L258), but when I later refactored the function to use the `User` type instead of `any`, I [initialized the `commits` attribute to a default empty array](https://github.com/reposense/RepoSense/pull/1903/commits/e4d202b08fdb3021f72dbe2fab14b9a392231d74) but mistakenly forgot to change the optional property on the `commits` attribute. Essentially, because of the default initialization, the `commits` attribute will always be defined and we can define it as non-optional.

This makes the `User` in `types/types.ts` basically equivalent to `utils/User.ts`, except the former is a TS interface while the latter is a JS class. They are mostly equivalent, with the difference being that the class still exists at runtime while the interface is only used for type-checking at compile time, and erased at runtime when the code is transpiled to JS. (Hence there's slightly less overhead using the interface). 

From some reading (e.g. the two StackOverflow threads I linked above), I feel that we don't really utilize the benefits of a JS class as there's no class methods, or any other behavior that's not already achieved by the interface (at least from what I can see). @ckcherry23 would love to get your thoughts on whether you think the class is necessary/beneficial here?

Otherwise if they are equivalent, I think it would be worth to remove either the interface or the class, so that we don't have two separate things with identical names that serve the same purpose (and avoiding confusing imports like [this](https://github.com/reposense/RepoSense/blob/master/frontend/src/types/window.ts#L3)). I tried removing the User class and replacing the references with the User interface in this branch: https://github.com/vvidday/RepoSense/tree/remove-user-class and it seems to be working fine.

This removal, if we go ahead with it, can be done in a separate PR.

HCY123902 (approved):
I will approve it for now. The changes regarding removing the `User` class with the interface can be done in a separate pull request

frontend/src/types/zoom.ts
```diff
@@ -0,0 +1,6 @@
+enum CommitsSortType {
```
zhoukerrr, line 1: I like this approach!

zhoukerrr (approved):
LGTM. Agree with @HCY123902 that a new PR can be made to address the problem.

vvidday (approved):

dcshzj (changes requested):
Will merge once the latest comment is resolved.

dcshzj (approved):
LGTM! Are you able to resolve the file conflicts too?

PR #1969, opened 2023-04-02
[#1936] Migrate c-authorship.vue to TypeScript

vvidday:
Part of #1936 

### Proposed commit message

```

Let's migrate one of the main components, c-authorship.vue, to
TypeScript. This will provide better type safety and reduce runtime
errors.
```

### Other information

Although this migration is fairly large, there aren't any logic changes and the PR mostly contains addition of types & minor refactoring to support said types. 

These are the notable changes:
### Removal of `hasCommits()` method
This PR [removed](https://github.com/reposense/RepoSense/commit/897c5fb116ea9ff51d6bd2275238ce5b966547ec#diff-892687ea1756a0d02b6050970127e76f77e5a10d6ab478f4bccd6692ea6dfd92L508) the `hasCommits()` method. This method is not used anywhere in the codebase. It was written [4 years ago](https://github.com/reposense/RepoSense/blame/8f2241733ae9f254baf50c34f88b71c55171aca4/frontend/src/views/c-authorship.vue#L508), but somewhere along the line the call to the method was removed but the method remained. The reason I had to remove it is because it references a non-existent property (`repo.commits.authorFinalContributionMap`) - this property [does not exist on the object](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/report/CommitReportJson.java#L43).

### Duplicate interface/class
Similar to the `User` interface and class that was [discussed in #1965](https://github.com/reposense/RepoSense/pull/1965#issuecomment-1488698009), there is a duplicate definition between the [`AuthorshipFileSegment` interface](https://github.com/reposense/RepoSense/commit/a4181ec1f8388ed205b52cc3a6ca8b70248f384f) in `types.ts` and the [`Segment` class](https://github.com/reposense/RepoSense/blob/master/frontend/src/utils/segment.ts). Since the context is identical (duplicate definitions, class only contains attributes and there's no class methods etc), we can address this in a separate PR, as discussed in #1965.

### `fileSize` attribute potentially redundant
The [`fileSize` attribute](https://github.com/reposense/RepoSense/commit/897c5fb116ea9ff51d6bd2275238ce5b966547ec#diff-892687ea1756a0d02b6050970127e76f77e5a10d6ab478f4bccd6692ea6dfd92R592) appears to be redundant. In the current codebase, [this is the only time it is referenced](https://github.com/reposense/RepoSense/blob/master/frontend/src/views/c-authorship.vue#L600), and since `file.fileSize` does not exist*, its value is always `undefined` (can be verified by opening Vue devtools and scrolling through the `files` array). ~~We can look to remove this, but since this does not concern TypeScript migration, it can be done in a separate PR.~~ Update: The assignment of this attribute, along with its type definition, has been removed in this PR due to the above reason.

*It is [not exported by the backend](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/authorship/model/FileResult.java#L14) when generating the authorship.json report. 

### `@types/minimatch`
The PR installs the above library as a dev dependency as it provides official typings for the `minimatch` library that we use in `c-authorship`. See https://www.npmjs.com/package/@types/minimatch


ckcherry23 (commented):
The `c-authorship` file migration looks excellent to me! I haven't really found any places where you missed out on adding the type. 

However, we may want to consider whether we want to restrict the return types of our functions to ensure the correct type is returned in functions with multiple return statements for example. Though the compiler may be able to infer what type we are returning, it may not be the type we intended. Hence, this will help us to avoid developer errors though it would make the code more verbose.

ckcherry23 (approved):

vvidday:
> The `c-authorship` file migration looks excellent to me! I haven't really found any places where you missed out on adding the type.
> 
> However, we may want to consider whether we want to restrict the return types of our functions to ensure the correct type is returned in functions with multiple return statements for example. Though the compiler may be able to infer what type we are returning, it may not be the type we intended. Hence, this will help us to avoid developer errors though it would make the code more verbose.

Great point! I originally didn't include explicit return type annotations as TypeScript can infer types, but as you pointed out, this could potentially result in incorrect return types (due to developer error) being 'accepted'. I think among the functions, the ones that are most important to annotate are the functions that have a non-void return type.

I'll update the function return types for this file and future files, but for previous files we could potentially add it as a first-time issue for new contributors, as it is relatively straightforward but will also give good exposure to the codebase (as it requires going through all the methods in the component and figure out how and what data flows between the functions).

HCY123902 (approved):

frontend/src/views/c-authorship.vue
```diff
@@ -591,15 +583,16 @@ export default {
...
-        out.fileSize = file.fileSize;
-        out.isIgnored = !!file.isIgnored;
-        out.isBinary = !!file.isBinary;
+        const out: AuthorshipFile = {
+          path: file.path,
+          lineCount: lineCnt,
+          active: lineCnt <= SINGLE_FILE_LINE_COUNT_THRESHOLD && !file.isBinary,
+          wasCodeLoaded: lineCnt <= SINGLE_FILE_LINE_COUNT_THRESHOLD,
+          fileType: file.fileType,
+          fileSize: undefined, // is always undefined, to potentially remove
```
HCY123902, line 592: If it is always undefined, can you check whether it is safe to remove it within this pull request

frontend/src/types/authorship.ts
```diff
@@ -0,0 +1,11 @@
+export enum FilesSortType {
+  LineOfCode = 'lineOfCode',
```
HCY123902, line 2: Just a grammar check. Will it be more suitable to use the plural form `lines` instead

HCY123902 (changes requested):

frontend/src/types/authorship.ts
vvidday: I agree with this, I'll change it to `lines`. It's also similar [in c-zoom](https://github.com/reposense/RepoSense/pull/1965/files#diff-f1f207cd3bfc1e4ba3264657c296cc74bdf556265c4879b07ca82d032f9a86c9R3), should we change it in #1965 too?

HCY123902: yeah sure. Please go ahead

vvidday: @ckcherry23 thanks!

HCY123902 (approved):

frontend/src/views/c-authorship.vue
```diff
@@ -437,7 +442,7 @@ export default {
       }
       if (repoCache.length === 2) {
         const toRemove = repoCache.shift();
-        if (toRemove !== this.info.repo) {
+        if (toRemove !== this.info.repo && toRemove !== undefined) {
```
zhoukerrr, line 445: 2 minor points here.

1. you can simply use `if (toRemove)` to check if it is undefined
2. It is always a better practice to check for undefined first

frontend/src/views/c-authorship.vue
```diff
@@ -627,7 +619,7 @@ export default {
 
       res.sort((a, b) => b.lineCount - a.lineCount).forEach((file) => {
         // hide files over total char count limit
-        if (!file.isIgnored && !file.isBinary && file.active) {
+        if (!file.isIgnored && !file.isBinary && file.active && file.charCount) {
```
zhoukerrr, line 622: May I check why is `file.charCount` added to the condition check? If `file.charCount` is 0, it will not enter the if branch. This is a little different from the original logic.

zhoukerrr (changes requested):
@vvidday Will appreciate if you can answer the clarification I have. Thank you!

frontend/src/views/c-authorship.vue
vvidday: Thanks for the pointers, will make the changes & keep this in mind in the future!

frontend/src/views/c-authorship.vue
vvidday: The check was added because in the following line we check `file.charCount <= SINGLE_FILE_CHAR_COUNT_THRESHOLD`, but `charCount` is an optional field.
Thanks for pointing this out, here I think we have to do an explicit `file.charCount !== undefined` check to still allow the code to enter the branch if `file.charCount` is 0 like you mentioned.

If we check `file.charCount !== undefined` it technically won't be changing any logic, since [charCount is only set when `!file.isIgnored && !file.isBinary`.](https://github.com/reposense/RepoSense/blob/897c5fb116ea9ff51d6bd2275238ce5b966547ec/frontend/src/views/c-authorship.vue#L603), so it will always be true.

zhoukerrr (approved):
LGTM!

Issue #1966, opened 2023-03-27
Add cypress test for merged groups


MarcusTXK:
### What feature(s) would you like to see in RepoSense

Adding a cypress test for testing merged groups when the page is refreshed will help in early detection of issues such as those in  #1960

MarcusTXK:
Strangely this issue was not automatically closed when #1970 was merged in. I am manually closing this as this issue has been completed.

PR #1970, opened 2023-04-02
[#1966] Add cypress test for merged groups contribution bars

### LINKED ISSUE #1959: Missing contribution bar for merged groups after refresh

MarcusTXK:
Fixes #1959 

### Proposed commit message

```
Currently, there are no tests to ensure that the length and the number 
of contribution bars for merged groups are correct. This resulted in 
issue #1959, where the contribution bar went missing, being undetected. 

Let's add a cypress test to check that the length and the number of 
contribution bars for merged groups are correct, even after refresh.
```

### Other information

The cypress test checks if the widths of `summary-chart__contrib--bar` are 100%, 100% and 50%.
image

Note: This should not be merged before #1960, as the cypress test will fail until the bug fix is merged in, below is the current `summary-chart__contrib--bar` due to the bug on reload:
image


frontend/cypress/tests/chartView/chartView_mergeGroup.cy.js
```diff
@@ -54,4 +54,45 @@ describe('merge group', () => {
...
+      .should('have.length', 3)
+      .then(($bars) => {
+        // calculate the percentage of the width relative to the parent container
+        const parentWidth = $bars.eq(0).parent().width();
+        const width1 = (parseFloat(window.getComputedStyle($bars[0]).width) / parentWidth) * 100;
+        const width2 = (parseFloat(window.getComputedStyle($bars[1]).width) / parentWidth) * 100;
+        const width3 = (parseFloat(window.getComputedStyle($bars[2]).width) / parentWidth) * 100;
+
+        // assert that the widths are close enough to 100% and 50%
+        expect(width1).to.be.closeTo(100, 1);
```
vvidday, line 75: I was wondering if instead of hardcoding the values, we could record the values of the number of `.summary-chat__contrib--bar` elements and their respective widths before reloading, then checking if they are the same after reload? So we can avoid hardcoding the values which might potentially change in the future.

vvidday (commented):
Thanks for your work on this @MarcusTXK! The test cases look good overall but I just had a question regarding the hardcoding aspect of it.

frontend/cypress/tests/chartView/chartView_mergeGroup.cy.js
MarcusTXK: Great idea, I disliked how hardcoded it was as well, let me modify it to be less hardcoded.

MarcusTXK:
I have updated the cypress test to include a non-hardcoded version, when refreshing and checking if there are the same number of contribution bars and if they are the same width.

However, I left a test checking if the width and the number of contribution bars expected, as I realised that there is currently no test for the contribution bars and I think it would be good to be able to test that the calculation is correct. However this does require a slight hardcoding.

frontend/cypress/tests/chartView/chartView_mergeGroup.cy.js
```diff
@@ -54,4 +54,59 @@ describe('merge group', () => {
       .should('be.visible')
       .should('be.disabled');
   });
+
+  it('should have the correct number of merge group contribution bars and correct length', () => {
+    cy.get('#summary label.merge-group > input:visible')
```
vvidday, line 59:
```suggestion
    // Assumption: The number of merge group contribution bars is 3 and the width of the third bar is 50%.
    cy.get('#summary label.merge-group > input:visible')
```

vvidday (commented):
Thanks for the changes, the tests LGTM! Just thought it would be better to include a small comment about the assumption regarding the hardcoded values, as in other similar test cases: https://github.com/reposense/RepoSense/blob/master/frontend/cypress/tests/zoomView/zoomView_selectFileTypes.cy.js#L52

HCY123902 (approved):

zhoukerrr (approved):
LGTM. More testings are always good

PR #1974, opened 2023-04-04
[1971] Add Vue router

### LINKED ISSUE #1971: Add Vue Router for RepoSense

Let's add Vue Router to the frontend to allow for routing.

This is as part of #1894, and involves adding Vue Router for RepoSense. The benefits of including VueRouter are that in the future, more routes can easily be added (e.g. widgets) and it will also potentially help solve #642 , allowing a user to navigate back upon clicking options.

MarcusTXK:
Fixes #1971

### Proposed commit message

```
Currently, the RepoSense frontend does not have a router as everything 
is in the root page. In the future, we might add other pages or other 
functionalities, such as widgets, which might require routing. Adding a 
router also allows for easier backward navigation, which can allow a 
user to click back to navigate to a previous view after changing 
something on the frontend. 

Let's add a Vue Router to enable routing and history management in the 
RepoSense frontend.
```

#### Main changes
The change mainly involves adding a Vue Router, and shifting everything that was formerly in `app.vue` into `home.vue`.

#### Reason for changing `getFontColor` to an exported function
I was stuck with a bug where the code panel would not render the selected files, even though the selectedfiles were present in the component's data. 
image
This was in the console.
image
After spending along time trying to debug it, it turned out that `getFontColor` was `undefined`. I thus had to fix it by changing `getFontColor` to be an exported function instead of a global function set in `window`. I suspect that somehow with the addition of the router, the global function `getFontColor` was not initialised before the components get rendered, which caused this bug.


frontend/cypress/tests/chartView/chartView_percentage.cy.js
```diff
@@ -61,6 +61,8 @@ describe('sort by contribution', () => {
   });
 
   it('percentage is NOT hidden even after changing other fields', () => {
+    cy.get('#summary-wrapper').scrollTo('top');
+
```
MarcusTXK, line 65: This was a strange bug where `mui-select.grouping` is not visible specifically for this test case and not for the previous test case. I added this to scroll to the top of the summary wrapper so it is visible.

frontend/cypress/tests/chartView/chartView_percentage.cy.js
```diff
@@ -67,6 +67,8 @@ describe('sort by contribution', () => {
     cy.get('#summary .summary-chart__title--percentile')
       .should('exist');
 
+    cy.get('#summary-wrapper').scrollTo('top');
+
```
MarcusTXK, line 71: This was a strange bug where mui-select.grouping is not visible specifically for this test case and not for the previous test case. I added this to scroll to the top of the summary wrapper so it is visible.

vvidday:
(Just commenting on the `getFontColor` part for now)
I think I found another solution that doesn't require changing `getFontColor` to an export - moving the [`const getFontColor = window.getFontColor`](https://github.com/vvidday/RepoSense/blob/master/frontend/src/views/c-summary.vue#L137) definition to [inside the method itself](https://github.com/vvidday/RepoSense/commit/8594fd6881380b0391f00feea423bbbd9aa8dddd#diff-5419037f2fc8c426a6a54e0614ee8033d10e16426d30573486c27e5811036869R915). This seems to work with vue router (branch: https://github.com/vvidday/RepoSense/commits/add-vue-router-test) and is pretty similar to the original code, just the definition is moved.

(Additional context for others, as what Marcus has mentioned): The original code references [window.getFontColor](https://github.com/vvidday/RepoSense/blob/master/frontend/src/views/c-summary.vue#L137) at the start of `<script>` but outside the Vue component (`export default ...`). Without vue-router, `window.getFontColor` returns the function as expected, but when vue-router is added, `window.getFontColor` is undefined. I don't know the reason for this, and understanding why this happens probably requires some niche/specialized knowledge on the specific inner workings of Vue & Vue router.

MarcusTXK:
Great solution @vvidday, thanks alot for suggesting it, I think that is alot cleaner then just changing that single function into an exported function.

To add on a point I discussed with some of the developers, I feel that unless there is a reason why global functions are required, we should consider changing them to modular imports. In my opinion, it's better to avoid using global functions and use modular imports and exports instead. This is because global functions pollute the global namespace and may cause conflicts or make code harder to maintain. This also prevents any such issue of global functions potentially being undefined. Perhaps this could be addressed in a separate issue.

That said, as I am less familiar with the frontend codebase, if there is a reason why the global functions are required please do add on the discussion.

zhoukerrr:
> Great solution @vvidday, thanks alot for suggesting it, I think that is alot cleaner then just changing that single function into an exported function.
> 
> To add on a point I discussed with some of the developers, I feel that unless there is a reason why global functions are required, we should consider changing them to modular imports. In my opinion, it's better to avoid using global functions and use modular imports and exports instead. This is because global functions pollute the global namespace and may cause conflicts or make code harder to maintain. Using modular imports/exports also prevents any such issue of global functions potentially being undefined. Perhaps this could be addressed in a separate issue.
> 
> That said, as I am less familiar with the frontend codebase, if there is a reason why the global functions are required please do add on to the discussion.

@MarcusTXK Yes. I fully agree. This is definitely something that we can work on in future. Feel free to open an issue on this and I think this will be a good issue for new developers to familiarise with our codebase.

I just want to check if the idea of this PR is just to wrap the entire thing in a router, which is now called `c-home.vue`. There are no changes to any of the logic. Scrolling through to code to compare is a little difficult so I want to double check on that.

MarcusTXK:
> I  just want to check if the idea of this PR is just to wrap the entire thing in a router, which is now called `c-home.vue`. There are no changes to any of the logic. Scrolling through to code to compare is a little difficult so I want to double check on that.

Yup that is correct, it is simply the addition of a router and shifting everything that was formerly in formerly in `app.vue` into `c-home.vue`, with no change in logic. `app.vue` now contains the `router-view` which renders the appropriate content based on the route (which currently is just `/` which renders `c-home.vue`).

frontend/src/router/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+const routes: Array<RouteRecordRaw> = [
+  {
+    path: '/',
+    component: Home,
+  },
+];
+
+const router: Router = createRouter({
+  history: createWebHashHistory(),
```
vvidday, line 14: We discussed it in class but I think it would be beneficial if you could elaborate on the reason you chose the hash based history mode instead of the default one

Docs for reference: https://router.vuejs.org/guide/essentials/history-mode.html#hash-mode

vvidday (approved):
Thanks for the changes, LGTM! Just thought it would be good to add context on the use of hash mode.

Also definitely agree wrt refactoring the global functions, thanks for creating the issue on that - will discuss further there!

zhoukerrr (approved):
LGTM. The back button feature has been neglected for a long time. Just remember to document down the reason you chose the hash based history mode instead of the default one in this PR.

ckcherry23 (changes requested):
Hi @MarcusTXK, I noticed that the font 'Titillium Web' is not being applied on the webpage and is falling back to default fonts. After inspecting the Network tab in the console, it appears that there is no request being made for the font file to load at all. I suspect that this could be due to the import not being passed down when using Vue Router. If you or anyone have any insights on how we can fix this issue, then let's go ahead with that. Otherwise, we can explore alternative ways to import the fonts. Thanks!

frontend/src/router/index.ts
MarcusTXK: Sure, I previously write it in #1894, but I think it is a good idea to transfer it here since I am splitting up the PR. I have added it to the description, thanks for the suggestion.

MarcusTXK:
> Hi @MarcusTXK, I noticed that the font 'Titillium Web' is not being applied on the webpage and is falling back to default fonts. After inspecting the Network tab in the console, it appears that there is no request being made for the font file to load at all. I suspect that this could be due to the import not being passed down when using Vue Router. If you or anyone have any insights on how we can fix this issue, then let's go ahead with that. Otherwise, we can explore alternative ways to import the fonts. Thanks!

Good catch, I have shifted the global styles into `app.vue`. This fixes the bug of the font not loading.

ckcherry23 (approved):
LGTM!

HCY123902 (approved):

PR #1975, opened 2023-04-04
[#642] Click back in browser to return to previous view

### LINKED ISSUE #642: Report: clicking back in browser does not go to previous view
Not critical but good to have

MarcusTXK:
Fixes #642

### Proposed commit message

```
Currently, after interacting with the inputs and changing the view of 
the page, the user is unable to navigate by the browser to the previous 
view. 

Let's add the ability to navigate to the previous view by using routers 
to handle updating of the url.
```

### Changes
The addition of `decodeURI(hash.mergegroup).replace(/%2F/g, '/').split(window.HASH_DELIMITER)` for `mergegroup` was required, otherwise the backslash and brackets would not be unencoded properly.

### Demo
backward navigation

This must be merged after #1974.
Note: I am merging into a branch with the same code as #1974 to make it easier for review. Once #1974 is merged in I will change it back to merging into master

Just an update, this PR might take longer than expected to fix as almost every cypress test is failing, so they need to be fixed individually. Some of them are failing weirdly and I have yet to figure out why. Overall this is time consuming and this I will likely only be able to finish by the end of the weekend/

### DIFF
--- a/frontend/cypress/tests/chartView/chartView_mergeGroup.cy.js
+++ b/frontend/cypress/tests/chartView/chartView_mergeGroup.cy.js
@@ -5,7 +5,8 @@ describe('merge group', () => {
 
     cy.get('#summary label.merge-group > input:visible')
       .should('be.visible')
-      .check()
+      .check();
+    cy.get('#summary label.merge-group > input:visible')
       .should('be.checked');
 
     // after checking merge group, only one merged repo group will show
@@ -14,7 +15,8 @@ describe('merge group', () => {
 
     cy.get('#summary label.merge-group > input:visible')
       .should('be.visible')
-      .uncheck()
+      .uncheck();
+    cy.get('#summary label.merge-group > input:visible')
       .should('not.be.checked');
 
     // after un-checking merge group, all 5 summary charts will show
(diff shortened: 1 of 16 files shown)


MarcusTXK:
Just an update, this PR might take longer than expected to fix as almost every Cypress test is failing, so they need to be fixed individually. Some of them are failing weirdly and I have yet to figure out why. I will try to fix all the bugs by this weekend.

frontend/cypress/tests/chartView/chartView_mergeGroup.cy.js
```diff
@@ -26,9 +28,12 @@ describe('merge group', () => {
     cy.get('div.mui-select.grouping > select:visible')
       .select('groupByAuthors');
 
+    cy.wait(200);
```
HCY123902, line 31: Just curious, is there any more robust way to make this compatible with asynchronous calls as setting a specific value for the waiting time may cause a stability issue here

MarcusTXK: Yeah I would like to avoid doing it this way as well, I am trying to look for better alternatives as I feel that this could break if the browser is slower and takes more than 200ms to refresh. I am open to any suggestions if anyone has any.

PR #1976, opened 2023-04-05
[#1936] Migrate c-summary-charts.vue to TypeScript

vvidday:
Part of #1936

### Proposed commit message

```

Let's migrate one of the main components, c-summary-charts.vue, to
TypeScript. This will provide better type safety and reduce runtime
errors.
```

## `window.getAuthorDisplayName` type definition
This PR changes the type for the `authorRepos` parameter [in `window.getAuthorDisplayName`](https://github.com/reposense/RepoSense/commit/ad1b25ca33083ce4bcb0aac1dfcfc8efad3c59e8), from `Repo[]` to `User[]`. This was a mistake I made in the initial type definition (as both `Repo` and `User` have the `displayName` attribute).


frontend/src/components/c-summary-charts.vue
```diff
@@ -479,31 +487,41 @@ export default {
       this.$store.commit('updateTabZoomInfo', info);
     },
 
-    getBaseTarget(target) {
+    getBaseTarget(target: HTMLElement | null): HTMLElement | null {
+      if (!target) {
+        // Should never reach here
```
zhoukerrr, line 492: I think adding a log message could be more helpful here? Just in case the code enters here for whatever reason, we could trace the issue quickly.

zhoukerrr (commented):
@vvidday Just a minor point I want to point out. The rest LGTM!

frontend/src/components/c-summary-charts.vue
vvidday: Agreed, just added a console statement & elaborated a bit more on the comment itself. The linter is currently setup to warn console statements, but I think that this rare false positive is fine as it can catch unintentional console statements from debugging that were accidentally committed.

ckcherry23 (approved):
I have reviewed the code and overall, the TypeScript migration looks great! 
However, I do have a comment about the prop types. 

I noticed we are using generics to define the types under `props` in `c-summary-charts` as `Array<string>` for example, but in `c-summary`, we are making use of `PropType` so the equivalent here would be `Array as PropType<string[]>`. 

Similarly, most of the attributes under `data()` are defined as `[] as string[]` but we could make use of generics here as well as `[] as Array<string>`. 

Do you think we can standardize how we define our types using a lint rule?

vvidday:
> I have reviewed the code and overall, the TypeScript migration looks great! However, I do have a comment about the prop types.
> 
> I noticed we are using generics to define the types under `props` in `c-summary-charts` as `Array<string>` for example, but in `c-summary`, we are making use of `PropType` so the equivalent here would be `Array as PropType<string[]>`.
> 
> Similarly, most of the attributes under `data()` are defined as `[] as string[]` but we could make use of generics here as well as `[] as Array<string>`.
> 
> Do you think we can standardize how we define our types using a lint rule?

Great point about generics vs `T[]` syntax, [according to TS docs](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#arrays) they are equivalent, but I agree we should stick to one for consistency. Do you have a preference on which?

There [does seem to be a lint rule for this](https://typescript-eslint.io/rules/array-type), we can either choose generics or the `T[]` syyntax.

frontend/src/components/c-summary-charts.vue
```diff
@@ -181,25 +181,30 @@
...
+export default defineComponent({
   name: 'c-summary-charts',
   components: {
     cRamp,
   },
   mixins: [brokenLinkDisabler],
   props: {
     checkedFileTypes: {
-      type: Array,
+      type: Array<String>,
```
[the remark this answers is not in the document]
vvidday: Thanks for catching!

zhoukerrr:
@vvidday Personally I think the `T[]` syntax is cleaner and easier to write. But of course the team will have to come to an agreement. The lint rule and refactoring of the code code be done in a separate PR or be made as a good first issue for new developers to familiarise with our codebase.

ckcherry23:
> Great point about generics vs `T[]` syntax, [according to TS docs](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#arrays) they are equivalent, but I agree we should stick to one for consistency. Do you have a preference on which?

I think I would prefer the `array-simple` rule as it uses the more modern `T[]` syntax for simpler types while allowing `Array<T | U>` for more complex types (where (T | U)[] may be a confusing syntax). I will create a new issue for this, so that we can tackle it in another PR.

zhoukerrr (approved):
LGTM. Not sure how to work with the cypress warning regarding the unexpected console statement. I do think it is useful in some cases. Maybe a separate issue could be raise for this for new developers to work on?

HCY123902 (approved):

Issue #1856, opened 2023-01-08
Possible Concurrency Issue - RepoConfigParserTest


yhtMinceraft1010X:
**Tell us about your environment**

* **RepoSense Version:** N/A (development)
* **OS and Version:** Ubuntu 22.04 CI
* **Web Browser and Version (if applicable):** N/A

**Please include the steps to reproduce the bug.**
(Concurrency issue - may not be reproducible)
1. Run a test

**What was expected to happen?**
All tests pass

**What actually happened? Please include a screenshot of the output.**
When I was creating a PR for #1853, there was one failing test - `RepoConfigParserTest - merge_twoRepoConfigs_success()` in one run of the CI workflow. Running the CI workflow again results in the test passing:
![Screenshot (506)](https://user-images.githubusercontent.com/69678785/211193746-c9a5aad1-f507-4c30-8816-2b3009b0688d.png)

**If possible, include the URL to your RepoSense report or log files (if any).**
The relevant workflow run is here: https://github.com/reposense/RepoSense/actions/runs/3866800055. There were 2 workflow run attempts for the Ubuntu OS.

**Note about test concurrency**
Backend tests were parallelized in #1806. This was done due to the sheer number of tests cases we have.

chan-j-d:
https://github.com/reposense/RepoSense/actions/runs/4130785264/jobs/7137805345#step:11:244 seems like it's not an isolated incident as well and just occurred again here

PR #1978, opened 2023-04-06
[#1908] Update how logger output is recorded

### LINKED ISSUE #1908: Changing how logger output is recorded
### What would you like to see in RepoSense
Output logging messages to a file `reposense.log` in the directory where reposense is run.
Only at the end of the reposense run do we move the `reposense.log` file to the output directory.

### Is the suggestion related to a problem?
Currently, `setLogFolderLocation` is called in `ArgsParser` which sets the output folder location for all loggers. This occurs in line 351.
https://github.com/reposense/RepoSense/blob/bae660c0c237bb52644bd3cb30b5ff8c40fda42a/src/main/java/reposense/parser/ArgsParser.java#L351

However, there is a possible logger warning earlier at line 336
https://github.com/reposense/RepoSense/blob/bae660c0c237bb52644bd3cb30b5ff8c40fda42a/src/main/java/reposense/parser/ArgsParser.java#L335-L337
which if triggered, does not get recorded to the final log file and only shows up in the standard output.

Besides that, the current structure causes `ArgsParser::parse` to have side effects that affects `LogsManager` which is almost like a global variable. I believe this causes the concurrency issue in #1856 as the `ArgsParser::parse` method is called multiple times in test cases indirectly. 

1. Write the logs to the directory where reposense is currently being run right at the start of execution.
2. At the end of a run, move the log file from the working directory to the output directory.

MarcusTXK:
Fixes #1908

### Proposed commit message

```
Currently, setLogFolderLocation is called in ArgsParser which sets the
output folder location for all loggers. However, any logs before the
output folder is set do not get recorded in the final log file and only
show up in the standard output. This also potentially causes concurrency
issues, since the current structure causes ArgsParser::parse to have
side effects that affect LogsManager.

Let's output logging messages to a file reposense.log in the directory
where RepoSense is run and only move the reposense.log file to the
output directory at the end of the RepoSense run.
```

### Other Information
`FileHandler` is now created with the default directory of the current directory that RepoSense is running in.

At the end of `RepoSense::main`, `LogsManager::moveLogFileToOutputFolder` is called, where the log files are shifted from the temporary output folder to the output folder specified by the user.


src/main/java/reposense/system/LogsManager.java
```diff
@@ -27,6 +31,7 @@ public class LogsManager {
     // eg. reposense.log.0, in the logs/ folder of the working directory
     private static final String LOG_FOLDER_NAME = "logs";
     private static final String LOG_FILE_NAME = "reposense.log";
+    private static final Path TEMP_LOG_DIR = Paths.get(System.getProperty("user.dir"));
```
sikai00, line 34: I'm not sure if this fixes the concurrency issue mentioned in #1908/#1856. According to the issue description, the problem seems to be coming from the fact that LogsManager is a static variable, which happened to be accessed at the same time.

At this point, I don't seem to recall having seen the problem as mentioned in #1856 which was closed and resolved. I believe that issue was primarily due to unsynchronized access to the LogsManager. I might be missing something though so please let me know otherwise!

src/main/java/reposense/RepoSense.java
```diff
@@ -124,6 +124,7 @@ public static void main(String[] args) {
             // help message was printed by the ArgumentParser; it is safe to exit.
         }
 
+        LogsManager.moveLogFileToOutputFolder();
```
sikai00, line 127: I believe that the warning not caught can possibly be fixed by just setting the location of the logs earlier before all the warnings, rather than moving things around.

sikai00 (commented):
Some general questions regarding the PR

src/main/java/reposense/system/LogsManager.java
chan-j-d: To clarify, it's not the LogsManager class itself that causes the concurrency issue but the list of loggers inside LogsManager where all their output folders have to be changed every time the `ArgsParser::parse` is called (see the `synchronized` block in `LogsManager`). On a normal run, the method is only called once but  in tests, it's called multiple times. So the list has a chance of being accessed concurrently which causes the issue. 

This fixes it in a roundabout way by only changing a single variable which could on paper lead to another type of concurrency issue (multiple writes to a single variable) that I believe is not caught but it's also not one that matters in this context since we just have to avoid the concurrency exception being thrown since the actual logs output folder does not affect our test cases.

src/main/java/reposense/RepoSense.java
chan-j-d: When bringing up the issue, I did think of this but it's a bit of a strange problem because this method would immediately require the out folder argument always being the first argument that we read so that we can set the logs output folder. Though not really a big issue now, some scenarios this might be an issue is for example in your setup wizard where if you  want to include an output folder prompt together with preemptive argument checking, you would need to provide the output folder prompt as a first option otherwise you can't log anything.

I guess the main consideration is whether you believe there's anything worth logging between the start of a reposense run and the earliest moment the output folder location can be known. I think it's a relatively trivial change to implement to be more flexible which is why I suggested it.

sikai00: In this case, I do believe it will be better since it covers all grounds rather than rely on the positioning of the output from parse.

src/main/java/reposense/system/LogsManager.java
```diff
@@ -88,7 +91,7 @@ private static void removeHandlers(Logger logger) {
      * Creates {@link FileHandler} if it is null.
      */
     private static void addFileHandler(Logger logger) {
-        Path path = logFolderLocation.resolve(LOG_FOLDER_NAME);
+        Path path = TEMP_LOG_DIR.resolve(LOG_FOLDER_NAME);
 
         try {
             if (!Files.exists(path)) {
```
sikai00, line 97: Is this check still relevant? Considering we use the current working directory as per TEMP_LOG_DIR

src/main/java/reposense/RepoSense.java
MarcusTXK: Yup I agree, for instance, a log could be required before `ArgsParser::parse` is even called, in that case having the logs output in a temp directory first would be ideal.

src/main/java/reposense/system/LogsManager.java
MarcusTXK: Good catch, thanks for noticing this

src/main/java/reposense/system/LogsManager.java
```diff
@@ -105,13 +108,51 @@ private static void addFileHandler(Logger logger) {
...
+        }
+
+        try {
+            // Close the FileHandler to release the lock on the log file
+            fileHandler.close();
+            Path tempLogDir = TEMP_LOG_DIR.resolve(LOG_FOLDER_NAME);
+            Path outputLogDir = logFolderLocation.resolve(LOG_FOLDER_NAME);
+
+            if (!Files.exists(outputLogDir)) {
+                Files.createDirectories(outputLogDir);
```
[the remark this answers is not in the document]
MarcusTXK: I've added a log here to log when the log output folder is successfully created

src/main/java/reposense/system/LogsManager.java
MarcusTXK: I left the creation of the directory as it is still required since the temp folder might not exist. However,  I have shifted the logging to below, after the addition of the `fileHandler` to the logger, so that the logs will be output correctly.

src/main/java/reposense/system/LogsManager.java
```diff
@@ -88,30 +91,75 @@ private static void removeHandlers(Logger logger) {
      * Creates {@link FileHandler} if it is null.
      */
     private static void addFileHandler(Logger logger) {
-        Path path = logFolderLocation.resolve(LOG_FOLDER_NAME);
+        Path path = TEMP_LOG_DIR.resolve(LOG_FOLDER_NAME);
 
         try {
-            if (!Files.exists(path)) {
+            boolean isFileExist = Files.exists(path);
+            if (!isFileExist) {
```
sikai00, line 98: should this and line 107 be combined? since this actually creates the log temp folder

sikai00, line 98: should this and line 107 be combined? since this actually creates the log temp folder

sikai00 (changes requested):

src/main/java/reposense/system/LogsManager.java
MarcusTXK: The reason why I shifted the logging down to 107, is that if the temp log folder is created, but the file handler is not created( which usually seems to be created together), the log won't be output. 

However, we need to create the temp log folder before the file handler can be created, as otherwise, an error will occur as the lock file, `reposense.log.0.lck` cannot be created as the directory does not exist.

sikai00 (approved):
LGTM!

src/main/java/reposense/RepoSense.java
```diff
@@ -108,6 +108,7 @@ public static void main(String[] args) {
...
         LogManager.getLogManager().reset();
```
chan-j-d, line 112: I think with the new implementation, this line can be removed since you already close the relevant file handler before deleting the logs folder. This was previously added due to the .lck file lingering and I wasn't sure what was the cause but it seems you've addressed it.

src/main/java/reposense/system/LogsManager.java
```diff
@@ -27,6 +31,7 @@ public class LogsManager {
     // eg. reposense.log.0, in the logs/ folder of the working directory
     private static final String LOG_FOLDER_NAME = "logs";
```
chan-j-d, line 32: I think we should go for a far less generic name like `reposense-logs` or something similar since we're now creating this in the running directory, there's far higher chance of name clash

chan-j-d (changes requested):
Some small changes, looks mostly quite good.

src/main/java/reposense/system/LogsManager.java
MarcusTXK: Good idea, it would be bad to conflict with existing files

chan-j-d:
@MarcusTXK checkstyle is failing, could you get it fixed?

chan-j-d (approved):
LGTM!

PR #1979, opened 2023-04-07
[#1004] Standardize fonts for consistency across panels

ckcherry23:
Part of #1004 

### Proposed commit message
```
There are inconsistencies in the design between the different panels. 
One aspect that varies is the fonts used in terms of the font sizes and 
families. 

Let us standardize the font sizes across the Vue front-end using mixins.
Let us also set the font family for sans-serif to be Titillium Web and
monospace to be GitHub's default font.
```

### Other information
This PR will standardize the font sizes to be consistent across the panels. I have also changed all monospace text to follow GitHub's font style. 

To facilitate this, I have defined 5 font mixins in `_utils.scss`:
* `mini-font`: Used for secondary details. Includes tags, contribution percentile.
* `small-font`: Used for primary details. Includes file types, file/author legends, toolbar components, tooltips, etc.
* `medium-font`: Used by default. Includes file names, repo names, and author names.
* `large-font`: Used for headings. Includes panel headings, and dates in the Commits Panel.
* `mono-font`: Used for code. Includes code in the Code Panel, commit messages in the Commits panel

This will make it easier to switch font sizes or families later based on the group they belong to.

There are some more changes to do regarding #1004, such as removing the panel title for consistency's sake and changing the style of the 'filter by glob' component to match the rest. However, I have limited the scope of this PR to only CSS font changes.

#### 1. Default
Screenshot 2023-04-07 at 11 42 59 PM

#### 2. With Code Panel
Screenshot 2023-04-07 at 11 43 29 PM

#### 3. With Commits Panel
Screenshot 2023-04-07 at 11 44 04 PM


ckcherry23:
The Cypress tests are failing because the test case "percentage is NOT hidden even after changing other fields" in `chartView_percentage.cy.js` is failing on the CI despite passing locally. A fix for this has been implemented by @MarcusTXK  in #1974, but I am not sure what is causing this disparity.

frontend/src/styles/_utils.scss
```diff
@@ -11,12 +11,23 @@
   font-weight: normal;
 }
 
+@mixin small-font {
+  font-size: .8rem;
```
vvidday, line 15: Was just wondering if there's any specific reason you chose these values for the sizes?

The small font looks great on my laptop screen but "feels" small on my desktop. I'm not familiar at all with the best practices for font size etc across devices, so maybe this isn't an issue.

<details>
<summary>Screenshots</summary>

## Before
image
## After
image
</details>

vvidday (approved):
The code LGTM! I like the approach of using mixins throughout to make the styles consistent. Just wanted to check regarding the font sizes chosen.

frontend/src/styles/_utils.scss
ckcherry23: Thanks for testing on your desktop @vvidday!

The font sizes are defined using `rem` units (and monospace with `px` units), which are relative units and will look similar on every device. This is the same method we were using before this PR as well, but I have just replaced individual font size styling with mixin imports. 

The general practice is to use relative units like these instead of `cm`, `pt`, etc. However, there is still the issue of text appearing too small on desktops. The way to get around this is by using media queries, that set different font sizes based on the size of the device. 

Since the previous behavior of RepoSense was similar to this (not ideal on desktop), I think we can keep it as is for now and add media queries for responsiveness in a separate PR.

vvidday: Sounds good, thanks for clarifying!

ckcherry23: If you are specifically referring to the size of the commit messages, I have changed it to use the same font sizes as the code in the Code Panel, which is why it is smaller now.

frontend/src/views/c-authorship.vue
```diff
@@ -803,20 +803,21 @@ export default {
...
 
       .mui-form--inline {
+        @include small-font;
         align-items: flex-end;
         display: flex;
         flex-wrap: wrap;
 
         #search {
-          @include medium-font;
+          @include small-font;
```
HCY123902, line 812: Can I ask about the reason for changing the font size here?

frontend/src/views/c-authorship.vue
```diff
@@ -803,20 +803,21 @@ export default {
...
           margin-top: 1.25rem;
           min-width: 130px;
           padding: .5rem 1.0rem .25rem 1.0rem;
           width: 70%;
         }
 
         #submit-button {
-          @include medium-font;
+          @include small-font;
```
HCY123902, line 820: Similar question herd

frontend/src/views/c-authorship.vue
ckcherry23: The toolbar controls on the summary panel were using a smaller font whereas the ones on the authorship panel were using `medium-font`. To unify the design for these controls that have similar purposes, I have defined a `small-font` with a size between the initial 2 sizes.

HCY123902 (approved):

ckcherry23:
@zhoukerrr Can you please help to review this PR, thanks!

To everyone, would it be beneficial to document how to use the font styles in the codebase with comments or maybe within the HTML report section of the developer guide?

zhoukerrr:
> @zhoukerrr Can you please help to review this PR, thanks!
> 
> To everyone, would it be beneficial to document how to use the font styles in the codebase with comments or maybe within the HTML report section of the developer guide?

@ckcherry23 I think it will certainly be helpful to include that in the HTML section. You can just put what you have attached in this PR under the section. Comment can also be added next to the mixins so that people know where to use them in future without going to the DG.

zhoukerrr (approved):
LGTM!

PR #1983, opened 2023-04-07
[#1982] Use correct "sort within groups by" when visiting by url

### LINKED ISSUE #1982: Wrong `sort within groups by` when visiting by url

1. If the url contains anything for `sort within groups by` for anything other than `↑ title`, it will get replaced by `↑ title`.
2. Observe  the gif below, where `↓ contribution` is selected and the changes are updated in the url. However upon copying the url and revisiting the page, the sort is changed back to `↑ title`.

sortWithinGroupsByBug

The sort should remain as configured by the user in the url.

MarcusTXK:
Fixes #1982 

### Proposed commit message

```
Currently, if the URL contains anything for "sort within groups by" for 
any value other than "↑ title", it will get replaced by "↑ title" when 
the url is visited. 

Let's fix it so the sort remains as configured by the user in the URL.
```

### Other information

This was caused by refactoring in #1953. `hash.sortWithin && Object.values(SortWithinGroupSelection).includes(hash.sort as SortWithinGroupSelection` would always return false.

 I suggest adding cypress tests for `renderFilterHash()` to ensure that the parameters in the URL are loaded as per expected in the future for early detection and prevention of such bugs.


ckcherry23 (approved):
LGTM, Nice catch!

zhoukerrr (approved):
LGTM. Nice catch. Just a side note, I think for **"low hanging fruits"** such as this one, we can keep the issue open and let new developers to work on this kind of issues.

MarcusTXK:
@zhoukerrr 
> LGTM. Nice catch. Just a side note, I think for **"low hanging fruits"** such as this one, we can keep the issue open and let new developers to work on this kind of issues.

Yeah I agree, I would have done that, but unfortunately I need these fixed for #1975 to work, as the reason why alot of cypress tests failing and some strange bugs occurring are partly due to a few undetected bugs, especially in regard to getting the params from the url, which are frustrating to hunt down. 

Another one that I have not made a PR for is if `checkedFileTypes` is an empty array in the url, when visited/refreshed, it becomes fully populated again. There will probably be others as well as I slowly debug through the failing cypress tests/bugs there. 

Do you think I should continue make seperate issues and PR fixes for these? I did so with the intention of keeping each PR with one purpose + documenting the bugs in issue, but alternatively I could solve them at one go in #1975.

zhoukerrr:
@MarcusTXK Understood. Yes please open each PR for one purpose. That will be easier to review and keep track as well.

frontend/src/views/c-summary.vue
```diff
@@ -405,7 +405,8 @@ export default defineComponent({
       if (hash.sort && Object.values(SortGroupSelection).includes(hash.sort as SortGroupSelection)) {
         this.sortGroupSelection = hash.sort as SortGroupSelection;
       }
-      if (hash.sortWithin && Object.values(SortWithinGroupSelection).includes(hash.sort as SortWithinGroupSelection)) {
+      if (hash.sortWithin
+        && Object.values(SortWithinGroupSelection).includes(hash.sortWithin as SortWithinGroupSelection)) {
```
HCY123902, line 409: Nice observation. I agree with the point that the URL parameters need to be coverer by `cypress` as well, which can be put in a sepearate pull request

HCY123902 (approved):

PR #1269, opened 2020-05-15
[#1240] Remove insertion and deletion of each commit in commits.json

0blivious:
Fixes #1240 
```
We now have the line contribution info of each file type within a
commit, we can remove the insertion and deletion info of each
commit in the commits.json file. Same for
authorFinalContributionMap.

The insertion and deletion info can be calculated in the front
end instead, by summing up all the insertion and deletion of
all filetypes within a commit.

Let's remove the insertion & deletion in commit.json file and
calculate them in front end.
```

Issue #1240, opened 2020-04-17
Remove insertion and deletion of each commit in commits.json


jamessspanggg:
> > Update: can't really remove the insertions and deletions of each commit as they are required in the backend to calculate the variance.
> 
> Please raise an issue for this, we should aim to resolve this in the near future

_Based on my discussion with @eugenepeh in #1192_ 

Since we have the line contribution info of each file type within a commit, we can remove the insertion and deletion info of each commit in the `commits.json` file. 

The insertion and deletion info can be calculated in the front end instead, by summing up all the insertion and deletion of all filetypes within a commit. More info shown in screenshot below:

<img width="536" alt="Screen Shot 2020-04-17 at 8 13 29 PM" src="https://user-images.githubusercontent.com/32864116/79568233-11587800-80e8-11ea-94d7-005478ab6ab1.png">

In the meantime, the variance can also be calculated in the front end, as currently the calculation of variance is coupled with the insertion and deletion info of each commit.

PR #1985, opened 2023-04-08
[#1984] Frontend: fix CheckedFileTypes unable to be set as empty

### LINKED ISSUE #1984: `CheckedFileTypes` unable to be set as empty

1. If `breakdown by file type` is selected and no file type is selected, if the page is refreshed/the url is visited, all file types will be selected again.

checkedFileTypesBug

The url gets populated with all file types again:
* original: `...&checkedFileTypes=`
* after refreshing/visiting url: `...&checkedFileTypes=md~puml~java~scss~fxml~css~gradle~json~gitignore`

The selected file types should remain as configured by the user in the url.

MarcusTXK:
Fixes #1984 

### Proposed commit message

```
Currently, if "breakdown by file type" is selected and no file type is 
selected, if the page is refreshed/the URL is visited, all file types 
will be selected again. 

Let's fix it so the selected file types remain as configured by the user
in the URL, for both summary and code commit panels.
```

### Other information

The bug was introduced since `renderFilterHash()` is checking using the truthiness of `hash.checkedFileTypes`. This is dangerous as it lead to the case where if the string is empty (which is a valid input, such if `All` was unchecked), it is also treated as if no `checkedFileTypes` were set in the url (as if it were undefined), and the if block to set the `checkedFileTypes` is skipped.

Update:
I have fixed the behaviour for the commit panel to be consistent.

I have also fixed the bug where the contribution bar disappears. This is caused by the `checkedFileTypeContribution` being set to 0, but not being updated when the new `avgContributionSize` is calculated once `All` is rechecked, since it is not 'undefined'.


ckcherry23:
Hi @MarcusTXK, good catch! 
However, is there a case where a user would knowingly choose no file types at all in the summary panel? I don't think RepoSense will have any useful functionality when no file types are selected. 

I feel that such a selection would most likely be an error, so I personally would prefer the existing behavior of resetting the file types. But if the rest of the team disagrees then we can go ahead with this PR.

MarcusTXK:
> Hi @MarcusTXK, good catch! However, is there a case where a user would knowingly choose no file types at all in the summary panel? I don't think RepoSense will have any useful functionality when no file types are selected.
> 
> I feel that such a selection would most likely be an error, so I personally would prefer the existing behavior of resetting the file types. But if the rest of the team disagrees then we can go ahead with this PR.

Good points. However, in my opinion, the URL is meant to store the state of the user's inputs and we should not simply reset what the user has set. I feel if a user clicks `All` to uncheck all filetypes, it should correctly be reflected in the url and rendered accordingly.

This is more visible in the future should #642 be merged in and the user's inputs are updated in the routing history, as when the user unchecks `All` and no file type is selected, the browser rerenders the updated URL and since no file type is selected, sets all file types as selected again (which renders as all of the file types being ticked). This prevents a user from unchecking the `All` button at all.

vvidday (commented):
Great points @ckcherry23 @MarcusTXK, I agree that functionally it doesn't make sense to preserve the 'all file types unchecked' state, but I don't really feel that the downside is significant (In the likely case it was unintentional, the user can re-check the options again). And considering that this is negatively affecting #1975, I think we can go ahead with this fix.

Also noticed that the behaviour is inconsistent for the code/commits panel -  if all files are similarly unchecked, the code panel maintains the state on refresh but the commits panel re-checks all options, much like the summary panel in this PR. We can look into standardizing this behaviour (either here or in a separate PR?).

Lastly there's unfortunately a small bug introduced here, where the contribution bars in merge view don't re-appear correctly after a full uncheck -> refresh -> recheck. 
Reproduce:
1) Merge all groups
2) Uncheck 'all'
3) Refresh
4) Recheck 'all'
5) Notice contribution bar is missing

This bug most definitely exists before this PR, but it was never encountered because of the re-checking behaviour as mentioned above.

damithc:
I think it's best to preserve what the user selected, to make the behavior less 'surprising' to the user. Users get annoyed when the software tries to 'outthink' them and prevent them from doing the exact thing they want, for which they might have a good reason that we did not anticipate. For example, a user might want to share a view without any file types selected to let the viewer select any type of file they wish easily.

HCY123902:
@MarcusTXK According to the above discussion, may I suggest standardizing the behavior of the commits panel to allow empty checked file types as well within this pull request? Additionally, if possible, try to fix the contribution bar problem within this pull request as well, since the changes made here surface the problem

ckcherry23 (commented):
@MarcusTXK This PR might give some insights into why this behavior was introduced: [[#1240] Remove insertion and deletion of each commit in commits.json](https://github.com/reposense/RepoSense/pull/1269)

The initial implementation for file type selection was `if (hash.checkedFileTypes !== null)`, but the reason it was changed was specifically to fix the contribution bar bug.

MarcusTXK:
Thanks alot for all the feedback. I will standardize the behavior of the commits panel as well as standardizing the behavior of the commits panel in this PR.

@ckcherry23 Thanks alot for the resource, I will look through it.

ckcherry23 (approved):
LGTM, thanks for fixing this!

HCY123902 (approved):

zhoukerrr (approved):
LGTM

Issue #1962, opened 2023-03-24
Modify screenshots to include new changes in RepoSense report


HCY123902:
### What problem(s) did you run into that caused you to request additional documentation? What, if any, existing documentation relates to this proposal?

Several screenshots in the [User Guide](https://reposense.org/ug/usingReports.html#using-reports) and the [Developer Guide](https://reposense.org/dg/report.html#html-report) are obsolete with the frontend changes introduced in #1453, #1883, and #1939, such as the title highlight in the summary view, number of deletions in the commits panel, and the inclusion of author labels for a merged group in the authorship contribution panel.

![image](https://user-images.githubusercontent.com/54735123/227433818-825639cf-3433-4696-b551-fbe9304070d3.png)
![image](https://user-images.githubusercontent.com/54735123/227433860-422f6e17-5120-46e5-8989-1a75b691f508.png)

### What would you like added or modified in the documentation?

The screenshots should include the new frontend changes indicated above, and related description can be updated accordingly

Issue #1929, opened 2023-03-05
Add dynamic positioning support for tooltip


HCY123902:
### What feature(s) would you like to see in RepoSense

Currently, most tooltips are shown above buttons and text. When these tooltips appear at the top of the viewport, part of the tooltips will not be rendered.

![image](https://user-images.githubusercontent.com/54735123/222944970-1dfcbc38-a595-42cb-9dd5-8825fa504bbc.png)

A previous pull request #1860 has implemented a solution to add dynamic positioning support for collapse file tooltips so that they will move below the title when appearing at the top of the viewport. The same changes can be applied to the other tooltips as well, including those on the chart panel, authorship contribution panel, and commits panel.

### If possible, describe the solution

A possible solution has been proposed in #1860

PR #1988, opened 2023-04-08
[#1894] Add embeddable ramp widget

### LINKED ISSUE #1894: Add embeddable ramp widget

MarcusTXK:
Fixes #1894 

### Proposed commit message

```
The ability to copy a link for an embeddable ramp widget with the click 
of a button and embed the ramp widget into any existing website will be 
a useful feature. This allows users who intend to feature any 
contributions made to a repository/projects in their portfolio/any 
website, especially as I feel it is both eye-catching and provides a 
lot of useful information. The embedded ramps should follow any filters 
configured by the users above so that the information displayed by the 
ramps can be customized. 

Let's add a button to allow users to copy and embed widgets.
```

### Screenshots
RepoSense Report
image

Embedded Group Ramp (Unmerged)
image
* Note that elements such as the numbering and certain buttons that do not have usage in iframes are hidden. An additional button that redirects the user to the RepoSense report with the current inputs is added to each user.

Embedded Group Ramp (Merged)
image

Embedded Single Ramps
image
* Similarly here, the numbering and unnecessary buttons are removed.

The height of the iframe are calculated based on `the number of charts in the iframe * chart height + title (if present)` such that no scrolling of the iframe is required when viewing it. The width is set to an arbitrary value of `800`. These dimensions can be edited by the user in the iframe should they prefer other values.

Tested the iframes on the following html file (green bg to see the widget margins better) :
```html
<!DOCTYPE html>
<html>
<head>
	<title>Example with iFrame:</title>
</head>

<style>
body {
  background-color: #7bc96f;
}
</style>
<body>
	<h1>Embedded Widget</h1>
        <iframe src="XXX" frameBorder="0" width="800px" height="XXXpx"></iframe>
</html>
```

#### Why iframes?
For embeddable widgets I initially considered two options, the first being [standard Web Components](https://vuejs.org/guide/extras/web-components.html) that can be embedded in any HTML or iframes.

I ultimately decided to use iframes as using standard Web Components required prebuilding of the components and for users to add two items, a link in the head for css and a script in the body for the widget source, which is troublesome for the user. Iframes on the otherhand can be directly embedded in the body of any html file with no hassle.

#### Reuse of `c-summary-charts.vue`
I initially intended to create a seperate widget component, however this has the drawback of being hard to maintain as if the summary chart is updated, the widget would have to be updated separately. Since `c-summary-charts.vue` renders the charts in a double for loop, I realized that we can easily reuse it but filter out non relevant elements and only render the specified charts. This can be thought of as a matrix, where we access the required row or required element directly by keeping track of indexes. This is done via the url parameters `chartGroupIndex` and `chartIndex`.

#### Overall
This are the main changes:
* I have extracted out all common logic between `c-home` and `c-widget` in `app.vue`. The common logic are passed down as props to the respective pages.
* `router-view` renders `c-home` or `c-widget` depending on if the routing is `/` or `/#/widget`. 
* `c-summary` does not render the `summary-picker` if `isWidgetMode` prop is passed in, and now passes in `chartGroupIndex` and `chartIndex` to `c-summary-charts`  which only renders the specified indexes.
* `c-summary-chart` conditionally renders certain buttons, depending on if it is currently a widget (which is determined by if `chartGroupIndex` and `chartIndex` are passed present).
* Common CSS for summary chart between c-widget and c-home has been shifted into `summary-chart.scss`

This must be merged after #1974.
Note: I am merging into a branch with the same code as #1974 to make it easier for review. Once #1974 is merged in I will change it back to merging into master

### tl;dr
Common logic moved into `app`, `app` renders either `c-home` or `c-widget` and passes in common logic as props.
`c-widget` renders `c-summary` with some extra props.

Edit: Documented bugs:
* clicking ramp does not lead to report
* non-localhost for local deployment does not work
* warnings not hidden


vvidday (commented):
Thanks for your work on this @MarcusTXK! The structure looks great, just found some small issues/bugs

- Currently clicking on the "view breakdown of commits" (the button that will usually open the commits panel on the report), the user is redirected to the report itself (with commits panel not open). I was wondering if it would be better to open the corresponding commits panel on the report as well?
- Similarly, if granularity is set to "week" and the user clicks on the ramp, ordinarily on the report it opens the corresponding commit in the commits panel, but in the widget it does nothing. I think it would be good to minimally redirect to the report, and preferably open the corresponding commit like it does in the original report
- Currently the widget renders any errors in the summary panel, could this be removed in widget view?

<details>
<summary>Screenshot</summary>

image
</details>

ckcherry23 (commented):
Hi @MarcusTXK, thanks for working on this PR! 
On some initial testing, I noticed a few things to address:
* The copy to clipboard function works on `localhost:8080` but not on the IP addresses (eg. `http://0.0.0.0:8080/`). I don't know what the issue is here, but can you take a look to ensure it works in production?
* The tooltips on hover appear outside the viewport of the iframe. This can probably be fixed when resolving #1929. However, the issue is more amplified on the iframes, as there is no way to see what's on the tooltips even after some scrolling. 
* As discussed in class, the embeddable ramp widget has some bugs when tested using the RepoSense backend test report data. Please take a look, thanks!

MarcusTXK:
Thanks a lot for the feedback, I have addressed all of them below.

@vvidday 
> Currently the widget renders any errors in the summary panel, could this be removed in widget view?

Thanks for noticing this! I have fixed this as well as breakdown filetypes (that I realised was not hidden as well)

> Currently clicking on the "view breakdown of commits" (the button that will usually open the commits panel on the report), the user is redirected to the report itself (with commits panel not open). I was wondering if it would be better to open the corresponding commits panel on the report as well?

I tried implementing this logic, however after spending quite awhile, I realized that it requires quite a bit of changing the code in order to render the URL for `href` as one that points to the open tab (which needs to be "computed" on render of `c-summary-charts`), as it involves invoking `openTabZoom` and replicating all of its logic as well as the other functions it calls, but without modifying the actual state (since we only want to get the url and not actually render what happens if it is clicked). 

I think a simple workaround for now would be for the user to open the tab first, and then copy the iframe, if the user wants the page to be redirected to the report with their tab open. If we are to add logic for this, I feel it should be in a seperate PR as part of a further enhancement of the widget feature, as the non-trival refactoring needed will make this PR more confusing (as the current one already involves a degree of refactoring).

>Similarly, if granularity is set to "week" and the user clicks on the ramp, ordinarily on the report it opens the corresponding commit in the commits panel, but in the widget it does nothing. I think it would be good to minimally redirect to the report, and preferably open the corresponding commit like it does in the original report

Good catch for this! For now, I have updated the logic to redirect to the report for ramps that are of granularity "week". I think this functionality can be added together with the one above in a separate PR, for the same reasons of adding even more complexity to this PR.

@ckcherry23 
> The copy to clipboard function works on localhost:8080 but not on the IP addresses (eg. http://0.0.0.0:8080/). I don't know what the issue is here, but can you take a look to ensure it works in production?

This was an interesting issue to solve. It turns out that `navigator.clipboard` requires a secure origin — either HTTPS or localhost (or disabled by running Chrome with a flag), otherwise it will be undefined. I have added a workaround for non-secure origins. I originally tried using other libraries such as `copy-to-clipboard` and `Clipboard.js`, but I found that they did not work either, so I decided to use `document.execCommand('copy')`, which unfortunately requires spawning of a temporary text area that is removed subsequently, but I think that this is fine as it is only for development.

> The tooltips on hover appear outside the viewport of the iframe. This can probably be fixed when resolving https://github.com/reposense/RepoSense/issues/1929. However, the issue is more amplified on the iframes, as there is no way to see what's on the tooltips even after some scrolling.

Yes, I agree that this is an issue, but I feel that this is best fixed in a separate PR. Perhaps in widget mode I could temporarily disable tooltips until this is fixed.

> As discussed in class, the embeddable ramp widget has some bugs when tested using the RepoSense backend test report data. Please take a look, thanks!

Thanks for noticing this bug, I did not notice this as the reports I tested it on were smaller. I have fixed this issue.

Additionally, @ckcherry23 brought up a good idea during our CS3281 meeting to add feedback to the user when the copy button is clicked, so I have modified the tooltip to now show "Copied iframe" upon clicking of the copy button, for 2 seconds, before it reverts back to its original text.

ckcherry23:
Hi @MarcusTXK, thanks for your work on this!

The copy iframe function is working correctly and the user feedback is clear. The bug with the wrong summary charts has also been fixed.

While the current implementation seems correct, there are some usability improvements that could make it even better. If you have the time and resources, I would highly encourage you to consider making these changes before merging the code. Else we can leave it for future contributors as per the team's decision.

I've mentioned them here in order of importance: 
* When I see the below iframe on a webpage, it lacks context, as the viewer has no idea what this ramp chart is for. It may be beneficial to include which repo these contributions belong to. If possible, we can add the date range too, but this is not as critical.
image

* Consider disabling the tooltips in the report since they are often only partially visible and not very helpful to the user.
* It would be a nice touch to add a "Generated by RepoSense" footer on the embedded widgets to promote our software.

Moreover, the UG needs to be updated to include a how-to guide for widgets, and the HTML report in the DG should be updated to include the new files added along with their purpose. Please fix the merge conflicts as well.

I will be asking @reposense/active-reviewers for their review so they can add their opinions about this as well.

MarcusTXK:
Thanks alot for reviewing the changes, I really appreciate it.

> When I see the below iframe on a webpage, it lacks context, as the viewer has no idea what this ramp chart is for. It may be beneficial to include which repo these contributions belong to. If possible, we can add the date range too, but this is not as critical.

Currently, the widget was designed with simplicity in mind, so I chose not to include date range, granuality,  etc, as in my opinion, context can be provided by the user when they use the iframe in their website (e.g. in their portfolio website) and further context can be viewed upon clicking on the widget to view the full report. However, if others disagree, we can discuss what are the important contexts to include.

> Consider disabling the tooltips in the report since they are often only partially visible and not very helpful to the user.

For this point I agree, as stated previously, I think it is best fixed in a separate PR, so lets temporarily disable the tooltip for a user.

> It would be a nice touch to add a "Generated by RepoSense" footer on the embedded widgets to promote our software.

I think this is an interesting point. The iframe will have to be made taller so that if this is added, it does not obfuscate any of the important information so the footer can be displayed. I am personally impartial to this, while a user might not appreciate the additional bulk by including the logo, it would definitely help with the promotion of RepoSense, so I think that whether this is added should be based on if the promotion of RepoSense is an important aspect of the widget. 

> UG needs to be updated to include a how-to guide for widgets
Definitely, I will update information about the widget, perhaps in the [Sharing Reports](https://reposense.org/ug/sharingReports.html) section. For updating of the screenshots, I think it can be done in issue #1962, together with all the other frontend changes.

HCY123902:
Great work. I have several suggestions. I agree that it is beneficial to include at least the repository and branch title for Single Ramp and Group by Author Ramp. This provides the readers with the basic details without requiring them to click on the report link.

image

Additionally, I suggest changing the icon for the redirect button, since the current icon collides with the commits panel icon while the redirect button may not necessarily open the commits panel after opening the report.

MarcusTXK:
@HCY123902 Thanks for the feedback, Ive updated the icon to redirect the user to the RepoSense Report
 I have also updated the widget to include the repository and branch title for Single Charts as well as  Group by Author Ramp.
 
Single Chart for group by Repo/Branch:
image

Single Chart for group by Author:
image

docs/ug/sharingReports.md
```diff
@@ -29,10 +29,10 @@ You can point the recipients to the [_**Using reports**_](usingReports.html) sec
 
 As RepoSense reports are in a web page format, you can publish a report by simply uploading it onto any web hosting service. Given below are several options that not only allow to publish reports, but **also enable various levels of automating** the entire process (e.g., automatically update the report daily).
 
+Published reports can additionally be embedded in other websites through `iframes`. Simply click the clipboard icon to generate and copy the iframe for your desired section of the report. Paste the iframe in any HTML supported document to render it.
```
HCY123902, line 32: May I suggest adding a separate section for this to [Sharing Reports](https://reposense.org/ug/sharingReports.html) and including an example for each use case in the new section?

frontend/src/components/c-summary-charts.vue
```diff
@@ -6,12 +6,15 @@
...
     .summary-charts__title(
-      v-if="filterGroupSelection !== 'groupByNone' && !isChartWidgetMode",
+      v-if="filterGroupSelection !== 'groupByNone'",
+      v-bind:ref="'summary-charts-title-' + i",
       v-bind:class="{ 'active-background': \
         isSelectedGroup(repo[0].name, repo[0].repoName) && !isChartGroupWidgetMode}"
     )
       .summary-charts__title--index(v-if="!isChartGroupWidgetMode") {{ i+1 }}
-      .summary-charts__title--groupname
+      .summary-charts__title--groupname(
```
HCY123902, line 15: Can you also try to align the format of the Group by Author titles and single ramp titles with those in the original report

HCY123902 (changes requested):

frontend/src/components/c-summary-charts.vue
MarcusTXK: Apologies, could you clarify what you mean by this? Currently, the title already seems to be already aligned with the text below when in widget mode (by removing the left padding, as the index is no longer shown). Or do you mean to align it in a different way?

image

HCY123902: Oh I mean the font and styling of the title

MarcusTXK: Ah okay, apologies about that, with the recent refactor in #1979, the new mixins were not imported

HCY123902 (approved):

ckcherry23 (commented):
Hi @MarcusTXK, the embedded iframe with the title looks cleaner and almost ready. I noticed a few more issues:

* There is a tooltip for the 'number of lines' shown beside the title (for both `group by author` and `group by repo`). Can you disable this tooltip as well for widget mode?

* It doesn't seem like the copy iframe feature works for ramp charts when using the `None` option for `group by`. I am not sure if you intended this, but I think we should allow embedding ramps even when using the`None` option. Can you recheck or clarify this?

MarcusTXK:
@ckcherry23 Thanks for helping to take a look, appreciate your help!

> There is a tooltip for the 'number of lines' shown beside the title (for both group by author and group by repo). Can you disable this tooltip as well for widget mode?

Thanks for spotting this, I have hidden it for widget mode.

> It doesn't seem like the copy iframe feature works for ramp charts when using the None option for group by. I am not sure if you intended this, but I think we should allow embedding ramps even when using theNone option. Can you recheck or clarify this?

This was an accidental bug introduced with the inclusion of title for single charts, and the title being undefined for `group by None`. Good catch, I have fixed this.

ckcherry23:
> Thanks for spotting this, I have hidden it for widget mode.

I think there might still be an issue with the tooltip
Screenshot 2023-04-22 at 10 22 37 PM

ckcherry23 (approved):
LGTM!

frontend/src/components/c-summary-charts.vue
```diff
@@ -17,20 +24,20 @@
...
-        v-if="!isGroupMerged(getGroupName(repo))",
+        v-if="!isGroupMerged(getGroupName(repo)) && !isChartGroupWidgetMode",
         v-on:click="handleMergeGroup(getGroupName(repo))"
       )
         .tooltip
           font-awesome-icon.icon-button(:icon="['fas', 'chevron-up']")
           span.tooltip-text Click to merge group
       a(
-        v-if="isGroupMerged(getGroupName(repo))",
+        v-if="isGroupMerged(getGroupName(repo)) && !isChartGroupWidgetMode",
```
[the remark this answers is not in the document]
MarcusTXK: Yeah I definitely agree it would be great if we could refactor these repeated individual checks into a common one, though I unfortunately cant think of a clean way to do it at the moment.

PR #1991, opened 2023-04-10
Refactor ArgsParser::parse method

chan-j-d:
### Proposed commit message

```
The ArgsParser::parse method is long and convoluted, with different
lines of code relating to different flags interweaved amongst each
other.

Additionally, there are large chunks of logic performed pertaining to
just a few flags that can be abstracted out into a separate method.

Let's refactor the ArgsParser::parse method to group flag-related code
together and abstract out more complex flag-related logic into separate
methods.
```

### Other information

The bulk of the method relates to just setting the `sinceDate` and `untilDate` for analysis so that has been abstracted out as a separate private method. The `ReportConfiguration` has also been abstracted out as its logic is also somewhat complex. 

The remainder of the flags are grouped into two types.
1. Direct read-offs from the supplied arguments
    - these are all grouped at the top and directly set to the first builder code chain
2. Slightly more complex logic
    - these have been grouped together with the logic for the flag immediately followed by the setting it to the builder

There is also one minor logical change which is the logging of the `MESSAGE_USING_DEFAULT_CONFIG_PATH`. The old if statement was impossible to result in logging as reposense will fail to run if `-c` is used without supplying with an argument. The default config path was actually only used if neither `-r` or `-c` is specified, which would make it use a config run by default with the `DEFAULT_CONFIG_PATH` as the argument for `configFolderPath`.


yhtMinceraft1010X (approved):
LGTM

HCY123902 (commented):
Do you think it is necessary or beneficial to refactor arguments with simpler logic such as `isViewModeOnly` and `isAutomaticallyLaunching` into separate methods as well so that the level of abstraction will be consistent in the `parse` method

chan-j-d:
@HCY123902 I didn't think it was important to do that since those flags were mostly a combination of the other booleans whereas the two I abstracted out had to do additional analysis like opening a file, etc. I'm not opposed to it though and am open to changing it.

HCY123902 (approved):

PR #1992, opened 2023-04-12
[#1967] Use GitHub environmental files instead of set-output

### LINKED ISSUE #1967: Use GitHub environmental files instead of using soon-to-be deprecated set-output

Currently, our `surge.yml` uses `set-output` to extract and save the PR number for use as part of the surge.sh preview domain. However, GitHub will be deprecating `set-output` to "avoid untrusted logged data to use save-state and set-output workflow commands without the intention of the workflow author."

image

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

This feature will be fully disabled by 31st May 2023, which is 2 months from now. 

As described in the article, we can use GitHub environmental files available within an action to save the PR number instead.

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#sending-values-to-the-pre-and-post-actions

sikai00:
Fixes #1967

### Proposed commit message

```
surge.yml currently uses set-output to extract and save the PR number
for use as part of the surge.sh preview domain. However, GitHub will be
deprecating set-output for security reasons.

Let's update surge.yml to use the new GITHUB_OUTPUT environment files
instead.
```

### Other information

The change were as per [GitHub's article](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

While our surge.sh preview is still currently down, the correctness of this PR can be checked against the link produced.

### DIFF
--- a/.github/workflows/surge.yml
+++ b/.github/workflows/surge.yml
@@ -37,7 +37,7 @@ jobs:
 
     - name: Extract PR number
       id: pr-number
-      run: echo '::set-output name=ACTIONS_PR_NUMBER::'$(cat ./pr/NUMBER)
+      run: echo 'ACTIONS_PR_NUMBER=$(cat ./pr/NUMBER)' >> $GITHUB_OUTPUT
 
     - name: Download deployment status artifacts
       uses: dawidd6/action-download-artifact@v2
--- a/.github/workflows/surge.yml
+++ b/.github/workflows/surge.yml
@@ -37,7 +37,7 @@ jobs:
 
     - name: Extract PR number
       id: pr-number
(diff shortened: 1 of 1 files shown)


sikai00:
### Example
I have a [completed workflow](https://github.com/sikai00/RepoSense/actions/runs/4719192310) on my forked repo, which shows the correctness of the PR. The warning no longer shows up.

Note that the workflow is ran using a different commit than the PR, so that the workflow is indeed using the updated line to run. Furthermore I disabled codecov in my repo, as it will fail and prevent the surge workflow from happening.

ckcherry23 (approved):
LGTM!

chan-j-d (approved):
LGTM!

yhtMinceraft1010X (approved):
LGTM

PR #2007, opened 2023-06-18
[#1881] Check if author config file exists instead of detecting it via an exception

### LINKED ISSUE #1881: Check if author config file exists instead of detecting it via an exception
Currently, for both detection of author config and group config, it relies on running the code normally as if the file is always present and catching the `FileNotFoundException` to determine if the file is not present. 

https://github.com/reposense/RepoSense/blob/c162ff20d9c055bddb66ad2f57b3a62906b541d2/src/main/java/reposense/RepoSense.java#L148-L155

It would be preferable to start with an `if` statement that checks if a file exists at the file path before attempting to parse it. Something similar can be done for the group config parsing a few lines lower.

germainelee02:
Fixes #1881

```
Currently, for both detection of author config and group config, it relies 
on running the code normally as if the file is always present and 
catching the FileNotFoundException to determine 
if the file is not present.

Let's refactor the code to detect if the file does not exist using 
an `if` block instead of catching it in the try-catch 
block
```

### DIFF
--- a/src/main/java/reposense/model/ConfigRunConfiguration.java
+++ b/src/main/java/reposense/model/ConfigRunConfiguration.java
@@ -1,7 +1,8 @@
 package reposense.model;
 
-import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


sikai00 (changes requested):
Thanks for doing this up. The changes looks good to me.

1. We include the issue that we are fixing in the title so that it is easier to track from the pull requests list. So something like [#1881] Check if... in your case.
2. For the commit message, keep to 80 characters per line.

Other than that LGTM!

sikai00 (approved):
LGTM

PR #2009, opened 2023-06-23
[#1968] Incorrect dropdown box for config in UG: Customizing Report

### LINKED ISSUE #1968: Incorrect dropdown box for config in UG: Customizing Report
### What problem(s) did you run into that caused you to request additional documentation? What, if any, existing documentation relates to this proposal?

Under this section https://docs-1950-pr-reposense-reposense.surge.sh/ug/customizingReports.html#customize-using-csv-config-files 
image
I believe it should be showing the config syntax instead of the assets one. 

### What would you like added or modified in the documentation?

Fix the documentation to show the config flag instead of the assets one.
Additionally, I believe it should also include a link to the config files format page in some form.

germainelee02:
Fixes #1968

```
There is a bug in the UG where the assets syntax is incorrectly
displayed under the config section instead of the correct config
syntax. This could potentially confuse users.

Let's fix the bug by displaying the correct config syntax, and add a
link to the page that explains the format of config files, for user
convenience.

```
https://docs-1950-pr-reposense-reposense.surge.sh/ug/customizingReports.html#customize-using-csv-config-files

### DIFF
--- a/docs/ug/cli.md
+++ b/docs/ug/cli.md
@@ -56,7 +56,7 @@ The section below provides explanations for each of the flags.
 
 * Cannot be used with `--repos`. The `--repos` flag will take precedence over this flag.
 * If both `--repos` and `--config` are not specified, RepoSense looks for config files in the `./config` directory.
-* Config files must follow [this](./#configFiles.html) format.
+* Config files must follow [this](../#configFiles.html) format.
 </box>
 </div>
 
--- a/docs/ug/cli.md
+++ b/docs/ug/cli.md
@@ -28,7 +28,7 @@ The section below provides explanations for each of the flags.
 
 ### `--assets`, `-a`
 
-<div id="section-config">
+<div id="section-assets">
 
 **`--assets ASSETS_DIRECTORY`**: Specifies where to place assets for report generation.
 * Parameter: `ASSETS_DIRECTORY` The directory containing the assets files. A `favicon.ico` file can be placed here to customize the favicon of the dashboard.
--- a/docs/ug/cli.md
+++ b/docs/ug/cli.md
@@ -56,7 +56,7 @@ The section below provides explanations for each of the flags.
 
 * Cannot be used with `--repos`. The `--repos` flag will take precedence over this flag.
 * If both `--repos` and `--config` are not specified, RepoSense looks for config files in the `./config` directory.
-* Config files must follow [this](#configFiles.html) format.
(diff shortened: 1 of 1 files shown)


docs/ug/cli.md
```diff
@@ -56,6 +56,7 @@ The section below provides explanations for each of the flags.
 
 * Cannot be used with `--repos`. The `--repos` flag will take precedence over this flag.
 * If both `--repos` and `--config` are not specified, RepoSense looks for config files in the `./config` directory.
+* Config files must follow [this](#configFiles.html) format.
```
vvidday, line 59: Is the link pointing to the correct path? I think the `#` is for linking to sections on the current page instead of other pages.

vvidday (changes requested):
Thanks for your work on this @germainelee02!

Also have a minor comment on the commit message - we usually use `Let's ...` for the changes being done in the PR, and use an imperative mood instead of past tense. Also, it would be good to include a short statement on "why it needs to change" (even though it's quite obvious in this case).

[More info under 'commit message body'](https://se-education.org/guides/guidelines/PRs.html#submitting-a-pr)

Something like the following (feel free to change as you see fit!):
```
There is a bug in the UG where the assets syntax is incorrectly
displayed under the config section instead of the correct config
syntax. This could potentially confuse users.

Let's fix the bug by displaying the correct config syntax, and add a
link to the page that explains the format of config files, for user
convenience.
```

docs/ug/cli.md
germainelee02: Thanks for pointing it out! Would an absolute path like this (https://reposense.org/ug/configFiles.html) work better?

vvidday: That might work on the production site, but may not be too ideal as it won't point to the corresponding page when deployed on different environments (e.g. localhost, or the surge preview deployments), and would also need to be changed if we change the domain in the future.

Markbind [has a feature](https://markbind.org/userGuide/formattingContents.html#intra-site-links) to get around this problem by specifying absolute paths as `{{ baseUrl }}/path`. Alternatively, it might be worth looking into if you can use relative paths for pages (i.e, removing the `#` in the link).

vvidday (approved):
LGTM!

MarcusTXK (approved):
LGTM, nice job!

chan-j-d (approved):
LGTM!

PR #2010, opened 2023-06-24
[#1899] Add graphic for commit diffstat

### LINKED ISSUE #1899: Add graphic for commit diffstat in Commits Panel
Along with each commit on the Commits Panel, show the number of insertions and deletions with a simple visualization. This can be placed next to the LoC changed count.

### Purpose
This graphic must be useful to get a high-level overview of the *impact* of a commit:
* Compare the ratio of additions/deletions per commit
* Compare commit sizes across commits

#### 1. GitHub style: Fixed-width
* Histogram-style visualization of the number of additions and deletions in a fixed-width bar. 
* The number of green and red boxes is calculated based on the ratio of additions to deletions.
* Due to the fixed size of each graphic, we cannot compare commit sizes across commits. 
* Example:
image
[Image source](https://www.google.com/url?sa=i&url=https%3A%2F%2Fstackoverflow.com%2Fquestions%2F41956260%2Fgit-diff-get-files-changes-with-type-of-change-additions-and-deletions&psig=AOvVaw1t3bIdZ8mZREP_oCwQ56TF&ust=1676020119005000&source=images&cd=vfe&ved=0CBAQjRxqFwoTCJixvtGLiP0CFQAAAAAdAAAAABBF)

#### 2. cgit style: Proportional to commit size
* (Right-aligned) Contribution bar that visualizes the number of additions and deletions. 
* The size of each bar should be calculated based on the actual commit size, rather than ratio between additions/deletions. 
* Due to the varying size of each graphic, it may introduce some challenges in size calculations or alignment (responsiveness). Good to define a maximum width.
* One large commit (like package.json) may make other significant commits seem too small.
* Example:
image
[Image source](https://www.google.com/url?sa=i&url=https%3A%2F%2Fwww.reddit.com%2Fr%2FAmd%2Fcomments%2Fo7i1zl%2Fphoronix_amd_sfh_linux_driver_updated_for_next%2F&psig=AOvVaw1t3bIdZ8mZREP_oCwQ56TF&ust=1676020119005000&source=images&cd=vfe&ved=0CBAQjRxqFwoTCJixvtGLiP0CFQAAAAAdAAAAABAl)

### Proposed solution
I believe the contribution bar proportional to commit size meets most of our requirements, though it may require more work to implement. We could also hide the contribution bar when the commit messages are hidden, to reduce clutter. Please do add any alternative suggestions or thoughts!

nseah21:
Fixes #1899 

### Add graphic for commit diffstat 

```
A diffstat graphic is useful in providing a visual representation
of the changes within and between commits. 

Let's add a diffstat graphic emulating the diffstat used by cgit, 
to help users have a high-level overview of the impact of a
commit. 
```
### Other information

NA


damithc:
@nseah21 Thanks for taking this up. I personally prefer the other alternative (i.e., similar to the contribution bar), which is more consistent with RepoSense. Plus RepoSense is supposed to complement what GitHub does, rather than duplicate it.
We can show the bar as a separate line, similar to how we do in the chart view.

ckcherry23:
Hi @nseah21, your code changes look great for a first attempt at creating the GitHub style diffstat! 
However, since we prefer the contribution bar (cgit) style, let us start with that as our first/default.

The reason we prefer the cgit style is that we can reuse the same code we have for the contribution bars below each ramp chart for creating the diffstat as well.

https://github.com/reposense/RepoSense/blob/3878389c8ea7891f12feeeb373c575d1003c496a/frontend/src/components/c-summary-charts.vue#L221-L242

While you can copy over the same code for the diffstat, I would suggest that we move out this contribution bars code outside of `c-summary-charts.vue` into a separate file called `c-stacked-bar-chart.vue` and generalize it to be used by both the contribution bars and the diffstat. @vvidday can give his thoughts as well.

We can discuss whether we want a toggle mode for the 2 styles in a separate PR, where we can make use of the current set of changes.

Feel free to reach out to us if you have any questions with the `c-stacked-bar-chart.vue` implementation and we will be glad to help!

nseah21:
@damithc @ckcherry23 Thank you for your comments. I have pushed some changes based on the initial feedback. 

I will look into using the code in `c-summary-charts.vue` for this feature.

damithc:
@nseah21 Good progress. A few comments:

1. The bar should be at the be below the tags, not between title and tags.
2. The same scale should be used for all commits in a report (not just one author) so that a visual comparison is possible. A 100 line commit of person A should be the same length as that of person B. 
3. For consistency (similar to ramp charts view), the bar should overflow to the next line if too long to show in one line. That way, very big commits will stand out by appearing as multiple lines.

vvidday:
Great job so far and thanks for taking up this issue! Definitely agree with @ckcherry23 on re-using some of the code for the contribution bar - I believe it also has some logic that will help with what Prof mentioned above (overflow & scale).

Also, as we are trying to unify the frontend to use TypeScript (#1936), it would be great if all new features/files could also be started in TypeScript. This can be done by adding `lang="ts"` inside the opening `<script>` tag, and [wrapping the export in `defineComponent`](https://vuejs.org/guide/typescript/overview.html#definecomponent).

vvidday (changes requested):
Great job @nseah21 with the progress! This wasn't an easy issue to take up and your progress on this has been impressive. The functionality works well, I left some comments w.r.t implementation details.

Regarding your questions, I think (1) can go either way - I don't think we necessarily need to follow the average across report like in summary, since that view contains charts across the whole report but the zoom view is restricted to a repo. Not too sure about (2), the non-linear scaling might be a bit misleading visually? Would love to get more opinions on these 2 questions!

frontend/src/components/c-stacked-bar-chart.vue
```diff
@@ -0,0 +1,79 @@
...
+        v-for="(barWidths, fileType) in widthMappings"
+        )
+        .summary-chart__contrib--bar(
+          v-for="width in barWidths",
+          v-bind:style="{ width: `${width}%`,\
+          'background-color': fileTypeColors[fileType] }",
+          v-bind:title="`${fileType}: ${fileTypeLinesChanged} lines, \
+          total: ${totalLinesChanged} lines (contribution from ${minDate} to \
+          ${maxDate})`"
+          )
```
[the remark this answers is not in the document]
nseah21: I agree that the bar chart component can be better abstracted out to reduce coupling between the parent component and itself. 

I will try making some changes based on your suggestions.

frontend/src/views/c-zoom.vue
```diff
@@ -5,12 +5,12 @@
...
-      v-on:click="toggleAllCommitMessagesBody(true)"
-    ) show all commit messages
+      v-on:click="toggleAllCommitMessagesBody(true); toggleDiffstatView(true);"
+    ) show all commit details
     a(
       v-if="expandedCommitMessagesCount > 0",
-      v-on:click="toggleAllCommitMessagesBody(false)"
-    ) hide all commit messages
+      v-on:click="toggleAllCommitMessagesBody(false); toggleDiffstatView(false);"
+    ) hide all commit details
```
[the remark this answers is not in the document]
nseah21: I feel that the individual toggling of the commit message body should not also toggle the visibility of the diffstat. This is because the purpose of the diffstat is to provide the user with a "bird's eye view" of the changes in a commit in relation to one another. Users already zooming in on a particular commit can refer to the lines changed.

Therefore, toggling each of the diffstats individually might not be necessary, save for the (rare) scenarios where diffstats occupying many lines may need to be hidden to reduce clutter.

In these cases, perhaps it would be better to add a separate individual toggle button for showing/hiding the diffstat, so that users can still tell, based on the presence of the existing toggle button, which commits have message bodies.

frontend/src/views/c-zoom.vue
```diff
@@ -478,6 +541,7 @@ export default defineComponent({
   /* Commit Message Body in Zoom Tab */
   .commit-message {
     border: 1px solid transparent;
+    overflow: hidden;
```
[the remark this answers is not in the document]
nseah21: Thank you for pointing this out. I will try to find a fix that still shows the tooltip.

Previously, some diffstats were spilling over onto the next commit, causing the contribution bar to "stick" to the commit title of the next commit.

image

damithc:
@nseah21 remember to let reviewers know when the PR is ready for the next round of reviews.

nseah21:
I have attempted some changes for this feature based on the feedback. The PR is ready for the next round of reviews.

frontend/src/components/c-stacked-bar-chart.vue
```diff
@@ -0,0 +1,23 @@
+<template lang="pug">
+#summary
+  .summary-chart__contrib--bar(
```
vvidday, line 3: Since this is no longer only linked to `summary-charts`, would it be possible to rename the class here?

You might have to update some test cases since I believe they are using the `.summary-chart__contrib--bar` selector.

frontend/src/views/c-zoom.vue
```diff
@@ -132,6 +132,16 @@
...
+      template(
+        v-if="showDiffstat"
+        v-for="(widths, color) in getContributionBars(slice)"
+      )
+        c-stacked-bar-chart(
+          v-bind:bars="widths.map((width) => {\
+            return {width: width,\
+                    color: color}\
+          })"
+        )
```
vvidday, line 144: Since the `c-stacked-bar-chart` component is now accepting an array of bars, I don't think it makes sense to render two different stacked-bar-chart components (one for green, one for red) - or in the case of the `checkedFileTypeContributionBars` in `c-summary-charts`, rendering different stacked-bar-charts for each file type.

Can the `getContributionBars` and `getFileTypeContributionBars` be refactored to return all the bars needed for the chart instead?

 i.e. instead of `getContributionBars` returning `{ red: [40], limegreen: [60, 80] }` and constructing a different `stacked-bar-chart` for red and limegreen, can it instead return 
```typescript
[
    { color: "red", width: 40 },
    { color: "limegreen", width: 60 },
    { color: "limegreen", width: 80 },
]

**Review by vvidday — CHANGES_REQUESTED** [review]
Thanks for the changes @nseah21 and apologies again for the late reply.

Will reply regarding the CSS overflow issue in the old thread above.

Also, before this pr is merged, it will be good to add cypress test cases for the new feature.

File: frontend/src/views/c-zoom.vue (hunk 3 above)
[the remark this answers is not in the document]
Comment (vvidday) [reply frontend/src/views/c-zoom.vue]: Think this is due to the `float: left` property on the contribution bars which makes the parent container have a height of 0, resulting in the elements below not being pushed down.

From searching online the solutions seem to be either using `overflow: hidden` as you've tried (but results in our tooltip being hidden), or some CSS tricks (clearfix?)

Alternatively, it might be worth trying to remove the `float: left` and find an alternative method to position the bars, so the bars will actually have the correct height. Since the bars are just divs with a `width: xx%` property, maybe putting them in a flex container with flex wrap will work?

**Review by ckcherry23 — COMMENTED** [review]
The new changes look great @nseah21! I agree with David's comments and have a few more to add.

Regarding your questions,
For (1), according to prof's earlier comments, I think we should consider the average contribution size across the entire report.
> 2. The same scale should be used for all commits in a report (not just one author) so that a visual comparison is possible. A 100 line commit of person A should be the same length as that of person B.

For (2), I also think the length of the diffstat should scale linearly as it's easier to comprehend for users.

---

Also, I think we should avoid the use of default red-colored bars on the summary chart contribution bars so that they are not confused with the deletions bar in the commit diffstat. I'd suggest using a shade of blue as the default for contribution bars.

Screenshot 2023-07-08 at 3 56 32 PM

File: frontend/cypress/tests/chartView/chartView_mergeCommits.cy.js [hunk 6]
```diff
@@ -1,7 +1,7 @@
 describe('include merge commits in chart view', () => {
   it('show merge commits in summary chart', () => {
     // ramp chart should have merge commit slices
-    cy.get('[title="[2023-03-03] Merge branch \'new-branch\' into cypress: +0 -0 lines "]')
+    cy.get('[title="[2023-03-04] Merge branch \'new-branch\' into cypress: +0 -0 lines "]')
```
ckcherry23, line 4: Thanks for taking a look into this @nseah21. The date `2023-03-03` may not pass Cypress tests in your local machine but is the date that is required for the CI frontend tests to pass. The discrepancy is probably due to a time zone issue as the commit was made close to midnight. Perhaps we can create a new issue to investigate why the date varies on different machines.

frontend/cypress/tests/zoomView/zoomView_contributionBars.cy.js
ckcherry23, line 1: Thanks for starting with the Cypress tests for this feature @nseah21. I think this file can be renamed to something like `zoomView_diffstat.cy.js`. 

Apart from checking if the diffstat has rendered, we can also check:
1. if both green and red bars are rendered for a general commit with > 0 insertions and deletions.
2. if the contribution bar is not visible for empty commits with 0 insertions/deletions.
3. if the size of the green/red bars is proportional to the insertion/deletion size. I think a naïve way to do this is to check whether the green bar length is greater or less than the red bar length depending on the contribution size. 

@vvidday

ckcherry23 (changes requested):
Thanks for the quick iteration! I've left a few comments to address

frontend/src/views/c-zoom.vue
```diff
@@ -300,6 +310,72 @@ export default defineComponent({
...
+      // Pad with transparent bars to prevent overflowing bars
+      // from sticking to message title of subsequent commits
+      /*
+        const transparentBars = [];
+        transparentBars.push(100 - currentBarWidth);
+        for (let i = 0; i < 3; i += 1) {
+          transparentBars.push(100);
+        }
+        allContributionBars.transparent = transparentBars;
+      */
```
vvidday, line 375: Can this be removed?

frontend/cypress/tests/zoomView/zoomView_diffstat.cy.js
```diff
@@ -0,0 +1,89 @@
...
+    cy.get('#tab-zoom .commit-message')
+      .first()
+      .within(() => {
+        cy.get('.stacked-bar__contrib--bar')
+          .then((element) => {
+            expect(element.length).to.be.equal(2);
+            expect(element[0].style['background-color']).to.be.equal('limegreen');
+            expect(element[0].style.width).to.be.equal('0.1%');
+            expect(element[1].style['background-color']).to.be.equal('red');
+            expect(element[1].style.width).to.be.equal('0.1%');
```
vvidday, line 27: Would be good to add a short comment here regarding the assumptions that this test makes (like [this](https://github.com/reposense/RepoSense/blob/master/frontend/cypress/tests/zoomView/zoomView_rampChart.js)).

frontend/cypress/tests/zoomView/zoomView_diffstat.cy.js
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  it('should render empty contribution bars for commits with no changes', () => {
+    cy.get('.icon-button.fa-list-ul')
+      .should('be.visible')
+      .first()
+      .click();
+
+    cy.get('#tab-zoom .commit-message')
+      .eq(1)
```
vvidday, line 39: same here re: assumptions

vvidday (changes requested):
Thanks @nseah21, overall LGTM, just some minor things to tie up

vvidday (approved):
LGTM!

ckcherry23 (approved):
Great work, LGTM! Just remember to update your commit message with the latest changes and follow the guidelines.

nseah21:
I have updated the commit message accordingly. 

@ckcherry23

PR #2015, opened 2023-07-02
 [#1931] Update frontend integration tests for summary and zoom view

### LINKED ISSUE #1931: Update frontend integration tests for summary and zoom view

On the frontend, changes in the summary panel affect the zoom panel as well (e.g. changing the date range, changing the zoom range affect the commits shown in the zoom panel). 

However, this functionality is not well tested in our current Cypress tests - for example, in `chartView_zoomFeature.js`, the test only checks if the zoom panel is visible after changing the zoom region, but does not check that the commits in the zoom panel actually correspond to the new zoom region.

It would be good to add test cases to test for this functionality.

Other test cases, such as in #1930, rely on the correctness of this functionality between the summary and zoom panels.

We can add additional Cypress test cases for:
- Check that changing the zoom range in the summary panel correctly filters the commits in the zoom panel
- Check that changing the date range in the summary panel correctly filters the commits in the zoom panel

Ideally, these test cases wouldn't rely on any order/assumptions of the data, and thus won't break due to changes to the `cypress` branch.

germainelee02:
Fixes #1931 

```
On the frontend, changes in the summary panel affect the zoom 
panel. However, this functionality is not well tested in our current 
Cypress tests.

Let's add additional Cypress tests for the changing of zoom and date 
range.
```


frontend/cypress/tests/chartView/chartView_zoomFeature.cy.js
```diff
@@ -39,3 +39,260 @@ describe('zoom features in code view', () => {
...
+      .first()
+      .click();
+
+    cy.get('#tab-zoom')
+      .should('be.visible');
+
+    cy.get('#tab-zoom .ramp .ramp__slice')
+      .last()
+      .invoke('attr', 'title')
+      .should('eq', '[2018-06-12] Setup AppVeyor CI (#142): +19 -0 lines ');
```
vvidday, line 63: For the date tests, the functionality/requirement we're interested in testing is that changing the date in the summary view will filter the commits in zoom to only be between the start/end dates. So instead of asserting that the last commit is this specific commit, would it be better to do something like loop through the commits and asserting that their date falls between the since/until dates? 

This would also make this test not rely on the correctness of the sort.

frontend/cypress/tests/chartView/chartView_zoomFeature.cy.js
```diff
@@ -39,3 +39,260 @@ describe('zoom features in code view', () => {
...
+      .invoke('attr', 'title')
+      .should('eq', '[2019-03-08] [#587] Fix unoriented output messages (#593): +105 -69 lines ');
+
+    cy.get('#tab-zoom .ramp .ramp__slice')
+      .last()
+      .invoke('attr', 'title')
+      .should('eq', '[2018-08-27] [#311] code view: differentiate untouched code more (#314): +12 -5 lines ');
+  });
+
+  it('changing the \'until\' and \'since\' date again will result in a different zoom view', () => {
```
vvidday, line 162: Could the name of these two tests be more specific in what they are testing so that they aren't identical?

vvidday (changes requested):
Thanks for taking this up @germainelee02 and apologies for the late response.

I understand that the nature of these tests might make it necessary to make certain assumptions about the data in the report which might result in some hardcoding of the test actions/values. 

Would it be possible to indicate whenever there's an assumption regarding the data in the comments like [this](https://github.com/reposense/RepoSense/blob/master/frontend/cypress/tests/zoomView/zoomView_rampChart.js)? Since the tests that rely on assumptions/hardcoded data might fail in the future if the test repo gets more commits, and it would be easier for future maintainers to know exactly what assumptions are being made in the tests.

vvidday (approved):
LGTM

ckcherry23 (approved):
Nice work on abstracting out common functionalities and the test coverage is also good.

germainelee02:
> Nice work on abstracting out common functionalities and the test coverage is also good.
> 
> Would it be possible to add test cases for some more functionalities?
> 
> * setting the 'granularity' changes the zoom view
> * setting the 'granularity' again results in a different zoom view
> * filtering by a certain file type changes the zoom view
> * filtering by a different file type results in a different zoom view

Should I make a separate PR for that then?

ckcherry23:
> Should I make a separate PR for that then? @ckcherry23

@nseah21 I think this can be done in the same PR, as they have the same purpose of testing whether changes in the summary panel are reflected in the zoom panel.

ckcherry23 (approved):
@germainelee02 I will approve this for now, and add a new issue for the suggested test cases. 
Please take up that issue if you would like.

Issue #2016, opened 2023-07-02
Breakdown by file type not retained after refresh


germainelee02:
### Please include the steps to reproduce the bug.

1. Open a report
2. Check 'breakdown by file type'
3. Check 'all'
4. Refresh

### What was expected to happen?
The checked files should be retained after refreshing.

### What actually happened? Please include a screenshot of the output.
After each refresh, the files become unchecked one by one

https://github.com/reposense/RepoSense/assets/97350904/2492c5f6-63b4-44af-bfc8-83ea6ac66f6f

ckcherry23:
To anyone working on this issue, please add Cypress test cases and check that the test cases in frontend/cypress/tests/codeView/codeView_renderFilterHash.cy.js have the expected behaviour.

PR #2017, opened 2023-07-06
[#1986] Add cypress tests for renderFilterHash

### LINKED ISSUE #1986: Add cypress test for `renderFilterHash()`

Let's add cypress tests for `renderFilterHash()` to ensure that the parameters in the URL are loaded as per expected in the future for early detection and prevention of such bugs. This could be done by calling `cy.reload()` and ensuring all the previously set data remains as configured.

There were a few instances of the parameters in the URL not being loaded properly due to bugs introduced in the code. However, due to a lack of cypress tests, it was not able to be detected early. Adding the relevant cypress tests for this will help with early detection and prevention of bugs such as #1982 and #1894.

nseah21:
Fixes #1986 

### Add cypress tests for renderFilterHash

```
There were a few instances of URL parameters being incorrectly
loaded due to bugs in the code. 

These bugs were not able to be detected early due to a lack of 
tests for the renderFilterHash method.

Let's add tests to check that the URL parameters are loaded
correctly after the application reloads.
```

### Other information

NA


ckcherry23 (commented):
Thanks for taking this up @nseah21 and sorry for the late review. 
I think the test cases you have added now seem pretty comprehensive. 

However, we may want to wait until issue #2016 is fixed before adding a test case for that scenario within this PR. This same issue is also apparent during the test case `'checked file types: url params should persist after change and reload'` as the `yml` option gets unchecked by itself after reload.

ckcherry23 (approved):
Since it is taking a while for issue #2016 to be resolved, I will approve this PR and update that issue regarding these Cypress tests. Thanks @nseah21 for your work!

MarcusTXK (approved):
LGTM, merging this in so that more tests can be added in #2086.

PR #2021, opened 2023-07-11
[#1943] Make ErrorSummary non-static by removing singleton design pattern 

### LINKED ISSUE #1943: Make ErrorSummary non-static by removing singleton design pattern

germainelee02:
Fixes #1943 

```
Currently, ErrorSummary implements the singleton design pattern.
However, ErrorSummary have to be made non-static if we 
want to parallelize/have concurrent runs of RepoSense.

Let's modify ErrorSummary to be non-static.
```

### DIFF
--- a/docs/dg/learningBasics.md
+++ b/docs/dg/learningBasics.md
@@ -226,7 +226,7 @@ Here are some small tasks for you to gain some basic knowledge of the code relat
   Add this to the catch block of `spawnCloneProcess` and `waitForCloneProcess`, so that the message will be captured in `summary.json`.
 
   ```
-  ErrorSummary.getInstance().addErrorMessage(config.getDisplayName(), e.getMessage());
+  new ErrorSummary().addErrorMessage(config.getDisplayName(), e.getMessage());
   ```
   </panel>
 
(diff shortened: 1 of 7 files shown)


germainelee02:
Close PR after looking at this: https://github.com/reposense/RepoSense/pull/1947#issuecomment-1466624134

PR #2034, opened 2023-08-31
[#2027] Fix date range bug

### LINKED ISSUE #2027: Commits panel not displaying commits at end of date range
### Tell us about your environment

* **RepoSense Version:** v2.5 (also latest dev build)
* **OS and Version:** Windows 11 Version 22H2
* **Web Browser and Version (if applicable):** Google Chrome Version 115.0.5790.111

1. Generate a report with only the **--repo** flag specified (flag likely does not matter).
2. Get the date of any commit. (In the screenshot below, it is 2020-09-16)
image
3. Set the date in the until field to the date obtained in step **2**.
image
4. Using the shortcut (hold CTRL while selecting a range of commits to display in the commits panel), select a range that includes the commit chosen in step **2**.
image
5. Commits panel does not include the commits made on the same date as obtained in step **2**.
image

Commits panel should display the blue commits selected in step **4**.

Commits panel does not display the blue commits.
image

### If possible, include the URL to your RepoSense report or log files (if any).
RepoSense report used in the above demonstration can be found [here](https://nus-cs2103-ay2021s1.github.io/ip-dashboard/?search=&sort=groupTitle&sortWithin=title&since=2020-08-14&until=2020-09-16&timeframe=commit&mergegroup=&groupSelect=groupByRepos&breakdown=false&tabOpen=true&tabType=zoom&zFR=false&zA=AaronnSeah&zR=AaronnSeah%2Fip%5Bmaster%5D&zACS=253.51968503937007&zS=2020-09-13&zFS=&zU=2020-09-15&zMG=true&zFTF=commit&zFGS=groupByRepos).

jq1836:
Fixes #2027 

### Proposed commit message

```
Currently, users are unable to select a zoom range that includes the until date.

This results in misleading data being presented to users.
```

### Other information

Upon further investigation, the bug is caused by inaccurate logic in the `openTabZoomSubrange` method in `c-summary-charts.vue`.

```
if (this.drags.length === 2 && this.drags[1] - this.drags[0]) {
        const tdiff = (new Date(this.filterUntilDate)).valueOf() - (new Date(this.filterSinceDate)).valueOf();
        const idxs = this.drags.map((x) => (x * tdiff) / 100);
        const tsince = window.getDateStr(new Date(this.filterSinceDate).getTime() + idxs[0]);
        const tuntil = window.getDateStr(new Date(this.filterSinceDate).getTime() + idxs[1]);
        this.drags = [];
        this.openTabZoom(user, tsince, tuntil, isMerged);
}
```
In line 2, the const `tdiff` is calculated as the time difference in ms between the `filterUntilDate` and the `filterSinceDate`. This is inaccurate as the after creation of each `Date` object using the arguments `this.filterUntilDate` and `this.filterSinceDate`, which results in `Date` objects with the correct date, but at 00:00 (GMT+0000). This results in the date represented by `this.filterUntilDate` not being accounted for when deciding which regions of the ramp chart would be considered a certain date. As such, by adding a day worth of time in ms to `tdiff`, we can adjust for this.

### Screenshot of GUI after fix:
image

### DIFF
--- a/frontend/src/components/c-summary-charts.vue
+++ b/frontend/src/components/c-summary-charts.vue
@@ -535,7 +535,9 @@ export default defineComponent({
 
       // skip if accidentally clicked on ramp chart
       if (this.drags.length === 2 && this.drags[1] - this.drags[0]) {
-        const tdiff = (new Date(this.filterUntilDate)).valueOf() - (new Date(this.filterSinceDate)).valueOf();
+        // additional day was added to include the date represented by filterUntilDate
+        const tdiff = (new Date(this.filterUntilDate)).valueOf() - (new Date(this.filterSinceDate)).valueOf()
+            + window.DAY_IN_MS;
         const idxs = this.drags.map((x) => (x * tdiff) / 100);
         const tsince = window.getDateStr(new Date(this.filterSinceDate).getTime() + idxs[0]);
         const tuntil = window.getDateStr(new Date(this.filterSinceDate).getTime() + idxs[1]);


damithc:
@jq1836 is this ready for review?

jq1836:
Yep, this PR is ready to be reviewed.

ckcherry23:
Hi @jq1836, thanks for working on this. Adding a day's worth of time will definitely help in providing a more accurate representation of the commits within a zoom range compared to what we have now. 

To update the @reposense/active-reviewers-1 team, I have added some screenshots of the new behaviour of the zoom date range. In the first scenario (After), the zoom range has exactly those commits that are included within our selection. But in the second scenario (After), selecting a few commits of the day selects all of those commits plus a few more after them (which is better than no commits at all). This is not the ideal case, so please respond if you have any objections.

### Scenario 1: Single commit in a day

#### Before
Screenshot 2023-09-19 at 1 05 44 PM

#### After
Screenshot 2023-09-19 at 1 06 26 PM

### Scenario 2: Multiple commits in a day

#### Before
Screenshot 2023-09-19 at 1 06 05 PM

#### After
Screenshot 2023-09-19 at 1 06 33 PM

ckcherry23 (approved):

jq1836:
Something to note would be that the granularity of the zoom window, is currently on a day by day basis, i.e. if the region selected contains the day, even if it were a small part of the day, the zoom window would display all commits made on that day.

Note that it is possible to alter the behavior of the zoom window such that the zoom selection more accurately chooses a time period for commits to display, for e.g. by the hour instead of by the day. Though such specificity provides minimal convenience to the user, apart from extreme cases with many commits within the day, with the user wanting to analyse commits made during a certain period of the day, which can still be done to a smaller extent with the current implementation, by zooming into the day and looking at the relative positions of the commits.

vvidday (approved):
Thanks for the work on this @jq1836, I think the fix of adding a day is fine - something similar is also done when calculating the positions in the ramps (https://github.com/reposense/RepoSense/blob/master/frontend/src/components/c-ramp.vue#L158), and the behaviour of the zoom view showing all the commits on the day even though not all commits are visually within the drag selection is an existing behaviour and not caused by the changes in this PR, as you mentioned.

Just one minor comment about the commit message - it would be nice to include a brief elaboration on why the change is being made (e.g. what benefit will fixing this bug bring/ why the fix needs to be made), even though it might seem obvious.

PR #2035, opened 2023-09-13
[#1936] Migrate c-segment.vue to typescript

jq1836:
Part of #1936 

### Proposed commit message

```
Currently, there is still some JavaScript code which remains 
unmigrated. This allows for type unsafe code to be written, 
potentially resulting in unintended behavior.

Let's migrate the rest of the JavaScript code to TypeScript 
code to facilitate future changes to the code.
```

### Changes made
* Converted JavaScript code in `c-segment.vue` to TypeScript.
* Modified potentially dangerous code (see below).

### Dangerous code
The following code makes use of the behavior of the `||` operator where the result will be the first operand given that the first operand is not `false` or `null` or any value that is represented by the value `0`. In this case, `this.segment.knownAuthor` has been declared to be `string` or `null` in `segment.ts`. The consequence of this would be that the result, given that the first operand is a `string` that is not the empty string, `''`, would be the `string` itself. Due to common naming conventions, one would expect `isOpen` to be of type `boolean`, however, in actuality, it would be a `string`, hence the potentially dangerous behavior which currently works as intended, but may not in the future with more features relating to `segment` being added. 

The same can be said for `canOpen`, however, it is not as dangerous due to the `&&` operator taking into consideration both operands and returning a `boolean` value only. But in order to prevent future misunderstandings where someone mistakenly assumes that `this.segment.knownAuthor` is a `boolean` value from its use in the `canOpen` data field, I have opted to disambiguate this.
```
data() {
    return {
      isOpen: this.segment.knownAuthor || this.segment.lines.length < 5,
      canOpen: this.segment.knownAuthor && this.segment.lines.length > 4,
      transparencyValue: '30' as string,
    };
},
```

### DIFF
--- a/frontend/src/components/c-segment.vue
+++ b/frontend/src/components/c-segment.vue
@@ -31,11 +31,13 @@
     )
 </template>
 
-<script>
+<script lang='ts'>
+import { defineComponent } from 'vue';
 import { mapState } from 'vuex';
 import Segment from '../utils/segment';
+import { StoreState } from '../types/vuex.d';
 
-export default {
+export default defineComponent({
   name: 'c-segment',
   props: {
     segment: {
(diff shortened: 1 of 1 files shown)


frontend/src/components/c-segment.vue
```diff
@@ -49,22 +51,22 @@ export default {
   },
   data() {
     return {
-      isOpen: this.segment.knownAuthor || this.segment.lines.length < 5,
-      canOpen: !this.segment.knownAuthor && this.segment.lines.length > 4,
-      transparencyValue: '30',
+      isOpen: (this.segment.knownAuthor !== null) || this.segment.lines.length < 5 as boolean,
+      canOpen: (this.segment.knownAuthor === null) && this.segment.lines.length > 4 as boolean,
```
ckcherry23, line 55: Thanks for fixing this part of the code with null checks!

ckcherry23 (approved):
Changes LGTM! @jq1836

vvidday (approved):
LGTM, thanks for providing the detailed explanation regarding the changes

PR #2037, opened 2023-09-23
[#1936] Migrate c-ramp.vue to typescript

jq1836:
Part of #1936 

### Proposed commit message

```
unintended behavior.

Let's migrate the rest of the JavaScript code to TypeScript code to
facilitate future changes to the code.
```

### Notable changes made
* Moved `mergeCommitSize` and `deletesContributionRampSize` from `data` to `computed`
* Added null check to `getContributionMessage` (more on this)

#### Current implementation of `getContributionMessage` method
```
// commit of type CommitResult must be provided if tframe === 'commit', only pass null if tframe !== 'commit
getContributionMessage(slice: Commit, commit: CommitResult | null) {
    if (this.tframe === 'commit' && commit !== null) {
      return `[${slice.date}] ${commit.messageTitle}: +${commit.insertions} -${commit.deletions} lines `;
    }

    let title = this.tframe === 'day'
        ? `[${slice.date}] Daily `
        : `[${slice.date} till ${slice.endDate}] Weekly `;
    title += `contribution: +${slice.insertions} -${slice.deletions} lines`;
    return title;
}
```
`getContributionMessage` method presently takes in `Commit` and `CommitResult` arguments, however, in the pug portion of `c-ramp.vue`, the method was dangerously invoked with only the `Commit` argument. (shown below)
```
template(v-else)
    a(v-bind:href="getReportLink()", target="_blank")
      .ramp__slice(
        draggable="false",
        v-for="(slice, j) in user.commits",
        v-bind:title="getContributionMessage(slice)",
        v-on:click="openTabZoom(user, slice, $event)",
        v-bind:class="`ramp__slice--color${getSliceColor(slice)}`",
        v-bind:style="{\
          zIndex: user.commits.length - j,\
          borderLeftWidth: `${getWidth(slice)}em`,\
          right: `${(getSlicePos(tframe === 'day' ? slice.date : slice.endDate) * 100)}%` \
          }"
)
```
Currently, this results in the desired behavior, as `CommitResult` is not used when `this.tframe !== 'commit'`, which is the case where the dangerous invocation of the method is used. To minimise the risk of unintentional misuse of the method, I have split the method into two, `getContributionMessage` and `getContributionMessageByCommit`, with the former taking in only a `slice: Commit` and the latter taking in both `slice: Commit` and `commit: CommitResult`.


damithc:
@reposense/active-reviewers-1 this PR is ready for review

frontend/src/components/c-ramp.vue
```diff
@@ -111,29 +121,31 @@ export default {
...
       }
-      const newSize = 100 * (slice.insertions / this.avgsize);
+      // '+' unary operator here attempts to convert this.avgsize to number, if it is not already one
+      const newSize = 100 * (slice.insertions / +this.avgsize);
       return Math.max(newSize * this.rampSize, 0.5);
     },
-    getContributionMessage(slice, commit) {
-      let title = '';
-      if (this.tframe === 'commit') {
+    // commit of type CommitResult must be provided if tframe === 'commit', only pass null if tframe !== 'commit
```
vvidday, line 128:
```suggestion
    // commit of type CommitResult must be provided if tframe === 'commit', only pass null if tframe !== 'commit'
```

frontend/src/components/c-ramp.vue
```diff
@@ -111,29 +121,31 @@ export default {
...
+    getContributionMessage(slice: Commit, commit: CommitResult | null) {
+      if (this.tframe === 'commit' && commit !== null) {
```
vvidday, line 130: Instead of enforcing that the caller must pass null as the second argument when `tframe` is not `commit`, which might not be very obvious and could potentially be missed out, do you think it would be better to separate the two scenarios into different methods instead?

frontend/src/components/c-ramp.vue
```diff
@@ -111,29 +121,31 @@ export default {
...
-            ? `[${slice.date}] Daily `
-            : `[${slice.date} till ${slice.endDate}] Weekly `;
+      let title = this.tframe === 'day'
+          ? `[${slice.date}] Daily `
+          : `[${slice.date} till ${slice.endDate}] Weekly `;
       title += `contribution: +${slice.insertions} -${slice.deletions} lines`;
       return title;
     },
-    openTabZoom(user, slice, evt) {
+    // Only to be used when this.tframe !== 'Commit'
```
vvidday, line 140: If this shouldn't be called when `tframe` is `Commit`, would it be beneficial to explicitly check for this within the method and abort/exit early? Will it break the application if the method was called when this condition is met?

frontend/src/components/c-ramp.vue
```diff
@@ -111,29 +121,31 @@ export default {
...
+    openTabZoom(user: User, slice: Commit, evt: MouseEvent) {
       // prevent opening of zoom tab when cmd/ctrl click
       if (window.isMacintosh ? evt.metaKey : evt.ctrlKey) {
         return;
       }
 
       const zoomUser = { ...user };
-      zoomUser.commits = user.dailyCommits;
+      zoomUser.commits = user.dailyCommits as Commit[];
```
vvidday, line 148: Not sure if this is a safe cast since `DailyCommit` doesn't have the `insertions` and `deletions` fields. Will this break anything in the zoom view? (since if that file tries to access these properties, it will just be undefined).
Understand that this is existing behaviour and not introduced by this PR, but if there is an existing bug because of this, we should file an issue & fix that

vvidday (commented):
Thanks for your work on this @jq1836, have some minor comments below

frontend/src/components/c-ramp.vue
jq1836: Probably a great idea to do so, as this may cause confusion. Will update to incorporate your suggestion.

frontend/src/components/c-ramp.vue
jq1836: After playing around with removing `openTabZoom` and using it with `this.tframe === 'Commit'`, it seems that `openTabZoom` in `c-ramp` is somehow redundant, at least on my Windows OS. It seems that the logic for opening the Zoom tab is in `c-summary-charts.vue`, under the `openTabZoom` method. As I am unsure of its significance, I will remove the comment and keep the method. Perhaps this can be investigated in a separate issue/PR.

frontend/src/components/c-ramp.vue
jq1836: Similarly, due to the unknown nature of the `openTabZoom` method, this seemingly unsafe cast may not even matter. Seems like removing the statement entirely does not affect the behavior of the program, at least on my Windows OS.

frontend/src/components/c-ramp.vue
vvidday: I can't test right now, but if i'm not wrong the method is called when the granularity is set to "day" or "week" (one of the drop-downs on the top left), and you click on a ramp on the summary charts panel - it should open the zoom view for commits represented by the ramp that you clicked on. This was on MacOS though.

jq1836: I've done further testing as you have described, and it seems that at least of Windows OS, the `openTabZoom` method might be redundant. Here are some screenshots to showcase the testing process.

Summary report settings
image

Prop data stored in CZoom CRamp before removal of `openTabZoom` method
image

Prop data stored in CZoom CRamp after removal of `openTabZoom` method
image

As seen from the prop data, the data seems to be identical before and after removing the `openTabZoom` method, indicating a possibly redundant `openTabZoom` method in `c-ramp.vue`. It seems that the `openTabZoomSubrange` method in `c-summary-charts.vue` that was provided as an event handler in the following code may be either overriding any data modified by the `openTabZoom` method in `c-ramp.vue`, or that the method is not even triggering. The following is an image of the code segment where the `openTabZoomSubrange` is passed as an event handler to the `summary-chart__ramp` element containing the `c-ramp` component that is being used to select a zoom range.

image

Perhaps we could test in a similar manner for MacOS and Linux to ensure that the `openTabZoom` method in `c-ramp.vue` is truly redundant.

Notes:
* Testing done on Windows OS, Windows 11
* The data debugging tool used is the Chrome Vue extension recommended in the DG
* Page was refreshed and the same zoom region reselected before screenshots were taken

vvidday: Sorry for the delay! Here's a reproduction of what I was referring to in my earlier message - this calls the `openTabZoom` method in `c-ramp.vue`:


<details>

ramp-zoom
</details>


<details>
<summary>Difference in `commits` object received by zoom tab</summary>

ramp-zoom2
</details>

jq1836: Thanks for the clarification, I now understand what you meant. I have modified the `openTabZoom` method such that the total insertions and deletions are calculated to allow for a safe creation of `Commit`.

frontend/src/components/c-ramp.vue
jq1836: Refer to [this comment](https://github.com/reposense/RepoSense/pull/2037#discussion_r1341945935) for solution to the unsafe casting

jq1836:
I believe that the failing Cypress tests are caused by the changes made in PR #2034. Will create a new PR to fix the tests.

frontend/src/components/c-ramp.vue
vvidday: Sorry for the delay. Did some investigation regarding this issue, will summarize for the benefit of others/future contributors as well:

- In the current code, `DailyCommit[]` is passed in place of `Commit[]` to the zoom view, which expects `Commit[]`. `Commit` is essentially a supertype of `DailyCommit` - it contains [two additional fields, `insertions` and `deletions`](https://github.com/reposense/RepoSense/blob/master/frontend/src/types/types.ts#L30).
- This means that anytime code in the zoom view (provided the zoom view is generated by this method call, which happens when a user clicks on a commit in the ramp with granularity set to day/week as outlined above & in the video), accesses these properties (`commit.insertions`/`commit.deletions`), something might break
- There are currently 2 bugs that I found with the zoom view that is generated from this action,
  - Sorting by lines of code in the zoom view doesn't work properly, since it sorts by `commit.insertions`, which would be undefined (https://github.com/reposense/RepoSense/blob/master/frontend/src/views/c-zoom.vue#L191)
  - The message body of commits aren't shown, as can be seen in the video in the earlier comment. This might be a potential regression caused by #1866, as the `isOpen` property that enables the toggling of message bodies is only added to the commits in `user.commits` instead of `user.dailyCommits`. (https://github.com/reposense/RepoSense/pull/1866/files#diff-5419037f2fc8c426a6a54e0614ee8033d10e16426d30573486c27e5811036869)

The fix you introduced fixes the first bug, but not the second. I'm actually not sure what the purpose of the original line `zoomUser.commits = user.dailyCommits;` is. It seems that by removing the line (and leaving `zoomUser.commits` alone), both of the above bugs don't occur. 

Given that these bugs are due to existing code and that this fix technically changes logic but doesn't fully fix the existing bugs, it's my preference that we don't introduce the new adding of insertions/deletions in this PR. IMO we can just do the original unsafe typecast (`as Commit[]`) and leave a comment warning that this typecast is unsafe, and file the current bugs as an issue to be fixed in a separate PR, since fixing those are out of the scope of migrating the file to TypeScript. I think this is just one of the instances where converting files to TypeScript help us uncover existing bugs.

The above approach is just my preference - please let me know if you have any thoughts on it (other reviewers please feel free to chime in!)

jq1836: Once again, thanks for investigating the above issue. Regarding the message body bug, I have come up with a modification to my fix that would address the message body bug, as well as the sorting bug, however, I do agree with the idea of creating a separate issue and PR to address this, for the purposes of maintaining a clear scope for each PR.

vvidday: Sounds good! Will approve this PR once the fix is reverted and feel free to request a review when you open the new PR. Thanks!

jq1836:
@vvidday Change has been reverted, PR is ready for review. Will put out the PR with the fix once this PR has been merged.

vvidday (approved):
LGTM!

vvidday:
> @vvidday Change has been reverted, PR is ready for review. Will put out the PR with the fix once this PR has been merged. Also, will an issue describing the bugs be necessary or will just the PR suffice?
Would be good to make an issue so it's clearer what's being fixed. Just a short description of the current bugs will do!

ckcherry23 (approved):
LGTM!

PR #2038, opened 2023-09-23
[#1936] Migrate c-resizer.vue to typescript

jq1836:
Part of #1936 

### Proposed commit message

```
unintended behavior.

Let's migrate the rest of the JavaScript code to TypeScript code to
facilitate future changes to the code.
```

### Notable changes made
* Function returned by `throttledEvent` function now only takes in a `MouseEvent` and ignores all other arguments, making it clearer what the `throttledEvent` function does


jq1836:
@reposense/active-reviewers-1 this PR is ready for review

frontend/src/components/c-resizer.vue
```diff
@@ -21,24 +21,25 @@
...
+<script lang='ts'>
 import { mapState } from 'vuex';
+import { defineComponent } from 'vue';
 
 const DRAG_BAR_WIDTH = 13.25;
 const SCROLL_BAR_WIDTH = 17;
 const GUIDE_BAR_WIDTH = 2;
 
-const throttledEvent = (delay, handler) => {
+const throttledEvent = (delay: number, handler: Function) => {
```
vvidday, line 32:
```suggestion
const throttledEvent = (delay: number, handler: (event: MouseEvent) => unknown) => {
```
Do you think it will be beneficial to be more strict on the type of the handler, by specifying that it takes in a `MouseEvent` as a parameter?

jq1836: Definitely. Was not aware that of such a possibility while working on the PR. Thanks for pointing this out!

vvidday: ESLint seems to be treating this as an error because the variable `event` in the type definition was not used. I think this case is a clear false positive so it should be fine to just declare an exception to this line (`eslint disable next line`)

vvidday (approved):
LGTM!

MarcusTXK (approved):
LGTM, great job with the migration

PR #2052, opened 2023-10-07
[#1936] Migrate repo-sorter.js to typescript

jq1836:
Part of #1936 

### Proposed commit message
```
unintended behavior.

Let's migrate repo-sorter.js to TypeScript code to facilitate future
changes to the code.
```

### Relevant changes made
* Implemented `checkKeyAndGetValue` method to retrieve the value stored in `element` with key, `key`. If a key is not provided, or if there is no such key in `element`, it returns undefined. In both these cases, the array is not sorted.


damithc:
@reposense/active-reviewers-1 this PR seems to be ready for review

frontend/src/utils/repo-sorter.ts
```diff
@@ -1,12 +1,27 @@
...
 }
 
-function sortingHelper(element, sortingOption) {
+function checkKeyAndGetValue(element: object, key: string | undefined) {
+  if (key === undefined || !(key in element)) {
+    return undefined;
+  }
+  /** Casting here is safe as guard clause above ensures that element has an attribute with the same name as the string
+   *  represented by key. */
+  return (element as any)[key];
```
ckcherry23, line 19: See if it is possible to use generics here. 
Else disable eslint for this line using `// eslint-disable-next-line  @typescript-eslint/no-explicit-any`.

ckcherry23 (approved):
LGTM. Just left a comment about the lint warnings.

frontend/src/utils/repo-sorter.ts
jq1836: Thanks for the review @ckcherry23, I have experimented with making use of generics over here and it seems like there is no real need for generics (although I have left a function that uses generics for general use). All that is needed would be casting the `key` string into a `keyof User`. However, this causes the problem of handling invalid keys and values that are not `ComparablePrimitive`, i.e. `string | number` to resurface (it was being suppressed due to the `any` typecast previously used). I have tried to handle these cases, but am unsure whether such behavior is appropriate (details below). Do let me know if the current behavior is not as intended.

### Details of behavior
When a `key` that is not valid or simply if a key is not provided to the function, `checkKeyAndGetValue` will return undefined, which will be caught in the `getComparablePrimitive` function's guard clause, resulting in an empty string being returned.
When a `key` that is associated with a value that is not a `ComparablePrimitive`, i.e. not a `string | number`, is provided to the function, it is caught in the `getComparablePrimitive` function's guard clause, also resulting in an empty string being returned.
When an empty string is returned, this results in the associated `User` being sorted to the front. If all `User`s in the `User[]`  result in an empty string, no sorting occurs.

jq1836:
Kindly requesting for your attention in reviewing PRs #2052 and #2053 related to the migration to TypeScript. Once done, issue #1936 can be closed.

PR #2060, opened 2023-10-21
[#944] Fix wrong credit information inherited by annotated author

SkyBlaise99:
Part of #944

### Proposed commit message

```
AnnotatorAnalyzer only overwrites the author but not the credit
information, when an author tag is found. If the analyze authorship
flag is enabled, credit information based on the blame author will be
wrongly inherited by the annotated author.

Lets assign partial credit if the annotated author is not the same as
the blame author, and keep the analyzed credit information if the 2
are the same.
```


SkyBlaise99:
The frontend tests are failing even though this PR did not touch the frontend at all. Seems to be because of #2066.
And codecov seems to refuse to work. #2067 proves that nothing is wrong with this PR.

SkyBlaise99:
codecov is working now, I tried the fix in #2066 but that didn't work.

SkyBlaise99:
Can someone review and merge this? Thanks!

chan-j-d (approved):
LGTM!

PR #2070, opened 2023-11-13
[#944] Improve visualization for full and partial credit

SkyBlaise99:
Part of #944

### Proposed commit message

```
Currently light green is used for full credit and dark green for partial
credit. This can be unintuitive as darker color often means a stronger
and more confirmed attribution.

Let's swap the color code around.
```

### Other information

`partial-authored-code-background` is also introduced into `github` to make the code more readable.


ckcherry23:
Hi @SkyBlaise99, I haven't taken a close look at all your new changes, but I'm just curious if the scope of your work includes showing partial/full credit code when all author contributions of a repo are merged? (For example, the following scenario)

Screenshot 2023-11-20 at 1 36 41 AM

ckcherry23 (approved):

SkyBlaise99:
@ckcherry23 currently it isn't, but what you've mentioned is definitely something I can explore and add on.

The credit info should remain after the merge, so I just need a way to update the color of the lines based on the credit info.

Thanks for the great idea. I'll probably make another PR for that.

SkyBlaise99:
Comment to remove stall label.

PR #2072, opened 2023-12-09
[#944] Change to originality score and new threshold value

SkyBlaise99:
Part of #944

### Proposed commit message

```
Currently, lowering the similarity threshold increases the algorithm’s
stringency. Typically, threshold functions in a way where meeting or
surpassing it results in a better score. However, in this case, full
credit is awarded when the similarity threshold is not reached.

Also, it is unclear why the similarity threshold was set to 0.8, which
means that the last author only need to change roughly 20% of the line
to be awarded full credit.

Let's use a new metric called the originality score, where exceeding the
originality threshold will then be awarded a full credit. Furthermore,
let's update the originality threshold to to 0.51, where the last author
needs to change at least half the line to be considered as making a
significant change and obtain full credit for it.
```

### Other information

To be merged after #2070


ckcherry23 (approved):
LGTM

PR #2074, opened 2023-12-23
[#1989] Reduce scope of try-catch block in ArgsParser::parse

### LINKED ISSUE #1989: Reduce the scope of the try-catch block in ArgsParser::parse
https://github.com/reposense/RepoSense/blob/53d5197d841fb1b801541b8f8414c04878d04913/src/main/java/reposense/parser/ArgsParser.java#L260-L261

Currently, the above `try` block covers the entire method. The scope of it should be reduced to cover only the method/s that throw these exceptions.

nseah21:
Fixes #1989

### Proposed commit message

```
Reduce scope of try-catch block tin ArgsParser::parse

The old scope of the try block in ArgsParser::parse 
is too permissive, making it less apparent which 
methods throw the associated exceptions.

Let's reduce the scope of the try block in 
ArgsParser::parse.
```

### Other information

NA

### DIFF
--- a/src/main/java/reposense/parser/ArgsParser.java
+++ b/src/main/java/reposense/parser/ArgsParser.java
@@ -258,76 +258,78 @@ public class ArgsParser {
      * @throws ParseException if the given string arguments fails to parse to a {@link CliArguments} object.
      */
     public static CliArguments parse(String[] args) throws HelpScreenException, ParseException {
+        ArgumentParser parser = getArgumentParser();
+        Namespace results;
+
         try {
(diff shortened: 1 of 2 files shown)


MarcusTXK (approved):
LGTM, nice job.

ckcherry23:
Looks good!

ckcherry23 (approved):

chan-j-d:
Hi @nseah21, thanks for doing up this PR! Though note that we generally restrict one first timer issue per contributor, so do look up issues without the tag for future contribution. Thanks!

Issue #1990, opened 2023-04-10
Move TimeUtil ParseException throwing to ArgsParser::parse method


chan-j-d:
Currently, the methods `verifyDatesRangeIsCorrect` and `verifySinceDateIsValid` in the `TimeUtil` class are throwing `ParseException` which is a side effect that might not make sense when the methods are reused. 
https://github.com/reposense/RepoSense/blob/53d5197d841fb1b801541b8f8414c04878d04913/src/main/java/reposense/util/TimeUtil.java#L176-L193

The exception thrown should be in the `ArgsParser::parse` method itself where they are used. Since these methods have been changed since when they were written to be a simple `compareTo` call, we can consider removing these methods and replacing the equivalent lines in `ArgsParser::parse` with the appropriate `compareTo` check and throw a `ParseException` if necessary.

PR #2075, opened 2023-12-23
[#1990] Move TimeUtil ParseException throwing to ArgsParser::parse method

nseah21:
Fixes [#1990](https://github.com/reposense/RepoSense/issues/1990)
### Proposed commit message

```
ArgsParser::parse uses the methods verifyDatesRangeIsCorrect
and verifySinceDateIsValid from TimeUtil to throw a 
ParseException if either of the checks fail. The throwing 
of an exception in a utility method is a side effect
that makes the methods less reusable.

Let's move the throwing of ParseException from TimeUtil 
into ArgsParser.
```

### Other information

NA


src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -420,8 +424,13 @@ private static void addAnalysisDatesToBuilder(CliArguments.Builder builder, Name
...
 
-        TimeUtil.verifySinceDateIsValid(sinceDate, currentDate);
-        TimeUtil.verifyDatesRangeIsCorrect(sinceDate, untilDate);
+        if (sinceDate.compareTo(untilDate) > 0) {
+            throw new ParseException(MESSAGE_SINCE_DATE_LATER_THAN_UNTIL_DATE);
+        }
+
+        if (sinceDate.compareTo(currentDate) > 0) {
+            throw new ParseException(MESSAGE_SINCE_DATE_LATER_THAN_TODAY_DATE);
+        }
```
MarcusTXK, line 433: Could you add some test coverage for the two new if blocks added?

MarcusTXK (changes requested):
Great job, just a small fix and we can merge it in.

src/main/java/reposense/parser/ArgsParser.java
nseah21: Hi @MarcusTXK, I've added the tests for ArgParser. The PR is ready for review.

MarcusTXK (approved):
LGTM, nice job with the refactoring.

ckcherry23 (approved):
LGTM!

Issue #2076, opened 2023-12-23
Refactor RepoConfiguration to simplify constructor complexity


MarcusTXK:
### What feature(s) would you like to see in RepoSense

I propose the implementation of the Builder pattern for complex configuration classes in RepoSense, such as [`RepoConfiguration`](https://github.com/reposense/RepoSense/blob/master/src/main/java/reposense/model/RepoConfiguration.java). This pattern would enhance code readability, maintainability, and ease future modifications.

### Is the feature request related to a problem?

The current implementation of some classes, like `RepoConfiguration`, involve constructors with a large number of parameters (the largest constructor has 18 parameters) and multiple overloaded versions. This complexity makes the code difficult to understand and maintain, increasing the likelihood of errors such as mixing up order of parameters and hindering contributions from new developers.

```
    public RepoConfiguration(RepoLocation location, String branch, List<FileType> formats, List<String> ignoreGlobList,
            long fileSizeLimit, boolean isStandaloneConfigIgnored, boolean isFileSizeLimitIgnored,
            List<CommitHash> ignoreCommitList, boolean isFormatsOverriding, boolean isIgnoreGlobListOverriding,
            boolean isIgnoreCommitListOverriding, boolean isFileSizeLimitOverriding, boolean isShallowCloningPerformed,
            boolean isFindingPreviousAuthorsPerformed, boolean isIgnoredFileAnalysisSkipped,
            List<String> ignoredAuthorsList, boolean isIgnoredAuthorsListOverriding, String extraOutputFolderName) {
```

### If possible, describe the solution

Adopting the Builder pattern for these classes will simplify object creation, enhance code clarity, and provide a flexible approach to object construction. This pattern is particularly beneficial for classes with numerous parameters and complex initialization logic.

### If applicable, describe alternatives you've considered

Overloaded constructors or factory methods are alternatives, however in my opinion, they don't offer the same level of organization and extensibility as the Builder pattern, especially for classes with a high number of optional parameters.

[...]

PR #2078, opened 2023-12-29
[#2076] Refactor RepoConfiguration to simplify constructor complexity

asdfghjkxd:
Part of #2076

### Proposed commit message

```
Refactor RepoConfiguration to simplify constructor complexity.

Currently, the constructors for the `RepoConfiguration` class is 
very complex, involving up to 18 parameters at once, and is 
not easily extensible.

With the proposed changes, the constructor now follows the 
Builder pattern, by using a nested class `Builder` that 
helps to compose config parameters in a way that is much more 
modular and comprehensible, enhancing code quality and 
reducing confusion among developers.

However, it must be noted that `Builder` objects are not reusable as of
now, and that more combinations of arguments are permitted due to the
removal of constructors for `RepoBuilder`, which previously helped to
ensure the correctness of parameter combinations.

Let's refactor `RepoConfiguration` to simplify the 
complexity of the constructor and increase modularity 
of the code.
```

### Other information

Apologies if this PR seems rather rough, as this is my first time contributing to an OSS, especially one at this scale!

I labelled this as a partial fix as documentation for the changes has not been made yet, and I'm not sure if that is within the scope of this issue; if required, I am happy to work on the updated documentation, before opening another pull request that seeks to address that, otherwise, I am also happy to upgrade this PR from a partial fix to a full fix.


gok99 (commented):
@asdfghjkxd kudos for taking on a relatively meaty first issue - changes are looking good and I see no significant issues so far! 

for a start, let's keep the documentation changes within this PR for completion, although i suspect there isn't too much to update.

asdfghjkxd:
@gok99 I have gone back and reviewed the user/developer guides and didn't find any part that needs modification.

I believe the only documentation left would be the JavaDocs within the code itself, but all public methods and some private methods should have their JavaDocs already furnished!

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -52,69 +52,415 @@ public class RepoConfiguration {
...
+        /**
+         * Updates the {@code fileTypeManager} for {@code RepoConfiguration}.
+         *
+         * @param fileTypeManager List of file types and groupings permitted.
+         * @return This builder object.
+         */
+        public RepoBuilder fileTypeManager(FileTypeManager fileTypeManager) {
+            this.fileTypeManager = fileTypeManager;
+            return this;
+        }
```
MarcusTXK, line 279: Consider accepting a `List<FileType>` directly in the builder and internally creating the `FileTypeManager`, for simplicity and maintainability. This way, we encapsulate `FileTypeManager` creation and make the API easier for users, keeping the builder's usage clean and hiding unnecessary complexity.
```
public RepoBuilder formats(List<FileType> formats) {
    this.fileTypeManager = new FileTypeManager(formats);
    return this;
}
```

MarcusTXK (changes requested):
@asdfghjkxd Great work on this! I was really impressed that you tackled this as your first issue. I think you did a very good job refactoring it into the new builder pattern and making sure to capture all the logic carefully.

I've added some comments as minor nitpicks. Additionally, it would be good if you could address [some of these new lines that Codecov flagged](https://github.com/reposense/RepoSense/pull/2078/checks?check_run_id=20923199843) as not covered by tests.

asdfghjkxd:
@MarcusTXK Sure thing, I will work on it and add a new commit here once I'm done fixing the code and creating new test cases!

asdfghjkxd:
I have also made some changes to the way the `processName` and `processOrganisation` works, whereby if specified by `RepoBuilder`, the folder name, output folder name and display name will change to the one that is provided by `RepoBuilder`, rather than defaulting to the default names given to those 3 variables.

I believe that this was an oversight on my part and I had failed to implement checks and allow `RepoBuilder` configurations to override the default configuration.

This behaviour should be fixed now in the new commit!

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -52,69 +52,427 @@ public class RepoConfiguration {
...
+                    ? org + "_" + this.outputFolderName
+                    : repoBuilder.outputFolderName;
+        }
+    }
+
+    /**
+     * Builds the necessary configurations for RepoConfiguration.
+     * Obeys the Builder pattern as described in {@link CliArguments}.
+     */
+    public static class RepoBuilder {
```
gok99, line 149: minor: since CliArguments calls it's builder `Builder`, we could also use RepoConfiguration.Builder for consistency

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -52,69 +52,427 @@ public class RepoConfiguration {
...
+        private transient boolean isLastModifiedDateIncluded;
+        private transient boolean isShallowCloningPerformed = false;
+        private transient boolean isFindingPreviousAuthorsPerformed = false;
+        private transient boolean isFormatsOverriding = false;
+        private transient boolean isIgnoreGlobListOverriding;
+        private transient boolean isIgnoreCommitListOverriding = false;
+        private transient boolean isIgnoredAuthorsListOverriding;
+        private transient long fileSizeLimit = DEFAULT_FILE_SIZE_LIMIT;
+        private transient boolean isFileSizeLimitOverriding = false;
+        private transient boolean isIgnoredFileAnalysisSkipped = false;
```
gok99, line 176: for discussion: since the builder is so tightly coupled with `RepoConfiguration`, would it make sense for the builder to hold a `RepoConfiguration` and build directly into it? This would save us the duplicated fields and boilerplate copies that seem error-prone. 

I think the Oracle [blog post](https://blogs.oracle.com/javamagazine/post/exploring-joshua-blochs-builder-design-pattern-in-java#:~:text=Bloch's%20Builder%20pattern%20can%20be,for%20a%20missing%20language%20feature.&text=The%20Builder%20pattern%2C%20which%20is,complex%20objects%20step%20by%20step.) also criticizes this duplication - in our case it seems very unlikely that we'd need the Builder for other objects.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -52,69 +52,427 @@ public class RepoConfiguration {
...
+            return this;
+        }
+
+        /**
+         * Builds the {@code RepoConfiguration} object with the necessary configurations.
+         *
+         * @return {@code RepoConfiguration}.
+         */
+        public RepoConfiguration build() {
+            return new RepoConfiguration(this);
```
gok99, line 475: should minimally add an assertion to check that location has been set. in general, for all previously disallowed constructor combinations, we want to continue to disallow them (unless we've checked that it doesn't result in undefined/unexpected behavior).

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -52,69 +52,427 @@ public class RepoConfiguration {
...
-                isFileSizeLimitIgnored, ignoreCommitList, isFormatsOverriding, isIgnoreGlobListOverriding,
-                isIgnoreCommitListOverriding, isFileSizeLimitOverriding, isShallowCloningPerformed,
-                isFindingPreviousAuthorsPerformed, isIgnoredFileAnalysisSkipped, ignoredAuthorsList,
-                isIgnoredAuthorsListOverriding, DEFAULT_EXTRA_OUTPUT_FOLDER_NAME);
+    /**
+     * Processes the relevant names of the repository configs.
+     */
+    private void processNames(RepoBuilder repoBuilder) {
+        String repoName = this.location.getRepoName();
+        this.displayName = repoBuilder.displayName == null
```
gok99, line 114: would an `Optional` be a good idea for these? we probably should be using more `Optional`s in RepoSense to enforce a check, but that's a separate discussion...

gok99 (changes requested):
sorry about the review delays! have left some suggestions and discussion points - feel free to comment :)

src/main/java/reposense/model/RepoConfiguration.java
asdfghjkxd: Sure thing, calling it `Builder` would make the code's naming conventions more consistent throughout the codebase! Will work on this ASAP!

src/main/java/reposense/model/RepoConfiguration.java
asdfghjkxd: Please correct me if I'm wrong, but I would like to clarify if you meant something like this:

```java
public static class Builder {
    ...
    private RepoConfiguration repoConfig;

    public Builder() {
        this.repoConfig = new RepoConfiguration();
    }

    ...

    public Builder branch(String branch) {
        // set*() methods should be private
        this.repoConfig.setBranch(branch);
        return this;
    }

    public RepoConfiguration build() {
        return this.repoConfig;
    }
}
```

If so, I believe that this might unintentionally result in aliasing issues. Since the builder pattern allows us to build on existing builder objects to create new configurations or reuse old configurations, the `build` method may unintentionally return the reference to the same `RepoConfiguration` object in memory, regardless of what configurations have been previously applied to the `Builder` object before.

This might violate the builder pattern since the `build` method is supposed to return a new `RepoConfiguration` if the builder configurations are modified, but now it only returns the reference to the latest modified `RepoConfiguration` object, and all previous `RepoConfiguration` references now point to the latest modified `RepoConfiguration` (all previously built `RepoConfiguration` are "forgotten").

src/main/java/reposense/model/RepoConfiguration.java
asdfghjkxd: I understand, I will look into creating some methods to check the validity of the constructed `RepoConfiguration` object, using the past constructors as a rough guideline on the correct constructor combinations to allow!

src/main/java/reposense/model/RepoConfiguration.java
asdfghjkxd: I agree, `Optional` would eliminate the `== null` checks, making the code a lot more understandable and readable! Will look into using`Optional` here!

src/main/java/reposense/model/RepoConfiguration.java
gok99: That's reasonable! I don't think we currently have any builder re-use (since this wouldn't have been possible with the constructors), but it seems possible that we might want to keep a builder for several `RepoConfiguration`s. It also seems like Java's `clone` is a bit of a slippery slope, so it might be unfortunately that this is the best we can do.

asdfghjkxd: @gok99 Just a tiny update, I have gone ahead and tried to implement your suggestion by converting the constructor for `RepoConfiguration` into a copy constructor, and created a private empty constructor for `RepoConfiguration` (for use within the `Builder` class).

So far, I managed to remove most of the duplicated variable names within the `Builder` class, spare for the `displayName`, `repoFolderName` and `outputFolderName`, as they are needed for the changing of the names during the actual build step for `Builder`, however, I am looking at implementing flags within `Builder` instead to track if the user has made any modifications to these names, and stop the default action (which is to process the default names for the repo and tag on extra information to it) from taking place.

src/main/java/reposense/model/RepoConfiguration.java
asdfghjkxd: @gok99 Can I also check if it is fine to set default values for some of the parameters? I realised from previous code that the constructors chain their calls and set default values for certain fields if they are not provided. I was wondering if it is ok to leave it as it is for now, and check that `RepoLocation` is minimally provided by the user?

The current checks that I had implemented relate to the parameters used in the constructors of the old code, which might not be all too helpful especially when default values for some fields are set (hence they will always be populated and not null).

gok99: Yes, my concern was that in setting booleans for example, we might be violating downstream assumptions about what combinations of fields were previously not possible (disallowed by the constructors), but are now possible.

Unfortunately since these assumptions are mostly implicit, we can't know for sure, without tracing through the whole pipeline, so I think we can do with just the location check for now. Minimally, existing Builder invocations should still work, and if fields interfere undesirably, we can fix them incrementally.

asdfghjkxd:
I'm reverting the changes back to before the refactor, it seems like something has caused all the system test cases to break

src/main/java/reposense/model/RepoConfiguration.java
gok99: > I am looking at implementing flags within Builder instead to track if the user has made any modifications to these names, and stop the default action (which is to process the default names for the repo and tag on extra information to it) from taking place.

You could probably do the construction of the names within the builder. Null checks are okay, my previous comment on optionals was more to set `displayName` etc to be optionals rather than using them just for the control flow. As long we make sure that the field is eventually set before the object is built, nulls seem fine to use.

> If so, I believe that this might unintentionally result in aliasing issues.

Also regarding aliasing, I realized `RepoConfiguration` instances would still share non-primitive fields which could still cause undesirable aliasing issues. We could possibly reset the RepoConfiguration instance when `build` is called, to avoid this.

src/main/java/reposense/model/RepoConfiguration.java
asdfghjkxd: I see, I have implemented a simple check to see if `RepoLocation` has been defined during the build process of the `RepoConfiguration` object. Hope that this temporarily provides some safeguards against illegal combinations of parameters.

src/main/java/reposense/model/RepoConfiguration.java
asdfghjkxd: Regarding the first point, I have mostly converted the null checks into Optional checks, hopefully making the code much more readable and understandable by abstracting away the logic. I have also moved the construction of the name over into the builder class.

Regarding the second point, I have modified the `build` method accordingly and utilised the copy constructor along with the resetting of the built `RepoConfiguration` object within the builder after it is done building. I realise that this might impact the reusability of the builder objects, but as we are not reusing builder objects anywhere as mentioned previously, this should be fine for now.

src/main/java/reposense/model/RepoConfiguration.java
MarcusTXK: Should we look into creating an issue for these separately?

asdfghjkxd: @MarcusTXK I agree, most of the issues we had discussed above are rather complicated (parameter verification, duplication of `RepoConfiguration`, etc.) and seem to warrant the creation of separate issues.

src/main/java/reposense/model/RepoConfiguration.java
gok99: Should we get rid of the copy constructor if we're resetting the object? We could create a separate issue for implementing `Cloneable`s but even if we go that route, we wouldn't need the constructor.

asdfghjkxd: Sure, we could stick with the previous builder constructor to enable deep cloning later in the future!

src/main/java/reposense/model/RepoConfiguration.java
MarcusTXK: Yeah I think we can focus on improving the `RepoConfiguration` incrementally, as @gok99 has brought up alot of good points.

src/main/java/reposense/model/RepoConfiguration.java
```diff
@@ -25,96 +27,422 @@ public class RepoConfiguration {
...
+        /**
+         * Builds the {@code RepoConfiguration} object with the necessary configurations.
+         *
+         * @return {@code RepoConfiguration}.
+         * @throws ConfigurationBuildException if there was an issue building the {@code RepoConfiguration}
+         *     object.
+         */
+        public RepoConfiguration build() {
+            if (!validate()) {
+                throw new ConfigurationBuildException();
```
gok99, line 421: Perhaps you can add a simple test to cover this case, but looks good to me otherwise!

src/test/java/reposense/parser/ArgsParserTest.java
```diff
@@ -524,6 +524,8 @@ public void parse_configOrLocationsSimilar_success() throws Exception {
                 .getRunConfiguration(locationCliArguments)
                 .getRepoConfigurations();
 
+        System.out.println(actualRepoConfigs.get(0));
```
gok99, line 527: Did you intend to leave this in?

src/main/java/reposense/model/RepoConfiguration.java
asdfghjkxd: Ah okay, I will work on it ASAP!

src/test/java/reposense/parser/ArgsParserTest.java
asdfghjkxd: Oh no, I must have forgotten to remove it after debugging. My bad! I will remove it in another commit!

gok99 (approved):
LGTM, thanks for the work!

MarcusTXK (approved):

MarcusTXK (approved):
LGTM, great job with this PR!

logical-1985516:
Hi, should issue #2076 be linked to this PR, therefore closing the issue?

PR #2084, opened 2024-01-20
[#1980] Standardise Array Style for Frontend Files

### LINKED ISSUE #1980: Inconsistent use of `T[]` and `Array<T>` for arrays
Standardize the style used for arrays to be `T[]` or `Array<T>`, based on the complexity of T.

During the TypeScript migration of the front-end, both of these styles were used interchangeably. Since the 2 styles are functionally equivalent, we can use the same style consistently across the codebase to make it easier for developers to read and understand array types.  

* Add a [lint rule to enforce that the same array style](https://typescript-eslint.io/rules/array-type/#array-simple) is used across the code base. This should be in the global `.eslintrc.json`.

  We can consider using the `array-simple` option as this allows us to use the shorter, modern `T[]` or `readonly T[]` for simple types (i.e. types that are just primitive names or type references), and `Array<T>` or `ReadonlyArray<T>` for all other types (union types, intersection types, object types, function types, etc) where the shorter syntax could get confusing.
```
{
  "rules": {
    "@typescript-eslint/array-type": ["error", {
        "default": "array-simple",
        "readonly": "array-simple"
      }
    ]
  }
}
```
* Make edits to the codebase such that the newly introduced lint rule is followed. Most of the changes will be in the `props` and `data()` sections of the `.vue` files. 

We could also consider using the `array` or `generic` options for the lint rule if we want to stick to a single style at all times, no matter the complexity of the type.

### Additional context
Discussed in #1976

sopa301:
Fixes #1980 

### Proposed commit message

```
Standardise Array Usage

Array<T> and T[] can be used interchangeably in TypeScript. We can use
the same style consistently across the codebase to make it easier for
developers to read and understand array types.

In Vue, using T[] in defineComponent requires casting in the form of
"Array as PropType<T[]>", while Array<T> can be used as is. Therefore,
it would be more consistent to use Array<T> for .vue files.
Additionally, such a rule can be enforced more easily with eslint.

Let's include a rule for eslint to check array usage and change the
code to follow this rule.
```


jonasongg (commented):
Just a small nit, but might be better to avoid cluttering the commit with changes to other lines in `.eslintrc.json`!

sopa301:
I've removed the unrelated lines. It was added by mistake by the IDE :P

jonasongg (approved):
LGTM!

ckcherry23:
@sopa301
> Interestingly, while TypeScript makes no distinction between the two styles, Vue (particularly in defineComponent) only accepts `Array<T>` directly, while requiring something like `[] as T[]` for the other style. I would like some input from someone with more experience in Vue on this.
> 
> For now, I have implemented the rule only for the TS files as the Vue side doesn't seem so clear-cut.

Thanks for your investigation regarding this! For defining arrays with the `T[]` syntax within Vue's `defineComponent`, you can try using `Array as PropType<T[]>`.

However, it does seem that `Array<T>` would be shorter and better for such use cases. If possible, switch to the `Array<T>` syntax for all array uses within `defineComponent`. Otherwise, we can standardise to `Array as PropType<T[]>`. Thanks!

sopa301:
@ckcherry23 Thank you for the review! I'm thinking using Array<T> for all Vue files is more consistent and enforceable, but I can change only the defineComponent code if needed.

ckcherry23 (approved):
@sopa301 This looks good to me! Thanks for your work!

ckcherry23:
@sopa301 Can you please resolve the merge conflicts?

vvidday (approved):
LGTM, nice job

Issue #1973, opened 2023-04-03
Duplication between the `User` Interface and class


HCY123902:
### What feature(s) would you like to see in RepoSense

As is mentioned in #1965, there are currently a TypeScript interface of `User` in `types/types.ts` and a JavaScript class of `User` in `utils/User.ts`. ~~The difference between them is that the `commits` attribute in the interface is optional while it is not in the class.~~ However, as the interface and the class are mostly equivalent, it will be reasonable to remove either of them.

### Is the feature request related to a problem?

The interchangeable usage of the `User` Interface and class implies possible inconsistency.

### If possible, describe the solution

Remove either the interface or the class. Removing the class seems to give more advantages considering the runtime.

### If applicable, describe alternatives you've considered

Removing the interface may also be reasonable as it may allow more flexibility with class methods in a future context.

PR #2085, opened 2024-01-20
[#2004] Remove redundant Segment class

### LINKED ISSUE #2004: Duplicate interface and class for `Segment`
There is currently an `AuthorshipFileSegment` interface in `types/types.ts` and a `Segment` class in `utils/Segment.ts`. However, as the interface and the class are mostly equivalent, it will be reasonable to remove either of them.

The interchangeable usage of the `Segment` Interface and class implies possible inconsistency.

Remove either the interface or the class and replace all usages of the removed construct. Removing the class seems to have more advantages considering the runtime.

Removing the interface may also be reasonable as it may allow more flexibility with class methods in a future context.

### Additional context
Discussed in #1969 and [#1965](https://github.com/reposense/RepoSense/pull/1965#issuecomment-1488698009).

jonasongg:
Fixes #2004

### Proposed commit message

```
Remove Segment class, replace with AuthorshipFileSegment

The Segment class and AuthorshipFileSegment interface are equivalent in
their usage and their redundancy seems to be a remnant when the code
was migrated from JS to TS.

Let's remove the Segment class for consistency and to improve runtime
performance.
```

### Other information

A very similar issue #1973 exists, but I decided against putting addressing both issues in the same PR. If it is better to address both (since they are both quite minor changes), let me know!


sopa301 (approved):
LGTM. We should omit the full stop in the first line of the commit message if it is meant as the subject line.

I agree that #1973 should be done in a separate PR as these are related but parallel issues.

jonasongg:
@sopa301 Ah, thanks for spotting!

vvidday (approved):
Great job! Definitely agree with the call of addressing both issues in separate PRs (in line with our [PR guidelines](https://se-education.org/guides/guidelines/PRs.html#scoping-a-pr))

Issue #2058, opened 2023-10-11
URL Changes upon refresh


SkyBlaise99:
### Tell us about your environment

* **RepoSense Version:** Current latest [5cd529f](https://github.com/reposense/RepoSense/commit/5cd529fb46809a7b8e1c9c27dc04f428b83e89f5)
* **OS and Version:** Windows 11
* **Web Browser and Version (if applicable):** Chrome Version 118.0.5993.71 (Official Build) (64-bit)

### Please include the steps to reproduce the bug.

1. Generate a report and view it `java -jar RepoSense.jar -repo https://github.com/reposense/RepoSense.git -v`
1. Click to view author's (Eugene's) contribution
1. Input `README.md` into `Filter by glob`
1. The current URL is http://localhost:9000/?search=&sort=groupTitle%20dsc&sortWithin=title&since=2018-05-03&timeframe=commit&mergegroup=&groupSelect=groupByRepos&breakdown=false&tabOpen=true&tabType=authorship&zFR=false&tabAuthor=eugenepeh&tabRepo=reposense%2FRepoSense%5Bmaster%5D&authorshipIsMergeGroup=false&authorshipFilesGlob=README.md
1. Refresh the webpage, the URL changes to http://localhost:9000/?search=&sort=groupTitle%20dsc&sortWithin=title&since=2018-05-03&timeframe=commit&mergegroup=&groupSelect=groupByRepos&breakdown=false&tabOpen=true&tabType=authorship&zFR=false&tabAuthor=eugenepeh&tabRepo=reposense%2FRepoSense%5Bmaster%5D&authorshipIsMergeGroup=false&authorshipFilesGlob=README.md%23%2F (extra `%23%2F`)
1. The search bar is now `README.md#/`

### What was expected to happen?
The search bar should remain at `README.md`

### What actually happened? Please include a screenshot of the output.

Before:
![image](https://github.com/reposense/RepoSense/assets/69448615/71578a82-99ee-4194-af67-ae61d939619a)

After referesh:
![image](https://github.com/reposense/RepoSense/assets/69448615/c5e1d5a8-4073-46fd-b98c-93f2fb20ca50)

SkyBlaise99:
[...]

Issue #1987, opened 2023-04-08
Glob filter automatically reset after the second refresh


HCY123902:
### Tell us about your environment

* **RepoSense Version:** master 10
* **OS and Version:** Windows 10
* **Web Browser and Version (if applicable):** Edge

### Please include the steps to reproduce the bug.

1. Open a [report](https://dashboard-1911-pr-reposense-reposense.surge.sh/?search=&sort=groupTitle&sortWithin=title&since=2018-03-01&timeframe=commit&mergegroup=reposense%2Ftestrepo-Alpha%5Bmaster%5D&groupSelect=groupByRepos&breakdown=false&tabOpen=true&tabType=authorship&zFR=false&tabAuthor=luozhijie-tom&tabRepo=p-repo-sense-test-repo%2Fcode%5Bmaster%5D&authorshipIsMergeGroup=false&authorshipFileTypes=txt~java~md~fxml~gradle~xml~properties~gradlew~bat~MF~rej~TXT&authorshipIsBinaryFileTypeChecked=false&authorshipIsIgnoredFilesChecked=false)
1. Open any authorship contribution panel, such as this
![image](https://user-images.githubusercontent.com/54735123/230727701-8b4f1fc3-aca5-4af8-b83c-6d3ea051b6a6.png)

1. Enter a regex path in the glob filter to filter the file list, such as this
![image](https://user-images.githubusercontent.com/54735123/230727722-0f8d85ff-f9a4-40d6-9236-a0e87a3d68c2.png)

1. Refresh the page 2 times

### What was expected to happen?

1. The glob filter and list of remaining files should stay the same regardless of the number of times of refresh.

### What actually happened? Please include a screenshot of the output.

1. The glob filter input and list of remaining files stay the same after the first refresh
![image](https://user-images.githubusercontent.com/54735123/230727738-6d9058ba-a348-4506-b785-717e4d713d05.png)

1. However, after the second refresh, the glob filter and file list are reset to the initial state
![image](https://user-images.githubusercontent.com/54735123/230727826-0caef391-0a59-47f3-a3f9-0321e0d12ba8.png)

### If possible, include the URL to your RepoSense report or log files (if any).

[...]

PR #2086, opened 2024-01-20
[#2016] Remove hash symbol from URL when decoding hash

### LINKED ISSUE #2016: Breakdown by file type not retained after refresh

jonasongg:
Fixes #2016 

### Proposed commit message

```
Fix bug where some params were not retained after refresh

The issue pertains to breakdown of file types disappearing after a
refresh, but it seems to apply to all params that were last in the URL
due to symbols not being filtered out correctly.

Let's fix this issue to make sure the state of the report stays the
same when refreshing.
```

### Other information

The bug seems to have been caused by the URL containing `#/` at the end, which would cause the last param to not be parsed correctly. I couldn't definitively find what caused the anchor to appear in the URL, so I filtered it out of the URL when decoding `hashParams`. Let me know if this behaviour should be changed!

Regarding the Cypress tests, #2017 seems to be outdated/some of the tests don't seem to be checking for the correct thing, at least on my machine. If I should investigate this further in anticipation of its being merged, let me know as well!


sopa301:
Googling a bit suggests that this appending of `#/` is caused by Vue's [Hash Mode](https://router.vuejs.org/guide/essentials/history-mode), which appends a `#` followed by the route `/` (the default one) at the end of the URL. This could be a regression from the implementation of Vue Router.

I tested this using the 2324S1 CS2103T RepoSense dashboard. Some evidence for this hypothesis:

- Using `#/widget` gives a different view of the webpage.
- Using `#/test` gives a blank webpage

That said, the current fix is good enough to make it work. I propose 2 possible paths in the future:

1. Rewrite the code to ensure Vue Router's internal routing is finished and the URL is fixed before it is processed by `api.ts` (not recommended due to complexity)
2. Refactor `api.ts` to capture this consideration more distinctly.

jonasongg:
@sopa301 Thanks for the suggestions! 

Just wanted to link these related issues here since it seems they are caused by this same problem: #2058 #1987

sopa301 (approved):
LGTM. As commented above, this is sufficient to fix the bug, but may warrant further refactoring.

ckcherry23 (commented):
Thanks for your inputs @jonasongg and @sopa301. I think this temporary fix is a reasonable solution for now as it resolves the current issue without introducing new issues. We can switch to the Vue Router HTML5 mode in the future. 

@jonasongg Since this fix is somewhat unconventional, can you add a comment in the code explaining the rationale behind the change? This will help future developers understand the context and reasoning behind this decision.

Regarding the Cypress tests, we should ideally be adding a test case for the bug fixed in this PR itself. Let us wait to see the progress of PR #2017 over the next few days and make a call over the changes needed.

MarcusTXK (commented):
Nice fix. Agree with @ckcherry23 that a comment to explain the logic for future developers would be useful.

Regarding the tests, I think perhaps @jonasongg could reference the code written in #2017 and add the updated tests, then co-author credit could be added to @nseah21.

jonasongg:
@ckcherry23 @MarcusTXK Thanks for the reviews! Just to confirm, I should add the `codeView_renderFilterHash.cy.js` file from #2017 here and add tests that test this new expected behaviour?

ckcherry23:
@jonasongg #2017 was just merged. You can look into what additions/changes are required on the tests now, thanks!

MarcusTXK:
@jonasongg After discussion, we decided to merge in #2017, so you can work on the merged test directly on your branch.

jonasongg:
Just added the comment and some test cases! For checked file types, I added some tests to make sure it was testing the last file type to appear in the URL (yml) as that was what was being affected by the bug. Also added some non exhaustive test cases for the code panel options including one for authorshipFileGlob that is most likely the cause for #2058 and #1987. Let me know if I should add more/change anything!

frontend/cypress/tests/codeView/codeView_renderFilterHash.cy.js
```diff
@@ -309,31 +309,184 @@ describe('render filter hash', () => {
     cy.url()
       .should('not.contain', 'gradle');
 
+    cy.url()
+      .should('not.contain', 'yml');
+
```
ckcherry23, line 314: Can we make it clear why gradle and yml were the 2 file types selected here with a comment?
For example,
```
// Assumption: gradle is the first file type and yml is the last file type

File: frontend/cypress/tests/codeView/codeView_renderFilterHash.cy.js [hunk 2]
```diff
@@ -309,31 +309,184 @@ describe('render filter hash', () => {
...
+    cy.reload();
+
+    cy.url()
+      .should('not.contain', '%23%2F');
+
+    cy.url()
+      .should('contain', 'reverseAuthorshipOrder=true');
+  });
+
+  it('code panel: filter by glob: url params should persist after change and reload', () => {
```
ckcherry23, line 461: This function only tests after one reload, but there is a bug in the filter by glob string after 2 reloads. 
Hence, just wanted to add a note that we must add a test with 2 page reloads when the bug in #1987 gets fixed.

ckcherry23 (approved):
LGTM! Just have a few minor comments

jonasongg:
Should be clarified now!

MarcusTXK (approved):
LGTM, I think you did a great job

PR #2087, opened 2024-01-21
[#1878] Updating `SystemTestUtil::assertJson` to compare Json objects instead of line-by-line analysis

### LINKED ISSUE #1878: Updating `SystemTestUtil::assertJson` to compare Json objects instead of line-by-line analysis
Currently, `SystemTestUtil::assertJson` uses the default file content testing method which analyzes line-by-line. It might be helpful to convert the files into `Json` objects that can then be compared properly. This might be helpful to avoid future problems where changes in the code causes an equivalent Json file to be arranged slightly differently and causes the test case to fail. 

The current code:
https://github.com/reposense/RepoSense/blob/c162ff20d9c055bddb66ad2f57b3a62906b541d2/src/systemtest/java/reposense/util/SystemTestUtil.java#L64-L67

We can convert the file paths into JsonObjects similar to what is done in `assertSummaryJson` and take advantage of the `JsonObject`'s `equals` method

asdfghjkxd:
Fixes #1878

### Proposed commit message

```
`SystemTestUtil::assertJson` currently checks if two JSON files are
equal by comparing the file contents line by line rather than by
converting the files directly into JSON objects, which can then be
directly compared using the built-in `equals` methods for
`JsonElement`.

This process also does not automatically check if the file is indeed a
JSON file, and implicitly assumes that the input paths are paths to
JSON files. This may cause expected behaviours if a user accidentally
or intentionally includes a path to a non-JSON file within the method
calls.

Let's move to refactor the code, and utilise the built-in methods for
`JsonElement` objects to check for the equality of JSON objects.
```

### Other information

The original issue alludes to the use of `JsonObject` in the fixes, however, while trying to get the system test cases to run, it appears that this method is to be used on `JsonObject` and `JsonArray` as well. `JsonElement` was hence chosen as a suitable class that encapsulates the information of the parsed JSON objects, while allowing for equality of the JSON objects to be checked.

### DIFF
--- a/src/systemtest/java/reposense/util/SystemTestUtil.java
+++ b/src/systemtest/java/reposense/util/SystemTestUtil.java
@@ -7,10 +7,9 @@ import java.nio.file.Path;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
-import com.google.gson.JsonArray;
-import com.google.gson.JsonElement;
 import org.junit.jupiter.api.Assertions;
 
+import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.JsonParser;
 
--- a/src/systemtest/java/reposense/util/SystemTestUtil.java
+++ b/src/systemtest/java/reposense/util/SystemTestUtil.java
@@ -7,6 +7,8 @@ import java.nio.file.Path;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
 import org.junit.jupiter.api.Assertions;
 
 import com.google.gson.JsonObject;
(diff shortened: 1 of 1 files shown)


jonasongg (commented):
Not sure if this is strictly necessary, but you might want to put both FileReader instantiations in the try () block so that they will automatically close. Otherwise, LGTM!

asdfghjkxd:
That's a good suggestion! I will look into moving FileReader into the try-with-resources clause!

asdfghjkxd:
@jonasongg Just an update, but after having reviewed the code, it seems that both instances of `FileReader` are instantiated within the try-with-resources clause. Both `FileReader`s should close after the assertion is executed!

sopa301 (approved):
LGTM!

gok99 (approved):
LGTM!

MarcusTXK (approved):
LGTM, nice job

PR #2089, opened 2024-01-21
[#944] Differentiate full and partial credit when group is merged

SkyBlaise99:
Part of #944

### Proposed commit message

```
Currently, when viewing individual contributions, full and partial
credit are differentiated with dark and light green colors. However,
this distinction is not applied when the group is merged, potentially
causing confusion for users.

Let's introduce a clear differentiation between full and partial credit
when viewing a merged group.
```

### Other information

Follow-up of the discussion [here](https://github.com/reposense/RepoSense/pull/2070#issuecomment-1817927236)

Sample Output V3 (transparency level 50, 20)

image

Sample Output V2 (transparency level 60, 30)

image

Sample Output V1 (transparency level 75, 30)

image


frontend/src/components/c-segment.vue
```diff
@@ -53,7 +53,7 @@ export default defineComponent({
     return {
       isOpen: (this.segment.knownAuthor !== null) || this.segment.lines.length < 5 as boolean,
       canOpen: (this.segment.knownAuthor === null) && this.segment.lines.length > 4 as boolean,
-      transparencyValue: '30' as string,
+      transparencyValue: (this.segment.isFullCredit ? '75' : '30') as string,
```
ckcherry23, line 56: Thanks for this! However, I think a transparency value of '75' may make the highlight color too dark and have little contrast with the text when there are too many full credit lines. Is it possible for you to adjust the transparency a bit based on a more common scenario of full credit and partial credit lines instead?

ckcherry23 (approved):

frontend/src/components/c-segment.vue
ckcherry23: This could be a useful tool to check color contrast: https://coolors.co/contrast-checker/4e5966-a1ccff

SkyBlaise99: Thanks for the feedback. It's just that because the colors are randomly generated so I can't be really sure that all the colors are with high contrast.

75 was a number I tried personally and I think it fits well as seen in the picture.

I'll look into the link you provided and see if I can do something with it.

SkyBlaise99: Actually I just tried using the tool, 

Blue: #A1CCFE - 7.57
Pink: #F9C5C5 - 8.32
Yellow: #FFECA2 - 10.69
Purple: #E1B2EC - 7.09
Green: #AEFFC7 - 10.79

7 is the minimum for very good, so I think except for the purple which is kinda just on the line, the rest are all good. I'll see if lowering will make it better.

SkyBlaise99: I have now lowered it to 60 and I think this is a good number.

Blue: #B2D5FF- 8.34
Pink: #FBCFCF- 8.98
Yellow: #FFEFB3- 10.98
Purple: #E6C0EF- 7.91
Green: #BDFFD0- 11.06

ckcherry23: Thanks for this! I believe the contrast ratios you've mentioned are probably for a darker text color. But considering we use syntax highlighting with various lighter colors (line 36 below), opacity 50 (and 20 for partial) would be safer levels (pictured below). 

Screenshot 2024-01-24 at 12 52 05 AM

However, if you've tried the highlight colors for these lighter text colors and they seem to work, then ignore my comments.

SkyBlaise99: I see, I'm aware of the syntax colouring. Can I ask if there is a list of all the syntax colours? It would be nice to find a value that has a good contrast with all the syntax.

Do you think having a function to calculate the value would be better? For example, minimum set to say 60, then comparing it against the other syntax colours, try to increase this value higher but not exceeding say 80.

ckcherry23: This file might be useful: `frontend/src/styles/hightlight-js-style.css`.

ckcherry23: Regarding having a function to calculate the value, I do not think it is necessary if we go with a safe enough low opacity value. 

Another option we have is to make sure the colors used for highlighting are lighter colors and do not clash with the syntax highlighting colors.

SkyBlaise99: Thanks for the info. I agree with the 50-20 you proposed. I did some extra checks on the colours.
1. As transparency level decreases, the contrast generally gets better for any colour.
1. I was initially afraid that 20 for partial will make it too transparent and hard to observe, but it turns out that it isn't that bad. I confirmed that using the deltaE way and the lowest was 3.61 which is still okay. (2 - 10 is considered perceptible at a glance by this [link](https://zschuessler.github.io/DeltaE/learn/#toc-defining-delta-e))

PR #2092, opened 2024-01-22
[#2091] Improve memory usage by refactoring Regex compilation

### LINKED ISSUE #2091: Suggestions on improvement for memory performance regarding Regex matching

Currently, some areas of RepoSense use Regex checks within iterations.

For example, consider `StringsUtil::filterText`.

```java
public static String filterText(String text, String regex) {
    String[] split = text.split("\n");
    StringBuilder sb = new StringBuilder();
    for (String line: split) {
        if (line.matches(regex)) {
            sb.append(line + "\n");
        }
    }

    return sb.toString();
}
```

As observed, the method `String::matches` is called for as many times as there are lines in the input text, which is not performant as `String::matches` creates and compiles a new Regex pattern every time it is called. This behaviour is evident in the [source code](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/lang/String.java) for the `String::matches` method, which makes a call to `Pattern.matches(regex, string)`.

It is recommended by the documentation for the [`Pattern` class](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html) as well as some [external sources](https://raygun.com/blog/java-performance-optimization-tips/#:~:text=Use%20compiled%20regular%20expressions%3A%20One,Pattern%20class%20for%20this%20purpose.) to avoid such a pattern of code since it causes many unnecessary Regex compilations, which takes more time and consumes more memory.

After implementing some fixes for this behaviour, and from some preliminary profiling using Intellij's built-in Java Profiler, I have observed that the improvements lead to a small decrease in runtime and a significant decrease in memory usage when Regex patterns are precompiled first before being used in a `for` loop.

Here are some of my findings:

**Command used to test**

```
./gradlew run -Dargs="--repos https://github.com/reposense/RepoSense.git https://github.com/CATcher-org/CATcher.git https://github.com/MarkBind/markbind.git --output ./report_folder --since 31/1/2017 --formats java adoc xml --view --ignore-standalone-config --last-modified-date --timezone UTC+08 --find-previous-authors"
```

**Runtime**

| Before Improvement | After Improvement |
| --- | --- |
| Screenshot 2024-01-22 at 10 00 30 PM | Screenshot 2024-01-22 at 10 00 21 PM |

Some key points to note:

* The runtime being inspected is the CPU time (time that the code is executed, ignoring gradle builds)
* **11.69%** decrease in runtime

**Memory Consumption**

| Before Improvement | After Improvement |
| --- | --- |
| Screenshot 2024-01-22 at 10 07 12 PM | Screenshot 2024-01-22 at 10 07 01 PM |

Some key points to note:

* **43.32%** (!) decrease in memory usage
* It also appears that `StreamGobbler` consumes a significant chunk of memory, perhaps we could also look into reducing the memory usage of that class?

---

**Limitations**

* These preliminary profiling tasks are done on a M1 Mac with MacOS. These performance values may vary across different OS and processors
* Only 3 profiling samples were taken, more samples might have to be taken, across multiple platforms and processors, to decide if these enhancements have a positive impact on the performance of RepoSense

This feature is not exactly related to a particular problem in general, but it is linked to the overall goal of increasing RepoSense's performance and memory usage.

The solution involves the pre-compilation of the offending Regex pattern into a `Pattern` object, and using a `Matcher` object created from `Pattern` to test if the string matches the Regex pattern.

The fixes are found in a draft PR [here](https://github.com/reposense/RepoSense/pull/2092).

No other viable alternatives have been explored or tried at this time. It seems on preliminary inspection that these Regex operations are crucial and cannot be removed or replaced with other built-in String methods.

### Additional context

N/A

asdfghjkxd:
Fixes #2091

### Proposed commit message

```
Currently, Regex checking is used in conjunction with iteration. This
pattern of coding is frowned upon due to the excessive Regex pattern
compilation, causing the program to run slower and consume more
memory.

By moving the Regex pattern compilation outside of the iteration, and
by using `Matcher` objects to check if the strings match the Regex
performance, we can potentially remove this performance bottleneck.

Let's move to refactor the code and remove such instances of Regex use
in iterative loops.
```

### Other information

N/A


sopa301 (approved):
LGTM

gok99 (approved):
Thanks for the thorough investigation and fix! 
LGTM

MarcusTXK (approved):
LGTM, I think you did a great job catching this issue. Nice fix!

PR #2093, opened 2024-01-22
[#1973] Remove redundant User class

### LINKED ISSUE #1973: Duplication between the `User` Interface and class

jonasongg:
Fixes #1973

### Proposed commit message

```
Remove utils/user.ts, replace with User interface from types.ts

The User class and interface are equivalent in their usage and their
redundancy seems to be a remnant when the code was migrated from JS to
TS.

Let's remove the User class for consistency and to improve runtime
performance.
```

### Other information

Similar to #2085.


sopa301 (approved):
LGTM

ckcherry23 (approved):
LGTM!

PR #2102, opened 2024-01-29
[#467] Add Title Component

### LINKED ISSUE #467: Give a way to add an intro to the generated report

sopa301:
Fixes #467

### Proposed commit message

```
Implement Title Component

With a new focus on allowing users to use RepoSense as a portfolio
tool, more functionality supporting this focus is needed.

Let's allow users to add customizable content in Markdown format at the
top of the report for a personalized introduction.
```

### Other information

The input file 'title.md' is inserted with the --assets flag when the jar file is used.


asdfghjkxd (approved):
LGTM!

ckcherry23:
Hi @sopa301, thanks for working on this! I have not tested your code yet, but after skimming through the changes, I have some thoughts. 

* I think this feature would be more extensible if we allow the user to specify the markdown content needed as a config file in the RepoSense build command rather than a file within the frontend directory itself. 
* I am not fully sure if we are still looking for a simple text intro or a navbar like the CS2103T website. Would help if prof @damithc can clarify.
* If a navbar is more preferable, one approach that we could try out is to require YAML frontmatter in the markdown file provided by the user. The frontmatter obtained (for example, a list of links and the header color) using a tool like [`gray-matter`](https://github.com/jonschlinkert/gray-matter) could be inserted into a template header component as props and displayed accordingly on the report.

damithc:
> * I am not fully sure if we are still looking for a simple text intro or a navbar like the CS2103T website. Would help if prof @damithc can clarify.

@sopa301 Can you provide screenshots of how it looks and how one is supposed to configure the content?

sopa301:
Currently, the content is obtained by looking for a `title.md` file in the `markdown` folder. I can rewrite it in the way that is suggested to give more flexibility.

Regarding prof @damithc 's query of how it looks:
Input file (an excerpt of the RepoSense README):
image

Html report:
image

damithc:
> The rest of the report can be accessed by scrolling down and the title can be removed from view by scrolling down.

Sounds promising @sopa301 Is there any live preview for me to see how it work?
I assume if the title.md is not present, the report looks same as before adding this feature?

sopa301:
Unfortunately I don't have a way to provide a live preview now. Also there's a bug in the tests, so the github preview doesn't work. I can show it to you in person next monday if I don't solve the bug by then. If someone knows of another way to host a preview do let me know!

If the title.md is empty, the report will look the same as before. I forgot to handle the case where the file is missing, but I think I'll rewrite the code to account for this scenario.

sopa301:
I managed to host the html file [here](https://sopa301.github.io/alpha/#/), although I messed up somewhere and the data didn't get uploaded as well. I'll update the link if I got it fixed.

frontend/src/components/c-title.vue
```diff
@@ -0,0 +1,31 @@
...
+        if (!response.ok) { // file not found
+          return '';
+        }
+        return response.text();
+      }).then((text) => {
+        const md = new MarkdownIt();
+        this.markdownText = md.render(text);
+      });
+    } catch (error) {
+      // do nothing
```
ckcherry23, line 27: Is there any common scenario where the program control comes to the `catch` statement? If there is, we somehow need a way to tell the user that an error has occurred.

frontend/src/components/c-title.vue
```diff
@@ -0,0 +1,31 @@
+<template lang="pug">
+div(v-html="markdownText")
+</template>
```
ckcherry23, line 3: We may have to set some padding/margins here so that the title is aligned correctly with the rest of the website.

Screenshot 2024-02-11 at 8 30 45 PM

src/main/java/reposense/report/ReportGenerator.java
```diff
@@ -87,7 +87,7 @@ public class ReportGenerator {
     private static final String LOG_ERROR_CLONING_OR_BRANCHING = "Exception met while cloning or checking out.";
     private static final String LOG_UNEXPECTED_ERROR = "Unexpected error stack trace for %s:\n>%s";
     private static final List<String> assetsFilesWhiteList =
-            Collections.unmodifiableList(Arrays.asList(new String[] {"favicon.ico"}));
+            Collections.unmodifiableList(Arrays.asList(new String[] {"favicon.ico", "title.md"}));
```
ckcherry23, line 90: Can we update the CLI syntax reference in the User Guide to include the new `title.md` file? 
Should be this page: https://reposense.org/ug/cli.html

frontend/src/views/c-home.vue
```diff
@@ -120,6 +123,25 @@ const home = defineComponent({
...
   },
+  data() {
+    return {
+      isTitleVisible: false,
+    };
+  },
+  computed: {
+    setScroll() {
+      return {
+        overflowY: this.isTitleVisible ? 'hidden' : 'auto',
```
ckcherry23, line 134: Generally, it is recommended to not have two scrolls in the same direction as "double-scrolling" causes poor UX and leads to many errors. 

The current title component integration uses double-scrolling, and I foresee many bugs in the scrolling behaviour due to this. Have mentioned one of them below:

### Steps to replicate
* Scroll to the complete bottom using the left panel
* Scroll to the top from the right panel until you can see the full title again

### Result
* The left panel does not scroll along with the right one to show the left panel's topmost contents, but the title above it is shown
* The user cannot scroll to show the top contents of the left panel unless they scroll to the bottom again until the title is not seen
* The method to scroll the full view (to show/hide the title) and the left panel is the same, so it seems like the browser randomly picks which one to scroll.

We may have to discuss with Prof @damithc regarding alternative ways to design this. I'd recommend only having the title as part of the left Summary panel, to avoid any scrolling issues.

ckcherry23 (changes requested):
Thanks for the updated implementation @sopa301! 
Have left a few comments for us to look into.

frontend/src/components/c-title.vue
sopa301: Honestly, I have no clue if this try-catch does anything with regard to common problems, because the common problems are handled within the code itself (e.g. missing file is handled by `response.ok` check and markdown doesn't have invalid syntax). It's just to catch freaky errors in I/O that can crash the app, which I'm really not familiar with. Perhaps I could just display the error as a text in the title component, since it's likely that the generated report won't be used if there's an error in the title?

frontend/src/components/c-title.vue
sopa301: I was thinking it that we could give the user complete freedom on the styling of their title component, but if we're primarily using this as a header for text, I do agree that the margins would make it look better.

frontend/src/views/c-home.vue
sopa301: I agree that it's an issue and I've been unable to find a satisfactory solution to this. One thing Prof suggested was to disable panel scrolling until the title is removed from view, which was implemented, but I've been unable to resolve the double scrolling, without resorting to tricks like attempting to disable scrolling in one direction.

Would it be possible to disable title scrolling at the parts where panel scrolling is enabled, except at a certain transition point? Maybe that would cause more bugs?

frontend/src/components/c-title.vue
ckcherry23: Sure, that sounds good! You can add an error message in place of the title component.

docs/ug/cli.md
```diff
@@ -31,7 +31,8 @@ The section below provides explanations for each of the flags.
 <div id="section-assets">
 
 **`--assets ASSETS_DIRECTORY`**: Specifies where to place assets for report generation.
-* Parameter: `ASSETS_DIRECTORY` The directory containing the assets files. A `favicon.ico` file can be placed here to customize the favicon of the dashboard.
+* Parameter: `ASSETS_DIRECTORY` The directory containing the assets files. A `favicon.ico` file can be placed here to customize the favicon of the dashboard,
+while a `title.md` file can be placed to customize the header of the report.
```
ckcherry23, line 35: Just a suggestion, we can explciitly let the user know that they can use Markdown syntax for this.
```suggestion
while a `title.md` file can be placed to customize the header of the report using [Markdown syntax](https://www.markdownguide.org/basic-syntax/).
```

frontend/src/views/c-home.vue
ckcherry23: I want to clarify a few things about your suggestion:
* If title scrolling is disabled when a panel can be scrolled, at what point of time do we enable title scrolling again?
* Since both panels don't scroll together, I'm wondering how it would be possible to scroll both of them to their top when scrolling to the top again to see the title. 

I would prefer the title component to be restricted to just the left Summary panel unless this does not fit in with Prof's requirements. I think this makes scrolling easier to handle and does not affect the UI design much.

1. Across the top of the screen:

image

2. Only on top of the summary panel:

301182943-e9206574-a88e-4d75-8588-bc5c3f303def

docs/ug/cli.md
sopa301: I think that's a good idea for those unfamiliar with markdown. I'll include it!

frontend/src/views/c-home.vue
sopa301: I was initially thinking to enable both scrolling at the point where the title is out of view and either panel is at its top, but it would cause scrolling and UX issues. 

At this point, I agree it's probably better to just confine the panel to the left panel. I'll amend the PR to try this.

frontend/src/components/c-title.vue
```diff
@@ -0,0 +1,39 @@
...
+    } catch (error) {
+      this.markdownText = (error as Error).toString();
+    }
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+.title {
+  padding: 2rem 1.5rem;
```
ckcherry23, line 35: Suggestion: I would make the vertical padding zero so that the title takes as little vertical space as possible. Moreover, vertical spacing can easily be formatted with Markdown as well.
```suggestion
  padding: 0 1.5rem;
```

ckcherry23 (approved):
LGTM!
I would recommend getting approval for the UI from Prof as well, either by deploying an example or demoing it during the lecture slot.

frontend/src/components/c-title.vue
```diff
@@ -0,0 +1,39 @@
...
+        if (!response.ok) { // file not found
+          return '';
+        }
+        return response.text();
+      }).then((text) => {
+        const md = new MarkdownIt();
+        this.markdownText = md.render(text);
+      });
+    } catch (error) {
+      this.markdownText = (error as Error).toString();
```
vvidday, line 27: This is more of a Javascript "gotcha", but the catch block here doesn't catch any errors thrown by the promise chain. This is because the code doesn't wait for the promise to fulfill, and so the `fetch` function returns before the error is thrown asynchronously.

For example, if you try running this code:
```javascript
function f() {
  return Promise.reject("Error");
}

try {
  f()
    .then(() => {
      console.log("Success");
      return 1;
    })
} catch (err) {
  console.log(`Error inside try catch block: ${err}`);
}
```
You'll see that the statement inside the `catch` block never runs.

The two options we have are:
1) Using `.catch` (this makes the outer `try...catch` redundant)
```javascript
function f() {
  return Promise.reject("Error");
}

try {
  f()
    .then(() => {
      console.log("Success");
      return 1;
    })
    .catch((err) => {
      console.log(`This runs and catches the error: ${err}`)
    })
} catch (err) {
  console.log(`This code still never runs: ${err}`);
}
```

or 

2) Using `async/await`, which forces everything after await to wait for the promise to be fulfilled. So we can use the `try...catch` here:
```javascript
async function f() {
  return Promise.reject("Error");
}

try {
  await f()
  console.log("Success")
} catch (err) {
  console.log(`Now this runs: ${err}`);
}
```

frontend/src/shims-vue.d.ts
```diff
@@ -4,3 +4,8 @@ declare module '*.vue' {
   const component: DefineComponent<{}, {}, any>
   export default component
 }
+
+declare module '*.md' {
```
vvidday, line 8: Can I double check what this is for?

vvidday (changes requested):
Thanks for your work on this @sopa301!

Just wanted to highlight some things about the usage of `v-html`: There are some pretty important security concerns we need to be mindful of when using this, since rendering HTML that's partly controlled by users has the potential to lead to XSS vulnerabilities.

Reference: [Vue's docs](https://vuejs.org/guide/essentials/template-syntax.html#raw-html) and [React's equivalent](https://react.dev/reference/react-dom/components/common#dangerously-setting-the-inner-html) (React even deliberately named their property `dangerouslySetInnerHTML`)

Did some reading and it seems like it's OK in this case, as the library we're using (Markdown-it) [claims to output safe HTML as long as the raw HTML option is disabled (which it is by default)](https://github.com/markdown-it/markdown-it/blob/master/docs/security.md). Just wanted to highlight this point to make sure we are mindful of it if we want to introduce any more advanced features like custom styling in the future.

frontend/src/shims-vue.d.ts
sopa301: This used to be for importing markdown files into Vue components, but now that it's no longer being imported at compile time, it's no longer needed. I'll remove it!

ckcherry23 (commented):
@sopa301 Would it be possible to add tests for the title component?

vvidday (approved):
LGTM!

sopa301:
> @sopa301 Would it be possible to add tests for the title component?

I can implement some tests to check for correct handling of exceptions, I don't think testing the markdown itself should be necessary, right?

ckcherry23:
> On the other hand, I don't think testing the markdown itself should be necessary, right?

Yes, since rendering the Markdown is handled by an external library, we don't have to test it.

Since we have to do some setup for the unit tests, we could do that in a different PR. 
We could also look into Vitest instead of Jest since it is recommended by Vue and see if it benefits us (eg. speed).

sopa301:
I have discussed this with Prof @damithc and he suggested allowing raw HTML so that we can do styling with the `md` file. I have enabled it for now.

ckcherry23 (approved):

docs/ug/cli.md
```diff
@@ -31,7 +31,8 @@ The section below provides explanations for each of the flags.
 <div id="section-assets">
 
 **`--assets ASSETS_DIRECTORY`**: Specifies where to place assets for report generation.
-* Parameter: `ASSETS_DIRECTORY` The directory containing the assets files. A `favicon.ico` file can be placed here to customize the favicon of the dashboard.
+* Parameter: `ASSETS_DIRECTORY` The directory containing the assets files. A `favicon.ico` file can be placed here to customize the favicon of the dashboard,
+while a `title.md` file can be placed to customize the header of the report using [Markdown syntax](https://www.markdownguide.org/basic-syntax/).
```
damithc, line 35: @sopa301 Noticed while passing: There should not be a line break here (in markdown files, we should not wrap lines to a specific length).

damithc: @sopa301 Further comments:
1. There should be more details of this feature. For example, code for an example title component.
2. Doesn't feel like this is the right place to document this feature. Perhaps add as a separate section in https://reposense.org/ug/customizingReports.html ?

sopa301: Thanks for the review prof, I'll work on it!

damithc:
Feature in action https://nus-tic2002-ay2324s2.github.io/ip-dashboard/?search=&sort=groupTitle&sortWithin=title&timeframe=commit&mergegroup=&groupSelect=groupByRepos&breakdown=true&checkedFileTypes=java~md~fxml~sh~bat~gradle~txt&since=2024-01-15

PR #2104, opened 2024-01-30
[#2103] Refactor `parser` package for greater organisation of classes

### LINKED ISSUE #2103: Refactor `parser` package for greater organisation of classes

Currently, the way the Java files are organised within the `parser` package is not very neat, with the Exception and Type classes being located on the same level as the Parser classes.

This does not follow how the other classes in the other packages are being organised, with Exception and Model classes being moved into their separate sub-packages.

No.

We can look into moving the Exception classes into its sub-package called `exception` and the types into its sub-package called `types`.

This helps to keep the package neat while also isolating logically similar components from the others.

A PR that aims to solve this potential issue will be linked [here](https://github.com/reposense/RepoSense/pull/2104) once it is up.

No other alternatives have been considered.

### Additional context

NIL

asdfghjkxd:
Fixes #2103

### Proposed commit message

```
Classes in the `parser` package as of now are not properly organised,
with exceptions, types and parser classes being found in the same
directory level.

This does not follow the convention established by other packages,
whereby logically similar classes are further grouped into
sub-packages.

Let's move to refactor the `parser` package to ensure that it follows
the overall structure of the other packages and to increase the level
of organisation within the `parser` package.
```

### Other information

Some accidental commits from another branch have bled through into this branch, however, I have removed the offending code/comments and tried as best as I could to restore it back to the original state.


sopa301 (approved):
LGTM!

gok99 (approved):
LGTM, thanks!

ckcherry23 (approved):
LGTM, thanks for this refactor!

PR #2105, opened 2024-01-30
[#2098] Add show more button for error messages

### LINKED ISSUE #2098: Improve UX of error message box when there are many error messages
When there are many error messages (e.g., many repos missing), the error message box becomes quite tall, pushing the primary report content too far down the page. An example is given below:
image

While not a big problem, it can be an eyesore during early stages of a project course, and can even give the impression that the entire report is broken (because the viewer can't see any non-error content at the first glance).

Perhaps we can show only a limited number of error message lines with a `show all ...` link to show the rest?

jonasongg:
Fixes #2098 

### Proposed commit message

```
Add show more button for error messages

Currently, if there are a lot of error messages, it can block the rest
of the page to the point where users may thing the report failed to
generate at all.

Let's hide extra messages automatically add a show more button if there
are more to display the messages in full.
```

### Other information

Currently, I opted for an easier JS implementation of the feature that displays the button if the number of error messages are above 4. About this, I wanted to ask what the best place would be to put a constant that defines this number (maximum number of error messages displayed without showing more).

Regarding the design of the show more button, I wanted to also get some thoughts on what would work best for RepoSense:
image

image
(default a tag, bad contrast)

image
(default MUI text button)

Lastly, with JS, the transition when clicking show more is visually a bit jarring as the new elements appear programmatically. If needed, I can investigate a way to better animate showing more through CSS, which would also eliminate the need to have a constant to define the number of error messages displayed by default (preliminarily, this would involve adding a transition to max-height).


sopa301 (approved):
LGTM! I think I like the blue option more because it's distinct from the surrounding colours. Also, the rest of RepoSense also uses blue for toggles.

frontend/src/views/c-summary.vue
```diff
@@ -92,7 +94,10 @@
           v-bind:href="getReportIssueEmailLink(errorBlock.errorMessage)"
         )
           span {{ getReportIssueEmailAddress() }}
-      .error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}
+      .error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}\
+    .error-message-box__show-more-container(v-if="Object.keys(errorMessages).length >= 4")
+      a.error-message-box__show-more(v-if="!errorIsShowingMore" v-on:click="toggleErrorShowMore()") SHOW MORE...
+      a.error-message-box__show-more(v-else v-on:click="toggleErrorShowMore()") SHOW LESS...
```
asdfghjkxd, line 100: CI seem to have indicated this area as a potential violation of certain linting rules, might warrant further investigation to know what went wrong!

asdfghjkxd (changes requested):
Good work on the PR! 

However, it seems like CI is failing due to some linting mistakes!

jonasongg:
@sopa301 @asdfghjkxd Thanks for the insight! Fixed the linting issue also

frontend/src/views/c-summary.vue
```diff
@@ -92,7 +94,10 @@
           v-bind:href="getReportIssueEmailLink(errorBlock.errorMessage)"
         )
           span {{ getReportIssueEmailAddress() }}
-      .error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}
+      .error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}\
+    .error-message-box__show-more-container(v-if="Object.keys(errorMessages).length >= 4")
```
ckcherry23, line 98: This should not contain `=` as by default 4 errors are shown according to line 80. Else, we could also change it to show only 3 errors by default.

```suggestion
    .error-message-box__show-more-container(v-if="Object.keys(errorMessages).length > 4")
```

frontend/src/styles/style.scss
```diff
@@ -201,6 +201,20 @@ header {
...
+      }
+
+      &__show-more {
+        border: 1.5px solid mui-color('red', '900');
+        border-radius: 8px;
+        color: mui-color('red', '900');
+        font-weight: bold;
+        padding: .4rem .5rem;
+      }
+
```
ckcherry23, line 217: Can we move these styles to the view file instead of the global styles file? We could also move it later to a dedicated component to display errors.

ckcherry23 (changes requested):

ckcherry23:
@jonasongg Thanks for tacking this issue!

> About this, I wanted to ask what the best place would be to put a constant that defines this number (maximum number of error messages displayed without showing more).

For now, I think we can define it in the view file and later move it to a dedicated Vue component.

> Regarding the design of the show more button, I wanted to also get some thoughts on what would work best for RepoSense.

Currently, the 'Show More' button takes up a lot of vertical space at the bottom of the error box. Moreover, we do not expect most users to click on 'Show More' so it can be more subtle. I think what would work better to fit in with the compact style of RepoSense is using a right-aligned button styled like a red link.

image

> Lastly, with JS, the transition when clicking show more is visually a bit jarring as the new elements appear programmatically. If needed, I can investigate a way to better animate showing more through CSS, which would also eliminate the need to have a constant to define the number of error messages displayed by default (preliminarily, this would involve adding a transition to max-height).

I think the transition isn't too bad but if you're interested in implementing smoother transitions, please go ahead!

damithc:
One other comment about the UI: I feel the fact that some errors have been omitted should be more apparent. The current `Show more` appears like a regular UI control that people will automatically ignore.

For example, the last line of error message could be 
> Remaining error messages omitted to save space. [Show all ...]()

What do you all think? Other alternatives?

ckcherry23:
> For example, the last line of error message could be
> 
> > Remaining error messages omitted to save space. Show all ...

Just some mockups for discussion:

image
image

damithc:
> Just some mockups for discussion:
> 
> image image

Thanks @ckcherry23 I'm OK with either. Will leave you all to decide.

jonasongg:
@ckcherry23 Added the changes as specified!

jonasongg:
@ckcherry23 Wondering if I should add some tests in chartView_errorSummary_messageBox.cy.js to test this show more button; if I do, then I will need to add some repos to summary.json

ckcherry23:
Yes, it would be great if we added a test case to test this behaviour!

However, adding repos to `summary.json` may cause other test cases to break so we try to avoid making any changes to it (the `RepoSense/cypress` branch) unless absolutely necessary.

In your case, would it be okay to show only 3 errors by default?

jonasongg:
If I'm not wrong, the default cypress configuration will only result in one error message, so I'm not too sure on how best to test this!

ckcherry23:
> If I'm not wrong, the default cypress configuration will only result in one error message, so I'm not too sure on how best to test this!

@jonasongg Okay, I thought the Cypress configuration had 4 errors, which is why I suggested the default of 3.

In that case, let's make changes to the `cypress` branch. You can go ahead and add more repos for analysis to the branch locally and see if all the Cypress tests pass, and fix any breaking tests.

jonasongg:
@ckcherry23 Sorry for the late update, just added the aforementioned tests!

ckcherry23:
@jonasongg Please edit the title of the PR to include the issue number

frontend/cypress/tests/chartView/chartView_errorSummary_messageBox.cy.js
```diff
@@ -27,4 +27,24 @@ describe('error summary', () => {
     cy.get('.error-message-box')
       .should('not.be.visible');
   });
+
+  it('number of error messages', () => {
```
ckcherry23, line 31: We could make this test case name more descriptive
```suggestion
  it('can be expanded and collapsed if count > 4', () => {
```

jonasongg:
@ckcherry23 Made the changes! Thanks

frontend/src/views/c-summary.vue
```diff
@@ -92,7 +94,11 @@
           v-bind:href="getReportIssueEmailLink(errorBlock.errorMessage)"
         )
           span {{ getReportIssueEmailAddress() }}
-      .error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}
+      .error-message-box__failed-repo--reason(v-else) {{ errorBlock.errorMessage }}\
+    .error-message-box__show-more-container(v-if="Object.keys(errorMessages).length > 4")
```
vvidday, line 98: Functionality looks good! Can we store the "4" in a variable within this file as mentioned by @ckcherry23?

vvidday (changes requested):

jonasongg:
@vvidday @ckcherry23 I added the variable in data() as I'm not too sure what the idiomatic way to define constants in our Vue files are, let me know if I should change it!

jonasongg:
@vvidday thanks for pointing this out! a very simple fix for most of these tests would like you suggested to remove the `.should(be.visible)` assertion in the tests, but i'm not sure if this is the right way to fix them? another way could be to scroll down so that the elements are visible

sopa301:
We can either use the `exist` [assertion](https://docs.cypress.io/guides/references/assertions) or we can use [`scrollIntoView`](https://docs.cypress.io/api/commands/scrollIntoView) as you mentioned. I think it would be better to use `scrollIntoView` because it would more closely follow an actual user's interaction.

ckcherry23 (approved):
LGTM!

vvidday (approved):
Thanks for the changes, LGTM!

asdfghjkxd (approved):
LGTM!

PR #2108, opened 2024-02-06
[#944] Improve performance

SkyBlaise99:
Part of #944 

### Proposed commit message

```
Currently, the authorship credit analysis process exceeds an hour in
duration, which is significantly longer than the mere 5 minutes required
when the feature is deactivated.

Let's speed up the performance by implementing a caching mechanism and
refining the dynamic programming algorithm utilized for computing the
Levenshtein distance.
```

### Other information

Statistics obtained by running on CS2103/T AY22/23 Semester 2 student repositories.

Forked [here](https://github.com/SkyBlaise99/tp-dashboard) and some statistics recorded on the [Github Actions](https://github.com/SkyBlaise99/tp-dashboard/actions).

| Case | Runtime |
| --- | --- |
| Base Case | 1h 30min |
| Cache | 15min |
| Lev Distance | 1h 8min|
| Cache + Lev Distance | 13 min |


SkyBlaise99:
@damithc Hi prof, the part related to lev dist is [here](https://github.com/reposense/RepoSense/pull/2108/files#diff-5d52a19e6938f176f7308726bbff380beafbd38d0daa8b9ebd418c63d9ee85b5).

MarcusTXK (approved):
LGTM, code is well commented and clean. Great job with the improvement in the performance

gok99 (approved):
LGTM - haven't reasoned through the details of the algorithm updates, but I'll trust the comprehensiveness of the tests.

I'm curious what the memory consumption hit is like if you happened to measure this too - the intelliJ profiler should be able to provide some information.

SkyBlaise99:
> I'm curious what the memory consumption hit is like if you happened to measure this too - the intelliJ profiler should be able to provide some information.

I'm not too familiar with intellij profiler. I tried using it initially but due to the long runtime on local env, I stopped using it. Instead I used Github [telemetry](https://github.com/catchpoint/workflow-telemetry-action) which also measures the ram usage, but it shows a rough scale and not the exact number.

Currently I'm looking at 2GB vs 1.5GB, a 0.5GB increase in ram usage when caching is enabled. As compared to 7GB limit for github actions, this should still be okay.

PR #2116, opened 2024-02-16
[#2109] Add search by tag functionality

### LINKED ISSUE #2109: Give a way to find repos with a specific tag
Suggestion: enhance the search feature to find repos with a specific tag e.g, `tag:A-JUnit`
image

This is a niche use case but could be useful for courses such as CS2103, for students/instructors to easily find repos which has a specific tag

Not sure how hard it is to implement though.

jonasongg:
Fixes #2109

### Proposed commit message

```
Add search by tag functionality

It can be useful to search author/repos by git tags.

Let's add this functionality to make it easier to filter by tags.
```


damithc:
> ### Proposed commit message
> ```
> Add search by tag functionality
> 
> Currently, there is no way to search repos/users by tags. 
> 
> Let's add this functionality to make it easier to filter by tags.
> ```

Two minor comments about the commit message:

It's common to omit the `Currently`, as it is implied anyway.
When adding a new feature, we can omit the 'problem' section i.e., `Currently, there is no way to search repos/users by tags. `
Instead, justify the feature.

```
Add search by tag functionality

It can be useful to ...

Let's add ...

```

damithc:
Another thing: any changes to the UI must come with a corresponding change to the UG, and most likely, tests as well.

jonasongg:
Thanks prof for the feedback! I've made the changes

frontend/src/views/c-summary.vue
```diff
@@ -487,25 +498,51 @@ export default defineComponent({
...
+              if (user) {
+                this.updateCheckedFileTypeContribution(user);
+                res.push(user);
+              }
             }
-            this.updateCheckedFileTypeContribution(user);
-            res.push(user);
+          });
+
+          if (res.length) {
```
sopa301, line 522: This check is replicated in both the `if` and `else` clause. Is there a reason why we do it this way?

sopa301 (approved):
LGTM, I just have one question.

frontend/src/views/c-summary.vue
```diff
@@ -487,25 +498,51 @@ export default defineComponent({
...
+      const res: Array<User> = [];
 
-        // filtering
-        repo.users?.forEach((user) => {
-          if (this.isMatchSearchedUser(this.filterSearch, user)) {
-            this.getUserCommits(user, this.filterSinceDate, this.filterUntilDate);
-            if (this.filterTimeFrame === 'week') {
-              this.splitCommitsWeek(user, this.filterSinceDate, this.filterUntilDate);
+      if (this.filterSearch.startsWith('tag:')) {
+        const tagSearch = this.filterSearch.split('tag:')[1];
```
ckcherry23, line 504: Suggestion: slight naming adjustments
This could be renamed `searchedTags` (plural) to show users can search for multiple tags. You can make similar adjustments in the rest of your changes accordingly.

jonasongg:
thanks @sopa301 @ckcherry23! just implemented the changes suggested

frontend/src/views/c-summary.vue
jonasongg: no actually haha, let me change this

sopa301 (approved):
LGTM!

frontend/cypress/tests/chartView/chartView_toolBar_searchBox.cy.js
```diff
@@ -33,4 +33,53 @@ describe('search bar', () => {
       expect(children).to.equal(1);
     });
   });
+
```
ckcherry23, line 36: We could add more test cases:
* if the search term does not start with the 'tag:' prefix but still has it: 'tag-1 tag:tag-2'
* if there is no 'tag:' prefix in the search term at all: 'tag-1 tag-2'
* if there are multiple tags after the prefix: 'tag: tag-1 tag-2'

These tests may be considered too much for end-to-end testing, but since we do not have unit tests currently, I think they should be added.

jonasongg: just added the tests!

frontend/src/views/c-summary.vue
```diff
@@ -487,25 +498,46 @@ export default defineComponent({
...
+      const res: Array<User> = [];
 
-        // filtering
-        repo.users?.forEach((user) => {
-          if (this.isMatchSearchedUser(this.filterSearch, user)) {
-            this.getUserCommits(user, this.filterSinceDate, this.filterUntilDate);
-            if (this.filterTimeFrame === 'week') {
-              this.splitCommitsWeek(user, this.filterSinceDate, this.filterUntilDate);
+      if (this.filterSearch.startsWith('tag:')) {
+        const searchedTags = this.filterSearch.split('tag:')[1];
```
vvidday, line 504: small nit - could we extract out the search string `tag:` into a variable?

frontend/cypress/tests/chartView/chartView_toolBar_searchBox.cy.js
```diff
@@ -33,4 +33,100 @@ describe('search bar', () => {
...
+    // Enter does not work. Related issue:
+    // Let's manually submit form
+    cy.get('#summary-wrapper form.summary-picker')
+      .submit();
+
+    cy.get('#summary-wrapper #summary-charts')
+      .should('be.empty');
+  });
+
+  it('searching for multiple tags shows results containing all the tags searched', () => {
```
vvidday, line 116: For the tests, would it be possible to also assert that the results contain the tags that are searched for? If I understand correctly, currently the tests would pass even if the search results don't match the searched tags (but still return the correct expected number of results).

If this is not possible, it might be a good idea to add a comment about the assumptions of each test since we are hardcoding the search values (e.g. Assumption: There are exactly 2 charts that match the tag "bb v1.10".)

vvidday (commented):
Thanks for the work on this! Have a question about the tests.

jonasongg:
@vvidday thanks for the comments! i have implemented some changes to address them

vvidday (approved):
thanks for the changes, LGTM!

frontend/cypress/tests/chartView/chartView_toolBar_searchBox.cy.js
```diff
@@ -33,4 +33,131 @@ describe('search bar', () => {
...
+      });
+  });
+
+  it("search field doesn't start with 'tag:' prefix but still contains it shows no results", () => {
+    cy.get('#app #tab-resize .tab-close').click();
+    cy.get('#summary-wrapper input[type=text]')
+      .type('v1.10 tag: v1.10')
+      .type('{enter}');
+
+    // Enter does not work. Related issue:
```
ckcherry23, line 107: The related GitHub issue is missing for some of the comments in this file.

ckcherry23 (approved):
LGTM! Thanks for adding this feature @jonasongg!

frontend/cypress/tests/chartView/chartView_toolBar_searchBox.cy.js
jonasongg: the comment was copied from the existing tests in the file. if that's the case, maybe i can remove the comment since submitting seems self-explanatory anyway?

ckcherry23: Sure you can remove them

PR #2118, opened 2024-02-17
[#2117] Refactor `CliArguments` to conform to `RepoConfiguration`'s Builder Pattern

### LINKED ISSUE #2117: Refactor `CliArguments` to conform to `RepoConfiguration`'s Builder Pattern

Currently, `CliArguments`' Builder patterns is not in line with how `RepoConfiguration`'s Builder pattern was implemented.

As observed in PR #2078, the Builder objects in the codebase are not being reused, and the Builder pattern was adopted as a way to reduce constructor complexity and increase flexibility in how we as developers can construct `RepoConfiguration` or `CliArguments` objects. The current implementation of the Builder pattern for `CliArguments` assumes that `Builder` objects will be reused, but since this is not the case for most parts of the codebase, refactoring the `Builder` class might be beneficial for reducing duplication of class attributes and for increasing overall code equality.

We could look into refactoring `CliArguments` in such a way that it conforms to `RepoConfiguration`'s implementation of the Builder pattern.

This refactoring is related to issue #2076 and PR #2078.

We can adopt the variation of the Builder pattern as detailed in #2078, by simplifying the Builder objects by storing an instance of `CliArguments` within the `Builder` object, and building into that instance directly, reducing attribute duplication.

N/A

### Additional context

N/A

asdfghjkxd:
Fixes #2117 

### Proposed commit message

```
Current implementation of the Builder pattern within `CliArguments`
does not conform that in `RepoConfiguration`.

Let's move to refactor `CliArguments` to reduce class attribute
duplication across `CliArguments` and `CliArguments::Builder` and to
enhance overall code quality.
```

### Other information

N/A


src/main/java/reposense/model/CliArguments.java
```diff
@@ -508,7 +461,9 @@ public Builder reportConfiguration(ReportConfiguration reportConfiguration) {
          * @return CliArguments
          */
         public CliArguments build() {
-            return new CliArguments(this);
+            CliArguments built = this.cliArguments;
+            this.cliArguments = new CliArguments();
```
sopa301, line 465: Can I clarify why we choose to reset `cliArguments` here?

sopa301 (approved):
LGTM, though I have a minor query.

src/main/java/reposense/model/CliArguments.java
asdfghjkxd: Sure! The main motivation behind resetting the Builder object on every `build` operation is to prevent aliasing issues when the user chooses to rebuild `CliArguments` multiple times. Since the cloning operation has not been implemented yet, we decided that the temporary fix would be to reset the Builder object on every build.

A new issue has been released which aims to tackle this issue and to allow duplication of `CliArguments` to avoid resetting the Builder object on every build.

gok99 (approved):
LGTM, could you also link this PR in issue #2119

asdfghjkxd:
Sure thing, I have linked this issue from #2119!

MarcusTXK (approved):
LGTM, thanks for standardising the builders

PR #2124, opened 2024-02-19
[#2119] Implement Proper Deep Cloning for `RepoConfiguration` and `CliArguments`

### LINKED ISSUE #2119: Implement Proper Deep Cloning for `RepoConfiguration` and `CliArguments`

Currently, both Builder patterns implemented in `RepoConfiguration` and `CliArguments` are implemented in such a way that the `Builder` objects are one-use and cannot be reused to build new instances of `RepoConfiguration` and `CliArguments`.

This is not ideal since this prevents us from reusing `Builder` instances to construct duplicate instances of `RepoConfiguration` or `CliArguments`, which violates the Builder pattern and incurs an overhead if we would like to construct duplicate copies of `RepoConfiguration` or `CliArguments`.

This issue is related to issues #2076, #2117 and PR #2078.

We could change the codebase in such a way that classes used in `RepoConfiguration` and `CliArguments`, as well as `RepoConfiguration` and `CliArguments`, implement the [`Cloneable`](https://docs.oracle.com/javase/8/docs/api/java/lang/Cloneable.html) interface and provide an implementation of the `clone()` method to allow for easier duplication and cloning of `RepoConfiguration` and `CliArguments` instances.

N/A

### Additional context

N/A

asdfghjkxd:
Fixes #2119 

### Proposed commit message

```
The current implementation of `RepoConfiguration` and `CliArguments` do
not permit Builder reuse. When a Builder `build` an instance of
`RepoConfiguration` or `CliArguments`, the stored instance of
`RepoConfiguration` or `CliArguments` within the Builder objects is
dereferenced and reset to the base instance of `RepoConfiguration` or
`CliArguments`.

This might result in unnecessary repetition of code to create duplicate
instances of `RepoConfiguration` or `CliArguments`, if needed.

Let's move to implement a cloning method in both `RepoConfiguration`
and `CliArguments`, to reduce code duplication.
```

### Other information

This issue is currently being discussed, and this PR serves as a proof of concept for implementing `clone` for `RepoConfiguration` and `CliArguments`.

### DIFF
--- a/src/main/java/reposense/model/Author.java
+++ b/src/main/java/reposense/model/Author.java
@@ -10,7 +10,7 @@ import java.util.regex.Pattern;
 /**
  * Represents a Git Author.
  */
-public class Author {
+public class Author implements Cloneable {
     public static final String NAME_NO_AUTHOR_WITH_COMMITS_FOUND =
             "NO AUTHOR WITH COMMITS FOUND WITHIN THIS PERIOD OF TIME";
     private static final String UNKNOWN_AUTHOR_GIT_ID = "-";
@@ -207,6 +207,15 @@ public class Author {
         if (!emails.contains(standardGitLabEmail)) {
             emails.add(standardGitLabEmail);
         }
+    }
+
+    @Override
+    public Author clone() throws CloneNotSupportedException {
+        Author clone = (Author) super.clone();
+        clone.emails = new ArrayList<>(this.emails);
+        clone.authorAliases = new ArrayList<>(this.authorAliases);
+        clone.ignoreGlobList = new ArrayList<>(this.ignoreGlobList);
+        return clone;
     }
 }
 
--- a/src/main/java/reposense/model/AuthorConfiguration.java
+++ b/src/main/java/reposense/model/AuthorConfiguration.java
@@ -13,7 +13,7 @@ import reposense.system.LogsManager;
 /**
  * Represents author configuration information from CSV config file for a single repository.
  */
-public class AuthorConfiguration {
+public class AuthorConfiguration implements Cloneable {
     public static final String DEFAULT_BRANCH = "HEAD";
     public static final boolean DEFAULT_HAS_AUTHOR_CONFIG_FILE = false;
     private static final Logger logger = LogsManager.getLogger(AuthorConfiguration.class);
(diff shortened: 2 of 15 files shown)


gok99 (commented):
Thanks for your work on this throrough proof-of-concept!

As it stands, I am still a little ambivalent about this feature, largely because it seems tricky to maintain and at the moment has no clear use cases.

What we're trying to avoid is having to rewrite the same builder on subsequent use (perhaps builders should be unapologetically single use in this sense and error on subsequent builds instead of simply resetting the instance).  However, given that this doesn't happen in a way that causes code clutter (only repeated a small number of times or in a loop), this seems much more readable and maintainable than cascading `Cloneable`.

Possibly null / immutable fields make writing `clone` not as trivial as `equals`, and requires a fair bit of testing to check. I am not sure if all this work is ultimately worth it - let me know what you think!

chan-j-d:
Dropping my two cents, I mostly agree with what @gok99 says. I think we can start with some possible uses of such deep cloning features in testing to see if it is worth the merit. If I understand the changes correctly, there is considerable overhead to maintain this functionality, for example if we wish to add or change some of the fields of a Cloneable class, we would have to also update the clone() method and its tests. Any additional new class used in RepoConfiguration would also require it. 

The main utility is likely in writing tests, as a common template builder can be cloned multiple times to be used in different tests working on a similar RepoConfiguration/CliArguments but with a few fields changed. Correct me if I'm wrong, but the current changes to test code in this PR is testing the cloning method itself. Perhaps, we could see if there are test classes/methods where the cloning can be used in place of what is there.

asdfghjkxd:
@gok99 @chan-j-d Apologies for the late reply, but thank you for your feedback and comments! I agree with @gok99 in that this may be rather troublesome to maintain, and perhaps we should make our builders explicitly one-use only. I also agree with @chan-j-d in that it would require constant updates to `clone` should we decide to modify `RepoConfiguration` and any of its dependent classes sometime in the future, complicating maintenance and hampering the extensibility of our code.

After reviewing the codebase, I too realise that cascading cloning offers little benefit to the overall efficiency of the codebase; perhaps we could look into how we can make some of the crucial classes immutable instead, choosing to minimise side effects rather than explicitly implementing `Cloneable` and forcing developers to call a potentially throwing function `clone`? I believe one of the other initial goal of implementing`Cloneable` was to also minimise side effects and prevent users from accidentally or intentionally altering the contents of a crucial object during runtime. Not sure if this is something that we should look into, but I would love to hear your thoughts on this!

gok99:
> perhaps we could look into how we can make some of the crucial classes immutable instead

I think having more immutability across the board would be great, but would require some significant architectural changes that should probably happen in smaller stages.

@reposense/active-reviewers-1 Regarding the work in this PR (and other PRs like #1094 and #1454 which have been idle for up to 4 years) - PRs of this kind seem to be accumulating in the open PR section. Perhaps we can close them and give them a special tag (more descriptive than `s.DoNotMerge`, maybe something like `ParkedForLater`) for later revisiting.

asdfghjkxd:
@gok99 Perhaps we could first take a look at how we can make our model classes immutable before we take a look at the other classes. I believe the model classes would be the easiest (and with minimal impacts on the entire codebase) classes to refactor first.

gok99:
> I believe the model classes would be the easiest (and with minimal impacts on the entire codebase) classes to refactor first.

Yes, we should start with the model classes but that doesn't seem entirely simple to me either. Perhaps you can create an issue with a checklist to track the immutability refactors for the model classes

PR #2126, opened 2024-02-19
[#2123] Fix zoom bug if zUser is undefined

### LINKED ISSUE #2123: TypeError when visiting a certain URL
### Tell us about your environment

* **RepoSense Version:** Latest version d2018d246e74fa04ae70d2f1f0f16c9e8e29e187
* **Web Browser and Version (if applicable):** Edge (but reproducible on Chrome)

1. Visit this URL: http://localhost:8080/?search=tag%3A&sort=groupTitle%20dsc&sortWithin=title&timeframe=commit&mergegroup=&groupSelect=groupByRepos&breakdown=true&since=2017-01-31&checkedFileTypes=&tabOpen=true&tabType=zoom&zFR=false&zA=eugenepeh&zR=reposense%2FRepoSense%5Bmaster%5D&zACS=133.3170731707317&zS=2017-01-31&zFS=tag%3Av1.8&zU=2024-02-05&zMG=false&zFTF=commit&zFGS=groupByRepos&zFT=gradle~java~js~scss~yml~md&zCST=time
2. RepoSense will fail to load, and a TypeError is thrown in the console: 
image

### If possible, include the URL to your RepoSense report or log files (if any).
RepoSense was run with `./gradlew run -Dargs="--repos https://github.com/reposense/RepoSense.git https://github.com/AY2324S1-CS2103T-W17-4/tp.git https://github.com/AY2324S1-CS2103T-W17-3/tp.git https://github.com/AY2324S1-CS2103T-W17-2/tp.git https://github.com/AY2324S1-CS2103T-W17-1/tp.git --since 31/1/2017"` + `./gradlew hotReloadFrontend`

jonasongg:
Fixes #2123

### Proposed commit message

```
Fix error thrown when zUser is undefined

Currently, an error is thrown when zUser is defined in c-zoom.vue. This
might be caused by zFilterSearch containing a query that returns no
results. Though this happening through normal user interaction is rare
(editing URL, for example), RepoSense should handle this case
gracefully rather than throwing an error.

Let's fix this bug by resetting the zoom view if the error happens.
```


sopa301 (approved):
LGTM!

frontend/src/views/c-zoom.vue
```diff
@@ -404,15 +410,16 @@ export default defineComponent({
...
-      this.filteredUser.commits.forEach((commit) => {
+      this.filteredUser!.commits.forEach((commit) => {
         commit.commitResults.forEach((slice) => {
           Object.keys(slice.fileTypesAndContributionMap).forEach((fileType) => {
             commitsFileTypes.add(fileType);
           });
         });
       });
-      this.fileTypes = Object.keys(this.filteredUser.fileTypeContribution).filter(
+      this.fileTypes = Object.keys(this.filteredUser!.fileTypeContribution).filter(
```
vvidday, line 422: Since the non-null assertion operator is ignored during runtime, this method will throw an error if it is actually called when `this.filteredUser` is undefined.

It is true that in the current logic flow of our code, `this.filteredUser` will not be undefined (since you return early in `created()`), however there is no guarantee that this behaviour will be maintained in the future (e.g. If a developer mistakenly calls `updateFileTypes` before the return statement in `created`, or in another part of the code before `created` is run).

Hence it might be better (more type safety) to handle the undefined case explicitly instead - let me know what you think about this!

Related eslint rule that explains why using `!` might not be super ideal: https://typescript-eslint.io/rules/no-non-null-assertion/

vvidday (commented):
Great job catching the bug & thanks for the fix! Have a minor comment about the use of `!`:

frontend/src/views/c-zoom.vue
jonasongg: thanks for the insight! it seems then the best way to handle the case when `this.filteredUser == null` is to simply return early - what do you think?

vvidday: Yup, sounds good!

jonasongg: ok, fixed!

vvidday (approved):
Thanks, LGTM

ckcherry23 (approved):
Great catch and thanks for the bug fix!

PR #2129, opened 2024-02-19
Fix Blurry Favicon

### LINKED ISSUE #2128: favicon blurry in generated reports
Reported by a student user @aureliony in https://github.com/nus-cs2103-AY2324S2/forum/issues/334

For the dev team's follow up ...

sopa301:
Fixes #2128

### Proposed commit message

```
Fix Blurry Favicon on Reports

The favicon looks blurry on generated reports.

Let's replace this with a clearer image.
```

### Other information

To view the new image, clear the cache.


ckcherry23 (approved):
Thanks for taking this up @sopa301! 

Upon some research, it seems like various sources online have differing opinions on what size(s) of favicons should be used for websites. 
* While the original size we used (16x16) is considered one of the [standard sizes](https://learn.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/samples/gg491740(v=vs.85)?redirectedfrom=MSDN) that should work well for most browsers, it seems like larger sizes such as the 200x200 we are using now can be [scaled down correctly](https://stackoverflow.com/a/48646940/152266610) most of the time. 
* Also, our MarkBind docs site uses the same 200x200 `favicon.ico` file as the site header. 

Considering this, and that this will increase the clarity of the favicon, I think the larger `.ico` file should be okay.

asdfghjkxd (approved):
LGTM!

MarcusTXK (approved):
LGTM, I agree the 200 x 200 ico is okay

Issue #2111, opened 2024-02-14
Fix failing cypress test for range changes in chartview 


MarcusTXK:
### What was expected to happen?

Currently the following frontend test cases are failing in the github CI

```
  1) range changes in chartview should reflect in zoom
       selecting the initial righthand and lefthand boundary:

      Timed out retrying after 30000ms
      + expected - actual

      -'[2019-12-31] [#785] ChartView: fix authors with unaccounted lines (#789): +49 -42 lines '
      +'[2019-12-20] [#46] Show total time after batch processing (#758): +43 -0 lines '
      
      at Context.eval (webpack:///./tests/chartView/chartView_zoomFeature.cy.js:270:7)

  2) range changes in chartview should reflect in zoom
       changing the righthand boundary:

      Timed out retrying after 30000ms
      + expected - actual

      -'[2019-12-31] [#785] ChartView: fix authors with unaccounted lines (#789): +49 -42 lines '
      +'[2019-12-20] [#46] Show total time after batch processing (#758): +43 -0 lines '
      
      at Context.eval (webpack:///./tests/chartView/chartView_zoomFeature.cy.js:296:7)
```

### What actually happened? Please include a screenshot of the output.
![image](https://github.com/reposense/RepoSense/assets/50147457/01eaa660-66a5-448b-b731-6ad2d864e938)

Link to one such failing CI instance
https://github.com/reposense/RepoSense/actions/runs/7862260141/job/21464243013?pr=2078#step:10:415

jonasongg:
Just investigating for a bit, this might be caused by the ramp shrinking as time goes on (since the until date is not set, defaults to today).

For now, I managed to fix this by setting "until" in the filters to an arbitrary time in the past before this problem arose (I chose 31/12/2023)

![image](https://github.com/reposense/RepoSense/assets/120372506/41a14aa9-9e4c-49b4-a573-dce6599df0b4)

This seems to be the same problem as #2066. Let me know if this fixes the problem for you and if I should create a PR for this!

MarcusTXK:
Hi @jonasongg, thanks for the quick investigation. I think your suggestion is a good way to solve it, works for me too. Yeah I think it is fine to just set an arbitrary time as this will prevent the issue from happening again as more time passes. Please do feel free to create a PR for this if you have the time to do so.

ckcherry23:
Hi @jonasongg, I'd recommend prioritising the PR to fix this first as other PRs would be blocked from merging until all CI tests pass. Thank you!

Issue #1613, opened 2022-01-14
Add plugin to perform visual testing with cypress


gerhean:
**What feature(s) would you like to see in RepoSense?**
Testing to prevent visual regressions.

**Is the feature request related to a problem?**
Cypress is unable to test for visual regressions as [stated here](https://docs.cypress.io/guides/tooling/visual-testing#Functional-vs-visual-testing)

If implemented, we can catch problems such as regressions to #1584

**Additional context**

The cypress documentation recommends (cypress-plugin-snapshots)[https://github.com/meinaart/cypress-plugin-snapshots] but other plugins may be better.

This is a very low priority feature, I don't expect this to be done soon.

LuoZhijie-tom:
Hi is this issue still open? It is OK to work on this issue?

LuoZhijie-tom:
Hi is this issue still open? It is OK to work on this issue?

dcshzj:
Hi @LuoZhijie-tom, this issue is still open but it might be a little difficult for new contributors. Perhaps I can suggest to [work on other tasks suitable for contributors](https://github.com/orgs/reposense/projects/1/views/1?filterQuery=status%3A%22For+contributors%22) instead?

LuoZhijie-tom:
OK sure!

PR #2137, opened 2024-02-22
[#2136] Add Tests for Segment CSS

### LINKED ISSUE #2136: Test code highlighting in Code Panel

Add tests to check if code highlighting is working:
* when there is a single author
* when multiple authors are merged in a repo group
* non-attributed lines are not highlighted

### Additional context

#2134

sopa301:
Fixes #2136

### Proposed commit message

```
Write Tests for Code Highlighting for Segment

A regression has occurred during refactoring of the Segment component.
It would be good to write tests to catch regressions if it happens
again for this component.

Let's write tests to catch further regressions of the same nature.
```

### Other information

Hex values [cannot](https://stackoverflow.com/questions/66438459/is-there-a-way-to-make-check-on-hex-color-cypress) be used directly because of Cypress limitations. Since we'll need an external library ([Chai](https://www.chaijs.com/plugins/chai-colors/)) for that, I have opted to use `rgb` values instead.

I chose to suppress warnings for testing the colors of a commit history of `_colors.scss` instead of using another file because it's a `scss` file which is unlikely to be touched, while the other files are `java` files. This may not be so necessary if the below concern is addressed.

Also, it seems that for quite a few tests (including the ones these are derived from), the `until` date is set to the current date, which can produce regressions like #2111. Additionally, for certain types of tests (like the ones which group a repo's entire commit history), certain assumptions (like the latest author for a line of code) can be broken which may cause the test to break. 

The most convenient solution would be to set the `until` date for relevant tests to a fixed date in the past (like 2021), while the generated report for testing could also be anchored to a past date. Perhaps this should be addressed in another PR so we can avoid more nasty surprises.


sopa301:
I've also looked at #1613 suggesting snapshot testing, but I have a few concerns. For example, there may be differences between systems that make snapshots unreliable.

Here's a sample report from dashboard #2137 (The relevant commit history is taken from Jor, testrepo-Delta):
image

And here's the locally built equivalent:
image

Perhaps this points to a deeper bug, but I've opted to stay away from this solution for now. If anyone has an idea of what's causing or how to fix this, do let me know!

frontend/cypress/tests/codeView/codeView_codeHighlighting.cy.js
```diff
@@ -0,0 +1,68 @@
...
+      .click();
+
+    cy.get('#tab-authorship .files', { timeout: 90000 })
+      .should('be.visible');
+
+    cy.get('.hljs-comment').contains('* MUI Colors module') // eugenepeh
+      .parent() // .line-content
+      .parent() // .code
+      .should('have.css', 'background-color')
+      .and('not.eq', 'rgb(255, 255, 255)') // #ffffff
```
ckcherry23, line 41: I think the background color used is always consistently chosen, is it not possible to check for the exact color here instead of checking for not white? 

Doing this would make the test case more robust, ensuring the merged group code is highlighted based on author's color and not all just green.

ckcherry23 (changes requested):

frontend/cypress/tests/codeView/codeView_codeHighlighting.cy.js
sopa301: Thanks for pointing that out! I went to look at how the colours are generated (in `c-authorship.vue`) and it turns out that the first 10 colours are consistently chosen, while the 11th onward is randomly generated. I'll amend the code since we're only testing the first two colours.

frontend/cypress/tests/codeView/codeView_codeHighlighting.cy.js
```diff
@@ -0,0 +1,67 @@
...
+      .click();
+
+    cy.get('#tab-authorship .files', { timeout: 90000 })
+      .should('be.visible');
+
+    cy.get('.hljs-comment').contains('* MUI Colors module') // eugenepeh
+      .parent() // .line-content
+      .parent() // .code
+      .should('have.css', 'background-color', 'rgba(30, 144, 255, 0.19)') // #1e90ff, transparencyValue 30
+      .then((color) => {
```
ckcherry23, line 41: This could be more descriptive.
```suggestion
      .then((firstAuthorColor) => {
```

ckcherry23 (approved):
LGTM, thanks for adding these tests!

vvidday (approved):
LGTM, thanks for providing context & great job on the comments in the test file describing assumptions & the navigation of parent elements!

ckcherry23:
@reposense/active-developers Please help us with a final review for this PR!

jonasongg (approved):
LGTM!

Issue #2145, opened 2024-03-04
Dockerisation of RepoSense


asdfghjkxd:
### What feature(s) would you like to see in RepoSense

Currently, we are facing issues regarding the standardisation of tests for both the frontend and backend.

Considering that there is a wide range of possible combinations of dependencies that need to be tested to ensure a consistent look and performance across all platforms, the use of a containerisation framework such as [Docker](https://www.docker.com/) would help accelerate our current workflows and increase the overall coverage of our tests.

Moreover, Docker provides us with a new way to help our users who are facing dependency issues, since the dependencies and codebase are bundled and abstracted away from the user, who just needs to use Docker to run RepoSense without worrying if the dependencies would cause issues for them.

This would also allow us to more confidently upgrade our existing dependencies while providing support for older dependencies through older, backwards-compatible builds of RepoSense.

For the more tech-savvy users or admins, the introduction of Docker containers provides them with a way to use other container orchestration tools such as Kubernetes to better manage multiple running instances of RepoSense on their cloud/web services, increasing the overall availability of RepoSense for their target users.

Docker integrates nicely with Github Actions [(more here)](https://docs.docker.com/build/ci/github-actions/) and can be easily orchestrated using Docker Compose, allowing us to create multiple Docker containers and store them on Github Packages or Docker Hub, or to use it for mainly testing purposes using Github Actions.

There are also ways for the Docker CLI to accept new parameters at the CLI during execution and to pass them to the Docker container, mimicking how our existing CLI commands work. We can also mount a user's local file system into the Docker container so that they can view and analyse any other local repositories on their local file system, maintaining RepoSense's key selling points and features.

### Is the feature request related to a problem?

The use of Docker might help resolve current issues related to frontend and backend testing issues, as well as issues related to dependency clashes.

### If possible, describe the solution

We could look into integrating Docker into our codebase and workflows, by creating `Dockerfile`s and `docker-compose` files, and experiment with how we can most efficiently pass in parameters from the Docker CLI into the Docker containers, as well as how we can mount our local file systems into the Docker containers we create and run.

### If applicable, describe alternatives you've considered

This feature might prove to be another dependency that we need to resolve now since we have to ensure that the Docker containers we generate and upload would be the most up-to-date version and that there could be multiple Docker containers at any time with differing dependencies for different user groups.

This might prove to be more cumbersome than leaving things as it is right now.

### Additional context

If we were to go ahead with the implementation of Docker, we need to settle a few things:

[...]

PR #2149, opened 2024-03-09
[#2145] Dockerisation of RepoSense

asdfghjkxd:
Part of #2145 

### Proposed commit message

```
Currently, there are some issues with regard to consistent testing of
code, especially for the frontend code, due to the difference in
environments where the test cases are run (developer's machine VS
Github Action's workers). This inconsistency in testing results makes
it hard for us to really distinguish false negatives/false positives.

Moreover, this inconsistency in the environment where RepoSense is
installed may also result in dependency issues.

In this PR, a proof-of-concept Dockerfile and docker-compose.yml file
will be provided first to test the viability of such an integration
into RepoSense. Future implementations would see further integration of
Docker into Github Actions, more complicated Docker build processes and
tests and documentation on how to maintain these Docker files and
builds.

Let's move to integrate Docker and Docker Compose into RepoSense to
mitigate and potentially solve some of the aforementioned issues, by
providing users and developers a more standardised environment where
they can run or test RepoSense.
```

### Other information

This PR is part of the following steps outlined in the parent issue:

- [X] Create Ubuntu Dockerfiles
- [X] Integrate Ubuntu Dockerfiles with Github Actions

For this PR, only Ubuntu/Linux images are explored, as the rest are in development. Ubuntu/Linux images are also easier to work with and far less fussy compared to Windows and MacOS images, hence starting work on Ubuntu/Linux might be a helpful proof of concept for the viability of Docker for use for testing.


asdfghjkxd:
Do note that the Docker containers currently do not conduct frontend tests using cypress, even though the dependencies are installed on the base Ubuntu containers, as they are consistently failing CI. Will look into the cause and update this thread once I have isolated the cause of the error.

docker/Dockerfile.ubuntu2204
```diff
@@ -0,0 +1,74 @@
...
+#     - https://stackoverflow.com/questions/40873165/use-docker-run-command-to-pass-arguments-to-cmd-in-dockerfile
+# - Installation of Java on Ubuntu
+#     - https://stackoverflow.com/questions/49914574/install-jdk-8-update-172-in-dockerfile-with-ubuntu-image
+# - Installation of NodeJS via Node Version Manager (NVM)
+#     - https://stackoverflow.com/questions/36399848/install-node-in-dockerfile
+
+FROM asdfghjklxd/reposense-images:ubuntu-22-04 AS setup-env
+
+LABEL os=ubuntu
+LABEL "os.version"="20.04"
```
sopa301, line 26: This should probably be 22.04.
```suggestion
LABEL "os.version"="22.04"
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,49 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout V4
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Build Docker Image
+        run: |
+          docker build -t ${{ matrix.tag }} -f ${{ matrix.dockerfile }} -o ${{ env.OUTDIR }} .
+      # - name: Upload Code Coverage (Java 8)
```
sopa301, line 36: Is this going to be re-added?

docker-compose.yaml
sopa301, line 1: As discussed, since we're not using this for the CI, we can probably put this in a folder marked for local testing.

sopa301 (changes requested):
LGTM! We just need to fix the Cypress issues.

asdfghjkxd:
As mentioned in issue #2145, we are not pursuing this idea, hence I am closing out this PR.

Issue #2151, opened 2024-03-10
Frontend DevOps: Update Node.js dependencies


sopa301:
### What feature(s) would you like to see in RepoSense

Some of our frontend dependencies have been flagged as out of date:
![image](https://github.com/reposense/RepoSense/assets/96387349/f1db6da0-eeeb-4ef5-a221-d9cae8e9cc32)
![image](https://github.com/reposense/RepoSense/assets/96387349/aeed0cf5-3e04-400d-9abf-854cab42cce2)

### Additional context
This issue is a continuation of #1871.

PR #2153, opened 2024-03-10
[#2151] Update Stylelint

sopa301:
Part of #2151.

### Proposed commit message

```
Update Stylelint

The depreciated rules have been moved to a dedicated plugin. We can now
update our Stylelint linter to the latest version.

Let's do so, as well as update related dependencies.
```

### Other information

This was done in a separate PR because:
1. It involves breaking changes
2. It replaces a large part of the Stylelint installation

There's a few new rules that the updated Stylelint has added. I have disabled them for now, and we should discuss which ones to enable. Here's the list:
- `shorthand-property-no-redundant-values`
- `declaration-empty-line-before`
- `selector-class-pattern`
- `color-function-notation`
- `alpha-value-notation`
- `comment-empty-line-before`
- `property-no-vendor-prefix`
- `selector-pseudo-element-colon-notation`
- `at-rule-empty-line-before`
- `font-family-name-quotes`
- `color-hex-length`
- `scss/at-import-no-partial-leading-underscore`
- `scss/at-import-partial-extension`
- `scss/load-no-partial-leading-underscore`
- `scss/double-slash-comment-empty-line-before`
- `scss/double-slash-comment-whitespace-inside`
- `scss/dollar-variable-pattern`


asdfghjkxd (approved):
LGTM! Good work on the updating of NodeJS to version 18!

frontend/.stylelintrc.json
```diff
@@ -1,43 +1,60 @@
 {
   "plugins": [
+    "@stylistic/stylelint-plugin",
     "stylelint-order"
   ],
   "extends": [
-    "stylelint-config-recommended-scss",
+    "stylelint-config-standard",
+    "stylelint-config-standard-scss",
```
vvidday, line 8: Can I check why we are moving from the recommended config to the standard + standard scss config?

vvidday (commented):
Thanks for taking this on @sopa301, definitely agree with disabling the new rules for now and leave those up for discussion (feel free to file a separate issue for that)

frontend/.stylelintrc.json
sopa301: `standard` was suggested by the [15.0.0 migration guide](https://stylelint.io/migration-guide/to-15) and `standard-scss` was suggested by the `recommended-scss` [github](https://github.com/stylelint-scss/stylelint-config-recommended-scss).

Do let me know if you want to keep it at the old configs!

vvidday: Got it, thanks! Let's go with these configs.

vvidday (approved):
LGTM, thanks!

Will have to trouble you to fix the conflicts before merging

frontend/babel.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
   presets: [
     '@vue/cli-plugin-babel/preset',
-  ],
-};
+    ],
+}
```
ckcherry23, line 5: I think this should be restored?

ckcherry23 (approved):
LGTM!

PR #2163, opened 2024-03-18
[#2148] Show tags on the ramp chart

### LINKED ISSUE #2148: Show tags on the ramp chart
Somewhat related to #2109 

I'm wondering if we should support the ability to see the tags on the ramp chart side. For example,
image

Benefits: Easy to see which repos has which tags, can do a Ctrl+F for a tag

Considerations: performance, UI clutter

There is some overlap with #2109 but they are not exactly the same

If we were to implement it, most likely tags shouldn't be shown by default, but the user can make the appear by ticking a checkbox.

jonasongg:
Fixes #2148

### Proposed commit message

```
Show tags of repo on the ramp chart

It can be useful to view all the tags of each repo that appears on the
ramp chart. Though this information is visible if the repo is merged
and the user click to see the breakdown of each commit, it can still be
useful to see this information upfront (for searching Ctrl+F perhaps).
This feature would be toggleable with a checkbox to reduce clutter.

Let's add this functionality to add convenience when searching through
RepoSense reports with tags.
```

### Other information

Pending changes to documentation and tests!


damithc:
@jonasongg can you post a screenshot of how the feature appears. I couldn't see any tags in the dashboard preview (I guess the repo we use for the preview doesn't have any tags?).

jonasongg:
@damithc 
here's a screenshot showing:
1. the checkbox to toggle viewing all tags on top
2. what it looks like if the tags overflow into a new line
3. what it looks like if the tags don't overflow 
image

jonasongg:
@sopa301 that's a great suggestion, thanks!

damithc:
Thanks for the screenshot @jonasongg 
Looks good. Perhaps `view repo tags` -> `show tags`?

jonasongg:
@damithc changed the name, thanks!

sopa301 (approved):
LGTM!

vvidday (commented):
Thanks for working on this @jonasongg!

Can I check what the behaviour is like if we sort by author, and there's multiple different repositories under the same author? I haven't tested myself, but it seems that the current version will display all the tags (which may be from different repositories) at the top (author) level

jonasongg:
@vvidday thanks for spotting this! i changed it such that if we group by author, the tags will appear on the repo (second) level instead. i also made it such that viewing all tags is stored in the url

frontend/src/components/c-summary-charts.vue
```diff
@@ -258,6 +270,19 @@
           v-if="filterGroupSelection === 'groupByNone' && sortGroupSelection.includes('totalCommits')"
         ) {{ getPercentile(j) }} %&nbsp
           span.tooltip-text.right-aligned {{ getPercentileExplanation(j) }}
+        .summary-chart__title--tags(
+          v-if="(filterGroupSelection === 'groupByAuthors' || filterGroupSelection === 'groupByNone')\
+            && viewRepoTags"
+        )
+          a.tag(
+            v-for="tag in getTags(repo, user)",
+            v-bind:href="getTagLink(repo[0], tag)",
```
vvidday, line 279: Is it correct to always get the link from the first repo?

frontend/src/components/c-summary-charts.vue
```diff
@@ -139,6 +139,18 @@
           v-if="sortGroupSelection.includes('totalCommits')"
         ) {{ getPercentile(i) }} %&nbsp
         span.tooltip-text.right-aligned {{ getPercentileExplanation(i) }}
+      .summary-charts__title--tags(
+        v-if="filterGroupSelection === 'groupByRepos' && viewRepoTags"
+      )
```
vvidday, line 144: Can we store this boolean in a computed variable with a name that describes what this condition is? (similar to `isChartGroupWidgetMode` and `isChartWidgetMode` in this file)

frontend/src/components/c-summary-charts.vue
```diff
@@ -258,6 +270,19 @@
           v-if="filterGroupSelection === 'groupByNone' && sortGroupSelection.includes('totalCommits')"
         ) {{ getPercentile(j) }} %&nbsp
           span.tooltip-text.right-aligned {{ getPercentileExplanation(j) }}
+        .summary-chart__title--tags(
+          v-if="(filterGroupSelection === 'groupByAuthors' || filterGroupSelection === 'groupByNone')\
+            && viewRepoTags"
```
vvidday, line 275: Same as above re: descriptive boolean name

vvidday (commented):
Thanks for the changes @jonasongg, definitely agree with the new proposed way of displaying tags when group by author is selected. And great job on adding the state of the toggle to the url!

Have a question about the logic of getting tag links when grouped by author:

frontend/src/components/c-summary-charts.vue
jonasongg: should be yes, because this part only runs when we groupByRepos, and the array is of users of the same repo. meaning each link retrieved from the repo is the same!

frontend/src/components/c-summary-charts.vue
jonasongg: fixed!

frontend/src/components/c-summary-charts.vue
jonasongg: fixed!

frontend/src/components/c-summary-charts.vue
vvidday: Sorry, might be misunderstanding but isn't this part run when we groupByAuthors?
```
    .summary-chart__title--tags(v-if="isViewingTagsByAuthor")
          a.tag(
            v-for="tag in getTags(repo, user)",
            v-bind:href="getTagLink(repo[0], tag)",
```

vvidday (approved):
Sorry for the delay - LGTM

ckcherry23 (approved):
LGTM @jonasongg!
As discussed, let us create an issue for the remaining Cypress test required.

Issue #2164, opened 2024-03-19
Reports show only one team


damithc:
Looks like something broke in the last 1-2 commits. The report shows all students under only one team 

Example: https://nus-cs2113-ay2324s2.github.io/tp-dashboard/?search=&sort=groupTitle&sortWithin=title&timeframe=commit&mergegroup=&groupSelect=groupByRepos&breakdown=true&checkedFileTypes=docs~functional-code~test-code~other&since=2024-02-23

![image](https://github.com/reposense/RepoSense/assets/1673303/641b4402-b91f-4ade-9387-cca8e89de5c7)

Confirmed: the problem goes away when I switched to a version 2 commits aga

PR #2167, opened 2024-03-22
[#2109] Add search by tag functionality

### LINKED ISSUE #2109: Give a way to find repos with a specific tag

jonasongg:
Fixes #2109

### Proposed commit message

```
Add search by tag functionality

It can be useful to search author/repos by git tags.

Let's add this functionality to make it easier to filter by tags. This
commit also fixes a bug that existed in a previous version of the
feature which resulted in all users being shown to belong to same group.
```


sopa301 (approved):
LGTM! I realised that the bug may be caused by my comment in the original PR suggesting to move the `full.push(res)` statement block out of the if-else statement (oops).

I think the `getFilteredRepos` method should probably be refactored sometime soon though, since there are some lines going up to 5(!) levels of indentation. In another PR maybe?

vvidday (approved):
Thanks @jonasongg - can I confirm that we can't add a test for the bug in #2164 due to our testing report only containing one group?

vvidday:
> I think the `getFilteredRepos` method should probably be refactored sometime soon though, since there are some lines going up to 5(!) levels of indentation. In another PR maybe?

@sopa301 Great point, I agree that it could be refactored to increase readability - feel free to file an issue for this!

ckcherry23 (approved):
LGTM, thanks for fixing the bug @jonasongg! 

Regarding the regression, I think we can file a new issue to add more repos to our Cypress setup and implement tests so that this doesn't happen again.

jonasongg:
@vvidday Sorry for the late response, but yes, it's because our testing config contains only one repo. I will be creating an issue to add more repos soon!

PR #2173, opened 2024-03-25
[#2001] Extract c-file-type-checkboxes from Summary, Authorship and Zoom

sopa301:
Part of #2001.

### Proposed commit message

```
Extract c-file-type-checkboxes from Summary, Authorship and Zoom

The checkbox collection is duplicated across many components.

Let's abstract them into a separate file.
```

### Other information

Progress:

- [x] Extract from Summary
- [x] Extract from Authorship
- [X] Extract from Zoom


asdfghjkxd (approved):
LGTM!

frontend/src/components/c-file-type-checkbox.vue
```diff
@@ -0,0 +1,124 @@
...
+      },
+    },
+  },
+  methods: {
+    getFontColor(color: string): string {
+      return window.getFontColor(color);
+    },
+    updateSelectedFileTypesHash(): void {
+      this.$emit('update-selected-file-types-hash');
+    },
```
vvidday, line 91: Is there a need to differentiate the `update-selected-file-types-hash` and `update:selectedFileTypes` emits? I personally feel that it might be okay to remove the `update-selected-file-types-hash` emit since this component doesn't need to know about the concept of hashes at all - it can just emit `update:selectedFileTypes` whenever the selectedFileTypes is updated/changed, and let the parent component decide what to do in response to that (which might include updating the hash).

For example, in c-summary, we listen to the emit and run the `getFiltered` method, but we run it _in repsonse to the checkedFileTypes changing_, rather than in response to needing to update the selected file types hash (updating the hash is one of the things done in `getFiltered`, among others).

vvidday (commented):
Really great job on this PR @sopa301! LGTM, just a minor nit on the naming/design of the emits of the new component.

frontend/src/components/c-file-type-checkbox.vue
sopa301: I agree that it would be better to hide this interaction from the component! I have updated the code.

vvidday (approved):
Thanks! LGTM

frontend/src/components/c-file-type-checkbox.vue
ckcherry23, line 1: Nit: I think we should rename this file to `c-file-type-checkboxes.vue` to better represent what it does.

ckcherry23 (approved):
LGTM! Thanks for the refactor @sopa301! I think we can merge this once the new file is renamed.

frontend/src/components/c-file-type-checkbox.vue
sopa301: Done!

frontend/src/components/c-file-type-checkboxes.vue
```diff
@@ -0,0 +1,120 @@
...
+    },
+    getLabel(label: {
+      fileTitle: string,
+      fileType: string,
+      lineCount: number,
+      blankLineCount: number,
+    }): string {
+      return `${label.fileType}\xA0\xA0`
+        + `${label.blankLineCount}\xA0\xA0`
+        + `(${label.lineCount - label.blankLineCount})\xA0`;
```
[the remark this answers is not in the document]
sopa301: Oops, I'll create a PR to fix it!

Issue #2157, opened 2024-03-12
Give an  easy way for a person to set up a report of past OSS contributions


damithc:
Usage: A user (e.g., a student) setting up an online report of their open-source coding work.

Steps:
1. Forks a provided repo (e.g., `https://github.com/reposense/code-report`) to own GitHub account
2. Updates a config file in the fork to specify repos contributed to, author names used in each repo
3. After committing the change, and CI run, the report is available for viewing (and can be linked from their resume, home page etc.)

A further optimization is to show each ramp for the time period in which there are relevant commits. i.e., in the ramp chart below, show the part in the red box only.
![image](https://github.com/reposense/RepoSense/assets/1673303/85692376-632b-48a6-ba72-060bd650780f)
This means each ramp chart will have its own timeline, and we'll have to show the timeline for each ramp chart.

ckcherry23:
Related to #1431

PR #2180, opened 2024-04-01
Add optimise timeline feature

### LINKED ISSUE #2162: Add an option to only show active period a user was working on a repo
Add a feature that allows users to display only the part of the summary chart where the user has been active.

For example, if a student worked on project A for 6 weeks, and then worked on project B for 6 weeks, currently there would be a sizeable amount of blank space for the latter half of project A and earlier half of project B. 

An alternate solution as described in #1431 is to instead allow the user to specify the display time range for each summary chart.

### Additional context
Related to #1431

jonasongg:
Fixes #2162
Part of #2157

### Proposed commit message

```
Add optimise timeline feature

Currently, the ramps have a lot of empty space if the since and until
dates are specified far apart. 

Let's add a checkbox that can eliminate this empty space for each ramp.
This is especially useful if RepoSense is to be used to set up a report
of an individual's past OSS contributions.
```


sopa301 (approved):
LGTM! Just some nitpicks:

1. The feature seems to be slightly bugged on small time scales, like < 1 week. Example:
image
It's not a big deal since the new graph still looks better than before.

2. The "state" of the button doesn't persist across refreshes. The "breakdown by file type" and "merge all groups" checkboxes retain their state when refreshed, so I think it would be good for the new checkbox to follow this behaviour as well!

jonasongg:
@sopa301 thank you and fixed!

frontend/src/components/c-ramp.vue
```diff
@@ -267,4 +298,17 @@ export default defineComponent({
...
   }
 }
+
+.ramp-padding {
+  height: 100%;
+  position: relative;
+}
+
+.date-indicators {
+  @include mono-font;
```
ckcherry23, line 308: Is there any reason we are using a monospaced font for the dates? I think we should follow the rest of the site, and use the default sans serif font for all dates.

ckcherry23 (changes requested):
Great start @jonasongg! The optimised timelines are looking good and would definitely improve the usability of RepoSense!

While most of the changes you've made look good, there are some things left out and some design decisions to discuss too.

1. [High priority] The chart zoom feature (selected using Ctrl/Cmd + click) does not work correctly when the timelines are optimised, i.e., the dates used for the zooming in are the same as the original dates instead of the optimised dates, which is misleading to the user. 
2. [Low priority] When I open the Commits Panel for an optimised chart, the chart shown at the top of the Commits Panel is not optimised. 
3. [Discuss] When the report is grouped by repos, the various authors under the repo may have different timelines. This discrepancy may make it difficult to compare when different teammates are contributing relatively. We may want to make the timeline duration consistent across a repo group (@damithc). However, when grouping by authors, this doesn't matter. 
4. [Discuss] We may want to think of a more intuitive name for "Optimise timeline", because it is not immediately clear what this feature does to end users. I would suggest something along the lines of "Condense charts".

ckcherry23:
Example repo before and after the optimise timelines option is activated.

Before:
Screenshot 2024-04-20 at 6 29 51 PM

After:
Screenshot 2024-04-20 at 6 29 35 PM

damithc:
Good inputs @ckcherry23 

> * [Discuss] When the report is grouped by repos, the various authors under the repo may have different timelines. This discrepancy may make it difficult to compare when different teammates are contributing relatively. We may want to make the timeline duration consistent across a repo group (@damithc). However, when grouping by authors, this doesn't matter.

One option (probably easier too) is to take the stance that 'optimized' timelines isn't suitable for timeline comparisons. What do you all think?

> * [Discuss] We may want to think of a more intuitive name for "Optimise timeline", because it is not immediately clear what this feature does to end users. I would suggest something along the lines of "Condense charts".

'Trim'?

jonasongg:
@ckcherry23 @damithc thank you for your comments! will do the necessary changes asap.

as for when we are grouping by repos, it shouldn't be too difficult to make the timeline consistent across a group (i think we can just select the minimum and maximum commit date of the group instead of the user). i'm ok to implement either option!

i think "trim timeline" works!

damithc:
> as for when we are grouping by repos, it shouldn't be too difficult to make the timeline consistent across a group (i think we can just select the minimum and maximum commit date of the group instead of the user). i'm ok to implement either option!

@jonasongg This might dilute the  effect of the feature, if different authors have authored code at different times. For exampe, if the following  author did one more commit in the place pointed by the arrow.

image

So, I'd rather keep the current behavior first, and revisit this after using it for a while.

damithc:
@jonasongg BTW, just now I realized the zoom feature and this feature must have lot in common, this that this one is like an auto-zoom. Is there some code reuse? Just curious.

jonasongg:
@damithc got it. for the code reuse question, there is a bit of reuse in the sense that the logic to calculate the position of the ramp slices is the same, but the code that this feature adds is just finding the minimum and maximum commit dates

frontend/src/components/c-summary-charts.vue
```diff
@@ -627,6 +656,15 @@ export default defineComponent({
...
       const clonedUser = JSON.parse(JSON.stringify(user));
+
+      if (this.optimiseTimeline) {
+        const optimisedMinimumDate = this.getOptimisedMinimumDate(user);
+        const optimisedMaximumDate = this.getOptimisedMaximumDate(user);
+
+        if (optimisedMinimumDate != null) since = window.getDateStr(optimisedMinimumDate);
+        if (optimisedMaximumDate != null) until = window.getDateStr(optimisedMaximumDate);
+      }
+
```
ckcherry23, line 667: Currently, using the tab zoom on an optimised/trimmed timeline doesn't work. The zoomed-in chart uses the duration of the optimised timeline min and max dates instead of the user selected zoom duration. I think removing the logic I highlighted here may help, as the `since` and `until` values passed to `opneTabZoom()` already use the expected since and until values for the zoom function.

frontend/cypress/tests/chartView/chartView_optimiseTimeline.cy.js
ckcherry23, line 1: Can we add a Cypress test for whether the optimised timeline feature works with the chart zoom feature? This can be done in `chartView_zoomFeature.cy.js`.

frontend/src/components/c-summary-charts.vue
```diff
@@ -610,14 +628,25 @@ export default defineComponent({
...
+        } else {
+          // additional day was added to include the date represented by filterUntilDate
+          const tdiff = (new Date(this.filterUntilDate)).valueOf() - (new Date(this.filterSinceDate)).valueOf()
+              + window.DAY_IN_MS;
+          const idxs = this.drags.map((x) => (x * tdiff) / 100);
+          const tsince = window.getDateStr(new Date(this.filterSinceDate).getTime() + idxs[0]);
+          const tuntil = window.getDateStr(new Date(this.filterSinceDate).getTime() + idxs[1]);
+          this.drags = [];
+          this.openTabZoom(user, tsince, tuntil, isMerged);
+        }
```
ckcherry23, line 649: This part also has a lot of duplicated calculation logic. If possible, define the `since` and `until` dates locally and use them in common calculation logic.

ckcherry23 (changes requested):
Thanks for the updates @jonasongg! I've left a few comments to address below.

Can you also add brief documentation for this feature in the UG?

ckcherry23 (commented):
Awesome work @jonasongg! The ramps and the chart zoom work well now with the trim timelines feature!

However, fixing the zoom feature broke the optimization of the default ramp chart shown on the Commits Panel when "trim timelines" is enabled, sorry for the oversight. Maybe the best option is to pass the optimized timeline min/max dates conditionally along with `filterSinceDate` and `filterUntilDate` for this in lines `103` and `221` of `c-summary-charts.vue`? 

The rest of the changes look good to me. As discussed, since we need other collaborators' help for adding Cypress tests, I think we can ignore that comment and ask for our second reviewer feedback now.

ckcherry23 (approved):
LGTM! Thanks for this @jonasongg! This would be a super useful feature for setting up personal RepoSense reports!

frontend/cypress/tests/chartView/chartView_optimiseTimeline.cy.js
```diff
@@ -0,0 +1,148 @@
...
+      .click();
+
+    cy.get('#tab-zoom')
+      .should('be.visible');
+
+    cy.get('#tab-zoom .ramp a')
+      .first()
+      .invoke('css', 'right')
+      .then((val) => parseFloat(val))
+      .should('gt', 0);
```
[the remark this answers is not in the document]
jonasongg: just fixed, thank you!

MarcusTXK (approved):
LGTM, tested and works great. Definitely a very useful feature to have.

Regarding adding of other tests, such as to test it with the zoom feature in `chartView_zoomFeature.cy.js`, let's create an issue for these so we don't forget about it.

MarcusTXK:
@jonasongg Ah OK, apologies I missed it. No worries, this works, let's leave it as it is then.

Issue #2177, opened 2024-03-28
Migrate to Java 11 Syntax and Features


asdfghjkxd:
### What feature(s) would you like to see in RepoSense

The current language support on which our codebase is written is based on Java 8. However, we should look into migrating to Java 11 LTS for better security, features and APIs.

**EDIT:** After checking the API documentation, it appears that the `HttpServer` implementation introduced in JDK 11 is part of the `com.sun` package, which is discouraged from use due to the JRE-specific implementation of these packages, meaning that users of other JREs other than `openjdk` may not be able to use it without breaking. As such, it might not be prudent to change the existing HTTP server packages that we are using right now for the new, implementation-specific HTTP server internal package offered by the JDK.

### Is the feature request related to a problem?

Related to #1304.

### If possible, describe the solution

We could refer to these websites on the new features introduced in [Java 9](https://www.baeldung.com/new-java-9), [Java 10](https://www.baeldung.com/java-10-overview) and [Java 11](https://www.baeldung.com/java-11-new-features), and evaluate if the new features offered by Java 9, 10 and 11 can be implemented into the current codebase.

### If applicable, describe alternatives you've considered

N/A

### Additional context

This is potentially a large issue and may require multiple issues and PRs to resolve. The following are tasks that we need to complete to successfully migrate from Java 8 language support to Java 11 language support.

- [x] Update CI/CD to Java 11 [in progress at #2183, related to #2161 and #2172]
- [x] Update checkstyle to v10
- [x] Update APIs to Java 11
- [x] Update Syntax to Java 11

asdfghjkxd:
Closing this issue as we have migrated over to Java 11, and have introduced Java 11 syntax that is non-compatible with older versions of Java.

ckcherry23:
We can create a new issue for upgrading checkstyle.

Issue #1304, opened 2020-07-19
Migration to Java 11


Tejas2805:
We should start migrating our code to Java 11 because from Gradle Version 6.0 onwards, it's compatible with Java 11.

As seen recently, our Gradle version was outdated and this led to 1/2 breaking changes. If we start moving to Java 11 from beforehand,  we can be ready and update our Gradle version as and when required.

fzdy1914:
@damithc What do you think of this issue? Do you think we still need java8 compatibility?

damithc:
> Do you think we still need java8 compatibility?

Would be nice to have (I think it was the last 'free' version of Java and some people might not upgrade beyond Java 8 fora while) but OK not to have it too.

fzdy1914:
OK, will only upgrade it when we truly have to do so.

dcshzj:
Lowering priority, as OpenJDK 1.8 is [supported until at least May 2026](https://access.redhat.com/articles/1299013).

yhtMinceraft1010X:
As a side note, [Checkstyle 10](https://checkstyle.sourceforge.io/releasenotes.html) is now available but requires a minimal version of Java 11. If we decide to move fully to Java 11 and raise the minimum Java version, we can consider upgrading Checkstyle at the same time.

supermii2:
The current build of RepoSense works on Java 11 and with Checkstyle 10.10.0, so we are able to migrate if needed

damithc:
I don't mind upgrading to Java 11

ckcherry23:
Let's continue the discussion in #2177

PR #2183, opened 2024-04-01
[#2177] Migrate to Java 11 Syntax and Features

asdfghjkxd:
Part of #2177 

### Proposed commit message

```
Our current codebase is written on Java 8, which is around 2
generations behind on the latest supported LTS version for Java.

Let's move to migrate our CI/CD from Java 8 to Java 11 first to prepare
for any upcoming changes in our syntax, APIs and features.
```

### Other information

N/A


sopa301 (commented):
I think we should also update the docs, particularly the required Java version in the developer guide!

asdfghjkxd:
@sopa301 Done! I have updated the docs to reflect the migration from Java 8 to Java 11!

asdfghjkxd:
@supermii2 I agree, we should discuss the different components that can be upgraded and begin implementing the changes in this PR!

sopa301 (approved):
LGTM!

asdfghjkxd:
@supermii2 I have upgraded checkstyle to version `10.10.0` as described in #1304. Do let me know if there are any other areas that need upgrading!

supermii2:
LGTM!

supermii2 (commented):
LGTM!

build.gradle
```diff
@@ -99,7 +99,7 @@ run {
 }
 
 checkstyle {
-    toolVersion = '9.3'
+    toolVersion = '10.10.0'
```
ckcherry23, line 102: I think it would be better if we do the checkstyle upgrade in a different PR, because there may be some changes that break backward compatibility.

asdfghjkxd: I see, sure thing, I will revert back to the previous commit before the introduction of the new checkstyle tool!

ckcherry23 (approved):
LGTM!

gok99 (commented):
Really sorry for the review delays these last 2 weeks! Thanks for the thorough CI and docs updates. Looks most good to go!

Can we remove the `killDescendants` in `serveTestReportInBackground` at line 202 of `build.gradle`?

asdfghjkxd:
@gok99 I have removed the flag from the `build.gradle` file. Do let me know if there are any more changes that needs to be made!

gok99 (approved):
LGTM

ckcherry23:
Right now, the status checks are not passing because the branch protection rules require the "JDK 8" version to pass, whereas this PR changes the CI name to use "JDK 11". 

Hence, the "Branch protection rules" or "Status checks that are required" settings must be updated to reflect the new name for this PR to be merged.

chan-j-d:
@ckcherry23 I've updated the branch protection rules to require the new JDK 11 checks instead. Though the frontend tests seem to be failing somewhat randomly, based on the history of CI runs on this PR. Should probably fix that soon and merge this as it breaks the other PRs.

PR #2185, opened 2024-04-02
[#2184] Fix Inconsistent Line Number Colours

### LINKED ISSUE #2184: Unintended color change in line numbers
image

All line numbers should be the same color, right?

sopa301:
Fixes #2184.

### Proposed commit message

```
Fix Inconsistent Line Number Colours

A regression has occurred where the line numbers no longer have a
consistent grey color.

Let's fix the bug and add a test to catch further regressions of the
same nature.
```


jonasongg:
LGTM!

jonasongg (approved):
LGTM!

frontend/cypress/tests/codeView/codeView_codeHighlighting.cy.js
```diff
@@ -0,0 +1,16 @@
...
+      .should('exist')
+      .first()
+      .click();
+
+    cy.get('#tab-authorship .files', { timeout: 90000 })
+      .should('be.visible');
+
+    cy.get('.line-number')
+      .each((el) => {
+        expect(el.css('color')).to.equal('rgb(158, 158, 158)'); // #9E9E9E
```
MarcusTXK, line 13: Since the issue is regarding consistency, shall we instead ensure that all the line numbers are the same color? This way we could avoid hard coding of the colors

frontend/cypress/tests/codeView/codeView_codeHighlighting.cy.js
```diff
@@ -0,0 +1,19 @@
...
+    cy.get('#tab-authorship .files', { timeout: 90000 })
+      .should('be.visible');
+
+    cy.get('.line-number')
+      .first()
+      .should('have.css', 'color')
+      .then((firstColor) => {
+        cy.get('.line-number')
+          .each((el) => expect(el.css('color')).to.equal(firstColor));
+      });
```
ckcherry23, line 17: This test needs to be updated as it would pass even without the bug fix. The `line-number` CSS color here is still always grey (default line number colour) but the colour shown on the report is the CSS color of`hljs-comment` (dark grey) or `hljs-number` (blue). 

Please make sure the test is changed such that it fails before the fix, and passes after it. Thanks!

ckcherry23 (changes requested):

frontend/cypress/tests/codeView/codeView_codeHighlighting.cy.js
sopa301: My bad, I'll remember to do this for future tests!

asdfghjkxd (approved):
LGTM!

ckcherry23 (approved):
Thanks for fixing this @sopa301! 

Seems like this test case takes 2 seconds to run in the CLI which is reasonable. However, I had longer wait times (~12 seconds) when running using the automated browser for some reason. The test case also caused crashes of the Cypress app sometimes, possibly due to repeatedly getting the child elements. If this causes more issues in the future, perhaps we can consider limiting how many line numbers are tested instead of checking all of them.

This LGTM for now though!

vvidday (approved):
LGTM

PR #2188, opened 2024-04-08
Add more repos to cypress tests

### LINKED ISSUE #2187: Add more repos for Cypress frontend tests

It can be useful to have more repos for Cypress tests.

This is related to the bug that necessitated #2165. If we had more repos in the tests, we could have caught that behaviour.

jonasongg:
Fixes #2187

### Proposed commit message

```
Add more repos to Cypress tests

Currently, our Cypress config only tests with a single repo. This can
cause our testing to be limited and cause certain bugs to go
unnoticed.

Let's add more repos so that our testing can be more extensive.
```


ckcherry23 (approved):
LGTM! We can add test cases for the bug in #2164 in a later PR.

vvidday (approved):
Great initiative, thanks! Will definitely help us catch more bugs in the future.

sopa301 (approved):
LGTM!

PR #2197, opened 2024-04-30
[#2195] Add cypress test for show tag when grouping by author and by none

### LINKED ISSUE #2195: Add Cypress tests for show tags on ramp chart
#2163 is the PR to add Cypress tests for

jonasongg:
Fixes #2195 

### Proposed commit message

```
Add cypress test for show tag when grouping by author and by none

Currently, there are no Cypress tests to test the behaviour of the show
tag feature when the report is grouped by author or grouped by none.

Let's add a test for this to ensure the feature is working properly.
```

### DIFF
--- a/frontend/cypress/config/author-config.csv
+++ b/frontend/cypress/config/author-config.csv
@@ -1,0 +1,2 @@
+Repository's Location,Branch,Author's Git Host ID,Author's Emails,Author's Display Name,Author's Git Author Name,Ignore Glob List
+https://github.com/reposense/publish-RepoSense.git,master,yong24s,,Yong Hao,,
--- a/frontend/cypress/config/author-config.csv
+++ b/frontend/cypress/config/author-config.csv
@@ -1,2 +1,2 @@
 Repository's Location,Branch,Author's Git Host ID,Author's Emails,Author's Display Name,Author's Git Author Name,Ignore Glob List
-https://github.com/reposense/publish-RepoSense.git,master,yong24s,,Yong Hao,,
+https://github.com/reposense/publish-RepoSense.git,master,yong24s,,,Yong Hao TENG;yong24s,
--- a/frontend/cypress/tests/chartView/chartView_showTags.cy.js
+++ b/frontend/cypress/tests/chartView/chartView_showTags.cy.js
@@ -77,10 +77,10 @@ describe('show tags', () => {
       .check()
       .should('be.checked');
 
-    cy.get('.summary-chart:first summary-chart__title--tags')
+    cy.get('.summary-chart:first .summary-chart__title--tags')
       .should('not.exist');
 
-    cy.get('.summary-chart:nth-child(2) summary-chart__title--tags')
-      .should('not.exist');
+    cy.get('.summary-chart:nth-child(2) .summary-chart__title--tags')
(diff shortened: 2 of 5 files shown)


frontend/cypress/tests/chartView/chartView_showTags.cy.js
```diff
@@ -64,4 +64,23 @@ describe('show tags', () => {
...
+    cy.get('#summary label.show-tags > input:visible')
+      .should('be.visible')
+      .check()
+      .should('be.checked');
+
+    cy.get('.summary-chart:first summary-chart__title--tags')
+      .should('not.exist');
+
+    cy.get('.summary-chart:nth-child(2) summary-chart__title--tags')
+      .should('not.exist');
```
ckcherry23, line 84: Based on what I see in the testing data, shouldn't the second chart have tags? This seems to check if they don't exist instead.

frontend/cypress/config/author-config.csv
```diff
@@ -0,0 +1,2 @@
+Repository's Location,Branch,Author's Git Host ID,Author's Emails,Author's Display Name,Author's Git Author Name,Ignore Glob List
+https://github.com/reposense/publish-RepoSense.git,master,yong24s,,Yong Hao,,
```
ckcherry23, line 2: Seems like this author config is not accurately recognizing Yong Hao's commits in `publish-RepoSense`, as it says 0 lines contributed when he has 148 lines. Perhaps look at this RepoSense documentation to specify multiple Git author names for one user.

ckcherry23 (commented):
Thanks for adding the `author-config.csv` to test this. I've left a few comments about the tests.

Also, I just realised that the tags feature does not work as expected when using the `Group by None` option. Each author gets all of the tags of the repo instead of only the ones they tagged. Let's create a new issue for this.

Screenshot 2024-05-01 at 2 03 52 PM

frontend/cypress/tests/chartView/chartView_showTags.cy.js
jonasongg: fixed!

jonasongg:
@ckcherry23 the bug with group by none actually happened because of a change I made in this PR to fix group by author. should i discard this change and create a new issue + PR to address the original bug?

frontend/cypress/config/author-config.csv
jonasongg: fixed!

ckcherry23:
> @ckcherry23 the bug with group by none actually happened because of a change I made in this PR to fix group by author. i fixed the bug in this PR already, so all group by options should be working correctly. (also added some tests for this!) do i need to create a new issue + PR to address the original bug?

If everything works as expected now, we do not need a new issue + PR. The current PR seems to have fixed all the issues, is there anything missing?

ckcherry23 (approved):
LGTM @jonasongg! Thanks for making the test cases more robust by comparing chart tags with those on the Zoom tab!

Please update the PR title and description to reflect that both test cases were added ('group by author' and 'group by none').

jonasongg:
@ckcherry23 done, thanks!

vvidday (approved):
LGTM, thanks

PR #2178, opened 2024-03-29
[#2176] Move from Vue CLI to Vite

sopa301:
Fixes #2176

### Proposed commit message

```
Move from Vue CLI to Vite

Vue CLI has been put on maintenance mode and its dependencies have
security issues that have not been fixed for a while. Switching to Vite
also allows us to introduce Vitest for our unit testing.

Let's install Vite and remove Vue CLI.
```

### Other information

A few things worthy of discussion:
- The [Vite migration guide](https://vueschool.io/articles/vuejs-tutorials/how-to-migrate-from-vue-cli-to-vite/) suggests removing Babel (which supports older browsers), which has been done so. If we decide to retain support for them, we can do so with a [plugin](https://github.com/vitejs/vite/tree/main/packages/plugin-legacy).
- The Vite changes remove the ability to configure the title and source folder for the `html` file from the config file. However, now that the `html` file has been moved out of the `public` folder (by design of Vite), this may not be a big concern. We can use a `.env` file if we want to be able to edit the title and root folder from another file.
- The Cypress tests on CI did not work initially and had to be hotfixed by adding `"chromeWebSecurity": false` into the Cypress config file. Since this is not very extensible to other browsers, we should also find another way to get the tests to work.

PR #2203, opened 2024-05-14
Revert to Vue CLI to fix GitHub Pages Deployment

### LINKED ISSUE #2202: Generated report is blank
Example https://nus-tic2002-ay2324s2.github.io/ip-dashboard/

ckcherry23:
Reverts reposense/RepoSense#2178 to fix #2202 

### Proposed commit message
```
Revert to Vue CLI to fix GitHub Pages Deployment

Although Vite boasts better performance and is maintained actively,
its usage for GitHub Pages deployment requires specifying the base of the
deployment URL in the config. This makes it inconvenient as all existing
deployment dashboards must update the base respective to their URLs.  

Let's revert to Vue CLI temporarily to fix this.
```

### Other information
[Deploying sites built with Vite on GitHub Pages](https://vitejs.dev/guide/static-deploy.html#github-pages) requires configuring the `base` in `vite.config.js` as the subdirectory (or repository name) where the report is deployed (eg. `/ip-dashboard/`). However, since RepoSense reports will be deployed on multiple GitHub Pages URLs, this will require all users to define a new environment variable to specify the custom `base`, which is not ideal. 

We need to decide if we should just revert the migration PR or introduce the env variable into the Vite config. For now, I have reverted the PR. Feel free to suggest alternatives.


sopa301 (approved):
LGTM, I'll look into a fix!

vvidday (approved):

PR #2219, opened 2024-05-16
Fix missing PageNav in docs

damithc:
Fixes a typo that caused the page navigation menu to not appear:

before:
image

after:
image


ckcherry23 (approved):
Thanks prof!

PR #2232, opened 2024-07-03
[#2196] Use git blame line info for aggregate blame author modified and date info

### LINKED ISSUE #2196: Use `GitBlameLineInfo` for processing git blame output in `aggregateBlameAuthorModifiedAndDateInfo`

https://github.com/reposense/RepoSense/blob/bf78bf2a11a4ba703403c75f442410ad56ecca98/src/main/java/reposense/authorship/FileInfoAnalyzer.java#L158-L183

`aggregateBlameAuthorModifiedAndDateInfo` in `FileInfoAnalyzer.java` does some manual string wrangling of git blame output which is hard to interpret, contains magic numbers and breaks the git blame abstraction.

#2140 Introduces `GitBlameLineInfo`, returned by `blameLine`, which encapsulates this string wrangling logic. We can replace this manual string processing by using this method in `aggregateBlameAuthorModifiedAndDateInfo`.

logical-1985516:
Fixes #2196 

### Proposed commit message

```
Uses blameLine to abstract away the string wrangling and improve
understandability. Magic numbers are also replaced to improve code
quality.

In order to use blameLine however, it is also changed to use 
author-time instead of commit-time, as there is a discrepancy
between the two, causing some test cases to fail.

The naming of the timestamp field in GitBlameLineInfo is changed
to reflect that it is in seconds, as author-time is in seconds.
```

### Other information

Surrounding code which states milliseconds is also changed to state seconds correctly, as this PR is meant to resolve a code quality issue anyways. If this is not ok, I can revert the last commit.


logical-1985516:
Hi, I have noticed that some test cases take a very long time to finish. Could this due to the additional overhead incurred from running git blame command in blameLine?

logical-1985516:
Hi, would anyone like to approve the workflow? I am not sure why the local test cases run indefinitely, so I am unable to verify that it passes all test cases. Thank you!

damithc:
> Hi, would anyone like to approve the workflow? I am not sure why the local test cases run indefinitely, so I am unable to verify that it passes all test cases. Thank you!

done

logical-1985516:
Hi, would anyone like to help me for this issue? I could not seem to figure why the test cases are failing. More information at this [discussion](https://github.com/reposense/RepoSense/discussions/2235).

src/main/java/reposense/git/model/GitBlameLineInfo.java
```diff
@@ -7,13 +7,13 @@ public class GitBlameLineInfo {
     private final String commitHash;
     private final String authorName;
     private final String authorEmail;
-    private final long timestampMilliseconds;
+    private final long timestampInSeconds;
```
gok99, line 10: Good catch!

src/main/java/reposense/git/GitBlame.java
```diff
@@ -81,7 +81,7 @@ private static GitBlameLineInfo processGitBlameResultLine(String blameResult) {
         String commitHash = blameResultLines[0].substring(0, FULL_COMMIT_HASH_LENGTH);
         String authorName = blameResultLines[1].substring(AUTHOR_NAME_OFFSET);
         String authorEmail = blameResultLines[2].substring(AUTHOR_EMAIL_OFFSET).replaceAll("[<>]", "");
-        long timestampMilliseconds = Long.parseLong(blameResultLines[5].substring(COMMIT_TIME_OFFSET));
+        long timestampMilliseconds = Long.parseLong(blameResultLines[3].substring(AUTHOR_TIME_OFFSET));
```
gok99, line 84: It seems like `author-time` is semantically what a user would expect from the diff. We can offer options of `AUTHOR_TIME_OFFSET` or `COMITTER_TIME_OFFSET` if necessary, but probably not worth it. Let's look at the tests this affects and see if we can update them.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
```diff
@@ -161,17 +158,15 @@ private void aggregateBlameAuthorModifiedAndDateInfo(RepoConfiguration config, F
...
-            String authorName = blameResultLines[lineCount + 1].substring(AUTHOR_NAME_OFFSET);
-            String authorEmail = blameResultLines[lineCount + 2]
-                    .substring(AUTHOR_EMAIL_OFFSET).replaceAll("<|>", "");
-            long commitDateInMs = Long.parseLong(blameResultLines[lineCount + 3].substring(AUTHOR_TIME_OFFSET)) * 1000;
-            LocalDateTime commitDate = LocalDateTime.ofInstant(Instant.ofEpochMilli(commitDateInMs),
-                    config.getZoneId());
-            Author author = config.getAuthor(authorName, authorEmail);
-
-            int lineNumber = lineCount / 5;
+        for (int lineCount = 0; lineCount < blameResultLines.length; lineCount += BLAME_LINE_INFO_ROW_COUNT) {
```
gok99, line 161: I was originally intending for most of this logic to be encapsulated in GitBlame, largely because we don't want to run GitBlame for every line. So, ideally something like `blameFile` in `GitBlame` would output a list of `GitBlameLineInfo`. We might have to augment `GitBlameLineInfo` with additional information like line numbers.

gok99 (changes requested):
Really sorry for the review delays! I've left some clarifications about the intent of the original issue. Let me know if you've got questions.

src/main/java/reposense/git/GitBlame.java
logical-1985516: Hi, after some investigation, this indeed caused one of the test cases to fail, AuthorAnalyzerTest > analyzeAuthorship_beforeSinceDate_success(). I have added an option to use either author time or committer time in this method.

src/main/java/reposense/authorship/FileInfoAnalyzer.java
logical-1985516: Hi, I have followed what you said and created a new method, blameFile, that converts the blameResult into a list of GitBlameLineInfo. The reason why I used blameLine initially, was because of its mention in issue, which got me too focused on using it. Either way, thanks for your input!

src/main/java/reposense/git/GitBlame.java
```diff
@@ -57,6 +63,23 @@ public static String blameWithPreviousAuthors(String root, String fileDirectory)
...
+    /**
+     * Returns the processed git blame result, created from the raw git blame result.
+     */
+    public static List<GitBlameLineInfo> blameFile(String blameResults) {
+        String[] blameResultLines = StringsUtil.NEWLINE.split(blameResults);
+        List<GitBlameLineInfo> blameFileResult = new ArrayList<>();
+        for (int lineCount = 0; lineCount < blameResultLines.length; lineCount += BLAME_LINE_INFO_ROW_COUNT) {
+            String blameResultLine = Arrays.stream(Arrays
+                    .copyOfRange(blameResultLines, lineCount, lineCount + BLAME_LINE_INFO_ROW_COUNT - 1))
+                    .reduce("", (curr, next) -> curr + next + "\n");
```
logical-1985516, line 75: I am thinking of moving StringsUtil.NEWLINE.split(blameResult) to blameLine, so that this reduce function is no longer needed. What do you think?

gok99: Sure, go ahead!

src/main/java/reposense/git/GitBlame.java
```diff
@@ -57,6 +63,23 @@ public static String blameWithPreviousAuthors(String root, String fileDirectory)
         return StringsUtil.filterText(runCommand(rootPath, blameCommandWithFindingPreviousAuthors), COMBINATION_REGEX);
     }
 
+    /**
+     * Returns the processed git blame result, created from the raw git blame result.
+     */
+    public static List<GitBlameLineInfo> blameFile(String blameResults) {
```
gok99, line 69: Could we have the interface to `blameFile` be similar to that of `blameLine` where the blame is done internally? We could have `blameFile` require `root`,  `fileDirectory` and a flag for `withPreviousAuthors` for now. We could then probably get rid of `getGitBlameResult` in `FileInfoAnalzyer`.

gok99 (changes requested):

src/main/java/reposense/authorship/FileInfoAnalyzer.java
gok99: I should have been clearer in the issue, sorry for the confusion!

src/main/java/reposense/git/GitBlame.java
```diff
@@ -68,21 +85,41 @@ public static GitBlameLineInfo blameLine(String root, String commitHash, String
...
+        return blameFileResult;
     }
 
     /**
-     * Returns the processed result of {@code blameResult}.
+     * Returns the processed result of {@code blameResultLines}, with reference to {@code timeOption}.
      */
-    private static GitBlameLineInfo processGitBlameResultLine(String blameResult) {
-        String[] blameResultLines = StringsUtil.NEWLINE.split(blameResult);
+    private static GitBlameLineInfo processGitBlameResultLine(String[] blameResultLines, String timeOption) {
```
gok99, line 110: I'd prefer to use a boolean for timeOption since it doesn't seem like there are likely to be new options in git in the future.

src/main/java/reposense/git/GitBlame.java
```diff
@@ -5,6 +5,9 @@
```
gok99, line 5: Could you add / update tests for functions / flags you've added here?

gok99 (changes requested):
Thanks for the updates! Should be good to go after the final few comments are addressed.

gok99:
@logical-1985516 Are you planning to wrap up this PR and #2238? I think this PR is mostly good to merge and is just pending some updates to tests.

logical-1985516:
Hi @gok99, for this PR, I may be able to wrap up this PR within next week, if I have the time.
However, I probably would not wrap up PR #2238. Sorry about that.

gok99:
@logical-1985516 thanks for the fixes. made a minor name change.

gok99 (approved):

jonasongg (approved):
LGTM!

georgetayqy (approved):
LGTM!

Though, we should look into upgrading the macOS runners since macOS 12 runners are being deprecated and GitHub Actions is not running the tests for macOS 12.

PR #2238, opened 2024-07-18
2100 widget info

### LINKED ISSUE #2100: Add date range and footer to widget
Add more relevant information to the widget such as:
* Date range of the chart
* Attribution to RepoSense in the Footer

Currently, this is how the widget UI looks. We can add more relevant information to make the ramp/bar charts more meaningful and link back to the RepoSense website.

Screenshot 2024-01-29 at 8 01 19 PM

### Additional notes
This issue may involve some UI design as well.
For the footer, we can refer to the current footer of the full RepoSense web report for consistency in design.

logical-1985516:
Fixes #2100 

### Proposed commit message

```
The ramp widget UI is currently quite empty and more information can be 
added to be more useful.

Let's add date range and footer to the ramp widget so that it is easier 
for people to view the timelime of the commits and the RepoSense 
website.
```

### Other information

This is how the new UI looks:
image


logical-1985516:
How should I go about writing test cases for the widget?

logical-1985516:
Hi @ckcherry23, does this new UI look good to you? Is there any other useful information that could be added to the widget?

ckcherry23 (changes requested):
Hi @logical-1985516, thank you for looking into this! Not sure if you were able to explore the difference between the main report view and the widget mode, but you can read more about the widget mode here [[Widget view](https://reposense.org/dg/report.html#widget-view-c-widget-vue), [Embeddable widgets](https://reposense.org/ug/sharingReports.html#embeddable-widgets)].

I don't think we should show the date indicators and "Generated by RepoSense" footers for all ramp charts as seen below in the Report view. This is because the since/until dates would be the same for all, and these dates are mentioned at the top of the report, so it is not that meaningful.

Report view:
Screenshot 2024-08-03 at 7 43 37 AM

Widget mode:
Screenshot 2024-01-29 at 8 01 19 PM

Hence, the date range must be visible conditionally, only when widget mode is `true`. Another option for this is to trim/optimise timelines by default for widgets so that the date range is shown. This would also make the widget ramps easier to differentiate/interpret, so I would prefer this.

For now, I think we can remove the "Generated by RepoSense" attribution footer from the centre of the widget as it seems to be too distracting. We can explore adding a small RepoSense logo at the top/bottom right with a link to the RepoSense website.

ckcherry23:
> Should the date indicators still be displayed if there are no commits? Currently, trim timeline does not show the date indicators if there are no commits found.

They don't need to be displayed.

> Some frontend test cases that pass on CI are failing locally - how do I fix this? Furthermore, it takes about 40 minutes to finish running the test cases. One example is:
> 
> ```
>   1) include merge commits in chart view
>        show merge commits in summary chart:
>      AssertionError: Timed out retrying after 30000ms: Expected to find element: `[title="[2023-03-03] Merge branch 'new-branch' into cypress: +0 -0 lines "]`, but never found it.     
>       at Context.eval (webpack:///./tests/chartView/chartView_mergeCommits.cy.js:5:7)
> ```

The test case you mentioned is a flaky test known to have issues locally on some devices. Might be issues related to author vs commit date. This has always passed on CI, so we can ignore it. Do let us know if there are any other test cases that are failing.

PR #2241, opened 2024-07-28
[#1832] Move report generation timestamp

### LINKED ISSUE #1832: Repeat footer on both panels
It's a bit odd that this footer appears on one panel only.
image

Should we repeat it on both sides?

Other suggestions welcome.

jedkohjk:
Resolves #1832

### Proposed commit message

```
Shift report generation timestamp up

Removes footer so that there is no inconsistency regarding 
the presence/absence of a footer between panels.

Rephrases message so it does not look like it is part of title.
```

### Other information

Before:

before-top

before-bottom

After:

after-top

after-bottom


damithc:
Thanks for the PR @jedkohjk 
hmm... I don't think the timestamp/version etc. fits in the top/middle of the report either, as it is just a 'small print' that doesn't deserve space in high-value screen real estate.

jedkohjk:
Thanks for the review @damithc 

Hmm, I agree that the version and user guide would fit a duplicated footer, as suggested in the original issue. But I felt like the timestamp may not feel like it fits on the right panel as the timestamp is for the entire report. I'm not sure if that makes sense.

As of now, I don't really have any other suggestions. Perhaps we could shift the reposense version and user guide back to the footer and duplicate it, and leave the time stamp on top?

Something like this (only for the left panel, the bottom of the right panel should have the reposense link in what I'm proposing, but I made this with microsoft paint):

Screenshot

Without the blue hyperlinks, the timestamp is less attention grabbing.

ckcherry23:
CS2103/T students may prefer finding the timestamp more easily on the top/middle as it can help them figure out why certain commits have not been updated yet.

However, a timestamp at the top may not be ideal for our other use cases, such as creating a project code portfolio.

jedkohjk:
Hi, I made a separate pull request, #2243 as another alternative. There, I left the timestamp as a footer on the left panel, and duplicated only the reposense links.

jedkohjk:
Hi, should I close this PR as CP3108A is coming to an end for me?

PR #2242, opened 2024-07-28
[#2240] Format datetime

### LINKED ISSUE #2240: Update report generation datetime format

Currently, the report generation datetime is shown as `E MMM dd HH:mm:ss yyyy z`.

Screenshot

Separating the year from the rest of the date is unintuitive. A format like `E dd MMM yyyy HH:mm:ss z` may be more intuitive.

Change the datetime formatter to use the proposed format.

jedkohjk:
Resolves #2240

### Proposed commit message

```
Format datetime as "day date month year time timezone"

Before, it was "day month date time year timezone".

"day date month year time timezone" puts the year with the rest of 
the date, and puts the time with the timezone, which is more 
intuitive.

Using "date month year" instead of putting the month before the date 
keeps the date format consistent with the since and until dates, and 
is also the norm in most of Singapore.
```

### Other information

Before:

Before

After:

After


jonasongg (approved):
LGTM!

ckcherry23 (approved):
LGTM!

PR #2272, opened 2025-01-31
[#2266] Fix Cypress tests that fail locally

### LINKED ISSUE #2266: Fix Cypress tests that fail locally but pass on GitHub due to time zone difference
When running Cypress tests on a local machine (`./gradlew frontendTest` or otherwise), certain tests such as in `optimiseTimeline` will fail as the time zone of Singapore is different from that of the GitHub runner, and the tests are currently configured to pass on GitHub. This causes confusion to new contributors when running Cypress tests. Let's change this behaviour such that tests will pass on both local machines and on GitHub.

There could be some Cypress configs that allow changing the time zone the tests are run in. Otherwise, some other workaround to ensure the tests that fail purely because of this time zone issue pass on local machines will work as well.

Airiinnn:
Fixes #2266

### Proposed commit message

```
Set UTC timezone for test reports

Running Cypress tests locally uses system's default timezone
(UTC+8), causing 3 tests that check for dates to fail. These 3 tests
pass on GitHub Actions as the system's default timezone is UTC.

Let's standardize the timezone for all cypress tests to UTC.
```

### Other information

Run `./gradlew frontendTests` or `./gradlew cypress` locally to verify all tests are passing.

3 previously failing tests:
* chartView/chartView_mergeCommits.cy.js show merge commits in summary chart
* chartView/chartView_optimiseTimeline.cy.js start and end date indicators should exist
* chartView/chartView_optimiseTimeline.cy.js zoom panel range should work correctly when timeline is optimised

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -197,7 +197,7 @@ def serveTestReportInBackground = tasks.register('serveTestReportInBackground', 
     workingDir = 'build/serveTestReport'
     main = mainClassName
     classpath = sourceSets.main.runtimeClasspath
-    args = ['--config', './exampleconfig', '--since', 'd1', '--view', '-A']
+    args = ['--config', './exampleconfig', '--since', 'd1', '--view', '-A', '-t', 'UTC']
     String versionJvmArgs = '-Dversion=' + getRepoSenseVersion()
     jvmArgs = [ versionJvmArgs ]
     waitForPort = 9000
--- a/frontend/cypress/tests/chartView/chartView_optimiseTimeline.cy.js
+++ b/frontend/cypress/tests/chartView/chartView_optimiseTimeline.cy.js
@@ -57,7 +57,6 @@ describe('optimise timeline', () => {
       .find('.summary-chart__ramp .date-indicators span')
       .last()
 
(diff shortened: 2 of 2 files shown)


build.gradle
```diff
@@ -197,7 +197,7 @@ def serveTestReportInBackground = tasks.register('serveTestReportInBackground',
     workingDir = 'build/serveTestReport'
     main = mainClassName
     classpath = sourceSets.main.runtimeClasspath
-    args = ['--config', './exampleconfig', '--since', 'd1', '--view', '-A']
+    args = ['--config', './exampleconfig', '--since', 'd1', '--view', '-A', '-t', 'UTC']
```
JoanneHing, line 200: Good fix. All CI passed too. LGTM.

JoanneHing (approved):

jonasongg (changes requested):
Looks good! could you remove the comment in `chartView_optimiseTimeline.cy.js` (line 59-60) which is related to this issue and is no longer needed? thank you and great fix!

jonasongg (approved):
LGTM! thanks for the fix :)

PR #2287, opened 2025-02-10
[#2190] Add ESLint rule to enforce commas between types

### LINKED ISSUE #2190: Add a typescript linting rule to enforce commas between types
Please include the steps to reproduce the bug.

TypeScript does not require a separator between types. However, for style and consistency, it should be enforced that there is always a comma separator between types

What was expected to happen?

An error should be thrown if the separator is missing. 

Let's fix these using an eslint rule.

https://eslint.style/rules/ts/member-delimiter-style

joeng03:
Fixes #2190

```
Add ESLint rule to enforce commas between types

This PR overrides the default behavior of requiring semicolon
separators for both interface and type in ESLint, to require semicolon
separator for interface and comma separator for type
```

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -82,7 +82,23 @@
             ]
           }
         ],
-        "@stylistic/member-delimiter-style": "error",
+        "@stylistic/member-delimiter-style": [
+          "error",
+          {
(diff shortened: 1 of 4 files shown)


sopa301 (approved):
LGTM

PR #2303, opened 2025-02-17
[#2283] Display time periods for each repository

### LINKED ISSUE #2283: Display time periods for each repository

Display time periods for individual projects to support allowing different periods for different repos.

Since `sinceDate` and `untilDate` will be provided in `summary.json` for each repository, display those dates under the ramp, like how it is done when trim timeline option is enabled. The date indicators will now always be visible.

Airiinnn:
Fixes #2283 

### Proposed commit message

```
Add frontend support for specifying individual repository time periods

Previously, there is no way to indicate the different time periods for
each repository.

Let's display date indicators to show the time periods for each
repository.
```

### Other information

**Changes:**

* Date indicators are always visible now.
* Date indicators should display the tighter date range (between the dates provided in `summary.json` and filter range). Refer to last summary chart in screenshot to see intended behavior.

**Old:**

image

Previously, date indicators are only visible when "trim timeline" is enabled.

image

**New:**

image


Airiinnn:
On hold. Waiting for backend changes to be merged to add the cypress tests

damithc:
@Airiinnn For UI changes, also give before-and-after screenshots in the PR.

frontend/src/components/c-summary-charts.vue
```diff
@@ -785,7 +798,7 @@ export default defineComponent({
       const regexToRemoveWidget = /([?&])((chartIndex|chartGroupIndex)=\d+)/g;
       return url.replace(regexToRemoveWidget, '');
     },
-    getRepo(repo: Array<Repo>): Array<Repo> {
+    getRepo(repo: Array<User>): Array<User> {
```
CYX22222003, line 801: Is there any particular reason of changing the type of `getRepo`?

CYX22222003, line 801: Is there any particular reason to change the typing here?

Airiinnn: Yes, following the call stack, the previous `Repo` type was wrong and should be `User` instead. 

Naming was very confusing to begin with, maybe we can fix this in another PR for refactoring and cleanup. I'm worried changing parts of it now may make it even more confusing.

CYX22222003: I see, maybe we should create another PR to handle this issue.

jonasongg (changes requested):
good work so far! can i ask - if let's say we have 4 dates A B C D, and `summary.json` specifies A to C, and i filter B to D, what's the outcome?

what if `summary.json` specifies A to B, and i filter C to D?

Airiinnn:
> good work so far! can i ask - if let's say we have 4 dates A B C D, and `summary.json` specifies A to C, and i filter B to D, what's the outcome?
> 
> what if `summary.json` specifies A to B, and i filter C to D?

Thanks for catching the 2nd case, I didn't consider that. Sorry for the oversight, will fix and add the tests for more edge cases

Here's what's I intend, do lmk if there's smth btr:

Let's say
A: 2020-01-01
B: 2021-01-01
C: 2022-01-01
D: 2023-01-01

If summary.json specifies A - D and you filter B - C, it should show B - C (the tightest range possible). Currently working.

If summary.json specifies A - C and you filter B - D, it should show B - C (the tightest range possible). Currently working.

If summary.json specifies A - B and you filter C - D, it should show B - B. Currently unaccounted.

If summary.json specifies C - D and you filter A - B, it should show C - C. Currently unaccounted.

jonasongg:
@Airiinnn I think for 3rd and 4th cases, it would be better to show something like a dash to indicate some form of invalidity. because if you show B - B, users might think that it's displaying commits on the date of B itself!

sopa301:
I kinda like number 3. I think it's really unusual for a user to deliberately put invalid date ranges, so we can be fairly certain it's an unintended error, and using red text would make it easier to identify.

Prof @damithc what are your thoughts on this?

jonasongg:
agreed, i think number 3 looks best!

Airiinnn:
As discussed on Monday, custom repository time periods will only be supported in portfolio view mode, due to its intended use case.

Here will be the intended behavior:

**Default view:** Disallow / ignore custom repository time periods (#2334). Frontend follows existing behavior as custom repository time periods are not supported in this view.

**Portfolio view:** Displayed summary charts to be unaffected by filter dates, solely relying on the date ranges in `summary.json`. No longer need to consider interaction with filter dates. Trim timeline will still be allowed.

I'll make these FE changes in #2320 and close this PR as the changes are interdependent.

---

For future reference, in case supporting custom repository time periods in default view becomes needed (non-exhaustive):

1. Decide how custom repository time periods should interact with the filter dates. 
a. Intersection of the 2 date ranges likely requires changes to adjust selectable filter dates to make sure all custom repository date ranges are included.
b. Ignore filter dates and only display the custom repository date range if provided. Backend needs to add a field to indicate if custom repository date range was provided.
c. Union and filter date range priority does not seem to make sense as commits may seem missing.
2. Consider what should be displayed when grouped by author

PR #2314, opened 2025-03-03
[#1941] Remove test mode flag

### LINKED ISSUE #1941: Removal of test mode flag in reposense code
With PR #1911 that removes the only use of the test mode flag in reposense, I think we can consider removing it to avoid additional maintenance. The original test mode flag was introduced in this PR #1816 where it was necessary since it was not known that the system tests were not quite functioning properly. 

The test mode flag is primarily used for implementing some test specific behavior in a reposense run which is currently not required. As such, I would like to put forward the suggestion that we remove related code and documentation for the test mode flag.

lyuanww:
Fixes #1941 

### Proposed commit message

```
Remove test mode flag

Currently, the test mode flag is not utilised in our codebase. The test
mode flag was added to pass system tests due to legacy reasons and the
related logic was removed in a previous PR.

Let's remove the related code and documentation for the test mode flag
to avoid additional maintenance.
```


CYX22222003 (approved):
LGTM! Thanks for the PR

joeng03 (approved):
LGTM!

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -279,12 +272,7 @@ private static ArgumentParser getArgumentParser() {
...
-        argumentGroup.addArgument(TEST_MODE_FLAG)
-                .dest(TEST_MODE_FLAG[0])
-                .action(Arguments.storeTrue())
-                .help("Enables testing mode.");
-
-        argumentGroup.addArgument(FRESH_CLONING_FLAG)
+        parser.addArgument(FRESH_CLONING_FLAG)
                 .dest(FRESH_CLONING_FLAG[0])
                 .action(Arguments.storeTrue())
                 .help("Enables fresh cloning. Requires testing mode to be enabled.");
```
sopa301, line 278:
```suggestion
                .help("Enables fresh cloning.");
```

Now we don't have testing mode, I think we can remove this part.

Issue #2106, opened 2024-01-30
Give a way to compare a subset of files across repos


damithc:
In some usage scenarios (e.g., CS2103), it might be useful to compare a file across repositories. That is, the user specifies a glob (e.g., `**/TaskList.java`) and the code panel shows the code changes to all files matching the glob, across all repositories. This feature can be useful for students to see how others have implemented something, without needing to click through many repos. It might help to catch cases of plagiarism as well (i.e., easy for a viewer to spot two authors implementing something in the exact same 'unusual' way).

One challenge would be to integrate this (rather niche) feature into the current feature set in a general way (i.e., it should not look like 'made for CS2103' feature).

PR #2320, opened 2025-03-10
[#2305] FE: Support optimised portfolio view

### LINKED ISSUE #2305: FE: Support optimised portfolio view

`summary.json` should include the type of view to render on the frontend (Details TBC in #2304). Based on this field, the frontend should decide whether to render the optimised portfolio view or the regular view.

Portfolio view should exclude certain components as mentioned in #2215.

On first render, display the right panel with the first repository's contributions.

Part of #2215

1. Logic to decide the view to render based on the value of the field in `summary.json`.
2. Optimised portfolio view needs to be implemented.

### Additional context

May be beneficial to refactor `c-summary.vue` and `c-summary-charts.vue` into smaller components.

To be discussed.

Airiinnn:
Fixes #2305 

### Proposed commit message

```
Brief description of current behaviour and reason(s) for the change.

Call-to-action statement describing what you are doing in this PR.
```


Airiinnn:
Tests for portfolio view needed the `--portfolio` flag enabled (or `isPortfolio: true` in `summary.json`), hence the changes to the gradle tasks and vite config. Unfortunately, this means report is generated twice now (once without `--portfolio` flag, once with) and takes slightly longer to setup.

I considered having separate tasks for each view (`cypressDefault`, `cypressPortfolio`, `frontendTestDefault`, `frontendTestPortfolio`) but I think having all the tests available like how it is currently is more convenient.

If anyone has any opinions, please feel free to suggest!

Airiinnn:
On hold. Likely need to integrate with #2283.

docs/dg/devOpsGuide.md
```diff
@@ -37,7 +37,7 @@ Some of the jobs execute some commands that are too complicated to be included i
 
 This workflow is run for both incoming pull requests to any branch as well as direct commits to any branch in the repository.
 
-Cypress frontend tests are run against reports generated from config files in `frontend/cypress/config`. It uses the `cypress` branch of the RepoSense repository which is kept independent of `master` and should be updated only when there are new frontend tests that need to be accommodated.
+Cypress frontend tests are run against reports generated from config files in `frontend/cypress/configs`. It uses the `cypress` branch of the RepoSense repository which is kept independent of `master` and should be updated only when there are new frontend tests that need to be accommodated.
```
sopa301, line 40: Perhaps we could rewrite this part while we're modifying this line. We don't use a separate branch for our frontend tests anymore.

sopa301, line 40: Perhaps we could rewrite this part while we're modifying this line. We don't use a separate branch for our frontend tests anymore.

sopa301 (approved):
LGTM, with some minor comments. We can leave the refactoring for another issue since this feature's quite high priority. Notably, the error message boxes, functions and styles are duplicated when they logically follow the non-portfolio version. Could I trouble you to create the issue for that, thanks!

Since this is a big feature with significant changes, let's have one more reviewer to ensure things are all good

jonasongg (changes requested):
i think the logic between `c-summary-charts` and `c-ramp` is getting a bit complicated! i think we should either:
1. pass in `this.isPortfolio` and `this.optimiseTimeline` from `c-summary-charts` to `c-ramp`, and calculate `sdate` and `udate` accordingly inside `c-ramp`
2. or calculate all the above in `c-summary-charts`

either way, let's push all the calculating logic inside functions called `getIndicatedMinimumDate` and `getIndicatedMaximumDate` (because they'll handle not just optimise timeline anymore), and we can write it something like:
```
if (this.optimiseTimeline)
  return user.commits.reduce...
if (this.isPortfolio && user.sinceDate)
  return user.sinceDate
return this.filterSinceDate
```
and to decide whether to display the date indicators, we should have a computed property `isDisplayDateIndicators` that returns `this.optimiseTimeline || this.isPortfolio`

this would also make sure you handle the case where isPortfolio is true, but the repo-specific since/until date is not specified

let me know if you've any questions!

jonasongg:
also if i'm not wrong #2340 should be merged first?

frontend/src/styles/hightlight-js-style.css
```diff
@@ -47,27 +47,27 @@
...
 }
 
 .hljs-link {
-    text-decoration: underline;
+  text-decoration: underline;
 }
 
 .hljs-number {
-    color: #005cc5;
+  color: #005cc5;
```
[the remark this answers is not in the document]
Airiinnn: I know it's general not a good idea to mix formatting changes in, but it was causing the frontend tests to fail. Likely due to a change in config even though the change was completely unrelated to checks.

Airiinnn:
> i think the logic between `c-summary-charts` and `c-ramp` is getting a bit complicated! i think we should either:
> 
> 1. pass in `this.isPortfolio` and `this.optimiseTimeline` from `c-summary-charts` to `c-ramp`, and calculate `sdate` and `udate` accordingly inside `c-ramp`
> 
> 2. or calculate all the above in `c-summary-charts`

@jonasongg That's a good idea, I think the first option makes more sense so will try that first. Would it be better to do those refactoring changes together in another refactoring PR that Ryan suggested?

> also if i'm not wrong https://github.com/reposense/RepoSense/pull/2340 should be merged first?

Yeah would be better to. This PR assumes the `summary.json` is correct (ie. repository custom date range always present when `isPortfolio: true`, and never present otherwise) and works without it assuming correct config.

frontend/src/components/c-summary-header.vue
```diff
@@ -0,0 +1,347 @@
+<template lang="pug">
```
CYX22222003, line 1: This part may conflict with #2106 which introduce a new global search. May need some efforts to synchronize the merging of these two PRS

jonasongg (approved):
@Airiinnn yep, i think we can leave that refactoring for another PR! and since user.sinceDate will always be present when isPortfolio === true (based on backend behaviour), LGTM

sopa301:
@Airiinnn could you resolve the merge conflicts? Thanks!

PR #2321, opened 2025-03-10
docs: Remove Netlify for docs

### LINKED ISSUE #2307: Remove Netlify References from README and index.html

Now that we do not use Netlify for our deployment previews or deployments, we can remove them from the relevant parts of our documentation.

joeng03:
Fixes #2307

```
Remove Netlify References from README and index.html](https://github.com/reposense/RepoSense/issues/2307

Remove Netlify references in the documentation since we are no longer
using it for deployments and previews.
```


jonasongg (approved):
LGTM!

sopa301 (changes requested):
I think we can leave the deployment method on our docs because they still work. It's just that we don't use them for our own deployments or previews!

PR #2322, opened 2025-03-14
[#2106] Implement comparison of files across repos

### LINKED ISSUE #2106: Give a way to compare a subset of files across repos

CYX22222003:
Part of #2106 

### Proposed commit message

```
Enable users to compare a file across repositories

Users can specify a blob and the code panel shows code changes to all
files matching the glob

Let's add the input field for the blob and update code panel using
URL parameters
```

### Other information

May need to update the relevant images in the UG and DG

New change in UI:

https://github.com/user-attachments/assets/c03b3c42-ac14-41bf-9f21-31a340ba5b6b


joeng03 (approved):
LGTM. Probably don't need a local search bar for globs anymore since we already have a global one. A good starting point as it integrates a global search bar for globs. Full requirements may require another PR as code panel is currently used to display 1 repository only, and changes to include all repositories can be potentially complex.

CYX22222003:
> LGTM. Probably don't need a local search bar for globs anymore since we already have a global one. A good starting point as it integrates a global search bar for globs. Full requirements may need another PR as code panel is currently used to display 1 repository only, and changes to include all repositories can be potentially complex.

Thanks for the review!

I think the issue currently does not explicitly require display all documents in one code panel. But it may be a good feature to discuss for the future development. 

Moreover, I think that whether to remove the local search is a good issue for discussion. Maybe we still want to keep the local search bar for each ramp. There may be a case where users are not comparing across repos but just trying to find a local glob. Removing the local search bar may cause them to reset global search when they switch to a different repo.

frontend/cypress/tests/codeView/codeView_renderFilterHash.cy.js
```diff
@@ -1,4 +1,33 @@
 describe('render filter hash', () => {
+  it('filter files: url params should persist after change and reload', () => {
+    cy.get('div.mui-textfield.filter_file > input:visible')
+      .should('be.visible')
+      .invoke('val')
+      .should('eq', '');
+
+    cy.get('div.mui-textfield.filter_file > input:visible')
+      .should('be.visible')
+      .type('**java**{enter}');
```
Airiinnn, line 10: Is pressing enter necessary for filter to update? Seems like the implementation updates on change

frontend/src/views/c-home.vue
```diff
@@ -91,11 +91,11 @@ const home = defineComponent({
       required: true,
     },
     repos: {
-      type: Object,
+      type: Object, //{ [key: string]: Repo }
       required: true,
     },
     users: {
-      type: Array,
+      type: Array, //Array<Repo>
```
Airiinnn, line 98: Thanks for clarifying! Would it be possible specify the type directly instead of comments? Or would it potentially break things?

frontend/src/views/c-summary.vue
```diff
@@ -401,12 +407,29 @@ export default defineComponent({
...
       this.filterSearch = (evt.target as HTMLInputElement).value;
       this.getFiltered();
     },
 
+    resetFilteredFileName() : void {
+      this.filteredFileName = '';
+      window.removeHash('authorshipFilesGlob');
+      this.$store.commit("updateAuthorshipRefreshState", false);
+      this.getFiltered();
+      window.location.reload();
```
Airiinnn, line 422: nit (Sorry if I am wrong):

I think `this.getFiltered();` may not be necessary since you're already setting the hash here, and this filter file name does not affect filtered repos. Also, why is there a need for a reload?

Airiinnn (approved):
Changes looks good, just some comments for consideration. 

Just want to bring up that this doesn't seem to match the desired behavior in #2106. Has the requirements changed or is this a part of a bigger implemention? I understand that the behavior stated in #2106 would likely be a big change.

CYX22222003:
> Changes looks good, just some comments for consideration.
> 
> Just want to bring up that this doesn't seem to match the desired behavior in #2106. Has the requirements changed or is this a part of a bigger implemention? I understand that the behavior stated in #2106 would likely be a big change.

Regarding the changes, I think this implementation is satisfied with the current requirements. The idea of displaying all relevant blobs in a single page is a possible feature we can discuss about.

frontend/src/views/c-summary.vue
CYX22222003: the reload is to update the opened code panel when attempting to change the global search keyword for blurb.

Without `this.getFiltered` the hash will not be updated properly after you click cancel and try to clear the inputs

frontend/src/views/c-home.vue
CYX22222003: Good catch. I remember you have discovered several flaws in typing on frontend. Shall we create a separate issues to work on this in the future?

frontend/cypress/tests/codeView/codeView_renderFilterHash.cy.js
CYX22222003: Yes, it follows the similar mechanism of filter by repo/author name.

sopa301 (approved):
LGTM!

docs/ug/usingReports.md
```diff
@@ -90,6 +90,10 @@ We allow contribution bars to overflow into multiple lines (rather than adjust t
 #### Tool bar
 
 The `Tool Bar` at the top of the Chart panel provides a set of configuration options that control the Chart panel.
+* `filter file`: filters the files and globs by keyword.
```
jonasongg, line 93:
```suggestion
* `Filter File`: filters the files and globs by keyword.
```

docs/ug/usingReports.md
```diff
@@ -90,6 +90,10 @@ We allow contribution bars to overflow into multiple lines (rather than adjust t
...
+  * Specifies a glob (e.g., **/TaskList.java)
+  * users can see the code changes to all the files matching the glob across all repositories by opening the code panel
```
jonasongg, line 95:
```suggestion
  * Users can see the code changes to all the files matching the glob across all repositories by opening the code panel
```

jonasongg (approved):
LGTM! Should we also ask Prof Damith how he find this UI change, or have you already asked him?

CYX22222003:
> LGTM! Should we also ask Prof Damith how he find this UI change, or have you already asked him? We can merge this as soon as this is ok :)

I have demonstrated the new feature to Prof last session and he is fine with the change we made.

damithc:
> Should we also ask Prof Damith how he find this UI change, or have you already asked him? We can merge this as soon as this is ok :)

In the UI, should it be `filter files` rather than `filter file`?
Should we also give a tool tip for this one, to give a bit more info on how to use this filter?

damithc:
> Just want to bring up that this doesn't seem to match the desired behavior in #2106. Has the requirements changed or is this a part of a bigger implemention? I understand that the behavior stated in #2106 would likely be a big change.

@Airiinnn Good point. Yes, this one doesn't meet the initial requirement but perhaps a 'good enough' solution for the time being. Yes, meeting the initial requirement of #2106 will take a lot more work, as it requires a significantly different view to be loaded into the code panel.

sopa301 (changes requested):
As Prof mentioned, let's add a tool tip to make it easier to use. For now, I think using the same one as the Code Panel should be fine!

sopa301:
image

The tooltip is cut off by the top of the screen, and it only shows when mousing over the clear input button.

CYX22222003:
> image
> 
> The tooltip is cut off by the top of the screen, and it only shows when mousing over the clear input button.

It is a bit hard to find a place to insert the tooltip with another icon and all other search fields doesn't seem to have tooltip embedded. Any suggestions?

sopa301:
I think this would be sufficient for now:
image

CYX22222003:
Updated UI/UX demo: [filter-files-demo.webm](https://github.com/user-attachments/assets/52c2ff41-1ca8-41b7-b0fe-ac21c1d09b53)

frontend/src/components/c-summary-header.vue
```diff
@@ -1,6 +1,20 @@
...
+      )
+        .mui-textfield.filter_file(v-if='!isPortfolio')
+          label filter files
+          input(
+            type="text",
+            @change="setFilteredFileName",
+            v-model="localFilteredFileName"
+            )
+          button.mui-btn.mui-btn--raised(type="button", @click.prevent="resetFilteredFileName") x
+          span.tooltip-text(ref='filter-files-label') Type a glob keyword to filter the list
```
sopa301, line 16:
```suggestion
          span.tooltip-text(ref='filter-files-label') Enter a glob to filter the files
```

Nit

sopa301 (approved):
LGTM!

PR #2325, opened 2025-03-15
[#2326] Optimize Vite Chunk Size

### LINKED ISSUE #2326: Vite Chunk Size Exceeds 500kB after Minification
Currently, we have this warning when building the Vite frontend:

(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rollupOptions.output.manualChunks to improve chunking: https://rollupjs.org/configuration-options/#output-manualchunks
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.

Optimizations can be done to cut down the chunk size to reduce the size of files we have to load to the frontend, potentially improving performance.

joeng03:
Fixes #2326

```
Optimize Vite Chunk Size

Did the following optimizations to reduce max chunk size from ~1.5MB to
~0.8MB. 


Manually split the Vue libraries and large utility libraries into
their own chunks. This reduces the max chunk size since we now have 3
main files instead of having everything in the index file as one large
chunk.

Also set the chunkSizeWarningLimit to 1000 for now, since the
highligh.js package is already 862.99kB on its own.

Note:
Command to generate the breakdown of chunk sizes:  
npx vite-bundle-visualizer -t sunburst 

```

Before:
Screenshot 2025-03-15 at 4 56 02 PM

Screenshot 2025-03-15 at 5 44 58 PM

After:
Screenshot 2025-03-15 at 7 15 49 PM
Screenshot 2025-03-15 at 7 20 21 PM

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -38,6 +38,8 @@
       "devDependencies": {
         "@stylistic/eslint-plugin": "^1.7.0",
         "@stylistic/stylelint-plugin": "^2.1.0",
+        "@types/markdown-it": "^13.0.7",
+        "@types/minimatch": "^5.1.2",
         "@types/node": "^20.12.4",
         "@vue/eslint-config-typescript": "^13.0.0",
         "eslint": "^8.57.0",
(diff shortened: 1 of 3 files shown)


JoanneHing (commented):
LGTM. But the last line of commit message might need to be split into two lines as it seems to exceed 72 characters.

sopa301:
@JoanneHing could I trouble you to add an approving review? Thanks!

JoanneHing (approved):
LGTM!

PR #2331, opened 2025-03-16
[#2210] Add default sort order for groups

### LINKED ISSUE #2210: Support custom sort order in the report
Current: 
image

The problem: There is no way to display items in a custom order that we want (e.g., order of importance)

Suggestion: Allow a `none` (or `custom`) option that will sort items in the order they are specified in config files. That way, we can order the items in config files in any way we want and get the report to follow that same order.

This can apply to sorting of both groups and within groups.

joeng03:
Fixes #2210

```
Add custom sort order for groups in
RepoSense. 

Now, we have 2 more options:
custom and custom desc. It will order
the groups based on the initial order
of the rows in repo-config.csv


Note: If we group by repos or group by authors, then the order among
groups will take the custom rank of the first element in the group. For
example, if the groups to be sorted are: {[1, 6, 10], [4, 7, 9], [3,
5], [2, 8]}, where the numbers are the custom rank of the user repos,
then the final sorted groups will be {[1, 6, 10], [2, 8], [3, 5], [4,
7, 9]}
```


sopa301 (changes requested):
I think it would be good to update the documentation as well!

Nit: I'm not sure if `custom` would be an apt naming, but I don't really have a better idea. Maybe `default`?

joeng03:
> Just a clarification about an edge case: when it is group by author, what is the sorting order of case like {[1], [1], [1, 2], [1]} ? This may be common in the case of CS2103T tp where more than one coder has the same order.


After looking into it, I realized that several other sorting options also face the same issue where there is no defined order if the values have a tie, for example, when grouped by repo, sort the groups by author displayName (there is a possible case where an author contributed to multiple repos, and is the lexicographically largest display name among contributors of both repos.

I would propose to resolve this ambiguity in another PR, where each grouping option (by author or repo) can only sort by a fixed set of fields (for example, sort by default order might not make sense for group by author, and sort by author displayName might not make sense for group by repo). Then for sort fields that might have the same value (for example, totalCommits might be the same for several authors/repos), maybe define a default way to break ties (for example, the default ordering of repos, or lexicogaphical order of author displayName.

joeng03:
> I think it would be good to update the documentation as well!
> 
> Nit: I'm not sure if `custom` would be an apt naming, but I don't really have a better idea. Maybe `default`?

Just updated the naming from custom to default.

frontend/src/types/summary.ts
```diff
@@ -11,6 +11,8 @@ export enum SortGroupSelection {
   TotalCommitsDsc = 'totalCommits dsc',
   Variance = 'variance',
   VarianceDsc = 'variance dsc',
+  Custom = 'defaultSortOrder',
+  CustomDsc = 'defaultSortOrder dsc'
```
Airiinnn, line 15: I think you may have missed renaming these from custom to default as well?

frontend/src/views/c-summary.vue
```diff
@@ -517,8 +517,8 @@ export default defineComponent({
             if (this.isMatchSearchedUser(this.filterSearch, user)) {
               this.getUserCommits(
                 user,
-                new Date(this.filterSinceDate) > new Date(user.sinceDate) ? this.filterSinceDate : user.sinceDate,
-                new Date(this.filterUntilDate) < new Date(user.untilDate) ? this.filterUntilDate : user.untilDate,
+      new Date(this.filterSinceDate) > new Date(user.sinceDate) ? this.filterSinceDate : user.sinceDate,
+       new Date(this.filterUntilDate) < new Date(user.untilDate) ? this.filterUntilDate : user.untilDate,
```
Airiinnn, line 521: nit: I think indentation was previously correct?

frontend/src/types/types.ts
```diff
@@ -56,6 +56,7 @@ export interface User {
 
 // We add these three fields in loadCommits and loadAuthorship of utils/api.ts
```
Airiinnn, line 57: nit: Could you update this comment? 

Seems like `defaultSortOrder`, `commits`, `users` added in `loadCommits`.
`files` added in `loadAuthorship`.

Airiinnn (changes requested):
The default descending sorting doesn't seem to be working. Both default and default descending is sorting in the same order (screenshots below). I also don't see any logic handling the descending case. 

Also, I think would be good to include cypress tests for these types of changes.

Default:

image

Default descending:

image

`repo-config.csv`
```
Repository's Location,Branch,File formats,Ignore Glob List,Ignore standalone config,Ignore Commits List,Ignore Authors List,Shallow Cloning,Find Previous Authors,Since Date,Until Date
https://github.com/reposense/RepoSense.git,cypress,,,,,,,,,
https://github.com/reposense/testrepo-Empty.git,master,,,,,,,,,
https://github.com/reposense/testrepo-Empty2.git,master,,,,,,,,,
https://github.com/reposense/testrepo-Empty3.git,master,,,,,,,,,
https://github.com/reposense/testrepo-Empty4.git,master,,,,,,,,,
https://github.com/reposense/testrepo-Empty5.git,master,,,,,,,,,
https://github.com/reposense/publish-RepoSense.git,master,,,,,,,,,
https://github.com/reposense/repoSense-action.git,main,,,,,,,,,
https://github.com/reposense/RepoSense-auth-helper.git,master,,,,,,,,,
```

frontend/src/types/summary.ts
joeng03: I think that was the cause of the unexpected behavior. Thanks for spotting!

frontend/src/views/c-summary.vue
joeng03: Linter gave an error previously hence the adjustment. Will try to revert it and test again.

sopa301:
I think a line or two in the UG to explain the sort order would be helpful for some users!

sopa301 (approved):
LGTM!

Airiinnn (approved):
LGTM!

PR #2337, opened 2025-03-17
[#2332] Frontend Sort Function Bugfix and Enhancement

### LINKED ISSUE #2332: Frontend Sort Function Enhancement
Currently, the frontend code sort functions just do a .reverse() if the order is descending. We can save some time (could potentially be significant if there are a lot of items) by modifying the default sorting logic to take into account whether order is descending by using a positive/negative flag.

joeng03:
Fixes #2332, #2362

```
Frontend Sort Function Enhancement

Improve frontend sort function efficiency for descending order by
refactoring to use boolean, and modify the sort comparator to take in 1
additional argument `isDesc` instead of calling `.reverse()` after
doing an ascending sort.


Important: A Cypress test case was modified to account for corrected
behavior because previous code for 'groupByAuthors' contains a bug
where we use `sortingWithinOption` instead of `sortWithinOption`, thus
causing the sort order within a group to be arbitrary if we sort by
'title' 
since the property doesn't exist and we are sorting an array of blank
strings.
```

### DIFF
--- a/frontend/src/utils/api.ts
+++ b/frontend/src/utils/api.ts
@@ -104,9 +104,9 @@ window.comparator = (fn, isDesc = false, sortingOption = '') => function compare
   const descMultiplier = isDesc ? -1: 1;
 
   if (a1 === b1) {
-    return  0;
+    return 0;
   }
-  return (a1 < b1 ? -1 : 1) * descMultiplier as -1 | 0 | 1;
+  return (a1 < b1 ? -1 : 1) * descMultiplier as -1 | 1;
 };
 
 window.filterUnsupported = function filterUnsupported(string) {
--- a/frontend/src/utils/api.ts
+++ b/frontend/src/utils/api.ts
@@ -84,7 +84,7 @@ window.decodeHash = function decodeHash() {
   window.hashParams = hashParams;
 };
 
-window.comparator = (fn, sortingOption = '') => function compare(a, b) {
+window.comparator = (fn, isDesc = false, sortingOption = '') => function compare(a, b): -1 | 0 | 1 {
   let a1;
   let b1;
   if (sortingOption) {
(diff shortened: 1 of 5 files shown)


frontend/src/utils/api.ts
```diff
@@ -100,12 +100,13 @@ window.comparator = (fn, sortingOption = '') => function compare(a, b) {
...
     b1 = b1.toLowerCase();
   }
+
+  const descMultiplier = isDesc ? -1: 1;
+
   if (a1 === b1) {
-    return 0;
-  } if (a1 < b1) {
-    return -1;
+    return  0;
```
jonasongg, line 107:
```suggestion
    return 0;
```

frontend/src/utils/api.ts
```diff
@@ -100,12 +100,13 @@ window.comparator = (fn, sortingOption = '') => function compare(a, b) {
...
   }
-  return 1;
+  return (a1 < b1 ? -1 : 1) * descMultiplier as -1 | 0 | 1;
```
jonasongg, line 109:
```suggestion
  return (a1 < b1 ? -1 : 1) * descMultiplier as -1 | 1;
```

jonasongg (approved):
LGTM! just some very minor notes

frontend/src/utils/repo-sorter.ts
```diff
@@ -65,28 +65,21 @@ function groupByRepos(
...
     if (sortWithinOption === 'totalCommits') {
-      users.sort(window.comparator((ele) => ele.checkedFileTypeContribution ?? 0));
+      users.sort(window.comparator((ele) => ele.checkedFileTypeContribution ?? 0, isSortingWithinDsc));
     } else {
-      users.sort(window.comparator((ele) => getComparablePrimitive(ele, sortWithinOption)));
-    }
-
-    if (isSortingWithinDsc) {
-      users.reverse();
+      users.sort(window.comparator((ele) => getComparablePrimitive(ele, sortWithinOption), isSortingWithinDsc));
```
CYX22222003, line 70: Good to use the additional flag for sorting order

CYX22222003 (approved):
LGTM

PR #2338, opened 2025-03-17
[#2212] Refresh Only Text

### LINKED ISSUE #2212: Provide a way to refresh text of a report
When I was tweaking content of title.md and/or blurbs.md, I had to rerun the `java -jar RepoSense.jar --view` command every time I wanted to see how my changes appeared on the report. As my report included some big repos, this took about 5 minutes each time.

I wonder if we can provide a way to just refresh the text only e.g., `java -jar RepoSense.jar --text --view` that will allow a user to quickly preview the report (i.e., reuse the data generated from the previous runs but regenerate the text content from title.md and blurbs.md).

lyuanww:
Fixes #2212 

### Proposed commit message

```
Add flag to refresh report text without regenerating data.

Currently, users must regenerate all report data when making changes 
to title.md or repo-blurbs.md or author-blurbs.md files, which can take
several minutes for large repositories.

Let's add a flag that allows users to quickly refresh text content from
markdown files while reusing previously generated data.
```


lyuanww:
Performance comparison results w/o the --text flag across 5 runs:

Test environment: On my Macbook Pro
Test command: ./gradlew run --args='--text'

| Metric          | without flag | with flag | Improvement |
|-----------------|-----------|-----------|-------------|
| Execution Time  | 32.8s     | 10.8ms     | 67% faster  |

CYX22222003 (commented):
This PR indeed reduce the time for repeatedly cloning and analyzing repos after rerun. LGTM

When I looked at the logger information, it seems that the rerun is still triggering the parsing the CSV config file. Is this the expected behavior? I suppose it should just copy the `title.md` and update the `blurbMap` of the `summary.json`.

image

src/main/java/reposense/RepoSense.java
```diff
@@ -85,11 +85,11 @@ public static void main(String[] args) {
...
                     TimeUtil::getElapsedTime, cliArguments.getZoneId(), cliArguments.isFreshClonePerformed(),
                     cliArguments.isAuthorshipAnalyzed(), cliArguments.getOriginalityThreshold(),
-                    blurbMap, cliArguments.isPortfolio()
+                    blurbMap, cliArguments.isPortfolio(), cliArguments.isOnlyTextRefreshed()
             );
 
-            FileUtil.zipFoldersAndFiles(reportFoldersAndFiles, cliArguments.getOutputFilePath().toAbsolutePath(),
-                    ".json");
+            FileUtil.handleZipFilesAndFolders(reportFoldersAndFiles, cliArguments.getOutputFilePath().toAbsolutePath(),
+                    cliArguments.isOnlyTextRefreshed(), ".json");
```
CYX22222003, line 92: A way to prevent re-parsing the CSV file is to handleZipFilesAndFolders when there is a text flag earlier before line 55 `getRepoConfigurations` is executed.

lyuanww: Thanks for the suggestion. The reason why I didn't move `handleZipFilesAndFolders` above is because the summary.json is modified in `reportGenerator.generateReposReport` in line79. The summary.json file can only be zipped into the zip folder after it is modified.

lyuanww:
@CYX22222003 Thanks for the review! Yes, the re-parsing of the CSV files is expected. I kept this behavior for two main reasons:

1. Even when the configuration files are re-parsed, it doesn't introduce any side effects to later runs because the system skips the repo analysis and cloning steps in `reportGenerator.generateReposReport`. Only some information in the summary.json file gets updated, and these updates aren't affected by configuration changes.
2. I could have added code to skip parsing the configuration files, but it wouldn't make the system much faster. Using the same logic, I'd need to skip other parts too, and the performance benefit just isn't worth the extra complexity.

Please let me know what you think. Thank you!

CYX22222003:
I agree that the main factors affecting the time is the cloning and generating JSON. The time for parsing of CSV file may not be so significant compared to these two. 

However, a change should not be that complicated. I try to insert below code before running of `getRepoConfigurations` and it runs OK.

```
if (cliArguments.isOnlyTextRefreshed()) {
                FileUtil.addOrReplaceFileInZipFile(
                        cliArguments.getOutputFilePath().toAbsolutePath(),
                        cliArguments.getOutputFilePath().toAbsolutePath(),
                        SummaryJson.SUMMARY_JSON_FILE_NAME,
                        FileUtil.ZIP_FILE);

                logger.info(TimeUtil.getElapsedTimeMessage());
                // ReportServer.startServer(SERVER_PORT_NUMBER, cliArguments.getOutputFilePath().toAbsolutePath());
                return;
            }
``` 

This further reduced the time taken to generate report compared to my previous experiment.
image

src/main/java/reposense/util/FileUtil.java
```diff
@@ -387,4 +393,48 @@ public static boolean pathExists(String path) {
...
+    }
+
+    /**
+     * Handles the file zipping logic.
+     * If there is a flag which indicates that only text is refreshed, only summary.json is overwritten.
+     */
+    public static void handleZipFilesAndFolders(List<Path> pathsToZip, Path sourceAndOutputPath,
+                                                boolean refreshOnlyText, String... fileTypes) {
+        if (refreshOnlyText) {
+            addOrReplaceFileInZipFile(sourceAndOutputPath, sourceAndOutputPath, SummaryJson.SUMMARY_JSON_FILE_NAME,
```
CYX22222003, line 431: If the `outputPath` and `sourcePath` are the same and this seems to be the only use case of `addOrReplaceFileInZipFile`, is it possible to avoid duplicate parameters?

lyuanww:
In my opinion, I don't think the performance improvement will not very significant for us to implement this explicitly for the sake of skipping parsing.

CYX22222003 (approved):
LGTM

joeng03 (approved):
Tested and it works on my end. LGTM

sopa301 (approved):
LGTM!

PR #2319, opened 2025-03-08
[#2308][#2311] Add author blurbs and repo blurbs to the frontend

CYX22222003:
Part of #2311, #2308

### Proposed commit message

```
Add frontend support for adding blurbs for both users and ramps.

Previously, blurbs can only be added for each repo.

Let's add user-specific and ramp-specific markdown blurb.
```

### Other information

UI after changes:

new_UI

new_UI2

new_UI3

Issue #2170, opened 2024-03-25
Add Blurbs for Repos


sopa301:
### What feature(s) would you like to see in RepoSense

Add blurbs for repositories. These blurbs will be displayed above the relevant repository on the left panel when it is grouped by Repo/Branch. 

Something to be decided is the granularity of the blurb. It can either:
1. Be specific to the repository as a whole
2. Be specific to each branch

### Is the feature request related to a problem?
No.

### If possible, describe the solution
Create a yaml file that can store the blurb mappings as key-value pairs and store it in the `assets` directory.
We can then either:
1. Process the file during the running of the jar file.
2. Process the file when the report is opened.

ckcherry23:
As discussed in issue #2191, this issue can have the following 3 subtasks:

- [x] 1. Add blurbs for repos (Done in #2191)
- [ ] 2. Add blurbs for authors
- [ ] 3. Add blurbs for individual charts

The current code is implemented such that blurbs can only be used for repos. In case we foresee adding blurbs for subtask 2: author groups (eg. CS3281 summary for a student) or for subtask 3: repo-author charts (eg. each CS2103T student mentions their contribution to the tP), then we can track these discussions here.

PR #2339, opened 2025-03-20
[#2311] Add blurbs for Authors 

### LINKED ISSUE #2311: Add blurbs for Authors
Add blurbs for repository authors.

As purposed in #2170, we can use a Markdown file with a specialized format to record the blurbs.

```
(link to repo branch)
(blurb)
<--delimiter-->
etc
```

We can also refer to the PR #2191  for relevant updates.

JoanneHing:
Fixes #2311 
Part of #2170 
Closes #2319 
### Proposed commit message
```
Add support for author blurbs

RepoSense currently only supports repository blurbs. Adding author  
blurbs allows developers to include personal messages about their  
contributions.
```
UI after changes:

new_UI

new_UI2

new_UI3


frontend/src/types/zod/summary-type.ts
```diff
@@ -45,9 +45,15 @@ export const summarySchema = z.object({
...
   isAuthorshipAnalyzed: z.boolean().default(false), // for backwards compatability
   supportedDomainUrlMap: supportedDomainUrlMapSchema,
-  blurbs: z.object({
+  repoBlurbs: z.object({
     urlBlurbMap: z.record(z.string(), z.string()),
   }),
+  authorBlurbs: z.object({
+    authorBlurbMap: z.record(z.string(), z.string())
+  }).optional(),
+  chartsBlurbs: z.object({
```
CYX22222003, line 54: Actually is it really necessary to use different key name for each of the blurb map? Maybe use the same name such as `blurbMap` can reduce the need to override the parsing in each specific blurb parser since they are just doing the same thing. This may better follow the DRY.

It is unlikely that the the format of blurb map to change.

src/main/java/reposense/model/RepoBlurbMap.java
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Adds a key-value record into the {@code BlurbMap}.
+     *
+     * @param key Key value.
+     * @param value Blurb value.
+     */
+    public void withRecord(String key, String value) {
+        this.urlBlurbMap.put(key, value);
+    }
```
CYX22222003, line 31: This two methods look very similar to the ones in `AuthorBlurbMap`. For better abstraction and avoid repetition, is it better if we implement this in the interface and not overriding it?

src/main/java/reposense/parser/AuthorBlurbMarkdownParser.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        public List<String> getBlurb() {
+            return blurb;
+        }
+
+        public int getNextPosition() {
+            return nextPosition;
+        }
+    }
```
CYX22222003, line 62: This static class is appearing in `RepoBlurbMarkdownParser.java` again. Should we further abstract this out?

src/main/java/reposense/parser/AuthorBlurbMarkdownParser.java
```diff
@@ -0,0 +1,176 @@
...
+            throw new InvalidMarkdownException("Empty author-blurbs.md file");
+        }
+
+        // Prepare the blurb map
+        AuthorBlurbMap authorBlurbMap = new AuthorBlurbMap();
+
+        // Define temporary local variables to track blurbs
+        String authorGitId = "";
+        StringBuilder blurb = new StringBuilder();
+        int counter = 0;
```
CYX22222003, line 92: This also seems to be very similar to the implementation in `RepoBlurbMardownParser`. Shall we provide higher level abstraction for this method?

CYX22222003 (commented):
Good to update the mapping and parsing classes to accommodate more blurb formats. Just some suggestions to improve the abstraction and avoid repetition. If it is not so significant and out of scope, the current implementation will be good enough.

src/main/java/reposense/model/RepoBlurbMap.java
JoanneHing: Thank you for pointing this out! Do you mean to abstract them down in abstract class? I did try to refractor out some similar code when I trying to built up the `authorBlurbs` and currently looking over if I could update the code to be more clean. Thank you for pointing this out I could try see if I could abstract the methods better!

src/main/java/reposense/parser/AuthorBlurbMarkdownParser.java
JoanneHing: Thanks for pointing this out too!

src/main/java/reposense/parser/BlurbMarkdownParser.java
```diff
@@ -1,163 +1,22 @@
...
-        public int getNextPosition() {
-            return nextPosition;
-        }
-    }
-
-    public BlurbMarkdownParser(Path markdownPath) throws FileNotFoundException {
-        super(markdownPath);
-    }
-
+public interface BlurbMarkdownParser {
```
joeng03, line 11: Since interface `BlurbMarkdownParser` defines `BlurbMap parse() throws IOException, InvalidMarkdownException;` but MarkdownParser class also defines `public abstract T parse() throws IOException, InvalidMarkdownException;`,
 there is no need to declare implements `BlurbMarkdownParser` in `AuthorBlurbMarkdownParser` and `RepoBlurbMarkdownParser` because they already extend from MarkdownParser. Instead, MarkdownParser should declare implements `BlurbMarkdownParser`.

src/main/java/reposense/parser/AuthorBlurbMarkdownParser.java
joeng03: Suggest placing common logic/class in the parent class, `MarkdownParser`.

joeng03 (changes requested):

src/main/java/reposense/parser/AuthorBlurbMarkdownParser.java
JoanneHing: Had created an abstract class `BlurbMarkdownParser` to handle the common functionality and classes! Thanks for the suggestion!

src/main/java/reposense/parser/BlurbMarkdownParser.java
JoanneHing: Had implemented this in the `BlurbMardownParser` too. Thanks for pointing this out!

CYX22222003 (approved):
LGTM!

joeng03 (approved):
LGTM!

frontend/src/utils/api.ts
```diff
@@ -237,13 +237,23 @@ window.api = {
       names.push(repoName);
     });
 
-    const blurbMap: { [key: string]: string } = data.blurbs.urlBlurbMap;
+    const repoBlurbMap: { [key: string]: string } = data.repoBlurbs.blurbMap;
+    const authorBlurbMap: {[key: string]: string} | undefined =
```
jonasongg, line 241: let's simplify this line and the line below to:

```suggestion
    const authorBlurbMap: {[key: string]: string} | undefined = data.authorBlurbs?.blurbMap
```

jonasongg (changes requested):
FE LGTM!

frontend/src/utils/api.ts
JoanneHing: Alright! The code is updated!

jonasongg (approved):

frontend/src/components/c-summary-charts.vue
```diff
@@ -162,7 +162,13 @@
       v-if="filterGroupSelection === 'groupByRepos'",
     )
       c-markdown-chunk.blurb(
-        :markdown-text="getBlurb(repo[0])"
+        :markdown-text="getRepoBlurb(repo[0])"
+      )
+
+    .blurbWrapper(
```
jonasongg, line 168: one more minor thing - could we change this to .blurp-wrapper thank you!

Issue #2368, opened 2025-03-31
Use Text Input for Since and Until date only if Safari version < 14.1


joeng03:
Currently for all Safari versions, we use text input instead of HTML date picker to input dates for the since and until date. However, since support for HTML date picker on MacOS has already been added to Safari 14.1 ~4 years ago, propose that we use the text input only if the detected Safari version is <14.1 to improve user experience for MacOS users.

Refer to: https://webkit.org/blog/11648/new-webkit-features-in-safari-14-1/

joeng03:
Addressed in #2369

PR #2369, opened 2025-03-31
[#2366] Support datetime for since and until CLI config

### LINKED ISSUE #2366: Support time specification for since/until
Image

Reason: In some cases, we need to filter based on specific times, not just the date

In fact, even more than in the filter bar, this feature is needed for the --until --since flags.

joeng03:
Fixes #2366, #2368

```
Support datetime (previously just date) for since and until in both
Backend CLI and Frontend.

3 supported formats for since and until dates: 
1. DD/MM/YYYY
2. DD/MM/YYYY'T'HH:mm
3. DD/MM/YYYY'T'HH:mm:ss

For example:
`./gradlew run -Dargs="--since 31/12/2017T10:10"`


Rationale for using 'T' instead of blank space is to adhere to ISO 8601
format and prevent complicating the CLI input where it will need an
additional set of quotation marks to identify the space-separated
date.

Behavior: For since date, if hour/minute/seconds not supplied then
default to 0 for each. For until date, if hour/minute/seconds not
supplied then default to 23, 59, 59 respectively. If only one date
(since or until) is provided, and a period is also provided (default to
1 month in our existing project), the time of since is truncated to
00:00:00 if until is supplied, and time of until is truncated to
23:59:59 if since is supplied.

We use text input instead of datetime input only if the
detected Safari version is <14.1 to improve user experience for MacOS
users. 



Note for systemtest: For "Asia/Singapore" timezone, on dates before
01-01-1982, where Singapore and Malaysia switched from UTC+07:30 to
UTC+08:00, all date conversions will result in +7.5 hours instead of +8
hours. For example, the Unix epoch 1970-01-01 00:00:00 will be
1970-01-01 07:30:00 with "Asia/Singapore" timezone, but 1970-01-01
08:00:00 with "UTC+08" timezone.
Reference: https://en.wikipedia.org/wiki/UTC%2B07:30.
```


damithc:
@joeng03 My review is usually from the user's POV; So, I usually look at changes to the UG. Does this PR updates the UG?

docs/ug/cli.md
```diff
@@ -254,10 +254,16 @@ Cannot be used with `--last-modified-date`. This may result in an incorrect last
 ### `--since`, `-s`
 
 **`--since START_DATE`**: Specifies the start date for the period to be analyzed.
-* Parameter: `START_DATE` The first day of the period to be analyzed, in the format `DD/MM/YYYY`.<br>
-  Default: one month before the current date
+* Parameter: `START_DATE` The first day of the period to be analyzed, in one of the following formats:<br>
```
JoanneHing, line 257: Would changing this line to
```
* Parameter: `START_DATE` The first day of the period to be analyzed (with optional time specification), in one of the following formats:<br>
```
be slightly clearer for the user that time specification is optional?

docs/ug/cli.md
```diff
@@ -254,10 +254,16 @@ Cannot be used with `--last-modified-date`. This may result in an incorrect last
...
+  ```
+  1. DD/MM/YYYY
+  2. DD/MM/YYYY'T'HH:mm
+  3. DD/MM/YYYY'T'HH:mm:ss
+  ```
+  Default: one month before the current date. 
+  If hour/minute/seconds not supplied then default to 0 for each of them.
```
JoanneHing, line 264: Would organize this part as
```
Default:  
  - If `START_DATE` is not specified, it defaults to one month before the current date at `00:00:00`.  
  - If hours/ minutes/ seconds are not provided, each will default to `00`.  
 ```
be more clear?

File: docs/ug/cli.md [hunk 3]
```diff
@@ -280,10 +286,16 @@ Cannot be used with `--last-modified-date`. This may result in an incorrect last
...
-* Parameter: `END_DATE` The last date of the period to be analyzed, in the format `DD/MM/YYYY`.<br>
-  Default: current date
+* Parameter: `END_DATE` The last date of the period to be analyzed, in one of the following formats:<br>
+  ```
+  1. DD/MM/YYYY
+  2. DD/MM/YYYY'T'HH:mm
+  3. DD/MM/YYYY'T'HH:mm:ss
+  ```
+  Default: current date 
+  If hour/minute/seconds not supplied then default to 23, 59, 59 respectively
```
JoanneHing, line 296: Similar to above, the suggested format is
```
Default:  
  - If `END_DATE` is not specified, it defaults to the current date at `23:59:59`.  
  - If hours/ minutes/ seconds are not provided, they will default to `23`, `59`, and `59`, respectively.   
  ```

**Review by JoanneHing — COMMENTED** [review]
Shared some suggestion for updating the UG. Also I reformat the proposed commit message because some of the lines are truncated with very few characters before the next word.

```Support datetime (previously just date) for since and until in both
Backend and Frontend.

3 supported formats for since and until dates: 
1. DD/MM/YYYY
2. DD/MM/YYYY'T'HH:mm
3. DD/MM/YYYY'T'HH:mm:ss

For example:
`./gradlew run -Dargs="--since 31/12/2017T10:10"`


Rationale for using 'T' instead of blank space is to adhere to ISO
8601 format and prevent complicating the CLI input where it will need
an additional set of quotation marks to identify the space-separated
date.

Behavior: For since date, if hour/minute/seconds not supplied then
default to 0 for each. For until date, if hour/minute/seconds not 
supplied then default to 23, 59, 59 respectively. If only one date
(since or until) is provided, and a period is also provided (default
to 1 month in our existing project), the time of since is truncated
to 00:00:00 if until is supplied, and time of until is truncated to 
23:59:59 if since is supplied.


Note for systemtest: For "Asia/Singapore" timezone, on dates before
01-01-1982, where Singapore and Malaysia switched from UTC+07:30
to UTC+08:00, all date conversions will result in +7.5 hours instead
of +8 hours. For example, the Unix epoch 1970-01-01 00:00:00 will be 
1970-01-01 07:30:00 with "Asia/Singapore" timezone, but 
1970-01-01 08:00:00 with "UTC+08" timezone.
Reference: https://en.wikipedia.org/wiki/UTC%2B07:30.
```

frontend/src/components/c-summary-header.vue
```diff
@@ -161,9 +161,9 @@ export default defineComponent({
       type: Boolean,
       default: false,
     },
-    isSafariBrowser: {
+    inputDateNotSupported: {
       type: Boolean,
-      required: true,
```
sopa301, line 166: Sorry, can I check why is the `required` field replaced with a default value? Would the component always work if we don't supply this field?

sopa301 (approved):
LGTM, just one comment

config/checks/check-eof-newline.sh
lyuanww, line 1: I noticed the file mode seems to be accidentally changed. Do you mind removing it?

docs/ug/cli.md
```diff
@@ -280,10 +287,16 @@ Cannot be used with `--last-modified-date`. This may result in an incorrect last
...
-* Example:`--until 21/10/2017` or `-u 21/10/2017`
+* Parameter: `END_DATE` The last date of the period to be analyzed (with optional time specification), in one of the following formats:<br>
+  ```
+  1. DD/MM/YYYY
+  2. DD/MM/YYYY'T'HH:mm
+  3. DD/MM/YYYY'T'HH:mm:ss
+  ```
+Default:
+- If `END_DATE` is not specified, it defaults to the current date at `23:59:59`.
+- If hours/ minutes/ seconds are not provided, they will default to `23`, `59`, and `59`, respectively.   * Alias: `-u`
```
jonasongg, line 298:
```suggestion
- If hours/ minutes/ seconds are not provided, they will default to `23`, `59`, and `59`, respectively.
* Alias: `-u`
```

src/main/java/reposense/util/TimeUtil.java
```diff
@@ -18,10 +18,11 @@
 public class TimeUtil {
     private static Long startTime;
     private static final String DATE_FORMAT_REGEX =
-            "^((0?[1-9]|[12][0-9]|3[01])\\/(0?[1-9]|1[012])\\/(19|2[0-9])[0-9]{2})";
+            "^((0?[1-9]|[12][0-9]|3[01])\\/(0?[1-9]|1[012])\\/(19|2[0-9])[0-9]{2}"
+                    + "(([T ])([01][0-9]|2[0-3]):([0-5][0-9])(:([0-5][0-9]))?)?)";
```
jonasongg, line 22: is it intended to accept either T or a blank space between date and time?

jonasongg (changes requested):

frontend/src/components/c-summary-header.vue
joeng03: After looking into it, I realized this field should be required because it affects whether we should display a html <input> with type="datetime-local". Fixed it, thanks for the review!

src/main/java/reposense/util/TimeUtil.java
joeng03: Fixed to accept only T. May be a good future enhancement to accept either blank space or T for better user experience, but have to handle the double quotation in CLI correctly first.

jonasongg (approved):
LGTM :)

jonasongg:
could you also help to write a short sentence in the commit message explaining the safari change? thank you!

damithc:
Let's merge this soon. I need to use this feature soon (tomorrow!) :-p

sikai00 (approved):

sopa301:
Let's skip the usual waiting period since this is an urgent feature.

joeng03:
Hi Prof @damithc, could you give a local configuration to reproduce the regression? Thanks!

damithc:
> Hi Prof @damithc, could you give a local configuration to reproduce the regression? Thanks!

@joeng03 You can get it from https://github.com/nus-cs2103-AY2425S2/tp-dashboard
While doing something else today, I also realised that the config in the above site has fallen behind the Publish-Reposense latest config, namely .github/workflows/main.yml
Not sure if the two are related.

PR #2388, opened 2025-04-05
[#2366] Support datetime since until

### LINKED ISSUE #2366: Support time specification for since/until

joeng03:
Fixes #2366

Refer to #2369


joeng03:
Resolved issue of ramps not displaying. It was due to the color generation that relies on the commit datetime.

Screenshot 2025-04-05 at 11 20 56 PM

sopa301:
Thanks for the investigation! Perhaps we could add a quick test to catch regressions of a similar nature?

joeng03:
Screenshot 2025-04-06 at 10 18 47 AM

Hi Prof @damithc, since we currently group the commits by dates, and each date has multiple commits, we can't do filtering based on datetime as of now. Should we changed the structure of `authorDailyContributionsMap` to support this?

damithc:
> Hi Prof @damithc, since we currently group the commits by dates, and each date has multiple commits, we can't do filtering based on datetime as of now. Should we changed the structure of `authorDailyContributionsMap` to support this?

@joeng03 That's an internal implementation decision, so, I'll leave it to the team. Also, OK to keep filtering by date only (i.e., no time) for now.

joeng03:
> Thanks for the investigation! Perhaps we could add a quick test to catch regressions of a similar nature?

Will be slightly challenging to test since in the buggy version, the elements are still there, just that it’s not displaying. 

Will suggest to merge this first, then discuss how to filter by commit datetime instead of just date. Then, more tests could be added for the correct display of ramps.

sopa301 (approved):
LGTM!

frontend/src/components/c-ramp.vue
```diff
@@ -238,8 +238,8 @@ export default defineComponent({
         return '-deletes';
       }
       const timeMs = this.fromramp
-          ? (new Date(this.sdate)).getTime()
-          : (new Date(slice.date)).getTime();
```
Airiinnn, line 242: nit: Might be good to add a comment to explain why parsing local time as UTC time works in this case to avoid potential confusion.

Eg.
// Force interpretation in UTC to preserve local day boundaries as dates are in local time format

frontend/src/views/c-summary.vue
```diff
@@ -530,8 +545,8 @@ export default defineComponent({
             if (this.isMatchSearchedUser(this.filterSearch, user)) {
               this.getUserCommits(
                 user,
-                new Date(this.filterSinceDate) > new Date(user.sinceDate) ? this.filterSinceDate : user.sinceDate,
-                new Date(this.filterUntilDate) < new Date(user.untilDate) ? this.filterUntilDate : user.untilDate,
+      new Date(this.filterSinceDate) > new Date(user.sinceDate) ? this.filterSinceDate : user.sinceDate,
+      new Date(this.filterUntilDate) < new Date(user.untilDate) ? this.filterUntilDate : user.untilDate,
```
Airiinnn, line 549: nit: Indentation. Maybe we should look into linter settings for FE in a separate issue

Airiinnn (approved):
Nice fix, LGTM!

PR #2460, opened 2025-06-22
[#2447] Add support for multiple git author names

### LINKED ISSUE #2447: report-config.yaml: support for multiple git author names
Image

From the above, it looks like we can specify one author name per author; is that the case?
If yes, there is a need to support multiple author names (e.g., in csv config files, we can provide multiple names separated by semicolons.

CYX22222003:
Fixes #2447

### Proposed commit message

```
Add support for multiple author names.

This is similar to how the parser handles
authors emails.
```

### Other information
Users can specify multiple author-name in the following way:
```yaml
author-git-author-name:
  - name1
  - name2
  - name3
```

### DIFF
--- a/docs/ug/reportConfig.md
+++ b/docs/ug/reportConfig.md
@@ -36,7 +36,8 @@ repos:
         authors:
           - author-git-host-id: Your username on GitHub, GitLab or Bitbucket
             author-display-name: Your display name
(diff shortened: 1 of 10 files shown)


Airiinnn (approved):
LGTM

sopa301 (approved):
LGTM

PR #2462, opened 2025-06-26
[#2436] Change portfolio mode to use custom sort order

### LINKED ISSUE #2436: Change portfolio mode to use custom sort order

Now that custom sort order (follows initial order of rows in `repo-config.csv`) has been added in #2331, it would be better for portfolio mode to use that ordering instead of the current sort by ascending group name.

Reasoning is that there is currently no way for the user (portfolio creator) to specify the order of the repositories in portfolio mode. It would be better for the user to be able to customize the order through the config file, so they can put more important / recent projects at the top.

For portfolio mode, change the filter settings when filtering repositories to use the custom sort order. 

It should be a small change in `c-summary-portfolio.vue`.

gladwintan:
Fixes #2436 

### Proposed commit message

```
Change sort order in portfolio mode to use custom sort order.

Users can customise the order of repositories in portfolio mode using the config file.
```

### DIFF
--- a/frontend/cypress/tests/portfolioView/portfolioView_widget.cy.js
+++ b/frontend/cypress/tests/portfolioView/portfolioView_widget.cy.js
@@ -62,7 +62,7 @@ describe('widget',
 
         cy.get('.summary-charts__title--groupname')
           .should('have.length', 1)
-          .contains('reposense/repoSense-action[main]');
+          .contains('reposense/RepoSense-auth-helper[master]');
 
         cy.get('.summary-chart')
           .should('have.length', 1);
--- a/frontend/src/views/c-summary-portfolio.vue
+++ b/frontend/src/views/c-summary-portfolio.vue
@@ -172,7 +172,7 @@ export default defineComponent({
 
       const filterControl = {
         filterGroupSelection: FilterGroupSelection.GroupByRepos,
-        sortingOption: SortGroupSelection.GroupTitle,
+        sortingOption: SortGroupSelection.Default,
         sortingWithinOption: SortWithinGroupSelection.Title,
         isSortingDsc: false,
         isSortingWithinDsc: false,


frontend/cypress/tests/portfolioView/portfolioView_widget.cy.js
```diff
@@ -62,7 +62,7 @@ describe('widget',
 
         cy.get('.summary-charts__title--groupname')
           .should('have.length', 1)
-          .contains('reposense/repoSense-action[main]');
```
Airiinnn, line 65: Could we keep this and instead change the index on line 43?

Would be even better if we just got by name (`reposense/repoSense-action[main]`) instead of index. That was an oversight on my part.

This test is supposed to check that the widget link for users in the repo are correct when there are multiple users, hence why we want the `repoSense-action[main]` repo which has 2 users.

Airiinnn (commented):
Thanks for the PR! Tested locally and looks to be working.

Left a comment regarding the test change.

Could you also add a test that verifies it is following the order in `repo-config.csv` to catch future regressions? Perhaps in a new file `cypress/tests/portfolioView/portfolioView_sorting.cy.js`.

Thanks!

gladwintan:
I have updated the widget test to get the repo by name and added tests for sorting in portfolio mode.

frontend/src/views/c-summary-portfolio.vue
```diff
@@ -172,7 +172,7 @@ export default defineComponent({
 
       const filterControl = {
         filterGroupSelection: FilterGroupSelection.GroupByRepos,
-        sortingOption: SortGroupSelection.GroupTitle,
+        sortingOption: SortGroupSelection.Default,
```
CYX22222003, line 175: Tested on windows. It is working.

CYX22222003 (approved):
LGTM!

Issue #2512, opened 2026-01-18
Filter file not apply to chart panel


FisherSkyi:
### What feature(s) would you like to see in RepoSense
When I use `Filter Files` function to search for specific files across all repositories, I would like the chart in chart panel on the left side to be updated as well. (For example, only commit with files matching the pattern will be shown in the chart).

Currently only the code panel (the right half) will update (if code panel is shown).

https://github.com/user-attachments/assets/816cf461-8133-4a54-be7f-06b88dfd5724

### Is the feature request related to a problem?

I have a issue when I cannot see the visualized commit history of specific kind of files.

We have **break down by file type** but user may want to self define the type to be plotted.

### If possible, describe the solution

Redraw the chart when filter by a specific pattern.

### Additional context

This may increase the lagging.

PR #2526, opened 2026-01-25
[#2512] Add global search panel

FisherSkyi:
#2512

### Proposed commit message

```
Add global search browser

The file search feature only supports searching within a single
repository under one author.

Students in SWE courses find it useful to search across all
repositories to see how others implement certain files (e.g., testing
files, **/ug.md) by glob.

Let's add global search panel with mode toggle button in the summary
header. 

The implementation preserves the existing local search functionality
while introducing a new global search mode. A mode toggle button
replaces the text field to switch between local and global search. The
global search panel displays results across all authors and
repositories.
```


FisherSkyi:
@lyuanww In order to pass test case `generateReposReport_isOnlyTextRefreshedTrueButInvalidPath_throwsIoException` in `src/test/java/reposense/report/ReportGeneratorTest.java`, I did the following changes:
1. change the `outputPath` parameter in testing function to a `tempDir`
2. move line `prepareTemplateFile(outputPath);` in method `generateReposReport` in file `src/main/java/reposense/report/ReportGenerator.java` from the first line to the line after the `isOnlyTextRefreshed` check. 

I just want to make sure these change do not compromise the intention of testing. Also I didn't manage to find the reason why IOException isn't thrown after my change while is thrown on a fresh clone. Any help would be much appreciated.

My investigation is that the `summary.json` exist in `test/resources/ReportGeneratorTest/` folder while the testing expect it to not exist there. And `prepareTemplateFile(outputPath);` might be the cause of creating this unwanted `summary.json`.

FisherSkyi:
@jonasongg testcase `subzoom panel range should work correctly when timeline is optimised` in `chartView_optimiseTimeline.cy.js` fails as it cannot find any ramp in x-pixel range 110 to 120.
```js
cy.get('body').type(zoomKey, { release: false })
  .get('#summary-charts .summary-chart__ramp .ramp')
  .first()
  .click(110, 20)
  .click(120, 20);
```
```
AssertionError: Timed out retrying after 30000ms: Expected to find element: `#tab-zoom .ramp .ramp__slice`, but never found it.

    at Context.eval (webpack:///./tests/chartView/chartView_optimiseTimeline.cy.js:141:7)
```
Will browser window size and screen resolution influence where should we click? I think the issue is in the hardcoded 110 and 120 as the commit ramp is actually there in the chart.
image

docs/images/global-file-browser.gif
CYX22222003: better not include this in the user guide

frontend/src/components/c-summary-charts.vue
```diff
@@ -441,6 +441,10 @@ export default defineComponent({
       type: Boolean,
       default: false,
     },
+    fileFilterScope: {
+      type: String as PropType<'global' | 'local'>,
```
CYX22222003: It seems that PropType is not imported in the script

frontend/src/components/c-segment.vue
```diff
@@ -112,6 +112,7 @@ export default defineComponent({
 
   .line-content {
     padding-left: 2rem;
+    white-space: pre-wrap;
```
CYX22222003: Just to reconfirm that this will affect the behavior of existing `c-authorship-file` as this component is also being used there

frontend/src/views/c-home.vue
CYX22222003: Apologies for the oversight. It seems that we forgot to import dependencies here as well. Will be great if can help to fix this as well.

frontend/src/views/c-global-file-browser.vue
```diff
@@ -0,0 +1,517 @@
...
+        }
+        repoMap[file.repoName].push(file);
+      });
+      return Object.keys(repoMap)
+        .sort((a, b) => a.localeCompare(b))
+        .map((repoName) => ({ repoName, files: repoMap[repoName] }));
+    },
+  },
+  methods: {
+    onSearchInput(): void {
```
CYX22222003: Can remove this void methods.

CYX22222003 (commented):
In general LGTM. Just some small nitpicks and clarifications

frontend/src/components/c-summary-charts.vue
CYX22222003: Nice! Thanks for exploring this issue further. Maybe we should create a small PR to fix this one

docs/ug/usingReports.md
```diff
@@ -189,3 +188,96 @@ The `Commits panel` allows users to see the commits attributed to a specific aut
 * To promote and encourage the 50/72 rule for commit messages, a dotted vertical line will be shown for:
   * Commit message subject that exceeds 50 characters.
   * Commit message body after the 72nd character mark.
+
+## Global file browser panel
+
+The `Global File Browser` provides a way to search and browse files across all repositories in the report. Unlike the [Code panel](#code-panel), which shows code attributed to a single author within one repository, the Global File Browser lets you explore files from every repository at once.
```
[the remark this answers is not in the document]
FisherSkyi: Fixed, thanks

docs/ug/usingReports.md
```diff
@@ -189,3 +188,81 @@ The `Commits panel` allows users to see the commits attributed to a specific aut
 * To promote and encourage the 50/72 rule for commit messages, a dotted vertical line will be shown for:
   * Commit message subject that exceeds 50 characters.
   * Commit message body after the 72nd character mark.
+
+## Global file browser panel
+
+The `Global file browser` provides a way to search and browse files across all repositories in the report. Unlike the [Code panel](#code-panel), which shows code attributed to a single author within one repository, the global file browser lets you explore files from every repository at once.
+
+#### Accessing the Global File Browser
```
CYX22222003: It will be good the make the capitalization uniform and use "Global file browser" instead.

PR #2556, opened 2026-03-10
[#2542] Add scroll button component and implement on views

### LINKED ISSUE #2542: Add roll back to top button
Roll back to the top on report page. Just like markbind website.

FisherSkyi:
Fixes #2542 

### Proposed commit message

```
Add roll back to top button

Currently, users have no quick way to navigate back to the top of the
page after scrolling down.

Let's add a floating button that scrolls the user back to the top of
the page on click.
```

### Other information

N/A

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -8,7 +8,15 @@
       "groupName": "all non-major dependencies except typescript",
       "groupSlug": "all-minor-patch",
       "matchPackageNames": [
-        "!typescript"
+        "!typescript",
+        "!eslint",
+        "!eslint-*",
+        "!@eslint/*",
+        "!@stylistic/*",
+        "!stylelint",
+        "!stylelint-*",
+        "!@stylelint/*",
+        "!puglint"
       ],
       "matchDepTypes": [
         "dependencies",
@@ -22,6 +30,30 @@
       "automerge": true
     },
     {
+      "groupName": "linting dependencies",
+      "groupSlug": "linting",
+      "matchPackageNames": [
+        "eslint",
+        "eslint-*",
+        "@eslint/*",
+        "@stylistic/*",
+        "stylelint",
+        "stylelint-*",
+        "@stylelint/*",
+        "puglint"
+      ],
+      "matchDepTypes": [
+        "dependencies",
+        "devDependencies"
+      ],
+      "matchUpdateTypes": [
+        "minor",
+        "patch"
+      ],
+      "enabled": true,
+      "automerge": false
+    },
+    {
       "matchPackageNames": [
         "typescript"
       ],
@@ -32,5 +64,6 @@
       "automerge": true
     }
   ],
-  "dependencyDashboard": true
+  "dependencyDashboard": true,
+  "baseBranches": ["master", "update-renovate-config"]
 }


FisherSkyi:
## Update on copilot's review:
  1. Make `c-scroll-top-button` presentational only.
     In frontend/src/components/c-scroll-top-button.vue:
      - remove scrollContainerId, mounted, beforeUnmount, handleScroll, topFunction
      - add a boolean prop like `showBackToTop`
      - emit one event like `scroll-to-top` on click
      - change the root element from a plain div to button type="button"
      - replace the hard-coded id with a class such as .go-back-button
  2. Move scroll ownership to frontend/src/views/c-home.vue.
     c-home already owns both real scroll containers:
      - #summary-wrapper
      - #tabs-wrapper

     Add:
      - `showSummaryBackToTop`
      - `showTabsBackToTop`
      - Or to prevent duplicate method, we may use 
          `c-scroll-top-button(@scroll-to-top="scrollPaneToTop('summary')")`
          `c-scroll-top-button(@scroll-to-top="scrollPaneToTop('tabs')")`
          Then in home.vue:
          ```js
          methods: {
            scrollPaneToTop(pane: 'summary' | 'tabs') {
              const el = pane === 'summary'
                ? this.$refs.summaryWrapper
                : this.$refs.tabWrapper;
              (el as HTMLElement | undefined)?.scrollTo({ top: 0, behavior: 'smooth' });
            },
          }
          ```
      - scroll handlers for each wrapper
      - click handlers that scroll the corresponding wrapper to top
  3. Render the two button instances in c-home.vue, one per pane.
     Example idea:
      - one inside #summary-wrapper, bound to showSummaryBackToTop and scrollSummaryToTop
      - one inside #tabs-wrapper, bound to showTabsBackToTop and scrollTabsToTop
  4. Remove c-scroll-top-button usage from child views.
     Remove it from:
      - frontend/src/views/c-summary.vue
      - frontend/src/views/c-authorship.vue
      - frontend/src/views/c-global-file-browser.vue
  5. Update Cypress selectors.
     Since the button should no longer use #go-back-button, change tests to use something like:
      - #summary-wrapper .go-back-button
      - #tabs-wrapper .go-back-button

## Rationale:

  - Keeps the child reusable.
    c-scroll-top-button becomes a generic control that only says “user clicked me” and “show/hide me”.
  - Avoids container binding.
    No `scrollContainerId` and `document.getElementById(...)` inside the reusable child.
  - Makes event handling unambiguous.
    Both buttons can emit the same scroll-to-top event, because each instance is wired separately in c-home.vue.

frontend/src/components/c-scroll-top-button.vue
```diff
@@ -0,0 +1,49 @@
...
+  color: white;
+  cursor: pointer;
+  display: flex;
+  font-size: 20px;
+  height: 50px;
+  margin-left: auto;
+  padding: 0;
+  place-content: center;
+  place-items: center;
+  position: sticky;
```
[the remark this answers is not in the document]
FisherSkyi: From a practical perspective `fixed` will cause the left and right buttons overlap as they are both positioned relative to the viewpoint. But help and suggestions are welcome here as I don't really understand the mechanism of css layout.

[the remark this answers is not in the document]
ashmitahaldar: Regarding this issue, yes, `position: fixed` won't work here. `position: sticky` does technically work (the button sticks to the bottom of the visible scroll area since it's the last child in the container), but it stays in document flow and can cause layout shifts.

The cleaner fix might be `position: absolute` scoped to each panel:
  1. Add `position: relative` to `#summary-wrapper` in `style.scss`).
  2. Change the button to `position: absolute; bottom: 20px; right: 20px` and remove `margin-left: auto`.
  3. Make sure `c-scroll-top-button`in the tabs panel out of `.tab-pane/.tab-content` and make it a direct child of `#tabs-wrapper`, so it floats over the panel content rather than scrolling with it. The summary panel button is already placed correctly.
This keeps each button scoped to its own panel, out of document flow, and visible at any scroll depth regardless of content length.

Can try this out though and see if it works!

frontend/cypress/tests/chartView/chartView_scrollButton.cy.js
```diff
@@ -0,0 +1,25 @@
+describe('scroll button', () => {
+  it('button exists', () => {
+    cy.get('#summary-wrapper .go-back-button').should('exist');
+  });
+
+  it('button is not shown when only scroll 100 pixels', () => {
+    cy.get('#summary-wrapper .summary-charts').should('exist');
+    cy.get('#summary-wrapper').scrollTo(0, 100);
+    cy.get('#summary-wrapper .go-back-button').should('not.be.visible');
```
ashmitahaldar, line 9: The tests in `globalBrowserView_scrollButton.cy.js` scroll back to (0, 0) at the end of each test, why not in the tests in this file?

frontend/cypress/tests/globalBrowserView/globalBrowserView_scrollButton.cy.js
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  it('button is not shown when only scroll 100 pixels', () => {
+    cy.get('#tabs-wrapper').scrollTo(0, 100);
+    cy.get('#tabs-wrapper .go-back-button').should('not.be.visible');
+    cy.get('#tabs-wrapper').scrollTo(0, 0);
+  });
+
+  it('shows the button when scroll more than 200 pixels', () => {
+    cy.get('#tabs-wrapper').scrollTo(0, 300).trigger('scroll');
```
ashmitahaldar: What does this `.trigger('scroll')` do? Why does test 2 and 3 use it but not test 1? Just a bit confused haha, would be good to understand.

frontend/src/components/c-scroll-top-button.vue
```diff
@@ -0,0 +1,52 @@
...
+    showBackToTop: {
+      type: Boolean,
+      default: false,
+    },
+  },
+  emits: ['scroll-to-top'],
+});
+</script>
+
+<style lang='scss'>
```
ashmitahaldar: Should you add a `scoped` here within the tag to avoid the styles from leaking globally?

frontend/src/components/c-scroll-top-button.vue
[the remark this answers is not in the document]
jonasongg: agree with ashmita here !

jonasongg (commented):
lgtm other than the small comment!

frontend/src/components/c-scroll-top-button.vue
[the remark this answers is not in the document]
FisherSkyi: The behavior with proposed fix: the button will scroll along with the segment, which is not desirable. Right now I think we can keep the `sticky`.

frontend/cypress/tests/chartView/chartView_scrollButton.cy.js
FisherSkyi: Added

frontend/cypress/tests/globalBrowserView/globalBrowserView_scrollButton.cy.js
FisherSkyi: [trigger](https://docs.cypress.io/api/commands/trigger) will emit the event "scroll".

`scroll` is an event. It's handler is in c-home.vue line 48 `#summary-wrapper(ref="summaryWrapper", @scroll="handleSummaryScroll")`

`trigger` is needed, as `cy.scrollTo()` changes the element's `scrollTop` property in the DOM. But in test environments (on CI/CD env specifically), the native scroll event does not fire automatically

ashmitahaldar: @FisherSkyi Thanks for clarifying! Do you need to include it then in the other tests or no need?

FisherSkyi:
> Generally LGTM! Would just like you to check if `.trigger()` is needed on other tests, otherwise looks good to go.

I double checked and it seems `scrollTo` along is enough. Chaining with `.trigger()` would emit scroll event twice, which is not desirable.

Also while exploring I found another separate issue #2574 but it should be addressed in separate pr.

frontend/src/views/c-home.vue
```diff
@@ -191,6 +216,26 @@ const home = defineComponent({
...
+
+    scrollSummaryToTop(): void {
+      const el = this.$refs.summaryWrapper as HTMLElement | undefined;
+      el?.scrollTo({ top: 0, behavior: 'smooth' });
+    },
+
+    scrollTabsToTop(): void {
+      const el = this.$refs.tabWrapper as HTMLElement | undefined;
+      el?.scrollTo({ top: 0, behavior: 'smooth' });
+    },
```
CYX22222003, line 238: Scrolling logic appears to be sound.

CYX22222003 (approved):
LGTM!

PR #2557, opened 2026-03-12
[#2528] Add YAML configuration file wizard

### LINKED ISSUE #2528: Add interactive configuration wizard for generating config files
A configuration file generator tool that provides a user-friendly interface for creating RepoSense configuration files (author-config.csv, repo-config.csv, group-config.csv, report-config.yaml) through interactive form fields, with automatic file generation and placement in the appropriate directories.

Currently, new users must manually create and edit CSV/YAML configuration files, which has several pain points:
- Steep learning curve: Users need to understand the exact schema, field names, and formatting requirements for each config file type before they can generate their first report
- Syntax errors: Manual editing is error-prone (typos, incorrect delimiters, YAML indentation errors, missing required fields)
- Documentation dependency: Users must constantly reference documentation to understand valid values for fields like file-formats, ignore-glob-list, or since/until date formats
- Time-consuming setup: Setting up configs for multiple repositories or authors requires repetitive manual data entry
- Validation gaps: No immediate feedback on whether config values are valid until running RepoSense, leading to trial-and-error cycles
This creates friction for adoption, especially for first-time users or teams trying to quickly set up RepoSense for their projects.

A standalone configuration wizard (web-based GUI) that:
1. Interactive prompts/form fields for each configuration type:
  - Repo config: repository location, branch, file formats, ignore patterns, etc.
  - Author config: author names, aliases, ignore patterns
  - Group config: group definitions and member assignments
  - Report config: title, repository display options, date ranges
2. Built-in validation with real-time feedback:
  - Check date formats, glob patterns, branch names
  - Validate repo URLs/paths before generating config
  - Warn about common mistakes (e.g., conflicting ignore patterns)
3. File generation & placement:
  - Generate properly formatted CSV/YAML files
  - Option to place files directly in workspace config directory
  - Prompt user with next steps (e.g., "Run java -jar RepoSense.jar --config ./config")
4. Edit existing configs:
  - Import existing config files for editing
  - Preserve comments and custom formatting where possible

### Additional context
- There is an existing issue #1715 and corresponding PR #1950 that attempts to implement a CLI-based RepoSense wizard.
- There has been some exploration on simplifying the config file setup process through an interactive GUI (#2199), but it involved provisioning cloud resources. Our solution would involve packaging the tool/wizard alongside RepoSense.
- A previous attempt at a similar tool #1430 that aims to walk users through creating their first report.

ashmitahaldar:
Fixes #2528

### Proposed commit message
```
Creating report-config.yaml currently requires users to manually
edit files with exact schema knowledge, which is error-prone and
has a steep learning curve for new users.

Add an interactive web-based Configuration Wizard that guides
users through creating a report-config.yaml file via a multi-step
form with live YAML preview, launched via:

java -jar RepoSense.jar --config-wizard
```

### Other information
The wizard opens at http://localhost:9000/config-wizard (browser is opened automatically) and walks users through four steps:
  1. Report Settings — report title
  2. Repos & Branches — repo URL with location validation, per-branch config (blurb, ignore globs, ignore authors, file size limit, date ranges), and per-branch author management
  3. Groups (optional) — per-repo file grouping with glob patterns
  4. Review & Generate — full config validation, summary of repos/branches/authors/groups, and generation to
  generated-configs/report-config.yaml

#### Backend changes (src/main/java/reposense/wizard/):
- ConfigWizardServer — embedded HTTP server (net.freeutils.httpserver) serving the Vue.js SPA and REST API:
    - GET /api/config — server status
    - POST /api/validate — validates a repo location (local path or remote URL)
    - POST /api/validate-glob — validates glob patterns
    - POST /api/validate-config — full config validation via ReportConfigYamlParser
    - POST /api/preview — returns YAML preview string
    - POST /api/generate — writes generated-configs/report-config.yaml
    - POST /api/quit — shuts down the server
- ConfigWizardService — service layer handling validation and config generation logic
- ConfigFileWriter — writes report-config.yaml via Jackson (nulls excluded)
- ArgsParser — adds --config-wizard flag (incompatible with all other flags)
- RepoSense.main() — routes to ConfigWizardServer.startWizard(9000) when flag is set

#### Frontend changes (frontend/config-wizard/):
- 4-step wizard with a typed Vuex-style store mirroring the report-config.yaml hierarchy (WizardConfig, Repo, Branch, Author, Group interfaces)
- Two-pane layout: form on the left, live YAML preview on the right with a draggable resizer
- Author fields use correct YAML keys matching ReportAuthorDetails @JsonProperty annotations (author-git-host-id, author-emails, author-display-name, author-git-author-name)
- Quit button in header and success screen on Step 4, both calling /api/quit to stop the server cleanly

### DIFF
--- a/docs/ug/cli.md
+++ b/docs/ug/cli.md
@@ -299,12 +299,12 @@ Cannot be used with `--last-modified-date`. This may result in an incorrect last
 
 ### `--text`, `-T`
 
-**`--text`**: Refreshes text content (intro.md/title.md, repo-blurbs.md and author-blurbs.md) without reanalyzing repository data.
+**`--text`**: Refreshes text content (intro.md, repo-blurbs.md and author-blurbs.md) without reanalyzing repository data.
 * Alias: `-T` (uppercase T)
 
 <box type="info" seamless>
 
-* This flag is used to update only the text content (intro.md/title.md, repo-blurbs.md and author-blurbs.md) of the report. The new report will be generated with the existing data from the previous report.
+* This flag is used to update only the text content (intro.md, repo-blurbs.md and author-blurbs.md) of the report. The new report will be generated with the existing data from the previous report.
 * Ensure that there is an existing valid report generated before using this flag.
 * Cannot be used with any other flags except from `--view`, `--assets` and `--config`.
 </box>
--- a/build.gradle
+++ b/build.gradle
@@ -353,7 +353,6 @@ def syncFrontendPublic = tasks.register('syncFrontendPublic', Sync) {
         include 'index.html'
         include 'favicon.ico'
         include 'intro.md'
-        include 'title.md'
         include 'logo.png'
     }
 }
--- a/docs/ug/customizingReports.md
+++ b/docs/ug/customizingReports.md
@@ -89,9 +89,9 @@ To ensure that their PRs are correct, you can use [Netlify _deploy previews_](ht
 There is a default `favicon.ico` file provided in the template zip folder. If you want to self-configure the `favicon.ico`, please ensure that the file is valid and has the file name of `favicon.ico`. This icon will appear in the browser tab when your report is viewed. Copy or move your `favicon.ico file` into the `assets` folder of the config directory.
 
 ##### Add an intro
-Add the introductory section of the dashboard by creating an `intro.md` file in the config directory. Existing `title.md` files still work for now, but will be deprecated in favour of `intro.md`.
+Add the introductory section of the dashboard by creating an `intro.md` file in the config directory.
 
-The intro can render a combination of Markdown/HTML and plaintext ([example](https://github.com/reposense/RepoSense/blob/master/docs/ug/title.md)), and will appear on the top of the left panel as shown below:
+The intro can render a combination of Markdown/HTML and plaintext, and will appear on the top of the left panel as shown below:
 ![Intro Component Example](../images/title-example.png)
 
 Do note that the width of the intro is bound by the width of the left panel.
--- a/frontend/src/components/c-intro.vue
+++ b/frontend/src/components/c-intro.vue
@@ -8,7 +8,6 @@ import { defineComponent } from "vue";
 import CMarkdownChunk from "./c-markdown-chunk.vue";
 
 const INTRO_FILE_PATH = "intro.md";
-const LEGACY_TITLE_FILE_PATH = "title.md";
 
 export default defineComponent({
   name: "c-intro",
@@ -20,19 +19,7 @@ export default defineComponent({
   },
   async beforeMount() {
     try {
-      const introText = await this.fetchMarkdown(INTRO_FILE_PATH);
-      if (introText) {
-        this.markdownText = introText;
-        return;
-      }
-
-      const legacyText = await this.fetchMarkdown(LEGACY_TITLE_FILE_PATH);
-      if (legacyText) {
-        console.warn(
-          "Detected deprecated title.md. Please rename it to intro.md.",
-        );
-      }
-      this.markdownText = legacyText;
+      this.markdownText = await this.fetchMarkdown(INTRO_FILE_PATH);
     } catch (error) {
       this.markdownText = (error as Error).toString();
     }


docs/ug/configWizard.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Troubleshooting
+
+**The browser did not open automatically.**
+
+Manually navigate to `http://localhost:9000/config-wizard`.
+
+**Port 9000 is already in use.**
+
+Stop the process using port 9000 and try again. Future versions of the wizard may allow configuring the port via a flag.
```
lyuanww, line 179: Remember to open an issue after merge about this future enhancement so this doesn't get lost

src/main/java/reposense/parser/ArgsParser.java
```diff
@@ -327,6 +336,17 @@ public static CliArguments parse(String[] args) throws HelpScreenException, Pars
...
         boolean shouldPerformFreshCloning = results.get(FRESH_CLONING_FLAG[0]);
         boolean shouldRefreshOnlyText = results.get(REFRESH_ONLY_TEXT_FLAG[0]);
+        boolean isConfigWizard = results.get(CONFIG_WIZARD_FLAGS[0]);
+
+        if (isConfigWizard) {
+            boolean hasOtherFlags = Arrays.stream(args)
+                    .filter(arg -> arg.startsWith("--"))
+                    .anyMatch(arg -> !arg.equals(CONFIG_WIZARD_FLAGS[0]));
+            if (hasOtherFlags) {
+                throw new ParseException(MESSAGE_CONFIG_WIZARD_INCOMPATIBLE_FLAGS);
```
lyuanww: After some thought, I wonder it's ok to not enforce a hard check on all the flags. Does the config wizard work without it? It might clash only with `--view` flag.

Looking at the `--version ` and `--help` documentation, there seems to have a bit of a clash.

ashmitahaldar:
> the YAML preview only refresh after clicking the 'Next' button, or is it intended to be live?

Hi @lyuanww, yes the YAML preview only refreshes/updates after clicking the 'Next' button. This is the case because the preview refers to `store.config` when it generates. The form fields write to local component state during editing, and store.config is updated only when the user clicks 'Next'.

docs/ug/configWizard.md
ashmitahaldar: Will do! Thanks for pointing this out 👍

src/main/java/reposense/parser/ArgsParser.java
ashmitahaldar: @lyuanww Actually even without the hard check, including other flags along with `--config-wizard` doesn't pose an issue, as the program launches the config wizard and returns before any other parsed value is used.
The `--version` and `--help` is a good observation though, as running with either of those flags will bypass the config wizard flag. I will add this behaviour in the documentation for config wizard.
What do you suggest as the best approach? Do you think eliminating the hard check is best, or should we display a warning if other flags are also mentioned?

frontend/config-wizard/App.vue
CYX22222003, line 1: @ashmitahaldar for the frontend, we strongly recommend to add some cypress tests, at least the happy path test to make it easier for future develpers to review and understand the implementation.

ashmitahaldar: @CYX22222003 Added in the recent commit!

src/main/java/reposense/parser/ArgsParser.java
ashmitahaldar: @lyuanww Since the config wizard is a standalone mode that ignores all analysis/report flags anyway, the only genuine logical conflict is with `--view`, because a user passing both might reasonably expect the view server to launch instead. 
I pushed a commit that updated the code to:
  - Throw a ParseException only for `--config-wizard` used with `--view` (using the already-parsed `reportFolderPath` to handle both `--view` and `-v`)
  - Emit a warning log for any other flags that will be silently ignored
 Please have a look!

lyuanww: @ashmitahaldar I think it's a good call for ParseException!
But I think for the warnings, I think it's ok if we remove that part of code and document that behaviour in the documentation since your code already safely ignores the other flags? To keep the codebase cleaner

docs/ug/cli.md
```diff
@@ -81,6 +81,23 @@ Refer to the [Advanced Customization](./customizingReports.html#advanced-customi
...
+
+**`--config-wizard`**: Launches an interactive browser-based wizard for generating a `report-config.yaml` file without writing YAML by hand.
+* Default: the wizard is not launched
+* Example: `java -jar RepoSense.jar --config-wizard`
+
+<box type="info" seamless>
+
+* Starting the wizard opens a local web server on port 9000 and automatically launches `http://localhost:9000/config-wizard` in your default browser.
+* The generated file is written to `./generated-configs/report-config.yaml` in the current working directory.
+* Cannot be used with any other flags.
```
lyuanww: > The --version and --help is a good observation though, as running with either of those flags will bypass the config wizard flag. I will add this behaviour in the documentation for config wizard.

Yes. I agree. I think it would best to preserve the behaviour of the older flags if its not too difficult to do so

src/main/java/reposense/parser/ArgsParser.java
ashmitahaldar: Done!

docs/ug/cli.md
ashmitahaldar: @lyuanww Have implemented the change requested in the other thread (about removing the explicit other flags check) and have clarified behaviour with other flags in the documentation!

The application already preserves original behaviour of `--version` and `--help` since they are handled by argparse4j's `HelpArgumentAction` and `VersionArgumentAction` respectively, which trigger during `parser.parseArgs(args)` (line 314 of `ArgsParser.java`). They print their output and exit the JVM right there, before execution ever reaches the config wizard check.

CYX22222003 (approved):
LGTM

PR #2624, opened 2026-09-02
[#2622] Hide bucket time in ramp tooltips

### LINKED ISSUE #2622: Ramp tooltip timestamp can be misleading for daily commit buckets
### Problem

Ramp tooltips currently show dates in a datetime format such as:

```text
[2019-07-24T00:00:00] Commit message: +0 -9 lines
```

This can look like the actual commit timestamp, but it is not. The frontend tooltip uses `slice.date`, which represents the daily contribution bucket that the commit belongs to.

On the backend, commits are grouped by day. The bucket date is normalized to the start of the day:

```java
current.withHour(0).withMinute(0).withSecond(0).withNano(0)
```

The actual commit time is parsed during commit analysis, but it is stored in a transient field in `CommitResult` and is not serialized into `commits.json`.

### Why This Is Misleading

Users may interpret `T00:00:00` as meaning the commit was made exactly at midnight, when it actually means the commit belongs to that daily bucket.

### Possible Improvement

For ramp tooltips, display only the date portion of the bucket date, for example:

```text
[2019-07-24] Commit message: +0 -9 lines
```

This avoids implying that RepoSense is showing the exact commit time.

### Scope

This can likely be handled as a frontend display-only change in `frontend/src/components/c-ramp.vue`, without changing the underlying report JSON format or date filtering logic.

CYX22222003:
Fixes #2622

### Proposed commit message
```
Hide bucket time from ramp tooltips.

Format ramp tooltip dates as date-only values, because the timestamp
shown there represents the daily bucket start rather than an actual
commit timestamp.
```

### Other information
This is a display-only frontend change. The underlying date values used
for filtering, ramp positioning, report generation, and JSON output are
left unchanged.

Tested with:
- `.\gradlew.bat testFrontend -Pci`
- `.\gradlew.bat test --tests "reposense.commits.*" --tests "reposense.parser.LocalDateTimeParserTest"`

### DIFF
--- a/frontend/cypress/tests/chartView/chartView_mergeCommits.cy.js
+++ b/frontend/cypress/tests/chartView/chartView_mergeCommits.cy.js
@@ -1,7 +1,7 @@
 describe('include merge commits in chart view', () => {
   it('show merge commits in summary chart', () => {
     // ramp chart should have merge commit slices
-    cy.get('[title="[2023-03-03T00:00:00] Merge branch \'new-branch\' into cypress: +0 -0 lines "]')
+    cy.get('[title="[2023-03-03] Merge branch \'new-branch\' into cypress: +0 -0 lines "]')
       .should('have.class', 'ramp__slice')
       .should('exist');
   });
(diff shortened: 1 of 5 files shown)


damithc (commented):
Thanks @CYX22222003 for the PR. Based on changes to the test files, this PR seems to achieve the intended outcome. No issues from my side.

PR #2625, opened 2026-09-02
[#2621] Document delete-only ramp color

### LINKED ISSUE #2621: UG should document special ramp color for delete-only contributions
### Problem

The User Guide explains the general ramp chart color behavior, but it does not mention the special color used for delete-only contributions.

From the current frontend implementation in `frontend/src/components/c-ramp.vue`, a ramp is treated as a delete-only contribution when:

```ts
insertions === 0 && deletions > 0
```

In that case, `getRampColor` / `getSliceColor` returns `'-deletes'`, and the ramp is rendered using `ramp__slice--color-deletes`, which is red.

This can confuse users who are trying to understand what ramp colors represent, because red is not part of the normal rotating ramp color scheme.

### Expected Documentation

The User Guide should clarify that:

- Normal ramps use the regular rotating time-period colors.
- Delete-only ramps use a separate red color.
- Mixed contributions, such as `+10 -5`, still use the normal ramp color, not the delete-only color.

### Suggested Location

This can be documented in the Ramp charts section of the User Guide, near the existing explanation of ramp color behavior.

CYX22222003:
Fixes #2621

### Proposed commit message
```
Document delete-only ramp color.

Clarify that ramps containing only deleted lines are shown in red, while
ramps with both added and deleted lines continue to use the normal
time-period colors.
```

### Other information
Docs-only change. Verified with `git diff --check`.

### DIFF
--- a/docs/ug/usingReports.md
+++ b/docs/ug/usingReports.md
@@ -69,6 +69,7 @@ The `Chart panel` (an example is shown above) contains a series of _ramp chart_ 
   * At the commit-level of granularity, it links to the commit in the remote repository.
   * At higher levels of granularity, it displays the list of commits in the Code Panel.
 * To make the comparison between two authors easier, the color of the ramps that represent different authors' contributions at the same time period are the same.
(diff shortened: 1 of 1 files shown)


damithc (approved):
Thanks for the quick fix, @CYX22222003

